Issue #556, opened 2019-02-10
Problem in #1519 Paper UI


udo1toni:
Next Issue :)

When Installing a fresh OH2.5 #1519, in Paper UI when calling Configuration->System I get an Error 500 - Internal Server Error.

I did an update, when getting this error I used apt remove and apt purge, but after installing the current nightly, I still get this internal error.

kaikreuzer:
I am so damn stupid - #557 didn't fix this issue :-/

PR #557, opened 2019-02-10
handle arrays for pids

kaikreuzer:
Fixes https://github.com/openhab/openhab-core/issues/556
Fixes https://github.com/eclipse/smarthome/issues/6710

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/service/ConfigurableServiceResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/service/ConfigurableServiceResource.java
@@ -310,10 +310,16 @@ public class ConfigurableServiceResource implements RESTResource {
     private String getServiceId(ServiceReference<?> serviceReference) {
         Object pid = serviceReference.getProperty(Constants.SERVICE_PID);
         if (pid != null) {
-            return (String) pid;
-        } else {
-            return (String) serviceReference.getProperty(ComponentConstants.COMPONENT_NAME);
+            if (pid instanceof String) {
+                return (String) pid;
+            } else {
+                String[] pidArray = (String[]) pid;
+                if (pidArray.length > 0) {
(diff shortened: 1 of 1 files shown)


kaikreuzer:
PR builds now all green.

bundles/org.openhab.core.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/service/ConfigurableServiceResource.java
```diff
@@ -310,10 +310,16 @@ private String getConfigDescriptionByFactoryPid(String factoryPid) {
...
         Object pid = serviceReference.getProperty(Constants.SERVICE_PID);
         if (pid != null) {
-            return (String) pid;
-        } else {
-            return (String) serviceReference.getProperty(ComponentConstants.COMPONENT_NAME);
+            if (pid instanceof String) {
+                return (String) pid;
+            } else {
+                String[] pidArray = (String[]) pid;
+                if (pidArray.length > 0) {
```
maggu2810: Wouldn't it make sense, if we return the first one only and ignore all the others, to check if they are equal and if they differ log a warning that contains some information (component name, services pids, the used one).

kaikreuzer: As mentioned in https://github.com/eclipse/smarthome/issues/6710#issuecomment-449592856, I do not at all understand why there should ever be different values. So for a start, this fixes the issue at hand.

kaikreuzer: FTR, I also do not have a working IDE right now, so I cannot test changes thoroughly - only trying to fix the most critical issues asap, so that the distro gets usable again...

Question: While in fire-fighting mode, could we allow self-merging for fixing regressions to speed up things and address comments later on?

PR #565, opened 2019-02-11
[automation] Adapt input type of ItemCommandAction

davidgraeff:
Considering the bundle name has changed recently, changing this type should not be a breaking change

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/resources/ESH-INF/automation/moduletypes/ItemActions.json
+++ b/bundles/org.openhab.core.automation/src/main/resources/ESH-INF/automation/moduletypes/ItemActions.json
@@ -51,7 +51,7 @@
      "inputs":[
       {
         "name":"command",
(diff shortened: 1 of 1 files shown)


maggu2810:
Sorry, I had to remove that merge.

The package (not bundle) name has not been changed.
Why should this reference been changed?

davidgraeff:
Oh, ok. So instead the old package names should be used? Because at least in the other PR I have also fixed types. For example from "command" to the full class name.

maggu2810:
IMHO the package name of the currently used code base should be used:
https://github.com/openhab/openhab-core/blob/a37ccea/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/types/Command.java#L13

davidgraeff:
Alright, I'll adapt the other PR. Thanks for the hint, haven't seen (or thought about) that

PR #566, opened 2019-02-11
[automation] Fix types for ItemCommandTrigger

davidgraeff:
Considering the bundle name has changed recently, changing those types should not be a breaking change.

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/resources/ESH-INF/automation/moduletypes/ItemTriggers.json
+++ b/bundles/org.openhab.core.automation/src/main/resources/ESH-INF/automation/moduletypes/ItemTriggers.json
@@ -51,13 +51,13 @@
          "outputs":[
            {
                "name":"command",
-               "type":"command",
+               "type":"org.openhab.core.types.Command",
                "description":"the received command",
                "label":"Command"
            },
            {
                "name":"event",
-               "type":"org.eclipse.smarthome.core.events.Event",
+               "type":"org.openhab.core.events.Event",
                "label":"Event",
                "description":"The events which was sent.",
                "reference":"event"
@@ -115,13 +115,13 @@
          "outputs":[
             {
                 "name":"state",
-                "type":"state",
+                "type":"org.openhab.core.types.State",
                 "description":"the item state",
                 "label":"State"
             },
             {
                 "name":"event",
-                "type":"org.eclipse.smarthome.core.events.Event",
+                "type":"org.openhab.core.events.Event",
                 "label":"Event",
                 "description":"The events which was sent.",
                 "reference":"event"
@@ -213,19 +213,19 @@
          "outputs":[
             {
                "name":"newState",
-               "type":"state",
+               "type":"org.openhab.core.types.State",
                "description":"the new item state",
                "label":"New State"
             },
             {
                "name":"oldState",
-               "type":"state",
+               "type":"org.openhab.core.types.State",
                "description":"the old item state",
                "label":"Old State"
             },
             {
                "name":"event",
-               "type":"org.eclipse.smarthome.core.events.Event",
+               "type":"org.openhab.core.events.Event",
                "label":"Event",
                "description":"The events which was sent.",
                "reference":"event"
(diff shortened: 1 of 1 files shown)


maggu2810:
Sorry, I had to remove that merge.

The bundle name has been changed.
The package name of that classes has not been changed.
The FQCN should still be the same.

Only the package name of the automation bundle has been changed.

davidgraeff:
Also related to #565

davidgraeff:
I have adapted this PR

davidgraeff:
> Is the change necessary because something stopped working and continues to work after that change? Or why do you identify that place and would like to change it?

I don't know if you have heard about it, but I proposed a Paper UI NG design study, that includes a Rule editor. And I noticed that I couldn't connect "command" outputs to FQCN "command" inputs.

That's very likely because of sloppiness and because nobody has used inputs/outputs before. I mean you can manually link any input to any output in a rule json file, but a (graphical) tool requires the types to be the same.

kaikreuzer:
> @kaikreuzer Can you comment? Has the non fully qualified version been chosen by intention? Should we keep it to allow changing the package name easily?

The "type" field was originally never meant to be a FQCN, but rather a value from an (yet to be defined) enumeration. i.e. the type is "a command" or "a state" in this case.
Using FQCN nicely shows the fragility here - if we use them, all created rules will break once we change the namespace, which is not good.

davidgraeff:
So is the fix instead to change all types to "Event", "Command" and "State"?

kaikreuzer:
to be discussed, I guess...

davidgraeff:
Ok, that's reasonable. But can we in the meantime have a fix so that I can link my inputs/outputs in my rules? That would be great. So either FQCN or the short variant, I don't mind, but it should be unified. It is later on easier to `sed` over the files as well if they have a unified name.

openhab-5iver:
> From the eclipse rules docs about types

The [annotations](https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/type/Output.java#L30) say it can be anything. Same for [Input](https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/type/Input.java#L31).

Using a FQCN is a nice standard for keeping them consistent, but I think it's a little misleading, since people may think it _needs_ to be a FQCN. My preference would be nonFQCN. @davidgraeff , you just need them to be consistent, right?

I'm not so sure this would be a breaking change, unifying either way, since this info isn't stored in the JSON rules, and the rules are recreated after each OH restart. If it is, we really should get #565 reverted (why wasn't it?).

I have a number of rules built out that I use for regression tests, and I'd like to do some testing before it's merged... but I need to get my IDE working. :frowning:

davidgraeff:
> we really should get #565 reverted

It is reverted, at least it's not in the branch history

openhab-5iver:
Weird. I see a merge, but no revert. The changes show org.openhab.core.types.Command, but the actual file shows it back the way it was. Must have been reverted.

kaikreuzer:
> My preference would be nonFQCN

So is mine - so my proposal is to define "state" and "command" as the first elements for the "to-be-defined" enumeration of valid types. Which means that we do not have to change anything in the code base or have we?

davidgraeff:
I'd go through all files and will unify this then. There is a third type that I want to throw into the game: "event" (which is currently FQCN everywhere)

kaikreuzer:
I'm not convinced. "event" can be anything and it does not say anything about what comes along there. For "command" and "state" we can actually expect a String as a content and we would know what to do with it. For "event" this won't work because you actually expect a Java class as a structure for it (and thus a FQCN makes some sense).

davidgraeff:
> For "event" this won't work because you actually expect a Java class as a structure for it

Agreed. There is still "eclipse.smarthome" in the FQCN though. As long as future module-types are allowed to use the same event type, I do not see an Issue.

kaikreuzer:
> As long as future module-types are allowed to use the same event type, I do not see an Issue.

Not anymore for an openHAB 3, so this will have to be part of a migration. But not having to consider states and commands for migration is at least already something...

kaikreuzer:
So is the agreed solution now that we simply close this PR?

maggu2810:
> So is the agreed solution now that we simply close this PR?

It depends if

> I'd go through all files and will unify this then.

has been about the "openHAB Core" files or the "Paper UI NG" ones and if @davidgraeff finished already (if it has been about files in that repo).

Or miss I something?

davidgraeff:
Can be closed. There are no changes in this file required according to the consensus we had above.

Issue #513, opened 2019-02-01
Create Oomph Setup file


kaikreuzer:
As we are asking more and more people to do PRs against this repo now, we should provide them with a proper IDE setup (and not just refer to [the notes](https://github.com/openhab/openhab-core/blob/master/notes/notes.md#ide).

@maggu2810 I assume we should maintain the setup file here in the repo (and not add it to the "addons IDE", which is in openhab-distro). If so, could you adapt your setup file from the notes and add it here? We could then ask the Oomph maintainers to include it in the official Eclipse Installer.

maggu2810:
What about the one I provided with the migration PR already?
https://github.com/openhab/openhab-core/blob/master/tools/oomph/openHAB-Core.setup

kaikreuzer:
This one I did not see yet 😄.

> (We should remove / fix that part from / of the notes)

Yes!

openhab-5iver:
I have not been able to get an IDE setup where I can debug openhab-core. After Eclipse starts, I run into this...

```
Could not calculate build plan: Plugin org.apache.maven.plugins:maven-jar-plugin:3.0.2 or one of its dependencies could not be resolved: Failed to read artifact descriptor for org.apache.maven.plugins:maven-jar-plugin:jar:3.0.2
Plugin org.apache.maven.plugins:maven-jar-plugin:3.0.2 or one of its dependencies could not be resolved: Failed to read artifact descriptor for org.apache.maven.plugins:maven-jar-plugin:jar:3.0.2
```
![image](https://user-images.githubusercontent.com/21967853/52636457-a2a3dd00-2e9a-11e9-9260-6da32eaa0376.png)

Is there a particular version of Maven required, or some plugin that needs to be manually installed? Continuing with the steps, where Eclipse needed a restart, did not help. I have an older IDE setup, where I have several changes that I'd like to debug and submit, but unfortunately, I can't get Paper UI to work in this setup. I would _really_ appreciate some help with this!

![image](https://user-images.githubusercontent.com/21967853/52637166-80ab5a00-2e9c-11e9-9220-c1846c17232d.png)

maggu2810:
* Which IDE version has been chosen in the Eclipse Installer?
* If you close the Eclipse IDE and enter the directory of the git clone of `openhab-core`, can you try to enter `mvn -V clean install` and post the output

kaikreuzer:
[...]

PR #570, opened 2019-02-12
added a remark about compiler compliance level

afuechsel:
Signed-off-by: Andre Fuechsel <afuechsel@gmail.com>

### DIFF
--- a/notes/notes.md
+++ b/notes/notes.md
@@ -31,9 +31,10 @@
 
 ### Start a demo runtime
 
-* Open the project "org.eclispe.smarthome.demo.app"
+* Open the project "org.openhab.core.demo.app"
 * Open the file "app.bndrun"
 * Press "resolve" to resolve the "run requirements" and ensure that all requirements are able to satisfied.
+* Make sure, that the Compiler Compliance Level in Eclipse matches the installed JDK: set it to 8, if there is Java 8 installed
 * Press "Debug OSGi"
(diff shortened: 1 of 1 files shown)


notes/notes.md
```diff
@@ -31,9 +31,10 @@
 
 ### Start a demo runtime
 
-* Open the project "org.eclispe.smarthome.demo.app"
+* Open the project "org.openhab.core.demo.app"
 * Open the file "app.bndrun"
 * Press "resolve" to resolve the "run requirements" and ensure that all requirements are able to satisfied.
+* Make sure, that the Compiler Compliance Level in Eclipse matches the installed JDK: set it to 8, if there is Java 8 installed
```
maggu2810: AFAIK openHAB Core -- as it depends on Xtext < 2.17 -- does not support any Java version > 8 at all.

afuechsel: Yes, was just a reminder, because for me, installed only Java 8, the runtime did not start, because Eclipse has set compiler compliance level to 10... However, @kaikreuzer mentioned a better fix in the setup file: 
``` 
<setupTask
          xsi:type="setup:PreferenceTask"
          key="/instance/org.eclipse.jdt.core/org.eclipse.jdt.core.compiler.compliance"
          value="1.8"/>
```

kaikreuzer:
With [this discussion](https://github.com/openhab/openhab-core/issues/513#issuecomment-463616034), I think we can close this PR and rather remove the notes file altogether.

Issue #553, opened 2019-02-09
OH #1518: issues with "now" implementations


nesys:
Hi team,

after updating to #1518 I've encountered issues in all my rules where I implement solutions with persistence (eg. lastUpdate.isAfter(now.minusSeconds(6)) or averageSince(now.minusMinutes(2)) )

Let me share an example:

2019-02-09 12:14:43.879 [INFO ] [el.core.internal.ModelRepositoryImpl] - Validation issues found in configuration model 'presence.rules', using it anyway:
The method createTimer(Object, Procedure0) from the type ScriptExecution refers to the missing type Object
The method lastUpdate(Item) from the type PersistenceExtensions refers to the missing type Object
The method lastUpdate(Item) from the type PersistenceExtensions refers to the missing type Object
The method lastUpdate(Item) from the type PersistenceExtensions refers to the missing type Object

My rule is:

```
rule "Welcome home my dear"
when
    Member of gPhones changed to ON
then
    if(cameHomeTimer !== null) {
        return; // do nothing if there is already a Timer
    }
    logInfo("test", "Rule triggered by " + triggeringItem.name + " at " + now.toString)
    cameHomeTimer = createTimer(now.plusSeconds(5), [ |

        val cameHome = gPhones.members.filter[ p | p.state == ON && (p.name == triggeringItem.name || (p.lastUpdate !== null && p.lastUpdate.isAfter(now.minusSeconds(6)))) ].map[ name.replace("_Phone_Home_Unifi_online", "") ]

        gPhones.members.forEach[ p | logInfo("test", "Values for " + p.name + "\n" + "State = " + p.state.toString + "\n" + "LastUpdate = " + p.lastUpdate.toString)]

        var message = "Benvenuti a casa, "

        switch(cameHome.size) {
            case 0: {
                logError("came home", "Unexpected number of phones: " + cameHome.size)
                return;
            }
            case cameHome.size == 1 && cameHome.get(0) == "Andrea": message = "Benvenuto a casa, Andrea"
            case cameHome.size == 1 && cameHome.get(0) == "Cate": message = "Benvenuta a casa, Caterina"
            case cameHome.size == 1 && cameHome.get(0) == "Sofia": message = "Benvenuta a casa, Sofia"
            case cameHome.size == 1 && cameHome.get(0) == "Edo": message = "Benvenuto a casa, Edoardo"
            default: {
                message = message + cameHome.members.reduce[ String list, String name | list + ", " + name ]
[...]

PR #572, opened 2019-02-12
ensure rules can access joda time packages

maggu2810:
Fixes: https://github.com/openhab/openhab-core/issues/553

### DIFF
--- a/bundles/org.openhab.core.model.rule/bnd.bnd
+++ b/bundles/org.openhab.core.model.rule/bnd.bnd
@@ -34,7 +34,8 @@ Import-Package: \
  org.apache.*,\
  org.eclipse.jdt.annotation;resolution:=optional,\
  org.eclipse.xtext.xbase.lib,\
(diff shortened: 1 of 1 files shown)


kaikreuzer:
Question: Won't we need the same fix for https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.model.script/bnd.bnd#L48?

maggu2810:
The JAR file that are generated on my machine includes already that packages for the script models, as they are used by the code.
Are they missing on yours?

kaikreuzer:
I didn't check it, that's why I was asking - the code looked similar, so I thought it might have the same problem.
If there is no problem, even better!

maggu2810:
That's instructions for bnd.
The difference is that script is using the Joda packages in the code, the rule doesn't.
I assume the script executor should also fallback to another classloader (context?) to resolve classes that has been added and the caller does not need to know about.

PR #574, opened 2019-02-13
migrate the model.thing tests

maggu2810:
Please use **rebase + merge** to keep the two commits separated.


bundles/org.openhab.core.model.thing/.classpath
```diff
@@ -23,7 +23,7 @@
 			<attribute name="maven.pomderived" value="true"/>
 		</attributes>
 	</classpathentry>
-	<classpathentry kind="src" output="target/test-classes" path="notestsources">
+	<classpathentry kind="src" output="target/test-classes" path="src.moved/test/java">
```
kaikreuzer: What is `src.moved` folder about? Do we need this here?

maggu2810: The model bundles are using a different non-standard Maven directory layout.

Instead of `src/main/java` for normal sources and `src/test/java` for test sources it is using `src` for normal files (additionally `src-gen`, `xtend-gen`).

As `src/test/java` would be part of `src` (the normal sources), we have to move that directory.
So I used `src.moved` because the base is moved because of Xtext.

kaikreuzer: ok, thanks!

PR #575, opened 2019-02-14
Make ChannelTypeRegistry immediate

htreu:
In a build setup where the core.thing.test integration tests are run by the bnd test runner with the felix runtime we experienced a problem with service references being disposed w/o referencing services being disposed. This led to multiple service references being available during test execution and failing tests.
Although I could not track down the root cause of this, a solution is to make the `ChannelTypeRegistry` an immediate component. This way it stays active during all test executions. In addition I think it should really be immediately available also in the runtime.

Signed-off-by: Henning Treu <henning.treu@googlemail.com>

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/type/ChannelTypeRegistry.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/type/ChannelTypeRegistry.java
@@ -32,7 +32,7 @@ import org.osgi.service.component.annotations.ReferencePolicy;
  *
  */
 @NonNullByDefault
-@Component(service = ChannelTypeRegistry.class)
+@Component(service = ChannelTypeRegistry.class, immediate = true)
 public class ChannelTypeRegistry {
 
     private final List<ChannelTypeProvider> channelTypeProviders = new CopyOnWriteArrayList<>();


maggu2810:
> service references being disposed w/o referencing services being disposed

That's sound strange.
Would you be able to share that test?

maggu2810:
Also guys with much more knowledge than me about OSGi states that it is bad practice to use immediate "unless it really needs to be".

Perhaps we could identify the root cause instead and fix that one instead of making it immediate if there is no real need (only the common use case).

htreu:
This PR was actually an offer in case you experience similar failures. I did dig into the felix SCR components but couldn't see whether reference counting is broken or the dependency tree is messed up or if its actually the Bnd test runner causing trouble (running the test stand-alone is fine, running all tests from the bundle breaks).
I know that equinox did not make this problems and also in the runtime the ChannelTypeRegistry can be lazy loaded since some immediate components depend on it.

You may close this PR w/o doing any harm :-)

maggu2810:
@htreu  I did not run into similar failures (until now).
Thank you for your offer!

So, you see that problem sometimes using Apache Felix, but you do not see it using Eclipse Equinox?
In both cases you are using the same Apache SCR implementation (and no Eclipse DS)?

htreu:
Here is where my speculations come from:

What I can observe:
* In the good old ESH days the test execution was equinox and all tests passed.
* In our test environment Apache Felix is the runtime, no difference in the running bundles (as far as I can tell, at least not the DS declarations) but one test is failing when the whole bundle is executed.
* The test fails due to a different ChannelTypeRegistry being injected into the items created then the one picking up my custom ChannelTypeProvider (see https://github.com/openhab/openhab-core/blob/master/itests/org.openhab.core.thing.tests/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelStateDescriptionProviderOSGiTest.java#L131).
* When all tests run in a row with felix, the ChannelTypeRegistry originally created to fulfil the ThingTypeRegistries demand is shut down due to the AutoUpdateManager shutting down. Felix then has a reference count of 0. I can see the `deactivate` method being called. The ThingTypeRegistry however is _not_ being destroyed (which is strange because it depends on the ChannelTypeRegistry).
* For further test cases new ChannelTypeRegistries are being created and destroyed on and on.
* The original ThingTypeRegistry however still holds the original instance of the ChannelTypeRegistry. And this one does also pick up my custom ChannelTypeProvider.
* The items I create in the test however do have a StateDescriptionService with a totally different dependency tree (meaning: different instances of ThingTypeRegistry and ChannelTypeRegistry. The later one with an empty list of ChannelTypeProviders.

This is really screwed and my conclusion is:
*  Having used equinox in ESH successfully felix somehow fails in reference counting or something else I do not see or understand.
*

kaikreuzer:
I am currently not aware of any such strangely failing tests, so let's close this PR.

PR #578, opened 2019-02-15
Updated NOTICE files to openHAB

Hilbrand:


kaikreuzer:
I actually deliberately didn't add a NOTICE file in the root. I think it is hard to maintain and not standard - we have a LICENSE file, which is what Github projects are expected to have and which should just be fine.

kaikreuzer:
Wrt the NOTICE file within the bundles, we should use the same across all repos, so it should be changes to be like https://github.com/openhab/openhab2-addons/blob/master/addons/binding/org.openhab.binding.airquality/NOTICE.

Hilbrand:
Updated: removed root NOTICE and other NOTICE files should now be same as openhab2-addons.

PR #581, opened 2019-02-15
Replaces 'Eclipse SmartHome' with 'openHAB' in text

Hilbrand:
Adapted CONTRIBUTING.md file.

P.s. not sure if the change in the `@author` tag is correct.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,13 +2,10 @@
 
 Warning: Note that this is most likely not the project you are looking for.
 
-openHAB mainly consist out of the core components from Eclipse SmartHome (https://github.com/eclipse/smarthome),
-extensions that are openHAB1-compatible from https://github.com/openhab/openhab and extensions that are 
-openHAB2-compatible from https://github.com/openhab/openhab2. All is then packaged into a distribution in
+openHAB mainly consist out of the core components, extensions that are openHAB1-compatible from
+https://github.com/openhab/openhab1-addons and extensions that are openHAB2-compatible from
+https://github.com/openhab/openhab2-addons. All is then packaged into a distribution in
 https://github.com/openhab/openhab-distro.
-
-This project here only contains a few bundles of the core runtime that are openHAB specific and thus not contained
-in the Eclipse SmartHome project.
 
 ## Contribution guidelines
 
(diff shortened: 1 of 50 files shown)


bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/library/items/CallItem.java
```diff
@@ -28,7 +28,7 @@
  * This item identifies a telephone call by its origin and destination.
  *
  * @author Thomas.Eichstaedt-Engelen / Initially CallItem
- * @author Gaël L'hopital - port to Eclipse SmartHome
+ * @author Gaël L'hopital - port to openHAB
```
maggu2810: Hm, I assume this cannot be changed as it will change the "history".
Gaël L'hopital has been ported it to Eclipse SmartHome, regardless if we migrate that code from Eclipse SmartHome to openHAB Core.

bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/library/types/StringListType.java
```diff
@@ -27,7 +27,7 @@
  * This type can be used for items that are dealing with telephony functionality.
  *
  * @author Thomas.Eichstaedt-Engelen / Initially CallType
- * @author Gaël L'hopital - port to Eclipse SmartHome
+ * @author Gaël L'hopital - port to openHAB
```
maggu2810: Hm, I assume this cannot be changed as it will change the "history".
Gaël L'hopital has been ported it to Eclipse SmartHome, regardless if we migrate that code from Eclipse SmartHome to openHAB Core.

bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/auth/client/oauth2/AccessTokenResponse.java
```diff
@@ -27,7 +27,7 @@
  * 4.4.3 - Client Credentials Grant - Access Token Response
  *
  * @author Michael Bock - Initial contribution
- * @author Gary Tse - Adaptation for Eclipse SmartHome
+ * @author Gary Tse - Adaptation for openHAB
```
maggu2810: Hm, I assume this cannot be changed as it will change the "history".
Gary Tse has done the adaption for Eclipse SmartHome, regardless if we migrate that code from Eclipse SmartHome to openHAB Core.

maggu2810:
> P.s. not sure if the change in the `@author` tag is correct.

IMHO it is not correct to change the project that work has been done before.

kaikreuzer:
> IMHO it is not correct to change the project that work has been done before.

I agree, those should be left untouched.

Hilbrand:
Updated changes in`@author`  lines. I was already not sure about it.

PR #589, opened 2019-02-20
Added maven archetype project to generate binding skeletons

### LINKED ISSUE #582: Migrate ESH maven skeleton archetype projects to this repository
The maven archetype projects for the skeleton scripts are not yet migrated to the openhab-core repository. 

Adding this issue so it will is visible as I've started working on this and to avoid some else also started working on it (or if someone already started it please report it here).

Hilbrand:
Note: I think the generated test launch file is not yet correct. But this seems a problem in general to all existing launch files too. So I would suggest to fix this when those files are being fixed to.

Closes #582

### DIFF
--- a/tools/pom.xml
+++ b/tools/pom.xml
@@ -9,12 +9,10 @@
     <version>2.5.0-SNAPSHOT</version>
   </parent>
 
-  <groupId>org.openhab</groupId>
-  <artifactId>tools</artifactId>
-
+  <artifactId>org.openhab.core.tools</artifactId>
   <packaging>pom</packaging>
 
-  <name>openHAB Tools</name>
+  <name>openHAB Core :: Tools</name>
 
   <modules>
     <module>archetype</module>
--- a/tools/pom.xml
+++ b/tools/pom.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.openhab.core</groupId>
+    <artifactId>org.openhab.core.reactor</artifactId>
+    <version>2.5.0-SNAPSHOT</version>
+  </parent>
+
+  <groupId>org.openhab</groupId>
+  <artifactId>tools</artifactId>
+
+  <packaging>pom</packaging>
+
+  <name>openHAB Tools</name>
+
+  <modules>
+    <module>archetype</module>
+  </modules>
+
+</project>
--- a/pom.xml
+++ b/pom.xml
@@ -774,7 +774,7 @@ Import-Package: \\
     <module>demo</module>
     <module>bundles</module>
     <module>features</module>
-    <!--<module>tools</module> -->
+    <module>tools</module>
   </modules>
 
 </project>
--- a/tools/archetype/binding/.project
+++ b/tools/archetype/binding/.project
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>org.openhab.archetype.binding</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+		<buildCommand>
+			<name>org.eclipse.m2e.core.maven2Builder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+	</buildSpec>
+	<natures>
+		<nature>org.eclipse.m2e.core.maven2Nature</nature>
+	</natures>
+</projectDescription>
--- a/tools/archetype/binding/pom.xml
+++ b/tools/archetype/binding/pom.xml
@@ -3,17 +3,15 @@
   <modelVersion>4.0.0</modelVersion>
 
   <parent>
-    <groupId>org.openhab.tools</groupId>
-    <artifactId>archetype</artifactId>
+    <groupId>org.openhab.core</groupId>
+    <artifactId>org.openhab.core.tools.archetypes</artifactId>
     <version>2.5.0-SNAPSHOT</version>
   </parent>
 
-  <groupId>org.openhab.archetype</groupId>
-  <artifactId>org.openhab.archetype.binding</artifactId>
-
+  <artifactId>org.openhab.core.tools.archetypes.binding</artifactId>
   <packaging>maven-archetype</packaging>
-  <name>openHAB Binding Archetype</name>
 
+  <name>openHAB Core :: Tools :: Archetypes :: Binding</name>
   <description>This is the Maven-Archetype for building new bindings for openHAB.</description>
 
   <build>
@@ -50,20 +48,6 @@
         </plugin>
       </plugins>
     </pluginManagement>
-    <plugins>
-      <!-- Disabled sat because nothing to check, and it actually breaks on spotbugs -->
-      <plugin>
-        <groupId>org.openhab.tools.sat</groupId>
-        <artifactId>sat-plugin</artifactId>
-        <version>${sat.version}</version>
-        <executions>
-          <execution>
-            <id>sat-all</id>
-            <phase>none</phase>
-          </execution>
-        </executions>
-      </plugin>
-    </plugins>
     <extensions>
       <extension>
         <groupId>org.apache.maven.archetype</groupId>
--- a/tools/archetype/binding/pom.xml
+++ b/tools/archetype/binding/pom.xml
@@ -1,0 +1,75 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.openhab.tools</groupId>
+    <artifactId>archetype</artifactId>
+    <version>2.5.0-SNAPSHOT</version>
+  </parent>
+
+  <groupId>org.openhab.archetype</groupId>
+  <artifactId>org.openhab.archetype.binding</artifactId>
+
+  <packaging>maven-archetype</packaging>
+  <name>openHAB Binding Archetype</name>
+
+  <description>This is the Maven-Archetype for building new bindings for openHAB.</description>
+
+  <build>
+    <resources>
+      <resource>
+        <filtering>true</filtering>
+        <directory>src/main/resources</directory>
+        <includes>
+          <include>archetype-resources/pom.xml</include>
+        </includes>
+      </resource>
+      <resource>
+        <filtering>false</filtering>
+        <directory>src/main/resources</directory>
+        <excludes>
+          <exclude>archetype-resources/pom.xml</exclude>
+        </excludes>
+      </resource>
+    </resources>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-archetype-plugin</artifactId>
+          <version>3.0.1</version>
+        </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-resources-plugin</artifactId>
+          <version>3.1.0</version>
+          <configuration>
+            <escapeString>@</escapeString>
+          </configuration>
+        </plugin>
+      </plugins>
+    </pluginManagement>
+    <plugins>
+      <!-- Disabled sat because nothing to check, and it actually breaks on spotbugs -->
+      <plugin>
+        <groupId>org.openhab.tools.sat</groupId>
+        <artifactId>sat-plugin</artifactId>
+        <version>${sat.version}</version>
+        <executions>
+          <execution>
+            <id>sat-all</id>
+            <phase>none</phase>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+    <extensions>
+      <extension>
+        <groupId>org.apache.maven.archetype</groupId>
+        <artifactId>archetype-packaging</artifactId>
+        <version>3.0.1</version>
+      </extension>
+    </extensions>
+  </build>
+</project>
--- a/tools/archetype/binding/src/main/resources/META-INF/archetype-post-generate.groovy
+++ b/tools/archetype/binding/src/main/resources/META-INF/archetype-post-generate.groovy
@@ -1,0 +1,157 @@
+/**
+ * Copyright (c) 2010-2019 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+nl = System.getProperty("line.separator")
+bundleName = request.getProperties().get('bindingIdCamelCase')
+githubUser = request.getProperties().get('githubUser')
+outputDirectory = new File(request.getOutputDirectory())
+
+/**
+ * Find which bundle the new binding should be added
+ */
+def String findAfterBundle(newBundle, codeownersFile) {
+    def codeowners = codeownersFile.getText('UTF-8') + '/bundles/' + newBundle
+    def lines = codeowners.split('\n').sort()
+    def lastIndex
+
+    lines.eachWithIndex { line, index ->
+        if (line.contains(newBundle)) {
+            lastIndex = index-1
+        }
+    }
+    return lines[lastIndex].replaceAll(/\/bundles\/([^\/]+)\/.+/, '$1')
+}
+
+/**
+ * Add the new bundle to the CODEWONERS file
+ */
+def addUserToCodewoners(githubUser, codeownersFile, bundleAfter, newBundle) {
+     def newContent = ''
+     def lines = codeownersFile.eachLine { line, index ->
+         newContent += line + nl
+         if (line.contains(bundleAfter)) {
+             newContent += '/bundles/' + newBundle + '/ @' + githubUser + nl
+         }
+     }
+     codeownersFile.write(newContent)
+     println 'Added github user to CODEOWNERS file'
+}
+
+/**
+ * Add the new bundle to the addons bom pom.xml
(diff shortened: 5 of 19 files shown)


maggu2810:
IMHO the whole archetype does not fit into the openHAB Core setup.
The openHAB Core bundles are using Maven + bnd and no bundled manifest etc.

Hilbrand:
With all bindings moved out of this project. And this is specific for binding generation. Do you suggest to move this to the openHAB2-addons project?

As the addons2 still do use the 'old' bundling, and I think there is some work to make it also possible to use the maven + bnd, a future version might use the new way. And then a new version of this project could be added here that will generate the new variant and then that new variant can be put here and the old variant can then be removed. Does that make sense?

maggu2810:
I don't know which would be the best repository for that archetype.
This archetype will not be used by any openHAB Core bundle, so, why adding it here.
The archetype will be used by "new" bundles that don't want to use the new toolchain.
IMHO we should add an archetype for the new toolchain to this repository.

I assume my view differs from @kaikreuzer ones, but IMHO this should not added to this repository. :wink: 

For me it is the same as the tycho POM it is consumed by no artifact of this repository.
But also this one has been added...
I was tired of discussing further.

So, I assume the same will be true here.

davidgraeff:
Doesn't seem to be necessary here, does it?. Would be more useful in the addons2 repository.

Btw: I wonder if individual bindings in the addons2 repository are allowed to use a tycho free buildsystem like in here, so that an archetype can already propagate the new way and bindings are migrated over time.

cweitkamp:
Just another idea for the script: Consider the [CODEOWNERS](https://github.com/openhab/openhab2-addons/blob/master/CODEOWNERS) file for new bundles.

Hilbrand:
@maggu2810 do read it correctly you suggest the archetype should be here, but generate a build based on maven + bnd build system? Or shouldn't it be here at all? My idea was to first add the current archetype based on the original code, but adapted for OH and then gradually change it's implementation. 

@davidgraeff I think that was the idea to be able to gradually move addons2 bindings away from tyco build. But I believe it requires to know what maven to import to make it possible (or anything else?) and I don't know what the status is of that.
 
@cweitkamp Unfortunately basically there is no script. It's just a generic maven plugin that does the work here. So I don't know how easy it will be to modify files like that. For example it now adds the binding as the last entry to the pom and modifies header and last line, very often resulting in code review comments to fix this, it would be nice if things like that could be fixed, and also adding the binding to the feature.xml would also help.

maggu2810:
If a new bundle is created for openHAB (e.g. starting from today without respect all the current bundles that in "WIP" state) should a Tycho based build approach be used or would it make sense to use a Maven + bnd based approach?

For ... I assume the answer is:

* core: a Tycho based bundle is not accepted and we require a Maven + bnd based approach.
* webui: @martinvw is currently working on the mirgation a Maven + bnd based approach of the webui. So, I assume also here we would like to not use the Tycho approach for new bundles
* addons:
  * IMHO new bundles should the Maven + bnd based approach.
  * IMHO we should migrate one bundle (e.g. the astro one) to a Maven + bnd based approach and ensure that it compiles and the tests are working so it can be used as a demonstration.

So, IF new bundles should not start with a Tycho based setup anymore, I don't think the archetype needs to be merged (regardless wich repository) before converting it to a Maven + bnd based archetype.
As it should not be used, it does not matter if it is not available in the master branch and the work is done in a working branch.


I am not sure about the correct repository at all.
If we can design the archetype that it works for every bundle perhaps core makes the most sense.
Perhaps also another repository "general tooling" or the infrastructure would make sense, too.

davidgraeff:
> addons: * IMHO we should migrate one bundle (e.g. the astro one) to a Maven + bnd based approach and ensure that it compiles and the tests are working so it can be used as a demonstration.

I'd help migrating oh2-addons, but I need one reference migration (including a test bundle).

maggu2810:
@davidgraeff :+1: 

I could start with a reference as soon as there is a positive feetback of ???
Which one is required?
* @openhab/2-x-add-ons-maintainers ?
* @kaikreuzer ?
* ... ?

wborn:
It would a good simplification when openhab2-addons uses bnd as well! :+1: 

> but I need one reference migration (including a test bundle

As a reference it may also be helpful if it is a binding that uses its own libs and uses a transport (e.g. feed or modbus).

Similar to the webui's it may not be possible to use both tycho and bnd (https://github.com/openhab/openhab-distro/issues/879#issuecomment-466182667) in the same reactor.

So if the migration is done with a single PR it may help if we pause merging PRs to prevent merge conflicts on the migration PR.

When the migration is done, it would also be helpful to have some documentation for contributors on to update their pending PRs.

davidgraeff:
> So if the migration is done with a single PR it may help if we pause merging PRs to prevent merge conflicts on the migration PR.

But then we can't split up the migration work.

> helpful to have some documentation

I can write such a document as soon as I have migrated my first bundle.

wborn:
> But then we can't split up the migration work.

If we accept that the build is broken during the migration we could also only merge these smaller migration PRs. It's similar to pausing merging other PRs with a big PR. If the PRs are small we can also still review them using the GitHub UI. :-)

maggu2810:
Just an idea:

Wouldn't it be simpler to use two reactors in the addons repo on migration?
Jenkins can build the one after the other on its CI job.

We could also provide "run" scripts that executes the command.

We could also modify the base POM to print a message on build that tells:

```
For the moment the build is split into two separate reactors.
Please build both using something like:

cd tycho; mvn clean install; cd ..
cd bnd; mvn clean install; cd ..
```

maggu2810:
And what's about the directory layout?
Should we use a similar one then openHAB Core?
A directory `bundles` for the addons and a directory `itests` for the integration tests?
Pure JUnit tests can be migrated to the addons similar as it has been done in core.
The directories `bundles` and `itests` can contain a flat list of bundles without further subdirectories for categories.

davidgraeff:
From a packagers perspective the directory layout doesn't matter that much.
But if you are just browsing the repo to get an overview, it helps a lot to have subdirectories. We have different types of extensions, I would follow those names for subdirectories.

maggu2810:
It is up to the addons maintainer, but IMHO such subdirectories complicates the whole reactor.
You have to create a lot of modules and reference that parent just for another subdirectory.

Then it comes to the question about the group and artifactory ID.
For the core we decided to use a flat list and it simplifies the working on the POMs (IMHO).

If it is such important to have more levels while browsing the repo...
Wouldn't a flat list with names that prefix match the scheme as used in core already shows the categories?

It does not need to result in an immense number of merge conflicts.
The "migrated" bundles resides in a new place, so all new PRs will use the new layout.
We could migrate one addon after each other (sure, we will do it for multiple ones parallel) and can merge PRs in front of the migration to the new place (for the specific bundles).
If the move to another directory is done correctly and you rebase the branch, it should work (IIRC) because Git knows that the files has been moved and which one you would like to change.

But this is just my personal feeling.
I also don't think about the separate builds (Tycho, Bnd + Maven) as long as the old directory layout is used.
For a flat list it is relative simple, there would be a parent bundle POM for Tycho and a parent bundle POM for Bnd.
If there is e.g. a directory "io" and a POM for the "io" directory, all bundles below that directory will use "io" as its parent. What is used by "io"? The Tycho one or the Bnd one? So, we need also an "io" for Tycho and an "io" for Bnd, ...

maggu2810:
Regardless of the directory layout (as long as "io" is using a Tycho parent, the migrated one don't need to declare "io" as its parent but can use another one -- it can be changed later easily -- hopefully)...

... I suggest another group / artifact ID naming scheme -- the similar one we used in core.

All "bundles" are using the same group ID. In core it is `org.openhab.core.bundles` and the artifact ID is the full qualified name e.g. `org.openhab.core.automation.rest`. The group ID for the addons could be `org.openhab.addons.bundles` (regardless of the sub category).

The same for integration tests, they could all use the group ID `org.openhab.addons.itests`.

If you agree, I could start with the migration of e.g. the astro binding that could be used as a demonstration.

davidgraeff:
I would love that to happen soon. Just one think that Kai never answered, and that also required me to leave the openhab2 maintainers team but is directly related:

What belongs into the core?

As of Kais perception, it is everything that can potentially be used by more than one other addon bundle, if I understood him correctly. So should we move the transports and transformations over to core in this process?

wborn:
I briefly tested what restructuring the add-on directories does to open PRs:
* New add-on contributions don't have a lot of merge conflicts. These just need to be moved to the right directory. See https://github.com/wborn/openhab2-addons/pull/3
* Updates to existing add-ons have seem to be causing quite a few merge conflicts.
  See: https://github.com/wborn/openhab2-addons/pull/4

Currently 91 of the 132 open PRs (69%) are new add-on contributions. So we can also restructure the directories without impacting most open PRs.

> the similar one we used in core

For simplicity and consistency it indeed makes sense to use a similar naming scheme.

> e.g. the astro binding that could be used as a demonstration

Compared to openhab-core many projects in openhab2-addons also use their own libraries in `lib`. So perhaps another binding would make a better example? 

Some of these libraries haven't been released to Maven repositories. So it would also be nice to know how to add dependencies to JARs part of the repo.

> What belongs into the core?

Sorry to see you go as maintainer. Thanks for all your help as maintainer. :+1: As contributor you can also make big differences!

If this is about bin2json I think it makes sense to add it to the core since it has more reuse potential compared to transports. IIRC there were also plans for wrapping Gson in the core. Though this is a subject for another PR/issue.

davidgraeff:
> If this is about bin2json I think it makes sense to add it to the core since it has more reuse potential compared to transports.

But what is the logical reasoning for one or the other? I pushed for the jinya templates transport to be in addons2, because I want wrapping PR to be merged soon. But that's a transport that makes sense in the core as well (as it also provides another transformation). Also feed can be used from more than one binding. On the other hand, the mqtt transport is used by addons2, so it could as well be in the addons2 repo (like with mqtt1, where the transport is in the same repo). I'm highly confused about the project structure.

maggu2810:
@Hilbrand Sorry! It seems I hijacked your PR and started a more generic discussion.

I don't know we should start a new issue somewhere or we should continue the discussion here to keep all at one place.

WDYT (you = all the other ones that commented here)?

maggu2810:
> Compared to openhab-core many projects in openhab2-addons also use their own libraries in `lib`. So perhaps another binding would make a better example?
> 
> Some of these libraries haven't been released to Maven repositories. So it would also be nice to know how to add dependencies to JARs part of the repo.

I don't know about that library but if they has never been published to a Maven repository are you sure that libraries are common used ones?
IMHO the preferred approach would be to use stuff that is present in any Maven repository.
If there is really no Maven repository that hosts that library, we should work with upstream so it gets pushed to a Maven repository or add it to a Maven repository on e.g. Bintray our self.

davidgraeff:
AFAIK all libs are also present on maven repositories. It might still be a good idea to document how you would add an external jar file, so that it is considered for bnd. A use-case could be a binding that supports an archaic device via a private jar file. And that jar file would never ever be used outside of that binding.

wborn:
> I don't know we should start a new issue somewhere or we should continue the discussion here to keep all at one place.

For discoverability by others it might make more sense to create a new openhab2-addons issue for this.

> add it to a Maven repository on e.g. Bintray our self

AFAIK besides Kai no other openhab2-addons maintainers can do that currently. 

I think some examples of bindings with libraries that are not in any Maven repository are:
* Grohe ONDUS Binding ([grohe-ondus-api-java](https://github.com/FlorianSW/grohe-ondus-api-java))
* Harmony Hub Binding ([harmony-client](https://github.com/digitaldan/harmony-client))
* LG WebOS Binding ([Connect-SDK-Java-Core](https://github.com/sprehn/Connect-SDK-Java-Core))
* Zoneminder Binding ([zoneminder4j](https://github.com/Mr-Eskildsen/zoneminder4j))

maggu2810:
> It might still be a good idea to document how you would add an external jar file, so that it is considered for bnd.

That's not Bnd specific, but Maven specific. There are several approach but for me all of that are just workarounds. IMHO it does not really fit in the Maven approach to use only locally available libraries.
You could e.g. use Maven plugins to unpack the local JAR file's classes to the build directory.
You can also pack the JAR file inside your artifact and reference it in the bundle's classpath.
I don't know if I would like to invest time to find the "best fit solution" for that corner case (that's only about the JAR file's location itself, about bundling the content see comment below).

>> add it to a Maven repository on e.g. Bintray our self

> AFAIK besides Kai no other openhab2-addons maintainers can do that currently.

Why not?
Does Kai control all the Bintray repositories? And prevent any other openHAB maintainer to create a repository on Bintray? Just kidding... Perhaps Kai is the only one that controls the official openHAB Bintray repository.
But if I am allowed to use an artifact that I uploaded to Maven Central by adding a dependency section to my POM. Why should it be forbidden to add the dependency section and a repository section to my POM to find the artifact on a repository of my choice?
Now, we could argue that the JAR file's respository must be also be accessible at runtime if someone installs the feature. But...

... if Maven is able to download the JAR file then we can use the Bnd plugin to handle its content. At compile time the dependency is used but after compilation the Bnd could add the content at packaging phase.
That way, it works on command line and also in the Eclipse IDE. The generated JAR file will be the same in both cases (other plugins fail because they are not handled by the Eclipse IDE's m2e lifecycle).
Bnd provides the private package (add all specified packages to the JAR)  and conditional package (add used packages to JAR only) instructions.

hakan42:
> Why should it be forbidden to add the dependency section and a repository section to my POM to find the artifact on a repository of my choice?
> Now, we could argue that the JAR file's respository must be also be accessible at runtime if someone installs the feature. But...

The last time I talked with Kai about this (admittedly, it is already more than a year ago), he said that he wanted developers to be able to start eclipse and be able to work immediately. Without waiting for maven to download stuff.

But I'm more used to the maven way of declaring dependencies and have the build system download them as well. You never know that library foobar-4711.jar is really what is on maven central or something someone locally modified.

wborn:
> Why should it be forbidden to add the dependency section and a repository section to my POM to find the artifact on a repository of my choice?

I'd be perfectly happy with allowing artifacts to be downloaded from Maven Central. We just need an approach for working with artifacts that aren't in any repo during the migration. E.g. by asking contributors to add them to Maven Central, adding these to the openHAB repo or (if possible) by configuring Bnd to use certain JARs. The MaryTTS voice add-on for example already downloads its dependencies from the openHAB Bintray repo during the  build.

Does Maven Central allow existing artifacts to be updated/deleted? That might be problematic if contributors decide to re-release artifacts to fix bugs or add features. I've only used private Artifactory/Nexus instances myself. 

> Without waiting for maven to download stuff.

They'll be cached as soon as they've been downloaded so that shouldn't really be an issue. You'll have the same issue when clearing your local Maven repository.

maggu2810:
> Does Maven Central allow existing artifacts to be updated/deleted?

Maven Central contains releases and not snapshot.
A released version never changes.

If a release gets an update it is another version.

But that should not really Maven Central related but normal software release management.

> I'd be perfectly happy with allowing artifacts to be downloaded from Maven Central.

Hm, I don't understand. If you call a normal Maven build for the current Tycho setup, a lot of stuff is downloaded from Maven Central.
It has been never different.

> by asking contributors to add them to Maven Central

My comment has about custom repositories, not the Central one.
I don't think that we should require the artifacts on Maven Central. I assume it is to complicated for beginners. But it should be simple for Bintray.

We are moving to a Maven based approach, so why are we trying to use local JAR files at all?

maggu2810:
Let's continue the "offtopic" discussion in the upstream repo:
https://github.com/openhab/openhab2-addons/issues/5005

wborn:
> But it should be simple for Bintray.

It's sure a lot easier to create your own Bintray repo compared to uploading to Maven Central. Isn't it also easier for them being removed one day causing builds to fail? As fallback we'll probably still have such JAR artifacts in stable/milestone/snapshot releases.

But IMHO this would be just as problematic as when someone decides to delete their GitHub repo containing all the source code of a library. Since we don't store such sources or source JARs.

To mitigate these issues we could also setup an openHAB repo proxy to such 3rd party repos. 

> why are we trying to use local JAR files at all

I'd like to manage them using dependencies in POMs as well. It would be a good improvement to be able to update libraries and all their dependencies by just updating a version number. :-)

Hilbrand:
@kaikreuzer Should this archetype project to generate bindings be part of the openhab-core or openhab2-addons, or maybe in it's own repository?

kaikreuzer:
I would prefer it to be in openhab-core (as it was in ESH in the past) as the skeleton can be seen as a tooling that is closely related to the API that we define in the core. That is, whenever we change something on the APIs and the way we expect add-ons to be developed, it should be done in a single PR against this repo.

davidgraeff:
@Hilbrand The archetype would need to be adapted to bnd and finally merged

Hilbrand:
Build breaks on sat ` Unable to execute mojo: Execution null of goal com.github.spotbugs:spotbugs-maven-plugin:3.1.6:spotbugs failed: java.io.IOException: No files to analyze could be opened `
Since this project doesn't contain any scanable sources this should not cause a problem for sat.
This project should be excluded/suppressed from sat or sat should be fixed. What should be done?

maggu2810:
I don't think the archetype needs to be checked by SAT.
If the created bundles are not accepted by SAT, we can fix the archetype, too.

If SAT needs to be changed depends IMHO if the error is only specific for the archetype or if there is a general failure. But I assume an IOE should not be thrown ever but SAT should complain the error that caused it.

Hilbrand:
SAT doesn't need to be checked on archetype since it doesn't contain any sources to check. Spotbug seems to fail because it tries to open a sources directory or something, but that doesn't exist in this project. In general this is not a common use case, and should probably not happen in any other case. So changing SAT would be for this specific case. I've tried newer versions of spotbug plugin, but that didn't work out.

So I've disabled the sat check in the archetype binding pom, in the same way it's disabled in the root pom when `skipChecks` profile is enabled.

Hilbrand:
@wborn or @cweitkamp since we're almost done with migration of all bindings to the new bnd build it would be nice to have the new skeleton scripts in place too when it's finished. Can you have a look at this pr to see if it can be merged or needs changes?

tools/pom.xml
```diff
@@ -0,0 +1,23 @@
...
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.openhab.core</groupId>
+    <artifactId>org.openhab.core.reactor</artifactId>
+    <version>2.5.0-SNAPSHOT</version>
+  </parent>
+
+  <groupId>org.openhab</groupId>
```
maggu2810: Hm, most stuff in this repo (as soon as the tycho leftovers used by addons can be removed) if using the group ID prefix `org.openhab.core`.
Why can't this be used for the archetype?

IMHO: The group ID is an ID "who" provides something, not "who" consumes something.
So, as long as this "tooling" is part of openHAB Core it should be use that group ID.
If it is not part of core at all, why placing it in this repo?

Also for the other fields (e.g. name, artifact ID, ...) it would be nice to use the format used in the other POM files (e.g. `openHAB Core :: ...`).

Hilbrand: I didn't pay a lot of attention to the naming, so copied basically what was in there. I've updated the groupIds, artifactIds and names. But you might want to have a second look to see if that was correctly done. Otherwise please do suggest how you think it should be named.

Issue #591, opened 2019-02-20
NPE: The removeModelParser method has thrown an exception


cweitkamp:
Running OH2 snapshot 1526. Exception occurred during stop / restart:

```
2019-02-20 15:51:32.630 [ERROR] [org.openhab.core.model.core         ] - bundle org.openhab.core.model.core:2.5.0.201902190301 (160)[org.eclipse.smarthome.folder(116)] : The removeModelParser method has thrown an exception
java.lang.NullPointerException: null
	at org.eclipse.smarthome.model.core.internal.folder.FolderObserver.removeModelParser(FolderObserver.java:101) ~[?:?]
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:?]
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[?:?]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:?]
	at java.lang.reflect.Method.invoke(Method.java:498) ~[?:?]
	at org.apache.felix.scr.impl.inject.methods.BaseMethod.invokeMethod(BaseMethod.java:228) ~[42:org.apache.felix.scr:2.1.14]
	at org.apache.felix.scr.impl.inject.methods.BaseMethod.access$500(BaseMethod.java:41) ~[42:org.apache.felix.scr:2.1.14]
	at org.apache.felix.scr.impl.inject.methods.BaseMethod$Resolved.invoke(BaseMethod.java:664) ~[42:org.apache.felix.scr:2.1.14]
	at org.apache.felix.scr.impl.inject.methods.BaseMethod.invoke(BaseMethod.java:510) [42:org.apache.felix.scr:2.1.14]
	at org.apache.felix.scr.impl.inject.methods.BindMethod.invoke(BindMethod.java:42) [42:org.apache.felix.scr:2.1.14]
	at org.apache.felix.scr.impl.manager.DependencyManager.invokeUnbindMethod(DependencyManager.java:1927) [42:org.apache.felix.scr:2.1.14]
	at org.apache.felix.scr.impl.manager.DependencyManager.close(DependencyManager.java:1682) [42:org.apache.felix.scr:2.1.14]
	at org.apache.felix.scr.impl.manager.SingleComponentManager.disposeImplementationObject(SingleComponentManager.java:417) [42:org.apache.felix.scr:2.1.14]
[...]

PR #592, opened 2019-02-20
model.core: respect OSGi lifecycle

maggu2810:
Fixes: https://github.com/openhab/openhab-core/issues/591

"Fixes" it hopefully. @cweitkamp please check yourself.


wborn:
Looks like a call to [deleteModelsFromRepo](https://github.com/openhab/openhab-core/blob/a842bb7ed3a09181bf502a2908f06f2ff798fa32/bundles/org.openhab.core.model.core/src/main/java/org/eclipse/smarthome/model/core/internal/folder/FolderObserver.java#L204) might also result in an NPE if `modelRepo` is null. To prevent these issues it would make sense to annotate the class with  null annotations.

maggu2810:
@wborn the method you mention is called on deactivation only.
If the component gets deactivated it has been activated before.
As the model repository is mandatory it is set as long as the component is active.
So I don't see any potential NPE here.

cweitkamp:
> please check yourself.

Thank you @maggu2810. It is working in my environment. I do not see the exception anymore.

PR #596, opened 2019-02-22
Change log level for unknown item calls from info to warn

semperor:
Calling a method for an unknown item seems to be a misconfiguration or typo. Both should be mentioned with a warning, because you have to expect unwanted behavior, such as not working widgets on HabPanel.

Signed-off-by: Hagen Marianek <semperor@web.de> (github: github_handle)

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/item/ItemResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/item/ItemResource.java
@@ -267,7 +267,7 @@ public class ItemResource implements RESTResource {
             dto.editable = isEditable(dto.name);
             return JSONResponse.createResponse(Status.OK, dto, null);
         } else {
-            logger.info("Received HTTP GET request at '{}' for the unknown item '{}'.", uriInfo.getPath(), itemname);
+            logger.warn("Received HTTP GET request at '{}' for the unknown item '{}'.", uriInfo.getPath(), itemname);
             return getItemNotFoundResponse(itemname);
         }
     }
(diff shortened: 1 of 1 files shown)


davidgraeff:
Hm. You seem to look at the logs only apparently. The http response will be a 404, so HabPanel should actually tell you that something went wrong.

I would actually do the reverse and remove all logs. The output is clearly communicated already to the consumer.

maggu2810:
I agree with @davidgraeff.
And there is no warning to log. The log should not warn about "the user has doing something wrong". But there is something wrong in the code. And I don't think there is something wrong in the code.

semperor:
I would totally agree with you both if the REST interface would typically be used by callers we do not know. But honestly, the vast majority of openHABs out there are certainly one-man shows. Therefore, the one who configures / programs the caller is also the one who runs the REST interface. And then I just want to know if my currently configured call is correct or not, or if I have forgotten to rename the just renamed item in all GUIs as well. Of course it would be the task of the caller to tell me this. But there is no guarantee. For example HABpanel just doesn't do it. We can hope that the caller informs the admin or we change these lines and make sure that the admin gets the message.

davidgraeff:
> For example HABpanel just doesn't do it

open a bug report on HABPanels repo. It should show a notification on a 404.
HabPanel is actually the only consumer where you define items on site.
Sitemaps are throwing errors if items do not exist.

semperor:
Just for the record: So you think it's more important to keep the principles than to take this chance to make openHAB a little more noob friendly? That saddens me. Because there will always be faulty calls to the REST interface. Whether from HABpanel or elsewhere.

maggu2810:
@semperor See the logging coding guidelines here: https://www.openhab.org/docs/developer/development/guidelines.html#e-logging

Views may differ, sorry.

PR #603, opened 2019-02-25
bump to OHDR 1.0.40

maggu2810:


features/p2/poms/tycho/pom.xml
```diff
@@ -41,8 +41,8 @@
 
     <!-- Eclipse p2 repository -->
     <repository>
-      <id>eclipse-oxygen</id>
-      <url>http://download.eclipse.org/releases/oxygen/201709271000</url>
+      <id>eclipse-photon</id>
+      <url>http://download.eclipse.org/releases/photon</url>
```
kaikreuzer: Referencing a dedicated version actually speeds up the build as Tycho does not have to read the indexes from ALL the photon releases.

maggu2810: AFAIK http://download.eclipse.org/releases/photon aggregates ATM

* http://download.eclipse.org/technology/epp/packages/photon/
* http://download.eclipse.org/releases/photon/201806271001/

I cannot judge if any consumer needs a bundle of the EPP repository while bumping to Photon.

Need to rebuild all repositories again after changing that line.

Which repos do you consider as necessary to test?

kaikreuzer: Hm, I wasn't aware that it is also aggregating the EPP repo.
Never mind, let's leave it as it is, it was just a comment about potential optimisation (as the previous url pointed to a specific version).

Issue #561, opened 2019-02-11
Handle hexadecimal firmware versions


YordanDZhelev:
Currently it is not possible to have hexadecimal string as thing firmware versions. Even though the version is an integer, it cannot be handled correctly by the [isInt(String s)](https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/firmware/FirmwareImpl.java#L315) check. It will be either considered not an integer if the version string contains one of `A, B, C, D, E or F` or will be falsely parsed as an integer, but with incorrect value. Such hex versions are already in use in [zigbee binding](https://github.com/openhab/org.openhab.binding.zigbee/blob/master/org.openhab.binding.zigbee/src/main/java/org/openhab/binding/zigbee/discovery/ZigBeeNodePropertyDiscoverer.java#L188). 
I would suggest to change [version comparison](https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/firmware/FirmwareImpl.java#L275) to be able to work also with hexadecimal versions by relying on a `0x` prefix infront of the actual version string. 
Such change is already prepared for the zigbee binding in this [zigbee binding PR](https://github.com/openhab/org.openhab.binding.zigbee/pull/384)

maggu2810:
Hi @YordanDZhelev,
what has been the reason to use hexadecimal representation of the firmware version in the ZigBee binding?
Is there any advantage?

hsudbrock:
Maybe @cdjackson knows an answer to @maggu2810 's question above; Chris, do you maybe recall whether there was a particular reason to use hex numbers to represent the FW versions in the ZigBee binding?

cdjackson:
I believe it was done like this to make the versions more readable - as decimal, they mean nothing - but as hex they are usable since mostly each byte is an independent part of the version.

I'm a little surprised that this doesn't work as it has been working ok like this for well over a year. Maybe we got lucky and versions don't contain A-F values.

hsudbrock:
Thanks for your feedback, Chris!

> I'm a little surprised that this doesn't work as it has been working ok like this for well over a year. Maybe we got lucky and versions don't contain A-F values.

[...]

PR #604, opened 2019-02-25
Support hexadecimal version parts explicitly.

YordanDZhelev:
Related to [this open issue](https://github.com/openhab/openhab-core/issues/561)
Signed-off-by: YordanDZhelev <zhelev.yordan@gmail.com>


kaikreuzer:
@YordanDZhelev Did you see @maggu2810's question?

maggu2810:
@YordanDZhelev friendly ping

maggu2810:
@openhab/core-maintainers Can we close this because of "no feedback"?

YordanDZhelev:
@maggu2810 If we don't want to handle the case with FW versions in hexadeciaml format explicitly in the framework it si perfectly OK for me. The only place it is used by now is in one binding, and versions there are padded, so comparing them lexicographically as strings will handle this case. Maybe should close this PR.

maggu2810:
All fine. Thank you for your response.

PR #515, opened 2019-02-01
Initial implementation for JsonBindingService

kaikreuzer:
Ported over from https://github.com/eclipse/smarthome/pull/5856

Also-By: Flavio Costa <flavio.costa.br@gmail.com>
Signed-off-by: Kai Kreuzer <kai@openhab.org>

PR #611, opened 2019-02-26
[bin2json] Binary to JSON converter initial contribution

paulianttila:
This bundle can be used to convert binary data to JSON format. See further details from README.

Signed-off-by: Pauli Anttila <pauli.anttila@gmail.com>


paulianttila:
I can't run any integration tests on my environment (maven build doesn't work either). Should they work or is is ESH integration still in progress?

features/karaf/openhab-core/src/main/feature/feature.xml
```diff
@@ -136,6 +136,7 @@
     </capability>
 
     <bundle>mvn:org.openhab.core.bundles/org.openhab.core.voice/${project.version}</bundle>
+    <bundle>mvn:org.openhab.core.bundles/org.openhab.core.io.bin2json/${project.version}</bundle>
```
maggu2810: It should not be added to base (openhab-core-base) feature.

If you would like to ensure that it is present in every openHAB distribution, then add it to `openhab-runtime-base`.

I would prefer a separate feature for bin2json at all that references the specific TP feature (see below).

features/karaf/openhab-tp/src/main/feature/feature.xml
```diff
@@ -33,6 +33,9 @@
     <!-- Gson -->
     <bundle dependency="true">mvn:com.google.code.gson/gson/2.3.1</bundle>
 
+    <!-- JBBP -->
+    <bundle dependency="true">mvn:com.igormaznitsa/jbbp/1.4.1</bundle>
```
maggu2810: Can you create a separate feature for that bunde (e.g. like `openhab.tp-commons-net`)?
The separate bin2json feature in openhab-core feature repo could then depend on the tp (by adding a "dependency=true" dependency to that TP feature)

features/p2/repo/pom.xml
```diff
@@ -201,6 +201,11 @@
                   <transitive>false</transitive>
                   <source>true</source>
                 </artifact>
+                <artifact>
+                  <id>org.openhab.core.bundles:org.openhab.core.io.bin2json:${project.version}</id>
+                  <transitive>false</transitive>
```
maggu2810: In the openhab-core BOM you inserted it keeping the alphabetical order. Would it be possible to move it to the "right" place here, too?

itests/org.openhab.core.io.bin2json.tests/src/test/java/org/openhab/core/io/bin2json/tests/Bin2JsonTest.java
```diff
@@ -0,0 +1,74 @@
...
+import org.openhab.io.bin2json.ConversionException;
+
+import com.google.gson.JsonObject;
+
+/**
+ * Integration tests for {@link Bin2Json}.
+ *
+ * @author Pauli Anttila - Initial contribution
+ */
+public class Bin2JsonTest {
```
maggu2810: This looks like a pure JUnit test and not a test that requires an OSGi framework.
You don't need to use an integration test project at all.

It should be enough to use that class into a package that resides in the `src/test/java` folder of your bundle.
The JUnit test should then be run on compile time.

paulianttila:
Thanks @maggu2810. I did some changes. I'm struggling with the features. Could you take a look and give me some hint?

features/karaf/openhab-core/src/main/feature/feature.xml
```diff
@@ -215,6 +215,13 @@
     <bundle>mvn:org.openhab.core.bundles/org.openhab.core.config.discovery.usbserial.linuxsysfs/${project.version}</bundle>
   </feature>
 
+  <feature name="openhab-core-io-bin2json" version="${project.version}">
```
maggu2810: I see your commit fb4604fb832c188074235ba5d022e62947a53d54
Why did you remove it from the base runtime again?
I don't know your discussion with Kai. Should it be an optional feature that is installed on request or should it part of every openHAB installation?

maggu2810:
> Unable to create resource for bundle mvn:com.igormaznitsa/jbbp/1.4.1

Need to have a look at myself.

maggu2810:
The error is caused because jbbp is not an OSGi bundle -- the manifest is missing all the nice details OSGi is using.

There are some options:
* Create a PR on the upstream project that the build process adds the missing OSGi manifest entries and ask him to release a new version.
* Do it yourself and add the artifact to a Maven Repository
* Don't add the OSGi manifest entries and put the classes of that artifact into your bundle
* ...

bundles/org.openhab.core.io.bin2json/src/main/java/org/openhab/core/io/bin2json/Bin2Json.java
```diff
@@ -0,0 +1,239 @@
...
+ * </pre>
+ *
+ * @author Pauli Anttila - Initial contribution
+ *
+ */
+public class Bin2Json {
+
+    private final Logger logger = LoggerFactory.getLogger(Bin2Json.class);
+
+    private JBBPParser parser;
```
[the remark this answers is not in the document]
paulianttila: Done

bundles/org.openhab.core.io.bin2json/src/main/java/org/openhab/core/io/bin2json/Bin2Json.java
```diff
@@ -0,0 +1,239 @@
...
+
+    /**
+     *
```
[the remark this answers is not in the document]
paulianttila: Done

bundles/org.openhab.core.io.bin2json/src/main/java/org/openhab/core/io/bin2json/Bin2Json.java
```diff
@@ -0,0 +1,239 @@
...
+     * @param parserRule Binary data parser rule
+     * @throws ConversionException
```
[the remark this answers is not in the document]
paulianttila: Done

bundles/org.openhab.core.io.bin2json/src/main/java/org/openhab/core/io/bin2json/Bin2Json.java
```diff
@@ -0,0 +1,239 @@
...
+            throw new ConversionException(String.format("Unexpected error, reason: %s", e.getMessage(), e));
+        }
+    }
+
+    /**
+     * Convert data from {@link InputStream} to JSON object.
+     *
+     * @param inputStream
+     * @return Gson {@link JsonObject}
+     * @throws ConversionException
```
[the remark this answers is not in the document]
paulianttila: Done

bundles/org.openhab.core.io.bin2json/src/main/java/org/openhab/core/io/bin2json/Bin2Json.java
```diff
@@ -0,0 +1,239 @@
...
+        }
+    }
+
+    private JsonObject convert(JBBPFieldStruct data) throws ConversionException {
+        try {
+            LocalDateTime start = LocalDateTime.now();
+            final JsonObject json = convertToJSon(data);
+            if (logger.isTraceEnabled()) {
+                Duration duration = Duration.between(start, LocalDateTime.now());
+                logger.trace("Conversion time={}, json={}", duration, json.toString());
```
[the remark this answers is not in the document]
paulianttila: done

bundles/org.openhab.core.io.bin2json/src/test/java/org/openhab/core/io/bin2json/Bin2JsonTest.java
```diff
@@ -0,0 +1,72 @@
...
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+import org.junit.Test;
+
+import com.google.gson.JsonObject;
+
+/**
+ * Integration tests for {@link Bin2Json}.
```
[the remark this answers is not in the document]
paulianttila: Done

kaikreuzer:
Question: With https://github.com/openhab/openhab-core/pull/515, we introduce(d) a `org.openhab.core.io.json` bundle. If we consider bin2json a bundle that is part of openhab-base (as defined in the feature), I think it does not really make much sense to have a separate `bin2json` bundle, but these utilities should rather be a part of the `org.openhab.core.io.json` bundle (as this is "all about json support"). Wdyt?

maggu2810:
Hm, my understanding has been that `org.openhab.core.io.json` defines the API and the part that is independent of the specific JSON implementation. There will be support to use different JSON implementations e.g. `org.openhab.core.io.json.gson`, `org.openhab.core.io.json.jackson` or whatever.

If my reading of this code base has been correct it is specific for Gson -- at least the Gson's JsonObject is part of the API.

kaikreuzer:
> If my reading of this code base has been correct it is specific for Gson -- at least the Gson's JsonObject is part of the API.

So does it make sense to make it depend on Gson when we try to become independent of it?

If we cannot avoid gson, then the suitable place for the code would be the `...json.gson` bundle - but again, I think we should try to limit the number of bundles, where possible.

maggu2810:
Okay, seems I misunderstand that PR.
My understanding has been it has been reviewed and the missing part has been to migrate it to core.

Do we still think about the correct place (e.g. bundle) for this code?
Then @paulianttila should stop to migrate or implement.

> If we cannot avoid gson, then

we should think about #515 again.

Perhaps I also did not get this intention correctly.
Hasn't been one point that we don't need to depend on the specific JSON implementation and products can use Gson, jackson or whatever they want?

If we cannot avoid Gson, why should we add an abstraction layer at all.
Couldn't we declare Gson as a requirement for openHAB?
We should just ensure that we don't rely on the "internal" package of Gson, so we can use the official Gson bundle, too.

kaikreuzer:
> Do we still think about the correct place (e.g. bundle) for this code?

Well, at least I don't really think it should be a separate bundle - I regard the feature to be a utility around JSON processing, that's why my suggestion was the existing `io.json` bundle.

> Hasn't been one point that we don't need to depend on the specific JSON implementation and products can use Gson, jackson or whatever they want?

Right, that's I was asking (@paulianttila) above whether we really have to depend on GSON with this feature - if so, I agree that #515 is contradicting this and we need to clarify how to proceed.

> Couldn't we declare Gson as a requirement for openHAB?

This would be one option, which means we would drop #515. At ESH our plan was to be able to potentially smoothly move to JSONB one day...

kaikreuzer:
@maggu2810 For openHAB distro, I think staying with Gson as a requirement is acceptable.
I was always seeing the efforts of #515 rather to be relevant for making solutions more future proof by using official standards instead of specific implementations. But if this isn't something that you see valuable to pursue, let's close #515 and make our life simpler.

maggu2810:
Using official standards and to develop for APIs instead of specific implementation should always be preferred.
But #515 continues the specific ESH PR that has initial created on "Jul 4, 2018".

I don't know if the time that has been invested and needs to be further invested (development - review - cycles) makes sense.
If the PR is ready and just needs to be accepted, no need to turn it away.

So, how far are we from using a "official standard"?

kaikreuzer:
Now that #515 has been closed, I think we continue here just as if nothing happened.
@paulianttila Is it still WIP? If so, please ping us once you consider it ready for another review!

maggu2810:
@paulianttila Did you see @kaikreuzer message?

kaikreuzer:
@paulianttila I see you meanwhile pushed further commits. Shall I remove the "wip" label and do another review of the PR?

paulianttila:
@kaikreuzer, yes please

bundles/org.openhab.core.io.bin2json/.classpath
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kaikreuzer: This file seems to be missing entries for the Java source folders - at least they are not present, if I import the project in Eclipse.

features/karaf/openhab-tp/src/main/feature/feature.xml
```diff
@@ -85,6 +85,11 @@
     </config>
   </feature>
 
+  <feature name="openhab.tp-jbbp" description="Java Binary Block Parser library" version="${project.version}">
+    <capability>openhab.tp;feature=jbbp;version=1.4.1</capability>
+    <bundle dependency="true">wrap:mvn:com.igormaznitsa/jbbp/1.4.1</bundle>
```
kaikreuzer: @J-N-K Am I right that instead of wrapping the artifact here, it should be added to https://github.com/openhab/openhab-osgiify?

bundles/org.openhab.core.io.bin2json/src/main/java/org/openhab/core/io/bin2json/ConversionException.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * {@link ConversionException} generic exception for errors which occurs during conversion.
+ *
+ * @author Pauli Anttila - Initial contribution
+ */
+@NonNullByDefault
+public class ConversionException extends Exception {
+
+    private static final long serialVersionUID = 1L;
```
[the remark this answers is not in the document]
paulianttila: Just a curiosity, why do you think that generated is better than just start from 0 or 1? I mean if someone will make changes to class which are not backward compatible from serialisation point of view, serialVersionUID need to be manually changed by the developer as serialVersionUID is anyhow explicitly defined.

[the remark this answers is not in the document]
kaikreuzer: I never ever gave much thoughts on this myself, I must admit. I always used the generation feature of Eclipse for it. But I cannot give an answer, why this would be better than just 0...
I have also never ever experienced that anyone changed the serialVersionUID when doing changes to the class - in theory, we should enforce this, I guess. But as we don't use Java serialisation anywhere in openHAB, I think that it really hardly matters at all...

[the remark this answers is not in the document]
cweitkamp: Alright. It is clear to me now. We do not care very much - which is okay - because we do not need it.

@paulianttila Up to you if you like to change it or not.

features/karaf/openhab-tp/src/main/feature/feature.xml
kaikreuzer: @J-N-K Did you see my question?

J-N-K: Already there. Just need to change to 

```
<bundle dependency="true">mvn:org.openhab.osgiify/com.igormaznitsa.jbbp/1.4.1</bundle>
```

kaikreuzer: Excellent, thanks!

kaikreuzer:
```suggestion
    <bundle dependency="true">mvn:org.openhab.osgiify/com.igormaznitsa.jbbp/1.4.1</bundle>
```

bundles/org.openhab.core.io.bin2json/.settings/org.eclipse.core.resources.prefs
```diff
@@ -0,0 +1,4 @@
+eclipse.preferences.version=1
```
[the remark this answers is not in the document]
paulianttila: done

features/karaf/openhab-tp/src/main/feature/feature.xml
paulianttila: done

bundles/org.openhab.core.io.bin2json/.classpath
paulianttila: done

paulianttila:
@kaikreuzer, @cweitkamp All issues should be now fixed?

PR #617, opened 2019-03-01
Added standard gravity and dBm Units

Hilbrand:
Also did some clean up, which is in the first commit. The units are in the second commit.
Related to: https://github.com/openhab/openhab2-addons/pull/4913

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/internal/library/unit/UnitInitializer.java
+++ b/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/internal/library/unit/UnitInitializer.java
@@ -14,6 +14,7 @@ package org.eclipse.smarthome.core.internal.library.unit;
 
 import org.eclipse.smarthome.core.library.unit.ImperialUnits;
 import org.eclipse.smarthome.core.library.unit.SIUnits;
+import org.eclipse.smarthome.core.library.unit.SmartHomeUnits;
 
 /**
  * Make sure static blocks from {@link SIUnits} & {@link ImperialUnits} are executed to initialize the unit parser.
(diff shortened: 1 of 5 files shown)


maggu2810:
What's the reason to make the classes final and its constructor private?
Do you see any critical if someone wants to create a sub class so we need to prevent it that strong?
Sure, it would make no sense (perhaps) but does it harm if?

maggu2810:
I assume it has been wrong before but has been possible to use every "SmartHomeUnit" as "SIUnit" and "ImperialUnit".

```
        System.out.println(SmartHomeUnits.AMPERE);
        System.out.println(SIUnits.AMPERE);
        System.out.println(ImperialUnits.AMPERE);
```

It is not valid anymore.

It seems to be correct after the change because a SmartHomeUnit is not necessary an SIUnit.

But should we mark this as "API breaking" (and merge it) as it can break the consumers.

Hilbrand:
The classes already had private constructors, except for `SmartHomeUnits` that had a protected constructor. Because of the private constructors is seems logical to make this more explicit and make the classes final. `SmartHomeUnits` was kind of misused by the other to classes. They extended `SmartHomeUnits`, but created a singleton of their own class, even while `SmartHomeUnits` also is used as a singleton class. So I changed the extend to the `AbstractSystemOfUnits` class they should extend (I think) and with that change `SmartHomeUnits` can have the private constructor too (and be a final class).

Hilbrand:
I checked the openhab2 repo and found 2 issues of problems due to the changes, but only in tests. I've created a [pr (5016)](https://github.com/openhab/openhab2-addons/pull/5016) to fix them.

maggu2810:
Thanks

PR #618, opened 2019-03-02
Added new parent class for units to fix import problems.

Hilbrand:
The change with unit classes parent (#617) causes it to report missing imports of tec.uom.se when using static import. Before the change this was not a problem because the parent class was SmartHomeUnits. This change adds a new parent class for all units.


wborn:
Nice. This way we can also use static imports again with `SmartHomeUnits`. :-)

Issue #613, opened 2019-02-28
[Rules DSL] StringUtils Missing since at least 1536


semperor:
Hello.
As mentioned in this thread https://community.openhab.org/t/2-5-0-snapshot-build-1536-problems-with-string-methods/68799 the StringUtils methods are not longer available for strings. This started at least with build 1536 and exists in the build 1541.

lolodomo:
I use "replaceChars" in some of my DSL rules and I can confirm that it is no more working:
```
2019-03-03 10:20:32.391 [ERROR] [ntime.internal.engine.RuleEngineImpl] - Rule 'Controle par la voix': 'replaceChars' is not a member of 'java.lang.String'; line 769, column 24, length 45
```
I am running snapshot 1544.

lolodomo:
@kaikreuzer : I still have the issue using replaceChars with snapshot 1549. Is it expected ?

semperor:
Unfortunately still not fixed at all:

https://community.openhab.org/t/2-5-0-snapshot-build-1536-problems-with-string-methods/68799/12?u=semperor

lolodomo:
Can you reopen the issue ?

semperor:
> Can you reopen the issue ?

Unfortunately not, as it was closed by a repo collaborator. @maggu2810 or @kaikreuzer please reopen this issue. Thank you.

kaikreuzer:
Reopened it.
@maggu2810 Any idea why my fix didn't solve it? What else could have changed in the bundle by the bnd migration?

maggu2810:
@kaikreuzer I try to understand, how the packages for explicit imports has been choosen on #620?
As it seems no to solve this specific problem, are this additional packages the ones that needs to be present for scripts and rules?
Which package contains the missing class?
Can someone share a minimal example to reproduce the problem using openHAB Core only?

maggu2810:
@semperor It has been closed because a PR has been merged that contains a "fixes #613". This will trigger Github to close the related issue automatically.

@kaikreuzer I try to understand, how the packages for explicit imports has been choosen on #620?
As it seems no to solve this specific problem, are this additional packages the ones that needs to be present for scripts and rules?
Which package contains the missing class?
Can someone share a minimal example to reproduce the problem using openHAB Core only?

openhab-5iver:
> Can someone share a minimal example to reproduce the problem using openHAB Core only?

@maggu2810, you could test with this DSL rule...

```
rule "Test"
when
[...]

PR #620, opened 2019-03-03
made package imports explicit

kaikreuzer:
fixes https://github.com/openhab/openhab-core/issues/613


kaikreuzer:
I've added a Guava version constraint.
Will this also fix the PR build issues? I don't quite get why it cannot resolve Guava for rules, while it does for scripts...

maggu2810:
Yes it should. As you removed the version the version has been set to (AFAIK): 14 <= x < 15
Guava does not use semver and so we would like to say: we need at least version 14, so all >= 14.

kaikreuzer:
Indeed, all buids green, so feel free to merge.

PR #621, opened 2019-03-04
[Automation] DayOfWeekCondition

davidgraeff:
Remove inconsistent visibility "PUBLIC". This value is not documented (there is only HIDDEN and VISIBLE) and this is the only module type that uses PUBLIC instead of just not setting this value at all.

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/resources/ESH-INF/automation/moduletypes/DayOfWeekCondition.json
+++ b/bundles/org.openhab.core.automation/src/main/resources/ESH-INF/automation/moduletypes/DayOfWeekCondition.json
@@ -4,7 +4,6 @@
          "uid":"timer.DayOfWeekCondition",
          "label":"it is a certain day of the week",
          "description":"checks for the current day of the week",
-         "visibility":"PUBLIC",
          "configDescriptions":[  
             {  
                "name":"days",


openhab-5iver:
> there is only HIDDEN and VISIBLE

Just FYI... a null VISIBILITY [defaults to VISIBLE](https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/type/ModuleType.java#L117). 

Also, there is [EXPERT](https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/Visibility.java#L38), which is used in [private custom module types](https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/resources/ESH-INF/automation/moduletypes/PrivateScriptedTypes.json)... private in that these are [only visible from within scripts](https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/ScriptedPrivateModuleHandlerFactory.java#L28). When creating module types in scripts, you can choose private or non-private.

Issue #610, opened 2019-02-26
[JSR223] Most Java classes are no longer available after ESH migration


openhab-5iver:
I'm testing this on S1534. Classes that were previously available can no longer be imported. Here is a list of the ones I have tested, when trying to update the [openhab2-jython](https://github.com/OH-Jython-Scripters/openhab2-jython) libraries. The ones with ##### worked, but the others did not.
```
# core.date
from org.openhab.core.library.types import DateTimeType as LegacyDateTimeType
#####from org.eclipse.smarthome.core.library.types import DateTimeType

#####from org.eclipse.smarthome.core.items import Metadata
#####from org.eclipse.smarthome.core.items import MetadataKey

from org.joda.time import DateTime

from org.eclipse.smarthome.automation.core.util import RuleBuilder
#####from org.openhab.core.automation import Rule as SmarthomeRule
#####from org.openhab.core.automation.handler import TriggerHandler

# core.triggers
from org.eclipse.smarthome.automation.core.util import TriggerBuilder

#####from org.openhab.core.automation import Trigger
#####from org.eclipse.smarthome.config.core import Configuration
#####from org.eclipse.smarthome.core.thing import ChannelUID, ThingUID, ThingStatus
from org.eclipse.smarthome.core.thing.type import ChannelKind
#####from org.eclipse.smarthome.core.types import TypeParser

# DirectoryEventTrigger
from org.eclipse.smarthome.core.service import AbstractWatchService

# JythonTransform
from org.eclipse.smarthome.core.transform import TransformationService

# JythonThingTypeProvider
#####from org.eclipse.smarthome.core.thing.binding import ThingTypeProvider

# JythonThingProvider
#####from org.eclipse.smarthome.core.thing import ThingProvider

# JythonItemChannelLinkProvider
from org.eclipse.smarthome.core.thing.link import ItemChannelLinkProvider

# JythonBindingInfoProvider
from org.eclipse.smarthome.core.binding import BindingInfoProvider

# personal.utils
from org.eclipse.smarthome.model.script.actions.Exec import executeCommandLine
from org.eclipse.smarthome.model.persistence.extensions import PersistenceExtensions

# core.actions
from org.openhab.core.scriptengine.action import ActionService
from org.eclipse.smarthome.model.script.engine.action import ActionService
```
[...]

PR #622, opened 2019-03-04
[Automation] GenericCronTrigger

davidgraeff:
A cron trigger module is actually not that "Generic" and a useful trigger type for a lot of rules.

Therefore I suggest to remove the visibility "HIDDEN" with this PR.
I have also added a context "cronexpression" to the TEXT configuration value, so that UI's can offer a cron expression
selection widget.
(we currently have: 'time','item','script','channel','dayOfWeek','rule')

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/resources/ESH-INF/automation/moduletypes/GenericCronTrigger.json
+++ b/bundles/org.openhab.core.automation/src/main/resources/ESH-INF/automation/moduletypes/GenericCronTrigger.json
@@ -4,11 +4,11 @@
          "uid":"timer.GenericCronTrigger",
          "label":"Cron Trigger",
          "description":"This triggers a rule based on a cron expression",
-         "visibility":"HIDDEN",
          "configDescriptions":[  
             {  
                "name":"cronExpression",
                "type":"TEXT",
+               "context":"cronexpression",
                "label":"cron expression",
                "description":"the cron expression",
                "required":true


openhab-5iver:
@davidgraeff, this trigger wasn't available to be used in Paper UI rules, but it is definitely used in scripts. I don't see much risk to this change, but the only way to test it would be to use the raw API, until #610 is resolved. Or have you already tested this? But [if I'm reading this right](https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigDescriptionParameter.java#L323), 'cronexpression' would not be a valid value for context. I haven't looked to see how this is validated, but would 'time' suit your purpose?

davidgraeff:
@openhab-5iver The context field is a simple string, only relevant for UIs and nowhere validated.
The enum values need to be documented somewhere though, I agree.

I'll add the missing "channel", "rule", "item" and also "cronexpression" in a follow-up PR to the javadoc comment.

"time" is a hint to UIs to render a date/time-picker. For cron expressions we want a cron expression generator widget instead.

I will also create a PR later on to remove "dayOfWeek" and fill the options array with "monday,tuesday,..." (whatever the actual entry values will be, I need to look that up in the code).

> Or have you already tested this 

Yes, I'm using rules in Paper UI NG.

davidgraeff:
I have created https://github.com/openhab/openhab-core/pull/623 to address the missing context values.

davidgraeff:
> context of 'script'

I know. And I'm using that value in Paper UI NG to render the script editor as well. But it is not obvious which scripting language the configuration parameter contains. I have two solutions so far.

1. Remove "script" and introduce "script_javascript" and "script_python" etc.
2. Keep "script" but add a shebang or any other mean to the text value so that a UI can identify what scripting language is used.

PR #623, opened 2019-03-04
ConfigDescriptionParameter: javadoc getContext()

davidgraeff:
Improve javadoc by explaining the different defined context values. 
I have added "rule", "channel", "channeltype" and "cronexpression" as those are used within the automation module.


davidgraeff:
> Also, would it make sense to add a small note to the top of the constructor doc

Done

PR #527, opened 2019-02-04
[WIP] Fix thing creation for updated XML

J-N-K:
was: https://github.com/eclipse/smarthome/pull/6537

Signed-off-by: Jan N. Klug <jan.n.klug@rub.de>

PR #628, opened 2019-03-05
[WIP] Fix thing creation for updated XML

afuechsel:
This is a work-in-progress improved implementation as a basis for a deeper discussion of #527 (originally eclipse/smarthome#6537). The `ManagedItemProvider` now manages its own list of things. 

Note, that this is no final implementation, it still has some flaws. 

Signed-off-by: Andre <afuechsel@gmail.com>

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ManagedThingProvider.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ManagedThingProvider.java
@@ -20,21 +20,27 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 import org.eclipse.jdt.annotation.NonNull;
 import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.smarthome.config.core.Configuration;
+import org.eclipse.smarthome.core.common.ThreadPoolManager;
 import org.eclipse.smarthome.core.common.registry.DefaultAbstractManagedProvider;
 import org.eclipse.smarthome.core.service.ReadyMarker;
 import org.eclipse.smarthome.core.service.ReadyMarkerFilter;
 import org.eclipse.smarthome.core.service.ReadyService;
 import org.eclipse.smarthome.core.storage.StorageService;
+import org.eclipse.smarthome.core.thing.binding.ThingHandler;
 import org.eclipse.smarthome.core.thing.binding.ThingHandlerFactory;
 import org.eclipse.smarthome.core.thing.binding.builder.ThingBuilder;
 import org.eclipse.smarthome.core.thing.type.ChannelTypeUID;
 import org.eclipse.smarthome.core.util.BundleResolver;
 import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Deactivate;
 import org.osgi.service.component.annotations.Reference;
 import org.osgi.service.component.annotations.ReferenceCardinality;
 import org.osgi.service.component.annotations.ReferencePolicy;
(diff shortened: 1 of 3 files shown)


afuechsel:
@J-N-K would be nice to get your thoughts about this implementation as it was originally your PR that inspired me.

bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ManagedThingProvider.java
```diff
@@ -41,6 +75,159 @@ protected String keyToString(ThingUID key) {
...
+            }
+        } else if (bAcceptedItemType != null) {
+            return false;
+        }
+
+        return true;
+    }
+
+    private void createThingFromStorageForThingHandlerFactory(Thing storedThing, ThingHandlerFactory factory) {
+        logger.debug("createThingFromStorageForThingHandlerFactory({}, {}", storedThing.getUID(),
```
afuechsel: This logger is just for deeper debugging purposes and will be removed finally.

bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ManagedThingProvider.java
```diff
@@ -41,6 +75,159 @@ protected String keyToString(ThingUID key) {
...
+                factory.getClass().getName());
+        if (!loadedXmlThingTypes.contains(getBundleName(factory))) {
+            return;
+        }
+
+        logger.info("*** thing.getHandler() = {}", storedThing.getHandler());
```
afuechsel: This logger revealed the main issue with this approach: Since the order of ReadyMarker-Calls is not guaranteed, the following could happen, that the RadyMarker of the ThingManager fires first, then the handler is assigned and initalized and everything is fine here. But it could also happen, that the ReadyMarker here fires first. Then we do not have a handler (yet). So any update of the thing would not cause the handler to update the thing, as there is no handler. This ends up in an inconsistent model, where `thing.getHandler() != thing` inside the eventually assigned handler.

J-N-K: What about queuing that for later re-try? The GenericThingProvider also queues thing creation when not all XMLs are loaded. So if `handler==null` Add to queue and try if the handler is available let‘s say every 500ms or so.

bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ManagedThingProvider.java
```diff
@@ -41,6 +75,159 @@ protected String keyToString(ThingUID key) {
...
+                            thingBuilder.withoutChannel(factoryChannelUID).withChannel(factoryChannel);
+                            logger.trace("updated channel {} in thing {}", factoryChannel, storedThing.getUID());
+                        }
+                    }
+                });
+
+                // check if we have removed channels
+                storedChannels.forEach((storedChannelUID, storedChannel) -> {
+                    if (!factoryChannels.containsKey(storedChannelUID)) {
+                        thingBuilder.withoutChannel(storedChannelUID);
```
J-N-K: The reason why I didn’t implement that line is: it will remove user-added channels. As far as I know there is no way to find out if an additional channel was added by the user as `extensible` or if it was present in the XML before and was removed now.

bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ManagedThingProvider.java
afuechsel: I was also thinking about this but first searhed for a simpler solution.

bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ManagedThingProvider.java
afuechsel: I didn't think about user added channels to be honest. Maybe we could mark the channels somehow whether they are user created?

J-N-K: That's exactly what I was looking for in my initial comment in [#6537](https://github.com/eclipse/smarthome/pull/6537) but I didn't find it. 

Maybe `Channel` could get a `private boolean userAddedExtensible=false;` with additional setter/getter. This could be set to true if the user adds a channel via UI, serialized for storage and read back from the database later.

davidgraeff: You have the thing description that tells you about extensible channels. Just filter them out.

J-N-K: We would need to add the ThingTypeRegistry then. But that is also fine for the `GenericThingProvider`.

afuechsel:
Hmm, spent the whole day to play around with this stuff but no clear solution. Added a queue to process things without handler later but this did not solve the main issues. Several tests are still failing, mainly, because the lifecycle of a thing in the ManagedThingProvider now completely changed, but others are also failing and I don't know why. I feel very uncomfortable with this way to go and want to step back a bit and re-think, if there is a simpler and cleaner solution. Ideas welcome.

maggu2810:
@afuechsel Can this be closed as long as you re-think?

afuechsel:
Yes, I will close it ftb. I will come up soon with a new approach.

PR #633, opened 2019-03-06
Do not use Paths to concatenate resource paths

afuechsel:
On Windows `Paths.get().toString()` will correctly convert "/" to "\\" but this is not a valid resource path for `bundle.getResource()` and the resource will not be found. 

Signed-off-by: Andre <afuechsel@gmail.com>


openhab-5iver:
Is [this another one](https://github.com/openhab/openhab-core/blob/68efabe63ca4785acd629771921e70098147d217/itests/org.openhab.core.model.core.tests/src/main/java/org/eclipse/smarthome/model/core/internal/folder/FolderObserverTest.java#L328) that should be fixed? I didn't find any others.

afuechsel:
Don’t  think so. Here the path is not used as a resource identifier but as a file. This should work.

PR #635, opened 2019-03-07
Added ScriptModuleTypeProvider for Jython and Groovy scripted Actions and Conditions

### LINKED ISSUE #521: [Automation - Feature Request] Add Jython and Groovy scripted Actions to Paper UI
The drop down currently only has Javascript.

openhab-5iver:
Fixes #521. This PR replaces #519. As discussed there, this PR will provide a dynamic list of script languages when using ScriptAction or ScriptCondition, based on the ScriptEngines in use on the system. It implements the ScriptModuleTypeProvider, which replaces the ScriptTypes ResourceBundle. I do not see anything that would make this a breaking change. The ParameterOption labels and values for the currently known working script languages are...

|options | |
-- | --
0 |  
label | "ECMAScript (ECMA - 262 Edition 5.1)"
value | "application/javascript"
1 |  
label | "Groovy (2.4.12)"
value | "application/x-groovy"
2 |  
label | "Python (2.7)"
value | "application/python"

**Things to be aware of:**
1) The dropdown will not have Javascript as an option label, but ECMAScript (ECMA - 262 Edition 5.1). 
2) When an engine is upgraded, the label will change to the upgraded version, but no change will be required by the user, as the value of the option should not have changed. I've tested this by upgrading/downgrading the package.
3) This PR removes the NashornScriptEngineFactory, as it was a duplicate of GenericScriptEngineFactory after this PR, except for how the scopeValues() method was implemented. I did not see any need for both files, so I added special handling of ECMAScripts in GenericScriptEngineFactory.scopeValues(). It's a little strange to put something specific in something names Generic, but this elimiated a file and a lot of duplicated code.
4) This provider is setup to use localization, but it is not yet implemented.

@kaikreuzer, you had suggested a ConfigOptionProvider, but I could not find a way to add one to a ResourceBundle. I based this provider off of your MediaActionTypeProvider, in which you state...

> This is necessary since there is no other way to provide dynamic config param options for module types

ConfigOptionProvider had been around for about a year when you wrote this, so I figured you would have used one if it was possible. That is why I went this route.

@openhab/core-maintainers, please review. I've tested this, and everything seemed to function properly, but I don't think of myself as much of a Java developer. :wink:

Signed-off-by: Scott Rushworth openhab@5iver.com (github: openhab-5iver)


bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptEngineFactory.java
```diff
@@ -1,8 +1,8 @@
 /**
- * Copyright (c) 2014,2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2010-2019 Contributors to the openHAB project
```
maggu2810: Could you rebase your branch.
This change should already be part of master.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptEngineFactory.java
```diff
@@ -14,24 +14,30 @@
...
- *
+ * @author Scott Rushworth - added getLanguageNames() for use in ScriptModuleTypeProvider
  */
 public interface ScriptEngineFactory {
 
     /**
-     * @return the list of supported language endings e.g. py, jy
+     * @return the list of supported language endings, e.g. py, groovy, js
      */
     List<String> getLanguages();
```
maggu2810: Sounds like the "file extensions" (WRT your `getLanguageNames()` comment).

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptEngineFactory.java
```diff
@@ -14,24 +14,30 @@
...
 
+    /**
+     * @return a map of supported language name : file extension, e.g. {Python=py, Groovy=groovy, ECMAScript=js}
+     */
+    TreeMap<String, String> getLanguageNames();
```
maggu2810: Hm, I don't think "get language names" should return more that a language name.
The language name would be Python, Groovy, ECMAScript (WRT your comment).
Perhaps this is more about a language file extendsion mapping?

It is very uncommon to use the specific implementation of a map in an interface.
It should be enough to state a map will be returned. Which one should not matter.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptEngineManager.java
```diff
@@ -32,6 +34,11 @@
      */
     boolean isSupported(String fileExtension);
 
+    /**
+     * Gets the names and MimeTyoes of the supported languages
+     */
+    TreeMap<String, String> getSupportedLanguageNames();
```
maggu2810: see above.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptEngineFactory.java
maggu2810: Please fix your JavaDoc comments.
A comment should start with a short description, empty line, param(s), return and throws statements.

(I comment this only once, it fits for the other methods, too)

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
```diff
@@ -68,6 +71,14 @@ public void addScriptEngineFactory(ScriptEngineFactory provider) {
         for (String language : provider.getLanguages()) {
             this.supportedLanguages.put(language, provider);
         }
+
+        TreeMap<String, String> languageNames = provider.getLanguageNames();
+
+        for (Map.Entry<String, String> entry : languageNames.entrySet()) {
+            if (!this.supportedLanguageNames.containsKey(entry.getKey())) {
```
maggu2810: Use [computeIfAbsent](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#computeIfAbsent-K-java.util.function.Function-)

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
```diff
@@ -76,6 +87,19 @@ public void removeScriptEngineFactory(ScriptEngineFactory provider) {
         for (String language : provider.getLanguages()) {
             this.supportedLanguages.remove(language, provider);
         }
+
+        TreeMap<String, String> languageNames = provider.getLanguageNames();
+
+        for (Map.Entry<String, String> entry : languageNames.entrySet()) {
+            if (this.supportedLanguageNames.containsKey(entry.getKey())) {
```
maggu2810: see above

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
```diff
@@ -76,6 +87,19 @@ public void removeScriptEngineFactory(ScriptEngineFactory provider) {
...
+                this.supportedLanguageNames.remove(entry.getKey(), entry.getValue());
+            }
+        }
+    }
+
+    @Override
+    public TreeMap<String, String> getSupportedLanguageNames() {
+        return this.supportedLanguageNames;
```
maggu2810: Do not return your internal map if you do not use a non modifiable collection.
The caller can modify the map and this could break you and other consumers.

e.g. `getSupportedLanguageNames().put("foo", "bar");`

openhab-5iver:
> If you would like to remove duplicate code, then we should improve the generic implementation to allow the specific child to modify / implement only the special scopeValues part without code duplication of all the remaining one.

To expedite things, I'll add NashornScriptEngineFactory back in. That way it is no worse off than it was. Cleaning up the duplicated code can then be done in a separate PR.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptEngineFactory.java
openhab-5iver: I had put this in to remove the conflict, since I knew the header PR would be merged before mine, but I've rebased.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptEngineFactory.java
openhab-5iver: This method returns a list of file extensions _and_ MimeTypes. ScriptEngineManagerImpl takes the list and makes a Map out of it, adding in the ScriptEngines associated with each. I'll update the doc with more detail. 

getLanguageNames returns a Map with one name and one MimeType, for use in the ParameterOption when selecting a scripting language in the UI. I tried to combine these methods, but it was cleaner just having both.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptEngineFactory.java
openhab-5iver: I had a hard time trying to decide on a method name. getLanguages wasn't very descriptove, but I chose getLanguageNames to differentiate between the two. I agree... both of them are not clear what they do. How about scriptEngineIdentifiers and scriptEngineNameMap?

I will clean up the TreeMaps. Java is new to me, so I'm going to make some beginner mistakes. Thank you for the direction.

> Please fix your JavaDoc comments.

I'll look at them again. I was mainly following the pattern of what was already in the file.

openhab-5iver:
@maggu2810, I've reworked this PR, so hopefully all of your comments have now been addressed. I'm not looking forward to wait a couple weeks for a review, so I'm hoping @wborn, @cweitkamp, or @kaikreuzer might take a look at it. Hope your vacation is going well! I have tested this PR for a while now, and am running it on my production system without issues. 

* Completely changed how the ParameterOptions were being generated in the previous commit
* Added GenericScriptEngineFactory as a service. ESH #[6284](https://github.com/eclipse/smarthome/pull/6284) only setup NashornSEF as a service.
* Readded NashornSEF
* Added an AbstractSEF for the others to inherit from, in order to remove duplication of code in the SEFs. This is the one change I questioned most, as I was uncertain if there might be issues with making services from both a parent and child class. If there are no issues, then the other factories can be extended from GenericSEF. I will keep looking at this.
* Refactored ScriptEngineManagerImpl a bit, utilizing the fact that all factories are now services. 
* Moved the instantiation of the javax.script.ScriptEngineManagers from the factories to a static in the ScriptEngineFactory interface

I have noticed something though, which also occurs in the current production code. When OH starts up, the SEFs, or the ScriptEningManagerImpl, are started twice. Every now and then, after clearing the cache, this does not happen. The code seems to handle this, but I'm not sure what's causing it yet. This is unrelated to this PR, but I thought I'd mention it in case someone noticed it or had an explanation for it.

maggu2810:
Thank you for all your work and effort.
This PR is still part of the things I need to do.
Just the usual "spare time problem".
Please give me some time.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptEngineFactory.java
```diff
@@ -16,44 +16,49 @@
...
-     * @return
+     * @param scriptEngine
+     * @param scopeValues
      */
-    ScriptEngine createScriptEngine(String fileExtension);
+    void scopeValues(ScriptEngine scriptEngine, Map<String, Object> scopeValues);
 
     /**
-     * checks if the script is supported. Does not necessarily be equal to getLanguages()
+     * This method creates a new ScriptEngine based on the supplied file extension or MimeType.
```
maggu2810: The JavaDoc should declare how a invalid / unknown `scriptType` is handled.
Is `null` returned or is an `IllegalArgumentException` thrown?

If `null` is returned, it would make sense to add `@NonNullByDefault` to the class and add `@Nullable`

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/NashornScriptEngineFactory.java
```diff
@@ -12,74 +12,67 @@
...
+                scriptTypes.addAll(f.getMimeTypes());
             }
         }
-        String scriptToEval = String.join("\n", expressions);
-        try {
-            engine.eval(scriptToEval);
-        } catch (ScriptException e) {
-            logger.error("ScriptException while importing scope: {}", e.getMessage());
-        }
+        logger.trace("getScriptTypes(): {}", scriptTypes);
```
maggu2810: I assume this logging can be removed.
If we want to be informed about the script types, why not moving this logging statement to the caller, so we get that information for all implementations?

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/NashornScriptEngineFactory.java
```diff
@@ -12,74 +12,67 @@
...
+        return scriptTypes;
     }
 
     @Override
-    public ScriptEngine createScriptEngine(String fileExtension) {
-        ScriptEngine engine = engineManager.getEngineByExtension(fileExtension);
+    public void scopeValues(ScriptEngine scriptEngine, Map<String, Object> scopeValues) {
+        Set<String> expressions = new HashSet<String>();
```
maggu2810: You can change the second `<String>` to `<>`.

I would expect the save action called formatter replaces such stuff.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
```diff
@@ -62,63 +59,73 @@ public void unsetScriptExtensionManager(ScriptExtensionManager scriptExtensionMa
...
     }
 
-    @Override
-    public boolean isSupported(String fileExtension) {
-        return findEngineFactory(fileExtension) != null;
+    /**
+     * This method is used to determine is a given {@link ScriptEngineFactory} is generic or customized.
+     *
+     * @param engineFactory {@link ScriptEngineFactory}
+     * @return boolean
```
maggu2810:
```
@return true if the factory is a custom one otherwise true
```

IMHO it does not bring any value to state which data type is returned.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
```diff
@@ -62,63 +59,73 @@ public void unsetScriptExtensionManager(ScriptExtensionManager scriptExtensionMa
     }
 
     @Reference(cardinality = ReferenceCardinality.MULTIPLE, policy = ReferencePolicy.DYNAMIC)
-    public void addScriptEngineFactory(ScriptEngineFactory provider) {
-        this.scriptEngineFactories.add(provider);
-
-        for (String language : provider.getLanguages()) {
-            this.supportedLanguages.put(language, provider);
+    public void addScriptEngineFactory(ScriptEngineFactory engineFactory) {
+        engineFactory.logFactoryDetails();
```
maggu2810: I don't think we whould handle it this way.

If I would like to see the factory details of the added script engine factories in the log, I don't know how the logger needs to be configured.

The factory logs the details, so I don't know for which package I need to set the log level.
I also don't know which log level is used at all.
The factory A will e.g. use trace, the factory B will use debug another one info, ...

If we need that factory details for further information in the log at all, the `ScriptEngineFactory` should define a method that returns the factory details (e.g. a string).

Then this class (the engine manager) could define the log level. It will be equal for every implementation.
And we knows for which package we need to set the log level.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/provider/ScriptModuleTypeProvider.java
```diff
@@ -0,0 +1,157 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class dynamically provides ScriptActionType and ScriptConditionType {@link ModuleType}s. This class is necessary
+ * because there is no other way to provide dynamic {@link ParameterOption}s for {@link ModuleType}s.
+ *
+ * @author Scott Rushworth - Initial contribution
+ */
+@Component(immediate = true)
```
maggu2810: I assume there is a consumer of the `ModuleTypeProvider` components and if there is no one this service does not bring any value.

So, do you see any reason we need to use immediate here?
I assume it can (and if it can it should) be removed.

davidgraeff: See also https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/provider/file/ModuleTypeFileProvider.java#L30 as another example

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptEngineFactory.java
openhab-5iver: This was an existing method that I renamed and slightly modified, but I understand the need for better documentation. 

Worst case, [javax.script.ScriptEngineManager.getEngineFactories()](https://docs.oracle.com/javase/8/docs/api/javax/script/ScriptEngineManager.html#getEngineFactories--) will return an empty list and does not throw anything, as does [javax.script.ScriptEngineFactory.getExtensions()](https://docs.oracle.com/javase/8/docs/api/javax/script/ScriptEngineFactory.html#getExtensions--) and [javax.script.ScriptEngineFactory.getMimeTypes()](https://docs.oracle.com/javase/8/docs/api/javax/script/ScriptEngineFactory.html#getMimeTypes--), causing getScriptTypes() to return an empty list. If there is no ScriptEngineManager, there are some serious problems and I can't imagine we'd get far enough to call this method. So, this looks to me like getScriptTypes() would never return null.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/NashornScriptEngineFactory.java
openhab-5iver: This logging was handy in testing/debugging, so I left it in. The method is called twice in ScriptEngineManagerImpl, to generate an interable when adding and removing SEFs (rare)...
```
for (String scriptType : engineFactory.getScriptTypes()) {
```
So, I'll need to set a variable and pull it out of there to do the logging. It won't take much, but is it worth it to move the logging to the caller?

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/NashornScriptEngineFactory.java
openhab-5iver: Done. This was the original code... it only showed in the diff because I moved the method.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
openhab-5iver: Done.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/provider/ScriptModuleTypeProvider.java
maggu2810: @davidgraeff I'm not sure what this example is supposed to show.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptEngineFactory.java
maggu2810: `Collections.emptyList()` :wink:

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/NashornScriptEngineFactory.java
maggu2810: Sure, this one logging statement does not really harm.
But there will be a lot of them in the whole codebase if we keep all the "has been useful on testing" in the code.

The other point is:
There is an interface `ScriptEngineFactory`.
This factories are used by "someone".
An implementation needs to provide a `getScriptTypes()` method.
We provide an `AbstractScriptEngineFactory` that contians an implementation of that method and adds some logging.
If an implementation does not use the abstract base class there is no logging or a different one.
So, if we would like to be able to be informed about all the script types in the runtime, this works only for the subclasses of the abstract class but not for other implementations as long as the logging is in the abstract class.
If there is no need to get function calls traced anymore in production because it does not bring any value (as we know the method is working), IMHO we should remove it.
To use trace to detect some wrong behaviour on runtime systems of customers, I would like the same behaviour for all `ScriptEngineFactory` implementations (regardless if abstract base class is used) -- so my comment about moving it to the caller.

So, do you think we need to trace the if the abstract class is able to collect its types or which types are received by the caller.

openhab-5iver: > So, if we would like to be able to be informed about all the script types in the runtime, this works only for the subclasses of the abstract class but not for other implementations as long as the logging is in the abstract class.

That makes perfect sense. I think this logging could help in troubeshooting issues, so I will move it to the caller.

davidgraeff: Hm. If you start openhab with logging level INFO you will see exactly nothing in the resulting 1MB log file. If openHAB does not yet have a way to tell about the status of its services, then we should tackle that (I suggested a service status like a thing status once). But adding log statements makes the situation worse. Just my 2ct.

maggu2810: It is common practice to use debug and trace log to be able to diagnose some stuff at runtime.
This is not openHAB Core specific.
Some of that logging really makes sense.

davidgraeff: Most services are state machines though and it would be sufficient to be able to retrieve the current (error) state. In my vision of openHAB I would never need to go through logs, except to look up historic unexpected problems (errors), but not to find out why a particular thing does not work right now.

This is unrelated to the PR though.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/provider/ScriptModuleTypeProvider.java
openhab-5iver: > I assume there is a consumer of the ModuleTypeProvider components 

I've seached for one, but haven't found it. But when the component annotation is removed, the ModuleTypes do not show up in the REST API.

> So, do you see any reason we need to use immediate here?

For this file, I followed most of the patterns in [MediaActionTypeProvider](https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.automation.module.media/src/main/java/org/openhab/core/automation/module/media/internal/MediaActionTypeProvider.java#L48), which uses immediate. Same for [ScriptedCustomModuleTypeProvider](https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/ScriptedCustomModuleTypeProvider.java#L33). The ModuleTypes will also need to be available before any rules are loaded that require them.

However, on my test system, I did not see any issues removing immediate. I'll try it tonight on my production system.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
openhab-5iver: > If I would like to see the factory details of the added script engine factories in the log, I don't know how the logger needs to be configured.
> 
> The factory logs the details, so I don't know for which package I need to set the log level.
> I also don't know which log level is used at all.
> The factory A will e.g. use trace, the factory B will use debug another one info, ...

I don't follow your logic. These logs will all be coming in from under org.openhab.core.automation.module.script, so set the log level there. But even this is too granular to see much going on, so I usualy have org.openhab.core.automation set to DEBUG or TRACE.

As per the discussion about getScriptTypes(), I have moved this logging out of the SEFs and into the SEMImpl.

maggu2810:
@openhab-5iver Please ping me if you think all is done and I should start another review round.

openhab-5iver:
@maggu2810, will do! A few more tests and I'll commit the changes.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/provider/ScriptModuleTypeProvider.java
openhab-5iver: With immediate removed, my production system did not have an issue. I will remove this.

openhab-5iver:
@maggu2810, it's ready... and thank you!

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
openhab-5iver: Done

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
openhab-5iver: This was removed.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
openhab-5iver: This was removed.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptEngineManager.java
openhab-5iver: This was removed.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/provider/ScriptModuleTypeProvider.java
```diff
@@ -0,0 +1,157 @@
...
+ *
+ * @author Scott Rushworth - Initial contribution
+ */
+@Component
+public class ScriptModuleTypeProvider implements ModuleTypeProvider {
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+    private final List<ParameterOption> parameterOptions = new CopyOnWriteArrayList<>();
+
+    @SuppressWarnings("unchecked")
```
maggu2810: What is unchecked here?

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/ScriptConditionHandler.java
```diff
@@ -33,9 +33,9 @@
  */
 public class ScriptConditionHandler extends AbstractScriptModuleHandler<Condition> implements ConditionHandler {
 
-    public final Logger logger = LoggerFactory.getLogger(ScriptConditionHandler.class);
+    public static final String TYPE_ID = "script.ScriptCondition";
 
-    public static final String SCRIPT_CONDITION = "script.ScriptCondition";
+    public final Logger logger = LoggerFactory.getLogger(ScriptConditionHandler.class);
```
maggu2810: Hm, public loggers are a uncommon thing, but it has been present before, so no need to change anything at this time.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptEngineManager.java
```diff
@@ -16,47 +16,48 @@
...
-    void loadScript(String scriptIdentifier, InputStreamReader scriptData);
+    void removeEngine(String engineIdentifier);
 
     /**
-     * Unloads the ScriptEngine loaded with the scriptIdentifer
+     * Checks if the supplied file extension or MimeType is supported by the existing ScriptEngineFactories
      *
-     * @param scriptIdentifier
+     * @param scriptType a file extension (script) or MimeType (ScriptAction or ScriptCondition)
+     * @return boolean
```
maggu2810: IIRC this has been already comment. The JavaDoc should contain a description of the return value, not the type.
Why not using `true if supported` is in the old `isSupported` method?

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptEngineManager.java
```diff
@@ -16,47 +16,48 @@
...
 @NonNullByDefault
 public interface ScriptEngineManager {
 
     /**
-     * Checks if a given fileExtension is supported
+     * Creates a new ScriptEngine used to execute scripts, ScriptActions or ScriptConditions
      *
-     * @param fileExtension
-     * @return true if supported
+     * @param scriptType       a file extension (script) or MimeType (ScriptAction or ScriptCondition)
```
maggu2810: Please excuse me, it is really not your fault, but perhaps you can correct it...
There is something wrong with the IDE setup, the formatter should be selected but it seems there is a bug.

Go to your IDE settings, search for formatter, select the Java formatter, choose another formatter then "ESH", and then switch back to "ESH". Apply & close, reformat this file...

There should be no spaces after e.g. `scriptType` and its description.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptEngineManager.java
```diff
@@ -16,47 +16,48 @@
...
 
     /**
-     * Creates a new ScriptEngine based on the given fileExtension
+     * Loads a script and initializes its scope variables
      *
-     * @param fileExtension
-     * @param scriptIdentifier
-     * @return
+     * @param engineIdentifier the unique identifier for the ScriptEngine (script file path or UUID)
+     * @param scriptData       the content of the script
```
maggu2810: See above

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptEngineFactory.java
```diff
@@ -16,44 +16,45 @@
...
      *
-     * @param fileExtension
-     * @return
+     * @param scriptType a file extension (script) or MimeType (ScriptAction or ScriptCondition)
+     * @return ScriptEngine
      */
-    boolean isSupported(String fileExtension);
+    ScriptEngine createScriptEngine(String scriptType);
```
maggu2810: I still don't understand why the return value is not annotated with `@Nullable`.

What will be returned if i call `createScriptEngine("someVeryStupidString");`

As the method contains `@NonNullByDefault` and you do not annotate the return type, you state that there is no code path that will result into a return of `null`.
If a non existing / unknown `scriptType` is used, do you throw an `IllegalArgumentException` instead of returning `null`? But then it needs to be commented, too.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
```diff
@@ -62,63 +57,83 @@ public void unsetScriptExtensionManager(ScriptExtensionManager scriptExtensionMa
...
-            this.supportedLanguages.put(language, provider);
+        logger.trace("{}.getScriptTypes(): {}", engineFactory.getClass().getSimpleName(), scriptTypes);
+        for (String scriptType : scriptTypes) {
+            if (isCustomFactory(engineFactory)) {
+                this.customSupport.put(scriptType, engineFactory);
+            } else {
+                this.genericSupport.put(scriptType, engineFactory);
+            }
+        }
+        logger.info("Added {}", engineFactory.getClass().getSimpleName());
```
maggu2810: Reduce this from `info` to `debug` or `trace`.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
```diff
@@ -62,63 +57,83 @@ public void unsetScriptExtensionManager(ScriptExtensionManager scriptExtensionMa
...
-            this.supportedLanguages.remove(language, provider);
+        logger.trace("{}.getScriptTypes(): {}", engineFactory.getClass().getSimpleName(), scriptTypes);
+        for (String scriptType : scriptTypes) {
+            if (isCustomFactory(engineFactory)) {
+                this.customSupport.remove(scriptType, engineFactory);
+            } else {
+                this.genericSupport.remove(scriptType, engineFactory);
+            }
         }
+        logger.info("Removed {}", engineFactory.getClass().getSimpleName());
```
maggu2810: Reduce this from `info` to `debug` or `trace`.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/provider/ScriptModuleTypeProvider.java
openhab-5iver: Lol... I was wondering the same thing earlier today! I copied it over in the code I copied from [MediaActionModuleTypeProvider](https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.automation.module.media/src/main/java/org/openhab/core/automation/module/media/internal/MediaActionTypeProvider.java#L53).

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/ScriptConditionHandler.java
openhab-5iver: Noted! I forsee a @Null annotation cleanup coming for automation, and can clean up some of the other small stuff at that time.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptEngineManager.java
openhab-5iver: I fixed the other but forgot this one. I'll fix this...

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/provider/ScriptModuleTypeProvider.java
openhab-5iver: Type safety: The return type ModuleType for getModuleType(String, Locale) from the type ScriptModuleTypeProvider needs unchecked conversion to conform to T from the 
	 type ModuleTypeProvider

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/ScriptConditionHandler.java
openhab-5iver: Done

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptEngineManager.java
openhab-5iver: Done

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptEngineManager.java
openhab-5iver: Done

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptEngineManager.java
openhab-5iver: Done

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptEngineFactory.java
openhab-5iver: Done for now, but there's are a lot of places to fix up the null annotations in automation!

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
openhab-5iver: Done

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
openhab-5iver: Done

openhab-5iver:
@maggu2810, thank you very much for your patience and the time you have spent working with me. I have learned a lot from this! I've pushed the latest changes.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/provider/ScriptModuleTypeProvider.java
maggu2810: IMHO the API of `ModuleTypeProvider` needs to be changed. It does not make sense to use a unbound return value.
It will be unknown in any case...

We should improve that API.

PR #637, opened 2019-03-07
add missing synchronisation in Configuration

mlriess:
The Configuration class was synchronising all access to its properties member except for two methods, which could lead to data loss or ConcurrentModificationExceptions. This PR adds the missing synchronisation.


mlriess:
I changed too much ... will open another PR with fewer changes.

PR #639, opened 2019-03-07
remove PDE left overs from core.karaf project

maggu2810:


wborn:
@kaikreuzer do you know if the Jenkins tmp space is exhausted causing build issues? 

The logging in several builds seems to suggest so:

```
Java HotSpot(TM) 64-Bit Server VM warning: Insufficient space for shared memory file:
   28603
Try using the -Djava.io.tmpdir= option to select an alternate temp location.
```

kaikreuzer:
> @kaikreuzer do you know if the Jenkins tmp space is exhausted causing build issues?

Seems so. Will reach out to you soon for help on the docker configuration...
Cleaned it now and restarted Jenkins.

PR #640, opened 2019-03-07
specify directories so we do not rely on the current working directory

maggu2810:
The current working directory has been changed between Bnd 4.1.0 and 4.2.0 and we should not rely on a specific choosen one.
We should set the appropriate directories ourself.


maggu2810:
IMHO superseded by introduction of a separate demo project:
https://github.com/openhab/openhab-distro/pull/889#issuecomment-471050852

maggu2810:
As long as the discussion is in progress we should improve the demo app provided by this repo.
Otherwise it will be broken with the current Bnd release which we should bump ASAP.

Issue #608, opened 2019-02-25
Gson library version


paulianttila:
Used gson library v2.3.1 is pretty old. Can we bump it to latest which is v2.8.5?

I need at least v2.4 to be able to provide binary to json converter openhab/openhab2-addons#4979

API change review: https://abi-laboratory.pro/index.php?view=timeline&lang=java&l=gson

maggu2810:
At which place you would like to bump the Gson version.

The openhab-runtime-base feature provides 2.8.2 at runtime.

```
  <feature name="openhab-runtime-base" description="openHAB Runtime Base" version="${project.version}">
    <bundle>mvn:p2.osgi.bundle/com.google.gson/2.8.2.v20180104-1110</bundle>
```

openHAB Core uses v2.3.1 at compile time for their bundles.
But I don't see any reason to bump that version just because an addon needs a newer one -- it should be enough to use a newer one while compiling the addon.
The requirements at runtime for OSGi bundles are part of the manifest.

I assume you would like to request a bump of the Gson that is provided by the target platform because you use that bundles as compile time dependencies.

The target platform is (AFAIK) provided by https://github.com/openhab/openhab-deps-repo

maggu2810:
Sorry, didn't have a look at the PR itself and its comments.

So, you would like to migrate a bundle to core and the Gson version used in openHAB Core does not fulfill the requirements of your code.

Can you please check the minimum version that suffers your needs you need.

If we bump the compile time dependency used in the core (the file bom/compile/pom.xml) this will result into a new minimum version required by the bundle.

If a product currently uses e.g. 2.7.0 it will not work if we bump the minimum version from 2.3.1 to 2.7.1 or 2.8.5.

The minimum version should be set to the real minimum and not just to the newest one available.

I am personally fine with using a more recent version but we should choose the minimum version responsibly.

* Which function do you need to use?
* Is there an equivalent function in 2.3.1?
* Which version of Gson started to add this function?

paulianttila:
I think I found a way to be Gson 2.3.1 compatible

PR #641, opened 2019-03-08
[gson] bump gson version to 2.4

Hilbrand:
When I tried to migrate my tplinksmarthome binding to mvn + bnd (because I thought that would be easier to get test running, silly me) I found out it didn't work because the gson version it links to is 2.3.1. Unfortunately my binding uses a feature added in 2.4. Therefor I updated the gson version to the version also uses are runtime version. I looks like I was not the first with this problem #608

### DIFF
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -101,7 +101,7 @@
     <dependency>
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
-      <version>2.3.1</version>
+      <version>2.4</version>
       <scope>compile</scope>
     </dependency>
 
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -380,7 +380,7 @@
     <dependency>
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
-      <version>2.3.1</version>
+      <version>2.4</version>
       <scope>compile</scope>
     </dependency>
     <dependency>
--- a/features/karaf/openhab-tp/src/main/feature/feature.xml
+++ b/features/karaf/openhab-tp/src/main/feature/feature.xml
@@ -31,7 +31,7 @@
     <!--<bundle dependency="true">mvn:org.apache.commons/commons-lang3/3.4</bundle>-->
 
     <!-- Gson -->
-    <bundle dependency="true">mvn:com.google.code.gson/gson/2.3.1</bundle>
+    <bundle dependency="true">mvn:com.google.code.gson/gson/2.4</bundle>
 
     <!-- Measurement -->
     <bundle dependency="true">mvn:javax.measure/unit-api/1.0</bundle>


wborn:
Can't we bump it even higher? We ran into issues because of the older Gson versions before, see also https://github.com/eclipse/smarthome/issues/4440 and https://github.com/eclipse/smarthome/issues/5077.

Hilbrand:
I followed @maggu2810 advice to bump as minimal as possible. (my first version had 2.8.2 :smile: ) So if another version is desired I'll update it.

wborn:
It's a always a dilemma if you want people to spent time on fixing unknown issues (by upgrading) or known and already fixed issues (by not upgrading, see [changelog](https://github.com/google/gson/blob/master/CHANGELOG.md)). :wink: There seem to be a few fixes for supporting newer Java versions as well.

maggu2810:
In our products we already use Gson >= 2.8.0, so personally I am also fine with a bump to a more recent one.
We decided to keep openHAB Core as a framework usable for other products, so in general we should not bump to the most recent version just because it the most simple way for "our" products and we like to break other ones that cannot bump such easily.

I assume ATM there are no many consumers of openHAB Core, so if Kai agrees, too...
... use a 2.8.x version.

kaikreuzer:
> so if Kai agrees, too...

I do!

Hilbrand:
Runtime in feature is 2.8.2 already. Shall I use that version or bump all to 2.8.5?

wborn:
The newer versions do bring some additional Java 9 compatibility fixes that might ease migrating to Java 11. I don't see any major issue reported or already fixed on the gson master for 2.8.5.

maggu2810:
I am fine with 2.8.5, too.
You just need to remember that we need a bundle that exports the internal packages, too.
IIRC the internal ones are used by the ESH code (also by Xtext?).

Ensure the referenced GAV exists on a Maven repo.

Hilbrand:
I've set it to gson version 2.4. I've tried higher versions, 2.7, 2.8.2 and 2.8.5 but all seems to fail to build (at different points). It complains about not finding the gson. It looks like to have do something with the feature version. But this is all a bit magic for me. I found 2.7 and 2.8.2 in eclipse orbit. But I have no idea how that relates or if it relates and/or even how to fix the build problems. It seems to work for 2.4. Which is actually working enough for me. So if someone has a clue to what I'm missing please enlighten me.

BClark09:
Was it complaining about `com.google.gson.internal`?

I'm learning the magic of maven/Java so was interested in this issue, especially if it might easy migrating to Java 11 as @wborn mentioned. That also means you should tell me if I'm just making noise ;-)

From what I took from the eclipse issues @wborn cited above, this project uses an unusual mix of versions from different sources. Versions of gson `2.7.0` onwards do not export the internal packages anymore. 

Having looked at the latest xtext, lsp4j, and jsonpath, it looks like the only place the project uses these packages anymore is in [mapdb](https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.storage.mapdb/src/main/java/org/eclipse/smarthome/storage/mapdb/internal/PropertiesTypeAdapter.java#L25-L27) so it might be worth raising an issue for that to be changed.

maggu2810:
@BClark09 You don't make noise :wink: 

But it is already known that mapdb uses the internal packages, see: https://github.com/eclipse/smarthome/issues/4440#issuecomment-337913346

It would be great if someone is willing to change that.

For the time being, I will bump Gson to 2.8.2 (a PR will follow) that exports the internals (the bundle from Orbit).

wborn:
I'll close this PR since Gson was upgraded to 2.8.2 in https://github.com/openhab/openhab-core/pull/648.

PR #642, opened 2019-03-08
fix synchronisation and equals() in Configuration

mlriess:
We recently encountered a ConcurrentModificationException in the setProperties() method. This PR changes the implementation so that a synchronized Map is used throughout the class, including setProperties() and toString() where the synchronization was previously missing.

While doing that we saw that the equals() method was merely comparing the hashcodes of the maps. This could of course result in equals() returning true for Configurations which are actually different, and thus in subtle bugs. We changed that so that equals() delegates to the equals() method of the HashMap objects.

We also simplified the implementation of the setProperties() method as well as the constructors.

### DIFF
--- a/bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/Configuration.java
+++ b/bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/Configuration.java
@@ -12,14 +12,15 @@
  */
 package org.eclipse.smarthome.config.core;
 
+import static java.util.Collections.*;
+import static org.eclipse.smarthome.config.core.ConfigUtil.normalizeTypes;
+
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.Map;
-import java.util.Map.Entry;
 import java.util.Set;
 
 import org.eclipse.jdt.annotation.Nullable;
@@ -33,12 +34,14 @@ import org.eclipse.smarthome.config.core.internal.ConfigMapper;
  * @author Gerhard Riegler - added converting BigDecimal values to the type of the configuration class field
  * @author Chris Jackson - fix concurrent modification exception when removing properties
  * @author Markus Rathgeb - add copy constructor
+ * @author Michael Riess - fix concurrent modification exception when setting properties
+ * @author Michael Riess - fix equals() implementation
  */
 public class Configuration {
     private final Map<String, Object> properties;
 
     public Configuration() {
-        this(null, true);
+        this(emptyMap(), true);
     }
 
     /**
@@ -50,7 +53,7 @@ public class Configuration {
      * @param configuration the configuration that should be cloned (may be null)
      */
     public Configuration(final @Nullable Configuration configuration) {
-        this(configuration != null ? configuration.properties : null, true);
+        this(configuration == null ? emptyMap() : configuration.properties, true);
     }
 
     /**
@@ -59,7 +62,7 @@ public class Configuration {
      * @param properties the properties the configuration should be filled. If null, an empty configuration is created.
      */
     public Configuration(Map<String, Object> properties) {
-        this(properties, false);
+        this(properties == null ? emptyMap() : properties, false);
     }
 
     /**
@@ -68,20 +71,12 @@ public class Configuration {
      * @param properties the properties to initialize (may be null)
      * @param alreadyNormalized flag if the properties are already normalized
      */
-    private Configuration(final @Nullable Map<String, Object> properties, final boolean alreadyNormalized) {
-        if (properties == null) {
-            this.properties = new HashMap<>();
-        } else {
-            if (alreadyNormalized) {
-                this.properties = new HashMap<>(properties);
-            } else {
-                this.properties = ConfigUtil.normalizeTypes(properties);
-            }
-        }
+    private Configuration(final Map<String, Object> properties, final boolean alreadyNormalized) {
+        this.properties = synchronizedMap(alreadyNormalized ? new HashMap<>(properties) : normalizeTypes(properties));
     }
 
     public <T> T as(Class<T> configurationClass) {
-        synchronized (this) {
+        synchronized (properties) {
             return ConfigMapper.as(properties, configurationClass);
         }
     }
@@ -93,9 +88,7 @@ public class Configuration {
      * @return true if the key is part of the configuration, false if not
      */
     public boolean containsKey(String key) {
-        synchronized (this) {
-            return properties.containsKey(key);
-        }
+        return properties.containsKey(key);
     }
 
     /**
@@ -107,15 +100,11 @@ public class Configuration {
     }
 
     public Object get(String key) {
-        synchronized (this) {
-            return properties.get(key);
-        }
+        return properties.get(key);
     }
 
     public Object put(String key, Object value) {
-        synchronized (this) {
-            return properties.put(key, ConfigUtil.normalizeType(value, null));
-        }
+        return properties.put(key, ConfigUtil.normalizeType(value, null));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/Configuration.java
```diff
@@ -59,29 +62,21 @@ public Configuration(final @Nullable Configuration configuration) {
...
     public Configuration(Map<String, Object> properties) {
-        this(properties, false);
+        this(properties == null ? emptyMap() : properties, false);
     }
 
     /**
      * Create a new configuration.
      *
-     * @param properties the properties to initialize (may be null)
+     * @param properties        the properties to initialize (may be null)
```
maggu2810: Please remove the added spaces between `properties` and `the`

bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/Configuration.java
```diff
@@ -59,29 +62,21 @@ public Configuration(final @Nullable Configuration configuration) {
...
-        if (properties == null) {
-            this.properties = new HashMap<>();
-        } else {
-            if (alreadyNormalized) {
-                this.properties = new HashMap<>(properties);
-            } else {
-                this.properties = ConfigUtil.normalizeTypes(properties);
-            }
-        }
+        this.properties = synchronizedMap(alreadyNormalized ? new HashMap<>(properties) : normalizeTypes(properties));
```
maggu2810: Why has this been changed?
The argument `properties` is marked nullable.
I am pretty sure `new HashMap<>(properties)` will throw an exception if it is called with a null argument.

bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/Configuration.java
```diff
@@ -127,74 +116,67 @@ public Object remove(Object key) {
...
-    public void setProperties(Map<String, Object> properties) {
-        for (Entry<String, Object> entrySet : properties.entrySet()) {
-            this.put(entrySet.getKey(), entrySet.getValue());
-        }
-        for (Iterator<String> it = this.properties.keySet().iterator(); it.hasNext();) {
-            String entry = it.next();
-            if (!properties.containsKey(entry)) {
-                it.remove();
-            }
+    public void clear() {
```
maggu2810: Do we need an public accessible `clear` method? Why?

bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/Configuration.java
mlriess: The spaces are automatically added by the openhab workspace format settings - the param descriptions are aligned. I think this is intentional?

bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/Configuration.java
mlriess: I forgot to remove the nullable annotation - it's no longer possible for the properties parameter to be null.

bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/Configuration.java
mlriess: Since the class is mutable and I needed the method to simplify setProperties(), I saw no need to hide it. But if you think it shouldn't be exposed, I can make it private.

bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/Configuration.java
mlriess: In essence, I simplified this (private) constructor to accept only nonnull properties and made sure that all the other, public constructors call it with nonnull arguments.

bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/Configuration.java
maggu2810: Are you using the most recent IDE setup?
AFAIK this has been an formatter or formatting configuration error and has not been enabled by intention.

bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/Configuration.java
maggu2810: As it is used in `setProperties` only, why not calling `properties.clear()` direct?
I would prefer not to add another method if it is used at one place only AND just calls another method (so itself a one liner).
Public would make sense only if we want to provide the option that external users call `clear()` but there has not been any need till now - so we should not just add it because we can.

bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/Configuration.java
maggu2810: As this constructor is private, I am fine with the changes if the nullable annotation is removed and so `properties` is defined to be non null.

bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/Configuration.java
mlriess: I used the newest Oomph installer. I just disabled the setting and removed the change.

bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/Configuration.java
mlriess: I don't know how many clients do something like setProperties(emptyMap()) which could be improved by having a clear() method, but maybe it's not necessary to add the method in this PR. I removed it.

bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/Configuration.java
mlriess: I removed the nullable annotation.

maggu2810:
@openhab/core-maintainers Shouldn't there be a signed-off check?

@mlriess Sorry for the additional change request, but could you add a Signed-off-by footer to your commits (AFAIK we still require it).

mlriess:
I squashed the changes and added the signed-off-by. Sorry for the trouble, this is my first PR in the project :-)

maggu2810:
Thanks for your changes.

cweitkamp:
> Shouldn't there be a signed-off check?

:+1: Yes. How should we realize that? [Pull Approve](https://pullapprove.com/)?

maggu2810:
I think so.
At least you use it on openhab2-addons already: https://github.com/openhab/openhab2-addons/blob/679e14d/.pullapprove.yml#L3-L5

cweitkamp:
Okay, I can prepare a PR for it. But I do not know if it is sufficient to commit the YAML file or if the plugin has to be activated elsewhere in this repo.

Which other features we want to use? Dow we need [approve_by_comment](`https://www.pullapprove.com/v2/groups/approve_by_comment/)?

maggu2810:
> But I do not know if it is sufficient to commit the YAML file or if the plugin has to be activated elsewhere in this repo.

I will give it a try. As it is used on other openHAB repos, I don't think it will harm here.

Personally I don't need "approve by comment".
I don't like too much magic words as long as I don't see how this will speed up things. Approve by a button press is not really more time consuming that writing that comment.

maggu2810:
> A request to install PullApprove has been submitted on the @openhab account.

bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/Configuration.java
maggu2810: @mlriess Perhaps this will help:
https://github.com/eclipse/smarthome/issues/6424#issuecomment-471728524

maggu2810:
@cweitkamp The PullAprove request has been approved.
Will you create a PR with a `.pullaprove.yml` file?

cweitkamp:
@maggu2810 Done (see #649).

PR #643, opened 2019-03-10
Travis configuration improvements

wborn:
This PR makes it possible again to follow progress of a Travis build while it's running. While building it shows the reactor progress information while making sure the build doesn't timeout by appending zero width characters every 3 minutes.

While I was at it I've made some other configuration improvements as well:

* Uses the new Xenial Travis images which easily allows for running Java 11 builds
* Builds using OpenJDK 8 instead of Oracle JDK 8. Oracle JDK 8 is no longer available in the Xenial images. I've done a lot of builds already with it and it works equally well.
* Adds a new OpenJDK 11 build which is allowed to fail. This makes it easy to see what needs to be done for successfully compiling the code with Java 11. It disables the enforcer for the Java 11 build.
* On successful builds only the reactor summary is printed (instead of the most recent 200 lines)
* Removes JVM options that are no longer supported in Java 8 and which result in warnings:
  `-XX:PermSize=512m -XX:MaxPermSize=1g`
* Removes the unused Paper UI NPM cache from the build

I've also tested a similar configuration with the openhab2-addons build (see this [wborn/openhab2-addons build](https://travis-ci.org/wborn/openhab2-addons/builds/504365795)) where the timeout would occur during feature validation. The build times remain the same with this configuration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,27 +1,54 @@
 sudo: required
-dist: trusty
+dist: xenial
 
 language: java
-jdk: oraclejdk8
+
+jdk:
+  - openjdk8
+  - openjdk11
 
 cache:
   directories:
   - $HOME/.m2
   - $HOME/.p2
-  - extensions/ui/org.eclipse.smarthome.ui.paper/npm_cache
 
-before_install: echo "MAVEN_OPTS='-Xms1g -Xmx2g -XX:PermSize=512m -XX:MaxPermSize=1g'" > ~/.mavenrc
-
-#$ mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
-#$ mvn test -B
-
+before_install:
+  - echo "MAVEN_OPTS='-Xms1g -Xmx2g'" > ~/.mavenrc
 install:
-  - echo 'mvn clean verify -B -V -Dspotbugs.skip=true 1> .build.stdout 2> .build.stderr' > .build.sh
-  - chmod 0755 .build.sh
-script:
-  - travis_wait 60 ./.build.sh
+  - |
+    function extra_maven_opts() {
+        if [ "$TRAVIS_JDK_VERSION" != "openjdk8" ]; then echo "-Denforcer.skip=true"; fi;
+    }
+    function prevent_timeout() {
+        local i=0
+        while [ -e /proc/$1 ]; do
+            # print zero width char every 3 minutes while building
+            if [ "$i" -eq "180" ]; then printf %b '\u200b'; i=0; else i=$((i+1)); fi
+            sleep 1
+        done
+    }
+    function print_reactor_summary() {
+        sed -ne '/\[INFO\] Reactor Summary:/,$ p' "$1" | sed 's/\[INFO\] //'
+    }
+    function mvnp() {
+        set -o pipefail # exit build with error when pipes fail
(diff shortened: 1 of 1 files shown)


wborn:
> How is it working? Are the Xenial VMs more powerful?

It uses a [build matrix](https://docs.travis-ci.com/user/build-matrix/) so Travis typically builds these jobs in parallel. We use an even [bigger build matrix for openhab-docker](https://travis-ci.org/openhab/openhab-docker) where Travis always executes 3 jobs in parallel. The Java 11 job currently always fails before the Java 8 build succeeds so it shouldn't increase the Travis (wall clock) build time.

> Does it make sense to enable a build for 11 as long as we are using Xtext <2.17?

I think it does because it makes it easy to see what other issues there are. E.g. it now fails because of some Mockito issue that needs further investigation. Another reason for adding a Java 11 build is that it shows that we are actively working on supporting newer Java versions which is food for [community topics](https://community.openhab.org/t/openhab-and-future-jdk-versions-when/65572?u=wborn). It also helps to keep the discussion going so we actually get there (like SAT warnings nag people to fix them :wink:).

I wasn't aware that Xtext needs to be updated for supporting Java 11. It wasn't mentioned in https://github.com/eclipse/smarthome/issues/4369. Apparently Xtext 2.17 has only been released last week ([announcement](https://blogs.itemis.com/en/xtext-2-17-the-java-11-release)). Do you think we can easily upgrade Xtext from 2.14 to 2.17? The upgrade should also fix an issue with the xtext-maven-plugin when running a parallel maven build (see https://github.com/eclipse/xtext-maven/pull/37 by @pfink).

maggu2810:
@wborn I don't know at which placed I reported it. Just found this one: https://github.com/openhab/static-code-analysis/issues/325#issuecomment-421301492

cweitkamp:
Do not forget to edit the version constraint in the [pom.xml](https://github.com/openhab/openhab-core/blob/48d873a32ba0901a9b58f04c541f0f858fbfcd51/pom.xml#L586-L588):

```
[WARNING] Rule 0: org.apache.maven.plugins.enforcer.RequireJavaVersion failed with message:
Detected JDK Version: 11.0.2 is not in the allowed range [1.8.0-40,1.9).
```

wborn:
I'd prefer to only do that when we know the code successfully builds using JDK 11. That way it's immediately clear for developers what JDK needs to be installed for a successful build. In the Travis config I've disabled the enforcer with `extra_maven_opts` only for builds that aren't using openjdk8.

maggu2810:
@wborn To skip something, so changing the default behaviour should be only done if really wanted.
Wouldn't it make more sense to skip the enforcer plugin if the used JDK is the one we know to skip it (openjdk11) instead of skip it if it is not "openjdk8"?
Sure, it should not matter, just thinking about it.

maggu2810:
> Do you think we can easily upgrade Xtext from 2.14 to 2.17?

I could never get used to it.
Perhaps someone would like to give it a try (we could track it in a separate issue).

Anything missing in this PR or ready to be merged?

wborn:
It's ready to be merged. :-)

wborn:
> Perhaps someone would like to give it a try (we could track it in a separate issue).

When time permits I can give it a try. There's now https://github.com/openhab/openhab-core/issues/646 for this.

PR #647, opened 2019-03-12
Upgrade Mockito to 2.25.0 for Java 11 compatibility

wborn:
Currently the first exceptions that occur in a Java 11 build are due to Mockito/Byte Buddy being incompatible with Java 11:

```
Mockito cannot mock this class: interface java.lang.Runnable.

Mockito can only mock non-private & non-final classes.
If you're not sure why you're getting this error, please report to the mailing list.


Java               : 11
JVM vendor name    : Oracle Corporation
JVM vendor version : 11.0.1+13
JVM name           : OpenJDK 64-Bit Server VM
JVM version        : 11.0.1+13
JVM info           : mixed mode
OS name            : Linux
OS version         : 4.15.0-1026-gcp


Underlying exception : java.lang.UnsupportedOperationException: Cannot define class using reflection
	at org.eclipse.smarthome.test.java.JavaTestTest.waitForAssertShouldRunAfterLastCall_whenAssertionSucceeds(JavaTestTest.java:38)
Caused by: java.lang.UnsupportedOperationException: Cannot define class using reflection
	at org.eclipse.smarthome.test.java.JavaTestTest.waitForAssertShouldRunAfterLastCall_whenAssertionSucceeds(JavaTestTest.java:38)
Caused by: java.lang.IllegalStateException: Could not find sun.misc.Unsafe
	at org.eclipse.smarthome.test.java.JavaTestTest.waitForAssertShouldRunAfterLastCall_whenAssertionSucceeds(JavaTestTest.java:38)
Caused by: java.lang.NoSuchMethodException: sun.misc.Unsafe.defineClass(java.lang.String, [B, int, int, java.lang.ClassLoader, java.security.ProtectionDomain)
	at org.eclipse.smarthome.test.java.JavaTestTest.waitForAssertShouldRunAfterLastCall_whenAssertionSucceeds(JavaTestTest.java:38)
```

After upgrading Mockito these exceptions are resolved. The stacktraces then show that next Xtext needs to be upgraded to 2.17 (#646) to make the Java 11 build more successful.

We'll also need to update the itests dependencies in openhab2-addons.

PR #648, opened 2019-03-13
bump Gson to 2.8.2

maggu2810:
As long as we depend on the internal Gson packages, we need to use the Gson bundle provided by Eclipse Orbit instead of the official one.

Related to: https://github.com/openhab/openhab-core/pull/641

### DIFF
--- a/itests/org.openhab.core.config.discovery.tests/itest.bndrun
+++ b/itests/org.openhab.core.config.discovery.tests/itest.bndrun
@@ -7,11 +7,15 @@ Fragment-Host: org.openhab.core.config.discovery
 	bnd.identity;id='org.openhab.core.config.discovery.tests',\
 	bnd.identity;id='org.openhab.core.thing.xml'
 
+# If we would like to use a storage at all, we will use the "volatile" storage.
+-runblacklist: \
+    bnd.identity;id='org.openhab.core.storage.json',\
+    bnd.identity;id='org.openhab.core.storage.mapdb'
+
 #
 # done
 #
 -runbundles: \
-	com.google.gson;version='[2.7.0,2.7.1)',\
 	javax.measure.unit-api;version='[1.0.0,1.0.1)',\
 	org.apache.commons.collections;version='[3.2.1,3.2.2)',\
 	org.apache.commons.io;version='[2.2.0,2.2.1)',\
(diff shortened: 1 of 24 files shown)


maggu2810:
Will resolve the conflicts.

itests/org.openhab.core.config.discovery.tests/itest.bndrun
```diff
@@ -45,6 +49,7 @@ Fragment-Host: org.openhab.core.config.discovery
 	tec.uom.se;version='[1.0.8,1.0.9)',\
 	ch.qos.logback.core;version='[1.2.3,1.2.4)',\
 	slf4j.api;version='[1.7.25,1.7.26)',\
-	ch.qos.logback.classic;version='[1.2.3,1.2.4)',\
 	net.bytebuddy.byte-buddy;version='[1.9.7,1.9.8)',\
-	net.bytebuddy.byte-buddy-agent;version='[1.9.7,1.9.8)'
\ No newline at end of file
+	net.bytebuddy.byte-buddy-agent;version='[1.9.7,1.9.8)',\
+	com.google.gson;version='[2.8.2,2.8.3)',\
+	slf4j.simple;version='[1.7.21,1.7.22)'
```
wborn: Any idea why the plugin adds `slf4j.simple` in this PR and `ch.qos.logback.classic` in my PR (https://github.com/openhab/openhab-core/pull/647)? I didn't like that change and even less like the idea of it being switched depending on who generates these files. It's also changed in the other files.

It would be even nicer if we don't have to generate and commit these files and update them for every dependency change.

maggu2810: I assume it has been caused by an unclean development environment.
I forget to remove other commits in front of building and resolving.
slf4j.simple should not be part of the available bundles at all.

The files should be committed if there is any change (sure, we should eliminate changes if not necessary).

maggu2810: There are additional ones that could be cleaned up some time, but I assume now it is a little bit better.

Hilbrand:
@maggu2810 thanks for fixing this. Seeing your changes there were to many unknowns for me.

maggu2810:
In time, it becomes routine. :wink:

PR #649, opened 2019-03-13
Added PullApprove config

cweitkamp:

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -1,0 +1,20 @@
+version: 2
+
+requirements:
+  signed_off_by:
+    required: true
+
+always_pending:
+  title_regex: 'WIP'
+  explanation: 'Work in progress...'
+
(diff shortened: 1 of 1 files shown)


maggu2810:
Hm, it does not seem to work.

cweitkamp:
The YAML is bases on PullApprove v2 which is used in openHAB 2 Add-ons. Maybe we have PullApprove v3 over here now. How can we validate that?

maggu2810:
I cannot press the configure button: https://github.com/openhab/openhab-core/settings/installations
I assume @kaikreuzer is the single one that can have a look at.

kaikreuzer:
Hm, for openhab2-addons we use a Pullapprove webhook, while here for openhab-core the PullApprove Github app is now installed. I cannot see any version mentioned with it, though, but it is likely that this is the newest version.
If we align the repos, it'd probably make sense to use the Github app also for openhab2-addons, but I'll only change the config once we have it working here.

PR #653, opened 2019-03-14
add further imports to automation script module

maggu2810:
Fixes: https://github.com/openhab/openhab-core/issues/610

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/bnd.bnd
+++ b/bundles/org.openhab.core.automation.module.script/bnd.bnd
@@ -1,0 +1,10 @@
+Bundle-SymbolicName: ${project.artifactId}
+Import-Package: \
+	org.eclipse.smarthome.core.transform;resolution:=optional,\
+	org.eclipse.smarthome.model.persistence.extensions;resolution:=optional,\
+	org.eclipse.smarthome.model.script.actions;resolution:=optional,\
+	org.eclipse.smarthome.model.script.engine.action;resolution:=optional,\
+	org.openhab.core.library.types;resolution:=optional,\
+	org.openhab.core.scriptengine.action;resolution:=optional,\
+	org.joda.time;resolution:=optional,\
+	*
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/defaultscope/FQCNAccess.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/defaultscope/FQCNAccess.java
@@ -1,0 +1,96 @@
+/**
+ * Copyright (c) 2010-2019 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.automation.module.script.internal.defaultscope;
+
+/**
+ * @author Markus Rathgeb - Initial contribution
+ */
+public class FQCNAccess {
+
+    /**
+     * Ensure access.
+     *
+     * <p>
+     * This method is used to ensure that a defined set of classes is known by the classloader.
+     * The access of the class will ensure that the package imports are added.
+     */
+    public static void load() {
+        load(org.eclipse.smarthome.config.core.Configuration.class);
+        load(org.eclipse.smarthome.core.binding.BindingInfoProvider.class);
+        load(org.eclipse.smarthome.core.items.Metadata.class);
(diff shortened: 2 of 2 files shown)


maggu2810:
I don't know about a good naming.
The class exists only to ensusre that other classes can be accessed by the classloader.
So in this special case that the packages of that classes are added to the import package manifest entry.

openhab-5iver:
@maggu2810, as you know, I've tested this change, and it appeared to resolve the current issue at hand (thank you again!). But I'm curious... is there any benefit to having the FQCNAccess, or could everything be moved into the bnd.bnd?

Also, IMO allowing scripted automation to dynamically import whatever was required would provide more freedom for people using scripts, and would require less maintenance in the future. Is using 'DynamicImport-Package: *' an option that could be used here, or is this something that we would not want to consider?

maggu2810:
We should move away from `bnd.bnd` at all and use annotations instead.
This has some advantages, e.g. if you rename a package or a class (e.g. an Activator), the annotations that reference a class will be changed, too (if you use the IDE). The bnd.bnd as a simple text only file will not reflect such changes.
See e.g. https://virtual.osgiusers.org/2018/10/pure-annotation-driven-dev

Dynamic package import is AFAIK something you should avoid.
Why do we need to import "all"?
Also if we don't use special script classes at the moment (with refereing to the other comments using an official scripting API), we should only expose what may be used.

openhab-5iver:
> We should move away from bnd.bnd at all and use annotations instead.

A bit over my head ATM, but interesting. I'm not sure how that answers my question though... why add the new class, if everything could just go into the bnd file? Really just a asking out of curiousity.

> Dynamic package import is AFAIK something you should avoid.
Why do we need to import "all"?

My understanding is that this wasn't importing all packages, but allowing them to be optionally imported. I was hoping you could share some insight into why/how it could be detrimental :smile:. I figured you had a reason not to use it, but maybe it was...

> Also if we don't use special script classes at the moment (with refereing to the other comments using an official scripting API), we should only expose what may be used.

... just a matter of restricting access. Bnd (and this PR) is a step towards restricting access, where it was previously completely open. I think this PR reopens much of what was being used before, but I expect there will be some more classes that will be needed to recover lost funcitonality. So, where is the line drawn, as to what will be allowed and what won't be? It sounds to me like an architectual decision needs to be made, and documented, as to what scripted automation will be allowed to access. If it is restricted at all, it will likely turn away the more enthusiastic scripters. A simple example, metadata has been around for a while now, but there is still no way to use it in a DSL rule, other than add/remove (can't read it) through the REST API. However, JSR223 was able to use it immediately, without anything being done to expose it. It will be considered a major breaking change if/when the decision is made to restrict access to classes, and only allow access through a scripting API. I don't see why a scripting API couldn't be built out to provide access to everything that anybody wants to get to (or is would that be restricted too?), but that is a ways out in the future, and IMO, it should be optional.

It looks like there is a lot of potential for using JSR223, or GraalVM, outside of automation... but that's another conversation!

openhab-5iver:
I noticed there are three missing optional imports that I reported here... https://github.com/openhab/openhab-core/issues/610#issuecomment-471151469. The one for org.quartz is used in the [core.triggers](https://github.com/OH-Jython-Scripters/openhab2-jython/blob/master/Core/automation/lib/python/core/triggers.py) module, which builds all of the triggers used in rules, and uses this import to validate cron expressions. The other two are used for finding, creating, and getting OSGi services, and is a crucial piece of functionality made available though the [core.osgi](https://github.com/OH-Jython-Scripters/openhab2-jython/tree/master/Core/automation/lib/python/core/osgi) package.

openhab-5iver:
In case you haven't tried it, without the FQCNAccess and the following bnd file, I had no issues with imports.
```
Bundle-SymbolicName: ${project.artifactId}
DynamicImport-Package: *
```

maggu2810:
I did not check it myself.
As you may have already seen from the other topic, I had not used that scripting automation yet.

If you think dynamic package imports are the better approach, perhaps it makes the most sense.

WRT your comment above:
Sure, if we made all accessible, new features can be used as soon as they are available.
But I am pretty sure users will complain about broken setups as soon as we decide to drop a third party library or we reformat our code.

Why can't we easily replace Joda and drop that libraries?

As long as nobody will block changes because an user script can be broken, I don't care if everything is provided.
But we should add a big fat notice to the documentation that there is no stable scripting API yet.

openhab-5iver:
> If you think dynamic package imports are the better approach, perhaps it makes the most sense.

Functionally, it not only resolves #610, but also provides the same access to classes that we had before the reintegration, and there will be no future changes needed to expose new classes.

> But I am pretty sure users will complain about broken setups as soon as we decide to drop a third party library or we reformat our code.

More so from DSL users than JSR223. Jython scripters can always just import third party libraries directly. For major breaking changes, the helper libraries help to buffer that too, but an API is defintely the way to go. I plan to start working on that once the Jython and Groovy bundles are done.

Shall I create the PR for the DynamicImport change, or are you planning to?

maggu2810:
Please create the PR yourself, there are other things I need to do.

PR #655, opened 2019-03-14
Added i18n feature for dynamic state descriptions

cweitkamp:
- Added i18n feature for dynamic state descriptions via a `BaseDynamicStateDescriptionProvider` class
- Added null annotations to `StateOption`, `StateDescription` and `ChannelStateDescriptionProvider`
- Make use of `StateDescriptionFragmentBuilder`
- Changed null annotation of `CommandOption` label as we always allow `Nullable` labels
- Minor improvements

### DIFF
--- a/bundles/org.openhab.core.test.magic/src/main/resources/ESH-INF/thing/channel-types.xml
+++ b/bundles/org.openhab.core.test.magic/src/main/resources/ESH-INF/thing/channel-types.xml
@@ -28,7 +28,8 @@
 		<item-type>Color</item-type>
 		<label>Color</label>
 		<description>The color channel allows to control the color of a light.
-			It is also possible to dim values and switch the light on and off.
+			It is also possible to dim values and switch the
+			light on and off.
 		</description>
 		<category>ColorLight</category>
 		<tags>
@@ -89,6 +90,11 @@
 		<item-type>String</item-type>
 		<label>Text channel</label>
 		<autoUpdatePolicy>recommend</autoUpdatePolicy>
+	</channel-type>
+
+	<channel-type id="systemcommand">
+		<label>Send system command</label>
+		<description>Sends a system command to Kodi. This allows to shutdown/suspend/hibernate/reboot/quit Kodi.</description>
 	</channel-type>
 
 	<channel-type id="temperature">
--- a/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseDynamicStateDescriptionProvider.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseDynamicStateDescriptionProvider.java
@@ -1,0 +1,152 @@
+/**
+ * Copyright (c) 2010-2019 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.core.thing.binding;
+
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.smarthome.core.thing.Channel;
+import org.eclipse.smarthome.core.thing.ChannelUID;
+import org.eclipse.smarthome.core.thing.i18n.ChannelTypeI18nLocalizationService;
+import org.eclipse.smarthome.core.thing.type.ChannelType;
+import org.eclipse.smarthome.core.thing.type.ChannelTypeUID;
+import org.eclipse.smarthome.core.thing.type.DynamicStateDescriptionProvider;
+import org.eclipse.smarthome.core.types.StateDescription;
+import org.eclipse.smarthome.core.types.StateDescriptionFragmentBuilder;
+import org.eclipse.smarthome.core.types.StateOption;
+import org.osgi.framework.BundleContext;
+import org.osgi.service.component.ComponentContext;
+import org.osgi.service.component.annotations.Activate;
+import org.osgi.service.component.annotations.Deactivate;
+
+/**
+ * The {@link BaseDynamicStateDescriptionProvider} provides a base implementation for the
+ * {@link DynamicStateDescriptionProvider}.
+ * <p>
+ * It provides localized state patterns and dynamic state options while leaving other state description fields as
+ * original. Therefore the inheriting class has to request the reference for the
+ * {@link ChannelTypeI18nLocalizationService} on its own.
+ *
+ * @author Christoph Weitkamp - Initial contribution
+ */
+@NonNullByDefault
+public abstract class BaseDynamicStateDescriptionProvider implements DynamicStateDescriptionProvider {
+
+    private @NonNullByDefault({}) BundleContext bundleContext;
+    protected @NonNullByDefault({}) ChannelTypeI18nLocalizationService channelTypeI18nLocalizationService;
+
+    private final Map<ChannelUID, @Nullable String> channelPatternMap = new ConcurrentHashMap<>();
+    private final Map<ChannelUID, @Nullable List<StateOption>> channelOptionsMap = new ConcurrentHashMap<>();
+
+    /**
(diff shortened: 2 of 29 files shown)


cweitkamp:
What is wrong with the itest here?

```
[ERROR] Error   : Exit code remote process 1: java -cp /home/travis/build/openhab/openhab-core/itests/org.openhab.core.thing.tests/target/test/tmp/testing/itest/cnf/cache/4.2.0/org.openhab.core.bom.runtime-index/org.eclipse.osgi-3.13.100.v20180827-1536.jar:/home/travis/build/openhab/openhab-core/itests/org.openhab.core.thing.tests/target/test/tmp/testing/itest/cnf/cache/4.2.0/bnd-cache/biz.aQute.launcher/biz.aQute.launcher-4.2.0.jar -Dlauncher.properties="/home/travis/build/openhab/openhab-core/itests/org.openhab.core.thing.tests/target/test/tmp/testing/itest/generated/launch3088335997198286107.properties" -ea aQute.launcher.Launcher
```

Is it my fault? I cannot find the reason for it.

cweitkamp:
@maggu2810 Is there a specific reason why the tests for the Magic bundle are commented out in the `itests/pom.xml`?

https://github.com/openhab/openhab-core/blob/71b241ae21a89ae67eb2b935aaf64497c632c928/itests/pom.xml#L48

maggu2810:
No technical one, just not enough time...

IIRC the comment out modules has not yet been migrated -- or at least tested.
IIRC I just applied some common migration steps for all bundles.
After that I checked what is working out of the box and enabled them.
After that I invest some time and fixed some other ones so they are working.

So, feel free to have a look at the magic test (I assume also pure unit tests needs to be moved manually).

cweitkamp:
The unit test for Magic Bundle are working now. But from my point of view there are no integration tests yet. I guess we could move all existing tests.

bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseDynamicStateDescriptionProvider.java
```diff
@@ -0,0 +1,163 @@
...
+        bundleContext = componentContext.getBundleContext();
+    }
+
+    @Deactivate
+    public void deactivate() {
+        channelOptionsMap.clear();
+        bundleContext = null;
+    }
+
+    @Reference
```
maggu2810: Do not depend on DS annotation inheritance (see e.g. https://community.liferay.com/blogs/-/blogs/bnd-instruction-to-avoid -- it has already been discussion on ESH some time).
If a component use this class a parent, he should define the annotated methods itself and call the parent class implementation.

cweitkamp: It somehow makes sense. A subclass has to add a `@Component` annotation too to provide its service. I will update the class accordingly but will remove the setter and getter completely. Similar to the `AbstractDiscoveryService` - I always was wondering why it holds instances of two services without requiring them. That is clear now.

itests/org.openhab.core.thing.tests/src/main/java/org/eclipse/smarthome/core/thing/type/ChannelStateDescriptionProviderOSGiTest.java
```diff
@@ -77,14 +81,20 @@
...
     private ItemChannelLinkRegistry linkRegistry;
 
     @Mock
     private ComponentContext componentContext;
 
+    private Bundle testBundle;
+
     @Before
-    public void setup() {
+    public void setup() throws Exception {
```
davidgraeff: That's very generic ;)

cweitkamp: Unfortunately, I have to

https://github.com/openhab/openhab-core/blob/00edae4539f9ea4188d34afb68424ac8cfb33232/bundles/org.openhab.core.test/src/main/java/org/eclipse/smarthome/test/SyntheticBundleInstaller.java#L79-L92

maggu2810:
My mail UI didn't tell me that the WIP has been removed long time ago.
I will add this PR to my todo list.

maggu2810:
There are merge conflicts.
Would it make sense to split the "enable itests of magic bundle" changes to a separate PR?

cweitkamp:
> Would it make sense to split the "enable itests of magic bundle" changes to a separate PR?

Yes, I thought about that too. Give me a minute.

cweitkamp:
Done: Split-up PR into #756 and this one.

maggu2810:
There is still a merge conflict.

cweitkamp:
Rebased again.

bundles/org.openhab.core.test.magic/src/main/resources/ESH-INF/thing/channel-types.xml
```diff
@@ -91,6 +92,11 @@
 		<autoUpdatePolicy>recommend</autoUpdatePolicy>
 	</channel-type>
 
+	<channel-type id="systemcommand">
+		<label>Send system command</label>
+		<description>Sends a system command to Kodi. This allows to shutdown/suspend/hibernate/reboot/quit Kodi.</description>
```
maggu2810: I really don't care, but I don't think there is some interaction with code. Is there? :wink:

maggu2810:
Magic bundle test failed on Travis CI.

bundles/org.openhab.core.test.magic/src/main/resources/ESH-INF/thing/channel-types.xml
cweitkamp: No, just copy & paste from the binding where the issue has its origin.

cweitkamp:
Alright. Green again.

bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseDynamicStateDescriptionProvider.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    @Override
+    public @Nullable StateDescription getStateDescription(Channel channel, @Nullable StateDescription original,
+            @Nullable Locale locale) {
+        // can be overridden by subclasses
+        ChannelUID channelUID = channel.getUID();
+        String pattern = channelPatternMap.get(channelUID);
+        List<StateOption> options = channelOptionsMap.get(channelUID);
+        if (pattern == null && options == null) {
```
Hilbrand: This suggests both `setStatePattern` and `setStateOptions` need to be set. Given the javadoc of `setStatePattern` that if not set the default pattern is used I would suspect if `pattern` is null here first `original.getPattern()` should be used and checked if that isn't null either?

cweitkamp: I am not sure if I get your point. The `StateDescriptionProvider` only should return an new `StateDescription` instance if at least one values is set. In general the `StateDescriptionService` is responsible for merging it with the original `StateDescription`, so we do not need not take care of that here.

Hilbrand: Ok. I don't know what I was thinking here, maybe I shouldn't write code when tired :smile:. This code seems fine. So I'll have to search further why my other changes didn't work...

cweitkamp: Never mind. I know that :laugh:.

Do you already created a PR (or have a piece of code) to point me to? I can have a look too. And I have some idea related to the component declaration of a `StateDescriptionProvider` which I like to discuss with you. It does not really fit in here.

Hilbrand: I haven't created a pr, but the code is here: https://github.com/Hilbrand/openhab2-addons/commits/spotif_base_dynamic
I tested it locally, but not a lot. It doesn't show the dynamic lists anymore, so I started debugging, but not a lot. Interested in your idea.

cweitkamp: Thanks for the link. See my https://github.com/Hilbrand/openhab2-addons/commit/9c4f586cf6bb64e3cd0c229290a5641e38113521 over there.

PR #656, opened 2019-03-14
Switched PullApprove version to v3

cweitkamp:

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -1,20 +1,14 @@
-version: 2
+version: 3
 
-requirements:
-  signed_off_by:
-    required: true
-
-always_pending:
-  title_regex: 'WIP'
-  explanation: 'Work in progress...'
-
-# Group settings to apply to all groups by default, optionally being overridden later
-group_defaults:
-  reset_on_push:
-    enabled: true
+pullapprove_conditions:
+- condition: "'WIP' not in title"
+  unmet_status: pending
+  explanation: "Work in progress"
 
 groups:
   code-review:
-    required: 1
-    teams:
-    - core-maintainers
+    reviews:
+      required: 1
+    reviewers:
+      teams:
+      - core-maintainers


wborn:
It talks so it must be working. :smile:

cweitkamp:
Yes, that are the good news. But now I know how to continue - in a private repo to not bother you with my tests ...

maggu2810:
@cweitkamp Thanks for your effort.

cweitkamp:
I guess I can stop digging. The feature we are looking for has been removed in version 3 (see https://www.pullapprove.com/docs/migration/).

> - Removed PR "requirements"
> Static PR requirements (such as signed-off-by) are no longer directly a part of PullApprove. Instead we're refocusing on the human element of code review, and you can integrate with other status checks via conditions.

And other features are imho nice to have for openhab-core.

Any ideas?

wborn:
How about using probot DCO instead?

https://probot.github.io/apps/dco/

They have lots of other handy apps too like auto labeling, preventing merging WIP PRs etc.!

maggu2810:
I stumbled across it yesterday while I was looking for alternatives.
I have seen DCO from time to time on other repositories.

But to be honest, I cannot judge as I never used it myself.
If it contains the desired check(s), let's (@kaikreuzer your turn :wink:) give it a try.

cweitkamp:
Looks fine. Would be really helpful if it automatically send a notification to the user if the `sign-off` is missing.

```
Hey @{{ author }}, thanks for the PR! The review will start once the tests and CI checks have passed. You will also need to sign your commits and agree with the DCO (see [Sign your Work](https://www.openhab.org/docs/developer/contributing/contributing.html#sign-your-work)).
```

kaikreuzer:
We already used DCO on openhab-android, so I have now activated it on all repos.

maggu2810:
> We already used DCO on openhab-android, so I have now activated it on all repos.

Thanks!

cweitkamp:
:heavy_check_mark:  DCO is working

I changed the syntax of the config file to handle "WIP" status and add how many approvals are required for a PR to be approved.

kaikreuzer:
> add how many approvals are required for a PR to be approved.

This is actually already configured in github itself for the project, see https://github.com/openhab/openhab-core/settings/branch_protection_rules/3893707

kaikreuzer:
I am actually not too sure if we need PullApprove at all by now... If the WIP is the only "feature" for us, this might also be outdated through the new "WIP PR" feature of Github (I didn't yet have a closer look at how that works).

cweitkamp:
Alright. Meaning we can remove PullApprove completely over here?

kaikreuzer:
Would be fine for me, but if anybody wants to keep/use it, go for it.

PR #658, opened 2019-03-15
Revert "Added PullApprove config"

cweitkamp:
Reverts openhab/openhab-core#649


cweitkamp:
@openhab/core-maintainers If we need to override the DCO you can click on "Details" and then on "Set DCO to pass".

DCO

PR #659, opened 2019-03-15
Split ui bundle into ui.charts and ui.sitemap

### LINKED ISSUE #587: [org.openhab.core.ui] Split
I have read through the code base yesterday and I have issues to understand "org.openhab.core.ui".

* It defines an interface to provide charts via persistence.
  And a default chart provider is also included, which adds http endpoints to access charts.
  Can this be factored out into "org.openhab.core.io.charts" or so? (Makes it more obvious that this bundle adds an IO interface and interacts with IO [persistence])

* ItemUIProvider, ItemUIRegistry is only used by the current sitemaps concept.
  Can those be moved to a sitemap bundle? ("org.openhab.core.sitemap.items" or so).

* Proxies: There is currently a proxy implementation in async and blocking variant (for servlet API 2.4). Can we remove the blocking part? servlet API 2.4 is 10 years old and all supported OSGI containers have newer APIs supported.
  From the javadoc: "The servlet registers as "/proxy" and expects the two parameters "sitemap" and "widgetId"". So it is also tailored towards sitemaps. Mid-term it should work without any sitemap relation, but for now, can we move that to a package name that indicates that this is sitemap related? (e.g. "org.openhab.core.sitemap.mediaproxy" or so).

Reasoning: I'm looking into sitemaps2, and it's easier if everything sitemap related can be identified more easily.

davidgraeff:
The charts bundle can be of use also outside of the sitemap concept.

* The karaf feature name was not changed, it just depends on both ui.charts and ui.sitemap now.
* compat1x: Depends on both ui.charts and ui.sitemap now.
* Internal package names were not changed.

Fixes #587

## Next
Next PR for this series is to remove the blocking proxy implementation and only keep the async variant
and make the (media)proxy sitemap independent. 

## Overall goal

Isolate the sitemap concept.


davidgraeff:
Hm. Why is there still a `features/p2/repo/pom.xml`?

maggu2810:
I assume for the ones that would like to use the target platform inside the IDE to work with PDE bundles.
The repository is part of `openhab-distro/launch/openhab.target`

davidgraeff:
I have adapted the p2 feature file, but I'm not really happy that a contributor has to care about two build systems.

maggu2810:
@davidgraeff I am not happy, too. But IMHO the only "solution" will be to migrate all addons... then we will all happy again (until other things make us sad) :wink: 

What's your suggestion?
Should we drop all PDE stuff at hold until they are migrated?
I can work with it (I can migrate what I need), but I am pretty sure, I am the minority and no one would like to get a openHAB distribution that loose a lot of bundles.

davidgraeff:
If all pde vanishes after openhab2-addons is migrated, that's good enough for me. I guess no one will care about openhab1-addons though. Those addons might really be lost somehow. 

I doing my first bundle migrations today. But I guess it makes sense to use the automatic conversion script at some point (at least for every bundle that is not using .jar file dependencies).

davidgraeff:
Jenkins:

> 13 Build timed out (after 60 minutes). Marking the build as aborted.

kaikreuzer:
I am not sure what benefit this split should bring, see my comment at https://github.com/openhab/openhab-core/issues/587#issuecomment-475916106.

davidgraeff:
> Isolate the sitemap concept.

That is the purpose. The charts part can be re-used by non sitemap concepts.

kaikreuzer:
What does isolation mean? As stated in my comment, the different concerns are already isolated by the split in different packages.

davidgraeff:
Isolation in the context of being able to turn the sitemap concept off without affecting other functionality.

Personally I'm using a setup without the eclipse modeling framework and sitemap does not work then of course, which is fine because it doesn't have a storage backend anyway. But I like to keep charts in.

kaikreuzer:
If you want to be able to use the charts separately, it is ok for me to factor this out into a `ui.chart` (please use singular) bundle, but the rest should simply be left untouched. A `ui.sitemap` bundle imho does not make much sense as its content is not about sitemaps, but only has dependencies (which is logical as sitemaps are the core UI concept of openHAB). And its package names should be reflected in the bundle names, which is not the case anymore, if you rename that bundle.

davidgraeff:
I'll do smaller steps then and factor charts out.

I have opened a question-issue here a few weeks ago about additional ui concepts.
Sitemaps is right now the only concept that is correct, but I had assumed it has been discussed and concluded that sitemaps will be replaced. 

I have prepared and implemented three replacement concepts, and to me it does not make sense to all have them in one "ui" bundle.

They are
* ui.widgets (a backend storage for habpanel, at the moment habPanel is storing one huge json string as service configuration),
* ui.cards (the factored out cards concept from Yannicks habot, slightly modified to not depend on the natural language interpretation parts)
* ui.pages (similar to the current sitemaps concept, but using the storage backend).

And the current sitemap concept just naturally fits in with "ui.sitemap" (and the corresponding model bundles) as a forth concept that need to be kept until consumers have migrated away.

kaikreuzer:
I think any approach to replace sitemaps won't go into 2.x, but in 3.x.
We should create a dev branch where we can do bigger changes.

davidgraeff:
I will propose a different bundle structure

PR #662, opened 2019-03-15
Delete crowdin.yml

davidgraeff:
ui.dashboard is no longer


maggu2810:
You need to fix your name in the Signed-off-by footer.

davidgraeff:
This is a web-commit, so I cannot amend the commit message :/

maggu2810:
Does it care?
I was not aware that the Signed-off-by rule depends on Github.
It is a common future by Git -- also for users without a Github account.

AFAIK the signed off by footer should contain the full name and the mail address.

PR #666, opened 2019-03-15
Update ohdr.version to 1.0.42

wborn:
Should be merged after https://github.com/openhab/openhab-deps-repo/pull/13

### DIFF
--- a/features/p2/poms/pom.xml
+++ b/features/p2/poms/pom.xml
@@ -42,7 +42,7 @@
     <groovy.eclipse.compiler.version>2.9.2-01</groovy.eclipse.compiler.version>
     <groovy.eclipse.batch.version>2.4.3-01</groovy.eclipse.batch.version>
 
-    <ohdr.version>1.0.41</ohdr.version>
+    <ohdr.version>1.0.42</ohdr.version>
     <repo.version>2.5.x</repo.version>
     <karaf.version>4.2.2</karaf.version>
     <jetty.version>9.4.12.v20180830</jetty.version>


wborn:
Please allow me to self merge this small change so we can get a working openhab2-addons build again.

maggu2810:
Which repos are using that POM except openhab2-addons?
Also the OH1 addons?
Any others?

wborn:
It's used by each openHAB add-on repository using Tycho/p2.

These are all the repos that I could find and currently use it:

* [openhab1-addons](https://github.com/openhab/openhab1-addons)
* [openhab2-addons](https://github.com/openhab/openhab2-addons)
* [org.openhab.binding.bacnet](https://github.com/openhab/org.openhab.binding.bacnet) (inherited by parent POM)
* [org.openhab.binding.zigbee](https://github.com/openhab/org.openhab.binding.zigbee)
* [org.openhab.binding.zwave](https://github.com/openhab/org.openhab.binding.zwave)


It was previously also used in the UI add-on repositories before they moved to [openhab-webui](https://github.com/openhab/openhab-webui):

* [org.openhab.ui.habmin](https://github.com/openhab/org.openhab.ui.habmin)
* [org.openhab.ui.habot](https://github.com/openhab/org.openhab.ui.habot)
* [org.openhab.ui.habpanel](https://github.com/openhab/org.openhab.ui.habpanel)

maggu2810:
Would it make sense to migrate the bacnet, zigbee and zwave bundle first?
After that the POM could be moved to openhab2-addons and openhab1-addons could use that parent.

wborn:
It will certainly give us more options to move the addon POM from core to another project when the zigbee and zwave bundles are migrated to a bnd build.

BACNet doesn't seem to be part of any CI build or release pipeline and is an OH1 binding and using a released parent. So it would at some (later) point in time perhaps be a good example for migrating OH1 addons to a bnd build (if we also want to spent time on doing that).

If the addons POM parent moves to openhab2-addons, @pfink will also need to update the release pipeline for that change because currently OH1 addons and OH2 addons are released in parallel:

pipeline

The projects might also expect to find the released pom-tycho artifact in the openhab-core Bintray repo and not openhab2-addons: https://bintray.com/openhab/mvn/openhab-core

I'd personally just leave pom-tycho in openhab-core until we complete migrating  zigbee, zwave and openhab2-addons to bnd.

kaikreuzer:
> After that the POM could be moved to openhab2-addons and openhab1-addons could use that parent.

I don't think this is a good idea. Imho, all "addons" repos should be buildable by themselves and only depend on openhab-core.  Instead of removing the parallel build of openhab1-addons and openhab2-addons, we should rather also include all the others (webui, zigbee, zwave, etc.) to be be built in parallel.

maggu2810:
I just realised that because of the missing parent in the tycho POM the version is not update if you use the versions plugin newVersion goal on the reactor (just FYI)

PR #668, opened 2019-03-16
Increase DateTimeType parsing and formatting precision

### LINKED ISSUE #667: DateTimeType implementation may result in loss of precision on Java 9+
Currently the Java 11 Travis build fails because of unit test failures in the `DateTimeTypeTest`:

```
TEST serializationTestZoned[1](org.eclipse.smarthome.core.library.types.DateTimeTypeTest) <<< ERROR: expected:<2019-03-15T23:19:55.832763Z> but was:<2019-03-15T23:19:55.832Z>
java.lang.AssertionError: expected:<2019-03-15T23:19:55.832763Z> but was:<2019-03-15T23:19:55.832Z>
```

After adding Java 9 and 10 to the build matrix it shows this is an issue since Java 9. It seems that since this version the maximum resolution from the underlying clock is used ([JDK-8164428](https://bugs.openjdk.java.net/browse/JDK-8164428)). So instead of just milliseconds the resolution can even be tenth of microseconds.

The unit tests fail because a higher precision value is used in the constructor. The `toString` and  `toFullString` methods only return values with millisecond precision. So the unit test assertions fail because these low resolution values are used in comparisons with the original high resolution ZonedDateTime value.

Some possibilities for solving this are: 
* change the `toString`/`toFullString` methods so they also return the high precision values
* only adapt the test if we are OK with this loss of resolution, the high precision value is still returned by `getZonedDateTime`
* adapt the test and also reduce the resolution of the `zonedDateTime` field so the values returned by `toString` and  `toFullString` match that of `getZonedDateTime`

wborn:
Since Java 9 ([JDK-8164428](https://bugs.openjdk.java.net/browse/JDK-8164428)) the maximum resolution from the underlying clock is used.
Instead of just milliseconds the resolution can now even be tenth of microseconds.

According to the Type JavaDocs `toFullString()` should return the full string representation of the type to be consumed by `valueOf(String)`.

With the changes in this PR `toFullString()` may return higher resolution date time strings on newer Java versions and `valueOf(value)` is able to parse these.

Code depending on a certain resolution returned by `toFullString()` should use the `format(pattern)` method instead so the resolution will not change.

Fixes #667

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/library/types/DateTimeType.java
+++ b/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/library/types/DateTimeType.java
@@ -33,6 +33,7 @@ import org.eclipse.smarthome.core.types.State;
  * @author Kai Kreuzer - Initial contribution
  * @author Erdoan Hadzhiyusein - Refactored to use ZonedDateTime
  * @author Jan N. Klug - add ability to use time or date only
+ * @author Wouter Born - increase parsing and formatting precision
  */
 @NonNullByDefault
 public class DateTimeType implements PrimitiveType, State, Command {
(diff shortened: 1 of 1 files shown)


wborn:
I've also made the (4) `DateTimeFormatter`s static because they are thread safe. This should result in some nice performance and memory benefits because this way they are no longer created for each instance. :-)

PR #161, opened 2017-06-29
updated repo version

kaikreuzer:
Signed-off-by: Kai Kreuzer <kai@openhab.org>

PR #679, opened 2019-03-26
bump SAT version to 0.6.1 for Tycho builds

J-N-K:
This should be merged immediately before https://github.com/openhab/openhab2-addons/pull/5026.

### DIFF
--- a/features/p2/poms/pom.xml
+++ b/features/p2/poms/pom.xml
@@ -48,7 +48,7 @@
     <jetty.version>9.4.12.v20180830</jetty.version>
     <jna.version>4.5.2</jna.version>
     <jackson.version>1.9.13</jackson.version>
-    <sat.version>0.5.0</sat.version>
+    <sat.version>0.6.1</sat.version>
     <nrjavaserial.version>3.15.0.OH2</nrjavaserial.version>
     <maven.compiler.plugin.version>3.3</maven.compiler.plugin.version>
     <build.helper.maven.plugin.version>1.9.1</build.helper.maven.plugin.version>


kaikreuzer:
The build is miserably failing, see https://ci.openhab.org/job/openHAB-Core/166/console.
Could this be related to SAT? Not really, but I do not see any other recent code changes...

@wborn Any idea what might have broken the build?

J-N-K:
That looks really strange. I see the same "Errors" in my local build, however (as stated in the text) only as warnings. Since the change should only affect tycho builds and this is bnd, it should not be due to this change.

wborn:
I think last time we saw these errors it was fixed by restarting Jenkins. It also prints:

```
17:15:22 Java HotSpot(TM) 64-Bit Server VM warning: Insufficient space for shared memory file:
17:15:22    8837
17:15:22 Try using the -Djava.io.tmpdir= option to select an alternate temp location.
```

Did you already restart Jenkins after I asked for that a few days ago @kaikreuzer ? (https://github.com/openhab/openhab2-addons/pull/5256#issuecomment-476780086). Are there also plans for improving Jenkins' speed/stability?

kaikreuzer:
I restarted Jenkins once a day...
And yes, I have just restarted it now after reconfiguring the log configuration (which was the culprit for filling up the disk space). So let's touch wood that it runs much more stable from now on!

kaikreuzer:
I just ran https://ci.openhab.org/job/openHAB-Core/167/console with `-DskipChecks=true`, though.
You think there is a chance it succeeds even with SAT enabled? I can try it...

wborn:
Yes it should also work with SAT enabled. The last succesful build `#161` also had SAT enabled, see the [console output](https://ci.openhab.org/job/openHAB-Core/161/consoleFull).

J-N-K:
#168 looks good.

kaikreuzer:
Excellent!

PR #685, opened 2019-03-31
Add required bundles to itests for successful Java 11 build

wborn:
Adds the following bundles required for building on Java 11 by default to the run requirements:

* org.apache.servicemix.specs.activation-api-1.1
* org.apache.servicemix.specs.annotation-api-1.3
* org.apache.servicemix.specs.jaxb-api-2.2

---

With this PR the last remaining Java 11 compatibility issues are fixed.
After it's merged we can also successfully build OHC with Java 11. :-)


maggu2810:
Would it be possible to create three separate PRs for the changes?
The changes address separate stuff and should be merged not in one commit.

The mDNS test fix can be merged easily.

The byte buffer one, too (and it is not related to the test fix).

The other one needs a little bit more detailed look (at least for me).
So we could keep that commit part of this PR.

Can you create two new PRs and split the respective changes out?

wborn:
I've now created #686 and #687 for these @maggu2810.
There's now also #688 for enforcing Java 11 compatibility if this PR is merged.

wborn:
It looks like the servicemix and geronimo bundles indeed already provide these packages. So I'll have a look if we can use those instead.

maggu2810:
It would be simpler if we could use one execution environment for the tests and a demo app.
That's what the bnd instruction `-runee` is used for.

If I change the execution environment for Bnd and run the resolver, it will add the specific specs without the need to add them to the requirements ourself.

```diff
diff --git a/app/app.bndrun b/app/app.bndrun
index 3747267..be32b14 100644
--- a/app/app.bndrun
+++ b/app/app.bndrun
@@ -63,7 +63,7 @@ feature.openhab-model-runtime-all: \
        bnd.identity;id='org.openhab.ui.paper'
 
 -runfw: org.eclipse.osgi
--runee: JavaSE-1.8
+-runee: JavaSE-11
 
 -runproperties: \
        osgi.console=,\
@@ -200,4 +200,7 @@ feature.openhab-model-runtime-all: \
        org.eclipse.xtext.xbase.lib;version='[2.17.0,2.17.1)',\
        org.objectweb.asm;version='[7.0.0,7.0.1)',\
        org.objectweb.asm.commons;version='[7.0.0,7.0.1)',\
-       org.objectweb.asm.tree;version='[7.0.0,7.0.1)'
+       org.objectweb.asm.tree;version='[7.0.0,7.0.1)',\
+       jaxb-api;version='[2.2.11,2.2.12)',\
+       org.apache.servicemix.specs.activation-api-1.1;version='[2.9.0,2.9.1)',\
+       org.apache.servicemix.specs.annotation-api-1.3;version='[1.3.0,1.3.1)'
```

We should think about the different approaches...

wborn:
I've reworked the PR so it now uses the existing bundles. The demo app can also be fixed in another PR.

The runee approach also seems promising! Perhaps there's already a bnd variable for the Java version? If not, we can also set it using Maven profiles.

maggu2810:
Another option would be to add the requirement for the API bundles to the commons itest bndrun file, so they are added for ever execution environment to every itest (as long as we want to support both).

Using something like that:

```
--- a/itests/itest-common.bndrun
+++ b/itests/itest-common.bndrun
@@ -21,3 +21,7 @@ Import-Package: org.osgi.framework.*;version="[1.8,2)",*
 # The integration test itself does not export anything.
 Export-Package: 
 -exportcontents: 
+
+-runrequires.ee: bnd.identity;id='jaxb-api',\
+    bnd.identity;id='org.apache.servicemix.specs.activation-api-1.1',\
+    bnd.identity;id='org.apache.servicemix.specs.annotation-api-1.3'
```

Will add the respective bundles to every runproperites on new resolution.

If we ever decided to require Java 11 we need to change the itest-common.bundrun file only by removing the runrequires.ee and change runee (and rerun the resolver).

wborn:
The PR has been updated to add them by default to all itests.
That will probably also save everyone a lot of debugging time.

PR #686, opened 2019-03-31
Avoid ByteBuffer incompatibility when compiling with JDK9+

wborn:
Split off from #685

### DIFF
--- a/itests/org.openhab.core.io.net.tests/src/main/java/org/eclipse/smarthome/io/net/http/HttpRequestBuilderTest.java
+++ b/itests/org.openhab.core.io.net.tests/src/main/java/org/eclipse/smarthome/io/net/http/HttpRequestBuilderTest.java
@@ -15,6 +15,7 @@ package org.eclipse.smarthome.io.net.http;
 import static org.junit.Assert.assertEquals;
 import static org.mockito.Mockito.verify;
 
+import java.nio.Buffer;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.time.Duration;
@@ -96,7 +97,8 @@ public class HttpRequestBuilderTest extends BaseHttpUtilTest {
     private String getContentFromProvider(ContentProvider value) {
         ByteBuffer element = value.iterator().next();
         byte[] data = new byte[element.limit()];
-        ((ByteBuffer) element.duplicate().clear()).get(data);
+        // Explicit cast for compatibility with covariant return type on JDK 9's ByteBuffer
+        ((ByteBuffer) ((Buffer) element.duplicate()).clear()).get(data);
         return new String(data, StandardCharsets.UTF_8);
     }
 


itests/org.openhab.core.io.net.tests/src/main/java/org/eclipse/smarthome/io/net/http/HttpRequestBuilderTest.java
```diff
@@ -96,7 +97,7 @@ public void testPostWithContent() throws Exception {
     private String getContentFromProvider(ContentProvider value) {
         ByteBuffer element = value.iterator().next();
         byte[] data = new byte[element.limit()];
-        ((ByteBuffer) element.duplicate().clear()).get(data);
+        ((ByteBuffer) ((Buffer) element.duplicate()).clear()).get(data);
```
maggu2810: This is fully correct, but I am not sure if this is really readable.

Why not using e.g.

```Java
final ByteBuffer duplicate = element.duplicate(); // work on a copy to not modify the source
duplicate.clear(); // Prepare buffer for reading
duplicate.get(data); // read buffer into data
```

So, we do not need any casting at all.

(I do not test the code just write it here on Github)

wborn: The solution in this PR is based on https://github.com/apache/felix/pull/114. Are you sure your approach will still work after reading the explanation for why it occurs?

maggu2810: I don't see why it shouldn't.

The methods (e.g. `clear`) return `this` and it differs between the version by using the parent class `Buffer` or the specific class `ByteBuffer`.
If you are using the return value by writing all in one line it does matter.

If we do not use the return value (it is `this`, so no need for), why should we (developer, compiler, runtime, ...) care about at all?

But my understanding could be wrong.

maggu2810: Ah, okay, while reading it again, it could matter...

maggu2810: Should we add a comment like the linked one (`// Explicit cast for compatibility with covariant return type on JDK 9's ByteBuffer`) in the code and copy the explanation in the first linked comment to the commit message?
Then a git blame will identify why this casts are necessary, without requiring to open Github). IMO the commit message should contain the information why the change is necessary.

WDYT?

wborn: It's tricky stuff and I'm happy to get it working. If I remember correctly it occurs if you compile this code with Java 8 and then use it with Java 11 or vice versa. :disappointed:

wborn: Yes you're right about adding some comments for this sorcery! :wink:

PR #688, opened 2019-03-31
Require successful Travis CI Java 11 build

wborn:
* Update enforcer to allow compilation with Java 8, 9, 10 and 11
* Update Travis CI configuration to require the Java 11 build to succeed

---

Depends on #685

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,9 +16,6 @@ before_install:
   - echo "MAVEN_OPTS='-Xms1g -Xmx2g'" > ~/.mavenrc
 install:
   - |
-    function extra_maven_opts() {
-        if [ "$TRAVIS_JDK_VERSION" != "openjdk8" ]; then echo "-Denforcer.skip=true"; fi;
-    }
     function prevent_timeout() {
         local i=0
         while [ -e /proc/$1 ]; do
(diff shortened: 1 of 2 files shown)


kaikreuzer:
What's the goal in making it compile with Java 11?
We should indeed make sure that running on Java 11 works nicely, but building is imho not really urgent.
I'm mainly asking as I remember that Xtext generates different code when Java 11 is used at build time, which makes the code incompatible with Java 8 at runtime.

maggu2810:
> What's the goal in making it compile with Java 11?

An example would be to support machines that set Java 11 globally for the system.
Also we don't need Java 8 on that machines at all.

kaikreuzer:
You mean machines that build+run openHAB at the same time?

maggu2810:
... or something different that is using openHAB Core.

Or machines that install 11 for building and other machines that install 11 for execution.

Perhaps some guy don't want to install a pre 11 environment on its machine.

Other question: Is there something wrong to support Java 11 for compilation?

I am fine with that changes regardless if it is used by me or not.
IMHO we should support compilation and runtime support for Java 11.

wborn:
For me it's mostly the same as supporting compilation on another OS or architecture. 

Another benefit is that the unit test results tell something about how well your code runs on Java 11. 

It's also a matter of time until Java 8 is no longer available for a certain OS or distro.

Personally I won't use Java 11 on my desktop if I can't also compile most of my code with it.

kaikreuzer:
My point is that it must be clear that when using Java 11 for compilation, the artifacts are not compatible with Java 8 anymore - so as long as we want to provide compatibility with Java 8 at runtime, no official built must use Java 11. Fine for me to make sure that it nonetheless compiles with Java 11 as long as this is clear.

wborn:
It's clear to me and I've had my own hands-on experience with such incompatibilities. The official releases should be fine if we only use Java 8 on Jenkins. We could even add a Maven profile that detects Jenkins and then enforces Java 8.

PR #691, opened 2019-04-01
Always send correct 'item' member in SSE events.

### LINKED ISSUE #690: SSE events for visibility changes sent with wrong data
When an item changes visibility (or color) of another item, the SSE event sent for this change contains the widget ID of the widget whose visibility was changed, but the item of the widget which triggered the change. This is inconsistent and will lead to wrong state updates on client side, since the client has no chance of knowing what widget the sent item belongs to.

Sample data for this can be found at https://github.com/openhab/openhab-android/pull/1174#issuecomment-478476101 - the problem can be reproduced using the 'Outside temperature' page of the demo server.

Issue is here: https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java#L239-L242 ... `event.item` should be either the item of `w` or be omitted entirely if `skipWidget` is true; same applies to `event.state`. Not sure what variant is preferable?

maniac103:
If a change to item A causes visibility changes to item B, we previously
sent an SSE event for B including the new visibility for B, but included
item and state of A. This leads to client confusion, as it'll update its
internal state for the widget belonging to B with the item state of A.
As the visibility change doesn't include updated item and state anyway,
just omit those from the SSE event.

Closes #690

### DIFF
--- a/bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
+++ b/bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
@@ -216,7 +216,8 @@ public class PageChangeListener implements StateChangeListener {
                 events.addAll(constructSitemapEvents(item, state, itemUIRegistry.getChildren((Frame) w)));
             }
 
-            boolean skipWidget = (w.getItem() == null) || !w.getItem().equals(item.getName());
+            boolean itemBelongsToWidget = w.getItem() != null && w.getItem().equals(item.getName());
+            boolean skipWidget = !itemBelongsToWidget;
             // We skip the chart widgets having a refresh argument
             if (!skipWidget && w instanceof Chart) {
                 Chart chartWidget = (Chart) w;
@@ -231,16 +232,19 @@ public class PageChangeListener implements StateChangeListener {
                 event.valuecolor = itemUIRegistry.getValueColor(w);
                 event.widgetId = itemUIRegistry.getWidgetId(w);
                 event.visibility = itemUIRegistry.getVisiblity(w);
-                // event.item contains data from the item including its state (in event.item.state)
+                // event.item contains the (potentially changed) data of the item belonging to
+                // the widget including its state (in event.item.state)
+                final Item itemToBeSent = itemBelongsToWidget ? item : getItemForWidget(w);
                 String widgetTypeName = w.eClass().getInstanceTypeName()
                         .substring(w.eClass().getInstanceTypeName().lastIndexOf(".") + 1);
                 boolean drillDown = "mapview".equalsIgnoreCase(widgetTypeName);
                 Predicate<Item> itemFilter = (i -> i.getType().equals(CoreItemFactory.LOCATION));
-                event.item = EnrichedItemDTOMapper.map(item, drillDown, itemFilter, null, null);
+                event.item = EnrichedItemDTOMapper.map(itemToBeSent, drillDown, itemFilter, null, null);
 
-                // event.state is an adjustment of the item state to the widget type.
-                event.state = itemUIRegistry.convertState(w, item, state).toFullString();
-                // In case this state is identical to the item state, its value is set to null.
+                // event.state is an adjustment of the item state to the widget type, but is only sent if both
+                // - the new state belongs to the widget in question and
+                // - the state actually changed
+                event.state = itemBelongsToWidget ? itemUIRegistry.convertState(w, item, state).toFullString() : null;
                 if (event.state != null && event.state.equals(event.item.state)) {
                     event.state = null;
                 }
@@ -249,6 +253,18 @@ public class PageChangeListener implements StateChangeListener {
             }
         }
         return events;
+    }
+
+    private Item getItemForWidget(Widget w) {
+        final String itemName = w.getItem();
+        if (itemName != null) {
+            try {
+                return itemUIRegistry.getItem(itemName);
+            } catch (ItemNotFoundException e) {
+                // fall through to returning null
+            }
+        }
+        return null;
     }
 
     private boolean definesVisibilityOrColor(Widget w, String name) {
--- a/bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
+++ b/bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
@@ -244,7 +244,8 @@ public class PageChangeListener implements StateChangeListener {
                 // event.state is an adjustment of the item state to the widget type, but is only sent if both
                 // - the new state belongs to the widget in question and
                 // - the state actually changed
-                event.state = itemBelongsToWidget ? itemUIRegistry.convertState(w, item, state).toFullString() : null;
+                final State stateToBeSent = itemBelongsToWidget ? state : itemToBeSent.getState();
+                event.state = itemUIRegistry.convertState(w, itemToBeSent, stateToBeSent).toFullString();
                 if (event.state != null && event.state.equals(event.item.state)) {
                     event.state = null;
                 }
--- a/bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
+++ b/bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
@@ -241,11 +241,10 @@ public class PageChangeListener implements StateChangeListener {
                 Predicate<Item> itemFilter = (i -> i.getType().equals(CoreItemFactory.LOCATION));
                 event.item = EnrichedItemDTOMapper.map(itemToBeSent, drillDown, itemFilter, null, null);
 
-                // event.state is an adjustment of the item state to the widget type, but is only sent if both
-                // - the new state belongs to the widget in question and
-                // - the state actually changed
+                // event.state is an adjustment of the item state to the widget type.
                 final State stateToBeSent = itemBelongsToWidget ? state : itemToBeSent.getState();
                 event.state = itemUIRegistry.convertState(w, itemToBeSent, stateToBeSent).toFullString();
+                // In case this state is identical to the item state, its value is set to null.
                 if (event.state != null && event.state.equals(event.item.state)) {
                     event.state = null;
                 }


maniac103:
@kaikreuzer @lolodomo You added and later modified the code in question, could you please review?

lolodomo:
I will review this PR this week.

lolodomo:
I understand the problem but your fix leads to the generation of events without event.item and event.state for widgets having their visibility impacted. I am not sure that it will be correctly handled by UI like Basic UI.
First, we have to check Basic UI code to see how it will behave in this case.

maniac103:
Please go for it :-)
AFAICT it previously should have suffered from the same issue (state of the items whose visibility was changed being replaced by the state of the item causing the visibility change), but I think the code block at https://github.com/openhab/openhab-webui/blob/master/bundles/org.openhab.ui.basic/web-src/smarthome.js#L1968-L1972 needs to be extended to handle the case of both missing item and state and using the existing state in that case.

lolodomo:
After checking Basiuc UI code, I can confirm that it requires the state(s) when receiving an event to make the widget visible.

lolodomo:
Yes, you clearly identified the location in Basic UI code.

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
```diff
@@ -231,18 +231,20 @@ public void changeStateTo(Item item, State state) {
                 event.valuecolor = itemUIRegistry.getValueColor(w);
                 event.widgetId = itemUIRegistry.getWidgetId(w);
                 event.visibility = itemUIRegistry.getVisiblity(w);
-                // event.item contains data from the item including its state (in event.item.state)
-                String widgetTypeName = w.eClass().getInstanceTypeName()
-                        .substring(w.eClass().getInstanceTypeName().lastIndexOf(".") + 1);
-                boolean drillDown = "mapview".equalsIgnoreCase(widgetTypeName);
-                Predicate<Item> itemFilter = (i -> i.getType().equals(CoreItemFactory.LOCATION));
-                event.item = EnrichedItemDTOMapper.map(item, drillDown, itemFilter, null, null);
+                if (!skipWidget) {
```
maniac103: Maybe instead of this we should do a ternary?
```
Item itemToBeSent = skipWidget ? item : w.getItem();
event.item = EnrichedItemDTOMapper.map(itemToBeSent, drillDown, itemFilter, null, null);
```

...
```

**Comment by lolodomo** [comment]
I am not sure that the existing and correct state is saved by Basic UI when widgets are invisible.

File: bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java [hunk 2]
```diff
@@ -231,18 +231,20 @@ public void changeStateTo(Item item, State state) {
...
+                    // event.item contains data from the item including its state (in event.item.state)
+                    String widgetTypeName = w.eClass().getInstanceTypeName()
+                            .substring(w.eClass().getInstanceTypeName().lastIndexOf(".") + 1);
+                    boolean drillDown = "mapview".equalsIgnoreCase(widgetTypeName);
+                    Predicate<Item> itemFilter = (i -> i.getType().equals(CoreItemFactory.LOCATION));
+                    event.item = EnrichedItemDTOMapper.map(item, drillDown, itemFilter, null, null);
 
-                // event.state is an adjustment of the item state to the widget type.
-                event.state = itemUIRegistry.convertState(w, item, state).toFullString();
```
maniac103: ... but still keep this null if `skipWidget` is true. That way, (AFAICT) the correct item will be sent, but no explicit state will be sent with the event.

lolodomo:
Yes, that would be better.
We should avoid setting event.state to null. Maybe we can recompute it using w.getItem() too. I am not sure if the item state in part of the item too.

maniac103:
> We should avoid setting event.state to null.

Why? The current state of the code clearly does *not* avoid setting it to null, and clearly doing so didn't cause any issues?

maniac103:
Logic updated as outlined in previous comments.
If anyone now could tell me how to run the openhab-core demo app in a way that makes accesses to '/rest/sitemaps/ not throw 404, that'd be great.

maniac103:
Actually got around to testing this patch (the IDE setup is a pretty frustrating procedure at the moment :-/ ), seems to work fine.

maggu2810:
I cannot judge as my knowledge of that part is rudimentary.
If no one veto, I will apply it.

@lolodomo It seems you know that interface much better, so will you add your review approval / change requests, too? The only one that add its agreement for the moment is @davidgraeff

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
```diff
@@ -231,16 +232,19 @@ public void changeStateTo(Item item, State state) {
...
-                event.item = EnrichedItemDTOMapper.map(item, drillDown, itemFilter, null, null);
+                event.item = EnrichedItemDTOMapper.map(itemToBeSent, drillDown, itemFilter, null, null);
 
-                // event.state is an adjustment of the item state to the widget type.
-                event.state = itemUIRegistry.convertState(w, item, state).toFullString();
-                // In case this state is identical to the item state, its value is set to null.
+                // event.state is an adjustment of the item state to the widget type, but is only sent if both
+                // - the new state belongs to the widget in question and
+                // - the state actually changed
+                event.state = itemBelongsToWidget ? itemUIRegistry.convertState(w, item, state).toFullString() : null;
```
lolodomo: Why not rather
`event.state = itemUIRegistry.convertState(w, itemToBeSent, itemBelongsToWidget ? state : event.item.state).toFullString() `

maniac103: Because the lines right below this line (248-250) would set it to null again anyway.

lolodomo: No that is wrong, that is all the interest of this field, having in certain cases a widget state that is different from the item state. Why not recomputing it. In case it is equivalent to the item state, yes it will then be set to null.

maniac103: https://github.com/openhab/openhab-core/pull/691/commits/df2284a89b656f33ebb17255167b3e5125155f88 should do what you mean, I guess?

lolodomo: Yes that is better like that.
Except the updated comment that is wrong

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
```diff
@@ -231,15 +232,18 @@ public void changeStateTo(Item item, State state) {
                 event.valuecolor = itemUIRegistry.getValueColor(w);
                 event.widgetId = itemUIRegistry.getWidgetId(w);
                 event.visibility = itemUIRegistry.getVisiblity(w);
-                // event.item contains data from the item including its state (in event.item.state)
+                // event.item contains the (potentially changed) data of the item belonging to
+                // the widget including its state (in event.item.state)
+                final Item itemToBeSent = itemBelongsToWidget ? item : getItemForWidget(w);
```
cweitkamp: @maniac103 May I ask you to take a second look at these changes? The newly introduced method `getItemForWidget()` may return a `null` value. But later in the code we pass the variable `itemToBeSent` towards two methods line [242](https://github.com/openhab/openhab-core/pull/691/files#diff-78e437b3d2798d795dfa76249a91c583R242) and line [246](https://github.com/openhab/openhab-core/pull/691/files#diff-78e437b3d2798d795dfa76249a91c583R246)) which require a `@NonNull` value.

This could e.g. lead to a NPE like this:

```
2019-06-25 15:19:17.216 [WARN ] [pse.smarthome.core.items.GenericItem] - failed notifying listener 'org.eclipse.smarthome.io.rest.sitemap.internal.PageChangeListener@71c0a5b7' about state update of item testString: null
	at org.eclipse.smarthome.core.items.dto.ItemDTOMapper.fillProperties(ItemDTOMapper.java:152) ~[137:org.openhab.core:2.5.0.201906250301]
	at org.eclipse.smarthome.core.items.dto.ItemDTOMapper.map(ItemDTOMapper.java:138) ~[137:org.openhab.core:2.5.0.201906250301]
	at org.eclipse.smarthome.io.rest.core.item.EnrichedItemDTOMapper.map(EnrichedItemDTOMapper.java:56) ~[?:?]
	at org.eclipse.smarthome.io.rest.sitemap.internal.PageChangeListener.constructSitemapEvents(PageChangeListener.java:242) ~[?:?]
	at org.eclipse.smarthome.io.rest.sitemap.internal.PageChangeListener.constructSitemapEvents(PageChangeListener.java:216) ~[?:?]
	at org.eclipse.smarthome.io.rest.sitemap.internal.PageChangeListener.constructAndSendEvents(PageChangeListener.java:174) ~[?:?]
	at org.eclipse.smarthome.io.rest.sitemap.internal.PageChangeListener.stateChanged(PageChangeListener.java:188) ~[?:?]
	at org.eclipse.smarthome.core.items.GenericItem$1.run(GenericItem.java:262) [137:org.openhab.core:2.5.0.201906250301]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [?:?]
	at java.lang.Thread.run(Thread.java:748) [?:?]
```

Reproducible by instructions in #897.

maniac103: @cweitkamp In that case we probably need to omit `event.item` and `event.state`, which is what I wanted to do to begin with, but concerns were raised on whether events without item are handled properly everywhere. I can't think of any other solution if the widget in question actually does not have any item though.
@lolodomo Thoughts?

lolodomo:
À null check Is probably missing in your new code. 
I will try to understand where. 
Unfortunately, as everybody, I have no more dev environment for testing or debugging OH core and UI.

maniac103:
> A null check Is probably missing in your new code.

Not sure. The basic question is: If we, as you wanted, always send event.item, from where do we take said item if the widget in question doesn't have any item? If omitting the item is OK, having a null check for itemToBeSent around the assignment of event.item/event.state should work.

(And yeah, the lack of a simple dev environment is really a problem)

lolodomo:
Of course,  if there is no item attached to the widget,  there is no need to build an item. I just have to understand why we encounter this case now and not before.
If there is no item attached to the widget,  I think there is just no need to send a SSE event and we should not run the involved code. This was the job of boolean skipWidget I think.

maniac103:
Let's recap the scenario this PR aimed to fix:
- We have two widgets A and B
- visibility of A is controlled by B
- when B changes, two SSE events are sent out: one for the change of B, one for the visibility change of A
- this PR is about the SSE event for the visibility change of A

Previous behavior was that in the visibility change event of A contained the new visibility flag for A, but item and state of B (!)
Intended new behavior is that that visibility change event of A contains new visibility, item and state of A

As previously B had both item and state, we didn't encounter this case before. We also can't drop the event, as clients need to be informed of the visibility change.

maggu2810:
> Unfortunately, as everybody, I have no more dev environment for testing or debugging OH core and UI.

Haven't had a look at the UI yet but tested the official IDE setup with its quirks here:
https://github.com/openhab/openhab-distro/issues/927

maniac103:
Proposed fix (null check for `itemToBeSent`): https://github.com/openhab/openhab-core/pull/909

PR #692, opened 2019-04-01
Removed apache commons-httpclient, added commons-codec.

Hilbrand:
While [migrating the neato binding](https://github.com/openhab/openhab2-addons/pull/5290) I encountered a problem with commons-codec versions.
Turns out the commons-codec version included is from the rather old commons-httpclient. Since all http goes through jetty I guess this library could be removed. SoI've removed the dependency on commons-httpclient and put back an up-to-date version of commons-code.
The only use in core was in the compat1x HttpUtil. This has be replaced by calls to the smarthome HttpUtil methods.

Disclaimer: I don't have a full overview of the impact of this change. This repository did build locally with this change.

### DIFF
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -71,9 +71,9 @@
       <scope>compile</scope>
     </dependency>
     <dependency>
-      <groupId>org.apache.servicemix.bundles</groupId>
-      <artifactId>org.apache.servicemix.bundles.commons-httpclient</artifactId>
-      <version>3.1_7</version>
+      <groupId>commons-codec</groupId>
+      <artifactId>commons-codec</artifactId>
+      <version>1.12</version>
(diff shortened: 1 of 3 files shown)


maggu2810:
* The compat1x tests are failing. Have you checked if you need to resolve the platform again?
* If the httpclient is not used anymore, it should be removed from the Karaf features, too.
* The karaf features currently use commons-codec 1.6. So, why do you bump the compile (min. version) to 1.12?

Hilbrand:
I'm going to close this pr. I found out core had already a class `HexUtils` that could replace the use of library in the binding. So I'm not sure if it would make sense to add the commons-codec library at all. Although removing commons-httpclient may be a good idea. But I think that is better handled when the compat1x part is refactored.

PR #693, opened 2019-04-02
[core] Improved thing / bridge performance

cweitkamp:
- Changed property `channels` from `List<Channel>` to `Map<ChannelUID, Channel>` for faster accessing channels of a thing
- Changed property `things` from `List<Thing>` to `Map<ThingUID, Thing>` for faster accessing things of a bridge
- Added new methods `getChannel(ChannelUID)` and `getThing(ThingUID)`
- Fixed SAT findings (e.g. null annotation on `BridgeImpl`)

I removed the method `getChannelsMutable()` from `ThingImpl` and replaced it by a `addChannel(Channel)` method. Imho the best solution will be to omit this new method too, but unfortunately we need it in the helper class `ThingHelper#addChannelsToThing()`. Any ideas how to get rid of that? It is used only in two places - first one can be refactored completely to make use of the `ThingBuilder`:

https://github.com/openhab/openhab-core/blob/48d873a32ba0901a9b58f04c541f0f858fbfcd51/bundles/org.openhab.core.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/thing/ThingResource.java#L221

and

https://github.com/openhab/openhab-core/blob/48d873a32ba0901a9b58f04c541f0f858fbfcd51/bundles/org.openhab.core.model.thing/src/org/eclipse/smarthome/model/thing/internal/GenericThingProvider.xtend#L327

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/Bridge.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/Bridge.java
@@ -21,10 +21,20 @@ import org.eclipse.smarthome.core.thing.binding.BridgeHandler;
 /**
  * A {@link Bridge} is a {@link Thing} that connects other {@link Thing}s.
  *
- * @author Dennis Nobel - Initial contribution and API
+ * @author Dennis Nobel - Initial contribution
+ * @author Christoph Weitkamp - Added method `getThing(ThingUID)`
  */
 @NonNullByDefault
 public interface Bridge extends Thing {
+
+    /**
+     * Gets the thing for the given UID or null if no thing with the UID exists.
+     *
+     * @param thingUID thing UID
+     * @return the thing for the given UID or null if no thing with the UID exists
+     */
+    @Nullable
+    Thing getThing(ThingUID thingUID);
 
     /**
      * Returns the children of the bridge.
--- a/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/Thing.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/Thing.java
@@ -29,11 +29,12 @@ import org.eclipse.smarthome.core.thing.binding.ThingHandler;
  * A {@link Thing} might be connected through a {@link Bridge}.
  * <p>
  *
- * @author Dennis Nobel - Initial contribution and API
+ * @author Dennis Nobel - Initial contribution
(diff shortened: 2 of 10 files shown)


openhab-5iver:
Has it been standard practice to remove deprecated methods in a minor release? I looked and was surpised to not find any policy for this in the Developer Guide. The methods you removed have been deprecated for over two years now, but IMO removing them should only be done in preparation for a major release.

cweitkamp:
Afaik, Yes. I have done it before (see https://github.com/eclipse/smarthome/pull/5244 or https://github.com/eclipse/smarthome/pull/5819). Two years sound like a reasonable duration for finally getting removed. From my point of view the switch from ESH to OHC is a second argument to do it. @openhab/core-maintainers wdyt?

maggu2810:
I personally don't care as it has been deprecated a really long time and I am fine with dropping all the deprecated methods.

The bumps of major, minor, ... version bumps and API breakage would only fit if openHAB would use semantic version.
But I was not aware that this is really strict.

But, also if we are not using semantic version I assume this methods can stay as long as we switch to openHAB 3.
Perhaps my reading has been wrong but the `ThingBuilder` and the `BridgeBuilder` are public API.

davidgraeff:
As long as OH extensions are not affected (mostly OH2-addons)...

I really like to have the speed improvements.

maggu2810:
That's the reason I prefer not to mix different / unrelated changes in one PR.
Except the removal of the deprecated method, it looks good to me.

davidgraeff:
@maggu2810 But @cweitkamp is right. Both removed deprecated methods are in "internal" packages, aren't they?

maggu2810:
> Perhaps my reading has been wrong but the `ThingBuilder` and the `BridgeBuilder` are public API.

As commented above my concerns are about the removal of that methods:

* https://github.com/openhab/openhab-core/pull/693/files#diff-44fdf6ebca4daa065ec6088297fb2859L50
* https://github.com/openhab/openhab-core/pull/693/files#diff-256bf6fd1cd5c729241c6bf6e423b994L55

But I assume I miss some points here.

davidgraeff:
Ah I see. It is about `@Deprecated	 public static ThingBuilder create(ThingUID thingUID)`.

True that specific method might not be removed in this PR to be on the safe site.

cweitkamp:
I am prepared for that. Will cherry-pick the removal into separate PRs and split this enhancement up.

maggu2810:
Thank you

cweitkamp:
Done.

PR #694, opened 2019-04-02
Add commons-net to bom

wborn:
The resolver failed to resolve the commons-net bundle used in the Max Cube Binding itests (https://github.com/openhab/openhab2-addons/pull/5376).

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -375,6 +375,12 @@
       <version>2.6</version>
       <scope>compile</scope>
     </dependency>
+    <dependency>
+      <groupId>commons-net</groupId>
+      <artifactId>commons-net</artifactId>
+      <version>3.3</version>
+      <scope>compile</scope>
+    </dependency>
 
     <!-- Gson -->
     <dependency>


maggu2810:
I have a question about that change:
Which dependencies of addons will be added to the BOMs of core if they are not used by core itself?

wborn:
I didn't find any other BOM managing such dependencies in openhab2-addons yet. So added them to this core BOM because it already managed many other Apache commons libraries. 

It does make sense to use it only for managing core dependencies in core and add-ons dependencies in their own project. Do you think these should be added to the [runtime-index POM](https://github.com/openhab/openhab2-addons/blob/master/bom/runtime-index/pom.xml) in openhab2-addons or should we create another index for this?

Is it also possible to add some configuration to the itest.bndrun files for this? If a dependency is only used by one add-on it might not make sense to add it to an index. Add-ons might also use different versions of a dependency. They may also depend on very specific bundles (e.g. a Chromecast or Logitech Harmony client) in an integration test. I'd rather not add these to such indexes.

maggu2810:
As it has been an Apache commons library and its just about "runtime", I approved it.

But I don't think the core BOMs should be "polluted" by dependencies that are not needed by core itself.

The runtime BOM should be similar to what is provided by Karaf and the Karaf Core features.

PR #695, opened 2019-04-03
Added units kvarh and kvar.

Hilbrand:
Also simplified gravity unit.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/library/unit/SmartHomeUnits.java
+++ b/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/library/unit/SmartHomeUnits.java
@@ -14,6 +14,7 @@ package org.eclipse.smarthome.core.library.unit;
 
 import java.math.BigInteger;
 
+import javax.measure.Quantity;
 import javax.measure.Unit;
 import javax.measure.quantity.Acceleration;
 import javax.measure.quantity.AmountOfSubstance;
@@ -63,6 +64,7 @@ import tec.uom.se.function.MultiplyConverter;
 import tec.uom.se.function.PiMultiplierConverter;
 import tec.uom.se.function.RationalConverter;
 import tec.uom.se.unit.AlternateUnit;
+import tec.uom.se.unit.BaseUnit;
 import tec.uom.se.unit.ProductUnit;
 import tec.uom.se.unit.TransformedUnit;
 import tec.uom.se.unit.Units;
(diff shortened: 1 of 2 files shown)


Hilbrand:
The build fails in things tests on spotbugs timeout:
```
21:37:20 [INFO] --- sat-plugin:0.6.1:spotbugs (sat-all) @ org.openhab.core.thing.tests ---
21:37:20 [INFO] Adding dependency to findbugs:0.6.1
21:37:20 [INFO] Adding dependency to bug-pattern:1.2.4
21:37:20 [INFO] Adding dependency to spotbugs:3.1.7
21:37:21 [INFO] Fork Value is false
21:37:22      [java] JVM args ignored when same JVM is used.
21:50:55 Timeout: sub-process interrupted
```
```
21:52:18 [ERROR] Failed to execute goal org.openhab.tools.sat:sat-plugin:0.6.1:spotbugs (sat-all) on project org.openhab.core.thing.tests: Unable to execute mojo: Execution null of goal com.github.spotbugs:spotbugs-maven-plugin:3.1.6:spotbugs failed: Timeout: killed the sub-process -> [Help 1]
```
A 'normal' build runs in seconds:
```
18:02:02 [INFO] Adding dependency to spotbugs:3.1.7
18:02:02 [INFO] Fork Value is false
18:02:04      [java] JVM args ignored when same JVM is used.
18:02:28 [INFO] Done SpotBugs Analysis....
```

hakan42:
Out of curiosity, how will those units be used? The description on wikiperia is not really informative for people like me who are not electrical engineers :)

Hilbrand:
Here is an explanation using the :beer:  analogy :smile:  
https://electrical-engineering-portal.com/beer-mug-and-power-factor

These units are used for a new smart meter thing. Where the meter gives those values.

Hilbrand:
@wborn or @cweitkamp can one of you have a look. It is now blocking my dsmr pr and thus blocking migrating the dsmr binding to bnd.

bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/library/unit/SmartHomeUnits.java
```diff
@@ -82,8 +84,8 @@
 
     // Alphabetical ordered by Unit.
     public static final Unit<Acceleration> METRE_PER_SQUARE_SECOND = addUnit(Units.METRE_PER_SQUARE_SECOND);
-    public static final Unit<Acceleration> STANDARD_GRAVITY = new TransformedUnit<>("gₙ",
```
cweitkamp: Don't we loose the symbol "g" when applying this.

J-N-K: I think so. Rest looks good IMHO.

PR #697, opened 2019-04-05
Enable Semantics Integration Tests

wborn:
They seem to work just fine.

### DIFF
--- a/itests/org.openhab.core.semantics.tests/itest.bndrun
+++ b/itests/org.openhab.core.semantics.tests/itest.bndrun
@@ -9,12 +9,7 @@ Fragment-Host: org.openhab.core.semantics
 # done
 #
 -runbundles: \
-	ch.qos.logback.classic;version='[1.2.0,1.2.1)',\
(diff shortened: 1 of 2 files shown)


cweitkamp:
I'd say it's okay. But where is Travis?

wborn:
Maybe we need to remove the Travis CI "Service" from the [Apps and Integrations](https://github.com/openhab/openhab-core/settings/installations) settings @kaikreuzer ? We don't have it in openhab2-addons anymore. So it may also be caused by https://github.com/openhab/openhab2-addons/issues/4935.

wborn:
Another difference is that it is missing a [Webhook](https://github.com/openhab/openhab-core/settings/hooks) to https://notify.travis-ci.org

maggu2810:
@kaikreuzer 
It seems all `PR-openHAB-Core` fail because of OOM.
Anything we can do?
Can you activate Travis CI again, so there is at least one CI active?

maggu2810:
Build it myself and it succeeded.

PR #702, opened 2019-04-08
runtime: add support for Paho MQTT

maggu2810:


davidgraeff:
Hm is that a long term solution? 
I mean to add all kind of dependencies to the pom although those parts are optional only (in this case, as long as you do not install the mqtt transport, you do not need paho).

maggu2810:
What do you mean by "long term solution"?
The purpose of runtime BOM of `openhab-core` is to provide all things that are necessary to use an openHAB Core bundle at runtime.
Similar as as openHAB Core's compile BOM defines all non openHAB Core artifacts we depend on compile time.

davidgraeff:
I'm just asking. So if we add more transports to this repo, the bom would need to be modified.

maggu2810:
Just want to mention that it is not (only) "the long term solution" but the **intention** of that POM.

Issue #699, opened 2019-04-07
MapDB: Not prepared to work with Metadata?


davidgraeff:
I'm working with item metadata in https://github.com/openhab/openhab2-addons/pull/5243 (Hue Emulation Service) and I'm using https://github.com/maggu2810/openhab-demo as the runtime for development. That runtime is configured to use mapdb instead of jsondb which exposes the following deserialize problem:

```
o.e.s.s.mapdb.internal.MapDbStorage:171 - Couldn't deserialize value 'org.eclipse.smarthome.core.items.Metadata@@@{"key":{"segments":["HUEEMU","DemoDimmer"]},"value":"5","configuration":{}}'. Root cause is: Unable to invoke no-args constructor for class org.eclipse.smarthome.core.items.Metadata. Registering an InstanceCreator with Gson for this type may fix this problem.
```

Afaik mapdb is still a supported db. Can someone confirm the above issue?

Quick demo code:

```java
@Component(immediate = true)
@NonNullByDefault
public class MyClass {
    @Reference
    protected @NonNullByDefault({}) MetadataRegistry metadataRegistry;

    @Activate
    public void activate(Map<String, Object> properties) {
         MetadataKey key = new MetadataKey("some_namespace", "existing_item_name");
         Metadata metadata = metadataRegistry.get(key);

         if (metadata != null) {
              metadataRegistry.add(new Metadata(key, "a_value", null));
         }
    }
}
```

Update: I have changed to jsondb and I'm experiencing the same problem. My next guess is that due to the update to a newer (and stricter!) gson library, gson can no longer (de)serialize classes with a no-args constructor. And MetaData does not have that!

davidgraeff:
@openhab/core-maintainers Comments are appreciated

maggu2810:
Just add them similar to e.g. #505, #509

davidgraeff:
I see. Didn't know about similar cases. A simple no-arg c'tor and non-final fields should be enough then to make it work again, I guess.

maggu2810:
This kind of no-arg constructor should not result in the removal of `final`s.
The `final` keyword makes member variables thread safe and this safety should not be removed.

PR #703, opened 2019-04-08
Update Metadata.java

davidgraeff:
Potential fix for #699 

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/items/Metadata.java
+++ b/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/items/Metadata.java
@@ -25,15 +25,17 @@ import org.eclipse.smarthome.core.common.registry.Identifiable;
  * This is a data class for storing meta-data for a given item and namespace.
  * It is the entity used for within the {@link MetadataRegistry}.
  *
- * @author Kai Kreuzer - Initial contribution and API
- *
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/items/Metadata.java
```diff
@@ -25,15 +25,17 @@
...
 public final class Metadata implements Identifiable<MetadataKey> {
 
-    private final MetadataKey key;
-    private final String value;
-    private final Map<String, Object> configuration;
+    private MetadataKey key;
+    private String value;
+    private Map<String, Object> configuration;
+
+    Metadata() {
```
maggu2810: Don't remove `final`.
Add the comment that is (should be) used on all the other such constructors, too.

davidgraeff: I get you. Immutable classes are preferable. Is gson able to modify final fields?

maggu2810: Gson uses reflection to set final fields (via `.setAccessible(true)`),

maggu2810:
Has this been closed by intention?
A PR tracks a branch, so no need to close and create a new PR as long as the branch keeps the same.

PR #704, opened 2019-04-08
Update Metadata.java

### LINKED ISSUE #699: MapDB: Not prepared to work with Metadata?

davidgraeff:
Fixes #699 

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/items/Metadata.java
+++ b/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/items/Metadata.java
@@ -35,7 +35,15 @@ public final class Metadata implements Identifiable<MetadataKey> {
     private final String value;
     private final Map<String, Object> configuration;
 
-    Metadata() { // For (de)serialization
+    /**
+     * Package protected default constructor to allow reflective instantiation.
+     *
+     * !!! DO NOT REMOVE - Gson needs it !!!
+     */
+    Metadata() {
+        key = new MetadataKey("", "");
+        value = "";
+        configuration = Collections.emptyMap();
     }
 
     public Metadata(MetadataKey key, String value, @Nullable Map<String, Object> configuration) {
--- a/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/items/Metadata.java
+++ b/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/items/Metadata.java
@@ -25,7 +25,7 @@ import org.eclipse.smarthome.core.common.registry.Identifiable;
  * This is a data class for storing meta-data for a given item and namespace.
  * It is the entity used for within the {@link MetadataRegistry}.
  *
- * @author Kai Kreuzer - Initial contribution and API
+ * @author Kai Kreuzer - Initial contribution
  *
  */
 @NonNullByDefault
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/items/Metadata.java
```diff
@@ -35,6 +35,9 @@
     private final String value;
     private final Map<String, Object> configuration;
 
+    Metadata() { // For (de)serialization
```
maggu2810: Can't we use the same comment for all such constructors?

bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/items/Metadata.java
```diff
@@ -35,6 +35,9 @@
...
+    }
```
maggu2810: You need to assign a "dummy" value to the final fields to pass the null checks (all members are non null).
You will see an example in one of the referenced PR.

davidgraeff: True. Regarding this other branch, I'm using the github editor to get this done as another core checkout and build would poison my local maven cache.

davidgraeff: fixed. But you are rather verbose here in terms of LOC for this actually very obvious fact. I think I liked my one liner more.

maggu2810: It has not been chosen by me (IIRC) but present in some ESH file for years.
My requirement is that it is the same for all files.

So, if we would like to search all files and replace it, I am still happy. :wink:

maggu2810: > as another core checkout and build would poison my local maven cache.

Just if you are interested in: `mvn -Dmaven.repo.local=/some/other/directory ...`

davidgraeff: I'm a gradle user and so to say "forced" to use maven for this project. I turns out it is not 10 times slower as I originally thought (that was tycho) and if you turn it silent, it also doesn't cause so much noise anymore. But I'm still not familiar with all those command line switches. Thanks for the hint.

maggu2810:
I tested the build locally.

PR #705, opened 2019-04-08
Update AbstractManagedProvider.java

davidgraeff:
The storage service must be set by the parent class at some point,
but the parent class is not able to check later on if a storage service is set.

Make the field protected to access the storage service for null checks and other usage.
The alternative is that the parent class stores the same field another time.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/common/registry/AbstractManagedProvider.java
+++ b/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/common/registry/AbstractManagedProvider.java
@@ -44,7 +44,7 @@ import org.slf4j.LoggerFactory;
 public abstract class AbstractManagedProvider<E extends Identifiable<K>, K, PE> extends AbstractProvider<E>
         implements ManagedProvider<E, K> {
 
-    private StorageService storageService;
+    protected StorageService storageService;
     private volatile Storage<PE> storage;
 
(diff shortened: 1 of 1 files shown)


maggu2810:
Isn't `AbstractManagedProvider` the parent class?

> The alternative is that the parent class stores the same field another time.

Am I correct that you mean "subclass" instead of "parent class"?

But I don't understand why this is necessary?
Can you give me some more details?
The storage service itself is just used to create a storage and its concrete value should not matter for the subclass.
All managed providers I remember are using a mandatory reference to the storage service. At which place you need to check its value? Do you consider to use an optional reference or non active components?

davidgraeff:
> Am I correct that you mean "subclass" instead of "parent class"

I tried hard but the prefix "sub" did not come to my mind when writing that post.

> All managed providers I remember are using a mandatory reference to the storage service.

Actually the recommended usage is optional, dynamic and I'm doing this in my services. So I need a way to check if a storageservice has been set. At the moment I'm using a flag in the subclass, but that should really not be necessary. 

A boolean function `isStorageServiceSet()` or so would suffice, but why not exposing the service itself.

davidgraeff:
> As we do not need to support R4.2 anymore

I'm with you. I'm using DS 1.3 in the hue emulation service. I do not see a reason to clutter code with old syntax to support a 10+ years old standard.

> Could you point me to that recommendation?

I got that recommendation when working on mqtt but it would take long to look for the comment.
The code is not prepared for optional anyways, so mandatory + greedy seems to be a logical choice.

PR #709, opened 2019-04-10
Fix BaseThingHandlerFactory.getServices()

davidgraeff:
If we only call activate() on the `ThingHandlerService` interface, and not on the service instance itself,
we skip super class activation. This breaks thing handler services that extend AbstractDiscoveryService.

@cweitkamp DYT this solves the discovery issue?

See: https://github.com/openhab/openhab2-addons/pull/5013#issuecomment-481470448
And: https://github.com/openhab/openhab2-addons/pull/5351#issuecomment-479118436
And find numerous other cases (Deconz, Yamaha etc)

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseThingHandlerFactory.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseThingHandlerFactory.java
@@ -188,7 +188,7 @@ public abstract class BaseThingHandlerFactory implements ThingHandlerFactory {
                         } else {
                             serviceRegs.add(serviceReg);
                         }
-                        ths.activate();
+                        serviceInstance.activate();
                     }
                 }
             } catch (InstantiationException | IllegalAccessException e) {
--- a/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseThingHandlerFactory.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseThingHandlerFactory.java
@@ -188,7 +188,8 @@ public abstract class BaseThingHandlerFactory implements ThingHandlerFactory {
                         } else {
                             serviceRegs.add(serviceReg);
                         }
-                        serviceInstance.activate();
+                        Method activateMethod = c.getMethod("activate");
+                        activateMethod.invoke(serviceInstance);
                     }
                 }
             } catch (InstantiationException | IllegalAccessException e) {


maggu2810:
@davidgraeff Can you give me some more details about your concerns?

For every class returned by `getServices()` we create an instance.
If the new object is not an instance of the `ThingHandlerService`, we drop them again (see comment below).
If the new object is an instance of `ThingHandlerService` we store the reference `serviceInstance` points to also in `ths`.
Your change does not call `ths.activate()` anymore but uses reflection to call `serviceInstance.activate()`.
The same reference is used in both cases, so I don't see the difference.
A interface only defines the method a class needs to implement. Methods are called for the referenced object in both cases.


But another improvement:
We should test if `ThingHandlerService.class` is assignable from `Class c` and already break the further processing if this is not true.
So, no need to create an instance first and then check after that (we could skip the instance creation at all).

davidgraeff:
Yes I agree with your change. The method is absolutely strange implemented.

Fact is: Classes that extend AbstractDiscoveryService are broken with the current design. We have no integration test yet for this case as that is a new pattern. In my reviews I always tell people to use this new getServices() pattern and they always come back telling me that it doesn't work.

My own implementation in the deconz service also doesn't work. If you read through the linked comments, @cweitkamp came to the conclusion that the wrong activate method is called (ie not the one from AbstractDiscoveryService)

maggu2810:
If a class implements the `ThingHandlerService` interface it must contain a `public` method that's signature is `void activate()`.
If a parent class of the implementation already contains that method, there is nothing to do.
If you implement an `activate` method in your class, it is up to the implementation to decide if you want to call the parents method or not.
From my understanding this is normal object orientated programming.

But I am pretty sure that there is another bug.
Can you give me a simple modification for e.g. the magic bundle or point me to a bnd based bundle that shows me this buggy behaviour (so I can reproduce it in my env).

maggu2810:
The classes that are returned by `getServices()` should not contain an `@Component` annotation at all.
They should not be activated by SCR or some similar mechanism.

They should return implementations of classes that would like to provide services using the thing handler lifecycle.

The `BaseThingHandlerFactory` will create an instance of that class, register the instance for all implemented interfaces and calls the activate method of this implementation.
There is no "activate" method.
It is a custom made service registration.
No DS involved, no activate annotation etc.

davidgraeff:
Exactly. But that didn't work and the instance creation tricked us in believing that the framework accepted the provided "service", even if a few moments later the framework decided to throw it away again.

Regarding my fix attempt. This works fundamentally different in Java then in C++. In C++ ThingHandlerService.activate would have been called with the cast like in the current code, but at the same time we don't have interfaces or even interfaces with default implementations.

maggu2810:
In C++ it depends if you are using `virtual` methods or not (at least the time I used the language). :wink: 

I can try to dig into your problems if you point me to a Bnd based binding (You've aroused my interest) .

cweitkamp:
@maggu2810 My old playground: https://github.com/openhab/openhab2-addons/compare/master...cweitkamp:feature-openweathermap-discovery. Unfortunately with a not yet bnd-migrated version of the binding. It should be no problem to rebase it or move the example to the Magic Bundle if you like. I guess the valuable information are the comments. If not - do not hesitate to ask. 

A related PR https://github.com/openhab/openhab2-addons/pull/4921 for the deCONZ binding.

maggu2810:
I have to admit, that I was not really aware of that concept before I read this PR.
After reading the current implementation I understand how it works and this is a little demonstration that works as expected (WRT to the current `BaseThingHandlerFactory.registerServices(Thing thing, ThingHandler thingHandler)` implementation.

https://github.com/maggu2810/openhab-core/commit/91ecadf635c1e9199bd4f25760b465b6d1811220

* The direct declared interfaces are used for service registration except `ThingHandlerService` itself (`Class[] interfaces = c.getInterfaces();`).
* There have to be exist a public accessible `activate()` method (no args). If there is an implementation in the class or a subclass, that one is used. As last resort the default implementation of the interface is used (but that's only the last resort, a method of a base class is always preferred -- as defined for `default` on interfaces).

cweitkamp:
Thank you very much for the explanation. In other words - What is missing in my example - and probably the deCONZ binding too - is the implementation of the `DiscoveryService` interface on the class we want to provide as a `ThingHandlerService` and an `activate()` method, even if the `AbstractDiscoveryService` already implements it.

cweitkamp:
I can confirm, that the above suggested implementation is working with one exception. The service got activated but it did not resolve dependencies on other OSGI bundles. Any ideas how to solve that?

```
    @Reference
    protected void setLocationProvider(LocationProvider locationProvider) {
        this.locationProvider = locationProvider;
    }
```

```
2019-04-11 08:32:15.168 [DEBUG] [org.openhab.binding.openweathermap  ] - BundleEvent INSTALLED - org.openhab.binding.openweathermap
2019-04-11 08:32:15.454 [DEBUG] [org.openhab.binding.openweathermap  ] - BundleEvent RESOLVED - org.openhab.binding.openweathermap
2019-04-11 08:32:15.485 [DEBUG] [org.openhab.binding.openweathermap  ] - BundleEvent STARTING - org.openhab.binding.openweathermap
2019-04-11 08:32:15.490 [DEBUG] [org.openhab.binding.openweathermap  ] - BundleEvent STARTED - org.openhab.binding.openweathermap
2019-04-11 08:32:15.575 [DEBUG] [org.openhab.binding.openweathermap  ] - ServiceEvent REGISTERED - {org.eclipse.smarthome.core.thing.binding.ThingHandlerFactory}={service.id=466, service.bundleid=263, service.scope=bundle, component.name=org.openhab.binding.openweathermap.internal.factory.OpenWeatherMapHandlerFactory, component.id=295} - org.openhab.binding.openweathermap
2019-04-11 08:33:24.490 [DEBUG] [overy.OpenWeatherMapDiscoveryService] - Called 'setThingHandler' in OpenWeatherMap DiscoveryService.
2019-04-11 08:33:24.505 [DEBUG] [org.openhab.binding.openweathermap  ] - ServiceEvent REGISTERED - {org.eclipse.smarthome.config.discovery.DiscoveryService}={service.id=467, service.bundleid=263, service.scope=singleton} - org.openhab.binding.openweathermap
2019-04-11 08:33:24.512 [DEBUG] [overy.OpenWeatherMapDiscoveryService] - Activate OpenWeatherMap DiscoveryService.
2019-04-11 08:33:24.515 [DEBUG] [overy.OpenWeatherMapDiscoveryService] - Start OpenWeatherMap Location background discovery job at interval 60 s.
2019-04-11 08:33:24.542 [DEBUG] [nal.handler.OpenWeatherMapAPIHandler] - Initialize OpenWeatherMap API handler 'openweathermap:weather-api:db416c85'.
2019-04-11 08:33:24.558 [DEBUG] [ap.internal.utils.ByteArrayFileCache] - Using cache folder '/var/lib/openhab2/cache/org.openhab.binding.openweathermap'
2019-04-11 08:33:24.567 [DEBUG] [nal.handler.OpenWeatherMapAPIHandler] - Start refresh job at interval 60 min.
2019-04-11 08:34:24.527 [WARN ] [mmon.WrappedScheduledExecutorService] - Scheduled runnable ended with an exception: 
	at org.openhab.binding.openweathermap.internal.discovery.OpenWeatherMapDiscoveryService.scanForNewLocation(OpenWeatherMapDiscoveryService.java:112) ~[?:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) ~[?:?]
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308) ~[?:?]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180) ~[?:?]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [?:?]
	at java.lang.Thread.run(Thread.java:748) [?:?]
2019-04-11 08:40:00.253 [DEBUG] [nal.handler.OpenWeatherMapAPIHandler] - Dispose OpenWeatherMap API handler 'openweathermap:weather-api:db416c85'.
2019-04-11 08:40:00.259 [DEBUG] [nal.handler.OpenWeatherMapAPIHandler] - Stop refresh job.
2019-04-11 08:40:00.275 [DEBUG] [org.openhab.binding.openweathermap  ] - ServiceEvent UNREGISTERING - {org.eclipse.smarthome.config.discovery.DiscoveryService}={service.id=467, service.bundleid=263, service.scope=singleton} - org.openhab.binding.openweathermap
2019-04-11 08:40:00.284 [DEBUG] [overy.OpenWeatherMapDiscoveryService] - Deactivate OpenWeatherMap DiscoveryService.
2019-04-11 08:40:00.287 [DEBUG] [overy.OpenWeatherMapDiscoveryService] - Stop OpenWeatherMap Location background discovery job.
```

maggu2810:
@cweitkamp Don't mix programmatic service registration and OSGi components.
The `@Reference` annotation is used for components. This is not related to the concept OH provides.

If you want to use the components mechanism AND a wiring to the service, your programmatic registered service should be used as a marker only.
Your component will reference that marker interfaces (multiples) and handle the logic.

PR #710, opened 2019-04-10
ThingHandler.getServices() apidoc

davidgraeff:
Add a usage example to getServices()

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandler.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandler.java
@@ -188,9 +188,13 @@ public interface ThingHandler {
     void handleRemoval();
 
     /**
-     * This method provides a list of classes which should be registered as services by the framework
+     * This method provides a list of classes which should be registered as services by the framework.
+     * <p>
+     * Use this method for example to return your thing specific discoveries (extending {@link AbstractDiscoveryService}).
+     * For more details have a look at the {@link ThingHandlerService} documentation.
      *
-     * @return - list of classes that will be registered as OSGi services
+     * @return - list of classes that will be registered as OSGi services,
+     *           for example "Collections.singletonList(MyThingDiscovery.class);"
      */
     default Collection<Class<? extends ThingHandlerService>> getServices() {
         return Collections.emptyList();


bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandler.java
```diff
@@ -188,9 +188,13 @@
...
 
     /**
-     * This method provides a list of classes which should be registered as services by the framework
+     * This method provides a list of classes which should be registered as services by the framework.
+     * <p>
+     * Use this method for example to return your thing specific discoveries (extending {@link AbstractDiscoveryService}).
+     * For more details have a look at the {@link ThingHandlerService} documentation.
      *
-     * @return - list of classes that will be registered as OSGi services
+     * @return - list of classes that will be registered as OSGi services,
```
maggu2810: What's the intention of the dash between `@return` and `list`.
Which documentation style is it?

bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandler.java
```diff
@@ -188,9 +188,13 @@
     void handleRemoval();
 
     /**
-     * This method provides a list of classes which should be registered as services by the framework
+     * This method provides a list of classes which should be registered as services by the framework.
+     * <p>
+     * Use this method for example to return your thing specific discoveries (extending {@link AbstractDiscoveryService}).
```
maggu2810: Hm, using the `AbstractDiscoveryService` this way is IMHO more complicated as you need to provide other services. Didn't we?

bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandler.java
davidgraeff: That you have to ask the original contributor I guess

maggu2810: Sorry, I did not realize that it has been present before.
But if we identify a bad style (at least we did not use this style if it exists), we can IMHO change it if we already touch the same line.
But absolutely not a must.

bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandler.java
davidgraeff: But `getServices()` was to my knowledge purely created to support this use-case. If the implementation is un-usable (I mean it really is), then we should replace it instead.

I would say we use `@Component(scope=ServiceScope.PROTOTYPE)` on those services and let the framework create them and not creating our own object instance first via newInstance.

maggu2810: I was not aware that we ever tried to create a new component logic.
For me it seems it has been created for simpler services like the ones it has been contributed: https://github.com/eclipse/smarthome/pull/6398

davidgraeff: But don't you occasionally need component dependencies in your thingActions as well?
I'm using this for discovery services tied to things because I was told to by other core maintainers and I liked that simpler approach. But if that is not the intention of the API, the documentation should be adopted to say so.

davidgraeff:
`getServices()` was not meant to be used for thing handler bound discovery services. The underlying implementation would need to be changed. Closing for now.

PR #711, opened 2019-04-10
ThingHandlerService apidoc

davidgraeff:
Add a usage example for getServices()

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandlerService.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandlerService.java
@@ -13,11 +13,25 @@
 package org.eclipse.smarthome.core.thing.binding;
 
 /**
+ * The {@link ThingHandler#getServices()} function allows you to return a list of classes,
+ * implementing this interface, that will be started as OSGi services. 
+ * <p>
+ * Use this for example to start thing specific discovery services (extending {@link AbstractDiscoveryService}).
+ * Example:
+ * <pre>
+ * @NonNullByDefault
+ * public class MyThingDiscovery extends AbstractDiscoveryService implements ThingHandlerService {
+ *     private @Nullable ThingHandler thingHandler; // A reference to your thing handler
  *
- * Interface for a service that provides access to a {@link ThingHandler}
+ *     public void activate() {
+ *         super(Collections.emptyMap()); // start AbstractDiscoveryService
+ *     }
+ *     public void setThingHandler(ThingHandler handler) { this.thingHandler = handler; }
+ *     public ThingHandler getThingHandler() { return this.thingHandler; }
+ * }
+ * </pre>
  *
  * @author Stefan Triller - initial contribution
- *
  */
 public interface ThingHandlerService {
 


bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandlerService.java
```diff
@@ -13,11 +13,25 @@
...
 
 /**
+ * The {@link ThingHandler#getServices()} function allows you to return a list of classes,
+ * implementing this interface, that will be started as OSGi services. 
+ * <p>
+ * Use this for example to start thing specific discovery services (extending {@link AbstractDiscoveryService}).
+ * Example:
+ * <pre>
+ * @NonNullByDefault
+ * public class MyThingDiscovery extends AbstractDiscoveryService implements ThingHandlerService {
```
cweitkamp: We have to slightly change this as the important thing is to implement the interface which has to be registered as service:

```suggestion
 * public class MyThingDiscovery extends AbstractDiscoveryService implements DiscoveryService, ThingHandlerService {
```

bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandlerService.java
```diff
@@ -13,11 +13,25 @@
...
+ *     private @Nullable ThingHandler thingHandler; // A reference to your thing handler
  *
- * Interface for a service that provides access to a {@link ThingHandler}
+ *     public void activate() {
+ *         super(Collections.emptyMap()); // start AbstractDiscoveryService
```
cweitkamp: I guess you want to call the super method. And passing `null` should be sufficient.

```suggestion
 *         super.activate(null); // start AbstractDiscoveryService
```

bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandlerService.java
```diff
@@ -13,11 +13,25 @@
...
+ *     }
+ *     public void setThingHandler(ThingHandler handler) { this.thingHandler = handler; }
```
cweitkamp: I am not sure if we should add the `@Override` to all three methods. The IDE takes care of that automatically if you safe the file.

```suggestion
 *     @Override
 *     public void setThingHandler(ThingHandler handler) { this.thingHandler = handler; }
```

bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandlerService.java
```diff
@@ -13,11 +13,25 @@
...
+ *     public ThingHandler getThingHandler() { return this.thingHandler; }
```
cweitkamp:
```suggestion
 *     public @Nullable ThingHandler getThingHandler() { return this.thingHandler; }
```

davidgraeff:
`getServices()` was not meant to be used for thing handler bound discovery services. The underlying implementation would need to be changed. Closing for now.

PR #721, opened 2019-04-14
[core] Removed deprecated methods

cweitkamp:
- Removed deprecated methods in `ThingImpl` and `BridgeImpl` and as well in the related builders

Follow-up PR off #693


kaikreuzer:
I'd agree with @maggu2810 that this should normally only be done with a version change, so ideally when we have introduced package versioning (or at least move the master to OH3).

If we are sure that no code in openHAB repos is using these methods/constructors, I am fine to remove them now as well. @cweitkamp Did you check whether they are still in use somewhere?

cweitkamp:
> Did you check whether they are still in use somewhere?

Yes, I did. `BridgeImpl` and `ThingImpl` are only used in OHC. `BridgeBuilder` and `ThingBuilder` are used here and in add-ons repositories. A spot check on the deleted method (see https://github.com/search?q=org%3Aopenhab+BridgeBuilder&type=Code and https://github.com/search?q=org%3Aopenhab+ThingBuilder&type=Code) revealed no external usage. I think we are safe.

PR #724, opened 2019-04-15
[JSR223] Sort by filename instead of path

openhab-5iver:
Currently, the scripts are loaded by ScriptFileWatcher based on the lexicographical order of the absolute paths of the scripts. This makes it difficult to control the load order. This change bases the load order solely on the filename, as was originally used before [ESH/#3855](https://github.com/eclipse/smarthome/pull/3855), and preserves the ability to use scripts with the same filename. This PR resolves https://github.com/OH-Jython-Scripters/openhab2-jython/issues/76.

## After this PR
Scripts will be loaded based on the filename. If more than one script have the same filename, their paths will also be taken into account for the sorting. This is especially helpful with the direcotry structure adopted by the Jython helper libraries. Currently, Community scripts (in a community directory), are being loaded before the Core scripts (in a core directory).

**Directory structure and scripts**
```
├── /automation/sr223
│   ├── 00
│   │   ├── 00
│   │   │   ├── 00_script.py
│   │   │   ├── 01_script.py
│   │   │   └── another.py
│   │   ├── 00_script.py
│   │   ├── 01
│   │   │   ├── 00_script.py
│   │   │   ├── 01_script.py
│   │   │   └── another.py
│   │   ├── 01_script.py
│   │   └── another.py
│   ├── 00_script.py
│   ├── 01
│   │   ├── 00
│   │   │   ├── 00_script.py
│   │   │   ├── 01_script.py
│   │   │   └── another.py
│   │   ├── 00_script.py
│   │   ├── 01
│   │   │   ├── 00_script.py
│   │   │   ├── 01_script.py
│   │   │   └── another.py
│   │   ├── 01_script.py
│   │   └── another.py
│   ├── 01_script.py
│   ├── another.py
```
**Load order**
```
/00_script.py
/00/00_script.py
/00/00/00_script.py
/00/01/00_script.py
/01/00_script.py
/01/00/00_script.py
/01/01/00_script.py
/01_script.py
/00/01_script.py
/00/00/01_script.py
/00/01/01_script.py
/01/01_script.py
/01/00/01_script.py
/01/01/01_script.py
/another.py
/00/another.py
/00/00/another.py
/00/01/another.py
/01/another.py
/01/00/another.py
/01/01/another.py
```
@lewie, I'm curious what your thoughts are about this change.

### DIFF
--- a/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/loader/ScriptFileWatcher.java
+++ b/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/loader/ScriptFileWatcher.java
@@ -21,6 +21,7 @@ import java.io.InputStreamReader;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.nio.file.WatchEvent;
 import java.nio.file.WatchEvent.Kind;
 import java.util.Comparator;
@@ -252,9 +253,17 @@ public class ScriptFileWatcher extends AbstractWatchService {
         SortedSet<URL> reimportUrls = new TreeSet<URL>(new Comparator<URL>() {
             @Override
             public int compare(URL o1, URL o2) {
-                String f1 = o1.getPath();
-                String f2 = o2.getPath();
-                return String.CASE_INSENSITIVE_ORDER.compare(f1, f2);
+                Path path1 = Paths.get(o1.getPath());
+                Path path2 = Paths.get(o2.getPath());
+                String name1 = path1.getFileName().toString();
+                String name2 = path2.getFileName().toString();
+                int nameCompare = name1.compareToIgnoreCase(name2);
+                if (nameCompare != 0) {
+                    return nameCompare;
+                } else {
+                    int pathCompare = path1.getParent().toString().compareToIgnoreCase(path2.getParent().toString());
+                    return pathCompare;
+                }
             }
         });
 


maggu2810:
I cannot judge, which sort order makes most sense, but I would expect (what reason ever) a sort order like:

```
function load(dir):
* sort all files in the current directory by name and load file by file
* sort all directories in the current directory by name and call "load(subdir)" for each of them
```

So, go down the directory hierarchy after the files of that directory has been checked.

Your example:

```
/00_script.py
/01_script.py
/another.py
/00/00_script.py
/00/01_script.py
/00/another.py
/00/00/another.py
/00/00/00_script.py
/00/00/01_script.py
/00/01/another.py
/00/01/00_script.py
/00/01/01_script.py
/01/00_script.py
/01/01_script.py
/01/another.py
/01/00/00_script.py
/01/00/01_script.py
/01/00/another.py
/01/01/00_script.py
/01/01/01_script.py
/01/01/another.py
```

openhab-5iver:
A script loading order based on the directory structure (absolute path) is _a lot_ harder to work with. Currently, with the load order based on the absolute path, the only way to have certain scripts load before any others, is to manipulate the directory names, or put them in the root directory and manipulate the file name. So, scripts in `/000_Startup_Scripts/` would load before scripts in directories that are not prefixed with a number. Sorting by absolute paths also means that I cannot distribute a solution that you can just copy into `/automation/jsr223/`, if it contains anything that has a specific load order, because I do not know what your directory structure is.

This PR allows for me to distibute solutions that are contained in their own directories, which can be easily copy/pasted, and the load order, relative to other directories that have been distributed, are preserved because the load order is based off of the file name. The openhab-jython repo relies on a specific directory structure. Without this PR, users will have to modify the directory and file names so that the script load order works properly with their implementation. Specific example...

Currently, the owm_daily_forecast.py will load first, and fail, since the 000_starup_delay.py is required to run first and delay the loading of all other scripts until the environment is full loaded. The script will also fail since it uses a module that depends on 200_JythonItemProvider.py being loaded.

├── automation/jsr223
│   ├── core
│   │   ├── 000_startup_delay.py
│   │   └── components
│   │       ├── 200_JythonBindingInfoProvider.py
│   │       ├── 200_JythonExtensionProvider.py
│   │       ├── 200_JythonItemChannelLinkProvider.py
│   │       ├── 200_JythonItemProvider.py
│   │       ├── 200_JythonThingProvider.py
│   │       ├── 200_JythonThingTypeProvider.py
│   │       ├── 200_JythonTransform.py
│   │       ├── 300_DirectoryEventTrigger.py
│   │       ├── 300_OsgiEventTrigger.py
│   │       ├── 300_ShutdownTrigger.py
│   │       └── 300_StartupTrigger.py
│   ├── community
│   │   ├── openweathermap
│   │   │   └── owm_daily_forecast.py

In reading through this several times, it all still looks like it would be confusing for someone who is not using scripted automation. I'll see if I can't drum up some others to help communicate the benefits of this change.

mjcumming:
We definitely need a reliable way to ensure scripts are loaded in the correct order. The proposed naming convention solves this problem without worrying about directory names or structure.

maggu2810:
Don't get me wrong, I don't argue for a specific sorting algorithm, I just want to understand why the one is better then the other.

> A script loading order based on the directory structure (absolute path) is _a lot_ harder to work with. Currently, with the load order based on the absolute path, the only way to have certain scripts load before any others, is to manipulate the directory names, or put them in the root directory and manipulate the file name. So, scripts in `/000_Startup_Scripts/` would load before scripts in directories that are not prefixed with a number. Sorting by absolute paths also means that I cannot distribute a solution that you can just copy into `/automation/jsr223/`, if it contains anything that has a specific load order, because I do not know what your directory structure is.

Hm, I don't understand.
What is a "solution" that you distribute?
Why is it not possible to distribute if an absolute path based sorting is used that respect directory names and file names?
If we distribute all the scripts you need, you can ensure the load order by its file names and they just need to be added in the same directory.
You could also use different directories to group all your stuff and create an archive of this directory structure.
If you rely on other stuff that you don't distribute, why should it work using file names only -- you don't know how the files are named on a specific system. Or do you know file names someone is using but not directory names?
What's the difference between directory names and file names?

> This PR allows for me to distibute solutions that are contained in their own directories, which can be easily copy/pasted, and the load order, relative to other directories that have been distributed, are preserved because the load order is based off of the file name. The openhab-jython repo relies on a specific directory structure. Without this PR, users will have to modify the directory and file names so that the script load order works properly with their implementation. Specific example...
> 
> Currently, the owm_daily_forecast.py will load first, and fail, since the 000_starup_delay.py is required to run first and delay the loading of all other scripts until the environment is full loaded. The script will also fail since it uses a module that depends on 200_JythonItemProvider.py being loaded.

Same here.
You state that the correct file names needs to be used otherwise the loading order will not work.
So the file name must not be changed.
Same can be true for directory names.

If my script is named `000a_foo.py` it will be loaded first.
If my script is names `100_foo.py` it will be loaded in front of the 200 ones.

You could also state that the script you provide needs to be placed in a directory "000" and the user must not use the root directory of `jsr223` but a directory names that is lexically ordered after "000".

openhab-5iver:
> Don't get me wrong, I don't argue for a specific sorting algorithm, I just want to understand why the one is better then the other.

I understand your position. You want to be certain that the best solution is chosen, but don't have experience with scripted automation. You are the gatekeeper and we need and appreciate you! 

> What is a "solution" that you distribute?

The openhab2-jython repo contains community provided modules and scripts, which a user can install by "cherry picking" them out of the directory structure and copying to their OH installation. This is just a stepping stone until we have an automation extension service. The organization is about to be renamed to openHAB Scripters, and separate repos will be created for each scripting language and their helper libraries, using the same directory format.

In a test environent, [setup the Jython helper libraires and scripts](https://github.com/OH-Jython-Scripters/openhab2-jython/blob/master/Docs/Getting-Started.md#quick-start-guide) and you might see what I mean. I have not added [much](https://github.com/OH-Jython-Scripters/openhab2-jython/blob/master/Docs/Community.md) to the doc for the Community stuff, mainly because it doesn't work due to the current load order. I've also held up submitting LOTS of examples until we get something like this PR merged. The alternative is to add numeric prefixes to the directory names, which is rather ugly and does not solve the issue in all cases.

Say that we have renamed the core script directory with a numeric prefix so that these scripts will load first using the current load order. But there is a community script that needs to load in the middle of the core scripts...

├── automation/jsr223
│   ├── 000_core
│   │   ├── 01_loadme_first.py
│   │   ├── 03_loadme_third.py
│   ├── community
│   │   ├── 02_loadme_second.py

With the current load order, these scripts would load 

`/000_core/01_loadme_first.py`
`/000_core/02_loadme_first.py`
`/community/02_loadme_second.py`

There is currently no way to interweave the load order of scripts from separate directories. This PR provides this functionality focusing on the file name.

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/loader/ScriptFileWatcher.java
```diff
@@ -252,9 +253,17 @@ private void checkFiles() {
...
-                return String.CASE_INSENSITIVE_ORDER.compare(f1, f2);
+                Path path1 = Paths.get(o1.getPath());
+                Path path2 = Paths.get(o2.getPath());
+                String name1 = path1.getFileName().toString();
+                String name2 = path2.getFileName().toString();
+                int nameCompare = name1.compareToIgnoreCase(name2);
+                if (nameCompare != 0) {
+                    return nameCompare;
+                } else {
+                    int pathCompare = path1.getParent().toString().compareToIgnoreCase(path2.getParent().toString());
```
maggu2810: AFAIK `getParent()` can return null, so this could result into a NPE.
But 
1. this file does not use nullness checks at all
2. I assume we can be sure that there will be a parent

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/loader/ScriptFileWatcher.java
```diff
@@ -252,9 +253,17 @@ private void checkFiles() {
...
+                    return pathCompare;
```
maggu2810: You should not create a temporary local variable if you only intend to return it immediately.
But lets move forward and fix it another time.

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/loader/ScriptFileWatcher.java
openhab-5iver: Noted. I will add nullness checks at some point when cleaning up automation. For this specific case, the parent cannot be null, since the scripts have to reside in /automation/jsr223/ for them to be loaded.

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/loader/ScriptFileWatcher.java
openhab-5iver: OK, thank you. I thought I had seen similar done elsewhere in the code. I will keep this in mind in the future.

PR #732, opened 2019-04-15
Proposal for some new profiles : Change and update timestamp

clinique:
Would like your comments.

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemProfileFactory.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemProfileFactory.java
@@ -55,15 +55,17 @@ public class SystemProfileFactory implements ProfileFactory, ProfileAdvisor, Pro
     @NonNullByDefault({})
     private ChannelTypeRegistry channelTypeRegistry;
 
-    private static final Set<ProfileType> SUPPORTED_PROFILE_TYPES = Stream.of(SystemProfiles.DEFAULT_TYPE,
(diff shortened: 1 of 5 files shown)


kaikreuzer:
> The question is if it makes sense to extend items / events in general to carry their last change timestamps?

This has been discussed a couple of times in the past years, but although it sounds easy, it isn't straight forward. An event can be passed on. Should its timestamp be updated in that situation or not? What if you receive an event with a lower timestamp AFTER one with a higher one? Shouldn't the timestamp be the time of the measurement and not the time of arriving in the framework? What timestamp should be put in the datebase for persistence (note that the HistoricItem already has a timestamp).

The best solution to avoid such questions might be to have some "lastUpdated" method on the ItemRegistry and not on events and or items themselves.

kaikreuzer:
But all the above is rather off-topic. Wrt this PR, I think the profile is ok, although it is pretty weird that the original state is not at all considered - it thus feels a bit like a misuse of the profile idea...

clinique:
> ..although it is pretty weird that the original state is not at all considered...

Can you elaborate a bit on this ? I do not get your point. Thks

clinique:
@cweitkamp : now that I'm on it, do you think it would be appropriate to turn expire binding into a profile ?

cweitkamp:
@clinique Hm, from my point of view it will fit. But I do not want to decide that on my own. @openhab/2-x-add-ons-maintainers Wdyt?

davidgraeff:
Actually I had the same thought a while ago. The issue that I'm seeing with it is that there is no way to get an overview of "running" expires. 
I know that this is also true for OH1-like expires, but I guess we can do better this time. Knowing how many expire timers are in the queue helps maintaining the system.

So maybe going with a profile and a `Registry` that keeps track of active expire profiles (that can be accessed by Rest later on)?

PR #734, opened 2019-04-16
RFC: Natural language processor (NLP) with modular skill architecture

davidgraeff:
Adapted from HaBot to be independent from the HaBot card system.
Thank you Yannick Schaus for this great piece of code.

This is a Request for comments PR. Do you think it belongs into core or does it classify as addon?

The base bundle (org.openhab.nlp) is to be used with openHAB voice services.
Coupled with speech-to-text and text-to-speech engines it allows to build
privacy-focused specialized voice assistant services.

The rest bundle (org.openhab.nlp.rest) allows to interact in a chat like form.
Just send your sentence to `/nlp/chat` and corresponding actions are performed and
and an answer is generated. Use `/nlp/greet` for an initial greeting.

Each bundle features its own readme including a
"Develop for NLP in openHAB" section for skill development.

Some of the original skills were user-interface only and are
not part of this PR. The generated `IntentInterpretation`
contains all relevant information to allow skills that are ui centric as well.
A user interface system would only need to provide a separate
rest interface and might provide additional ui-specific skills.

Also-by: Yannick Schaus <habpanel@schaus.net>


maggu2810:
IMHO It sounds like a good idea to separate the NLP stuff and add it to core.

maggu2810:
IMHO It sounds like a good idea to separate the NLP stuff and add it to core.
But for a better analysis I would have to look at the code.
Can the NLP stuff be enhanced by other bundles (addons) or is this an addon to the core itself and is not extended anymore?

davidgraeff:
> Can the NLP stuff be enhanced by other bundles (addons) or is this an addon to the core itself and is not extended anymore?

It can be extended by so called "skills". A user-interface specific REST bundle would also extend the `IntentInterpretation` interface for very specific needs, I guess.

kaikreuzer:
I didn't yet have a closer look yet, but wouldn't this qualify to be an [HLI implementation](https://www.openhab.org/docs/configuration/multimedia.html#human-language-interpreter) (maybe one to replace the "system" one)? (@ghys fyi)

ghys:
Why am I only now learning about this? 😒
A heads-up and a proper mention would have been nice when my code is being refactored...

> this qualify to be an HLI implementation

Agreed, HABot is one but it needed an "enriched" `reply` method (`interpret` only returns a string, and I didn't want to propose an core API change at the time due to the need for specific stuff like cards etc.)

lewie:
Thank you @davidgraeff and @ghys, I'm very impressed!!! When do we get this in master?

maggu2810:
> Adapted from HaBot to be independent from the HaBot card system.

So, is this taken from the HABot bundle and HABot needs to be refactored to base on this core component (if merged)?

@ghys Did you already reiew this addition? Do you agree? Do you complain something?

davidgraeff:
APIs are not changed. Only package renaming would be required for habot. The idea was to have no java code at all in webui.

lolodomo:
Can we make progress with the new HLI? 
I have not yet played with Habot (but I will) but if its NLP is so efficient, we should consider it as replacement of our system HLI ASAP.

lolodomo:
@davidgraeff : I take a look and I like this move allowing to define the clever stuff behind HaBot as a new Human Language Interpreter in openHAB. Are you still available for considering review comments ?

As @davidgraeff said, this is mainly package renaming + a new REST API to be used by HaBot (for example).

My first remarks are against the strange package naming you used.
IMHO, org.openhab.nlp should be renamed into org.openhab.core.voice.nlp and org.openhab.nlp.rest should be renamed into org.openhab.core.io.rest.voice.nlp.

They are few changes, the main one is of course that Card is no more in the ChatReply but is replaced by the data required to build the Card later. So HaBot will request the API to get a reply and then will build the Card with the data inside the reply.

Something to consider, I see that 2 skills are missing. Is it intentional or were these skills added after this PR ?

Of course I can do a full review but the one that is expected here is of course @ghys . Can you review please and provide a feedback ?

PS: I am still thinking how to integrate Snips into openHAB and one of my current ideas would be to let the user link any Snips intent to an openHAB NLP skill.

lolodomo:
Italian language is missing in the PR (while present in HaBot).

davidgraeff:
"org.openhab.core.voice.nlp" -> habot uses nlp also for the chat input and nlp in general can be used for chat like interfaces. "voice" is imo not necessary here.

And there are other packages that are not part of "core" and it is actually nowhere defined what should be in "core". It is imo redundant, because addons must be in "org.openhab.addon.x" so there is no confusion coming from that side. We could argue that every official bundle aka every bundle in this repo must be in "core". That's fine of course.

lolodomo:
All packages from the core framework are put in org.openhab.core.XXX
All REST API are defined in org.openhab.core.io.rest.XXX

nlp is for Natural Language Processor and is at 95% an implementation of what is called a Human Language Interpreter in the openHAB core framework.
As HLI is defined org.openhab.core.voice, it makes sense to put NLP in org.openhab.core.voice.nlp in my point of view.

@kaikreuzer : can you please confirm as a maintener of the core framework what is the package location the most appropriate for this new Natural Language Processor.

ghys:
About this PR - while it makes sense to add a standardized NLP API to the core - or even merge it with the HLI interface which have more a voice connotation), I still feel any non-trivial implementation of it should remain in an add-on that users can (un)load at will, rather than introducing a non-critical OpenNLP dependency to the core (or maybe it doesn't and I didn't understand?).

lolodomo:
If the dependency on OpenNLP is a problem in the core framework, we should move that to openHAB2 repo and create a HLI binding. Each user can then install it or not.
The only point I am not sure is the REST API: is a binding able to add a REST API ? Will it be viewed by REST docs in this case ?

lolodomo:
We already have one REST API for HLI in the existing core framework. So the new REST API is only for chat purpose, I doubt it is a real requirement for HABot.

lolodomo:
In fact, we need this in core because HABot needs them and HABot which is in an addon in webui repo will not be able to use packages from openhab2 repo.

lolodomo:
In fact the clean way would be I thing:

- defining in core an interface ChatBotHumanLanguageOnterpreter  + a chatbot manager (or enhancing the voice manager). 

- implement in openhab2 a nlp binding as a ChatBotHumanLanguageOnterpreter and HumanLanguageOnterpreter

- Update HABot to work with any installed ChatBotHumanLanguageOnterpreter

Does it make sense?

lolodomo:
I will first propose an extension of HumanLanguageOnterpreter covering the extended need (reply).

lolodomo:
@ghys : I have not yet looked at HABot front-end. Can you tell me if you really need the `Intent` object in the `ChatReply` ?

Issue #698, opened 2019-04-05
Fix and reenable disabled integration tests


wborn:
Some integration tests are still disabled ever since the migration from PDE to bnd. As new PRs are merged these tests will only be broken even more. So it would certainly help if we can fix and reenable these ASAP.

The currently disabled tests (commented in [itests/pom.xml](https://github.com/openhab/openhab-core/blob/master/itests/pom.xml#L23)) are:

- [x] org.openhab.core.automation.integration.tests (#1061)
- [x] org.openhab.core.automation.module.core.tests (#1062)
- [x] org.openhab.core.automation.module.script.defaultscope.tests (#1062)
- [x] org.openhab.core.automation.module.script.tests (#1062)
- [x] org.openhab.core.automation.module.timer.tests (#1062)
- [x] org.openhab.core.config.discovery.usbserial.linuxsysfs.tests (#1065)
- [x] org.openhab.core.config.discovery.usbserial.tests (#1065)
- [x] org.openhab.core.config.dispatch.tests (#995)
- [x] org.openhab.core.config.xml.tests (#971)
- [x] org.openhab.core.io.http.tests (#700)
- [x] ~org.openhab.core.io.rest.sse.tests~ moved to pure unit tests (#1044)
- [x] ~org.openhab.core.io.rest.tests~ moved to pure unit tests (#1055)
- [x] ~org.openhab.core.io.transport.mqtt.tests~ moved to pure unit tests (#1056)
- [x] ~org.openhab.core.io.transport.upnp.tests~ moved to pure unit tests (#1057)
- [x] org.openhab.core.magic.tests (#756)
- [x] org.openhab.core.model.core.tests (#728)
- [x] org.openhab.core.model.item.tests (#735)
- [x] org.openhab.core.model.lsp.tests (#730)
- [x] org.openhab.core.model.persistence.tests (#729)
- [x] org.openhab.core.model.rule.tests (#736)
- [x] org.openhab.core.model.script.tests (#737)
- [x] org.openhab.core.semantics.tests (#697)

I've already created a PR (https://github.com/openhab/openhab-core/pull/697) for reenabling `org.openhab.core.semantics.tests`. :-)

cweitkamp:
`org.openhab.core.magic.tests` is part of #655.

cweitkamp:
FYI: I am working on: `org.openhab.core.model.*`

openhab-5iver:
[...]

PR #737, opened 2019-04-16
Enable Model Script Tests

cweitkamp:
- Enable Model Script Tests

Related to #698

Currently failing.

### DIFF
--- a/itests/org.openhab.core.model.script.tests/src/main/java/org/eclipse/smarthome/model/script/actions/ScriptExecutionTest.java
+++ b/itests/org.openhab.core.model.script.tests/src/main/java/org/eclipse/smarthome/model/script/actions/ScriptExecutionTest.java
@@ -29,7 +29,6 @@ import org.quartz.impl.StdSchedulerFactory;
  * Tests for {@link ScriptExecution}
  *
  * @author Jon Evans - Initial contribution
- *
  */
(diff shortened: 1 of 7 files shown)


itests/org.openhab.core.model.script.tests/src/main/java/org/eclipse/smarthome/model/script/actions/ScriptExecutionTest.java
```diff
@@ -42,7 +41,7 @@
     @BeforeClass
     public static void setUp() throws SchedulerException {
         scheduler = new MockScheduler();
-        System.setProperty(StdSchedulerFactory.PROPERTIES_FILE, "quartz-test.properties");
+        System.setProperty(StdSchedulerFactory.PROPERTIES_FILE, "../../../../../target/classes/quartz-test.properties");
```
cweitkamp: It is working now but uses the same ugly path workaround like the Config Dispatch tests: https://github.com/openhab/openhab-core/pull/995#discussion_r318601336

wborn: Let's use this workaround and then we'll remove it when we stop using quartz.

Issue #660, opened 2019-03-15
[openhab.boot] Split


davidgraeff:
Hm. I keep finding sitemap stuff.

The `openhab.boot` package serves multiple different purposes. I propose to

* move "IconForwarder" to combat1x (it proxies /images to /icons for OH1 rest consumers).
* move "DefaultSitemapProvider" to the "ui.sitemap" bundle
* Rename "CoreActivator" to "ModelRuleEngineStart" and move it to "org.openhab.core.model.rule.runtime"
* Move "OpenHab.java" to "org.openhab.core.config". The class provides the static method `getVersion`.
Unfortunately it cannot be moved to " org.openhab.core", because it requires config.core classes to determine the user directory.

Additionally I like to add a REST endpoint "about" that allows to request the openhab version from that mentioned class. A UI cannot request any versions at the moment. Fascinating somehow.

davidgraeff:
@openhab/core-maintainers BUMP

maggu2810:
IMHO

* `OpenHAB`
  Methods like `getBuildVersion()`, `getVersion()` and similar should be provided by a service and this one should be used instead of static functions.
  I would also prefer if first an optional configuration provided by config admin is used and if there is no one, a fallback (e.g. the current code) is used. This allows others solutions to provide overwrite the version by a configuration.

> * Rename "CoreActivator" to "ModelRuleEngineStart" and move it to "org.openhab.core.model.rule.runtime"

okay for me

> move "IconForwarder" to combat1x (it proxies /images to /icons for OH1 rest consumers).

combat1x => compat1x
But could make sense as the redirect from old to new needs only be used if we want to provide a compatibility.

> * move "DefaultSitemapProvider" to the "ui.sitemap" bundle

makes sense

davidgraeff:
A service makes sense as yes

Issue #422, opened 2018-11-02
html/static folder still throws 404 after ui.start bundle commit


lsiepel:
The issue listed here: https://github.com/openhab/openhab-core/issues/420 should be fixed with https://github.com/openhab/openhab-core/pull/421 but with build #1409 apt version: 2.4.0~20181102181428-1 it still throws 404 on the static folder.

kaikreuzer:
It shows a slightly different one, but I agree that the issue isn't fixed :-(

kaikreuzer:
Grrr. While it all worked smoothly in the IDE, it still isn't working in the final distro - so I have to reopen this again.

PR #738, opened 2019-04-16
About REST endpoint

davidgraeff:
Provides core and distribution information via REST for user-interfaces.

As of now user-interfaces do not have the means to query the core for
the solution name, version, and distribution information and that information
was backed in into the deprecated dashboard concept.

This is a first, simple proposal. The rest endpoint is `/rest/about`.

The use case is to present those informational data
like here https://davidgraeff.github.io/paperui-ng/maintenance.html at the top part of the page.

### DIFF
--- a/features/karaf/openhab-core/src/main/feature/feature.xml
+++ b/features/karaf/openhab-core/src/main/feature/feature.xml
@@ -543,6 +543,7 @@
     <!-- It is temperarily disabled due to https://github.com/openhab/openhab-core/issues/422 -->
     <!-- <bundle start-level="30">mvn:org.openhab.core.bundles/org.openhab.ui.start/${project.version}</bundle> -->
     <bundle start-level="90">mvn:org.openhab.core.bundles/org.openhab.core.boot/${project.version}</bundle>
+    <bundle start-level="90">mvn:org.openhab.core.bundles/org.openhab.core.boot.rest/${project.version}</bundle>
     <bundle>mvn:org.openhab.core.bundles/org.openhab.core.karaf/${project.version}</bundle>
     <config name="org.eclipse.smarthome.audio">
       defaultSink = enhancedjavasound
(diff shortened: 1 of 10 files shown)


features/karaf/openhab-core/src/main/feature/feature.xml
```diff
@@ -543,6 +543,7 @@
     <!-- It is temperarily disabled due to https://github.com/openhab/openhab-core/issues/422 -->
     <!-- <bundle start-level="30">mvn:org.openhab.core.bundles/org.openhab.ui.start/${project.version}</bundle> -->
```
hakan42: Issue #422 has been fixed / closed some time ago :)

davidgraeff: I guess that line needs to be removed anyway, because there is no `org.openhab.ui.start` in this core repo.

maggu2810:
Sorry, did not read the changes at the moment, but I assume it can be combined with the first part of my comment here https://github.com/openhab/openhab-core/issues/660#issuecomment-484244355 (the one about `OpenHAB`).
There should be an option to provide that informations by a configuration and fallback to a default one.
The REST API just expose that methods.

davidgraeff:
Yes I agree. I was just reluctant touching .boot, without conferring with someone. I will amend this PR then.

kaikreuzer:
We should not start adding new bundles for any new REST resource. If you consider this useful for any installation (which I assume you do), it should be part of `io.rest.core` just as all the other "standard" resources.
Instead of adding a new `/about` url, I could even imagine to add this info directly to the `RootBean` (and thus to `io.rest`) - after all, this is what is called if you want to figure out what this REST API is about.

features/karaf/openhab-core/src/main/feature/feature.xml
lsiepel: More details: https://github.com/openhab/openhab-core/issues/422#issuecomment-435622005

Issue #597, opened 2019-02-23
SSE breaks due to BundleException in SseActivator.start() on first startup


ScottKinSF:
My test system is running OH 2.5.0-SNAPSHOT s1530 on up-to-date OpenHABian on an RPi3. When the bundle org.openhab.core.io.rest.sse is started, the following exception is logged:
```
2019-02-23 11:07:57.836 [ERROR][lisher@754ba872][org.openhab.core.io.rest.sse        ] - FrameworkEvent ERROR - org.openhab.core.io.rest.sse
org.osgi.framework.BundleException: Exception in org.eclipse.smarthome.io.rest.sse.internal.SseActivator.start() of bundle org.openhab.core.io.rest.sse.
        at org.eclipse.osgi.internal.framework.BundleContextImpl.startActivator(BundleContextImpl.java:800) ~[?:?]
        at org.eclipse.osgi.internal.framework.BundleContextImpl.start(BundleContextImpl.java:729) [?:?]
        at org.eclipse.osgi.internal.framework.EquinoxBundle.startWorker0(EquinoxBundle.java:933) ~[?:?]
        at org.eclipse.osgi.internal.framework.EquinoxBundle$EquinoxModule.startWorker(EquinoxBundle.java:309) ~[?:?]
        at org.eclipse.osgi.container.Module.doStart(Module.java:581) ~[?:?]
        at org.eclipse.osgi.container.Module.start(Module.java:449) ~[?:?]
        at org.eclipse.osgi.framework.util.SecureAction$23.run(SecureAction.java:474) ~[?:?]
        at org.eclipse.osgi.framework.util.SecureAction$23.run(SecureAction.java:1) ~[?:?]
        at java.security.AccessController.doPrivileged(Native Method) ~[?:?]
        at org.eclipse.osgi.framework.util.SecureAction.start(SecureAction.java:472) ~[?:?]
        at org.eclipse.osgi.container.ModuleContainer.start(ModuleContainer.java:750) ~[?:?]
        at org.eclipse.osgi.container.ModuleContainer.applyDelta(ModuleContainer.java:741) ~[?:?]
        at org.eclipse.osgi.container.ModuleContainer.resolveAndApply(ModuleContainer.java:511) ~[?:?]
        at org.eclipse.osgi.container.ModuleContainer.resolve(ModuleContainer.java:457) ~[?:?]
        at org.eclipse.osgi.container.ModuleContainer.refresh(ModuleContainer.java:1001) ~[?:?]
[...]

PR #739, opened 2019-04-16
[WIP] Use OSGi Compendium R7 JAX-RS Whiteboard pattern for REST

### LINKED ISSUE #726: Remove JAX-RS Connector (com.eclipsesource.jaxrs) dependency
<a alt="bounty" href="https://www.bountysource.com/issues/72862366"></a>

### What does this dependency do?

It scans all OSGi services (by STARTING THEM!!) for JAX-RS `@Path` or `@Provided` annotations and adds them to a servlet.

### The problems with this dependency

Unfortunately the package is not maintained (no major update since 5 years, no commit in general since 3 years). It is not using newer OSGi features (the entire codebase would shrink by half or more), it uses tycho as its main buildsystem with respective outdated examples, and it does really bad things like actually starting up services to find annotations.

I debugged my Hue Emulation Service for probably 3 hours to find out that double service starts are caused by that bundle. Especially if you do a lot of heavy initialization in activate() (even if async) to throw that away moments later, that adds a lot to OH start up time.

### Solution 

We have the huge advantage in openHAB, that all our REST endpoints not only are annotated with `@Path` or `@Provided` but also implement `RestResouce` and register themselves to the rest core bean. We can therefore spin up our own servlet in about 20 lines of code without this really badly designed dependency.

Example code (without debouncing/restarting to accommodate for dynamically appearing endpoints):

```java
ResourceConfig resourceConfig = ResourceConfig.forApplicationClass(JerseyApplication.class);

resourceConfig.property(ServerProperties.APPLICATION_NAME, "openHAB");

// don't look for implementations described by META-INF/services/*
resourceConfig.property(ServerProperties.METAINF_SERVICES_LOOKUP_DISABLE, false);

// disable auto discovery (class path scanning)
resourceConfig.property(ServerProperties.FEATURE_AUTO_DISCOVERY_DISABLE, true);
resourceConfig.property(ServerProperties.PROCESSING_RESPONSE_ERRORS_ENABLED, true);

resourceConfig.registerInstances(allOurRestResourceInstancesList);

try {
    Hashtable<String, String> initParams = new Hashtable<>();
    initParams.put("com.sun.jersey.api.json.POJOMappingFeature", "false");
    initParams.put(ServletProperties.PROVIDER_WEB_APP, "false");
    httpService.registerServlet("/rest", new ServletContainer(resourceConfig), initParams, null);
} catch (ServletException | NamespaceException e) {
}
```

### Notes

This dependency is not related to `com.eclipsesource.jaxrs.provider.swagger` which is also used by openHAB. `provider.swagger` configures swagger via a properties file and is apart from that unrelated to jax-rs.

davidgraeff:
Fixes #726

* Remove all associated quirks and workarounds ("core.io.rest.optimize")
* Simplify RootResource of org.openhab.core.io.rest. We don't need to update
  osgi-jax-rs-connector's configuration anymore.
* Add JerseyServletContainerInitializer to satisfy pax-web. Should be
  removed when pax-web is not used anymore.
* No double-service activation anymore
* About 3k lines less code by using modern JAX-RS features.
* Implementation independent. No glassfish, jersey or any other specific features required.

TODO:
* I'm not sure if the JAX-RS specified SSE works in a way that the Sitemap implementation is using it (in a subscription like fashion with sessions). That need to be verified, but not by me as I'm not using sitemaps.
* Finish adapting tests. Use test setup of Hue emulation service and spin up real jax-rs instances for testing.

The RESTdoc bundle in the webui repo probably requires an update
to query the introduced `RESTservice` interface for the REST root.


bundles/org.openhab.core.io.rest/src/main/java/org/eclipse/smarthome/io/rest/internal/resources/RootResource.java
```diff
@@ -90,45 +116,70 @@ private RootBean getRootBean() {
...
-                if (!RESTConstants.REST_URI.equals(rootAlias)) {
-                    properties.put(Constants.JAXRS_CONNECTOR_ROOT_PROPERTY, RESTConstants.REST_URI);
+        try {
+            Hashtable<String, String> initParams = new Hashtable<>();
+            initParams.put("com.sun.jersey.api.json.POJOMappingFeature", "false");
+            initParams.put(ServletProperties.PROVIDER_WEB_APP, "false");
+            httpService.registerServlet(restRoot, servletContainer, initParams, null);
+            logger.info("openHAB REST started on {}", restRoot);
+        } catch (ServletException | NamespaceException e) {
+            logger.warn("Could not start REST API on {}: {}", restRoot, e.getMessage(), e);
```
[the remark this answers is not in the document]
davidgraeff: That depends if we see REST as essential. But yeah I'll change that

bundles/org.openhab.core.io.rest/src/main/java/org/eclipse/smarthome/io/rest/internal/resources/RootResource.java
```diff
@@ -90,45 +116,70 @@ private RootBean getRootBean() {
...
+    public String getRESTroot() {
+        return restRoot;
     }
 
-    protected void unsetConfigurationAdmin(ConfigurationAdmin configurationAdmin) {
-        this.configurationAdmin = null;
+    /**
+     * Debounce reloading of the servlet container
+     */
+    private void scheduleReload() {
```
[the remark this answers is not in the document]
davidgraeff: It should be the one and only service resolution thread in my naive assumption, but yes I agree. That should be `synchronized`.

davidgraeff:
> As org.openhab.core.io.rest.sitemap and org.openhab.core.io.rest.sse already use glassfish API, we could also use it here -- IMHO.

The `org.glassfish.jersey.servlet.ServletContainer` is glassfish.
Is there other, simpler API to be used from glassfish elsewhere?

The replaced dependency was also using the ServletContainer so I'm not introducing new dependencies to the table.

bundles/org.openhab.core.io.rest/src/main/java/org/eclipse/smarthome/io/rest/internal/resources/RootResource.java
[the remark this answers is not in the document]
maggu2810: My understanding has been that the root resource is exactly used for this purpose (REST).
Does it provide any other stuff, so it would make sense to mark it `active` if the `ServletException` or `NamespaceException` has been thrown. So, does the service make sense without the servlet registration succeeded?

bundles/org.openhab.core.io.rest/src/main/java/org/eclipse/smarthome/io/rest/internal/resources/RootResource.java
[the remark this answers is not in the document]
maggu2810: IIRC there has been already some comments about that topic.
There is no word in the spec that optional dependencies and the activation etc. are called from the same thread (AFAIK -- but I could be wrong).

maggu2810:
> The `org.glassfish.jersey.servlet.ServletContainer` is glassfish.
> Is there other, simpler API to be used from glassfish elsewhere?

I don't know.

> The replaced dependency was also using the ServletContainer so I'm not introducing new dependencies to the table.

My comment should already state that it is okay for me. :wink:

bundles/org.openhab.core.io.rest/src/main/java/org/eclipse/smarthome/io/rest/internal/resources/RootResource.java
```diff
@@ -90,45 +116,70 @@ private RootBean getRootBean() {
...
+            httpService.unregister(restRoot);
+        } catch (IllegalArgumentException ignore) {
         }
     }
 
-    @Reference
-    protected void setConfigurationAdmin(ConfigurationAdmin configurationAdmin) {
-        this.configurationAdmin = configurationAdmin;
+    @Override
+    public String getRESTroot() {
```
kaikreuzer: CamelCase -> `getRESTRoot`

bundles/org.openhab.core.io.rest/src/main/java/org/eclipse/smarthome/io/rest/RESTservice.java
```diff
@@ -0,0 +1,22 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.io.rest;
+
+/**
+ * Allows to query and control the service that publishes the REST servlet.
+ * 
+ * @author David Graeff - Initial contribution
+ */
+public interface RESTservice {
```
kaikreuzer: CamelCase -> `RESTService`

bundles/org.openhab.core.io.rest/src/main/java/org/eclipse/smarthome/io/rest/RESTservice.java
```diff
@@ -0,0 +1,22 @@
...
+    String getRESTroot();
```
kaikreuzer: -> `getRESTRoot`

bundles/org.openhab.core.io.rest/src/main/java/org/eclipse/smarthome/io/rest/internal/resources/RootResource.java
```diff
@@ -54,17 +68,29 @@
...
 
-    private ConfigurationAdmin configurationAdmin;
+    protected @NonNullByDefault({}) @Context UriInfo uriInfo;
+    private @NonNullByDefault({}) String restRoot;
+
+    @Reference
+    protected @NonNullByDefault({}) HttpService httpService;
+    private @NonNullByDefault({}) ServletContainer servletContainer;
+
+    private final ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor();
```
kaikreuzer: This adds a permanent additional (and during runtime idle) thread. I'd prefer to instead use the [common thread pool](https://github.com/openhab/openhab-core/blob/48d873a32ba0901a9b58f04c541f0f858fbfcd51/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/common/ThreadPoolManager.java#L55-L60) for this job.

davidgraeff:
I had a good read on OSGi core and compendium releases recently and I'm proposing another solution later that day. It builds on JAX-RS 2.1 for SSE (This closes #743) and the [JAX-RS Whiteboard Pattern](https://osgi.org/specification/osgi.cmpn/7.0.0/service.jaxrs.html) for REST.
That's possible because we already have Apache Aries as dependency (at least in the BOMs?!).

maggu2810:
There has been already some work on a Karaf feature for the whiteboard pattern (see Karaf mailing list).
I don't know if the whole part is working in Karaf 4.2.x as it does not support R7 officially (just some compendium implementations already support their part e.g. SCR).

davidgraeff:
Fortunately the JAX-RS whiteboard pattern is part of the compendium specification. We don't need a R7 OSGi core.
And yes according to mailing lists people are using karaf and aries to already have JAX-RS whiteboard.

I will try to remove pax-web or jersey while I'm on it. We don't need 3 competing javax.servlet providing libraries.

maggu2810:
@davidgraeff Did you move forward?

davidgraeff:
I have a branch that fully works with the [R7 compendium Jax-RS whiteboard pattern](https://osgi.org/specification/osgi.cmpn/7.0.0/service.jaxrs.html).

I need to adapt this PR's title & description. It is a few more lines to review, because I'm using the newest variant of jax-rs and reworked our REST bundles. SLOC are dropped by 3k.

I just didn't came to test all the finicky SSE parts with the Sitemap bundle, where you register a client first etc. That's so much antipattern and I think websockets should have been used here. That's why I'm so unmotivated to finish this bundle also considering that sitemaps will be dropped.

bundles/org.openhab.core.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/item/EnrichedItemDTOMapper.java
```diff
@@ -93,41 +93,50 @@ private static EnrichedItemDTO map(Item item, ItemDTO itemDTO, URI uri, boolean
...
+        if (desc == null) {
+            return null;
+        }
+        String pattern = desc.getPattern();
+        if (pattern == null) {
+            return desc;
+        } else {
+            try {
+                return TransformationHelper.isTransform(pattern)
+                        ? new StateDescription(desc.getMinimum(), desc.getMaximum(), desc.getStep(), "",
```
davidgraeff: I need to cheery pick #655 changes and apply them to this file again.

maggu2810: Wouldn't it make sense to rebase the working branch on a regular basis?

bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/ModuleTypeResource.java
```diff
@@ -62,41 +62,24 @@
...
-    @Reference(cardinality = ReferenceCardinality.OPTIONAL, policy = ReferencePolicy.DYNAMIC)
-    protected void setLocaleService(LocaleService localeService) {
-        this.localeService = localeService;
-    }
-
-    protected void unsetLocaleService(LocaleService localeService) {
-        this.localeService = null;
-    }
+@Component(immediate = true)
+@JaxrsApplicationSelect("(osgi.jaxrs.name=" + RESTService.REST_APP_NAME + ")")
```
maggu2810: For here and other places: Shouldn't something similar to

```java
@JaxrsApplicationSelect
@JaxrsName(RESTService.REST_APP_NAME)
```

work too?

My understanding has been that the property type has been introduced to not use constructs like `"(osgi.jaxrs.name=" + RESTService.REST_APP_NAME + ")"` anymore.

bundles/org.openhab.core.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/item/EnrichedItemDTOMapper.java
davidgraeff: I rebased, but for this particular file I chose to just take mine to not break the PR. But will reapply the made changes.

bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/ModuleTypeResource.java
```diff
@@ -62,41 +62,24 @@
...
-
-    @Reference(cardinality = ReferenceCardinality.OPTIONAL, policy = ReferencePolicy.DYNAMIC)
-    protected void setLocaleService(LocaleService localeService) {
-        this.localeService = localeService;
-    }
-
-    protected void unsetLocaleService(LocaleService localeService) {
-        this.localeService = null;
-    }
+@Component(immediate = true)
```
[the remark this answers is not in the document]
davidgraeff: That might be true, not sure.

In contrast to the pure whiteboard pattern found in literature the jax-rs application is created procedural, via code, and not declarative. That's done to be able to change the root path of the application. And I don't know exactly if the components must be started for this to work.

bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/ModuleTypeResource.java
davidgraeff: Yes I guess so. Eclipse just went nuts on me and didn't resolve `JaxrsName` when I made those changes.

bundles/org.openhab.core.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/item/EnrichedItemDTOMapper.java
maggu2810: Why will this PR be broken if you rebase to the master?

davidgraeff: Again: I have rebased. Just for this one file I had conflicts that I didn't feel comfortable to hackish resolve them, so I just kept my changes, but annotated with a comment to remind me to cheery pick the changes back.

kaikreuzer:
What's the current state here? I didn't look into the details, but it feels to be a risky change, while if all works out, it is a worthwhile cleanup of the code base.

Background of my question is https://github.com/openhab/openhab-core/issues/597: This seems to be a rather critical issue by now as it occurs almost on every initial setup. If this PR here might soon be ready, we could check if it solves this problem on the way - if not, we will probably have to invest time to solve it in the current code base.

davidgraeff:
It will probably fix the issue. The Jax-rs implementation independent SSE parts are used.
But I'm also stuck because of SSE. It was used in the sitemaps part in a way that it was not intended to be used and the API does not allow the same pattern.

davidgraeff:
I will not continue to work on this PR. So it might be closed or taken over :)

maggu2810:
@davidgraeff That's really sad.
I don't use the sitemap, too.
I assume there will be no volunteer -- at least for me it seems people are more interested to not change anything instead of moving to a more stable and standardized technology.
The R7 JAX-RS whiteboard pattern would allow us to drop some old outdated unmaintained third party code but without dropping or rewriting the sitemamp stuff we cannot move forward.

davidgraeff:
> (At least) for me it seems people are more interested to not change anything instead of moving to a more stable and standardized technology.

Exactly. Unfortunately that makes OH unattractive to me. I picked OH because I had hoped to benefit from synergistic effects. Instead I wrote or rewrote all required addons myself (MQTT, Network Binding, Yamaha Binding, Hue Emulation, Deconz, Hue). That's not an issue itself, the community benefits, was my thought.

But this phobic clinging to old concepts here in core is what turns me away tbh. I see huge problems with the current REST interface (https://github.com/openhab/openhab-webui/issues/24#issuecomment-493075190 for details) and updating the software stack is just the tip of the iceberg. And I don't see that changing, now that Yannick has written a new UI that again uses Sitemaps (sigh...).

maggu2810:
I assume I don't share all your motivation or thinking about a fast moving target as I also need to support commercial products.
But I assume we should think about such topics in general.

Using specifications and simplifications of R7 makes IMHO really sense.
WRT your PR I would like to be interested which developer would like to work on the sitemap codebase or would like to keep them up to date.

If there is no one who would like to work on that code and there are just consumers that state we need to keep such code working, I don't know how to move forward...

@davidgraeff Can you tell me about the work that is done on openHAB 3? Who will contribute? Who will move openHAB forward (not organization but the code base)?

kaikreuzer:
> (At least) for me it seems people are more interested to not change anything instead of moving to a more stable and standardized technology.

Not sure if you refer to me? If so, you are wrong here. I was very happy to see this PR and hoped we can get rid off a lot of code and dependencies here. So I share the disappointment that @davidgraeff gives up on it.

> I just didn't came to test all the finicky SSE parts with the Sitemap bundle, where you register a client first etc. That's so much antipattern, so not really supported by how Jax-RS SSE works

If this is the part that led you to give up on it, could you at least provide some details on how far you got here, what obstacles there were and what exactly you consider an antipattern that cannot be resolved without removing complete features? Is SSE working for item streams and the problem is only with SSE for sitemaps? Could you exclude the sitemap SSE support from this PR and for the moment keep the Glassfish dependency for it and I can look into it separately?

davidgraeff:
> what exactly you consider an antipattern

I haven't read the SSE specification, but the subscribe/unsubscribe pattern that is used by Sitemap has no matching API and I haven't seen it anywhere else. I guess the idiomatic way would be to open / close connections on unique http endpoints if its about unique event streams. Right now some kind of manual multiplexing and session tracking is performed.

> exclude the sitemap SSE support from this PR

I can't recall it exactly, but I think that was my first attempt actually. Overlapping APIs and library responsibilities and therefore build and runtime failures made me start to touch that part as well.

maggu2810:
@davidgraeff While my modifications of the eclipsesource JAX-RS publisher I realized (again) that it uses a custom service tracker to register resources (also for message body reader / writer, our REST resources, ...).
I changed some of that service trackers to DS references to provide a running IDE for the community and this can be done easily instead of this whole migration...
If you have a look at e.g. https://github.com/maggu2810/openhab-core/commit/5b10f8d7ed3d30431b0a2ccac2b74ee0c98c600c#diff-d2b15dd840e0a539deed5f80784684eb
we could also change the remaining tracker and the DS references to specific services or e.g. hardcode the specific message body writer and reader impl. Would this be an option to use the old for non migrated stuff and the JAX-RS Whiteboard pattern for the migrated ones?
If we control the publisher and so the registration of all things for the Jersey connector, can't we migrate in smaller steps and use both parallel?

davidgraeff:
> Would this be an option

Yes of course. If you get the eclipsesource bundle to run with recent whiteboard pattern compatible jax-rs libraries. At the moment it is hard-wired to jersey .22 afaik. Should be doable.

This PR is does not only contain the whiteboard pattern migration (that would be a few hundred lines of annotation changes only). It is also modernizing, by using recent jax-rs features, cutting down a lot of code. This will also work with the old eclipsesource bundle, if jersey can be updated to a more recent version.

maggu2810:
@davidgraeff I played around with SSE today a little bit to migrate some projects to pure javax.ws.rs package space without depending on a specific implementation.
I used the Aries JAX-RS Whiteboard bundle at runtime.

As all the time a new `Sse` object (of the context) is injected it makes no sense to create a new broadcaster all the time (so for every peer). You could trace the `setSee` calls yourself.
For me it seems also if a broadcaster does not remove a event sink after registration if this one has been closed at any time e.g. because the peer terminates the connection.

A very simple working implementation can be found here https://github.com/maggu2810/playground-osgi/commit/0b05290aa5e2988d5fcbf9065c6550399409a9d2#diff-d969275f7520fc6d63d1ccc458191d88R1

There is one REST endpoint to register for listening and an broadcast interface that can be used by other components to broadcast events.

davidgraeff:
> migrate some projects to pure javax.ws.rs package space

:+1: 

> As all the time a new Sse object (of the context) is injected, it makes no sense to create a new broadcaster all the time (so for every peer)

I agree. Apparently I did not read the api specification / contract well enough.

It's a slightly different situation for Sitemaps (multiple SSE endpoints) and I think the original code tried to harmonize multiple use cases and I followed along with this implementation.

maggu2810:
For the documentation (Swagger) you could be perhaps interested in something like that:
https://github.com/maggu2810/playground-osgi/compare/bafb6e4a6d09000b8fd887020fd337c0a19b0be6...a11358ddbd93f221f3f847135366ab79dc5b5380

* It uses the wabjars swagger-ui (only the default URL is changed) that is registered using the HTTP Whiteboard spec.
* A short demo of documented REST API using the OpenAPI annotations (https://swagger.io/specification/)
* A simple oapi generator that currently tracks all JAX RS resources (using the filter `"(osgi.jaxrs.resource=true)"`)

boc-tothefuture:
I think @davidgraeff abandoned this due to sitemaps.   I can't tell... @maggu2810 are you picking it up to complete the sitemap side of it?

davidgraeff:
The question is if this can be done in smaller parts. There are quite some merge conflicts accumulating. I guess it needs to be merged in one go though.

maggu2810:
> @maggu2810 are you picking it up to complete the sitemap side of it?

No, I don't.

maggu2810:
I started to migrate my fork yesterday myself (without all the DTO cleanups -- just to get it running).

* https://github.com/openhab/openhab-core/compare/master...maggu2810:jax-rs-whiteboard?expand=1
* https://github.com/openhab/openhab-webui/compare/master...maggu2810:jax-rs-whiteboard?expand=1
* https://github.com/openhab/openhab-distro/compare/master...maggu2810:jax-rs-whiteboard?expand=1

Issue #645, opened 2019-03-11
REST: Requesting item namespaces


davidgraeff:
I have the usecase of listing all metadata namespaces for an item or items.

That should work with a request similar to `http://192.168.1.8/rest/items?metadata=.*` with a regex pattern of ".*".
But I only get the metadata namespaces, that have config-descriptions assigned (second part of the code references below), but the first part does not seem to let me retrieve all namespaces with the given regex.

Faulty code suspected in:

https://github.com/openhab/openhab-core/blob/48d873a32ba0901a9b58f04c541f0f858fbfcd51/bundles/org.openhab.core.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/item/MetadataSelectorMatcher.java#L61

### How to reproduce
* Have an item (in this case with item name "6b18869a8b86bba")
* Add an arbitrary metadata to it: `curl -X PUT --header "Content-Type: application/json" --header "Accept: application/json" -d "{
  \"value\": \"testv3\"
}" "http://192.168.1.8/rest/items/6b18869a8b86bba/metadata/test"` (here we add "testv3" to "test")
* Request `http://192.168.1.8/rest/items?metadata=.*`: You do not see the "test" namespace.

PR #744, opened 2019-04-18
[WIP] Migration of expire binding to profiles

clinique:
Hello, following discussion initiated in [previous PR,](https://github.com/openhab/openhab-core/pull/732#issuecomment-484453886) here is a first step of work on the Expire transposition to profile.
I'd like your thoughts, and especially regarding the naïve approach I found to transform state/command string to State / Command. Maybe a better way exists, but I found no link with the channel or item.

Also if someone can give me some pointers on the way to proceed with the `Registry` ... would be welcome :)

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/ExpireProfile.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/ExpireProfile.java
@@ -1,0 +1,233 @@
+/**
+ * Copyright (c) 2010-2019 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.core.thing.internal.profiles;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.smarthome.core.library.types.DateTimeType;
+import org.eclipse.smarthome.core.library.types.DecimalType;
+import org.eclipse.smarthome.core.library.types.HSBType;
+import org.eclipse.smarthome.core.library.types.IncreaseDecreaseType;
+import org.eclipse.smarthome.core.library.types.OnOffType;
+import org.eclipse.smarthome.core.library.types.OpenClosedType;
+import org.eclipse.smarthome.core.library.types.PlayPauseType;
+import org.eclipse.smarthome.core.library.types.PointType;
+import org.eclipse.smarthome.core.library.types.QuantityType;
+import org.eclipse.smarthome.core.library.types.RawType;
+import org.eclipse.smarthome.core.library.types.RewindFastforwardType;
+import org.eclipse.smarthome.core.library.types.StopMoveType;
+import org.eclipse.smarthome.core.library.types.StringListType;
+import org.eclipse.smarthome.core.library.types.StringType;
+import org.eclipse.smarthome.core.library.types.UpDownType;
+import org.eclipse.smarthome.core.thing.profiles.ProfileCallback;
+import org.eclipse.smarthome.core.thing.profiles.ProfileContext;
+import org.eclipse.smarthome.core.thing.profiles.ProfileTypeUID;
+import org.eclipse.smarthome.core.thing.profiles.StateProfile;
+import org.eclipse.smarthome.core.thing.profiles.SystemProfiles;
+import org.eclipse.smarthome.core.types.Command;
+import org.eclipse.smarthome.core.types.State;
+import org.eclipse.smarthome.core.types.Type;
+import org.eclipse.smarthome.core.types.TypeParser;
+import org.eclipse.smarthome.core.types.UnDefType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This is the default implementation for an expire profile.
+ *
+ * @author Gaël L'hopital - initial contribution and API.
+ * @alsoby Michael Wyraz -- initial v1 binding implementation
+ * @alsoby John Cocula -- added command support, tweaks
+ *
+ */
+@NonNullByDefault
+public class ExpireProfile implements StateProfile {
+    private static final String DURATION_PARAM = "duration";
+    private static final String STATE_PARAM = "state";
+    private static final String COMMAND_PARAM = "command";
+    private static final Pattern DURATION_PATTERN = Pattern
+            .compile("(?:([0-9]+)H)?\\s*(?:([0-9]+)M)?\\s*(?:([0-9]+)S)?", Pattern.CASE_INSENSITIVE);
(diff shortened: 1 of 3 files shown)


openhab-5iver:
Seems to me like you're addng automation functionality into profiles, and that this might be better as an ActionType. All of the OH1 and Core Actions need to be migrated, and then they can be utilized in the new rule engine, scripted automation, and the rules DSL. There's some discussion and prep work to be done for this, and I plan to get to it soon. Maybe this would interest you? 

IMO, an Expire Action sounds better than an Expire Profile.

kaikreuzer:
@openhab-5iver I agree that profiles have some overlap with rules. Actually, if you read what I had [written about profiles](https://www.openhab.org/blog/2018-12-17-openhab-2-4-release.html#profiles), you will see that I described them to be some implicit rules that the users do not have to create.
In this respect, I don't think an Expire action would be the right way to go - it would mean that users have to create many rules in order to set up this logic.

The one problem that I see with the profile approach is that it cannot be combined with other profiles. If you are e.g. using an Offset profile already, you cannot use the Expire profile at the same time. I am thus not convinced that this is a good approach. It might require a closer integration into the framework, similar to the autoupdate feature.

openhab-5iver:
I understand the overlap between profiles and automation, but timer functionality also needs to be implemented as an action. Something triggers, conditions are met, and then after a certain amount of time, something else happens (with the option of rescheduling). Having an action and profile doing the same thing will likely cause confusion, especially when a timer action is applied to an expire profile item...

kaikreuzer:
I consider the expire feature to be similar to the autoupdate feature: It automatically sends state updates for items. When and why this happens should not matter from the outside and thus it is imho not relevant that scheduled jobs are used internally - they must imho not be exposed, but only be treated as an internal implementation detail.

davidgraeff:
> they must imho not be exposed, but only be treated as an internal implementation detail.

@kaikreuzer I'm not sure if you have read my last comment in the linked PR. I do agree that implementation details must not be shown, but there must be away to enumerate current setup timers for maintenance reasons.

I have suggested to register those created timers to a `Registry` and expose them via rest. An alternative is item meta data which can already be requested by REST (if #645  is fixed at least).

kaikreuzer:
> there must be away to enumerate current setup timers for maintenance reasons.

Could you elaborate on this a bit? I do not see the need for a Registry, which would mean that we open them up to CRUD operations instead of treating them as an internal implementation detail.

davidgraeff:
It does not need to be a Registry, this is just what we have.

An overview of setup timers is just useful. Imagine a light turns on/off randomly and you want to find out why. (I'm all for a log-file-free smarthome usage, if that's not yet obvious.) So I would lookup that list of scheduled item changes. Another reason is resource usage. If your smarthome system is unresponsive around every midnight, I'd check my timed automation rules and that item-specific timer list.

openHAB should not hide anything when it comes to scheduled item changes.

kaikreuzer:
For rules, I agree that scheduled jobs should be manageable by the user. For anything that is done internally, I think it is dangerous to expose it. With the same logic, you could ask to expose all jobs in any scheduled executor that is used in the code.
For maintenance/performance analysis, this is a classical case for using the metrics API.

davidgraeff:
Actually yes I also like the idea of a list of what the scheduler bundle has scheduled. The thing with home-automation software is, you are the user AND the administrator. So it comes in handy to know that your system is scheduling about a ton of work and where that work originates from.

But back to the expire topic: Where do you see harm if an expire2 provides an enumeration interface? (Not CRUD, just read-only)

bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/ExpireProfile.java
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    @Override
+    public void onStateUpdateFromHandler(State state) {
+        applyExpire(state);
+    }
+
+    private void applyExpire(Type state) {
+        logger.trace("Received '{}'", state);
+        executor.shutdownNow();
```
maggu2810: Why do you shutdown the whole executor?
I assume you would like to remember the Future that is returned by `executor::schedule` and cancel that specific future.

kaikreuzer:
> Where do you see harm if an expire2 provides an enumeration interface? 

As mentioned above, to me such an information is a metric about the inner state of the software and thus the metric API would be the right thing to use. Yes, I am aware that we do not have an implementation that picks up and display those provided metrics, but it is something we should add - we would have other internal metrics like CPU, Java heap usage, no of threads, threadpool queue lengths, etc. at the same time.

davidgraeff:
Didn't know that metrics API was a thing. I was going to submit my solution this weekend. I need exactly that for my paper UI design study.

bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/ExpireProfile.java
clinique: Do you feel this is really needed, because at one time there is only 0 or 1 Future scheduled, and when receiving a state it will be either cancelled, either rescheduled. So I felt this was fine.

maggu2810: Perhaps I am wrong but my reading and understanding has been that there is one executor for all profiles.
See https://github.com/openhab/openhab-core/blob/48d873a/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ProfileContextImpl.java#L45

A executor per profile will result into a lot of executors and overhead so the current implementation (using one executor for all profiles) makes sense (but that does not care as it is already the case).

So, if one profile shutdown the executor this will affect all the other profiles and stop all the scheduled jobs of every profile.

kaikreuzer:
> Didn't know that metrics API was a thing.

The discussion happened quite a while ago at the [ESH forum](https://www.eclipse.org/forums/index.php?t=msg&th=1094339&goto=1793459&#msg_1793459) as well as in the [openHAB forum](https://community.openhab.org/t/new-add-on-bundle-for-prometheus-health-metrics/48094).

We had added the microprofile metrics api to ESH, but didn't yet make any use of it.
But @pokerazor offered to contribute an implementation for pushing the metrics to Prometheus - @pokerazor, maybe you want to comment on this?

davidgraeff:
I see. I'll have a look at Eclipse Microprofile Metrics then

pokerazor:
@kaikreuzer @davidgraeff yes, we have a working implementation of metrics compatible with Prometheus, as stated in the linked forum post. It is here: https://github.com/KuguHome/openhab-prometheus-metrics
It exposes lots of OH internal metrics (resource usage, scheduler jobs etc) for a detailed insight.
I would be quite happy to discuss how to introduce this into the core.

This is not using the Eclipse Microprofile Metrics though. I have no experience with that. Jochen Hiller of Telekom said they had some implementation using it and were willing to release it when the retraction happend...

This is leading a bit off the original issue topic here, though? :)

kaikreuzer:
> This is leading a bit off the original issue topic here, though? :)

You're right, I have created https://github.com/openhab/openhab-core/issues/774 as a follow-up!

Rossko57:
Comment from non-technical user:

expire-1 is frequently used with "virtual" Items i.e. no bindings (channels).  It's not clear to me how you could apply a profile to this.

A frequently noted limitation of expire-1 is that the user cannot change the time delay from rules.  But very often, that is a frustration because we're using a virtual item with expire-1 as a simple to manage general purpose timer (for example to block message flooding), and not the original "expire stale data" intent.

A less frequently noted limitation of expire-1 is that the user cannot test for the unexpired time value.  Again, that's really about (mis)use of expire for general timing purposes.

kaikreuzer:
I'll close this PR as I think it became clear that profiles aren't necessarily the best match for this functionality.
I have created https://github.com/openhab/openhab-core/issues/1620 with an alternative suggestion, which should hopefully also not mean too much effort to implement.

PR #757, opened 2019-04-26
[WIP] Add profile to control Dimmer Items via rawbutton channel

bodiroga:
Hi guys!

I have implemented a new profile to control Dimmer Items via a rawbutton trigger channel.

- "Short" press event (PRESSED and RELEASED in less than 550 milliseconds): send ON/OFF command.
- "Long" press event: after 550 milliseconds, start sending INCREASE/DECREASE commands every 200 milliseconds. If the dimmer is fully ON (100%), it will send DECREASE commands, otherwise, INCREASE. Perhaps I could make this a user configuration parameter, since I am not sure if it makes sense to start sending INCREASE commands if the dimmer is at 90%. The parameter would be a number, between 0 and 100, and it will establish the dimmer state above which the profile will start sending the increase commands (100 by default).

What do you think?

Many thanks and best regards,

Aitor

Signed-off-by: Aitor Iturrioz <riturrioz@gmail.com>

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawButtonDimmerProfile.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawButtonDimmerProfile.java
@@ -77,7 +77,7 @@ public class RawButtonDimmerProfile implements TriggerProfile {
                     IncreaseDecreaseType.INCREASE;
             buttonPressed(command);
         } else if (CommonTriggerEvents.RELEASED.equals(event)) {
-            OnOffType newState = previousState.equals(PercentType.HUNDRED) ? OnOffType.OFF : OnOffType.ON;
+            OnOffType newState = previousState.equals(PercentType.ZERO) ? OnOffType.ON : OnOffType.OFF;
             buttonReleased(newState);
         }
     }
--- a/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawButtonDimmerProfile.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawButtonDimmerProfile.java
@@ -1,0 +1,116 @@
+/**
+ * Copyright (c) 2010-2019 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.core.thing.internal.profiles;
+
+import static org.eclipse.smarthome.core.thing.profiles.SystemProfiles.RAWBUTTON_DIMMER;
+
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.smarthome.core.library.types.IncreaseDecreaseType;
+import org.eclipse.smarthome.core.library.types.OnOffType;
+import org.eclipse.smarthome.core.library.types.PercentType;
+import org.eclipse.smarthome.core.thing.CommonTriggerEvents;
+import org.eclipse.smarthome.core.thing.profiles.ProfileCallback;
+import org.eclipse.smarthome.core.thing.profiles.ProfileContext;
+import org.eclipse.smarthome.core.thing.profiles.ProfileTypeUID;
+import org.eclipse.smarthome.core.thing.profiles.TriggerProfile;
+import org.eclipse.smarthome.core.types.Command;
+import org.eclipse.smarthome.core.types.State;
+
+/**
+ * This profile allows a channel of the "system:rawbutton" type to be bound to a Dimmer item.
+ *
+ * If the dimmer is fully ON, the profile will send 'decrease' commands while the button is pressed. Otherwise,
+ * the profile will send 'increase' commands to the item. A short press will toggle the state of the item.
+ *
+ * @author Aitor Iturrioz - initial contribution.
+ *
+ */
+@NonNullByDefault
+public class RawButtonDimmerProfile implements TriggerProfile {
+
+    private final ProfileCallback callback;
+
+    private final ProfileContext context;
+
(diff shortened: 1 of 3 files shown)


Gozilla01:
Very interesting.
Could you have a button to recognize a long press?

maggu2810:
IIRC ON will be replaced by 100 and OFF will be replaced by 0 for a dimmer.

An idea about the state handling:

* previousState unknown => assume 0, previousDirUp = false

* short press, release
  * previousState == 0 => send on (previousState will be 100 after that), set previousDirUp = true
  * previousState != 0 => send off (previous State will be 0 after that), set previousDirUp = false

* long press
  * `previousDirUp == true` => start decrease future
  * `previousDirUp == false` => start increase future

* long press release
  * stop future, `previousDirUp = !previousDirUp`,

bodiroga:
> Could you have a button to recognize a long press?

Hi @Gozilla01! I don't understand what you mean, this profile is designed to work with triggers that only send PRESSED and RELEASED events.

> Please add the new profile types and profile type UIDs to the list of supported profile types / UIDs in the related factory and make sure they can be created by it.

Done!

> An idea about the state handling:

Yeah, I like it :+1: Although there are some weird situations. Imagine that you dim the lights to 40% and after a couple of hours, you turn them off with a single press. Next morning, you turn the lights on again with a single press, but the lights are restored to 40%. Trying to increase the brightness level, you long press the button, but as the previous event was an ON, the lights will start to decrease. Ouch!

Anyway, with a single button we can't implement a perfect solution, so I can adopt @maggu2810's suggestion!

Best regards,

Aitor

PS: I could also implement a double press action (for example, turn the lights on at 100%), but that should be another PR, right?

cweitkamp:
Thanks for the fast changes. Here are my thoughts.

I started a similar approach a while ago. A little bit more generic. See [here](addons/binding/org.openhab.binding.deconz/src/main/java/org/openhab/binding/deconz/internal/profiles/DeconzRepeatedCommandTriggerProfile.java) to get an idea what I mean. Nothing stable yet.

Currently I have the slight feeling that you want to put too much business logic into the profile. I am not sure if that is the way to go. From my point of view the framework should provide primitive tools for the users. Or highly individualizable. Thus I would suggest to give the user more options (e.g. choice of the repeated command for the triggered event, flexible time gaps, and so on). Like you proposed too.

One problem I saw with both implementations is an unpredictable state of the item or the command which will be send to the device. Imagine you have a channel linked to an item using this profile, the handler uses a polling strategy to retrieve data from a device and updates the state frequently. Once the profile gets activated its scheduled update job and the handler's polling job will begin to compete with each other which may result in flapping states. Do you know what I mean?

Do not get me wrong: I really like the idea behind it. Just want to discuss the options with you.

Gozilla01:
> Hi @Gozilla01! I don't understand what you mean, this profile is designed to work with triggers that only send PRESSED and RELEASED events.

Yes, the current button only handles pressure and release.
Reading your proposal, very interesting, because I can implement the Openwebnet binding with this function.
What I asked you if you could implement a new button with the ability to manage long press.
I need to send a different code to the bus at each event:
- Start pressure
- Short press
- Prolonged pressure every 500ms
- Pressure release.
 

**The bticino physical dimmers work like this:**
Short press, turns on the light at the previous value %
Long press, increase value in %
If we continue longer, the decreases value in % 
If we take a short pressure, turn off the light
The previous value is always the value in % never ON or OFF
It would be nice to have a parameter to manage the speed between 0 and 100%

bodiroga:
Hi @cweitkamp!

Thanks for the link to your repo, I can see that your profile is much more generic and gives users more configuration options, but sometimes I feel that we give openHAB users too many options and few easy-to-use things. My profile tries to mimic a manual push button-dimmer that I have at home, a purely electronic one, with the idea to replace it with a software approach using openHAB. Although in my physical dimmer, if you keep pressing the button when the lights have reached their 100% brightness, it starts to decrease the light level. It cycles all the time after reaching a extreme point.

> Once the profile gets activated its scheduled update job and the handler's polling job will begin to compete with each other which may result in flapping states.

What about the RawRockerDimmerProfile? (https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawRockerDimmerProfile.java). I have followed the same logic implemented in that profile. IMHO, the presented profile could fall in the "primitive tools" category, because that is how "dumb" push button-dimmers work and openHAB users expect them to be included in the system. But yeah, I understand that is not easy to build simple to use tools and highly configurable profiles 😢

Perhaps I should change the profile behavior to "increase" all the way up until 100% and then start to "decrease"? Do you know how other physical push button-dimmers work? I only know the one I have at home.

> What I asked you if you could implement a new button with the ability to manage long press.

Hi @Gozilla01! I think there is a little bit of misunderstanding here 😕 You, as a binding developer, should model one of your trigger channels using of the system trigger channels (rawbutton, button, rawrocker,... https://www.eclipse.org/smarthome/documentation/development/bindings/thing-definition.html#system-trigger-channel-types). You can create your own, sure, but the best option is to follow one of the defined ones, to ensure the reusability. Once you have defined your channel as a trigger channel, the user can use the "triggered" event information via rules (the complex approach) or using an already defined profile. Profiles are what the framework offers to encapsulate simple logic to the user, avoiding the need to create rules to achieve simple things. This PR tries to encapsulate the logic to control a Dimmer Item using a rawbutton channel (one that only emits PRESSED and RELEASED events).  How are you going to model your buttons in the binding?

Best regards,

Aitor

cweitkamp:
@bodiroga Any progress on this?

bodiroga:
> @bodiroga Any progress on this?

I'm so sorry for not coming back until now, excuse me :disappointed: 

Ok, I can implement what you suggested in your last comment, just one question. When you say "if OFF (== 0%) send ON" in the long pressed action, should we send a simple ON (in most cases this means restoring the previous ON value) or ensure that the light is powered to its 100% brightness?

@maggu2810 (and to everyone interested), any other suggestion?

Thanks!

maggu2810:
Hi @bodiroga,
great that you are back.
I assume using numeric (percent) values only should work and we don't need to use ON / OFF on the long press at all.

cweitkamp:
@bodiroga I am glad you are back. Lets stick to percent values like @maggu2810 suggested. So replace `ON` by 100% (`PercentType.HUNDRED`) and `OFF` by 0% (`PercentType.ZERO`).

cweitkamp:
I added the "work in progress" label.

cweitkamp:
@bodiroga So you have plans to finalize this PR? Otherwise I would think about closing it.

bodiroga:
Hi @cweitkamp!

I'm sorry, but I think that closing the PR would be the best decision, as my Eclipse IDE is now set up to for OH2 and I can't test my core changes :cry: I will try to came back to it when OH3 is more mature :wink:

PR #766, opened 2019-04-28
Allow to use Declarative Services 1.4

maggu2810:
Change compile dependency for compendium specification to R7.
This allows us to use the version 1.4 of the components and we can use the Constructor Injection.

### DIFF
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -29,7 +29,7 @@
     <dependency>
       <groupId>org.osgi</groupId>
       <artifactId>osgi.cmpn</artifactId>
-      <version>6.0.0</version>
+      <version>7.0.0</version>
     </dependency>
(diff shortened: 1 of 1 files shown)


maggu2810:
I compiled the whole openHAB Core and checked its usage in Karaf 4.2.4 and 4.2.5.
It works as Karaf is already using SCR that supports 1.4.

maggu2810:
The constructor injection is IMHO a great feature.

Mandatory references are present on object creation.
You can set the member variables in the constructor and so nullness checks work much better as you can remove all the nullness check workarounds (as the member will be non null on the whole lifecycle).
We don't also need to check anymore (if constructor injection is used) if a component is already active at the time an optional dependency is injected. The component is active on object creation and so if a member function is called it must be active.

Issue #772, opened 2019-04-30
CME: org.eclipse.xtext


maggu2810:
Some of that annoying race conditions that are only seen some time and cannot be reproduced:

```
2019-04-29T07:32:05,568 | ERROR | CM Event Dispatcher (Fire ConfigurationEvent: pid=org.eclipse.smarthome.folder) | CompositeEValidator              | 222 - org.eclipse.xtext - 2.17.0.v20190304-0545 | Error executing EValidator
java.util.ConcurrentModificationException: null
        at org.eclipse.emf.common.util.AbstractEList$EIterator.checkModCount(AbstractEList.java:758) ~[?:?]
        at org.eclipse.emf.common.util.AbstractEList$EIterator.doNext(AbstractEList.java:712) ~[?:?]
        at org.eclipse.emf.common.util.AbstractEList$EIterator.next(AbstractEList.java:692) ~[?:?]
        at java.lang.Iterable.forEach(Iterable.java:74) ~[?:?]
        at com.google.common.collect.Iterables$5.forEach(Iterables.java:621) ~[87:com.google.guava:21.0.0]
        at org.eclipse.xtext.xbase.validation.UniqueClassNameValidator.checkUniqueName(UniqueClassNameValidator.java:76) ~[?:?]
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:?]
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[?:?]
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:?]
        at java.lang.reflect.Method.invoke(Method.java:498) ~[?:?]
        at org.eclipse.xtext.validation.AbstractDeclarativeValidator$MethodWrapper.invoke(AbstractDeclarativeValidator.java:127) ~[222:org.eclipse.xtext:2.17.0.v20190304-0545]
        at org.eclipse.xtext.validation.AbstractDeclarativeValidator.internalValidate(AbstractDeclarativeValidator.java:318) ~[222:org.eclipse.xtext:2.17.0.v20190304-0545]
        at org.eclipse.xtext.validation.AbstractInjectableValidator.validate(AbstractInjectableValidator.java:71) ~[222:org.eclipse.xtext:2.17.0.v20190304-0545]
[...]

PR #773, opened 2019-04-30
model: prevent simultan access of singleton

maggu2810:
The singleton instance of "Diagnostician" is used without synchronization.
The singleton "Diagnostician" instance is using the singleton "EValidator.Registry" instance (without synchronization).
In very rare high load situations there has been CME detected.

My first "solution" has been to synchronize the access of Diagnostician's instance method by using

```java
synchronized (Diagnostician.INSTANCE) {
    ...
```

But after realize that EMF is using internally other singletons I tried to find any information about EMF and thread safety.

I found this one: https://javahacks.net/2016/07/13/emf-thread-safety/

  EMF models are not thread-safe by default and writing multithreaded
  applications is not that simple.
  The more complex our application became, the more often we got
  concurrent modification exceptions and had problems with filtering and
  sorting operations.

So, I assume instead of adding synchronizations to our code that is using EMF (IIRC this has been already done on the ESH hosted code base long time ago) we should try to execute EMF code in a safe manner.

This implementation adds a "SafeEMF" OSGi service that should be used to execute EMF code to ensure none code of EMF (or at least the ones that calls has been migrated to the SafeEMF usage) is accessed by separate threads at the same time.

Related to: https://github.com/openhab/openhab-core/issues/772

### DIFF
--- a/bundles/org.openhab.core.model.core/src/main/java/org/eclipse/smarthome/model/core/SafeEMF.java
+++ b/bundles/org.openhab.core.model.core/src/main/java/org/eclipse/smarthome/model/core/SafeEMF.java
@@ -1,0 +1,40 @@
+/**
+ * Copyright (c) 2010-2019 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.model.core;
+
+import java.util.function.Supplier;
+
+/**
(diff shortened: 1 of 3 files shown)


maggu2810:
Quote from https://wiki.eclipse.org/EMF/FAQ#Is_EMF_thread-safe.3F

Is EMF thread-safe?

EMF can safely be used in multi-threaded environments; however, EMF does not, itself, ensure thread-safety in application model implementations. Data structures are unsynchronized, for performance and to avoid potential deadlock situations. The expectation is that a complete instance of the model (including resources and resource set, if present) should only be accessed by one thread at a time, and that the synchronization should be provided by the application at a higher level.

Since it is shared among multiple instances of a given application model, Ecore metadata is intended to be thread-safe for read-only access. However, there was a thread safety bug involving Ecore annotations, which is now fixed but will still affect older releases of EMF. Also, the demand initialization of SDO metadata is not thread-safe. See the following two questions for workarounds to these issues.

Issue #636, opened 2019-03-07
[Automation] EventSubscriber is not registered on startup for rules created in Paper UI


openhab-5iver:
I'm seeing this in a test environment running S1549 (and S1534), but there was also at least [one report in the forum](https://community.openhab.org/t/rules-not-being-followed/68536), but that was on OH 2.4.0. After an OH restart, Paper UI rules can be triggered, and the log says they execute, but their actions don't run.
```
2019-03-05 12:37:54.031 [TRACE] [org.openhab.core.automation.internal.RuleEventFactory] - creating ruleEvent of type: RuleStatusInfoEvent
2019-03-05 12:37:54.031 [DEBUG] [org.openhab.core.automation.internal.RuleEngineImpl] - The rule '38e28c16-4594-418d-a5f5-2758d876b471' is executed.
2019-03-05 12:37:54.032 [TRACE] [org.openhab.core.automation.internal.RuleEventFactory] - creating ruleEvent of type: RuleStatusInfoEvent
```
By editing a rule and saving it, without making any changes, the rules will work again. I'm experiencing this every time I restart OH S1549, on a test machine with only one Item and no bindings. When the rules are saved, the EventSubscriber is registered, but this does not occur after an OH startup.
```
2019-03-07 07:04:37.875 [DEBUG] [org.openhab.core.automation] - ServiceEvent REGISTERED - {org.eclipse.smarthome.core.events.EventSubscriber}={service.id=421, service.bundleid=199, service.scope=singleton, event.topics=smarthome/items/Virtual_Switch_1/*} - org.openhab.core.automation
```

openhab-5iver:
I spent some time looking into this, and have some more details. When disabling a rule, it is added to automation_rules_disabled.json. When reenabling it, it is removed. I found that disabling all rules and restarting OH, the entries in automation_rules_disabled.json remained and the rules were not disabled, but all rules would worked properly!

This leads me to think that there is some issue in the serialization/deserialization of the JSON rule files. But since this behavior does not occur in earlier builds, it could be related to the ESH reintegration or use of maven/bnd.

When they work...
```
[...]

Issue #746, opened 2019-04-18
automation: unreliable execution after restart


maggu2810:
continues https://github.com/eclipse/smarthome/issues/6548#issue-382372469

maggu2810:
The commands to workaround the problem has been changed to:

```
scr:disable org.openhab.core.automation.internal.RuleRegistryImpl
scr:enable org.openhab.core.automation.internal.RuleRegistryImpl
```

davidgraeff:
Do you now the reason or have a suspect why it's unreliable?
I'm very much relying on the engine to work, but don't know how to reproduce this issue.

openhab-5iver:
This issue sounds a lot like #636, but I did not see it in OH builds before the ESH reintegration, and it is very consistent. @maggu2810, instead of restarting RuleRegistryImpl, do the rules work if you open them in Paper UI and save them?

maggu2810:
> but don't know how to reproduce this issue.

I assume you need a "slow" system.
I identify the problem on some RPis after updating to a new version (so, e.g. a first start of a completely new distribution just using the old userdata).

maggu2810:
> @maggu2810, instead of restarting RuleRegistryImpl, which I did not try when troubleshooting 636, do the rules work if you open them in Paper UI and save them?

I will give it a try next time.

maggu2810:
So, some further information.
As states in the other topic, the two rules are shown.

I added a new dummy rule that contains a trigger for an item change of item "DUMMY".
I set a breakpoint to inspect some internals.

After adding the new rule the rule engine implementation contains three rules.
There are three trigger instances but if I use the Karaf commands to list the available services provided by "org.openhab.core.automation" there is only one "EventSubscriber".

```
event.topics = smarthome/items/DUMMY/*
objectClass = [org.eclipse.smarthome.core.events.EventSubscriber]
service.bundleid = 262
service.id = 573
service.scope = singleton
```

To check it again, I filter by "event.topics":

```
karaf@root()> bundle:services -p org.openhab.core.automation | grep event.topics
event.topics = smarthome/items/DUMMY/*
```

or

```
karaf@root()> service:list org.eclipse.smarthome.core.events.EventSubscriber | grep event.topics
 event.topics = smarthome/items/DUMMY/*
```

So, it seems similar to #636

PR #776, opened 2019-04-30
automation: some cleanup

maggu2810:
Cleanup the rule engine implementation by using constructor injection. This allows us to use more final members. Also make some code "nicer looking"

If a rule is added, the status is set dependent on the "disabled storage" (see `addRule) to initializing or uninitialized-disabled.
The old code calls "set enabled" after adding all rules.
This is not necessary at all.

After removing the code my rules are working after an update, so perhaps this is also the root cause for #636 and #746. Sure, this could also just be timing dependent -- we will see.


davidgraeff:
Thanks, looks like a nice clean up and makes the enabled status more consistent indeed.

openhab-5iver:
I had planned on testing this _before_ merge... but it looks like this does resolve #636. Nice work!

PR #779, opened 2019-05-01
Initial contribution of 'Round' profile

cweitkamp:
- Initial contribution of 'Round' profile

In general bindings are meant to deliver precise data. If you like to have only one decimal for "visual purpose" you can archive it easily by applying a pattern / format on the state of your item linked to a channel. If a target device or entity requires a fixed decimal number the binding for it should take care of that on its own. We denied fixes in bindings for this purpose (see https://github.com/openhab/openhab2-addons/pull/4108).

But it often occurs that people are asking for rounded values e.g. to expose them via `say()` commands (see https://github.com/openhab/openhab2-addons/issues/4054). This profile can be used - if necessary - to round the states to a specific scale without using rules and introducing dummy items. It additionally provides the feature to choose a rounding mode.

Wdyt?

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemOffsetProfile.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemOffsetProfile.java
@@ -12,6 +12,8 @@
  */
 package org.eclipse.smarthome.core.thing.internal.profiles;
 
+import static org.eclipse.smarthome.core.thing.profiles.SystemProfiles.OFFSET;
+
 import java.math.BigDecimal;
 
 import javax.measure.UnconvertibleException;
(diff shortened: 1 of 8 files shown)


kaikreuzer:
Hm, I think we should be careful with the introduction of new profiles of this kind - I have the fear that the offset profile was a bad idea in the beginning and this now becomes an inspiration for further profiles of a similar kind.

In the past, we used transformations for such use cases - you had e.g. the chance to transform an input into an output for certain label in a sitemap. The "raw" data of the item was not touched. Note that with UoM-Items, rounding might not even make much sense, if the value is requested in a different unit than the one it was rounded for.

The major problem that I am seeing with having many such profiles is that you cannot combine them: What if you want a rounded offset value? You have to decide for the one or the other profile, but cannot use both.
We discussed that issue already when introducing profiles. I remember two potential solutions to that problem:
1. We use a Transform-Profile, which simply redirects the processing to transformations and such a profile could simply have a list of transformation strings as its configuration, so that you can build chains.
2. Offer a Script-Profile in which you can do complex calculations and thus easily do rounding, offsetting etc.

My take is, that we are probably more flexible through those options and don't have to implement many many new profiles for special situations.

davidgraeff:
When I wrote the documentation for Profiles, I had problems to find arguments why you would need Transformations in the first place.

* Profiles operate on real data types and don't do inefficient String->String mappings which makes me think that profiles offer the more advanced API.
* Some people don't use sitemaps at all. And if you don't, you find yourself using profiles only.
* Items are openHABs "variables". Considering this, for me it makes sense that if I want to have a rounded and non-rounded value, as an example, that I bind a channel to two items.

In the MQTT binding I offer people to use chained transformations for incoming and outgoing data. Why don't we allow chained profile execution?

PR #784, opened 2019-05-02
Update paho to 1.2.1 service release

davidgraeff:
Fixes unreleased resources, disconnecting problems and many more

### DIFF
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -207,7 +207,7 @@
     <dependency>
       <groupId>org.eclipse.paho</groupId>
       <artifactId>org.eclipse.paho.client.mqttv3</artifactId>
-      <version>1.2.0</version>
+      <version>1.2.1</version>
       <scope>compile</scope>
     </dependency>
 
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -207,7 +207,7 @@
     <dependency>
       <groupId>org.eclipse.paho</groupId>
       <artifactId>org.eclipse.paho.client.mqttv3</artifactId>
-      <version>1.2.1</version>
+      <version>1.2.0</version>
       <scope>compile</scope>
     </dependency>
 
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -438,7 +438,7 @@
     <dependency>
       <groupId>org.eclipse.paho</groupId>
       <artifactId>org.eclipse.paho.client.mqttv3</artifactId>
-      <version>1.2.0</version>
+      <version>1.2.1</version>
       <scope>compile</scope>
     </dependency>
 
--- a/features/karaf/openhab-tp/src/main/feature/feature.xml
+++ b/features/karaf/openhab-tp/src/main/feature/feature.xml
@@ -139,9 +139,9 @@
   </feature>
 
   <feature name="openhab.tp-paho" description="MQTT v3 Client" version="${project.version}">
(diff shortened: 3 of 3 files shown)


maggu2810:
You have missed https://github.com/openhab/openhab-core/blob/master/features/karaf/openhab-tp/src/main/feature/feature.xml#L141-L145

maggu2810:
@davidgraeff Do you consider to finish this PR?

davidgraeff:
done

maggu2810:
You changed the TP capability only, but not the used bundle:

`mvn:org.eclipse.paho/org.eclipse.paho.client.mqttv3/1.2.0`

maggu2810:
Have you changed your commits?
I have been sure that in your initial commit you changed the used paho version in the runtime BOM file (`bom/runtime/pom.xml`).
Now it is changed in the compile BOM file (`bom/compile/pom.xml`).

The compile can stay as it is as the API has not been changed and so there is no hard requirement to update (not changing the min version package version range part).
But to use the more recent version at runtime, we need to change the runtime BOM and the feature.

davidgraeff:
I'll change that later.
Is it intended that you need to change 4 different places for a dependency to update?

I know that we are using 2 different and independent mechanisms here which is maven as our buildsystem and karaf as the runtime container that need to know the version in form of karaf features. But why do I still need to change 4 instead of 2 places. Are there only properties in maven to centralize dependency version management?

maggu2810:
I assume we don't need to discuss about the difference between runtime and compile time dependencies.
So, I assume your "2 places" will be compile-version and runtime-version.

There is one file that contains the compile time (and so min. version): the compile `bom/compile/pom.xml`

There are two files for the runtime version:  the one used for bnd based tooling for the integration tests and for the demo applications + the one used for Karaf at runtime.

I miss the fourth place you are talking about. Can you point me to?

davidgraeff:
The 4th place in this case was the capability in the karaf feature. 

> There are two files for the runtime version: the one used for bnd based tooling for the integration tests and for the demo applications + the one used for Karaf at runtime.

That's what I think sounds unnecessary. We generate for bnd but not the karaf feature repository file.

davidgraeff:
done, this time for real

bom/compile/pom.xml
```diff
@@ -207,7 +207,7 @@
     <dependency>
       <groupId>org.eclipse.paho</groupId>
       <artifactId>org.eclipse.paho.client.mqttv3</artifactId>
-      <version>1.2.0</version>
+      <version>1.2.1</version>
```
maggu2810: Why do you change the minimum version we are able to work with?
Is there any point why products must update their paho version if they want to bump the mqtt transport?
If semver is used, we can keep the `[1.2.0,2)` and there is no need to force `[1.2.1,2)` for our imported paho packages.

davidgraeff:
The planned paho release 1.3 wasn't released for years now and there accumulated a high number of bug fixes and resource leaking. That's the least we can do to protect downstream from using the old version.

maggu2810:
We will not stop working with specific Java version just because there are known bugs.

Okay, that is a very "dramatic" example, but I never heard that OSGi bundles choose the package versions this way to prevent to work with "buggy" packages.

If we go that route, we will also change the minimum requirement for Jetty and a lot of other dependencies. This will make the framework unusable for any other product if they need to change its target platform all the time.

davidgraeff:
Paho is just a really bad example. The project looked even abandoned in between. I'd be better if we define 1.2 and 1.2.1 as minimum version if that's possible with maven.

PR #785, opened 2019-05-02
Added i18n feature for profiles

cweitkamp:
- Added i18n feature for profiles

One comment: There is an internal class `LocalizedProfileTypeKey` which is used in a very similar way for other translatable objects (e.g. https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/DefaultSystemChannelTypeProvider.java#L281-L328 and https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.config.xml/src/main/java/org/eclipse/smarthome/config/xml/AbstractXmlBasedProvider.java#L40-L79). I am not very happy with that and think about a refactoring. Let me know your opinion.

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/i18n/ThingTypeI18nUtil.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/i18n/ThingTypeI18nUtil.java
@@ -19,7 +19,6 @@ import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.smarthome.core.i18n.I18nUtil;
 import org.eclipse.smarthome.core.i18n.TranslationProvider;
 import org.eclipse.smarthome.core.thing.ThingTypeUID;
-import org.eclipse.smarthome.core.thing.profiles.ProfileTypeUID;
 import org.eclipse.smarthome.core.thing.type.ChannelDefinition;
 import org.eclipse.smarthome.core.thing.type.ChannelGroupDefinition;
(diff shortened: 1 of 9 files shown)


maggu2810:
What about a `LocalizedKey<K,V>`?

* `SystemProfileFactory.java.LocalizedProfileTypeKey` can be replaced by `LocalizedKey<UID, @Nullable String>`
* `DefaultSystemChannelTypeProvider.LocalizedChannelTypeKey` can be replaced by `LocalizedKey<UID, @Nullable String>
* `AbstractXmlBasedProvider.LocalizedKey` can be replaced by `LocalizedKey<Object, @Nullable String>`

Okay, while reading the classes, I assume we don't need generics at all.

`LocalizedKey` can use two members:
* `final Object key`
* `final @Nullable String locale`

The type of the key does not matter.

cweitkamp:
Yes, sounds good. Moving it to an own class in the i18n package.

Another question which came to my mind was if it is the right place to use the localized caches in the `XXXTypeProvider`s. Or should we maybe move them over to the `XXXTypeI18nLocalizationService`s components which makes imho much more sense because the services could be used in multiple places and the benefits may be even bigger.

Independently of the final approach we should do it in a separate PR to keep the history clean.

maggu2810:
> Independently of the final approach we should do it in a separate PR to keep the history clean.

:+1: 

IMHO also the `LocalizedKey` can be done in another PR as it is not about i18n feature in profiles itself.
Will you volunteer for that improvement? :wink:

cweitkamp:
> Will you volunteer for that improvement?

Of course. I already have it in the pipe.

PR #787, opened 2019-05-02
use mandatory storage services

maggu2810:


maggu2810:
Yes, I did :wink: 

I realized I did a breaking change by changing the constructor of the `DefaultAbstractManagedProvider`.
I will come up with a fix for the addons and Web UI repo.

Sorry about that.

PR #790, opened 2019-05-04
POM Code Convention

maggu2810:
There is a recommended ordering for all Maven POM files.
See: https://maven.apache.org/developers/conventions/code.html

The POM files has been "fixed" by using the "sortpom-maven-plugin".
The blank lines has been kept to keep the element separation for
readability.
The plugin also fixes indentation etc.
Have a look at: https://github.com/Ekryd/sortpom/wiki

The profile has been set to "recommended_2008_06" that states:
The POM Code Convention that was chosen by Maven developers in 2008

Command that has been executed:

    mvn \
      com.github.ekryd.sortpom:sortpom-maven-plugin:sort \
      -Dsort.keepBlankLines=true \
      -Dsort.predefinedSortOrder=recommended_2008_06

Signed-off-by: Markus Rathgeb <maggu2810@gmail.com>


maggu2810:
See also e.g. https://github.com/eclipse/smarthome/pull/6057

maggu2810:
IMHO this command should be used similar to the one to trigger the maven license plugin to generate / update the license headers.

It allows also to don't rely on specific IDE settings for the POM format.

PR #791, opened 2019-05-04
drop remaining tycho and p2 stuff

maggu2810:
* remove tycho based POM used as parent by external repos
* remove "mega" POM used as parent by external repos
* remove p2 repository generation

---

I don't know about repository that is using one of the parent POMs as soon as openhab/openhab-distro#906 has been merged (merge it in front of this PR).

---

I don't know about anyone using the p2 repository anymore.

---

After this PR has been merged, we could move the Karaf feature definitions to its parent directory (so, drop the explicit subdirectory "karaf" of "features").


maggu2810:
AFAIK after this one and as soon as `/launch/openhab.target` if removed from `openhab-distro`, the `openhab-deps-repo` is not necessary anymore.

wborn:
The changes look good! Can you resolve the merge conflicts caused by the POM sorting PR?

openhab1-addons is still using some [ancient](https://github.com/openhab/openhab1-addons/blob/master/pom.xml#L7) pom-tycho version (2.1.0 :open_mouth: and ohdr version) so it'll probably have no compilation/release issues. We can always create a solution for that if updates are needed in the future.

maggu2810:
Merge conflicts solved.

PR #793, opened 2019-05-04
add printf for console

maggu2810:


maggu2810:
Can we get that in or do you see any blocker here?

maggu2810:
> some JavaDoc on the methods would be nice.

You are absolutely right!

PR #795, opened 2019-05-06
Travis CI: check POM convention

maggu2810:
We could use Travis CI to execute some checks in front of the normal build.

This change adds a check if the POM files follow our POM conventions.

If something fails, it show which files violates the convention and the command that should be executed to fix the situation.

For example if the base POM file breaks the convention:

    At least one POM file breaks the convention, please use sortpom to fix the POM file(s).
    	modified:   pom.xml
    You should run the following command in the root directory of your working copy:
        mvn com.github.ekryd.sortpom:sortpom-maven-plugin:sort -Dsort.keepBlankLines=true -Dsort.createBackupFile=false -Dsort.predefinedSortOrder=recommended_2008_06

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,24 +39,9 @@ install:
         prevent_timeout $pid &
         wait $pid
     }
-    function check_pom_convention() {
-        mvnp com.github.ekryd.sortpom:sortpom-maven-plugin:sort -Dsort.keepBlankLines=true -Dsort.createBackupFile=false -Dsort.predefinedSortOrder=recommended_2008_06
-        if [ $? != 0 ]; then
-            echo "sortpom failed"
-            exit 1
-        fi
-        if [ $(git diff | wc -l) != 0 ]; then
-            echo "At least one POM file breaks the convention, please use sortpom to fix the POM file(s)."
-            git status | grep 'modified:'
-            echo "You should run the following command in the root directory of your working copy:"
-            echo "    $(grep 'sortpom-maven-plugin:sort' .travis.yml | grep -v grep | sed 's: *mvnp:mvn:g')"
-            exit 1
-        fi
-    }
 after_success:
   - print_reactor_summary .build.log
 after_failure:
   - tail -n 2000 .build.log
 script:
-  - check_pom_convention
   - mvnp clean verify -B -DskipChecks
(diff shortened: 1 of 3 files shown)


J-N-K:
Not speaking against that change, but shouldn't this be covered by SAT so that wrong POMs don't get pushed in the first place?

maggu2810:
Sure, that would be the better way.

Is there a generic SAT logic that can be configured to execute specified Maven goals and fail if files has been changed?

wborn:
SAT [executes the MOJOs](https://github.com/openhab/static-code-analysis/blob/8cc0af5e53a0d53f82837b2cdc165e7261bb4892/sat-plugin/src/main/java/org/openhab/tools/analysis/tools/AbstractChecker.java#L110-L111) defined by the [checkers](https://github.com/openhab/static-code-analysis/tree/master/sat-plugin/src/main/java/org/openhab/tools/analysis/tools).

So an additional sortpom checker could be added so SAT can generate issues based on the [violations file](https://github.com/Ekryd/sortpom/wiki/Parameters#for-the-verify-goal-only) generated by the sortpom verify goal.

maggu2810:
I will add an issue to that SAT repository about the inclusion of this check.

WDYT about adding that check to Travis CI at the moment and remove it from Travis CI again as soon as it is part of SAT?
Or do you see any advantage to not use that check at all?

wborn:
I think it's a good idea to prevent non-sorted POMs from entering the code base.

Did you also consider adding a the [verify](https://github.com/Ekryd/sortpom/wiki/Parameters#for-the-verify-goal-only) plugin execution goal when building the code? It only takes a couple of seconds to execute: 

`mvn com.github.ekryd.sortpom:sortpom-maven-plugin:verify -Dsort.keepBlankLines=true -Dsort.predefinedSortOrder=recommended_2008_06 -Dsort.verifyFail="Stop"`

so it wouldn't add much to the build time.

That would solve the issue that contributors only run into this when it is executed by the build server. The only issue with that would be that contributors would not know how to easily sort the POMs when they see the error. :neutral_face: 

`
[ERROR] Failed to execute goal com.github.ekryd.sortpom:sortpom-maven-plugin:2.10.0:verify (default-cli) on project org.openhab.core.reactor: The file /home/wouter/git/openhab/openhab-core/pom.xml is not sorted -> [Help 1]
`

maggu2810:
@wborn Thanks for the hint (verify goal). I updated the branch.

PR #798, opened 2019-05-06
Adding Milliseconds to Smarthome Units

clinique:
Adding test

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/library/unit/SmartHomeUnits.java
+++ b/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/library/unit/SmartHomeUnits.java
@@ -152,6 +152,8 @@ public final class SmartHomeUnits extends CustomUnits {
             new RationalConverter(BigInteger.valueOf(1852), BigInteger.valueOf(1000))));
     public static final Unit<SolidAngle> STERADIAN = addUnit(Units.STERADIAN);
     public static final Unit<Temperature> KELVIN = addUnit(Units.KELVIN);
+    public static final Unit<Time> MILLISECOND = addUnit(
+            new TransformedUnit<>("ms", Units.SECOND, Units.SECOND, new RationalConverter(1l, 1000l)));
     public static final Unit<Time> SECOND = addUnit(Units.SECOND);
     public static final Unit<Time> MINUTE = addUnit(Units.MINUTE);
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/library/unit/SmartHomeUnits.java
```diff
@@ -152,6 +152,8 @@
             new RationalConverter(BigInteger.valueOf(1852), BigInteger.valueOf(1000))));
     public static final Unit<SolidAngle> STERADIAN = addUnit(Units.STERADIAN);
     public static final Unit<Temperature> KELVIN = addUnit(Units.KELVIN);
+    public static final Unit<Time> MILLISECOND = addUnit(
```
Hilbrand:
```suggestion
    public static final Unit<Time> MILLISECOND = addUnit(MetricPrefix.MILLI(Units.SECOND));
```

itests/org.openhab.core.tests/src/main/java/org/eclipse/smarthome/core/library/types/QuantityTypeTest.java
```diff
@@ -291,6 +292,12 @@ public void testMWh() {
         assertEquals("1000000 Wh", energy.toUnit("Wh").toString());
     }
 
+    @Test
+    public void testMilliseonds() {
```
Hilbrand: This test doesn't test your changes. Because the conversion is already part of the automatic conversion even without your newly added const this test will succeed. What you want to test if your new constant works. Something like this:
```
    @Test
    public void testMilliSeconds() {
        Quantity<Time> oneSecond = Quantities.getQuantity(new BigDecimal("1"), Units.SECOND);
        Quantity<Time> millisecond = oneSecond.to(SmartHomeUnits.MILLISECOND);
        assertThat(millisecond.getUnit().toString(), is("ms"));
        assertThat(millisecond.getValue().doubleValue(), is(closeTo(1000, DEFAULT_ERROR)));
    }
```
This test maybe is better added in `SmartHomeUnitsTest`

cweitkamp:
I am not sure if we really need it. Doesn't `MILLI(SECOND)` do the job at all?

clinique:
@cweitkamp : You're right ! Forget this.

Issue #532, opened 2019-02-05
resource leak: audio manager


maggu2810:
If my understanding of the audio managers `playFile` implementation is correct then there is a resource leak.
At least for me it seems that the input stream if not closed reliable (if at all).

kaikreuzer:
It depends whether we consider it critical ;-)
I'd define critical to be:
- makes the system dysfunctional for a large number of users under normal operation
- vulnerabilities

I am personally using `playFile` on a daily basis and haven't seen any negative impact (system running for months without an issue). So I'd not really consider it to be critical unless I miss something.

maggu2810:
For me it is different.
For me resource leaks are critical because one component can eat all resources that are not available for other components anymore.
Just because nobody runs into it doesn't make it less critical.

If a productive system fails because it runs out of resources, it could be hard to find the root cause (network traffic, files, ...).

In the special case the input stream of an `FileAudioStream` is never closed.
Perhaps also my reading of the sources has been wrong and the problem itself does not exist.

On my working machine it fails on the "1023268" try to open an file input stream without closing it.
If we know that the usage of `FileAudioStream` is the only resource leak, perhaps you will not hit that number ever...

... but openHAB Core as a framework should not matter about its usage and do its best to be flexible and "safe" (what a magic word) to use.

So, okay if you don't want to mark that one as critical -- for me it stays on the list of critical issues (as soon as I start to use the audio support). :wink:

cweitkamp:
I think it is not handled correctly anymore. But currently we do not face a resource leak because of two things.

- Some of the `AudioSink` implementations in bindings use `IOUtils.closeQuietly()` in the `process()` method to handle the stream (see for example [here](https://github.com/openhab/openhab2-addons/search?utf8=✓&q=AudioSink+IOUtils&type=)) or close them on their own. That is obviously not what a framework expect the outside world to do.

- The `AudioServlet` implementation closes the streams via `IOUtils` once they will be expired (see https://github.com/openhab/openhab-core/pull/801/files#r281794201).

Latest [Apache Commons IO 2.6 API](https://commons.apache.org/proper/commons-io/javadocs/api-release/deprecated-list.html#method) removed the `IOUtlis.closeQuietly()` methods completely in favor of the `try-with-resources` statement. From my point of view we should follow their path and get rid of it too.

Do you think we should /could update Apache Commons IO dependency too? We include version 2.2.0. What are the plans for it?

maggu2810:
[...]

PR #801, opened 2019-05-06
audio servlet: don't change map while used by interation

maggu2810:
We should not remove entries (so keys) from the map while we are using the key set (a view) of that map (see JavaDoc: otherwise "results of the iteration undefined").

### DIFF
--- a/bundles/org.openhab.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioServlet.java
+++ b/bundles/org.openhab.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioServlet.java
@@ -15,7 +15,10 @@ package org.eclipse.smarthome.core.audio.internal;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.UUID;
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioServlet.java
```diff
@@ -143,16 +146,20 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
+        for (Entry<String, Long> entry : streamTimeouts.entrySet()) {
+            if (entry.getValue() < System.nanoTime()) {
+                toRemove.add(entry.getKey());
             }
         }
+        toRemove.forEach(streamId -> {
+            // the stream has expired, we need to remove it!
+            final FixedLengthAudioStream stream = multiTimeStreams.remove(streamId);
+            streamTimeouts.remove(streamId);
+            IOUtils.closeQuietly(stream);
```
cweitkamp: @maggu2810 I think in this line you can find a piece of the puzzle to answer #532.

maggu2810: Not really. While reading this implementation there are much more things that confused me. I answered to #532

PR #810, opened 2019-05-09
model: don't require to download antlr generator on build time

maggu2810:
The antlr generator has been downloaded by a specific download maven plugin at build time.
Instead of calling the plugin goals it would be much simpler to add the downloaded file to the repository (I assume it has been not possible on the Eclipse repository before).
Another benefit would be the IDE integration as the IDE does not need to know if the plugin needs to be executed or not...

The provided antlr generator has been taken from: http://download.itemis.com/antlr-generator-3.2.0-patch.jar


maggu2810:
The whole model build is more complicated then it should be...

Before the `xtend-maven-plugin` should do its job, the `exec-maven-plugin` is used to call the `org.eclipse.emf.mwe2.launch.runtime.Mwe2Launcher` with a mwe2 "generated foo" file as argument (e.g. for the items `file://${project.basedir}/src/org/eclipse/smarthome/model/GenerateItems.mwe2`).
For the Mwe2Launcher the JAR file "antlr-generator-3.2.0-patch.jar" is added to the classpath.

This has been already done this way in ESH.
I assume because of license reasons that file has not been added to the repository.
To ensure that it is present another Maven plugin has been added to download that file.
As we need to ensure that this part is executed in front of all model bundles, the logic has been added to the bundle POM (the parent of the models).

Don't ask me about all that emf, mwe2, xtend, ... stuff.
For me it is a kind of black magic.
Projects e.g. model.script and model.rules generates files in the other ones directory, they don't follow the "normal" Maven directory structure, we need the exec plugin, ...

I just want to drop the additional "download that file" part from the build to clean up the situation a little bit.

The location does not really matter, we just need to reference it in every model POM that needs to call the mwe2 launcher. I leave it the same as before just now added to the repo instead of downloaded dynamically...

PR #818, opened 2019-05-14
set http context ID for smarthome HTTP context

maggu2810:
The Pax Web Extender is much more happy if this property is present.
Let's make it happy...

Fixes: https://github.com/openhab/openhab-distro/issues/891


maggu2810:
Would be great is someone could test this before merge if it really fixes the linked issue.

PR #819, opened 2019-05-14
rule engine should not depend on specific registry implementation

maggu2810:
The current rule registry implementation provides a configuration for the "rule reinitialization delay". That configuration is not used by the rule registry itself but used by the rule engine implementation.
This required the rule engine implementation to contain a special code path that checks if the injected rule registry is an instance of that special implementation to access that configuration parameter.
As the configuration is not used by the registry itself, it does not make sense that it is a configuration of that component.
If the rule engine (at least this special implementation) would like to use a configurable parameter, it should be part of that specific component implementation.

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleEngineImpl.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleEngineImpl.java
@@ -74,6 +74,7 @@ import org.openhab.core.automation.type.TriggerType;
 import org.openhab.core.automation.util.ReferenceResolver;
 import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.ComponentPropertyType;
 import org.osgi.service.component.annotations.Deactivate;
 import org.osgi.service.component.annotations.Reference;
 import org.osgi.service.component.annotations.ReferenceCardinality;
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleEngineImpl.java
```diff
@@ -103,6 +104,14 @@
 @NonNullByDefault
 public class RuleEngineImpl implements RuleManager, RegistryChangeListener<ModuleType> {
 
+    @ComponentPropertyType
```
cweitkamp: This pattern is new in OSGi Compendium 7 too? Am I right?

Underscores will be replaced by "." thus the property name still is "rule.reinitialization.delay"?

maggu2810: It should :wink: 
You could have a look at the spec or to get a very short overview, have a look at: https://www.eclipsecon.org/europe2017/sites/default/files/slides/Constructor%20injection%20and%20other%20new%20features%20for%20Declarative%20Services%201.4.pdf page 14ff

cweitkamp: Yes, thanks. I found and read that presentation too.

maggu2810:
And to be honest I don't know if anyone is using that configuration property at all.
I just don't want to remove the option to configure it, as it has been possible before.

Issue #815, opened 2019-05-13
MQTT transport connection fails after upgrade of Paho client


mherwege:
The openHAB Niko Home Control binding uses the MQTT transport for connecting to an embedded MQTT broker in the Niko Home Control system. Since the Paho upgrade, the transport does not connect to the broker anymore.
See also [#5488](https://github.com/openhab/openhab2-addons/issues/5488).
In debugging, I see the MqttBrokerConnection.start() method ([L736](https://github.com/openhab/openhab-core/blob/4480e041375c6c9129c49764a1be598af7cae89a/bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java#L736)) run to the last line, but the future returns false on completion, indicating the connection failed. I can't figure out why it does so now, and did not do so with a previous version of the Paho client.
My code, calling start() is here: [L147](https://github.com/openhab/openhab2-addons/blob/6e057c17678a2477f7f3e12fa60fc53a996e22a2/bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/protocol/nhc2/NhcMqttConnection2.java#L147). I am getting false returned (and no exception from start or get).

davidgraeff:
The only other part where I set a boolean value to this future is here: https://github.com/openhab/openhab-core/blob/4480e041375c6c9129c49764a1be598af7cae89a/bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/MqttActionAdapterCallback.java#L35

So I'm puzzled why it returns false.

mherwege:
@davidgraeff 
What about [L153](https://github.com/openhab/openhab-core/blob/4480e041375c6c9129c49764a1be598af7cae89a/bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java#L153)? I believe the BrokerConnectionCallback is where this return value is being set.
[...]

PR #820, opened 2019-05-14
MQTT: disable https hostname verification

maggu2810:
After the bump of Paho from 1.2.0 to 1.2.1 the library enables the https hostname verification.
This breaks compatibility with current consumers.
Connections cannot be established anymore.

We disable the https hostname verification to keep the old behaviour.
We should add an API so the hostname verification can be enabled if desired.

This also makes the version 1.2.1 the lower version limit as we need to use the API to disable the verification internally.

Related to: https://github.com/openhab/openhab-core/issues/815

### DIFF
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -206,7 +206,7 @@
     <dependency>
       <groupId>org.eclipse.paho</groupId>
       <artifactId>org.eclipse.paho.client.mqttv3</artifactId>
-      <version>1.2.0</version>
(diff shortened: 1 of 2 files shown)


mherwege:
@maggu2810 I tested and I can confirm this solves the problem with my binding.

maggu2810:
Thank you for testing @mherwege 
We just need another @openhab/core-maintainers that agree with this PR and press the merge button.

PR #817, opened 2019-05-13
[archetype] Fixed generation of binding skeletons

cweitkamp:
- Fixed generation of binding skeletons

After https://github.com/openhab/openhab2-addons/pull/5555 which splitted the feature file into separate files we need some adjustments in the skeleton script

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-archetype-plugin:3.1.0:generate (default-cli) on project org.openhab.addons.reactor.bundles: java.io.FileNotFoundException: /home/christoph/Downloads/openhab-master/git/openhab2-addons/bundles/../features/karaf/openhab-addons/src/main/feature/feature.xml (Datei oder Verzeichnis nicht gefunden) -> [Help 1]
```

@J-N-K @Hilbrand May I ask you for validation? Thanks.

Also-by: Hilbrand Bouwkamp <hilbrand@h72.nl>
Signed-off-by: Christoph Weitkamp <github@christophweitkamp.de>

PR #825, opened 2019-05-15
Update archetype-post-generate.groovy

davidgraeff:
My dream for this file is to be almost empty. Remove features file modification, we have a local file per bundle now.

### DIFF
--- a/tools/archetype/binding/src/main/resources/META-INF/archetype-post-generate.groovy
+++ b/tools/archetype/binding/src/main/resources/META-INF/archetype-post-generate.groovy
@@ -75,33 +75,6 @@ def addBundleToBom(bundleAfter, newBundle) {
      println 'Added bundle to bom pom'
 }
 
-/**
- * Add the new bundle to the feature.xml
- */
-def addBundleToFeature(bundleAfter, newBundle, bundleName) {
-    def feature = '''
-    <feature name="##1" description="##2 Binding" version="${project.version}">
-        <feature>openhab-runtime-base</feature>
-        <bundle start-level="80">mvn:org.openhab.addons.bundles/##3/${project.version}</bundle>
-    </feature>
-'''.replace('##1', newBundle.replace('org.', '').replace('.', '-')).replace('##2', bundleName).replace('##3', newBundle)
-    def bomFile = new File(outputDirectory, '../features/openhab-addons/src/main/feature/feature.xml')
-    def newContent = ''
-    def insertIndex = 0;
-    def lines = bomFile.eachLine { line, index ->
-        newContent += line + nl
-        if (line.contains(bundleAfter) && line.contains('<bundle')) {
-            insertIndex = index + 1
-        }
-        if (insertIndex > 0 && index == insertIndex) {
-            newContent += feature
-            insertIndex = 0
-        }
-    }
-     bomFile.write(newContent)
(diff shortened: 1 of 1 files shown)


davidgraeff:
It might not be the smartest way to have the archetype for generating bundles for another repo in here. Each time we change something over there, we need to remember to also adapt here.

maggu2810:
> It might not be the smartest way to have the archetype for generating bundles for another repo in here. 

Shouldn't that archetype be valid for every openHAB bundle, also new openHAB Core bundles?

cweitkamp:
Similar to #817. Without adding the feature.xml

davidgraeff:
True. This is a duplicate

davidgraeff:
@maggu2810 I don't think that will work in the long run:
* openHAB Core: One project, multiple bundles, one release
* openHAB addons: Multiple independent projects, not necessary to release together

Therefore the directory structure might indeed be a bit different between core and addon repos.

maggu2810:
If the archetype is for "addons" only and cannot be used for "core" then I do not see any reason to maintain it here. Why not moving it to the "addons" repo if it is for addons only?

davidgraeff:
That's what I'm thinking as well. The archetype will still be uploaded to a maven repo and can be used by other openhab2-addons-inofficial repos, but it cannot be used by openhab-core itself.

maggu2810:
See also

---

 https://github.com/openhab/openhab-core/pull/589#issuecomment-467380732

> If we can design the archetype that it works for every bundle perhaps core makes the most sense.

---

https://github.com/openhab/openhab-core/pull/589#issuecomment-477097678

> I would prefer it to be in openhab-core (as it was in ESH in the past) as the skeleton can be seen as a tooling that is closely related to the API that we define in the core. That is, whenever we change something on the APIs and the way we expect add-ons to be developed, it should be done in a single PR against this repo.

---

For me it seems that this archetype is not only about the API anymore but also about the directory layout chosen by @openhab/2-x-add-ons-maintainers for its addons.
@kaikreuzer WDYT? Makes it sense to require PRs to core because the layout of the projects in addons changes?

davidgraeff:
I think there is just still a lot of confusion about each repos purpose. Understandable because it is nowhere written down.

The Eclipse Smarthome repo was more than just a core framework, it was a full solution.
It contained the docu, a demo app, example bundles, "official" extensions and of course the archetype for developing more extensions. That is no more. openHAB-core is just the core bundles.

kaikreuzer:
> If the archetype is for "addons" only and cannot be used for "core" then I do not see any reason to maintain it here. 

There shouldn't be "one archetype". The one that we have is the template how the core framework expects bindings to be implemented and provide a skeleton for it. So yes, for me, this clearly belongs in the core repo as it is valid for ANY binding being implemented (not just the ones in openhab2-addons).

> @kaikreuzer WDYT? Makes it sense to require PRs to core because the layout of the projects in addons changes?

Ideally, there should not be any dependency to a repo layout and we didn't have such a dependency in the past. If someone wants to create a custom binding in their own repo, this should be possible.
If we suggest a common directory layout that should be followed for any add-ons repo, then it is imho ok to also expect this to be followed when using the archetype.

davidgraeff:
So the solution would be to move "archetype-post-generate" to openhab2-addons?

PR #761, opened 2019-04-27
use upstream nrjavaserial 3.15.0 on runtime

maggu2810:
Fixes: https://github.com/openhab/openhab-core/issues/750

I see two possible fixes for #750:
* use the upstream library all the time and drop the OH specific release of nrjavaserial
* state that the OH nrjavaserial provides the specific capability (`openhab.tp;feature=serial;impl=rxtx`)

I don't know what's the difference between 3.15.0.OH2 of openHAB's nrjavaserial and 3.15.0 of the upstream one.
The nrjavaserial repository of openHAB does not contain any tag or commit that tells me anything about that.

So, let's wait for a reply to https://github.com/openhab/openhab-core/issues/750#issuecomment-487284749

At the moment I switch to the upstream one in this PR, but we can also remain at the OH spefic one, if working with upstream of nrjavaserial is not possible.

PR #833, opened 2019-05-25
Use nrjavaserial 3.15.0.OH2 in BOMs

wborn:
When debugging openHAB in Eclipse I ran into serial port locking issues again on Ubuntu 18.04.2.
We resolved these in the distro by using 3.15.0.OH2 which doesn't use liblockdev.

Errors similar to the one below may show after stopping/restarting openHAB when using serial ports on certain distros:

```
RXTX fhs_lock() Error: opening lock file: /var/lock/LCK..ttyUSB0: File exists. It is mine

testRead() Lock file failed
````

See also: https://github.com/openhab/openhab-core/pull/761

### DIFF
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -232,14 +232,9 @@
       <scope>compile</scope>
     </dependency>
     <dependency>
-      <groupId>org.openhab</groupId>
+      <groupId>com.neuronrobotics</groupId>
       <artifactId>nrjavaserial</artifactId>
-      <!-- Use "no liblockdev" version -->
-      <!-- See: https://github.com/openhab/openhab-core/pull/761 -->
-      <!-- See: https://github.com/openhab/openhab-core/issues/750 -->
-      <!-- <groupId>com.neuronrobotics</groupId> -->
-      <!-- <version>3.15.0</version> -->
-      <version>3.15.0.OH2</version>
+      <version>3.14.0</version>
       <scope>compile</scope>
     </dependency>
 
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -232,9 +232,14 @@
       <scope>compile</scope>
     </dependency>
     <dependency>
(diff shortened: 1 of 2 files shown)


maggu2810:
Why does it need to be changed for the compilation? Is there any API difference?
Are you that the qualifier (OH2) is not part of the version range in the resulting import package part of the manifests of the bundles using nrjavaserial API?

wborn:
It looks like it should also work without updating the BOM used for compilation. I see it still adds the 3.15.0.OH2 version to the index and shows it in the Maven dependencies of the Demo app.

To keep things simple I personally prefer to keep such versions in sync. It's easy to overlook updating one of these dependencies when the API does change if we upgrade to 3.16.0(.OH). As far as I can see the API didn't change between 3.14.0 and 3.15.0.OH2.

maggu2810:
> To keep things simple I personally prefer to keep such versions in sync.

To keep things for consumer of the framework simple I would prefer not to change the minimum version requirements of third party libraries if there is not any need for.
What's the reason to break distributions that are using 3.14 and that does not realize any problem the last years?

> It's easy to overlook updating one of these dependencies when the API does change if we upgrade to 3.16.0(.OH).

Why?
If you need a newly introduced API, you must change the compilation dependencies, otherwise you can't use the API.
If you change the compilation dependencies you must change the runtime dependencies (Bnd + Karaf), otherwise tests resolution, feature verification, ... will not work.

So, if there is not any reason to stop support products using 3.14 don't do it.
I fully agree to change the runtime BOM and Karaf features if openHAB Distri and Demo would like to use more recent stuff.
But changing requirements (e.g. minimum version) should not be done without any need (if there is one, this would be another topic).

wborn:
I understand the point of view from consumers is a bit different from that of maintainers so I've reverted the compile BOM changes @maggu2810 .

maggu2810:
Thank you!

PR #836, opened 2019-05-27
split activation logic

maggu2810:
If a component should be activated an instance is constructed and that object is activated.
Let's split the construction and activation logic also if constructor injection is used.


maggu2810:
> May I ask you to check Travis? I guess it failed because we have to call the `activate()` method in the `AutomaticInboxProcessor` instance now.

Done.

> Or change the test to extend `JavaOSGiTest` to make use of the `getService()` method.

In the long run I would prefer to use OSGi tests for components instead of mocking the whole world.

maggu2810:
The PR-openHAB-Core builder failed because of unrelated mDNS tests.
I triggered a new build.

Issue #834, opened 2019-05-27
NPEs in 'AutomaticInboxProcessor' during startup


cweitkamp:
```
2019-05-27 12:15:09.364 [ERROR] [nal.common.AbstractInvocationHandler] - An error occurred while calling method 'EventSubscriber.receive()' on 'org.eclipse.smarthome.config.discovery.internal.AutomaticInboxProcessor@36897ddd': null
java.lang.NullPointerException: null
	at org.eclipse.smarthome.config.discovery.internal.AutomaticInboxProcessor.receiveTypedEvent(AutomaticInboxProcessor.java:113) ~[?:?]
	at org.eclipse.smarthome.config.discovery.internal.AutomaticInboxProcessor.receiveTypedEvent(AutomaticInboxProcessor.java:1) ~[?:?]
	at org.eclipse.smarthome.core.events.AbstractTypedEventSubscriber.receive(AbstractTypedEventSubscriber.java:54) ~[133:org.openhab.core:2.5.0.201905250301]
	at sun.reflect.GeneratedMethodAccessor38.invoke(Unknown Source) ~[?:?]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:?]
	at java.lang.reflect.Method.invoke(Method.java:498) ~[?:?]
	at org.eclipse.smarthome.core.internal.common.AbstractInvocationHandler.invokeDirect(AbstractInvocationHandler.java:153) [133:org.openhab.core:2.5.0.201905250301]
	at org.eclipse.smarthome.core.internal.common.Invocation.call(Invocation.java:53) [133:org.openhab.core:2.5.0.201905250301]
	at java.util.concurrent.FutureTask.run(FutureTask.java:266) [?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [?:?]
	at java.lang.Thread.run(Thread.java:748) [?:?]
```

```
2019-05-27 12:15:09.388 [ERROR] [me.core.internal.events.EventHandler] - Dispatching/filtering event for subscriber 'org.eclipse.smarthome.core.events.EventSubscriber' failed: null
java.lang.NullPointerException: null
[...]

PR #837, opened 2019-05-28
Added nullness annotations

cweitkamp:
- Added nullness annotations

See https://github.com/openhab/openhab-core/issues/834#issuecomment-496301592

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ThingLinkManager.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ThingLinkManager.java
@@ -15,6 +15,7 @@ package org.eclipse.smarthome.core.thing.link;
 import java.util.List;
 import java.util.concurrent.ScheduledExecutorService;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.smarthome.core.common.ThreadPoolManager;
 import org.eclipse.smarthome.core.common.registry.Provider;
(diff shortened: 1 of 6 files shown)


bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ThingLinkManager.java
```diff
@@ -67,11 +69,11 @@
 
     private final ScheduledExecutorService scheduler = ThreadPoolManager.getScheduledPool(THREADPOOL_NAME);
 
-    private ThingRegistry thingRegistry;
-    private ManagedThingProvider managedThingProvider;
-    private ItemRegistry itemRegistry;
-    private ItemChannelLinkRegistry itemChannelLinkRegistry;
-    private ChannelTypeRegistry channelTypeRegistry;
+    private @NonNullByDefault({}) ThingRegistry thingRegistry;
```
maggu2810: Wouldn't it make sense to use constructor injection and ensure that every instance got a non null thing registry (and the other below)?

I agree that such a change can be separated, but perhaps we can do it before and so not need to add non null by default exclusions.

maggu2810: Can also be done later...

cweitkamp: I thought about that too. But it is more clear if we have it in a separate PR.

PR #842, opened 2019-05-30
Use bnd for including resources to workaround infinite Eclipse builds

wborn:
When "Build Automatically" is enabled in Eclipse, the build never ends.
It seems there is a bnd issue with the way resources are defined in the reactor POM.
The workaround seems to be to include all resources using bnd.

See also:

* https://github.com/bndtools/bnd/issues/3220
* https://github.com/openhab/openhab2-addons/issues/5554

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -139,18 +139,6 @@
   </pluginRepositories>
 
   <build>
-    <resources>
-      <resource>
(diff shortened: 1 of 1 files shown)


cweitkamp:
Thanks. Let's see how it is working.

PR #847, opened 2019-06-03
UIDs: Allow percent character

davidgraeff:
Extend the current regex expression to allow UID segments to contain percent characters.

This helps a lot if you need to map arbitrary utf8 strings (like in MQTT where I need to map topics to channel IDs),
because you can use URI percentage encoding.

See also: https://github.com/openhab/openhab2-addons/issues/5632#issuecomment-498185568

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/common/AbstractUID.java
+++ b/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/common/AbstractUID.java
@@ -28,7 +28,7 @@ import org.eclipse.jdt.annotation.Nullable;
 @NonNullByDefault
 public abstract class AbstractUID {
 
-    public static final String SEGMENT_PATTERN = "[A-Za-z0-9_-]*";
+    public static final String SEGMENT_PATTERN = "[\\w-%]*";
     public static final String SEPARATOR = ":";
     private final List<String> segments;
 


maggu2810:
If it is just about mapping there are several mapping functions that don't need to add another character to the allowed ones.
I don't know all the UID usages. Perhaps someone relies on the fact that every UID can be used as a filename or whatever. Isn't the percent character reserved for on some filesystems?

@kaikreuzer What has been the reasons about the subset of allowed characters?

davidgraeff:
File systems:
Windows NTFS/Fat: OK (https://docs.microsoft.com/en-us/windows/desktop/fileio/naming-a-file)
Linux btrfs/ext3/4: OK (tested)
Mac OS: OK (http://www.informit.com/articles/article.aspx?p=1144082&seqNum=5, Table 3.1)

kaikreuzer:
> @kaikreuzer What has been the reasons about the subset of allowed characters?

To avoid issues with DSL and rule syntaxes, url (and general) readability, unexpected side-effects due to unexpected characters, etc. I would very much hate to see percent signs within UIDs as they would become very cryptic and unreadable...

davidgraeff:
Is there any other way to encode utf8 strings? I mean I could also just assign random UUIDs to discovered MQTT topics. I think Base64 encoded IDs might be long for people to handle them.

maggu2810:
I assume base64 will use invalid characters.
Base36 should work.

maggu2810:
@davidgraeff Wait a moment, I will come with a suggestion.

maggu2810:
https://github.com/maggu2810/openhab-core/commit/4c6763996acc3b8d43daae92f06bf8bb145d14ae

Test output looks like:

```
in: test
 encoded: test
 decoded: test
 equals: true
in: TEST
 encoded: TEST
 decoded: TEST
 equals: true
in: test123TEST
 encoded: test123TEST
 decoded: test123TEST
 equals: true
in: test_test-test%test
 encoded: test_5Ftest_2Dtest_25test
 decoded: test_test-test%test
 equals: true
```

davidgraeff:
Thanks Markus. I think your implementation works for the usecase and I will adapt it in the MQTT binding.

kaikreuzer:
I assume this PR can hence be closed.

PR #848, opened 2019-06-03
Update ChannelUID.java

davidgraeff:
Extend the current regex expression to allow UID segments to contain percent characters.

This helps a lot if you need to map arbitrary utf8 strings (like in MQTT where I need to map topics to channel IDs),
because you can use URI percentage encoding.


kaikreuzer:
Looks related to https://github.com/openhab/openhab-core/pull/847.

kaikreuzer:
Closing, see https://github.com/openhab/openhab-core/pull/847#issuecomment-498555018.

PR #850, opened 2019-06-03
Added nullness annotations

cweitkamp:
- Added nullness annotations

### DIFF
--- a/bundles/org.openhab.core.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleEngineImpl.java
+++ b/bundles/org.openhab.core.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleEngineImpl.java
@@ -21,6 +21,8 @@ import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 
 import org.eclipse.emf.ecore.EObject;
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.smarthome.core.common.ThreadPoolManager;
 import org.eclipse.smarthome.core.events.Event;
 import org.eclipse.smarthome.core.events.EventFilter;
@@ -70,6 +72,7 @@ import com.google.inject.Injector;
  */
 @SuppressWarnings("restriction")
 @Component(immediate = true, service = { EventSubscriber.class, RuleEngine.class })
+@NonNullByDefault
 public class RuleEngineImpl implements ItemRegistryChangeListener, StateChangeListener, ModelRepositoryChangeListener,
         RuleEngine, EventSubscriber {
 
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.core.persistence/src/main/java/org/eclipse/smarthome/core/persistence/internal/PersistenceManagerImpl.java
```diff
@@ -51,13 +52,15 @@
 import org.osgi.service.component.annotations.Reference;
 import org.osgi.service.component.annotations.ReferenceCardinality;
 import org.osgi.service.component.annotations.ReferencePolicy;
+import org.quartz.Job;
```
cweitkamp: Those imports were added because of a reference in the JavaDoc for the `removeTimers()` method. As we are not using Quartz anymore we should fix the JavaDoc parts.

https://github.com/openhab/openhab-core/blob/ca1cda3f7521d56b85c4ee87a7417ab5aeb4e6f4/bundles/org.openhab.core.persistence/src/main/java/org/eclipse/smarthome/core/persistence/internal/PersistenceManagerImpl.java#L360-L374

maggu2810: full ack

bundles/org.openhab.core.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleEngineImpl.java
```diff
@@ -83,11 +86,10 @@
     private final ModelRepository modelRepository;
     private final ScriptEngine scriptEngine;
 
-    private RuleTriggerManager triggerManager;
+    private @NonNullByDefault({}) Injector injector;
+    private @NonNullByDefault({}) RuleTriggerManager triggerManager;
```
maggu2810: Can't we move the two lines https://github.com/openhab/openhab-core/blob/ca1cda3f7521d56b85c4ee87a7417ab5aeb4e6f4/bundles/org.openhab.core.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleEngineImpl.java#L109-L110 to the constructor and drop the additional `@NonNullByDefault({})`.
There is also no need set `triggerManager` to `null` on deactivation (it is already cleared).

Or do you prefer to separate the changes, too?

cweitkamp: Yes, that should be possible. I would like to do that in a separate PR because there are quite some `null` checks in the code on the `triggerManager` which can be removed in advance.

bundles/org.openhab.core.persistence/src/main/java/org/eclipse/smarthome/core/persistence/internal/PersistenceManagerImpl.java
cweitkamp: See #852

PR #681, opened 2019-03-28
Adding DataAmount (bit and derivatives) and DataTransferRates (bits p…

clinique:
…er second) to UoM

Signed-off-by: Gaël L'hopital <gael@lhopital.org>

PR #851, opened 2019-06-03
automation: Refactoring of 'DefaultScriptScopeProvider'

### LINKED ISSUE #846: [automation] RefreshType.REFRESH is not included in default imports
I plan to submit a PR for this. For reference, here is an ESH commit where some other things were added to [DefaultScriptScopeProvider](https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/defaultscope/DefaultScriptScopeProvider.java) (and don't forget to update the docs)... https://github.com/eclipse/smarthome/commit/22d5908d7bf5c6f521be5b921ab467c1bb28ce64#diff-00cac509a3b03592e161cd89469c365e.

cweitkamp:
- Use constructor injection to simplify lifecycle
- Added `RefreshType.REFRESH` and `BinaryPrefix` to imports
- General refactoring of class `DefaultScriptScopeProvider`

The `DefaultScriptScopeProvider` needed a little bit more refactoring to be NPE safe and to implement the OSGi lifecycle handling correctly. The changes made it possible to drop the workaround for any `thingActions` queued before the component got activated.

See https://github.com/openhab/openhab-core/issues/834#issuecomment-496833158

Closes #846

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/defaultscope/DefaultScriptScopeProvider.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/defaultscope/DefaultScriptScopeProvider.java
@@ -16,10 +16,8 @@ import java.io.File;
 import java.net.URLEncoder;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashMap;
-import java.util.LinkedList;
 import java.util.Map;
-import java.util.Queue;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.FilenameUtils;
@@ -43,6 +41,7 @@ import org.eclipse.smarthome.core.library.types.StopMoveType;
 import org.eclipse.smarthome.core.library.types.StringListType;
 import org.eclipse.smarthome.core.library.types.StringType;
 import org.eclipse.smarthome.core.library.types.UpDownType;
+import org.eclipse.smarthome.core.library.unit.BinaryPrefix;
 import org.eclipse.smarthome.core.library.unit.ImperialUnits;
 import org.eclipse.smarthome.core.library.unit.MetricPrefix;
 import org.eclipse.smarthome.core.library.unit.SIUnits;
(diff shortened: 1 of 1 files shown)


maggu2810:
Is there any reason to keep the old activation method at all?
Can't this all be moved to the constructor?

I don't see any usage of this or similar, so I assume no separation is needed.

openhab-5iver:
When I saw #840, I looked forward to when/if you'd be getting to the new rule engine! I hope to be wrapping up several commits for the helper libraries today, and I was then planning to get back into some Java things and the automation integration tests would be helpful for this PR. I'm sure you realize, this is a crucial piece of scripted automation. 

I'll try to start testing this later today, but are you planning similar PRs that would affect automation? It might save time if I test them all together.

cweitkamp:
> Is there any reason to keep the old activation method at all?

No, there is not.

> Can't this all be moved to the constructor?

Yes, but we even do not need to assign all those values in the ctor. We can use a `static final` immutable `Map` for them.

@openhab-5iver While I am touching it anyway, shall I add the `RefreshType.REFRESH` like requested in #846 too?

maggu2810:
> Yes, but we even do not need to assign all those values in the ctor. We can use a `static final` immutable `Map` for them.

You should not add non static members (e.g. `ruleRegistry`, `thingRegistry`, ...) to the map or am I missing something?

cweitkamp:
> You should not add non static members (e.g. `ruleRegistry`, `thingRegistry`, ...) to the map or am I missing something?

Arg, ... Yes, you are right.

openhab-5iver:
> shall I add the RefreshType.REFRESH

I was going to check to see if anything else was missing too, but that can be done separately. I haven't touched it yet, so go for it! I'm working a PR to update the [doc](https://www.openhab.org/docs/configuration/jsr223.html#default-preset-importpreset-not-required) right now, so I can take care of that side.

cweitkamp:
> I was going to check to see if anything else was missing too, but that can be done separately.

I double checked it. The recently introduced `BinaryPrefix`es (#681) are missing. Everything else should be available.

cweitkamp:
Finished.

@openhab-5iver A runtime test from your side will be very much appreciated.

maggu2810:
Do you see any critical potential regression (I did not), so we should wait for a runtime test or do you agree to merge it and fix a regression (if there will be one) afterwards?

openhab-5iver:
It'll take me a couple hours before I can test this. I think that this PR has the potential for breaking scripted automation. Without itests, my preference would be to at least smoke it before a merge, to prevent any potential regressions from getting into a build.

PR #853, opened 2019-06-03
model: Refactoring of 'RuleEngineImpl'

cweitkamp:
- Refactoring of `RuleEngineImpl`

See https://github.com/openhab/openhab-core/pull/850#discussion_r289796348


bundles/org.openhab.core.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleEngineImpl.java
```diff
@@ -245,23 +243,21 @@ private void internalItemRemoved(Item item) {
 
     @Override
     public void modelChanged(String modelName, org.eclipse.smarthome.model.core.EventType type) {
```
cweitkamp: Does it hurt if I will change the namespace here? And use an `import` for it?

maggu2810: You want to import `org.eclipse.smarthome.model.core.EventType` and use `EventType type` instead of the FQCN?

I don't know why the FQCN has been used which other `EventType` has been present.

If there is no other one anymore, feel free to import `org.eclipse.smarthome.model.core.EventType` and use `EventType`.

IMHO The API shouldn't be touched.

PR #763, opened 2019-04-27
safe caller: respect classloader of the class and the interfaces

maggu2810:
Related to: https://github.com/openhab/openhab2-addons/issues/5519

PR #861, opened 2019-06-09
don't use safe caller for event delegation to subscribers

maggu2810:
This changes should address #829 

At the moment we warn if an event subscriber exceeds the limit of expected max. runtime.
This has been done before, too.
It would also be possible (which another implementation I played around today, too) to interrupt the event subscriber if it exceeds the limit. I don't think we should do this. This could perhaps lead into a very confusion behaviour on very high load. The warning should be enough to detect a really bad working event subscriber.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/internal/events/EventHandler.java
+++ b/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/internal/events/EventHandler.java
@@ -16,10 +16,15 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
-import org.eclipse.smarthome.core.common.SafeCaller;
+import org.eclipse.smarthome.core.common.NamedThreadFactory;
 import org.eclipse.smarthome.core.events.Event;
 import org.eclipse.smarthome.core.events.EventFactory;
 import org.eclipse.smarthome.core.events.EventFilter;
@@ -33,26 +38,36 @@ import org.slf4j.LoggerFactory;
  * @author Markus Rathgeb - Initial contribution
  */
 @NonNullByDefault
-public class EventHandler {
+public class EventHandler implements AutoCloseable {
+
+    private static final long EVENTSUBSCRIBER_EVENTHANDLING_MAX_MS = TimeUnit.SECONDS.toMillis(5);
 
     private final Logger logger = LoggerFactory.getLogger(EventHandler.class);
 
     private final Map<String, Set<EventSubscriber>> typedEventSubscribers;
     private final Map<String, EventFactory> typedEventFactories;
-    private final SafeCaller safeCaller;
+
+    private final ScheduledExecutorService watcher = Executors
+            .newSingleThreadScheduledExecutor(new NamedThreadFactory("EventHandlerWatcher"));
+    private final ExecutorService executor = Executors
+            .newSingleThreadExecutor(new NamedThreadFactory("EventHandlerExecutor"));
 
     /**
      * Create a new event handler.
      *
      * @param typedEventSubscribers the event subscribers indexed by the event type
      * @param typedEventFactories the event factories indexed by the event type
-     * @param safeCaller the safe caller to use
      */
     public EventHandler(final Map<String, Set<EventSubscriber>> typedEventSubscribers,
-            final Map<String, EventFactory> typedEventFactories, final SafeCaller safeCaller) {
+            final Map<String, EventFactory> typedEventFactories) {
         this.typedEventSubscribers = typedEventSubscribers;
         this.typedEventFactories = typedEventFactories;
-        this.safeCaller = safeCaller;
+    }
+
+    @Override
+    public void close() {
+        watcher.shutdownNow();
+        executor.shutdownNow();
     }
 
     public void handleEvent(org.osgi.service.event.Event osgiEvent) {
@@ -131,17 +146,22 @@ public class EventHandler {
             EventFilter filter = eventSubscriber.getEventFilter();
             if (filter == null || filter.apply(event)) {
                 logger.trace("Delegate event to subscriber ({}).", eventSubscriber.getClass());
-                safeCaller.create(eventSubscriber, EventSubscriber.class).withAsync().onTimeout(() -> {
-                    logger.warn("Dispatching event to subscriber '{}' takes more than {}ms.",
-                            eventSubscriber.toString(), SafeCaller.DEFAULT_TIMEOUT);
-                }).onException(e -> {
-                    logger.error("Dispatching/filtering event for subscriber '{}' failed: {}",
-                            EventSubscriber.class.getName(), e.getMessage(), e);
-                }).build().receive(event);
+                executor.submit(() -> {
+                    ScheduledFuture<?> logTimeout = watcher.schedule(
+                            () -> logger.warn("Dispatching event to subscriber '{}' takes more than {}ms.",
+                                    eventSubscriber, EVENTSUBSCRIBER_EVENTHANDLING_MAX_MS),
+                            EVENTSUBSCRIBER_EVENTHANDLING_MAX_MS, TimeUnit.MILLISECONDS);
+                    try {
+                        eventSubscriber.receive(event);
+                    } catch (final Exception ex) {
+                        logger.error("Dispatching/filtering event for subscriber '{}' failed: {}",
+                                EventSubscriber.class.getName(), ex.getMessage(), ex);
+                    }
+                    logTimeout.cancel(false);
+                });
             } else {
                 logger.trace("Skip event subscriber ({}) because of its filter.", eventSubscriber.getClass());
             }
         }
     }
-
 }
--- a/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/internal/events/OSGiEventManager.java
+++ b/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/internal/events/OSGiEventManager.java
@@ -18,7 +18,6 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArraySet;
 
-import org.eclipse.smarthome.core.common.SafeCaller;
 import org.eclipse.smarthome.core.events.Event;
 import org.eclipse.smarthome.core.events.EventFactory;
 import org.eclipse.smarthome.core.events.EventSubscriber;
@@ -51,11 +50,9 @@ public class OSGiEventManager implements EventHandler {
 
     private ThreadedEventHandler eventHandler;
 
-    private SafeCaller safeCaller;
-
     @Activate
     protected void activate(ComponentContext componentContext) {
-        eventHandler = new ThreadedEventHandler(typedEventSubscribers, typedEventFactories, safeCaller);
+        eventHandler = new ThreadedEventHandler(typedEventSubscribers, typedEventFactories);
         eventHandler.open();
     }
 
@@ -114,15 +111,6 @@ public class OSGiEventManager implements EventHandler {
         for (String supportedEventType : supportedEventTypes) {
             typedEventFactories.remove(supportedEventType);
         }
-    }
-
-    @Reference
-    protected void setSafeCaller(SafeCaller safeCaller) {
-        this.safeCaller = safeCaller;
-    }
-
-    protected void unsetSafeCaller(SafeCaller safeCaller) {
-        this.safeCaller = null;
     }
 
     @Override
--- a/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/internal/events/ThreadedEventHandler.java
+++ b/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/internal/events/ThreadedEventHandler.java
@@ -22,7 +22,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
-import org.eclipse.smarthome.core.common.SafeCaller;
 import org.eclipse.smarthome.core.events.EventFactory;
 import org.eclipse.smarthome.core.events.EventSubscriber;
 import org.osgi.service.event.Event;
@@ -50,31 +49,32 @@ public class ThreadedEventHandler implements Closeable {
      *
      * @param typedEventSubscribers the event subscribers
      * @param typedEventFactories the event factories indexed by the event type
-     * @param safeCaller the safe caller to use
      */
     ThreadedEventHandler(Map<String, Set<EventSubscriber>> typedEventSubscribers,
-            final Map<String, EventFactory> typedEventFactories, final SafeCaller safeCaller) {
+            final Map<String, EventFactory> typedEventFactories) {
         thread = new Thread(() -> {
-            final EventHandler worker = new EventHandler(typedEventSubscribers, typedEventFactories, safeCaller);
-            while (running.get()) {
-                try {
-                    logger.trace("wait for event");
-                    final Event event = queue.poll(1, TimeUnit.HOURS);
-                    logger.trace("inspect event: {}", event);
-                    if (event == null) {
-                        logger.debug("Hey, you have really very few events.");
-                    } else if (event == notifyEvent) {
-                        // received an internal notification
-                    } else {
-                        worker.handleEvent(event);
+            try (EventHandler worker = new EventHandler(typedEventSubscribers, typedEventFactories)) {
+                while (running.get()) {
+                    try {
+                        logger.trace("wait for event");
+                        final Event event = queue.poll(1, TimeUnit.HOURS);
+                        logger.trace("inspect event: {}", event);
+                        if (event == null) {
+                            logger.debug("Hey, you have really very few events.");
+                        } else if (event == notifyEvent) {
+                            // received an internal notification
+                        } else {
+                            worker.handleEvent(event);
+                        }
+                    } catch (InterruptedException ex) {
+                        Thread.currentThread().interrupt();
+                    } catch (RuntimeException ex) {
+                        logger.error("Error on event handling.", ex);
                     }
-                } catch (InterruptedException ex) {
-                    Thread.currentThread().interrupt();
-                } catch (RuntimeException ex) {
-                    logger.error("Error on event handling.", ex);
                 }
             }
-        }, "ESH-OSGiEventManager");
+        }, "OH-OSGiEventManager");
+
     }
 
     void open() {


mhilbush:
Adding a summary of my testing from [here](https://github.com/openhab/openhab-core/issues/829#issuecomment-500243482).

Test parameters
- AMD 3.7 GHz 6 core CPU with 8 GB RAM
- openHAB build 1607
- Event rate of 31.4 events/second (see note 1 below)
- org.openhab.core built from recent clone with #861 applied

Build 1607:
- CPU: ranging between 25% and 85% on one core
- Memory: ~15.4%
- Event Manager thread CPU time rate of increase (see note 2 below)
```
   8322719   ESH-OSGiEventManager
   8328524   ESH-OSGiEventManager
   8334892   ESH-OSGiEventManager
   8341976   ESH-OSGiEventManager
```

Build 1607 with org.openhab.core as described above:
- CPU: ranging between 2% and 5% with occasional spikes to 20-25%
- Memory: ~8.0%
- Event Manager thread CPU time rate of increase (see note 2 below)
```
 2839      OH-OSGiEventManager
 10243     OH-EventHandlerExecutor-1
 758       OH-EventHandlerWatcher-1
-----------------------
 2854      OH-OSGiEventManager
 10301     OH-EventHandlerExecutor-1
 762       OH-EventHandlerWatcher-1
-----------------------
 2868      OH-OSGiEventManager
 10354     OH-EventHandlerExecutor-1
 766       OH-EventHandlerWatcher-1
-----------------------
 2884      OH-OSGiEventManager
 10412     OH-EventHandlerExecutor-1
 770       OH-EventHandlerWatcher-1
-----------------------
 2900      OH-OSGiEventManager
 10471     OH-EventHandlerExecutor-1
 775       OH-EventHandlerWatcher-1```
```

**Note 1**: Event rate measured with all event types in `org.ops4j.pax.logging.cfg` set to INFO level and the following command:
```
    tail -f /opt/openhab2/userdata/logs/events.log | pv --line-mode --average-rate > /dev/null
```

**Note 2**: Event Manager thread CPU time measured by taking snapshots every 10 seconds using the following command:
```
        /opt/openhab2/runtime/bin/client -p habopen -l 0 -- threads --list | \
            awk -F"│" '{ gsub(/[ \t]+$/, "", $2);  print $4 $2 }' | \
            sort --numeric-sort --reverse
```

maggu2810:
Thanks a lot to @mhilbush for his testing!

mhilbush:
I cranked the event rate up to 69 events per second. Pretty darn good!

CPU: 15-20% with spikes to 40%
Memory: ~8.5%
Event Manager thread CPU
```
 7030      OH-OSGiEventManager
 34676     OH-EventHandlerExecutor-1
 2200      OH-EventHandlerWatcher-1
-----------------------
 7060      OH-OSGiEventManager
 35206     OH-EventHandlerExecutor-1
 2216      OH-EventHandlerWatcher-1
-----------------------
 7087      OH-OSGiEventManager
 35696     OH-EventHandlerExecutor-1
 2232      OH-EventHandlerWatcher-1
-----------------------
 7118      OH-OSGiEventManager
 36207     OH-EventHandlerExecutor-1
 2249      OH-EventHandlerWatcher-1
-----------------------
 7147      OH-OSGiEventManager
 36718     OH-EventHandlerExecutor-1
 2265      OH-EventHandlerWatcher-1
-----------------------
 7178      OH-OSGiEventManager
 37254     OH-EventHandlerExecutor-1
 2283      OH-EventHandlerWatcher-1
-----------------------
 7207      OH-OSGiEventManager
 37752     OH-EventHandlerExecutor-1
 2301      OH-EventHandlerWatcher-1
-----------------------
 ```

> Thanks a lot to @mhilbush for his testing!

Thanks. Ran a Fortune 100 web site for 8 years. Our philosophy was:
1. Test it till it breaks
2. Fix issues
3. Go back to step 1

File: bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/internal/events/EventHandler.java [hunk 1]
```diff
@@ -33,26 +38,36 @@
...
     private final Logger logger = LoggerFactory.getLogger(EventHandler.class);
 
     private final Map<String, Set<EventSubscriber>> typedEventSubscribers;
     private final Map<String, EventFactory> typedEventFactories;
-    private final SafeCaller safeCaller;
+
+    private final ScheduledExecutorService watcher = Executors
+            .newSingleThreadScheduledExecutor(new NamedThreadFactory("EventHandlerWatcher"));
+    private final ExecutorService executor = Executors
+            .newSingleThreadExecutor(new NamedThreadFactory("EventHandlerExecutor"));
```
kaikreuzer: Afair, the safeCaller does not use a single thread, but a pool of x threads. Are we safe here to reduce it to a single thread or might we run into some other issues?

I also remember a feature that we assigned events to bindings and made sure that if one binding exhausts all threads of a pool, that this does not negatively impact the others. I cannot find this in the code right now, so I assume this was not part of the internal event handling itself.

maggu2810: Events should be handled in order.
Using multiple threads for the event handling can result into a different event order.
Assume changed events "1->2", "2->3".
If you are using different threads how do you want to ensure that the submitted execution does not change the order because the one thread comes in front of the other?
The event order will be non deterministic anymore.

So, regardless how it has been done, it should be done in order.
IIRC this has been also discussed some years ago.

As written above already, we could interrupt the event subscriber if the limit has been exceeded.
But if you realize that an addon consumes to much resources, perhaps it would be better to fix that addon (and realize the bad behaviour).

mhilbush: Just to put one more data point out there...

On my test system (with a ~9 year old AMD CPU with each of 6 cores running at 3.7 GHz) I max out one core at ~90 events per second. I produce this load with 5 systeminfo things set to update all channels (except disk because I don't know the performance penalty of that function) at a refresh rate of interval_high=1, interval_medium=4.

The top cpu consuming threads are:
```
 735734    OH-EventHandlerExecutor-1
 20119     Timer-7
 19935     ESH-usb-serial-discovery-linux-sysfs-1
 18066     pool-4-thread-1
 17530     OH-OSGiEventManager
 11320     EventAdminAsyncThread #11
 11261     EventAdminAsyncThread #12
 9450      safeCall-40
 9211      OH-EventHandlerWatcher-1
 9129      safeCall-38
 8779      pool-11-thread-1
 8565      Start Level: Equinox Container: 7dcb28cd-c6e7-4135-a1f8-6d3d46449a09
 8531      safeCall-39
 6941      sshd-SshServer[5e8ba151]-nio2-thread-1
 6514      sshd-SshServer[5e8ba151]-nio2-thread-2
 6321      safeCall-37
 6212      pool-9-thread-1
 6205      ZWaveReceiveInputThread
 4087      safeCall-41
 2974      Finalizer
 2922      Reference Handler
 2681      items-67
 2461      pool-706-thread-1
 2097      fileinstall-/opt/openhab2/userdata/etc
 2078      Karaf Lock Monitor Thread
 1925      items-68
 1643      openHAB-job-scheduler_Worker-1
 1580      openHAB-job-scheduler_Worker-2
 1555      ResourcePublisher
 1516      Active Thread: Equinox Container: 7dcb28cd-c6e7-4135-a1f8-6d3d46449a09
```

Overall, openHAB is consuming a little less than 2 CPUs worth of load.
```
top - 08:03:15 up 10 days, 14:04,  2 users,  load average: 2.03, 1.43, 1.13
Tasks: 330 total,   2 running, 263 sleeping,   0 stopped,   0 zombie
%Cpu(s): 32.6 us,  1.8 sy,  0.0 ni, 65.3 id,  0.1 wa,  0.0 hi,  0.2 si,  0.0 st
KiB Mem :  8167104 total,  2072852 free,  2031328 used,  4062924 buff/cache
KiB Swap: 16876540 total, 16876540 free,        0 used.  5811896 avail Mem 

  PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND                                                                                                                                                
31984 root      20   0 7042648 747712  23360 S 181.7  9.2  65:11.77 java                                                                                                                                                   
10228 mark      20   0  819608 100000  53848 S   9.3  1.2  23:08.14 Xorg                                                                                                                                                   
15095 xrdp      20   0   59036  28976   8084 S   7.0  0.4  20:59.19 xrdp                                                                                                                                                   
10900 mark      20   0  542100  41572  25628 S   4.3  0.5   8:15.48 xfce4-terminal                                                                                                                                         
10323 mark      20   0  429236  26368  20684 S   0.7  0.3   0:41.54 xfwm4                                                                                                                                                  
 1932 root      20   0 1668112  66936  38988 S   0.3  0.8  15:09.72 dockerd                                                                                                                                                
 3341 gdm       20   0  589708  24896  18020 R   0.3  0.3   5:55.77 gsd-color                                                                                                                                              
27420 mark      20   0   13824    876    776 S   0.3  0.0   0:01.59 tail                                                                                                                                                   
31641 mark      20   0   42052   3800   2912 R   0.3  0.0   0:00.07 top                                                                                                                                                    
    1 root      20   0  226512  10252   6708 S   0.0  0.1   0:42.78 systemd                                                                                                                                                
```

kaikreuzer: @maggu2810 We never had the requirement to respect a specific order and afair, older OSGi specs didn't ask for an ordered delivery either. I remember discussions where we found out that Equinox used single threaded delivery while Felix used a thread pool instead.

It seems that OSGi 7 has now introduced [a specific setting](https://osgi.org/specification/osgi.cmpn/7.0.0/service.event.html#i1556001) for defining whether ordering is required or not.

> But if you realize that an addon consumes to much resources, perhaps it would be better to fix that addon (and realize the bad behaviour).

Sure, but the idea was that its effect is constrained to its own functionality and does not impact others to misbehave at the same time. I remember situations where a binding took too long when consuming events and this resulted in late delivery of events to other bindings, which then showed very odd behavior and it was difficult to actually figure out, which binding was the culprit. Much better if only the one behaves badly, which actually has the bug.

kaikreuzer: @mhilbush I am just trying to find out whether we might come across any new regressions, because we are changing the currently multi-threaded processing to a single-threaded processing.
I agree, if we feel there is very little risk and if we might further look into these questions after merging, we can merge it to solve the current critical issue. I just think that for the overall stability, it is best to be as close to the original behavior that we had in the last releases (and which proved to work well).

mhilbush: I understand your POV, and agree that it makes sense to be careful about changing the behavior.

maggu2810: Just to comment the current functionality: The `SafeCaller` component uses a configuration `org.eclipse.smarthome.safecaller` that contains a setting `singleThread` (default false) that ensures that every `target` (identifier) received only one event after the other. This also allows to use different threads but only one per target at the same time.

There are several options now:
* Use a single thread and leave the changes as currently provided.
* Use a single thread and collect a statistic about the time the different subscribers need to handle an event. We could also add commands to display that diagnostic information on the console (or by REST, ...).
* Use multiple threads that number is hardcoded.
* Use multiple threads that number is configurable.
* Use the executor of the "safeCaller". This allows to use the same number as before and the old configuration to increase event handling power will be the same. Sure, it is not really a safe caller object that is added to the safeCaller executor but this should not matter at all.

For all multiple thread solutions we will loose the `singleThread` option.

I don't share your thinking that multiple threads are better to workaround an addon with a long running event subscriber. If we argue that way, we should also suggest to set the thread number to the number of addons so no addon will block the other.
Another point is as long as your event subscriber is not bound to one thread it can block every thread if it runs a long time. Assume there are 5 events that should be send to every event subscriber. The long running event subscriber will receive every 5 events and if it does not return it will receive every 5 events from a different thread of the executor. And it will eat every thread. So the number of threads should be also increased if more events are triggered in a small time period, ...
It is the same as for every OSGi component lifecycle methods. They should return as fast as possible and if there is a longer action to do it should use its own worker thread. Don't we require that for the initialize methods of thing handler already?

wborn: Does this change in behavior also mean that at most one ThingHandler will be handling a command at a time? So if your rule switches 10 lights simultaneously they again switch one after another instead of many in parallel (which I think was the behavior before the SafeCaller was introducded). If that's the case I'd rather have a higher load and put my resources to use. So they can finish work quickly instead of them sitting idle and me waiting for the work to complete.

mhilbush: Adding one other datapoint to help bring this to conclusion.

I ran another test with the following org.openhab.core config. This should give a feel for the baseline performance prior to any changes.
- without #861 (this PR)
- without #763 (CombinedClassLoader)

The test maxed out one CPU core at 100%, at which point I was processing 280 events per second using a single ESH-OSGiEventManager thread.

I generated this load with 12 systeminfo things configured this way.
```
systeminfo:computer:thing0 [interval_high=1, interval_medium=1]
systeminfo:computer:thing1 [interval_high=1, interval_medium=1]
systeminfo:computer:thing2 [interval_high=1, interval_medium=1]
systeminfo:computer:thing3 [interval_high=1, interval_medium=1]
systeminfo:computer:thing4 [interval_high=1, interval_medium=1]
systeminfo:computer:thing5 [interval_high=1, interval_medium=1]
systeminfo:computer:thing6 [interval_high=1, interval_medium=1]
systeminfo:computer:thing7 [interval_high=1, interval_medium=1]
systeminfo:computer:thing8 [interval_high=1, interval_medium=1]
systeminfo:computer:thing9 [interval_high=1, interval_medium=1]
systeminfo:computer:thinga [interval_high=1, interval_medium=1]
systeminfo:computer:thingb [interval_high=1, interval_medium=1]
```

So, to compare max performance, it looks like throughput has degraded by 68% with the addition of #861 and #763.

Test 1: Baseline (without #861 and #763)
Max events per second: 280

Test 2: With #861 and #763 applied
Max events per second: 90

So, I'm wondering if there's an opportunity to enhance the performance of the CombinedClassLoader to get the performance a little closer to the baseline.

@maggu2810 If you think there's room to enhance CombinedClassLoader, it only takes me a few minutes to build org.openhab.core and apply a large amount of load to the system.

kaikreuzer: @mhilbush Many thanks for your extensive testing!
Note that the CombinedClassLoader should only have an effect on the safeCaller, so a test with #861 but without #763 should actually lead to the exactly same result. Which also means that any enhancement of the CombinedClassLoader won't have ANY effect here.

So the question is rather: What else is slowing it down by 68% although we rather expected it to be faster as the code is simpler since we don't use any proxying here anymore. @maggu2810 Any idea what might be the reason?

mhilbush: Hmm. I didn't realize that. Thanks for the clarification.

@kaikreuzer Before asking @maggu2810 to any analysis, do I need to run a test with #861 and without #763 to eliminate the effect of #763 on other users of safeCaller?

maggu2810: Have you already tried to use the same number of threads as has been used before?
I list some options related to the number of threads above.

If an event subscriber (or anything that event subscriber is using) uses the safe caller while handling the event, then the CombinedClassLoader is still part of the game.

mhilbush: > Have you already tried to use the same number of threads as has been used before?

I did not vary the thread configuration in any way. My intent was to establish the baseline performance before any changes (i.e. without #861 and #763), then run various combinations  of #861 and #763.

I'm going to rerun those tests today, as I've been making changes to things and items along the way in order to continually increase load. Now that I have a stable config for generating event load, I'll be able to get a better comparison.

After that I can try varying the thread config.

kaikreuzer: @mhilbush Do you have any update on your performance test results?

mhilbush: Sorry, @kai. I was just getting ready to post an update...

My test box crashed. It's in another location, and I don't have the ability to restart it remotely. ☹️ To make matters worse, I was away this weekend, and didn't have much time to regroup.

So, today I built another test box (a 4 core 1.6 GHz box running Ubuntu 18.04). And, rather than mess with the systeminfo binding to generate the load, I decided to write a load generator binding that I can crank up significant state update volume.

On my first baseline test run using build 1613 with #763 reverted, I was producing around 1000 events per second and still had some CPU left. I'm going to run it again with even more volume to see if I can find the upper limit.

I also built a version of `org.openhab.core` that includes #861 and #763, and another version that includes #861 and reverts #763. I'm planning to test those over the next couple days (taking a break tomorrow to play some golf 😉).

kaikreuzer: Awesome - thanks for the update and enjoy golfing tomorrow! 🥇

mhilbush: @kaikreuzer @maggu2810 Here are the results of my testing. Let me know if there's anything else you think I should test.

**Load Generator Binding Config**
24 things
20 integer channels updated per thing per cycle
350 ms between cycles
==> Predicted event rate: 1370 events/second

All tests ran for around 10 minutes and averaged around 1370 events/second

Hardware had 4 cores running at 1.6GHz 

**Three tests using Build 1613**
Baseline: Without #761 and #863
Test 1: With #761 and #863
Test 2: Without #761, With #863

#763 -- safe caller: respect classloader of the class and the interfaces
#861 -- don't use safe caller for event delegation to subscribers

**Summary and Conclusion**
- Test 1 and Test 2 perform substantially better than the baseline in terms of CPU consumption
- Test 1 and Test 2 perform better than the baseline in terms of top CPU consuming threads
- The additional CPU consumed by OH-EventHandler* appears to be more than offset by the reduced consumption of safeCall-*
- #861 should be merged

**Baseline (baseline-861-763)**
- reverts #763 -- safe caller: respect classloader of the class and the interfaces
- without #861 -- don't use safe caller for event delegation to subscribers
```
commit 133c3113f5c255ad13e0b0c014db87fde085c851 (HEAD -> baseline-861-763)
Date:   Mon Jun 17 20:27:58 2019 -0400
    Revert "safe caller: respect classloader of the class and the interfaces (#763)"
    This reverts commit 25d7f14dfcc5f93c316cb64c75bd306721e7a36e.

commit 0f1c1bbb2dc03a0e3bb004ec081fdf7569253cd1 (origin/master, origin/HEAD, master)
Date:   Thu Jun 13 20:03:16 2019 +0200
```
```
$ ls -l org.openhab.core-2.5.0-SNAPSHOT-baseline.jar
-rw-rw-r-- 1 mark mark 419839 Jun 17 20:37 org.openhab.core-2.5.0-SNAPSHOT-baseline.jar

$ zipgrep "Bundle-Version" org.openhab.core-2.5.0-SNAPSHOT-baseline.jar 
META-INF/MANIFEST.MF:Bundle-Version: 2.5.0.201906180034

220 │ Active │  80 │ 2.5.0.201906180034    │ org.openhab.core
```
*CPU Utilization*
```
09:01:48 AM  CPU    %usr   %nice    %sys %iowait    %irq   %soft  %steal  %guest  %gnice   %idle
09:01:50 AM  all   60.62    0.00   14.93    0.00    0.00    1.80    0.00    0.00    0.00   22.65
09:01:50 AM    0   61.78    0.00   13.09    0.00    0.00    0.00    0.00    0.00    0.00   25.13
09:01:50 AM    1   55.10    0.00   19.90    0.00    0.00    0.00    0.00    0.00    0.00   25.00
09:01:50 AM    2   61.34    0.00   17.01    0.00    0.00    0.00    0.00    0.00    0.00   21.65
09:01:50 AM    3   64.25    0.00    9.84    0.00    0.00    7.25    0.00    0.00    0.00   18.65
```
*Top 30 CPU Consuming Threads*
```
 291765    ESH-OSGiEventManager
 180095    pool-5-thread-1
 85673     pool-10-thread-1
 44378     EventAdminAsyncThread #11
 44205     EventAdminAsyncThread #12
 25970     safeCall-25
 25006     safeCall-8
 21660     safeCall-26
 21639     safeCall-5
 19513     safeCall-42
 18950     safeCall-1
 18740     safeCall-40
 18590     safeCall-6
 18507     safeCall-32
 18455     safeCall-29
 18430     safeCall-18
 17813     safeCall-10
 16887     safeCall-50
 16801     safeCall-21
 16785     safeCall-15
 16168     safeCall-16
 16039     safeCall-7
 15740     safeCall-23
 15690     Start Level: Equinox Container: 3c3879aa-4b5e-46e2-a007-3498eaaf2cf6
 15084     safeCall-33
 14802     safeCall-44
 14507     safeCall-3
 14189     safeCall-38
 13114     safeCall-39
 12414     safeCall-24
```

**Test 1 (test1+861+763)**
- with #763
- with #861
```
commit 3244ae7ec86bb8bf07406214f4060d893a8576a9 (HEAD -> test1+861+763)
Merge: 0f1c1bbb e8758260
Date:   Mon Jun 17 20:45:14 2019 -0400
    Merge commit 'refs/pull/861/head' of https://github.com/openhab/openhab-core into test1+861+763

commit 0f1c1bbb2dc03a0e3bb004ec081fdf7569253cd1 (origin/master, origin/HEAD, master)
Date:   Thu Jun 13 20:03:16 2019 +0200
```
```
$ ls -l org.openhab.core-2.5.0-SNAPSHOT-test1.jar
-rw-rw-r-- 1 mark mark 423622 Jun 17 20:50 org.openhab.core-2.5.0-SNAPSHOT-test1.jar

$ zipgrep "Bundle-Version" org.openhab.core-2.5.0-SNAPSHOT-test1.jar 
META-INF/MANIFEST.MF:Bundle-Version: 2.5.0.201906180046

221 │ Active   │  80 │ 2.5.0.201906180046    │ org.openhab.core
```

*CPU Utilization*
```
09:17:23 AM  CPU    %usr   %nice    %sys %iowait    %irq   %soft  %steal  %guest  %gnice   %idle
09:17:25 AM  all   38.35    0.00   10.17    0.00    0.00    1.80    0.00    0.00    0.00   49.68
09:17:25 AM    0   33.51    0.00    8.38    0.00    0.00    0.00    0.00    0.00    0.00   58.12
09:17:25 AM    1   35.57    0.00   12.89    0.00    0.00    0.00    0.00    0.00    0.00   51.55
09:17:25 AM    2   39.09    0.00   12.18    0.00    0.00    0.00    0.00    0.00    0.00   48.73
09:17:25 AM    3   44.10    0.00    7.18    0.00    0.00    7.69    0.00    0.00    0.00   41.03
```
*Top 30 CPU Consuming Threads*
```
 367154    OH-EventHandlerExecutor-1
 219607    pool-5-thread-1
 97108     OH-OSGiEventManager
 67529     OH-EventHandlerWatcher-1
 43195     EventAdminAsyncThread #12
 43119     EventAdminAsyncThread #11
 15495     Start Level: Equinox Container: e7888463-015a-47d4-9e4f-d410bdf451de
 11929     items-4
 11574     items-5
 11538     items-1
 11297     items-2
 10809     items-3
 8688      qtp1312786476-82
 7750      qtp1312786476-202
 7630      qtp1312786476-91
 6652      qtp1312786476-204
 6465      qtp1312786476-89
 5468      qtp1312786476-207
 3318      ResourcePublisher
 3243      qtp1312786476-84
 3134      main
 3047      OH-thingHandler-24
 2988      OH-thingHandler-34
 2976      OH-thingHandler-4
 2950      OH-thingHandler-8
 2926      OH-thingHandler-37
 2921      OH-thingHandler-48
 2914      OH-thingHandler-47
 2894      OH-thingHandler-9
 2875      OH-thingHandler-3
```

**Test 2 (test2+861-763)**
- without #763
- with #861
```
commit 351ba579df7915324185e6b1266a1149a479a909 (HEAD -> test2+861-763)
Date:   Mon Jun 17 20:55:10 2019 -0400
    Revert "safe caller: respect classloader of the class and the interfaces (#763)"
    This reverts commit 25d7f14dfcc5f93c316cb64c75bd306721e7a36e.

commit e75a45120639910cd4bd954f4d5e129ede10f38e
Merge: 0f1c1bbb e8758260
Date:   Mon Jun 17 20:54:50 2019 -0400
    Merge commit 'refs/pull/861/head' of https://github.com/openhab/openhab-core into test2+861-763

commit 0f1c1bbb2dc03a0e3bb004ec081fdf7569253cd1 (origin/master, origin/HEAD, master)
Date:   Thu Jun 13 20:03:16 2019 +0200
```
```
$ ls -l org.openhab.core-2.5.0-SNAPSHOT-test2.jar
-rw-rw-r-- 1 mark mark 420180 Jun 17 20:59 org.openhab.core-2.5.0-SNAPSHOT-test2.jar

$ zipgrep "Bundle-Version" org.openhab.core-2.5.0-SNAPSHOT-test2.jar 
META-INF/MANIFEST.MF:Bundle-Version: 2.5.0.201906180056

222 │ Active │  80 │ 2.5.0.201906180056    │ org.openhab.core
```

*CPU Utilization*
```
09:56:11 AM  CPU    %usr   %nice    %sys %iowait    %irq   %soft  %steal  %guest  %gnice   %idle
09:56:13 AM  all   36.32    0.00    6.65    0.00    0.00    1.28    0.00    0.00    0.00   55.75
09:56:13 AM    0   35.71    0.00    7.14    0.00    0.00    0.00    0.00    0.00    0.00   57.14
09:56:13 AM    1   35.57    0.00    6.19    0.00    0.00    0.00    0.00    0.00    0.00   58.25
09:56:13 AM    2   36.55    0.00    7.11    0.00    0.00    0.00    0.00    0.00    0.00   56.35
09:56:13 AM    3   37.82    0.00    5.70    0.00    0.00    5.18    0.00    0.00    0.00   51.30
```
*Top 30 CPU Consuming Threads*
```
 350633    OH-EventHandlerExecutor-1
 208356    pool-5-thread-1
 104460    OH-OSGiEventManager
 63731     OH-EventHandlerWatcher-1
 42274     EventAdminAsyncThread #12
 41943     EventAdminAsyncThread #11
 15612     Start Level: Equinox Container: 9527c48e-cb0d-4fe5-84d4-a76efd03fff9
 11921     items-5
 11618     items-1
 11584     items-4
 11430     items-3
 11348     items-2
 8267      qtp139353611-82
 8251      qtp139353611-90
 8065      qtp139353611-84
 8036      qtp139353611-208
 7897      qtp139353611-85
 7172      qtp139353611-204
 6332      qtp139353611-205
 3180      ResourcePublisher
 3054      main
 2931      OH-thingHandler-5
 2920      OH-thingHandler-4
 2900      OH-thingHandler-31
 2896      OH-thingHandler-32
 2895      OH-thingHandler-12
 2891      OH-thingHandler-42
 2861      OH-thingHandler-35
 2858      OH-thingHandler-40
 2858      OH-thingHandler-21
```

mhilbush: BTW, if you're wondering why I didn't run the current build. I did run it. With 4 load generators (versus 24 in the other tests) producing 4x57 events per second, it wouldn't process above 40 events per second and the ESH-OSGiEventManager CPU consumption was 502185 after just a couple minutes of run time.

maggu2810: Thank you @mhilbush again for your tests.

> @kaikreuzer @maggu2810 Here are the results of my testing. Let me know if there's anything else you think I should test.

Not from my side.
I don't like the extensive usage of the safe caller because it is based on the creation of proxies.
I also don't think we should consider to revert #763 as this PR fixes an IMHO wrong usage of the classloading in OSGi.
This change has already been okay from my side on its creation.
Sure, the question about the number of used threads is still open (see https://github.com/openhab/openhab-core/pull/861#discussion_r292013308), can be discussed and changed (easily).
But I don't see any reason why we don't get it merged.

I don't know if the argument about "one addon can block others" is really valid at the moment. As your tests shown (multiple times already) the current event handling is MUCH slower without this PR. It seems for me that also using a single thread it will be much faster then before.
So why waiting? Can it make the current situation more bad then it is?

J-N-K: I‘m not a core expert, but so agree with @maggu2810.

kaikreuzer: @mhilbush, thanks again for the excessive tests, this is very helpful!

@maggu2810 & @J-N-K: I am not against making anything better and won't veto a merge. I just wanted to make sure that we are not semi-knowingly accept any bigger regressions with this.

The one issue that I wanted to have clarified is @mhilbush's [comment](https://github.com/openhab/openhab-core/pull/861#discussion_r292974031): "So, to compare max performance, it looks like throughput has degraded by 68% with the addition of #861 and #763."
I read this that with the latest tests, the baseline should have been much better than the Test 1 wrt throughput. So if the throughput is twice as much, I would claim that it is fully ok to also have a bit higher CPU load. Are we really comparing the same things here? From a user pov, throughput is much more important than CPU load, isn't it?

The second issue was my concern about changed behavior due to the single-threadedness. I remember that multithreading was introduced for a good reason, so there's a high chance for a regression that we are currently not seeing. It seems that at least @wborn [also vaguely remembers](https://github.com/openhab/openhab-core/pull/861#discussion_r292026168) that there were issues, so I think it isn't just a wrong memory from me. Sure, this is vague and won't bring us forward, so I can live with potentially finding out what the reason at the time was by merging this PR.

J-N-K: As far as I understood the latest test was 1370 events/second. My guess would be that this is sufficient, even if the multi-threaded version would probably be able to handle even higher numbers. If it’s really 90 vs 280 events/second we should think again. My feeling says that 90 events/second can be easily reached even with medium-sized systems.

mhilbush: > The one issue that I wanted to have clarified is @mhilbush's comment: "So, to compare max performance, it looks like throughput has degraded by 68% with the addition of #861 and #763."

I'd suggest we disregard tests other than the most recent. It was shortly after those tests that my test box crashed, so I'm considering those earlier tests suspect.

> I read this that with the latest tests, the baseline should have been much better than the Test 1 wrt throughput.

Again, I'm considering those earlier tests suspect. In the most recent set of tests, all three tests processed the same rate of events, but with Test 1 and Test 2 showing considerably lower CPU utilization.

> The second issue was my concern about changed behavior due to the single-threadedness.

This is what don't fully understand. I didn't see any evidence of multithreaded behavior in any of the tests. Wouldn't that have been revealed in the Karaf thread dump? If not, where would we see evidence of multithreaded behavior in the the Baseline?

kaikreuzer: > I'd suggest we disregard tests other than the most recent.

Ok, disregarded 😄.

 > where would we see evidence of multithreaded behavior in the the Baseline?

Well, my memory is that it was multi-threaded at least for different bindings. So if you had two load-test bindings, you should see two threads serving the events to them.
But maybe let's also wait for @wborn to comment as he has an extensive LIFX setup and wants them to be switched synchronously - maybe he can check, whether #861 makes any difference in behavior for his setup.

maggu2810:
@mhilbush To make it clear for us, can you fill such a table WRT your current machine?

Is my understanding correct, that you state the test makes no sense against the current master branch, because it is such slow?

| #763        | #861           | notes | events / seconds | cpu load |
|:-------------:|:-------------:| ----- | ----- | ----- |
| no | no | some time in the past as we noticed that the proxies using wrong classloading | | |
| no | yes | -- | | |
| yes | no | current master | very very rare events per second, see [here](https://github.com/openhab/openhab-core/pull/861#discussion_r295371424) | |
| yes | yes | master if this one gets merged | | |

mhilbush:
> Is my understanding correct, that you state the test makes no sense against the current master branch, because it is such slow?

Yes, the current master branch is too slow to run anything close to the volume I ran against Baseline, Test 1, and Test 2.

> can you fill such a table WRT your current machine?

I was just thinking about doing a table like this. :wink: Working on it now.

maggu2810:
Please reload and review, I edit the table already...

mhilbush:
Can you update the ID column with the following so that the rows map to the tests above? Otherwise, you've captured it correctly.

1 => Baseline
2 => Test 2
3 => Current Master
4 => Test 1

maggu2810:
I assume the naming "baseline" is very confusing. So I preferred not choosing a special name at all, but an ID, and a description, and a truth table. :wink: 
But if you are more happy, I will add another table to refer the naming in your comment above

mhilbush:
You can leave it as is. I used the term Baseline to reflect the performance of the system prior to PRs #763 and #861.

bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/internal/events/EventHandler.java
wborn: I had a look at the code and the changes in this PR don't impact how commands are sent to `ThingHandler`s using the `SafeCaller`: 

https://github.com/openhab/openhab-core/blob/e8758260f18215305b815a35fb72700c40465118/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/ProfileCallbackImpl.java#L71-L75

So commands will still be handled asynchronously which was my main concern. Commands will thus continue to be handled even if a couple of `ThingHandler`s hijack threads e.g. due to network timeouts or when waiting on locks.

mhilbush: FTR this is how I generated the event load.

https://github.com/mhilbush/openhab2-addons/tree/load-generator-binding/bundles/org.openhab.binding.loadgenerator

kaikreuzer: I was just about to ask @mhilbush to add some command handling to the load generator (with e.g. 300ms processing time of the handler).
But @wborn's observation that this had actually moved to the ProfileCallback should hopefully answer it - this should not be related to the event dispatching itself anymore. I consider my second concern to be clarified then.

Thank you all for the hard work and your patience with me!

kaikreuzer:
Thank you all for the hard work and your patience with me!

bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/internal/events/EventHandler.java
mhilbush: > Thank you all for the hard work and your patience with me!

No worries. It's a change to a critical area, so I get the need for diligence.

> I was just about to ask @mhilbush to add some command handling to the load generator (with e.g. 300ms processing time of the handler).

I wrote the load generator in a hurry, so there's room for improvement. As I plan to use it in the future, and possibly extend it, what were you looking for here?

kaikreuzer: > what were you looking for here?

A feature that makes the load generator not only create status updates, but also commands and then have the `handleCommand` method in the ThingHandler to block for e.g. 300ms to simulate some heavy work (like accessing external devices). This would then better simulate a setup where users can directly observe whether there is a delay in event processing or not - if switching on a device suddenly takes a few seconds, because the command hang in a single-threaded queue, it is instantly visible and annoying to the user. But as mentioned above: This PR should not have an issue on this behavior, so such an addition would only be interesting for future tests in other contexts.

PR #862, opened 2019-06-09
small improvements of hexutils API and impl.

maggu2810:
* improve HexUtils; e.g. don't create Byte array just to copy its content to a byte array
* migrate the HexUtils integration test to a normal JUnit test

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/util/HexUtils.java
+++ b/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/util/HexUtils.java
@@ -12,8 +12,7 @@
  */
 package org.eclipse.smarthome.core.util;
 
-import java.util.Arrays;
-import java.util.stream.Collectors;
+import java.nio.charset.StandardCharsets;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core/src/test/java/org/eclipse/smarthome/core/util/HexUtilsTest.java
```diff
@@ -1,3 +1,4 @@
+package org.eclipse.smarthome.core.util;
```
wborn: Your IDE probably tried to "help" you with moving this package declaration above the header. Can you move it back below the copyright header?

maggu2810: done

bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/util/HexUtils.java
```diff
@@ -46,10 +46,25 @@ private HexUtils() {
...
+        }
+        final StringBuilder sb = new StringBuilder(bytes.length * 2 + delimiter.length() * (bytes.length - 1));
+        int pos = 0;
+        while (pos < bytes.length - 1) {
+            final byte[] hex = byteToHex(bytes[pos++]);
+            sb.append((char) hex[0]);
+            sb.append((char) hex[1]);
+            sb.append(delimiter);
+        }
+        final byte[] hex = byteToHex(bytes[pos++]);
```
wborn: It seems unnecessary to increase `pos` here after using it: 

```suggestion
        final byte[] hex = byteToHex(bytes[pos]);
```

maggu2810: You are absolutely right, it has been done this way to make it visible that it is the same code as in the while loop except the delimiter.
Sure, we don't need to increment `pos` here, but I don't think that this instruction harm. :wink:

PR #864, opened 2019-06-10
UID: allow to encode arbitrary string (+decode)

maggu2810:
Related to: #847, #848


maggu2810:
I will move the static functions (encode and decode) to an UIDUtils class.

maggu2810:
@davidgraeff Is this still useful for you?

davidgraeff:
Yes, it will be used to deduct thing and channel ids based on MQTT topics.

maggu2810:
ping @openhab/core-maintainers

PR #867, opened 2019-06-12
Sort serial port parameter options

wborn:
When you have many serial ports its easier to find them in a sorted list.

### DIFF
--- a/bundles/org.openhab.core.config.serial/src/main/java/org/eclipse/smarthome/config/serial/internal/SerialConfigOptionProvider.java
+++ b/bundles/org.openhab.core.config.serial/src/main/java/org/eclipse/smarthome/config/serial/internal/SerialConfigOptionProvider.java
@@ -12,10 +12,11 @@
  */
 package org.eclipse.smarthome.config.serial.internal;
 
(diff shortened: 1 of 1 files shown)


maggu2810:
For me this seems to be a mixture of the UI layer and the business logic.

If an user interface would like to provide the options sorted (something I would expect and really prefer), the UI should sort the provided list.

I don't think non UI related code should be changed to use a strict order just because an UI would like to use that order.

If we don't want to adapt several UIs, can't we add a sort / comparation logic in the REST interface?

wborn:
Yes you're right about that this is about presentation and it should be in the presentation layer. I'll create a Paper UI PR for this instead.

> If we don't want to adapt several UIs, can't we add a sort / comparation logic in the REST interface?

Adding a parameter for sorting results would make life a bit easier for API consumers. Though when parts of a UI are lazy loaded it may only want to sort content at the moment it is actually shown. So if it were added, the sorting should be optional.

wborn:
There's now https://github.com/openhab/openhab-webui/pull/80 to get these lists of serial ports sorted. :-)

maggu2810:
Thanks

PR #870, opened 2019-06-12
[ephemeris] upgrade jollyday library to the latest released version

hakan42:
Signed-off-by: Hakan Tandogan <hakan@tandogan.com>

Updating to 0.5.8 so the turkish holidays are available, too 😄


hakan42:
please rebuild, Mr. Travis

PR #873, opened 2019-06-20
don't drop the itests projects from reactor on skipTests

maggu2810:
The bnd-testing-maven-plugin that is used to run the tests already checks for the skipTests property and do not execute any tests if defined.
We should keep the integration tests part of the reactor also if the tests are not executed.


maggu2810:
Same should be done for e.g. addons (I will create the respective PR later).

PR #877, opened 2019-06-20
we don't know if the rule has been deleted async.

maggu2810:


maggu2810:
Travis CI:
* Java 8: failed because of unrelated `Configuration Discovery Tests`
* Java 11: succeeded

PR #878, opened 2019-06-20
handle nullness in unget handler of composite module correctly

maggu2810:


maggu2810:
Travis CI:
* Java 8: succeeded
* Java 11: failed because of unrelated `Configuration Discovery Tests`

PR #904, opened 2019-06-28
add launch configuration to generate the model sources

maggu2810:
This launch configuration should be executed in the distros Oomph setup
as a one time job on first installation (I assume something like
`excludedTriggers="STARTUP MANUAL"`).

If it is executed by as setup instsructions no parallel auto build will
be triggered. If (for testing) executed inside a running IDE manually,
the automatic build should be disabled and a refresh should be triggered
in front of enabling it again.

If you integrate that in the Oomph Distro setup, I assume you can use
the core setup file (not used by others then me I assume) as a template
(and remove it now).

I assume it can be placed after the import task of core and beore a
build task of all projects.

```xml
    <setupTask
        xsi:type="projects:ProjectsImportTask">
      <sourceLocator
          rootFolder="${git.clone.openhabcore.location}"
          locateNestedProjects="true"/>
    </setupTask>
    <setupTask
        xsi:type="launching:LaunchTask"
        id="launch.model.gen"
        excludedTriggers="STARTUP"
        launcher="ModelGen"/>
    <setupTask
        xsi:type="projects:ProjectsBuildTask"
        excludedTriggers="STARTUP"
        refresh="true"
        clean="true"/>
    <setupTask
        xsi:type="setup:ResourceCreationTask"
        id="lifecycle-mapping"
```

Related to: https://github.com/openhab/openhab-distro/issues/927

### DIFF
--- a/tools/oomph/ModelGen.launch
+++ b/tools/oomph/ModelGen.launch
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<launchConfiguration type="org.eclipse.m2e.Maven2LaunchConfigurationType">
+<booleanAttribute key="M2_DEBUG_OUTPUT" value="false"/>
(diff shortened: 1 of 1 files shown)


kaikreuzer:
Thanks, that looks good. I think we should probably get rid off the "tools/oomph" folder from the repo - but we would then need a new location for this launch file. Would it make sense to put it into the `org.openhab.core.model.core` bundle directly?

maggu2810:
As it is about openhab core launch config and not only a model core one, would `tools/launch` directory a good idea?
It can be a place for Eclipse IDE launch configurations and simple to cleanup...

kaikreuzer:
I would want to avoid to have that many different projects with just 1-2 launch files in them. We already have one in the distro (which is meant to be the central one for the IDE setup) and we will have the same question for the webui repo as well. So I'd prefer to either have them all in the distro/launch folder (as they are executed by the setup file of the distro, this would imho be ok) or to put them within the projects that they deal with (ok to put it to openhab.core itself in this case).

maggu2810:
> I would want to avoid to have that many different projects with just 1-2 launch files in them. 

I absolutely agree, that also one of the reasons I don't like to add that specific launch configuration to the model.core project but use one directory in tue whole repository for all launch configuration.

Okay for me if you want to use one project for the whole distribution as you would like to support only that IDE setup (which makes sense for the openHAB community).
So, migrate that launch configuration to openhab-distro to integrate its call into the setup file.
No need to add it to that repo.

kaikreuzer:
Thanks - I have moved this as discussed to https://github.com/openhab/openhab-distro/pull/935.

PR #906, opened 2019-06-28
Do not create a second instance of the StateDescriptionFragment on build

cweitkamp:
- Do not create a second instance of the `StateDescriptionFragment` on build but return the internal instance like we do it in other builders too

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/types/StateDescriptionFragmentBuilder.java
+++ b/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/types/StateDescriptionFragmentBuilder.java
@@ -62,7 +62,7 @@ public class StateDescriptionFragmentBuilder {
      * @return a {@link StateDescriptionFragment} from the values of this builder.
      */
     public StateDescriptionFragment build() {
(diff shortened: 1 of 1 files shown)


cweitkamp:
Now I am confused. Travis failed and I have the feeling that he is right. A builder should return immutable instances. Right?

This is not the case for the `ThingBuilder`.

https://github.com/openhab/openhab-core/blob/01f472457524d41eb5b1006a019f72aafaba1fc3/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/builder/ThingBuilder.java#L127-L131

And a second question: Why does the builder use an internal `StateDescriptionFragmentImpl` property instead of the dedicated properties which will be assigned to the resulting object.

maggu2810:
I don't think a builder needs to be used for immutable objects only.
The instances created by the builder can be mutable (so contain e.g. set methods).

But (and I assume this is what you are thinking about) IMHO the build method of a builder should return a newly created object.
If I call build three times, I would like to get three different instances of the respective class.
If I call build and then call another method of the builder that changes an internal member this must not change my already built object.

I agree that the builder should use properties on its own and not the same state description fragment impl. as returned.

PR #910, opened 2019-07-03
[automation] Added nullness annotations

cweitkamp:
- Added nullness annotations

Therefore introduced two new classes `BaseActionModuleHandler` and `BaseConditionModuleHandler`. A [BaseTriggerModuleHandler](https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/handler/BaseTriggerModuleHandler.java) is already present.

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/AbstractScriptModuleHandler.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/AbstractScriptModuleHandler.java
@@ -21,6 +21,8 @@ import java.util.UUID;
 import javax.script.ScriptContext;
 import javax.script.ScriptEngine;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.automation.Module;
 import org.openhab.core.automation.handler.BaseModuleHandler;
 import org.openhab.core.automation.module.script.ScriptEngineContainer;
@@ -36,7 +38,9 @@ import org.slf4j.LoggerFactory;
  *
  * @param <T> the type of module the concrete handler can handle
  */
+@NonNullByDefault
 public abstract class AbstractScriptModuleHandler<T extends Module> extends BaseModuleHandler<T> {
+
     private final Logger logger = LoggerFactory.getLogger(this.getClass());
 
     /** Constant defining the configuration parameter of modules that specifies the mime type of a script */
@@ -45,13 +49,13 @@ public abstract class AbstractScriptModuleHandler<T extends Module> extends Base
     /** Constant defining the configuration parameter of modules that specifies the script itself */
     protected static final String SCRIPT = "script";
 
-    protected ScriptEngineManager scriptEngineManager;
+    protected final ScriptEngineManager scriptEngineManager;
 
     private final String engineIdentifier;
 
     private Optional<ScriptEngine> scriptEngine = Optional.empty();
-    private String type;
-    protected String script;
+    private @NonNullByDefault({}) String type;
+    protected @NonNullByDefault({}) String script;
 
     private final String ruleUID;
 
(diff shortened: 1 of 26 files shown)


bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/AbstractScriptModuleHandler.java
```diff
@@ -45,30 +49,28 @@
...
-    protected ScriptEngineManager scriptEngineManager;
+    protected final ScriptEngineManager scriptEngineManager;
 
     private final String engineIdentifier;
 
     private Optional<ScriptEngine> scriptEngine = Optional.empty();
-    private String type;
-    protected String script;
+    private @NonNullByDefault({}) String type;
+    protected @NonNullByDefault({}) String script;
```
maggu2810: We introduced the usage of `@NonNullByDefault({})` for variables we know that they are not null but this cannot be detected by the tooling. One common example has been references of components that are injected by the declarative service implementation on activation.
For this use case there is no need anymore because every mandatory reference can now be set by using constructor injection.
For optional references, we should already use nullable because they can be valid null.
Another use case are members that are injected by reflection because of specific deserialization (e.g. Gson).

Why do we need it in this case.
Wouldn't it be correct to mark these variable `@Nullable`?

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/AbstractScriptModuleHandler.java
```diff
@@ -91,16 +93,18 @@ private void loadConfig() {
         Object type = module.getConfiguration().get(SCRIPT_TYPE);
         Object script = module.getConfiguration().get(SCRIPT);
         if (!isValid(type)) {
-            throw new IllegalStateException(String.format("Type is missing in the configuration of module '%s'.", module.getId()));
+            throw new IllegalStateException(
```
maggu2810: You could add a method `String resolveValidType(final Object type)` that throws a ISE if the type is not valid and otherwise return a valid type.

cweitkamp: Thanks for the idea. I introduces a new method to get a valid config and assign it to the properties inside the contructor. This solution solves your other https://github.com/openhab/openhab-core/pull/910#discussion_r300031597 too and made it possible to make the properties final.

maggu2810:
@cweitkamp Did you see that errors?

| Description | Resource | Path | Location | Type |
| --- | --- | --- | --- | --- |
| Null constraint mismatch: The type 'Action' is not a valid substitute for the type parameter 'T extends @NonNull Module' | WebPushNotificationActionHandler.java | /org.openhab.ui.habot/src/main/java/org/openhab/ui/habot/notification | line 36 | Java Problem |
| Null type mismatch: required '@NonNull Map<String,Object>' but the provided value is null | DayOfWeekConditionHandlerTest.java | /org.openhab.core.automation.module.timer.tests/src/main/java/org/openhab/core/automation/module/timer/internal | line 56 | Java Problem |
| Null type mismatch: required '@NonNull Map<String,Object>' but the provided value is null | DayOfWeekConditionHandlerTest.java | /org.openhab.core.automation.module.timer.tests/src/main/java/org/openhab/core/automation/module/timer/internal | line 62 | Java Problem |
| Null type mismatch: required '@NonNull Map<String,Object>' but the provided value is null | DayOfWeekConditionHandlerTest.java | /org.openhab.core.automation.module.timer.tests/src/main/java/org/openhab/core/automation/module/timer/internal | line 69 | Java Problem |
| Null type mismatch: required '@NonNull Map<String,Object>' but the provided value is null | TimeOfDayConditionHandlerTest.java | /org.openhab.core.automation.module.timer.tests/src/main/java/org/openhab/core/automation/module/timer/internal | line 55 | Java Problem |
| Null type mismatch: required '@NonNull Map<String,Object>' but the provided value is null | TimeOfDayConditionHandlerTest.java | /org.openhab.core.automation.module.timer.tests/src/main/java/org/openhab/core/automation/module/timer/internal | line 59 | Java Problem |
| Null type mismatch: required '@NonNull Map<String,Object>' but the provided value is null | TimeOfDayConditionHandlerTest.java | /org.openhab.core.automation.module.timer.tests/src/main/java/org/openhab/core/automation/module/timer/internal | line 63 | Java Problem |
| Null type mismatch: required '@NonNull Map<String,Object>' but the provided value is null | TimeOfDayConditionHandlerTest.java | /org.openhab.core.automation.module.timer.tests/src/main/java/org/openhab/core/automation/module/timer/internal | line 68 | Java Problem |
| Null type mismatch: required '@NonNull Map<String,Object>' but the provided value is null | TimeOfDayConditionHandlerTest.java | /org.openhab.core.automation.module.timer.tests/src/main/java/org/openhab/core/automation/module/timer/internal | line 74 | Java Problem |

maggu2810:
The above errors will be fixed by:

* https://github.com/openhab/openhab-webui/pull/85
* https://github.com/openhab/openhab-core/pull/913

cweitkamp:
@maggu2810 I might have missed them because the itests for automation are still not enabled by default. Thanks for taking care.

PR #911, opened 2019-07-03
[automation] Added 'ThingStatusTrigger' for NGRE

cweitkamp:
- Added `ThingStatusTrigger` for NGRE

Currently I only check for the `ThingStatus.` But IIRC the trigger in DSL rules additionaly compares the `ThingStatusDetail` and the description. Just want to know right now if this is the way how it has to be done.

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ThingStatusTriggerHandler.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ThingStatusTriggerHandler.java
@@ -1,0 +1,146 @@
+/**
+ * Copyright (c) 2010-2019 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.automation.internal.module.handler;
+
+import java.util.Collections;
+import java.util.Dictionary;
+import java.util.HashMap;
+import java.util.Hashtable;
+import java.util.Map;
+import java.util.Set;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.smarthome.core.events.Event;
+import org.eclipse.smarthome.core.events.EventFilter;
+import org.eclipse.smarthome.core.events.EventSubscriber;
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ThingStatusTriggerHandler.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    @Override
+    public @Nullable EventFilter getEventFilter() {
+        return this;
+    }
+
+    @Override
+    public void receive(Event event) {
+        if (callback != null) {
```
maggu2810: Just a very minor cleanup:

```java
final ModuleHandlerCallback callback = this.callback;
if (!(callback instanceof TriggerHandlerCallback)) {
    return;
}

TriggerHandlerCallback thCallback = (TriggerHandlerCallback)
logger.trace("Received Event: Source: {} Topic: {} Type: {}  Payload: {}", event.getSource(),
        event.getTopic(), event.getType(), event.getPayload());
...
if (!values.isEmpty()) {
    values.put(OUT_EVENT, event);
    thCallback.triggered(module, values);
}
```

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ThingStatusTriggerHandler.java
```diff
@@ -0,0 +1,146 @@
...
+        return reqState.isEmpty() || reqState.equals(status.toString());
+    }
+
+    /**
+     * do the cleanup: unregistering eventSubscriber...
+     */
+    @Override
+    public void dispose() {
+        super.dispose();
+        eventSubscriberRegistration.unregister();
```
maggu2810: Also a very minor comment:

Shouldn't we do our "more specific" dispose logic apply in front of the dispose logic of the base class.
That's more a generic lifecycle handling (similar to constructor and destructor (if available) or activate, deactivate, ...)

1. init base
1. init specific
1. ...
1. deinit specific
1. deinit base

maggu2810:
@cweitkamp If the title is changed no mail is sent. So, I have not been informed that the WIP has been removed. It has been more or less coincidence that I realised it.

cweitkamp:
@maggu2810 Ah, sorry. I can remember, that you told me before, but I forgot. Next time I will send you a notification.

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ThingStatusTriggerHandler.java
cweitkamp: Done.

maggu2810:
@kaikreuzer Feel free to merge if you agree.

PR #684, opened 2019-03-30
provide Java EE modules 

J-N-K:
This provides easy dependencies for JAXB and JAX-WS (which were removed in JDK 11).

see: https://github.com/openhab/openhab2-addons/pull/5327

PR #920, opened 2019-07-11
provide Java EE modules

maggu2810:
Replaces: https://github.com/openhab/openhab-core/pull/684

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -867,6 +867,35 @@
       <version>0.5.8</version>
     </dependency>
 
+    <!-- JAXB -->
+    <dependency>
+      <groupId>org.apache.servicemix.specs</groupId>
+      <artifactId>org.apache.servicemix.specs.activation-api-1.1</artifactId>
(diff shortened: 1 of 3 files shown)


bom/runtime/pom.xml
```diff
@@ -867,6 +867,35 @@
       <version>0.5.8</version>
     </dependency>
 
+    <!-- JAXB -->
+    <dependency>
```
cweitkamp: It looks like we now included some of these dependency twice:

https://github.com/openhab/openhab-core/blob/b6aae6907f3d247e5d590bb25f6cb009e17584f8/bom/runtime/pom.xml#L135-L141

https://github.com/openhab/openhab-core/blob/b6aae6907f3d247e5d590bb25f6cb009e17584f8/bom/runtime/pom.xml#L127-L133

https://github.com/openhab/openhab-core/blob/b6aae6907f3d247e5d590bb25f6cb009e17584f8/bom/runtime/pom.xml#L159-L165

Issue #918, opened 2019-07-10
[REST API] REST API Docs UI cannot find swagger.json


rkoshak:
On recent snapshots (build 1635) the REST API Docs will no longer load complaining about an inability to find swagger.json.

The error on the UI states:

`Can't read swagger JSON from https://<address of OH>/rest/swagger.json`

When I try to load that file directly I get:

`{"error":{"message":"HTTP 404 Not Found","http-code":404,"exception":{"class":"javax.ws.rs.NotFoundException","message":"HTTP 404 Not Found","localized-message":"HTTP 404 Not Found"}}}`

I opened an issue over on the webuis repo but it looks like to root of the problem is here. It has been reported by at least two other users on the forum.

There are no errors in the logs. This error prevents the REST API Docs UI from loading.

maggu2810:
I can confirm and assume it is related to the publisher changes.

RobertHerter:
Same for me. Reported in webui-git a while ago.

sabefix:
Same problem already with build 1631.

lolodomo:
REST documentation is working in snapshot 1641.

PR #905, opened 2019-06-28
openHAB JAX-RS publisher

maggu2810:
Do not use the `com.eclipsesource.jaxrs` `publisher` anymore.
It's service tracker breaks our OSGi runtime a little bit.

There has been already:
https://github.com/eclipse/smarthome/pull/4153

This PR uses DS instead of a service tracker, this ensure the consumed services are already ready.

It drops the option to add a whitelist as fragment of the io.rest.optimize bundle, but I am pretty sure this is not used in openHAB anyhwere (IIRC it has been added to allow third party products to use a different whitelist, but I was not aware of anyone ever used it).

Perhaps it will also help us with a more step by step transition to the JAX-RS Whiteboard pattern as a positive side effect. If not, it does not harm.

The used EclipseSource Provider sources already uses EPL but v1.
I kept the origin license headers and copied a notice file from another bundle.
Perhaps this needs be fixed.

Using that publisher I do not see any CNFE or other exceptions in the demo application anymore.

PR #928, opened 2019-07-16
Reverts https://github.com/openhab/openhab-core/pull/905

kaikreuzer:
See https://github.com/openhab/openhab-core/issues/924#issuecomment-511787620

Fixes https://github.com/openhab/openhab-core/issues/924
Fixes https://github.com/openhab/openhab-core/issues/918


kaikreuzer:
I used Gits "reverse commit" feature, which does not add any sign-offs (as they don't make much sense since the commit does not include any copyrightable content from me).

kaikreuzer:
As it is approved, I'll merge it together with https://github.com/openhab/openhab-distro/pull/944.

maggu2810:
> I used Gits "reverse commit" feature, which does not add any sign-offs (as they don't make much sense since the commit does not include any copyrightable content from me).

Does it matter?
The git revert command supports the `-s, --signoff` option similar to the commit option.
And git revert is just some tooling to create a commit and commits should be signed off.

IMHO

Code has been removed by a commit and code will be added again by a revert commit.
The sign off should be used to state "I am allowed" to add code to the current code base.
Assume a commit removes code because the license has been not clear. If you revert the removal you should still state that you know what you are doing.

kaikreuzer:
Ok, sorry then. I used the revert feature of SourceTree and this has no sign-off option. Will try to do it with `-s` on the command line in future instead.

PR #929, opened 2019-07-18
change mqtt client from Paho to HiveMQ

J-N-K:
This changes the internal code for the mqtt transport from Paho to HiveMQ.

Reasons:
- Paho/Java is missing MQTTv5 support (has recently been removed from the code), HiveMQ does (not supported at the moment in openHAB code, but can be added)
- Paho is not thread-safe in all cases (especially publish). There have been reports of missing messages
- According to @davidgraeff's analysis Paho has some memory leaks
- The HiveMQ API is more modern and allows easier code (and extensions)
- The current Paho client sometimes loses subscriptions, reason unknown


kaikreuzer:
Could you elaborate in the PR description why the implementation should be changed?
What's wrong/bad about Paho, what benefits does HiveMQ bring that are worth any potential regressions that we might be seeing by such a change?

maggu2810:
A short summary would be nice about the cons for Paho MQTT and the pro for HiveMQT.
IIRC @davidgraeff already complained about the Paho MQTT implementation (memory leaks, delayed MQTT v5.0 support, bugs, ...) already but perhaps we could add the information to the commit message, so this information stays in the history and is part of the change docu itself.

J-N-K:
It seems to be very difficult to implement the "accept-all-certificates" solution that was used before. It is possible to supply a custom `TrustManagerFactory`, but I unfortunately it is not possible to provide a custom `TustManager` to a `TrustManagerFactory` that accepts all certificates. The reason is that the `getTrustManagers` method is final and can't be overridden.

IMO such a behaviour is faulty anyway. Can't we add a binding config that adds trusted certs? Accepting self-signed certs should be easy, it's only "accept all" that's difficult.

cweitkamp:
We have a [TrustAllTrustMananger](https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.io.net/src/main/java/org/eclipse/smarthome/io/net/http/TrustAllTrustMananger.java), does it help?

J-N-K:
Unfortunately not, I need a `TrustManagerFactory` and I didn’t manage to create that one because I can‘t override the final method.

cweitkamp:
Reading a little bit deeper into the `javax.net.ssl` library it looks like you can implement a custom `TrustManagerFactory` by supplying your own service provider interface (`TrustManagerFactorySPI`) which is responsible to hand over the `TrustManager` to the factory.

Maybe looking into the [code of other libraries](https://github.com/netty/netty/blob/00afb19d7a37de21b35ce4f6cb3fa7f74809f2ab/handler/src/main/java/io/netty/handler/ssl/util/InsecureTrustManagerFactory.java) will give you a hint how to proceed.

J-N-K:
Since HiveMQ uses netty anyway, we can probably use that directly. 

It's easy now to add MQTT5 support, but how should that be configured? Adding add additional parameters to the constructor or better have a setter/getter and use v3 as default?

davidgraeff:
I haven't looked into the hivemq startup, but if they support protocol negotiation I'd say that v3/v5 should be auto negotiated. A developer shouldn't care at this point in time, because our wrapping library does not yet expose any v5 only features.

J-N-K:
@davidgraeff unfortunately that's not possible, it is required to decide between v5 and v3 when creating the client. What would be possible is trying to create a v5 connection and fallback to v3 if that fails.

J-N-K:
The good news: I tested the new bundle on a pi today with unchanged bindings with a mosquito broker connection and it worked. Needs some more testing though.

J-N-K:
This works for me in my test setup. 

MQTTv3 and MQTTv5 clients are implemented, but currently always a v3 client is created. IMO we could either add a new getter/setter-pair for the version or add an additional parameter to the constructor. Both can be made backward-compatible.

davidgraeff:
A setter implies to the API consumer that the mqtt version could be changed on the fly, which is of course not the case. I'd say the constructor extension is the way to go. WDYT?

J-N-K:
The same holds true for most of the other getter/setter-methods. They need re-connection.

davidgraeff:
Getters are not an issue. I think there are no other setters that require a reconnection are there? At least there shouldn't.

J-N-K:
> 
> 
> Getters are not an issue. I think there are no other setters that require a reconnection are there? At least there shouldn't.

e.g. `setLastWill`

J-N-K:
Is there anything I can do to proceed here?

cweitkamp:
I have to admit that I did not follow this PR. I can dive into it but it would be nice to at least have a green build. Currently there are conflicts.

wborn:
Sorry if I caused these conflicts with my cleanup PRs! I'm done cleaning the repo now (except for a few minor SAT issues). So hopefully this will be the last time you have to resolve conflicts. 

I've quickly scanned the comments and it looks like this PR is the next step forward to resolve the open MQTT issues. The only known regression would be the TrustManagerFactory issues?

I'd personally rather first see the next milestone to be released before this is merged. That would  give us some time to fix any issues this PR might cause before they hit milestone users. Perhaps @kaikreuzer already has some plans for a M4? M3 was made a bit more than one month ago (September 7th).

There's also a [Karaf 4.2.7](https://issues.apache.org/jira/secure/ReleaseNote.jspa?projectId=12311140&version=12345539) we might want to upgrade to after creating the M4. :-)

kaikreuzer:
> Perhaps @kaikreuzer already has some plans for a M4? M3 was made a bit more than one month ago (September 7th).

I'm just sitting at the airport on my way back home from holidays - so yes, once back, I think we should head for a M4 within the next days.

J-N-K:
Too much work. Not worth the effort. I can't even rebase on master because of conflicts in other bundles.

wborn:
How did you rebase?

If I first squash your 5 commits (`git rebase -i HEAD~5`) and then rebase them on the current master (`git rebase origin/master`) there are only a few conflicts (or am I missing something?):


```
Unmerged paths:
  (use "git reset HEAD <file>..." to unstage)
  (use "git add <file>..." to mark resolution)

	added by us:     bom/openhab-core-index/.classpath
	added by us:     bom/openhab-core-index/.project
	added by us:     bundles/org.openhab.core.audio/NOTICE
	both modified:   bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
	added by us:     bundles/org.openhab.core.io.transport.mqtt/src/test/java/org/eclipse/smarthome/io/transport/mqtt/MqttAsyncClientEx.java
	added by us:     bundles/org.openhab.core.io.transport.mqtt/src/test/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnectionEx.java
	added by us:     itests/org.openhab.core.config.core.tests/pom.xml
	added by us:     itests/org.openhab.core.model.persistence.tests/itest.bndrun
	both modified:   itests/pom.xml
```

J-N-K:
```

$ git rebase upstream/master                                                                                                                                                                            error: The following untracked working tree files would be overwritten by checkout:                                                                                                                             bundles/org.openhab.core.storage.json/src/test/java/org/eclipse/smarthome/storage/json/internal/JsonStorageTest.java                                                                                    itests/org.openhab.core.config.discovery.tests/src/main/java/org/eclipse/smarthome/config/discovery/inbox/DynamicThingUpdateOSGiTest.java                                                               itests/org.openhab.core.config.discovery.tests/src/main/java/org/eclipse/smarthome/config/discovery/internal/InboxOSGiTest.java                                                                 Please move or remove them before you switch branches.                                                                                                                                                  Aborting
```

Renaming files with case-only-changes is causing conflicts on windows systems. Git can't handle that. Also some files exceed the file-name-length limit on windows systems.

wborn:
I've heard similar horror stories from other Windows users. :slightly_frowning_face: 

Since it does rebase properly on my Ubuntu machine, I've rebased your branch on the current master in my own [hivemq branch](https://github.com/wborn/openhab-core/tree/hivemq). It seems to [build well](https://travis-ci.com/wborn/openhab-core/builds/131684643) so maybe you can verify it has all your changes so we can continue with that?

PR #930, opened 2019-07-18
[facerecognition] WIP Tentative implementation of facerecognition as …

clinique:
…a profile.


clinique:
This is a RFC PR, following work done by @philippmeisberger  in his binding proposal [here](https://github.com/openhab/openhab2-addons/pull/3329).

I did create the Face Recognition as a core service that can be called by profiles.
Some parts still have to be designed / written : 
- Face recognition results in multiple elements : an image, a string, a level of recognition. I would like to forward the appropriate element through the profile depending on the target Item but have not found how to do this properly. So currently the profile only returns the name.
- I prepared an interface for the model training, but it can not be handled by Profiles. So I think this should be an action available from automation engine / scripts. Your thoughts ?

PR #931, opened 2019-07-19
start with an upper case on "initial contribution"

maggu2810:
Related to: https://github.com/openhab/openhab-core/pull/916/#discussion_r305082016

Signed-off-by: Markus Rathgeb <maggu2810@gmail.com>


maggu2810:
@cweitkamp there are some files :wink: 
But perhaps it make sense to be consistent.

cweitkamp:
>  But perhaps it make sense to be consistent.

But then we have to change it to "Initial contribution" only everywhere ... :grin: or add it if its missing ...

Issue #892, opened 2019-06-23
Multiple bundles for Guava and Jetty


5iver:
I'm seeing this in S1621...
```
 22 │ Active │  80 │ 18.0.0                │ Guava: Google Core Libraries for Java
 23 │ Active │  80 │ 21.0.0                │ Guava: Google Core Libraries for Java
 79 │ Active │  80 │ 9.4.11.v20180605      │ Jetty :: Http Utility
 80 │ Active │  80 │ 9.4.12.v20180830      │ Jetty :: Http Utility
 99 │ Active │  80 │ 9.4.11.v20180605      │ Jetty :: Websocket :: Common
100 │ Active │  80 │ 9.4.12.v20180830      │ Jetty :: Websocket :: Common
```

lolodomo:
I can confirm.

J-N-K:
For Guava this might be due to installed bindings which depend on 3rd-Party libraries which in turn depend on different versions of Guava. Not much we can do about that. For jetty we should find out why this happens.

lolodomo:
Other bundles I see in double:
```
157 │ Active │  80 │ 2.5.0.201906210315    │ openHAB Core :: Bundles :: REST Interface
158 │ Active │  80 │ 2.5.0.201906210319    │ openHAB Core :: Bundles :: REST Interface
```
and
```
211 │ Active │  80 │ 2.3.0                 │ OkHttp
212 │ Active │  80 │ 3.12.1                │ OkHttp
```

J-N-K:
OkHttp is definitely due to external libraries. This can‘t be solved. Two REST bundles is wrong, IMO.

openhab-5iver:
I can confirm the multiple REST Interface, but only have one OkHttp (3.12.1).

lolodomo:
In fact, for REST Interface, this is only a naming problem because the bundles are different:
```
157 │ Active │  80 │ 2.5.0.201906210315    │ org.openhab.core.io.rest
158 │ Active │  80 │ 2.5.0.201906210319    │ org.openhab.core.io.rest.core
```

J-N-K:
@openhab-5iver it depends on the installed bindings. The bundles are only installed when needed.

lolodomo:
For okhttp, that is fully clear for me:

* 2.3.0 is a dependency for the Netatmo binding
* 3.12.1 is a dependency for the openHAB Cloud connector

kaikreuzer:
[...]

Issue #567, opened 2019-02-11
MDNSDiscoveryServiceOSGiTest unstable


maggu2810:
https://ci.openhab.org/job/PR-openHAB-Core/1034/consoleText

```
[INFO] ------------------------------------------------------------------------
[INFO] Building openHAB Core :: Integration Tests :: Configuration mDNS Discovery Tests 2.5.0-SNAPSHOT
[INFO] ------------------------------------------------------------------------
[INFO] 
[INFO] --- maven-clean-plugin:3.0.0:clean (default-clean) @ org.openhab.core.config.discovery.mdns.tests ---
[INFO] Deleting /var/jenkins_home/workspace/PR-openHAB-Core/itests/org.openhab.core.config.discovery.mdns.tests/target
[INFO] 
[INFO] --- maven-enforcer-plugin:3.0.0-M2:enforce (enforce-java) @ org.openhab.core.config.discovery.mdns.tests ---
[INFO] 
[INFO] --- maven-resources-plugin:3.0.2:resources (default-resources) @ org.openhab.core.config.discovery.mdns.tests ---
[INFO] Using 'UTF-8' encoding to copy filtered resources.
[INFO] skip non existing resourceDirectory /var/jenkins_home/workspace/PR-openHAB-Core/itests/org.openhab.core.config.discovery.mdns.tests/src/main/resources
[INFO] Copying 1 resource
[INFO] 
[INFO] --- maven-compiler-plugin:3.8.0:compile (default-compile) @ org.openhab.core.config.discovery.mdns.tests ---
[INFO] Changes detected - recompiling the module!
[WARNING] /var/jenkins_home/workspace/PR-openHAB-Core/itests/org.openhab.core.config.discovery.mdns.tests/src/main/java/org/eclipse/smarthome/config/discovery/mdns/internal/MDNSDiscoveryServiceOSGiTest.java:[60,2243] Potential null pointer access: The variable configAdmin may be null at this location
[INFO] 
[INFO] --- bnd-maven-plugin:4.1.0:bnd-process (default) @ org.openhab.core.config.discovery.mdns.tests ---
[WARNING] /var/jenkins_home/workspace/PR-openHAB-Core/itests/org.openhab.core.config.discovery.mdns.tests/itest.bndrun [0:0]: Unused Import-Package instructions: [org.osgi.framework.*] 
[INFO] 
[...]

PR #934, opened 2019-07-19
Sync Jetty version with version used by Karaf 4.2.6

wborn:
Resolves the Jetty bundle duplicates as reported in #892 

---

I still had these changes on my local checkout. :smile:

### DIFF
--- a/features/karaf/openhab-core/pom.xml
+++ b/features/karaf/openhab-core/pom.xml
@@ -15,7 +15,7 @@
   <description>openHAB Core Features</description>
 
   <properties>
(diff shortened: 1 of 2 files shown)


lolodomo:
Cool

cweitkamp:
Thanks. Do we have to align the BOMs too? Do you think the Jenkins error (discovery.mdns tests) is related?

wborn:
The mDNS itest has been unstable for a while, see also #567.

cweitkamp:
Alright. Thanks.

PR #939, opened 2019-07-22
Fix "No SerialPortProvider found" on startup

### LINKED ISSUE #937: Be sure to start the serial providers before the bindings using them
It was discovered at least with the Z-Wave binding that at startup, the serial port is unknown by the openHAB serial abstraction. Just restarting the binding later solves the problem (the serial port is correctly identified).

https://github.com/openhab/org.openhab.binding.zwave/issues/1203

The Z-Wave binding handler factory bundle has a "immediate = true" property but it is not yet sure that it is the cause of the issue.

The current assumption is that the openHAB serial abstraction should be started with a higher priority, meaning with a start level < 80.

WDYT ?

wborn:
Change reference cardinality from `MULTIPLE` to `AT_LEAST_ONE` so the serial port provider can be used immediately when `ThingHandler`s are initialized.

Fixes #937

This would always result in the following logging on startup:

`[WARN ] [serial.internal.SerialPortManagerImpl] - No SerialPortProvider found for: /dev/ttyUSB0`

Bindings without serial port reconnection logic would need to be manually restarted (Z-Wave Binding).

### DIFF
--- a/bundles/org.openhab.core.io.transport.serial/src/main/java/org/eclipse/smarthome/io/transport/serial/internal/SerialPortRegistry.java
+++ b/bundles/org.openhab.core.io.transport.serial/src/main/java/org/eclipse/smarthome/io/transport/serial/internal/SerialPortRegistry.java
@@ -50,7 +50,7 @@ public class SerialPortRegistry {
      *
      * @param creator
      */
-    @Reference(cardinality = ReferenceCardinality.MULTIPLE, policy = ReferencePolicy.DYNAMIC, policyOption = ReferencePolicyOption.GREEDY)
+    @Reference(cardinality = ReferenceCardinality.AT_LEAST_ONE, policy = ReferencePolicy.DYNAMIC, policyOption = ReferencePolicyOption.GREEDY)
     protected void registerSerialPortCreator(SerialPortProvider creator) {
         synchronized (this.portCreators) {
             this.portCreators.add(creator);


maggu2810:
I don't think that this solves the problem correctly.
What do you think is happen if such a feature is installed after the system is fully up and running?
What do you think is happen if bundles are refreshed because a dependent bundle is update?

wborn:
When starting multiple bundles Karaf will start these before those with start level 80. What issues do you expect and what would be a better fix?

maggu2810:
Can you give me some more details how Karaf handles the start level used in features on the following situation:

* Start level 100 is reached and I install a feature that contains a bundle using start level 75.
  Is the system start level set to 60, all bundles using a start level > 60 are stopped and the system needs to restart all bundles again?
* A feature that defines a bundle is using start level 75 is installed and the bundle is started (assume we entered the level 75). Do we remain at start level 75 until all services (so, not only the execution of the activator) are activated? I don't think SCR will block the start level until all services are satisfied.

If there is any need to wait for something, we should work with requirements and capabilities.
If a service needs "something", ensure that there is a capability provided as soon as "something" is present and add a requirement for that capability.

Karaf feature is a Karaf specific implementation and the bundles and its services should work without relying on such external setup. An OSGi bundle contains a manifest and the declarative services contains XMLs that defines its requirements. We should rely only on that information that is contained in the bundle itself as this can be used by the specific implementations (core framework, SCR, ...) only.

lolodomo:
Another solution, nevertheless probably not ideal, could be to replace dynamic references to the serial providers in the registry by "hardcoded" static reference to the serial providers we need. Doiing that, we will be sure that the provider is started before.

wborn:
I also don't like using start-levels, unless there are no other options available.

> Can you give me some more details how Karaf handles the start level used in features on the following situation

Apparently that kind of information is hard to find in the Karaf documentation. :disappointed:  I think it just starts feature bundles according to the [OSGi Start Level API](https://osgi.org/specification/osgi.core/7.0.0/framework.startlevel.html). 

There are some add-ons using custom start-levels that don't seem to give any issues when they are installed after the runtime has started:

```
org.openhab.transform.jinja/src/main/feature/feature.xml:        <bundle start-level="75">mvn:org.openhab.addons.bundles/org.openhab.transform.jinja/${project.version}</bundle>
org.openhab.transform.exec/src/main/feature/feature.xml:        <bundle start-level="75">mvn:org.openhab.addons.bundles/org.openhab.transform.exec/${project.version}</bundle>
org.openhab.transform.xslt/src/main/feature/feature.xml:        <bundle start-level="75">mvn:org.openhab.addons.bundles/org.openhab.transform.xslt/${project.version}</bundle>
org.openhab.binding.mqtt.homeassistant/src/main/feature/feature.xml:        <bundle start-level="81">mvn:org.openhab.addons.bundles/org.openhab.binding.mqtt.generic/${project.version}</bundle>
org.openhab.binding.mqtt.homeassistant/src/main/feature/feature.xml:        <bundle start-level="82">mvn:org.openhab.addons.bundles/org.openhab.binding.mqtt.homeassistant/${project.version}</bundle>
org.openhab.transform.javascript/src/main/feature/feature.xml:        <bundle start-level="75">mvn:org.openhab.addons.bundles/org.openhab.transform.javascript/${project.version}</bundle>
org.openhab.transform.regex/src/main/feature/feature.xml:        <bundle start-level="75">mvn:org.openhab.addons.bundles/org.openhab.transform.regex/${project.version}</bundle>
org.openhab.binding.mqtt.homie/src/main/feature/feature.xml:        <bundle start-level="81">mvn:org.openhab.addons.bundles/org.openhab.binding.mqtt.generic/${project.version}</bundle>
org.openhab.binding.mqtt.homie/src/main/feature/feature.xml:        <bundle start-level="82">mvn:org.openhab.addons.bundles/org.openhab.binding.mqtt.homie/${project.version}</bundle>
org.openhab.io.mqttembeddedbroker/src/main/feature/feature.xml:        <bundle start-level="75">mvn:org.openhab.addons.bundles/org.openhab.io.mqttembeddedbroker/${project.version}</bundle>
org.openhab.transform.xpath/src/main/feature/feature.xml:        <bundle start-level="75">mvn:org.openhab.addons.bundles/org.openhab.transform.xpath/${project.version}</bundle>
org.openhab.transform.scale/src/main/feature/feature.xml:        <bundle start-level="75">mvn:org.openhab.addons.bundles/org.openhab.transform.scale/${project.version}</bundle>
org.openhab.transform.bin2json/src/main/feature/feature.xml:        <bundle start-level="75">mvn:org.openhab.addons.bundles/org.openhab.transform.bin2json/${project.version}</bundle>
org.openhab.transform.map/src/main/feature/feature.xml:        <bundle start-level="75">mvn:org.openhab.addons.bundles/org.openhab.transform.map/${project.version}</bundle>
org.openhab.transform.jsonpath/src/main/feature/feature.xml:        <bundle start-level="75">mvn:org.openhab.addons.bundles/org.openhab.transform.jsonpath/${project.version}</bundle>
org.openhab.binding.mqtt.generic/src/main/feature/feature.xml:        <bundle start-level="81">mvn:org.openhab.addons.bundles/org.openhab.binding.mqtt.generic/${project.version}</bundle>
```

I tried installing/uninstalling several (after startup finished) and it doesn't look like the system start level changes because bundles with start-level 80 remain Active.

> If a service needs "something", ensure that there is a capability provided as soon as "something" is present and add a requirement for that capability.

I tried that with https://github.com/wborn/openhab-core/commit/3a3fa36fb711f007e5b340355799cfb363a567ac but these capabilities/requirements don't seem to solve the issue. Maybe you know a better way?

The [Karaf documentation](https://karaf.apache.org/manual/latest/provisioning) is lacking details on the feature requirements too. I didn't find any explanation on how they influence startup behavior.

maggu2810:
> There are some add-ons using custom start-levels that don't seem to give any issues when they are installed after the runtime has started:

I already worked hard some time to get things running without rely on start levels.
So perhaps that ones are not necessary anymore at all.


Sorry, it seems I had not been clean enough about capabilities and requirements.
What I want to say is that you should not rely on Karaf only specific tooling. The Karaf features exist on Karaf only and so the capabilities and requirements of the Karaf features, too.
We should use capabilities and requirements of pure OSGi and its component specific ones.
Imported packages are requirements of the bundle, exported packages are capabilities.
Dependent services (references) of a component are requirements, provided ones capabilities.
You can use a target filter to specify further requirements, ...

wborn:
What also fixes the issue is changing the [reference cardinality](https://github.com/openhab/openhab-core/blob/c95dfcb1cc680272f16daf247069305d7f2f1e12/bundles/org.openhab.core.io.transport.serial/src/main/java/org/eclipse/smarthome/io/transport/serial/internal/SerialPortRegistry.java#L53) from `MULTIPLE` to `AT_LEAST_ONE`. It keeps the workaround simple and generic. WDYT?

lolodomo:
This will give a more chance to work, but I think you are using two serial providers, RxTxPortProvider and RFC2217PortProvider. If we have no chance, the bindings could be started while RFC2217PortProvider is loaded but not RxTxPortProvider.
Could we set the min cardinality rather to 2 ? In this case, I think it should work in all cases.

wborn:
Two is not a valid [reference cardinality](https://osgi.org/javadoc/osgi.cmpn/7.0.0/org/osgi/service/component/annotations/ReferenceCardinality.html). :wink: The only other way I see is using a `STATIC` [ReferencePolicy](https://osgi.org/javadoc/osgi.cmpn/7.0.0/org/osgi/service/component/annotations/ReferencePolicy.html) with some targets that needlessly hardcode the configuration to work with the openHAB distro only.

lolodomo:
Yes, I agree with you conclusion.
Let's see what @maggu2810 thinks.

maggu2810:
There are currently three serial transport implementations:

* openhab-core-io-transport-serial-javacomm
* openhab-core-io-transport-serial-rfc2217
* openhab-core-io-transport-serial-rxtx

The providers support different protocols.

What do you want to ensure?

Would you like that at least provider should be present without knowing which one?
Then the port you request can still be missing and it must be handled.

Would you like that at least x providers should be present?
You still don't know which one is present and missing (if more providers will be created to support other implementations, the problem is more visible).

Would you like to wait for an accepted protocol?
But if a setup don't want to install / support that protocol but others?
Then the specific serial provider needs to be installed also if not really used and not needed, too.

Mostly a binding communicate with a serial device and the binding should not care if the device is attached remote using RFC2217 or locally.

To make things more easier for binding developers why not thinking about adding "thing handler requirements" and the thing manager implementation should respect that requirements?
For example:
If a thing handler needs a serial port communication, the thing manager implementation could be informed about the requirement for a serial port "foo" (the one that is configured) and the thing manager will initialize the handler only if there is a serial port provider that provides that port...

We could also think about an serial port listener where you register for a specific port.
Some implementation could check every e.g. minute if this port is available (if not available yet) and inform the listeners about the presence of that port...

But the most simple and generic approach would be that a binding is able to handle the presence and absence of a serial port itself.
For a binding that runs on my SmartHome system I would require to handle the situation I plug off my USB to serial adapter and plug in it again without the need to restart anything.

lolodomo:
I agree with your conclusion and I hope most of the bindings are working like that (reconnection when the port is available) but apparently it is not something easy to implement in the Z-Wave binding.

Regarding your idea, the problem I see is that the port is not something that a developer can set as requirement because this is always a user parameter (configuration parameter of the thing). How do you imagine to declare this "thing handler requirement" ?

maggu2810:
> but apparently it is not something easy to implement in the Z-Wave binding.

Why is this so special in the Z-Wave binding / what is so special?

lolodomo:
Look at this [comment](https://github.com/openhab/openhab-core/issues/937#issuecomment-513859146)

wborn:
> What do you want to ensure?

I'm looking for a simple "quick fix" so the available serial port providers are added to the `SerialPortRegistry` when the ThingHandlers are initialized. Building/extending the API might be a long-term solution to properly fix the issue. I fully agree there are some (unlikely) scenarios where the workaround in this PR may not work.

If we opt for building/extending an API to fix this, the add-ons probably need to be adapted to use this API. That may be more effort as adding reconnection logic to the add-ons that are currently missing this. Adding reconnection logic needs to be done anyway and can already be done today without creating a new API for this. There are already add-ons using such reconnection logic (see https://github.com/openhab/openhab-core/issues/937#issuecomment-513843563).

The benefit of adding an API (and the workaround in this PR) would be that ThingHandlers can connect to the serial port as soon as the provider is registered in the `SerialPortRegistry`. They now have to wait until a scheduled job reconnects to the serial port.

This whole issue is of course only an issue for add-ons using the OHC serial API. Add-ons not using the OHC serial API (directly using rxtx) also don't use the `SerialPortRegistry`.

maggu2810:
> Look at this [comment](https://github.com/openhab/openhab-core/issues/937#issuecomment-513859146)

@lolodomo I was away of the referenced comment but "it is not super easy to implement so has not yet been implemented" does IMHO not explain what's the exact problem. Is it "not super easy" because it is not clear what needs to be done or is the binding itself such special that it will blow up the world if the connection to a non existing port is opened, ...

maggu2810:
Extending the API was just one possible option to handle the problem.
IMHO the binding is responsible for handling such situations because there are several connection errors that can occur. The port can disappear and appear again on a running system. A port can be available but the device is not responding, ...

It up to the core developers to decide if they agree with that "fix".
At least it does not fix the problem but workaround the issue. And the workaround is not stable because you just ensure that one provider is bound...

lolodomo:
@cdjackson if you want to develop for the specific case of the Z-Wave binding.

Even if the current fix is not perfect, it makes sense IMHO because until at least one serial port provider is started, there is no reason to start the serial stuff, it will not be usable in practice.

But we agree that even with this fix, we could still have a problem with the Z-Wave binding because started while only the "wrong" serial port provider is loaded.

maggu2810:
I assume it is clear, but just to note that this will also affect (potentially) the lifecycle of dependent services (the serial port registry gets deactivated if there is no serial port provider for a short moment e.g. on update, if the serial port registry gets deactivated all services and so dependent thing handler factories / bundles gets deactivated, ...)

kaikreuzer:
I'd tend to agree with @maggu2810 that this is some hacky workaround while the correct fix is to make the Z-Wave binding handle such issues as implementing reconnect features is a crucial feature for every binding.

But let's merge this in order to have a quick fix for the problem.

wborn:
> the correct fix is to make the Z-Wave binding handle such issues

I've created https://github.com/openhab/org.openhab.binding.zwave/issues/1213 for this.

PR #945, opened 2019-07-27
DateTimeType: new methods toZone and toLocaleZone

lolodomo:
Enhanced tests

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/library/types/DateTimeType.java
+++ b/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/library/types/DateTimeType.java
@@ -12,11 +12,13 @@
  */
 package org.eclipse.smarthome.core.library.types;
 
+import java.time.DateTimeException;
 import java.time.LocalDateTime;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
+import java.time.zone.ZoneRulesException;
 import java.util.Calendar;
 import java.util.GregorianCalendar;
 import java.util.Locale;
@@ -142,19 +144,43 @@ public class DateTimeType implements PrimitiveType, State, Command {
     }
 
     /**
-     * Translate to the locale time zone
+     * Create a {@link DateTimeType} being the translation of the current object to the locale time zone
+     *
+     * @return a {@link DateTimeType} translated to the locale time zone
+     *
+     * @throws DateTimeException if the converted zone ID has an invalid format or the result exceeds the supported date
+     *             range
+     * @throws ZoneRulesException if the converted zone region ID cannot be found
      */
-    public void toLocaleZone() {
-        toZone(ZoneId.systemDefault());
+    public DateTimeType toLocaleZone() throws DateTimeException, ZoneRulesException {
+        return toZone(ZoneId.systemDefault());
     }
 
     /**
-     * Translate to a given zone
+     * Create a {@link DateTimeType} being the translation of the current object to a given zone
      *
-     * @param zone the destination zone
+     * @param zone the target zone as a string
+     *
+     * @return a {@link DateTimeType} translated to the given zone
+     *
+     * @throws DateTimeException if the zone has an invalid format or the result exceeds the supported date range
+     * @throws ZoneRulesException if the zone is a region ID that cannot be found
      */
-    public void toZone(ZoneId zone) {
-        zonedDateTime = zonedDateTime.withZoneSameInstant(zone);
+    public DateTimeType toZone(String zone) throws DateTimeException, ZoneRulesException {
+        return toZone(ZoneId.of(zone));
+    }
+
+    /**
+     * Create a {@link DateTimeType} being the translation of the current object to a given zone
+     *
+     * @param zoneId the target {@link ZoneId}
+     *
+     * @return a {@link DateTimeType} translated to the given zone
+     *
+     * @throws DateTimeException if the result exceeds the supported date range
+     */
+    public DateTimeType toZone(ZoneId zoneId) throws DateTimeException {
+        return new DateTimeType(zonedDateTime.withZoneSameInstant(zoneId));
     }
 
     @Override
--- a/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/library/types/DateTimeType.java
+++ b/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/library/types/DateTimeType.java
@@ -34,6 +34,7 @@ import org.eclipse.smarthome.core.types.State;
  * @author Erdoan Hadzhiyusein - Refactored to use ZonedDateTime
  * @author Jan N. Klug - add ability to use time or date only
  * @author Wouter Born - increase parsing and formatting precision
+ * @author Laurent Garnier - added methods toLocaleZone and toZone
  */
 @NonNullByDefault
 public class DateTimeType implements PrimitiveType, State, Command {
@@ -108,7 +109,7 @@ public class DateTimeType implements PrimitiveType, State, Command {
             throw new IllegalArgumentException(zonedValue + " is not in a valid format.", invalidFormatException);
         }
 
-        zonedDateTime = date;
+        zonedDateTime = date.withFixedOffsetZone();
     }
 
     /**
@@ -138,6 +139,22 @@ public class DateTimeType implements PrimitiveType, State, Command {
 
     public String format(Locale locale, String pattern) {
         return String.format(locale, pattern, zonedDateTime);
+    }
+
+    /**
+     * Translate to the locale time zone
+     */
+    public void toLocaleZone() {
+        toZone(ZoneId.systemDefault());
+    }
+
+    /**
+     * Translate to a given zone
+     *
+     * @param zone the destination zone
+     */
+    public void toZone(ZoneId zone) {
+        zonedDateTime = zonedDateTime.withZoneSameInstant(zone);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


lolodomo:
@cweitkamp : here are the additional format methods with tests. This works well and as expected.

Now, we have the choice to link the old format methods with the new format methods using  ZoneId.systemDefault() as zone. This will impact all calls to the old format methods. Impact not yet evaluated.
Or we keep unchanged the existing format methods that will continue ignoring the system timezone for formatting dates and I just update the UI code to use the new format methods with ZoneId.systemDefault() as parameter to get dates in local timezone in UI.

bundles/org.openhab.core/src/test/java/org/eclipse/smarthome/core/library/types/DateTimeTypeTest.java
```diff
@@ -231,25 +231,71 @@ public void parsingTest() {
...
 
         DateTimeType dt2 = new DateTimeType(dt1.getZonedDateTime().withZoneSameInstant(ZoneId.of("Europe/Berlin")));
         assertThat(dt2, not(is(dt1)));
         assertThat(dt2.toString(), is("2019-05-25T19:53:10.000+0200"));
 
         assertThat(dt2.format(null), is("2019-05-25T19:53:10"));
         assertThat(dt2.format("%1$td.%1$tm.%1$tY %1$tH:%1$tM"), is("25.05.2019 19:53"));
-
-        assertThat(dt2.format(Locale.GERMAN, "%1$td.%1$tm.%1$tY %1$tH:%1$tM"), is("25.05.2019 19:53"));
+        assertThat(dt1.format(null, gmt), is("2019-05-25T17:53:10"));
```
cweitkamp: Should be `dt2` in this and the following lines.

lolodomo:
I am going to enhance the tests.

Ok for the idea of the toZone method rather than my new format methods.

lolodomo:
@cweitkamp : PR is now finished and ready for a review.

I will change the REST sitemap by using the new methods in a following PR.

bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/library/types/DateTimeType.java
```diff
@@ -108,7 +109,7 @@ public DateTimeType(String zonedValue) {
             throw new IllegalArgumentException(zonedValue + " is not in a valid format.", invalidFormatException);
         }
 
-        zonedDateTime = date;
+        zonedDateTime = date.withFixedOffsetZone();
```
cweitkamp: It looks like the right thing to streamline the SPI. But in opposition I found https://github.com/eclipse/smarthome/issues/5076 which questions if `withFixedOffsetZone()` is the right way to go. Unfortunately with no result. @openhab/core-maintainers wdyt?

lolodomo: withFixedOffsetZone() is applied to the other constructors so it makes sense to apply to this one too. It allows to get identical DateTimeType when calling the different constructors with the same data.

bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/library/types/DateTimeType.java
```diff
@@ -140,6 +141,22 @@ public String format(Locale locale, String pattern) {
...
+        toZone(ZoneId.systemDefault());
+    }
+
+    /**
+     * Translate to a given zone
+     *
+     * @param zone the destination zone
+     */
+    public void toZone(ZoneId zone) {
+        zonedDateTime = zonedDateTime.withZoneSameInstant(zone);
```
cweitkamp: You might consider to return a new copy of DateTimeType here.

lolodomo:
Ok.  I hesitated between the two options.

bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/library/types/DateTimeType.java
```diff
@@ -140,6 +141,22 @@ public String format(Locale locale, String pattern) {
...
+     * Translate to the locale time zone
+     */
+    public void toLocaleZone() {
+        toZone(ZoneId.systemDefault());
+    }
+
+    /**
+     * Translate to a given zone
+     *
+     * @param zone the destination zone
```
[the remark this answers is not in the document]
lolodomo: Ok I will add it.

lolodomo:
@cweitkamp : changes done.
Please let me know if you prefer that the methods throw an exception or return null in case of problem.

bundles/org.openhab.core/src/test/java/org/eclipse/smarthome/core/library/types/DateTimeTypeTest.java
```diff
@@ -232,79 +314,105 @@ public void parsingTest() {
...
+    public void changingZoneTest() {
+        DateTimeType dt = createDateTimeType();
+        DateTimeType dt2 = dt.toLocaleZone();
+        assertEquals(parameterSet.expectedResultLocalTZ, dt2.toFullString());
+        dt2 = dt.toZone(parameterSet.defaultTimeZone.toZoneId());
+        assertEquals(parameterSet.expectedResultLocalTZ, dt2.toFullString());
+        boolean thrown = false;
+        try {
+            dt2 = dt.toZone("XXX");
+        } catch (DateTimeException e) {
```
cweitkamp: This should be an own test which expects the thrown exception:

```
    @Test(expected = DateTimeException.class)
    public void changingZoneThrowsExceptionTest() {
        // TODO
    }
```

bundles/org.openhab.core/src/test/java/org/eclipse/smarthome/core/library/types/DateTimeTypeTest.java
```diff
@@ -104,13 +136,48 @@ public ParameterSet(TimeZone defaultTimeZone, Map<String, Integer> inputTimeMap,
...
+         * @param inputTimeMap
+         * @param inputTimeZone
+         * @param inputTimeString
+         * @param expectedResult
+         * @param expectedResultLocalTZ
+         * @param locale
+         * @param pattern
+         * @param expectedFormattedResult
+         */
+        public ParameterSet(TimeZone defaultTimeZone, Map<String, Integer> inputTimeMap, TimeZone inputTimeZone,
```
cweitkamp: I am just wondering if we can reuse this constructor called by the other constructors. But that is only a side comment. Up to you.

bundles/org.openhab.core/src/test/java/org/eclipse/smarthome/core/library/types/DateTimeTypeTest.java
```diff
@@ -38,6 +39,7 @@
  * @author Thomas.Eichstaedt-Engelen - Initial contribution
  * @author Gaël L'hopital - Added Timezone and Milliseconds
  * @author Erdoan Hadzhiyusein - Added ZonedDateTime tests
+ * @author Laurent Garnier - Enhanced tests
  */
 @RunWith(Parameterized.class)
 public class DateTimeTypeTest {
```
cweitkamp: Looking at the properties of the inner class `ParameterSet` does it make sense to add nullness annotations here?

lolodomo:
Do you know if this is possible to run a particular test only once and not for each ParsmeterSet?

bundles/org.openhab.core/src/test/java/org/eclipse/smarthome/core/library/types/DateTimeTypeTest.java
lolodomo: Yes we can.

bundles/org.openhab.core/src/test/java/org/eclipse/smarthome/core/library/types/DateTimeTypeTest.java
lolodomo: Done

bundles/org.openhab.core/src/test/java/org/eclipse/smarthome/core/library/types/DateTimeTypeTest.java
lolodomo: Done

bundles/org.openhab.core/src/test/java/org/eclipse/smarthome/core/library/types/DateTimeTypeTest.java
lolodomo: Done

cweitkamp:
No, I am afraid, I do not know. But does it hurt?

lolodomo:
No, it just certainly takes few additional ms to run.

kaikreuzer:
@cweitkamp I am honestly not really clear on the effect of that change, but as you have looked deeper into it and approved it, I don't see any reason not to merge it :-)

lolodomo:
Cool, I can move to the next step, using these methods to fix a bug.

cweitkamp:
Just for clarification: From https://docs.oracle.com/javase/8/docs/api/java/time/ZonedDateTime.html#withFixedOffsetZone--

> This returns a zoned date-time where the zone ID is the same as getOffset(). The local date-time, offset and instant of the result will be the same as in this date-time.
>
> Setting the date-time to a fixed single offset means that any future calculations, such as addition or subtraction, have no complex edge cases due to time-zone rules. This might also be useful when sending a zoned date-time across a network, as most protocols, such as ISO-8601, only handle offsets, and not region-based zone IDs.

Which basically means that it cuts off the "US/Central" (or whatever timezone you are in) string from the object (you can see that in the debugger). The quoted text above also states that cutting it off simplifies calculations based on this object later on (e.g. `2018-11-04T01:25:43-06:00[US/Central]` results in `2018-11-04T01:25:43-06:00`).

PR #948, opened 2019-07-29
[WIP] REST sitemap: datetime added for the tests

lolodomo:
Also partially test our ItemUIRegistry implementation

### DIFF
--- a/bundles/org.openhab.core.io.rest.sitemap/src/test/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapResourceTest.java
+++ b/bundles/org.openhab.core.io.rest.sitemap/src/test/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapResourceTest.java
@@ -24,6 +24,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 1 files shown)


lolodomo:
This enhanced test of the API shows in particular that the widget label build from a datetime item is currently formatted using the time zone of the channel state, not the local time zone.
As a consequence, the UIs consuming this API could retrieve formatted labels for dates that are not in the local time zone.

lolodomo:
Maybe I should rather create a new test in org.openhab.core.ui for ItemUIRegistry and update the test in the REST sitemap API just to show that it returns the label provided by the ItemUIRegistry ?

lolodomo:
I will fully rework this PR to add tests for the ItemUIRegistry rather than the REST sitemap.

lolodomo:
I just found we have already tests for ItemUIRegistry.
I just need to add one additional test like this following one but with a date different from the local zone.
https://github.com/openhab/openhab-core/blob/7b36b62c49bf5cdab44990b1b1ee9d1d29181507/itests/org.openhab.core.ui.tests/src/main/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImplTest.java#L292

lolodomo:
I will create a new PR.

PR #953, opened 2019-08-04
DSL rule: new sendCommand with a State as parameter

### LINKED ISSUE #950: [Rules DSL] enhance sendCommand to handle OnOffType
Forgive my probably incorrect terminology.

`myItem.sendCommand(xxx)`
We frequently advise using sendCommand method as it is "better out sorting out argument types" for xxx.   I imagine it usually reduces xxx to its string equivalent.

Working with Switch type Items, it is not uncommon to base a **command** on some other Switch's **state**.
`mySwitch.sendCommand(otherSwitch.state)`
This results in an error
```
2019-08-01 00:13:16.139 [ERROR] [ntime.internal.engine.RuleEngineImpl] - Rule 'test': An error occurred during the script execution: Could not invoke method: org.eclipse.smarthome.model.script.actions.BusEvent.sendCommand(org.eclipse.smarthome.core.items.Item,java.lang.Number) on instance: null
```
Circumvention is simple
`mySwitch.sendCommand(otherSwitch.state.toString)`
but the error is rather unexpected from the user viewpoint.   I guess an OnOffType state of ON is not quite the same as OnOffType command ON ??

This is a request for enhancement of sendCommand method to handle Item state as argument.

Use case-
https://community.openhab.org/t/two-switches-one-light-eg-staircase-switch-rule/79035/5

lolodomo:
Fix #950

### DIFF
--- a/bundles/org.openhab.core.model.script/src/org/eclipse/smarthome/model/script/actions/BusEvent.java
+++ b/bundles/org.openhab.core.model.script/src/org/eclipse/smarthome/model/script/actions/BusEvent.java
@@ -71,6 +71,20 @@ public class BusEvent {
     }
 
     /**
+     * Sends a state as a command for a specified item to the event bus.
+     *
+     * @param item the item to send the command to
+     * @param state the state from which to get the command to send
+     */
+    static public Object sendCommand(Item item, State state) {
+        if (item != null && state != null) {
+            return sendCommand(item.getName(), state.toString());
+        } else {
+            return null;
+        }
+    }
+
+    /**
      * Sends a command for a specified item to the event bus.
      *
      * @param itemName the name of the item to send the command to


lolodomo:
This is not tested as I don't have any IDE in which I can run openHAB.

openhab-5iver:
You could build org.openhab.core.model.script, drop it into a system running a recent snapshot, clear the cache, and test it that way. That's how I ended up testing a lot of the automation changes without an IDE to debug in.

lolodomo:
I have no problem to test bindings in such a way but to replace a core bundle in my production environment, I am less aware how to achieve that **properly**. What do you mean exactly by "clear the cache" ? And should I first uninstall or stop the existing bundle before putting the new jar in my addons folder ?

openhab-5iver:
Stop OH, copy in the file (/runtime/system/org/openhab/core/bundles/...) overwriting the old one or rename the file extension of the old one to revert later, delete /userdata/cache/org.eclipse.osgi (I found that was all I needed to clear for testing automation), restart OH. On recent snapshots, I've found that I need to restart twice after clearing the cache to startup without errors.

Hope this helps!

lolodomo:
Never try to do that in such a way. Will try.

J-N-K:
You can also replace a single bundle with `bundle:update <number> <uri>`. `<uri>` can be something like `file:///home/lolodomo/my-cool-new-bundle.jar`. Going back is the same, just update to the „old“ (unpatched) version.

lolodomo:
Looks cleaner ;)

openhab-5iver:
Wow... I've done that with addons, but never thought to do it with a core bundle. That will definitely save some time... if it works!

lolodomo:
I can confirm that this enhancement is working perfectly. I tested it with a Switch item and a Number item.

PS: regarding the way to update the bundle, the solution with bundle:update is ok but the bundle remained in status "Waiting". I restarted openHAB and that was OK with the new bundle.

J-N-K:
Maybe you need to `bundle:resolve <number>` and/or `bundle:start <number>`. `bundle:diag <number>` tells you what is missing to start the bundle.

lolodomo:
Yes, I should have tried at least bundle:start ! Will try the next time.

lolodomo:
@J-N-K : I tried again to restore the old bundle.
`bundle:start 171` does not help, the bundle remains in state Waiting.
`bundle:resolve 171` returns nothing.
bundle:diag 171 returns:
```
openHAB Core :: Bundles :: Model Script (171)
---------------------------------------------
Status: Waiting
Declarative Services
org.eclipse.smarthome.model.script.extension.ScriptEngineConsoleCommandExtension (291)
  missing references: ScriptEngine
```
Here are the bundles in a Waiting state:
```
164 x Waiting x  80 x 2.5.0.201907292052    x openHAB Core :: Bundles :: Language Server
170 x Waiting x  90 x 2.5.0.201907292033    x openHAB Core :: Bundles :: Model Rules Runtime
171 x Waiting x  80 x 2.5.0.201907292031    x openHAB Core :: Bundles :: Model Script
173 x Waiting x  80 x 2.5.0.201907292052    x openHAB Core :: Bundles :: Model Script Runtime
```
I tried to start 170 and 173 in addition to 171 but with no more success.

lolodomo:
What is done here is that we try to get a command from the state. It looks like what you request.
https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.model.script/src/org/eclipse/smarthome/model/script/actions/BusEvent.java#L85
If this is failing, th efollowing message will be logged: Cannot convert '{}' to a command type which item '{}' accepts: {}.

cweitkamp:
@lolodomo Do you agree with closing this PR?

lolodomo:
It is really a pity and I don't understand the reason why this enhancement is rejected
But yes, if mainteners don't want this improvement, it has no sense to keep it open.

PR #958, opened 2019-08-06
Fix null annotations order and cleanup

wborn:
* Move null annotations between modifiers and variable type
* Remove redundant public modifiers from interfaces where applicable
* Remove redundant `@NonNull` annotations


wborn:
These changes were carefully crafted by hand. :wink:

PR #965, opened 2019-08-10
AbstractFileTransformationService: stop watchService properly

lolodomo:


lolodomo:
Here is where it is done for our AbstractWatchService :
https://github.com/openhab/openhab-core/blob/48d873a32ba0901a9b58f04c541f0f858fbfcd51/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/service/AbstractWatchService.java#L72
https://github.com/openhab/openhab-core/blob/48d873a32ba0901a9b58f04c541f0f858fbfcd51/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/service/WatchQueueReader.java#L169

cweitkamp:
As we already have an abstract base class for OSGi services using the Java `WatchService` doesn't it make sense to let the `AbstractFileTransformatiomService` inherit from it?

lolodomo:
My intention was not to refactor the class, but only to fix a "little" bug.
This was implemented without using the `AbstractFileTransformatiomService`. There might be a good reason.
Sorry, I will not spend more time to refactor that.

PR #978, opened 2019-08-15
[IconServlet] New optional parameter anyFormat

### LINKED ISSUE #824: Fall back to other icon format if no icon is present for requested format
On high dpi mobile phones the PNG version of the default icon format looks pixelated (https://github.com/openhab/openhab-android/issues/1337). A user could change his icon format to svg, but then he has to provide all custom icons as svg, too.
I was thinking about this issue and IMO the best solution is to add a parameter to the icon request, like `fallback=true`. If added, the server may return icons in a different format, if no icon for the requested format was found. The client can look at the mime type to know how to parse the icon.
This would also allow us to change the default icon format in the clients to "svg" (for new installations), since it will only break the user experience if the user provides a png icon that overrides one of openHAB's icons.

lolodomo:
Fix #824

If this new parameter is not used by the caller (UI), the behavior is unchanged. So this change will break no existing UI.

When this new parameter is provided and set to true by the caller (UI), the server will search the icon as SVG and PNG formats. If the icon provider is the same in the two cases, it will return the icon with the requested format. In case the 2 icon providers are different for SVG and PNG, the server will return the icon in the format corresponding to the provider with higer priority.

This new parameter allows an UI to select a preferred format for icons without loosing the customized user icons that can be defined  in a different format. Of course, in this case, the UI must handle possible different icon formats returned by the servlet.

### DIFF
--- a/bundles/org.openhab.core.io.rest/src/main/java/org/eclipse/smarthome/io/rest/internal/resources/beans/RootBean.java
+++ b/bundles/org.openhab.core.io.rest/src/main/java/org/eclipse/smarthome/io/rest/internal/resources/beans/RootBean.java
@@ -24,7 +24,7 @@ import java.util.List;
  */
 public class RootBean {
 
-    public final String version = "2";
+    public final String version = "3";
 
     public final List<Links> links = new ArrayList<Links>();
 
--- a/bundles/org.openhab.core.ui.icon/src/main/java/org/eclipse/smarthome/ui/icon/internal/IconServlet.java
+++ b/bundles/org.openhab.core.ui.icon/src/main/java/org/eclipse/smarthome/ui/icon/internal/IconServlet.java
@@ -36,7 +36,6 @@ import org.osgi.service.component.annotations.Modified;
 import org.osgi.service.component.annotations.Reference;
 import org.osgi.service.component.annotations.ReferenceCardinality;
 import org.osgi.service.component.annotations.ReferencePolicy;
-import org.osgi.service.http.HttpContext;
 import org.osgi.service.http.HttpService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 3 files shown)


lolodomo:
@mueller-ma : please have a look.

mueller-ma:
Sounds good to, but IMO this PR should bump the rest api version.
@maniac103 What do you think about this change?

lolodomo:
Note that it enhance the API without breaking it.
But I understand that a new version will help you to know if the new parameter is available.
Where is set the REST API version ?

maniac103:
The version is defined in RootBean.java.
In general, the idea makes sense to me, as we can introduce an 'automatic' format mode with that (asking for SVG with anyFormat=true). That is assuming the sent mime type matches the sent image, but I'd be surprised if that weren't the case.

lolodomo:
Maybe you can simply use the server version ?

maniac103:
Server version as in ... ? You mean parsing e.g. `2.4.0`? That isn't really possible (or at least highly impractical) due to the huge number of corner cases in that approach.

kaikreuzer:
Increasing the version in the RootBean is perfectly fine - it is meant for the UIs to check whether they can expect a certain feature on the server or not.

lolodomo:
Fine, I am going to increase it. I am just surprised that the version is now only at 2 after so many years.

lolodomo:
I moved the version from 2 to 2.1 but I hesitated with 2.0.1.

kaikreuzer:
No, please increase it to 3, it is meant as an integer without any semver information on it.

lolodomo:
Done, version changed from 2 to 3.

lolodomo:
Can this one be merged ? ;)

kaikreuzer:
Indeed, everything looks sooooo green ;-)

PR #983, opened 2019-08-16
[WIP][money] Initial contribution of 'MonetaryType' and Money API

cweitkamp:
- Initial contribution of `MonetaryType` and Money API

Beside Units of Measurement for representing sensor data and a lot more features I have a need to do some calculations with money. After UoM has been introduced I always got the feeling I had to extend it to support different currencies. After some initial coding sessions I realized that this is the wrong path. Then I found JSR 354 via Google search, get in touch with it and gave it an try.

Based on [Money & Currency API (JSR 354)](https://javamoney.github.io/api.html) using [Moneta implementation](https://javamoney.github.io/ri.html).

@openhab/core-maintainers Is this an option for us? This PR is far away from being useful. Just want to have a hint if I should continue or not. Thanks.

### DIFF
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -202,6 +202,26 @@
       <scope>compile</scope>
     </dependency>
 
+    <!-- Money -->
+    <dependency>
+      <groupId>javax.money</groupId>
+      <artifactId>money-api</artifactId>
+      <version>1.0.3</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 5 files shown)


wborn:
I think it makes a lot of sense to add Money support. A common use case will be to measure power consumption/production and calculate the associated money loss/gain.

Building upon the JSR implementation will probably make it easier to switch to the default Java implementation if it is accepted one day. 

Another Money implementation which has been around a bit longer is [Joda-Money](https://www.joda.org/joda-money/).

PR #989, opened 2019-08-21
[semantics] Specified service for component description

cweitkamp:
- Specified service for component description
- Implement `ItemRegistryChangeListener` interface instead of `RegistryChangeListener<Item>`

### DIFF
--- a/bundles/org.openhab.core.semantics/src/main/java/org/eclipse/smarthome/core/semantics/internal/SemanticsMetadataProvider.java
+++ b/bundles/org.openhab.core.semantics/src/main/java/org/eclipse/smarthome/core/semantics/internal/SemanticsMetadataProvider.java
@@ -23,10 +23,10 @@ import java.util.TreeMap;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.smarthome.core.common.registry.AbstractProvider;
-import org.eclipse.smarthome.core.common.registry.RegistryChangeListener;
 import org.eclipse.smarthome.core.items.GroupItem;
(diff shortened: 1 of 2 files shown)


kaikreuzer:
Thanks for this fix!

Could you briefly explain why this resolves the circular reference?
And registering the service as a MetaDataProvider, does it mean that this was lost when removing the xmls (and thus the whole service wasn't working?)

cweitkamp:
I have to admit, that I am not sure if it will resolve the circular reference. I tried to identify it but I could not find it. If it really exists the construction injection changes will reveal it. Maybe there is another consumer involved which is not part of OHC. I will ask @digitaldan about it and remove the magic words from my OP.

 `MetadataProvider` (`SemanticsMetadataProvider`) -->
(1..1) `ItemRegistry` (`ItemRegistryImpl`) -->
(1..1) `MetadataRegistry` (`MetadataRegistryImpl`) -->
(optional, dynamic) `ManagedMetadataProvider` (`ManagedMetadataProviderImpl`) -->
(1..1) `StorageService` ...

What caught my attention was this message. SCR created a service for the `RegistryChangeListener` - which is not supposed to be registered as a service - I guess. The specific declaration now takes care of that.

```
{org.eclipse.smarthome.core.items.MetadataProvider, org.eclipse.smarthome.core.common.registry.RegistryChangeListener}={service.id=256, service.bundleid=186, service.scope=bundle, component.name=org.eclipse.smarthome.core.semantics.internal.SemanticsMetadataProvider, component.id=134}
```

I am afraid I cannot tell you if something got lost back then. The initial contribution (https://github.com/eclipse/smarthome/pull/6288) already used DS annotations.

PR #990, opened 2019-08-21
Fix Incorrect unit symbol for Byte and add unit symbol for Octet.

Cossey:
I was updating the Documentation for the Unit of Measurement as requested by @cweitkamp and found an issue with a couple of UoM symbols during this.

Signed-off-by: Stewart Cossey <stewart.cossey@gmail.com>


cweitkamp:
@clinique Introduced by you in #681, may I ask you for a review? Thanks in advance.

PR #995, opened 2019-08-24
Enable Config Dispatch Tests

cweitkamp:
- Enable Config Dispatch Tests

Related to #698

Currently failing because the tests try to load a configurations file from a folder placed in `src/main/resources`.


cweitkamp:
One step forward - but still failing: The test now complains about:
```
Unable to invoke no-args constructor for class org.eclipse.smarthome.config.dispatch.internal.ConfigDispatcher$ExclusivePIDMap. Registering an InstanceCreator with Gson for this type may fix this problem.
```

I already added it in https://github.com/openhab/openhab-core/pull/995/commits/563c316caf6b731bc1ce12ab38c1a8353583be25 but it did not help. @openhab/core-maintainers Any educated guess?

wborn:
Shouldn't both the class and constructor be public because Gson itself is in another package?

The stacktrace also shows from what class/package the constructor is called (using reflection):

```
java.lang.RuntimeException: Unable to invoke no-args constructor for class org.eclipse.smarthome.config.dispatch.internal.ConfigDispatcher$ExclusivePIDMap. Registering an InstanceCreator with Gson for this type may fix this problem.
	at com.google.gson.internal.ConstructorConstructor$14.construct(ConstructorConstructor.java:226)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:210)
```

Probably an `InstanceCreator` can be used with private class/constructor when it's defined in the `ConfigDispatcher` class. Or when the `InstanceCreator` is part of the same package you can use the default (package) scope with the class/constructor

cweitkamp:
The trick seems to be an inner `public static` class and a package private no-args constructor.

itests/org.openhab.core.config.dispatch.tests/src/main/java/org/eclipse/smarthome/config/dispatch/test/ConfigDispatcherOSGiTest.java
```diff
@@ -47,7 +50,7 @@
 
     private ConfigDispatcher cd;
 
-    private static final String CONFIGURATION_BASE_DIR = "configurations";
+    private static final String CONFIGURATION_BASE_DIR = "../../../../../target/classes/configurations";
```
cweitkamp: When using this path it runs on my local machine and on our build servers. It does not run from within the Eclipse IDE.

To let it succeed in the IDE I have to change the path to `"../../../../target/classes/configurations"`.

kaikreuzer: I guess we will have to live with this for now?

Issue #998, opened 2019-08-26
Allow multiple values for default configuration


cweitkamp:
The "default" value of a configuration is always a String. This could be a problem if a service / thing / channels uses a configuration description defined with `multiple="true"` and more than one default value should be available (see e.g. #919).

I worked out a three step plan to get it supported:

1. [x] Extend Config Dispatcher to support values containing a list of configuration options (see #999)
2. [x] Extend REST interface to return a list for `defaultValue` if `multiple="true"` (see #1031)
3. [x] (later) Consider more than one value(s) when applying the default value for Things and Channels. (see #1067)

My research on item 2:

When adding a default value for a service in the configuration description like the following example it will not be applied immediately:
```
	<config-description uri="system:ephemeris">
		<parameter name="dayset-weekend" type="text" required="true" multiple="true">
			<label>Weekend Days</label>
			<description>List of usual non workable days.</description>
			<default>SATURDAY,SUNDAY</default>
		</parameter>
	</config-description>
```

The wrong listing in Paper UI ([as described here](https://user-images.githubusercontent.com/12213581/62833289-aa232380-bc3c-11e9-8566-3f5f92ca7219.png)) results from a UI side merge of the configuration options and the default value. The UI calls the REST interface for information:

Get the configuration description (GET /rest/config-descriptions/system:ephemeris, options for parameters `country` and `region` are truncated):
```
{
  "uri": "system:ephemeris",
  "parameters": [
    {
      "defaultValue": "SATURDAY,SUNDAY",
      "description": "List of usual non workable days.",
      "label": "Weekend Days",
      "name": "dayset-weekend",
      "required": true,
      "type": "TEXT",
      "readOnly": false,
      "multiple": true,
      "advanced": false,
      "verify": false,
      "limitToOptions": true,
      "options": [
        {
          "label": "Montag",
          "value": "MONDAY"
        },
        {
          "label": "Dienstag",
          "value": "TUESDAY"
        },
        {
          "label": "Mittwoch",
          "value": "WEDNESDAY"
        },
        {
          "label": "Donnerstag",
          "value": "THURSDAY"
        },
        {
          "label": "Freitag",
          "value": "FRIDAY"
        },
        {
          "label": "Samstag",
          "value": "SATURDAY"
        },
        {
          "label": "Sonntag",
          "value": "SUNDAY"
        }
      ],
      "filterCriteria": [
      ]
    },
    {
[...]

PR #999, opened 2019-08-26
[config dispatch] Added support for values containing a list of configuration options

cweitkamp:
- Added support for values containing a list of configuration options (comma-separated list)

Related to #998

Syntax will be:
```
<service-pid>:<property>=<value_1>,<value_2>,...,<value_n>
```

Is this syntax okay? Or do we risk to break something existent because a comma (`,`) is a common used character in configurations. Should we maybe surround the value list by square brackets?
```
<service-pid>:<property>=[<value_1>,<value_2>,...,<value_n>]
```

I would like to add some unit tests. Before that we have to get #995 running.

### DIFF
--- a/bundles/org.openhab.core.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
+++ b/bundles/org.openhab.core.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
@@ -31,6 +31,8 @@ import java.util.stream.Collectors;
 
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
+import org.eclipse.jdt.annotation.NonNullByDefault;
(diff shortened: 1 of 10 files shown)


maggu2810:
Why not using the same syntax as e.g. configadmin is using

cweitkamp:
Yes, that was my first intention. But I am either not skilled enough to google it successfully or it is not proper documented. Do you have a link for me?

I guess it is using round brackets to indicate lists:
```
<service-pid>:<property>=(<value_1>,<value_2>,...,<value_n>)
```

One issue with the config dispatcher is that it only can handle single line entries.

maggu2810:
For configadmin it is e.g.

```
prop=["magic","stuff"]
```

cweitkamp:
I added the brackets. Syntax now will be:

```
<service-pid>:<property>=[<value_1>,<value_2>,...,<value_n>]
```

I also added integration tests. This PR now depends fully on #995.

cweitkamp:
Rebased to latest master.

PR #1004, opened 2019-08-27
Get and set the thing configuration from rules

J-N-K:


cweitkamp:
I like it. Looks a little bit like #583.

J-N-K:
Didn't see that. The only difference I see is that you get/set the whole map while I only get/set one parameter.

lolodomo:
By curiosity, what can be the usage ?

cweitkamp:
> The only difference I see is that you get/set the whole map while I only get/set one parameter

No problem. Yes, your approach maybe is the better one. But there are still some more things to consider.

1. We should think forward and add an action for the new rule engine too.
2. What about adding a Karaf console command?
3. When we allow setting the configuration parameters doesn't it make sense to take the configuration description into account? And normalize the values like we do in the REST endpoint?
4. What about channel configuations?

PR #1009, opened 2019-08-29
Selection widget as default widget for Number or String items with options

### LINKED ISSUE #984: Selection widget by default for items with state options
What do you think about changing the default widget from Text to Selection for items having state options and are not read only ?

lolodomo:
Fixes #984

Before the change (display from a group so with the default rendering):
image

And now with this change in Basic UI:
image

And in Classic UI:
image

### DIFF
--- a/bundles/org.openhab.core.ui/src/main/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImpl.java
+++ b/bundles/org.openhab.core.ui/src/main/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImpl.java
@@ -67,6 +67,7 @@ import org.eclipse.smarthome.core.library.types.StringType;
 import org.eclipse.smarthome.core.transform.TransformationException;
 import org.eclipse.smarthome.core.transform.TransformationHelper;
 import org.eclipse.smarthome.core.transform.TransformationService;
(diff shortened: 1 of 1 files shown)


lolodomo:
Maybe I should check that the item is not read only. If read only, use the text widget.

kaikreuzer:
Yes, if that information is available, it would be nice to be used. I'll wait with the merge, so you can amend that to the PR, if you like.

lolodomo:
@kaikreuzer : it is done.

lolodomo:
I installed the snapshot 1674, thinking this new feature will be included. Unfortunately, it isn't. Checking the core bundles versions, I discovered that they are from August 28 only (not 30). Is it due to failing builds of openHAB core since few days ?

kaikreuzer:
Yes.

lolodomo:
The problem seems to be a missing dependency in the DBus transport bundle. @maggu2810 : can you have a look please ? It is a dependency to something from you apparently.

kaikreuzer:
@lolodomo The problem came with https://github.com/openhab/openhab-core/pull/1011, see my comment https://github.com/openhab/openhab-core/pull/1011#issuecomment-526597868.
Still trying to understand why the build isn't checking for the correct repositories.

PR #1010, opened 2019-08-29
[rest] Changed property type to Boolean to allow null values

### LINKED ISSUE #985: [REST] Editable Field Not Eliminated Error
When requesting for all items filtering specific fields, the below error is generated and the "editable" field would be included in the response even though not selected.

```
2019-08-17 16:14:04.854 [WARN ] [thome.io.rest.internal.DTOMapperImpl] - Field 'editable' could not be eliminated: Can not set boolean field org.eclipse.smarthome.io.rest.core.item.EnrichedItemDTO.editable to null value
```

**Request**
http://localhost:8080/rest/items?fields=name

**Response**
```
[
  {
    "editable": false,
    "name": "device1"
  },
  {
    "editable": false,
    "name": "device2"
  }
]
```

cweitkamp:
- Changed property type to `Boolean` to allow `null` values

Fixes #985 

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/item/EnrichedItemDTO.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/item/EnrichedItemDTO.java
@@ -34,7 +34,7 @@ public class EnrichedItemDTO extends ItemDTO {
     public StateDescription stateDescription;
     public CommandDescription commandDescription;
     public Map<String, Object> metadata;
-    public boolean editable;
+    public Boolean editable;
 
     public EnrichedItemDTO(ItemDTO itemDTO, String link, String state, String transformedState,
             StateDescription stateDescription, CommandDescription commandDescription) {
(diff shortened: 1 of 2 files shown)


cweitkamp:
Tests are failing because we seem to expect the `editable` field in the response. Can we change it unscrupulously without introducing a regression?

https://github.com/openhab/openhab-core/blob/b3802d565db205ba6a04080c5b3286e4612dde4c/itests/org.openhab.core.io.rest.core.tests/src/main/java/org/eclipse/smarthome/io/rest/core/internal/item/ItemResourceOSGiTest.java#L160-L170

kaikreuzer:
We could set the `Boolean` to a value instead of leaving it null - in that case, the response will always contain the field.
I am just wondering, why the test expects a `true` while the `boolean` had `false` as a default value. Do you have any idea?

cweitkamp:
The value of `editable` is set every time. Regardless if we filter it or not. It is `true` when the item is managed by a `ManagedItemProvider` like it is done in the integration test.

If the `editable` field is mandatory we could skip the warning for it instead of changing the type of the `EnrichedItemDTO` property.

kaikreuzer:
Sorry, I guess I misunderstood your initial question:

> Can we change it unscrupulously without introducing a regression?

You mean changing the test to not expect `editable´ anymore? Yes, I think we can and we actually should as it is currently not really the response that one would expect and the test was obviously only written to fit to the implementation, but not to the theory.

PR #1017, opened 2019-08-31
REST Voice API: new API for TTS

### LINKED ISSUE #927: [REST API] Endpoint for "Say" under "Voice" missing
Currently the "Say" action is only available through the rule engine. There is a Rest endpoint "Voice" where there should be an endpoint "say", too.

lolodomo:
Fix #927

### DIFF
--- a/bundles/org.openhab.core.io.rest.voice/src/main/java/org/eclipse/smarthome/io/rest/voice/internal/VoiceResource.java
+++ b/bundles/org.openhab.core.io.rest.voice/src/main/java/org/eclipse/smarthome/io/rest/voice/internal/VoiceResource.java
@@ -32,7 +32,10 @@ import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
 import javax.ws.rs.core.UriInfo;
 
+import org.eclipse.smarthome.core.audio.AudioManager;
+import org.eclipse.smarthome.core.audio.AudioSink;
 import org.eclipse.smarthome.core.auth.Role;
+import org.eclipse.smarthome.core.voice.Voice;
 import org.eclipse.smarthome.core.voice.VoiceManager;
 import org.eclipse.smarthome.core.voice.text.HumanLanguageInterpreter;
 import org.eclipse.smarthome.core.voice.text.InterpretationException;
@@ -54,6 +57,7 @@ import io.swagger.annotations.ApiResponses;
  * This class acts as a REST resource for voice features.
  *
  * @author Kai Kreuzer - Initial contribution and API
+ * @author Laurent Garnier - add TTS feature to the REST API
  */
 @Component
 @Path(VoiceResource.PATH_SITEMAPS)
@@ -67,6 +71,7 @@ public class VoiceResource implements RESTResource {
     UriInfo uriInfo;
 
     private VoiceManager voiceManager;
+    private AudioManager audioManager;
     private LocaleService localeService;
 
     @Reference(cardinality = ReferenceCardinality.OPTIONAL, policy = ReferencePolicy.DYNAMIC)
@@ -76,6 +81,15 @@ public class VoiceResource implements RESTResource {
 
     public void unsetVoiceManager(VoiceManager voiceManager) {
         this.voiceManager = null;
+    }
+
+    @Reference(cardinality = ReferenceCardinality.OPTIONAL, policy = ReferencePolicy.DYNAMIC)
+    public void setAudioManager(AudioManager audioManager) {
+        this.audioManager = audioManager;
+    }
+
+    public void unsetAudioManager(AudioManager audioManager) {
+        this.audioManager = null;
     }
 
     @Reference(cardinality = ReferenceCardinality.OPTIONAL, policy = ReferencePolicy.DYNAMIC)
@@ -106,7 +120,7 @@ public class VoiceResource implements RESTResource {
     @GET
     @Path("/interpreters/{id: [a-zA-Z_0-9]*}")
     @Produces(MediaType.APPLICATION_JSON)
-    @ApiOperation(value = "Gets a single interpreters.", response = HumanLanguageInterpreterDTO.class)
+    @ApiOperation(value = "Gets a single interpreter.", response = HumanLanguageInterpreterDTO.class)
     @ApiResponses(value = { @ApiResponse(code = 200, message = "OK"),
             @ApiResponse(code = 404, message = "Interpreter not found") })
     public Response getInterpreter(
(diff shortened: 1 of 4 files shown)


lolodomo:
In addition to this enhancement, we need another REST API for audio to get the list of audio sinks. I will cover that in a separate PR.

bundles/org.openhab.core.io.rest.voice/src/main/java/org/eclipse/smarthome/io/rest/voice/internal/VoiceResource.java
```diff
@@ -67,6 +71,7 @@
     UriInfo uriInfo;
 
     private VoiceManager voiceManager;
+    private AudioManager audioManager;
```
kaikreuzer: You should imho do without that dependency - the VoiceManager offers methods to call without sinkIds and it handles the calls internally. The REST API should behave the same way as the Java API.

bundles/org.openhab.core.io.rest.voice/src/main/java/org/eclipse/smarthome/io/rest/voice/internal/VoiceResource.java
```diff
@@ -169,6 +183,91 @@ public Response interpret(@HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) @ApiParam(va
...
+        } else if (sink == null) {
+            return JSONResponse.createErrorResponse(Status.NOT_FOUND, "Default audio sink not found");
+        } else {
+            voiceManager.say(text, voice.getUID(), sink.getId());
+            return Response.ok(null, MediaType.TEXT_PLAIN).build();
+        }
+    }
+
+    @POST
+    @Path("/say/{sinkid: [a-zA-Z_:0-9]*}")
```
kaikreuzer: The sinkId is not a member of an entity "say", so it should imho rather be a QueryParam instead of a PathParam here.

bundles/org.openhab.core.io.rest.voice/src/main/java/org/eclipse/smarthome/io/rest/voice/internal/VoiceResource.java
```diff
@@ -169,6 +183,91 @@ public Response interpret(@HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) @ApiParam(va
...
+        } else if (audioManager.getSink(sinkId) == null) {
+            return JSONResponse.createErrorResponse(Status.NOT_FOUND, "Undefined audio sink");
+        } else {
+            voiceManager.say(text, voice.getUID(), sinkId);
+            return Response.ok(null, MediaType.TEXT_PLAIN).build();
+        }
+    }
+
+    @POST
+    @Path("/say/{sinkid: [a-zA-Z_:0-9]*}/{voiceid: [a-zA-Z_:0-9]*}")
```
kaikreuzer: Same here: Make sinkid and voiceid to QueryParams (this will actually allow you to combine the three methods into one as two params can both be optional).

bundles/org.openhab.core.io.rest.voice/src/main/java/org/eclipse/smarthome/io/rest/voice/internal/VoiceResource.java
lolodomo: Ok, this was necessary just to make a pre-control and avoid returning status 200 when say will fail because the provided audio sink is wrong.
But I can suppress this test and so this dependency to AudoManager.

bundles/org.openhab.core.io.rest.voice/src/main/java/org/eclipse/smarthome/io/rest/voice/internal/VoiceResource.java
lolodomo: Makes sense.
I think I do it like that because the interpret API  just above was using this pattern.

bundles/org.openhab.core.io.rest.voice/src/main/java/org/eclipse/smarthome/io/rest/voice/internal/VoiceResource.java
lolodomo: Is it possible to have QueryParam in a POST command ?

kaikreuzer: I would think/hope so.

bundles/org.openhab.core.io.rest.voice/src/main/java/org/eclipse/smarthome/io/rest/voice/internal/VoiceResource.java
kaikreuzer: Right, but for the interpreters call, it is selecting an entity by id from the list of interpreters, so it is a different situation.

lolodomo: I don't understand the difference.
Here I want to select a voice amongst a list of voices, and a sink amongst a list of audio sinks.

bundles/org.openhab.core.io.rest.voice/src/main/java/org/eclipse/smarthome/io/rest/voice/internal/VoiceResource.java
lolodomo: I have a doubt. I will check if we have such case in all our REST APIs (POST).

lolodomo: Doing a WEB search, it looks like this is possible.

bundles/org.openhab.core.io.rest.voice/src/main/java/org/eclipse/smarthome/io/rest/voice/internal/VoiceResource.java
kaikreuzer: Your uri is `say/sinkid` and not `sinks/sinkid` or `voices/voiceid`. Hence "sinkid" is not an entity of the "say" entities. See the difference?

lolodomo: I see what you mean now.

lolodomo:
Changes done. My tests are ok.

PR #1019, opened 2019-09-01
[voice] make getVoice public

lolodomo:


lolodomo:
This will for example allow checking the voice parameter in the new REST voice API before calling say.

bundles/org.openhab.core.voice/src/main/java/org/eclipse/smarthome/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -280,7 +280,7 @@ public String interpret(String text, String hliId) throws InterpretationExceptio
         return interpreter.interpret(localeProvider.getLocale(), text);
     }
 
-    private Voice getVoice(String id) {
+    public Voice getVoice(String id) {
```
kaikreuzer: This is an internal method that is not part of the interface. As nobody should depend on the impl class, it imho makes no sense to make this public here.

lolodomo: Good point. 
So this is the interface that has to be enhanced.

lolodomo:
I close this PR.

PR #1023, opened 2019-09-04
add reactor projects (with filters for subprojects)

maggu2810:
WRT to https://github.com/openhab/openhab-distro/issues/927#issuecomment-525915184 ensure that for core also the reactor projects are "known" by the IDE.


kaikreuzer:
I didn't test it, but won't this again mess up our working set structure and cause all resources be available multiple times again?
Wrt https://github.com/openhab/openhab-distro/issues/927#issuecomment-525915184, I don't see how this helps contributors like @cdjackson, who don't have the core development option selected for their IDE (and I don't think we want to make this a mandatory part?).

maggu2810:
> I didn't test it, but won't this again mess up our working set structure and cause all resources be available multiple times again?

As written in the topic there are "filters for subprojects" applied.
See e.g. https://github.com/openhab/openhab-core/pull/1023/files#diff-267c7db53673cb611d5c91164fdd974bR17-R48

> Wrt [openhab/openhab-distro#927 (comment)](https://github.com/openhab/openhab-distro/issues/927#issuecomment-525915184), I don't see how this helps contributors like @cdjackson, who don't have the core development option selected for their IDE (and I don't think we want to make this a mandatory part?).

What is the point?
I don't plan to add project files for every project some OH developer imports in its workbench.
I just want to ensure that all core projects are available for developers that use the core bundles in its environment.

PR #908, opened 2019-07-02
 Refactoring of builders

cweitkamp:
- [x] Refactoring of `StateDescriptionFragmentBuilder`
- [x] Switched `ChannelBuilderTest` to non OSGi test, added check for builder to create independent channels
- [x] Added unit test for `ThingStatusInfoBuilder`
- [x] Moved `ThingBuilderTest` to basic unit test and refactoring
- [x] Refactoring of `ThingBuilder` (and `BridgeBuilder`)

See https://github.com/openhab/openhab-core/pull/906#issuecomment-506859334

Signed-off-by: Christoph Weitkamp <github@christophweitkamp.de>

Issue #731, opened 2019-04-15
Use DTOs of Thing interface to store and restore the data


cweitkamp:
<a alt="bounty" href="https://www.bountysource.com/issues/72875689"><img src="https://api.bountysource.com/badge/issue?issue_id=72875689"></a>

> The storage for things does not use a data transfer object but the thing implementation itself to store and restore the data. So we are not allowed to change the members of the specific implementation

> We need first to use something like a ThingDTO for the storage similar as it has been done for rules.
>  But it will not be such easy as for rules the decision has been taken to break all existing storages.

_Originally posted by @maggu2810 in https://github.com/openhab/openhab-core/issues/717#issuecomment-482927908_

>> We definitely need a way - e.g. a service script - to adopt user storages after the introduction of a DTO.
>
> Absolutely! I am doing something like this in my solutions: maggu2810@b5c224e

_Originally posted by @cweitkamp in https://github.com/openhab/openhab-core/issues/717#issuecomment-482945783_

With the final implementation we can revert #719 to make use of #693.

PR #1027, opened 2019-09-10
[WIP] Fix thing creation for updated XML

afuechsel:
This is a new approach to solve the problem that I was trying to solve in #628 that in turn was based on another PR of @J-N-K (#527). 

It works as long as there are no user-defined channels, therefore it's still work in progress. I'd like to ask for our comments about this approach. 

`ThingManager#doInitializeHandler` calls `BaseThingHandler#setChannelsFromThingTypeDefinition` to only create channels, that are actually defined in the XML. This allows updated XML descriptions for already existing things to be taken into account.

Signed-off-by: André Füchsel <andre.fuechsel@telekom.de>

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseThingHandler.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseThingHandler.java
@@ -13,6 +13,7 @@
 package org.eclipse.smarthome.core.thing.binding;
 
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
@@ -20,10 +21,12 @@ import java.util.concurrent.ScheduledExecutorService;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.smarthome.config.core.ConfigDescriptionRegistry;
 import org.eclipse.smarthome.config.core.Configuration;
 import org.eclipse.smarthome.config.core.validation.ConfigValidationException;
 import org.eclipse.smarthome.core.common.ThreadPoolManager;
 import org.eclipse.smarthome.core.thing.Bridge;
+import org.eclipse.smarthome.core.thing.Channel;
 import org.eclipse.smarthome.core.thing.ChannelUID;
 import org.eclipse.smarthome.core.thing.Thing;
 import org.eclipse.smarthome.core.thing.ThingRegistry;
@@ -34,6 +37,8 @@ import org.eclipse.smarthome.core.thing.ThingTypeUID;
 import org.eclipse.smarthome.core.thing.ThingUID;
 import org.eclipse.smarthome.core.thing.binding.builder.ThingBuilder;
 import org.eclipse.smarthome.core.thing.binding.builder.ThingStatusInfoBuilder;
+import org.eclipse.smarthome.core.thing.internal.ThingFactoryHelper;
+import org.eclipse.smarthome.core.thing.type.ThingType;
 import org.eclipse.smarthome.core.thing.util.ThingHandlerHelper;
 import org.eclipse.smarthome.core.types.Command;
 import org.eclipse.smarthome.core.types.RefreshType;
@@ -633,6 +638,18 @@ public abstract class BaseThingHandler implements ThingHandler {
             logger.warn("Handler {} tried migrating the thing type although the handler was already disposed.",
                     this.getClass().getSimpleName());
         }
+    }
+
+    public void setChannelsFromThingTypeDefinition(ThingType thingType,
+            ConfigDescriptionRegistry configDescriptionRegistry) {
+        logger.info("Synchronizing channels for thing {}", getThing().getUID());
+        List<Channel> channels = ThingFactoryHelper.createChannels(thingType, getThing().getUID(),
+                configDescriptionRegistry);
+
+        ThingBuilder thingBuilder = editThing();
+        thingBuilder.withChannels(channels);
+        Thing editedThing = thingBuilder.build();
+        updateThing(editedThing);
     }
 
 }
--- a/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseThingHandler.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseThingHandler.java
@@ -644,7 +644,7 @@ public abstract class BaseThingHandler implements ThingHandler {
             ConfigDescriptionRegistry configDescriptionRegistry) {
         logger.info("Synchronizing channels for thing {}", getThing().getUID());
         List<Channel> channels = ThingFactoryHelper.createChannels(thingType, getThing().getUID(),
-                configDescriptionRegistry);
+                configDescriptionRegistry, getThing().getChannels());
 
         ThingBuilder thingBuilder = editThing();
         thingBuilder.withChannels(channels);
--- a/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/builder/BridgeBuilder.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/builder/BridgeBuilder.java
@@ -22,6 +22,7 @@ import org.eclipse.smarthome.config.core.Configuration;
 import org.eclipse.smarthome.core.thing.Bridge;
 import org.eclipse.smarthome.core.thing.Channel;
 import org.eclipse.smarthome.core.thing.ChannelUID;
+import org.eclipse.smarthome.core.thing.Thing;
 import org.eclipse.smarthome.core.thing.ThingTypeUID;
 import org.eclipse.smarthome.core.thing.ThingUID;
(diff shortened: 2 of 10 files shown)


cweitkamp:
Just FTR: Possible solution for https://github.com/eclipse/smarthome/issues/2555.

bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseThingHandler.java
```diff
@@ -635,4 +640,16 @@ protected void changeThingType(ThingTypeUID thingTypeUID, Configuration configur
...
     }
 
+    public void setChannelsFromThingTypeDefinition(ThingType thingType,
+            ConfigDescriptionRegistry configDescriptionRegistry) {
+        logger.info("Synchronizing channels for thing {}", getThing().getUID());
+        List<Channel> channels = ThingFactoryHelper.createChannels(thingType, getThing().getUID(),
+                configDescriptionRegistry);
+
+        ThingBuilder thingBuilder = editThing();
+        thingBuilder.withChannels(channels);
```
cweitkamp: IIRC this method clears the list of all existing channels. We have to filter the list of new channels to add only non existing ones. On the other hand we have to deal with removal of channels and changes in channel types too.

Be aware there currently is an issue regarding the `withoutChannels` for the `BridgeBuilder` - it returns a `ThingBuilder` instance - see https://github.com/openhab/openhab-core/pull/908/files#diff-44fdf6ebca4daa065ec6088297fb2859R82.

ivivanov-bg: > We have to filter the list of new channels to add only non existing ones

Why ?

Indeed this method clears all channels on the thing and replaces them with the one declared in the XML, but once the handler is initialized again - all other channels (added dynamically) will be re-added (it will behave just like if the thing is paired for a very first time)

cweitkamp: Yes, exactly but please keep in mind that channels can have a user specific configuration. You will throw away any of them when you re-add all the channels. We should try to keep them.

maggu2810: and there are channel that can be added by user interaction -- so manually per request. That information will be lost, too.

ivivanov-bg: Yes - I know - this is already mentioned in the description in the PR.

One question on that topic:
How are actually those channels used ?
How are the stated updated if the handler doesn't know about them. 
How commands sent to those channels handled ?

afuechsel:
Hmm, not sure why the build failed. Just pushed an update to my branch...

cweitkamp:
I think it failed because there are merge conflicts due to my recent builder refactoring (see #908).

afuechsel:
> I think it failed because there are merge conflicts due to my recent builder refactoring (see #908).

Maybe, but after a rebase I am not sure how to fix my code, because `BridgeBuilder` now cannot access the thing. I studied your PR about changing builders but I do not completely understand, why this change was necessary and what exactly is the improvement.

bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/builder/BridgeBuilder.java
```diff
@@ -101,4 +102,13 @@ public BridgeBuilder withLocation(@Nullable String location) {
         return (BridgeBuilder) super.withLocation(location);
     }
 
+    public BridgeBuilder withThings(@Nullable List<Thing> things) {
+        BridgeImpl bridge = ((BridgeImpl) thing);
```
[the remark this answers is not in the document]
afuechsel: Bear with me - I am short of time. I think, the only issue left open is the removal of channels during `editThing`, that are not defined in XML and therefore must not be removed. I have no clear idea yet, how to solve this.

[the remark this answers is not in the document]
wborn: It might help to add a field to the `Channel` class for tracking the origin. Then it can be removed if it was not created from the ThingType definition.

There are also some other border cases:
* The new ThingType defines a channel that was created by the user. In that case the user channel and configuration might get lost because it is replaced with that of ThingType.
* An existing channel is renamed. If the channel has configuration parameters its values will be lost when the old channel is removed.

I'm fine with skipping those border cases for now because I think they won't occur often and the changes in this PR would do more good than harm. :-)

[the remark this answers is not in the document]
cweitkamp: > It might help to add a field to the Channel class for tracking the origin.

Yes, that is true. But I am afraid it is not possible until we found a solution for #731.

btw: I prepared a new rebase for these changes including some unit tests. https://github.com/openhab/openhab-core/compare/master...cweitkamp:pr-1027

openhab-5iver:
How are things coming along with this PR? Any chance this could be added to the OH 3.0 backlog? Is that the intent of this project... https://github.com/orgs/openhab/projects/3?

cweitkamp:
> Any chance this could be added to the OH 3.0 backlog?

Done. I absolutely vote for it.

cdjackson:
> We have to filter the list of new channels to add only non existing ones. On the other hand we have to deal with removal of channels and changes in channel types too.

Just a comment on this - we also need to handle changes to channels. Channels have ```properties``` and we need to manage changes in these as well.

At some point in the past I had proposed to add a version number to the thing definition so that it was explicit that the definition had changed rather than trying to detect the all changes which might be problematic.

J-N-K:
Another year done and still no solution to it. The longer it takes, the more I am convinced that my original idea with versioned thing-types and upgrade-logic within the binding is a better and more viable solution than this general approach.

afuechsel:
@J-N-K Unfortunately it does not. I tried several times with your original solution and I was not able to get it to work completely. I was not able to get all tests green nor did the newly created tests work out well. In the runtime it only worked _sometimes_. This was due to the dynamics of the thing handler creation and their conflict with the ready marker triggers. Also the introduction of a queue of all not YET resolved thing types did not work out well. I have a 95% solution still as a branch. 

Therefore we decided to go with the now shown solution here and it worked out pretty well (with some modifications not yet part of this PR). In our solution that is still based on the frozen old ESH code base we now have a working implementation of it. 

I also tried to address the openHAB issue with user-created channels, something we don´t have in our solution. The easiest way was the introduction of a flag in the channels that indicate whether or not they have been created by the XMLs. 

Although I am not actively contributing to openHAB I was hoping I could eventually finish this PR but it seems I am lacking the time to. May be I could find some time to at least push the latest changes from our code base which works for us. Sorry for that.

J-N-K:
@afuechsel my original idea was implemented in the OneWire binding in ESH, before it got merged and removed there. It simply had a property `thingTypeVersion` for each thing-type, a number that has to be increased by the developer each time changes to the thing type are done. The thing handlers check the version number during initialization and upgrade the thing when necessary. Since the developer knows what changed, he can make the correct decision which channels need to be added, changed or removed and what needs to be kept.

afuechsel:
@J-N-K Feel free to revert the changes here and go with your idea. Maybe I made a mistake since the update with your implementation never worked for me. As I said, I have unfortunately not enough time to deal deeper with openHAB.

J-N-K:
IMO removing this from OH3 backlog is a big step backwards. It's now more than four years that we have to delete and re-create things after each update that adds new features.

kaikreuzer:
@J-N-K What do you mean? It is no step backward, it is just no step forward.

J-N-K:
It‘s a step backward to remove it from the backlog. Previously this was accepted as an issue to be solved for OH3.

kaikreuzer:
Keeping it in the backlog won't fix it, though. If you can come up with a PR and a fix for it in time for 3.0, I'll happily put it back.

J-N-K:
The biggest problem is that wedon't know which channels were originally added from the XML and which channels where added by the user (`extensible`). This could probably be solved in the REST call when the channel is added. This should be step 1 (and probably should be done before OH3, so that at least all the people starting from scratch will benefit later if we solved the other problems).

We also need to find a solution for already existing things (e.g. MQTT only has `extensible` channels), a lot of channels would be lost if the flag cannot be added for existing things. This needs to be done during the upgrade process. 

The third question is how to handle dynamically added channels. Removing and re-creating them during each startup is possible (and links should be preserved), but this is quite time-consuming and will probably impose bigger delays for larger installations until all things are configured.

If these issues are solved, it should be possible to go the route I originally proposed:

- read the stored thing
- create the same thing from scratch
- check which channels (that are not dynamically added and not extensible) are missing in the newly created thing, remove them and add those that are missing
- add thing to the thingmanager

wborn:
Hello @afuechsel, do you think you will find time to continue working on this PR?

afuechsel:
Unfortunately I no more time to work for openHAB. Feel free to close this PR or someone else could continue with it. It’s rather old now.

wborn:
Thanks for letting us know @afuechsel! :+1: 

I've archived you branch in my own [repo](https://github.com/wborn/openhab-core/tree/sync-thing-channels), so someone can pick this up in the future as it certainly looked like this was the most promising approach so far.

PR #1032, opened 2019-09-14
Mapview element by default for a location item

### LINKED ISSUE #1015: Default widget for location items: moving to mapview ?
By default, we are now using a text widget for a location item.
What do you thing to move to a mapview widget by default ?

lolodomo:
Fix #1015


lolodomo:
Tested with Basic UI, Classic UI and Android App.

I have set no height attribute, so each UI is defining its own default height for the map.

PR #1041, opened 2019-09-22
[core] Removed access to 'ThingRegistry' from 'BaseThingHandler'

cweitkamp:
- Removed access to `ThingRegistry` from `BaseThingHandler`

See https://github.com/eclipse/smarthome/issues/5182

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandlerCallback.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandlerCallback.java
@@ -160,10 +160,10 @@ public interface ThingHandlerCallback {
     /**
      * Returns the bridge of the thing.
      *
-     * @param bridgeUID {@link ThingUID} UID of the bridge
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandlerCallback.java
```diff
@@ -154,4 +157,13 @@
...
     boolean isChannelLinked(ChannelUID channelUID);
 
+    /**
+     * Returns the bridge of the thing.
+     *
+     * @param bridgeUID {@link ThingUID} UID of the bridge
+     * @return returns the bridge of the thing or null if the thing has no bridge
+     */
+    @Nullable
+    Bridge getBridge(@Nullable ThingUID bridgeUID);
```
maggu2810: IMHO it is not a good API design to accept a nullable argument here and (also my mean, your can vary) it is also not commonly used.

You define a method that give you the bridge for a UID and null if there is none.

Why should you accept a nullable UID?
The only reason is because you want to support lazy programming styles that just forward return value of one function to the argument of another one.

PR #835, opened 2019-05-27
discovery: Use constructor injection to simplify lifecycle

cweitkamp:
- Use constructor injection to simplify lifecycle

Fixes #834 

Signed-off-by: Christoph Weitkamp <github@christophweitkamp.de>

PR #1051, opened 2019-09-28
discovery: Fixed activation of inbox

cweitkamp:
- Fixed activation of inbox

After #835 and #836 the real "activation" failed because the method was not annotated.

Fixes https://github.com/openhab/openhab2-addons/issues/6144


kaikreuzer:
Jenkins failed with 2 errors in core.tests, but from the log I couldn't figure out which ones those should be... I suspect some hiccup rather than real failing tests, so let's merge anyway.

PR #1060, opened 2019-09-30
Removed orphaned 'registerHandlerLock'

### LINKED ISSUE #1022: ThingManagerImpl#migrateThingType will never work
Just stumbled across the following bug in the framework: https://github.com/openhab/openhab-core/blob/9724045bb60a6b3c31c7c2a91110c7a7d8b0e0ea/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java#L353 would never work, because it calls `waitForRunningHandlerRegistrations` which checks in https://github.com/openhab/openhab-core/blob/9724045bb60a6b3c31c7c2a91110c7a7d8b0e0ea/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java#L434 the `registerHandlerLock` for a handler. As this set is only created here: https://github.com/openhab/openhab-core/blob/9724045bb60a6b3c31c7c2a91110c7a7d8b0e0ea/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java#L348 and never written to this check will always fail. 

Originally, about 4 years in the past the implementation of `registerHandler` wrote to this set: 

```java 
        // this check is needed to prevent infinite loops while a handler is initialized
        if (!registerHandlerLock.contains(thingUID)) {
            registerHandlerLock.add(thingUID);
            ThingHandlerFactory thingHandlerFactory = findThingHandlerFactory(thing.getThingTypeUID());
            if (thingHandlerFactory != null) {
                registerHandler(thing, thingHandlerFactory);
            } else {
                logger.debug("Not registering a handler at this point since no handler factory for thing '{}' found.",
                        thingUID);
            }
            registerHandlerLock.remove(thingUID);
        } else {
            logger.warn("Attempt to register a handler twice for thing {} at the same time will be ignored.", thingUID);
        }
```  
Seems to be lost during some refactorings already back in the ESH era. @kaikreuzer could you please have a look here?

cweitkamp:
- Removed orphaned `registerHandlerLock`

As there is no place where a `ThingUID` will be added to the `registerHandlerLock` the `waitForRunningHandlerRegistrations` will always return immediately. Additionally the logic is wrapped by a second `Lock`. We can savely remove this orphaned code.

Closes #1022

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
@@ -350,8 +350,6 @@ public class ThingManagerImpl
 
     private final Set<Thing> things = new CopyOnWriteArraySet<>();
 
-    private final Set<ThingUID> registerHandlerLock = new HashSet<>();
-
     private final Set<ThingUID> thingUpdatedLock = new HashSet<>();
 
     @Override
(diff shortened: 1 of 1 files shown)


wborn:
Ping @cdjackson

cdjackson:
Sorry - I missed this. I'm not aware of any problem with thingType migration - the ZWave binding uses this extensively as all things are initially created with a generic type, and then changed later once the binding detects information from the device.

wborn:
Thanks for your confirmation! :+1:

PR #1067, opened 2019-10-02
Consider more than one value when applying the default value(s) for Things and Channels

cweitkamp:
- Consider more than one value when applying the default value(s) for Things and Channels

We have a really powerful normalization tool `ConfigUtil` for parameters. ~I tried to use this tool for determining the correct type of the default value which unfortunately did not work because the normalization tool returns the given object, if it was not possible to convert it. We expect `null` for the default value if we cannot convert it. Any ideas?~

My approach was not needed because the default value is always a `String`. I now moved the normalization of the default value and the method to apply the default configuration to the `ConfigUtil` tool. As s side effect it leads to a nice code clean-up / streamlining in the helper calsses.

TODOs:
- [x] Add unit test cases

Related to #998

### DIFF
--- a/bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
+++ b/bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
@@ -13,7 +13,6 @@
 package org.eclipse.smarthome.config.core;
 
 import java.math.BigDecimal;
-import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -48,16 +47,20 @@ public class ConfigUtil {
     private static final String DEFAULT_LIST_DELIMITER = ",";
 
     /**
-     * Maps the provided (default) value of the given {@link Type} to the corresponding Java type.
+     * Maps the provided (default) value of the given {@link ConfigDescriptionParameter} to the corresponding Java type.
      *
      * In case the provided value is supposed to be a number and cannot be converted into the target type correctly,
      * this method will return <code>null</code> while logging a warning.
      *
-     * @param parameterType the {@link Type} of the value (must not be null)
-     * @param defaultValue the value that should be converted (must not be null)
+     * @param parameter the {@link ConfigDescriptionParameter} wich default value should be normalized (must not be
+     *            null)
      * @return the given value as the corresponding Java type or <code>null</code> if the value could not be converted
      */
-    public static @Nullable Object normalizeDefaultType(Type parameterType, String defaultValue) {
+    public static @Nullable Object getDefaultValueAsCorrectType(ConfigDescriptionParameter parameter) {
+        return getDefaultValueAsCorrectType(parameter.getName(), parameter.getType(), parameter.getDefault());
+    }
+
+    static @Nullable Object getDefaultValueAsCorrectType(String parameterName, Type parameterType, String defaultValue) {
         try {
             switch (parameterType) {
                 case TEXT:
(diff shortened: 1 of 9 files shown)


bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
```diff
@@ -35,14 +40,89 @@
...
+     */
+    public static @Nullable Object normalizeDefaultType(Type parameterType, String defaultValue) {
+        try {
+            switch (parameterType) {
+                case TEXT:
+                    return defaultValue;
+                case BOOLEAN:
+                    return Boolean.parseBoolean(defaultValue);
+                case INTEGER:
+                    return new BigInteger(defaultValue);
```
cweitkamp: This is a major change as it returned a `BigDecimal` before which is imo wrong.

kaikreuzer: Not really - similar to the item types, we use `BigDecimal` for any number (be it int or float), so that consuming code can always cast to `BigDecimal` without having to worry.
The reason that we introduced a separation between INTEGER and DECIMAL for parameters was due to the JSON deserialisation - afair, integers always ended up being floats in Java (or something like that...).

bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
```diff
@@ -35,14 +40,89 @@
...
+    }
+
+    /**
+     * Applies the default values from a give {@link ConfigDescription} to the given {@link Configuration}.
+     *
+     * @param configuration the {@link Configuration} where the default values should be added (must not be null)
+     * @param configDescription the {@link ConfigDescription} where the default values are located (may be null, but
+     *            method won't have any effect then)
+     */
+    public static void applyDefaultConfiguration(Configuration configuration,
```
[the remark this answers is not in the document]
kaikreuzer: `limitToOptions`: Hm, hopefully there are no default values that aren't in the option list... I think requesting the (possibly dynamic) options might be too diffcult/expensive here, so I would not do this check and be optimistic.

`multipleLimit`: This would be easier to perform, but we have no clue if we might remove an important one and should have taken a different one. So again, I would just assume that the defaults are correctly defined and not do anything further here.

bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
cweitkamp: Good point.

We than have to be optimistic again because a decimal will be a valid default value for a integer config parameter. Should we maybe add a rounding strategy?

bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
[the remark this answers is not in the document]
cweitkamp: Yes, `limitToOptions` will be a quite a challenge. And we have the same issues with dynamic config parameters. AFAIK they will not validated too.

Regarding `multipleLimit` we maybe can log a warning instead of stripping them.

bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
kaikreuzer: Yes - maybe the same here: Log a warning and round it (down), so that no subsequent exceptions occur.

bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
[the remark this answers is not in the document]
kaikreuzer: Yes, logging a warning isn't a bad idea.

bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
cweitkamp: I changed it back to `BigDecimal`, added a warning and round down.

btw. do we have a proper solution to verify logger outputs in our unit tests?

bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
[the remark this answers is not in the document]
cweitkamp: Done.

bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
kaikreuzer: I am not aware of any such solution...

PR #1068, opened 2019-10-02
Move audio unit tests

wborn:
All audio tests are unit tests so there is no need for an itests project.


cweitkamp:
Great job. One question for all your move-the-tests-pull-requests: Did you consider the package of the class to be tested? I saw that some of them were in wrong locations.

wborn:
That's a good point @cweitkamp. So I've checked the tests in the bundles that all my PRs touched and made some fixes where applicable. :-)

PR #1070, opened 2019-10-02
Move core unit tests

wborn:
Moves the core unit tests from the itests project to the bundle project.


wborn:
Because of a cycle it's currently not possible to add a test dependency on `org.openhab.core.test` in `org.openhab.core`. So classes depending on that project currently remain in the itests project.

PR #1072, opened 2019-10-02
Bump compile version of jetty to 9.4 (equal to runtime)

ppieczul:
Current jetty version is missing ability to initialize WebSocketClient with HttpClient object that can be reused from the framework by Thing factory.
See this thread for more details: https://github.com/openhab/openhab2-addons/issues/6013
I would like to use this initialization in Loxone binding and get rid of custom thread pool.

Signed-off-by: Pawel Pieczul <pieczul@gmail.com>

### DIFF
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -16,7 +16,7 @@
   <description>The dependencies that are used to compile the core bundles</description>
 
   <properties>
-    <jetty.version>9.3.25.v20180904</jetty.version>
+    <jetty.version>9.4.11.v20180605</jetty.version>
   </properties>
 
   <dependencies>


wborn:
I think we didn't bump it because we only updated the versions that the OH distribution depends on. For compilation we also don't use the newest version of Karaf:

https://github.com/openhab/openhab-core/blob/807593aa94795eaf51fd6d63452bb429a03f4f04/pom.xml#L72-L73

IIRC @maggu2810 added that differentiation so it causes less issues with other distributions using OHC.

cweitkamp:
I see. So there is a reason.

As a workaround @ppieczul could add the jetty bundles with the required new version as dependencies to his bindings pom.

ppieczul:
Ok, I will add the new version to bindings pom, I think only for compile scope. Thanks.

PR #1077, opened 2019-10-02
Move voice unit tests

wborn:
Moves the voice unit tests from the itests project to the bundle project.


cweitkamp:

PR #1078, opened 2019-10-02
Move UI unit tests

wborn:
All UI tests are unit tests so there is no need for an itests project.


cweitkamp:

PR #1081, opened 2019-10-02
Move IO HTTP unit tests

wborn:
All IO HTTP tests are unit tests so there is no need for an itests project.


cweitkamp:

wborn:
Thanks for the quick reviews and merge @cweitkamp! There are still a few itests projects I have to check but the majority has been done now. :-)

PR #1097, opened 2019-10-05
Add STOP to IncreaseDecreaseType

cdjackson:
Maybe there's another way to do this, but currently an ```IncreaseDecreaseType``` can only have ```INCREASE``` and ```DECREASE``` - there seems no way to  set this to ```STOP``` to stop the movement.

It seems that this should therefore be a useful addition - I don't think it should be a breaking change. I'm happy to hear if there's an alternative to this if this change isn't needed, but I'd like to be able to use this in a ```Dimmer``` when someone pressed a button, and then to stop the movement when someone releases the button. This directly maps to a ZigBee concept which uses a ```Move``` command (with an Increase / Decrease parameter), and a ```Stop``` command.

https://community.openhab.org/t/knx-dpt-3-increasedecreasetype/18401

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/library/types/IncreaseDecreaseType.java
+++ b/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/library/types/IncreaseDecreaseType.java
@@ -23,7 +23,8 @@ import org.eclipse.smarthome.core.types.PrimitiveType;
 @NonNullByDefault
 public enum IncreaseDecreaseType implements PrimitiveType, Command {
     INCREASE,
-    DECREASE;
+    DECREASE,
+    STOP;
 
     @Override
     public String format(String pattern) {


kaikreuzer:
The semantics of INCREASE command is "increase by a step" and not "start increasing".
Hence, a STOP command does not make sense here.

The behaviour you describe (button press/release) is best done through the [RawRockerDimmerProfile](https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawRockerDimmerProfile.java) (typically done e.g. by the EnOcean binding).

For the dimmer/device side, if you do not want to send out individual commands to the device (as it might not be smooth enough), you can still implement your binding in a way that it starts dimming when receiving INCREASE commands and automatically stop it, if there aren't any further ones after 300-400ms).

cdjackson:
Thanks Kai,
I don’t know that I would agree (sorry). Even if this is a step, you might want to stop it half way through the step (again, this is possible with the ZigBee Step command) - it depends on the size of the step.

If I understand the profile, it will send an IncreaseDecrease command, but is it also sending OnOff when the buttons are released? Maybe @wborn can comment as I think you wrote this?

```
     if (CommonTriggerEvents.DIR1_PRESSED.equals(event)) {
            buttonPressed(IncreaseDecreaseType.INCREASE);
        } else if (CommonTriggerEvents.DIR1_RELEASED.equals(event)) {
            buttonReleased(OnOffType.ON);
        } else if (CommonTriggerEvents.DIR2_PRESSED.equals(event)) {
            buttonPressed(IncreaseDecreaseType.DECREASE);
        } else if (CommonTriggerEvents.DIR2_RELEASED.equals(event)) {
            buttonReleased(OnOffType.OFF);
        }
```

The times here are far too large - I guess I can write another profile, but it seems a very complex system - I'd need to write a profile to crease a timer that periodically sends INCREASE commands. Then in the binding I need to create another timer to detect if the INCREASE commands stop coming through.

Is there really a problem if we add the STOP? This issue seems to exist for all the UpDown IncreaseDecrease types.and as far as I can find this concept isn’t documented anywhere either. Another option I thought of is to use the ```UnDef``` type, but this is not available as a command.

Maybe it’s a question for OH3, but I think some of these types could really do with an overhaul to make the more generic, and more useful. Maybe we can reopen this while it's discussed?

wborn:
> I think you wrote this?

I've only touched [RawRockerDimmerProfile](https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawRockerDimmerProfile.java) with some cleanup PRs. It was added in https://github.com/eclipse/smarthome/pull/4621 by Jan (@kungfoolfighting). Perhaps he is still around to answer your questions? That PR also has some comments which may add some context to the current code.

cdjackson:
Thanks @wborn - I forgot this was copied from ESH and I just saw that your name was at the top ;)

Unfortunately the comments in the PR don't really help too much.

kaikreuzer:
> Even if this is a step, you might want to stop it half way through the step 

No, steps are discrete commands that are processed. ANY command will cause some transition from one state to another and we don't have any concept to stop or revert such transitions.
And as INCREASE/DECREASE commands are expected to come roughly every 300ms, I don't think there is any need to stop a dimming activity quicker than this.

> but is it also sending OnOff when the buttons are released?

It is sending ON/OFF on short button press and INCREASE/DECREASE on long press. Just as you would expect a dimmer to behave if you have a rocker.

> Is there really a problem if we add the STOP?

As explained, it does not fit the concepts here at all. There simply is nothing to "stop". And we cannot simply change the semantics of INCREASE/DECREASE.
Besides this, we have >100 bindings that implement dimmer channels, but don't know about such a command and which clearly expect the current semantics.

> I guess I can write another profile

My suggestion was not to write another profile, but to simply deal with the existing commands in the binding. When receiving a first INCREASE command, you can start your activity and if no further such commands are received within 500ms, you can stop the activity again. This should make the UX pretty smooth.

cdjackson:
Ok, if there’s no scope for creating a state full type that allows a start/stop concept, then I would propose to reduce the period if the commands. With the current delay I’m finding on a slow machine that the period can be 400ms - occasionally more, so the binding really needs to be waiting for at least 500ms between commands to prevent false Stop commands being sent. 500ms is really too long and I’d propose to reduce the period to something like 150ms - is that ok with you?

PR #1106, opened 2019-10-08
Fix SAT constant/variable name findings

wborn:
Fixes wrong names identified by the following rules:

* ConstantNameCheck
* LocalFinalVariableNameCheck
* LocalVariableNameCheck
* StaticVariableNameCheck

Most mismatches identified by the MemberNameCheck have also been fixed except for those where the variables are used in events/DTOs etc which would cause issues.


wborn:
This was done manually. All variable references also need to be updated so creating something to automate it might have taken more time.

PR #1112, opened 2019-10-10
Changed pid from 'org.eclipse.smarthome.core.i18nprovider' to 'org.eclipse.smarthome.i18n'

cweitkamp:
- Changed pid from 'org.eclipse.smarthome.core.i18nprovider' to 'org.eclipse.smarthome.i18n'

### DIFF
--- a/itests/org.openhab.core.binding.xml.tests/src/main/java/org/eclipse/smarthome/core/binding/xml/test/BindingInfoI18nTest.java
+++ b/itests/org.openhab.core.binding.xml.tests/src/main/java/org/eclipse/smarthome/core/binding/xml/test/BindingInfoI18nTest.java
@@ -89,20 +89,31 @@ public class BindingInfoI18nTest extends JavaOSGiTest {
     public void assertUsingDefaultLocale() throws Exception {
         // Set german locale
         ConfigurationAdmin configAdmin = getService(ConfigurationAdmin.class);
+        assertThat(configAdmin, is(notNullValue()));
+
         Configuration config = configAdmin.getConfiguration("org.eclipse.smarthome.i18n", null);
-        Dictionary<String, String> localeCfg = new Hashtable<>();
-        localeCfg.put("language", "de");
-        localeCfg.put("country", "DE");
-        config.update(localeCfg);
+        assertThat(config, is(notNullValue()));
+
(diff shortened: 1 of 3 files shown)


itests/org.openhab.core.binding.xml.tests/src/main/java/org/eclipse/smarthome/core/binding/xml/test/BindingInfoI18nTest.java
```diff
@@ -89,7 +89,7 @@ public void assertUsingOriginalBindingInfosIfProvidedLocaleIsNotSupported() thro
     public void assertUsingDefaultLocale() throws Exception {
         // Set german locale
         ConfigurationAdmin configAdmin = getService(ConfigurationAdmin.class);
-        Configuration config = configAdmin.getConfiguration("org.eclipse.smarthome.core.i18nprovider", null);
+        Configuration config = configAdmin.getConfiguration("org.eclipse.smarthome.i18n", null);
         Dictionary<String, String> localeCfg = new Hashtable<>();
         localeCfg.put("language", "de");
         localeCfg.put("country", "DE");
```
cweitkamp: This parameter does not have any effect. Should be `"region"` instead.

wborn:
The change itself is good because all other PIDs also don't use the ".core" part.

Will this change require everyone to again reconfigure their "i18nprovider" settings? If so, is there a way to prevent this? Looks like this configurationPid was [already used in ESH](https://github.com/eclipse/smarthome/blob/b8455de15d65512e8fac4e94d42de6ab2fccf1c5/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/i18n/I18nProviderImpl.java#L81-L83).

When we update all configurationPids from ESH to OHC it would also be nice to automate this for smooth migrations.

cweitkamp:
Will this change require everyone to again reconfigure their "i18nprovider" settings? If so, is there a way to prevent this?

Yes it does. I hope we can fin a way to use the `MOVE` file feature of the [distro update](https://github.com/openhab/openhab-distro/blob/master/distributions/openhab/src/main/resources/bin/update) script (see also https://github.com/openhab/openhab-distro/pull/979#issuecomment-540426507).

PR #1117, opened 2019-10-11
[rulehli] Moved config description  into 'src/main/resources/' folder

### LINKED ISSUE #1115: Rule Voice Interpreter: missing setting in Paper UI
Paper UI is showing no settings in configuration > services > voice, the block is here but showing no available settings. We should have a setting "item".
As a consequence, the interpreter is no more working.

Can someone please confirm as my server is now patched for voice stuff (but not this part).

cweitkamp:
- Moved config description into `src/main/resources/` folder

Closes #1115


lolodomo:
Yes I just saw that the config was not in the right directory.

lolodomo:
I let someone else validating the change in class path.

bundles/org.openhab.core.model.script/.classpath
```diff
@@ -18,11 +18,6 @@
 			<attribute name="maven.pomderived" value="true"/>
 		</attributes>
 	</classpathentry>
-	<classpathentry including="**" kind="src" output="target/classes" path="resources">
```
kaikreuzer: Hm, as this is a "model" bundle, we have "src" and "resource" as source folders and not "src/main/java" and "src/main/resources". I hence think that this PR is not correct.

lolodomo:
At least, with these changes, I can confirm the setting is again dispayed.

kaikreuzer:
So the question is, why is the ESH-INF folder not copied to https://ci.openhab.org/job/openHAB-Core/ws/bundles/org.openhab.core.model.script/target/classes/, although the `resources` folder is added as a source folder in the project's pom.xml?

bundles/org.openhab.core.model.script/.classpath
cweitkamp: Just as a side note - this is the last OHC bundle where the `resources/` folder is not located in`src/main/`.

Does it matter the classpath entry uses `including="**"` whereas all other bundles use `excluding="**"` in this place?

cweitkamp: I stick to this PR. In any other case I got this message:

```
[INFO] --- maven-resources-plugin:3.0.2:resources (default-resources) @ org.openhab.core.model.script ---
[INFO] Using 'UTF-8' encoding to copy filtered resources.
[INFO] skip non existing resourceDirectory ~/openhab-master/git/openhab-core/bundles/org.openhab.core.model.script/src/main/resources
```

and the config files are not copied into `target/classes/`

lolodomo:
Is there any conlusion ?
It is more than annoying to not be able to use the rule HLI.

cweitkamp:
@lolodomo You can put the configuration `org.eclipse.smarthome.rulehli:item=<ITEM_NAME>` for the service into your `services/runtime.cfg` file.

bundles/org.openhab.core.model.script/.classpath
kaikreuzer: Shouldn't it be as easy as https://github.com/openhab/openhab-core/pull/1160?
This makes Maven consider the "resources" folder and put its content into the jar.
Yes, the "skip non existing..." message will be shown, but this is only an INFO as Maven always checks for its default resource folder to be present and it wants to make sure that the user hasn't accidentially forgotten it. We can just ignore this message.

Issue #1118, opened 2019-10-11
Extension of the Human Language Interpreter


lolodomo:
This topic follows the discussion started in #734 .
The general idea is:

1. extend the core framework to support chat dialog and include the good generic stuff from HAbot NLP
2. make openNLP a new OH2 binding that will support chat dialog and voice control. This avoids a dependency to opennlp in the core framework
3. update HABot to support these changes without any regression

Here are the detailed steps and their status:

1. Core framework: extend the human language interface to support chat, include the card registry from HABot, include the stuff with Intent and Skill from HABot NLP and update the REST API. Status: **done** - PR #1122
2. HABot: update it to be compatible with the core framework changes and to have NLP packages not directly referenced by the HABot back-end. Status: **done** - PR openhab/openhab-webui#126
3. HABot: add a setting to select the interpreter. Status: **done** - included in PR 126
4. OH2: create a voice binding from HABot OpenNLP. Status: **done** - PR openhab/openhab2-addons#6264
5. HABot: remove embedded OpenNLP. State: waiting for the steps 1,2,3,4 to be merged.
6. OH2: extend the OpenNLP binding to make it fully compatible with voice control.
7. OH2: moving to a more recent version of OpenNLP. Status: waiting for the step 5 to be merged.
8. HABot: cleanup the REST internal interface and update the front-end to use the core REST API when the API are available in the core framework.
9. Core framework: make the built-in interpreter compatible with intent/skill and usable with HABot as an alternative to OpenNLP

ghys:
> update HABot to support these changes without any regression

This, of course, is a must as far as I'm concerned...

> make openNLP a new OH2 binding that will support chat dialog and voice control

You'll note that some skills in HABot make little sense in voice-only scenarios i.e. "get-status", "create-rule", the charting skills etc. - they provide natural language answers for increased "friendliness" but suppose the user will get a card. This was the main reason behind the design decision to include the card building as the responsibility of the skills. I'm not saying that decision can't be challenged, just stating some facts and explaining the thought process behind it.

lolodomo:
Yes, I azlready introduced in the skill a way to distinguish intents suitable for chat and/or voice control.
Of course, the charting skills are only suitable for chat.
The "get-status" skill is THE example that requires a new alternative skill for voice control, just because we need a different answer including the requested value.
But other skills like "activate-object", "deactivate-object" and "set-value" are already ready for voice control too, maybe with a minor change in the answer message.
I will do that in step 3.

[...]

PR #1122, opened 2019-10-12
Voice: extension for chat dialog + intent/skill added for HLI

lolodomo:
HLI interface extended for chat and intent interpretation
Cards management, item resolver from semantics and intent/skill interfaces retrieved from HABot
Voice REST API enhanced accordingly

Relative to #1118


lolodomo:
@ghys for review

lolodomo:
I stop now updating the code as it looks to me ready for a review.
Can you please review and merge this PR so that I can continue my long journey with voice control ?

lolodomo:
One last integration test was failing due to a missing dependency, I hope the build will be OK now.

lolodomo:
Build is now OK.

wborn:
Can you also help us with reviewing this @ghys?

ghys:
Okay, I'll bite; first I'll offer some general remarks on the functionality that is being added here instead of a pure code review.

1. First, this moves HABot's [components/slots and card system](https://community.openhab.org/t/habot-walkthrough-5-n-the-card-designer/55759) - cards are simply the root-level component of a tree with additional attributes and a registry + supporting storage provider - and their API to the core as a non-optional new concept in a "chat" package, with next to no changes; nothing inherently wrong with that, but I see several fundamental problems with it though because it feel "forced" and not generic enough as it should be for a core feature:

    * the API is really too much oriented towards HABot-specific features, notably the mechanics of card management - bookmark/unbookmark, the "card deck", timestamps, ephemeral cards etc.. Who knows if these concepts would be relevant to other UIs (not only web/chatbots, think also voice-controlled devices with a screen like the Echo View or Google Nest Hub - the very concept of a "card" as a visual support to an natural language answer is not, in fact, guaranteed). Even for UIs which _may_ have a similar card concept accompanying a reply, like Facebook Messenger or Telegram, their API and capabilities may differ significantly from what's specified here.
A generic `InterpretationResult` should therefore IMO have a way of providing a generic `Object` (or a generic interface), not a `Card`, defining what this object will actually be should remain the responsibility of implementation of an interpreter and its dedicated UI. Which leads to another remark: I also believe decoupling an interpreter and its skills from its intended UI is a recipe for nightmares, because the nature of the skills that an interpreter offers does depend heavily on what the end-user UI can offer, i.e. there is no point in having a skill that presents a chart if the UI has no charting capabilities, or if there's no visual UI at all. I'll address the _"suitableFor{Voice|Chat}/interpretFor{Voice|Chat}"_ solution to this problem proposed in this PR in my second point.

    * Cards designed for HABot's UI might be incompatible with other UIs; therefore providing a central storage and management API in a common core registry with no way of telling which end-user interface (or component framework) they're targeting does not really work. There could be an initiative to have some kind of convention on a set of common components and properties that every UI must implement (like a container, a list, a label, a slider, a toggle and so on), but restricting it too much limits the ability to use the underlying UI frameworks to their full potential.

    * I do however believe the components/slots concepts are worthy of being given first-class citizen status in the core, that is, move them out of the `o.e.s.c.voice.chat` package to something more generic and less "chat-oriented". Then, several types of components could be defined with these concepts: UI widgets for other UIs not related to voice control, maybe dashboards or managed sitemap replacements, etc. It feels odd having a card registry and `cards.json` JSONDB specifically for the chatbot use case, even without any chatbot addon installed. UI widgets implementations reusing the components/slots concept could then possibly either use a (tbd) central registry to manage & store their top-level components with a "namespace" to tell them apart from other implementations - or they could provide their own.

2. The duality of having skills "suitable for voice" and/or "suitable for chat" (which here means natural language answer + card) is IMO way too restrictive and feel like a attempt to add the HABot use case on top of the voice-only use case, see the examples above with voice-control + fullscreen visual feedback à la Echo Dot, or the need to make sure the UI can handle charts before offering skills that deal with charts. Another example: if there's a visual feedback, can it be interacted with? This information is needed before a skill can provide interactive UI elements like buttons or switches.
It should be up to the UI to provide the skills they support, which doesn't mean they can't be reused between them - for instance, HABot could implement a "activate object" skill which would (by subclassing or instantiating an "inner" skill) reuse the voice-control-only "activate object" skill, letting it do the heavy lifting, and simply add its card at the end. Additional HABot skills absolutely needing a card to make sense would be defined alongside, in the HABot bundle, and added to the interpreter's trainer using OSGi dependency injection.
In fact, I don't think we can share an interpreter instance for voice only _and_ "chatbot w/ card" ("`interpretForVoice` vs. `interpretForChat`") scenarios because you need specific training for the specific set of outcomes you want. Indeed, in the OpenNLP interpreter example, the set of intents define how the categorizer is trained and therefore what its possible outcomes can be - and you don't want the categorizer's outcome to be an intent your UI doesn't support. If you want to use the OpenNLP interpreter for both the voice-only and chatbot use cases at the same time, you therefore need two separate trainer/interpreter instances for the different set of skills supported by these two use cases.

Sorry for the long post, I hope it's clear enough, and provide some hindsight and explanation why I architectured HABot as a monolithic self-contained end-user product (just so you know, decoupling it into multiple bundles was considered at first), not only because I liked the "one-click install" of a single bundle, or because the maintenance is easier in a single-code owner scenario, but also because I didn't have answers to all these challenges at the time.

lolodomo:
The original intention was to get OpenNLP as a voice HLI without duplicating the work already done in HABot.
Considering all your remarks, I understand now that the best solution is just to create a new OpenHAB HLI as an OH2 binding and keep the one embedded in HABot as it is. The code will be  at 75% the same.
For Snips integration, I will reuse your good ideas with intent and skiils by creating similar objects. Same for the item resolver, I will duplicate your code.
So no absolute need to enhance the core framework, I close this PR.

ghys:
Sorry @lolodomo for reacting this late. The short version of what I had in mind is somewhat along those lines:

Core:
------
- `Component` can go in there (in a generic non-voice package), if @openhab/core-maintainers agree
- `Intent` & `Skill`
- `InterpretationResult` provides a generic object, not a `Card` (which is too specific)

OpenNLP HLI add-on
------------
- `IntentTrainer`
- OpenNLP HLI
- Skills compatible with voice-only control

HABot bundle
----------------
- dependency to the OpenNLP HLI
- `Card` _extends Component_ (in core)
- `CardRegistry`, `CardProvider` and the card management REST API (for now, don't know how to properly make a generic component registry in core, so this can stay there - other registries/providers and their API can be provided for other types of top-level Component derivatives)
- `HABotSkill` _extends Skill_
- `HABotIntentTrainer` _extends IntentTrainer_
- Skills which only make sense with a Card _extends HABotSkill_
- `GenericHABotSkill<T extends Skill>` _extends HABotSkill_ which is a wrapper for a voice-only skill but adds a Card with the matched items afterwards
- An interpreter service which would be able to collect all the voice-only skills (w/ dependency injection?) and build GenericHABotSkills with them, also add all the specific card-based HABotSkills defined in this bundle, and feed them all to the `HABotIntentTrainer`

If it makes sense :)

lolodomo:
This will require a dependency between openhab-webui and openhab2-addons and make the change in HABot very complicated (looks like).

I assume your idea is to change the interpret method in the HLI to return a InterpretationResult rather than a simple String ?

Maybe it is finally simpler to change nothing in core and HABot and I will just create a new binding for OpenNLP voice control, taking from HABOT what can be reused.

PR #1125, opened 2019-10-13
move MQTT client from Paho to HiveMQ

J-N-K:
Supersedes #929.


J-N-K:
Should we define a netty core-feature?`We have a lot of bundles that use netty and - at the moment - all can use `4.1.34.Final`.

wborn:
Thank for pushing this forwards. :+1: 

I think adding a netty feature is a good addition. Personally I would go for an additional constructor if the only configuration option is v3/v5. To future proof choosing between many options using a builder might be another option. But it's probably overkill if there's just this one configuration option.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -847,38 +839,44 @@ protected boolean finalizeStopAfterDisconnect(boolean v) {
...
      *
      * @param topic The topic
      * @param payload The message payload
+     * @param qos The quality of service for this message
+     * @param retain Set to true to retain the message on the broker
      * @param listener A listener to be notified of success or failure of the delivery.
      */
-    public void publish(String topic, byte[] payload, MqttActionCallback listener) {
-        publish(topic, payload, qos, retain, listener);
+    public void publish(String topic, byte[] payload, @Nullable Integer qos, @Nullable Boolean retain, MqttActionCallback listener) {
```
davidgraeff: Hm. How about forcing the API consumer to set qos and retain? Would that be soo bad?

The argument comments could be "qos The quality of service for this message. **0 is sufficient for most cases**" and "retain Set to true to retain the message on the broker. **In most cases you want to set this to false**".

The api consumer has to specify "null" at the moment anyway, there is not even a typing benefit.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -847,38 +839,44 @@ protected boolean finalizeStopAfterDisconnect(boolean v) {
...
+        final MqttAsyncClientWrapper client = this.client;
+        if (client == null) {
+            listener.onFailure(topic, new MqttException(new Throwable()));
+            return;
+        }
+
+        // use defaults when not provided by caller
+        qos = qos == null ? this.qos : qos;
+        retain = retain == null ? this.retain : retain;
```
davidgraeff: You are using deprecated class fields in this non-deprecated method.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -890,7 +888,7 @@ public void publish(String topic, byte[] payload, MqttActionCallback listener) {
      *         exceptionally on an error or with a result of false if no broker connection is established.
      */
     public CompletableFuture<Boolean> publish(String topic, byte[] payload) {
```
davidgraeff: Mark this deprecated maybe? Implicitly it uses deprecated class fields ("retain", "qos").

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -900,23 +898,30 @@ public void publish(String topic, byte[] payload, MqttActionCallback listener) {
...
         // publish message asynchronously
-        CompletableFuture<Boolean> f = new CompletableFuture<>();
-        try {
-            client.publish(topic, payload, qos, retain, f, actionCallback);
-        } catch (org.eclipse.paho.client.mqttv3.MqttException e) {
-            f.completeExceptionally(new MqttException(e));
-        }
-        return f;
+        CompletableFuture<Boolean> future = new CompletableFuture<>();
+        client.publish(topic, payload, retain, qos).whenComplete((m, t) -> {
```
davidgraeff: Why can't you use the clients future directly as return value? Maybe add the (unnecessary) boolean argument to the future to make it compatible?

This conversion stuff happening here right now is probably more inefficient than carrying a useless boolean in the type signature IMO.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/ClientCallback.java
```diff
@@ -64,14 +66,32 @@ public synchronized void connectionLost(@Nullable Throwable exception) {
...
+            logger.error("MQTT message received. MqttMessageSubscriber#processMessage() implementation failure", e);
+        }
     }
 
-    @Override
-    public void messageArrived(String topic, MqttMessage message) {
-        byte[] payload = message.getPayload();
+    public void messageArrived(Mqtt5Publish message) {
+        byte[] payload = message.getPayloadAsBytes();
+        String topic = message.getTopic().toString();
```
davidgraeff: Right now there is code duplication for those two `messageArrived` methods. How about an internal third one `void messageArrived(byte[] payload, String topic)` that gets called by the two others?

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/sslcontext/SSLContextProvider.java
```diff
@@ -24,6 +24,7 @@
  *
  * @author David Graeff - Initial contribution
  */
+@Deprecated
 public interface SSLContextProvider {
```
davidgraeff: Using this interface is a NO-OP I guess? Deprecating something means, that it still works but will be removed in the future. If this is a NO-OP now with the custom trust manager, an interface comment "DO NOT USE. This interface does nothing." might help to clarify?

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -206,12 +221,50 @@ public MqttBrokerConnection(String host, @Nullable Integer port, boolean secure,
...
+     * @param host A host name or address
+     * @param port A port or null to select the default port for a secure or insecure connection
+     * @param secure A secure connection
+     * @param clientId Client id. Each client on a MQTT server has a unique client id. Sometimes client ids are
+     *            used for access restriction implementations.
+     *            If none is specified, a default is generated. The client id cannot be longer than 65535
+     *            characters.
+     * @throws IllegalArgumentException If the client id or port is not valid.
+     */
+    public MqttBrokerConnection(MqttVersion mqttVersion, String host, @Nullable Integer port, boolean secure,
```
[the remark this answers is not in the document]
J-N-K: Agreed.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -661,95 +714,47 @@ MqttConnectOptions createMqttOptions() throws ConfigurationException {
...
-        MqttAsyncClient localClient;
-        try {
-            localClient = createClient(serverURI.toString(), clientId, localDataStore);
-        } catch (org.eclipse.paho.client.mqttv3.MqttException e) {
-            future.completeExceptionally(new MqttException(e));
-            return future;
-        }
+        MqttAsyncClientWrapper client = createClient();
 
+        logger.info("client created");
```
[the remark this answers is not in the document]
J-N-K: Done.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
J-N-K: Removing the nullable makes it much more difficult to maintain backward compatibility with the QoS and retain parameters on the connection.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
J-N-K: No way to remin backward compatible otherwise. When we remove the retain-setter/getter, we can default that to false (what is the default now, if not explicitely set otherwise).

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
J-N-K: It'll use the default (QoS 0 and retain false) after the getter/setter is removed.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
J-N-K: The clients return different futures: CompletableFuture<Mqtt3Publish> and CompletableFuture<Mqtt5Publish>.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/ClientCallback.java
J-N-K: Done

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/sslcontext/SSLContextProvider.java
J-N-K: It still works. I changed the two implementations so that this is fully transparent.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
davidgraeff: To stay backwards compatible all you need to offer is the deprecated `publish` method without qos and state. And in that one you would call `publish(topic, payload, this.getQos(), this.getRetain())`.

That's why I do not understand why you need nullable objects here. Maybe I'm overlooking something?

J-N-K: QoS is not deprecated on connection level. If we decide to do that, we can keep both as @NonNull. The reason for the @Nullable is to make it possible to set retain with the defualt qos (qos=null) or change the qos and  keep the retained (which is deprecated but should still work). The other option is to have three seperate methods instead of one (qos, retain,qos+retain).

davidgraeff: Ah ok, your argument is that you may want to set only one of the parameters.
But then again, you can just use the getter of the other parameter to get the default, eg:

`conn.publish(topic, payload, conn.getQos(), the_retain_value)`

J-N-K: good point.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
J-N-K: While we are discussing deprecation: What is the use-case for `subscribeRaw` and `unsubscribeRaw`? If there are no subscribers, why should the connection subscribe to a topic?

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -847,38 +839,44 @@ protected boolean finalizeStopAfterDisconnect(boolean v) {
...
-            return;
-        }
-        try {
-            IMqttDeliveryToken deliveryToken = localClient.publish(topic, payload, qos, retain, listener,
-                    actionCallback);
-            logger.debug("Publishing message {} to topic '{}'", deliveryToken.getMessageId(), topic);
-        } catch (org.eclipse.paho.client.mqttv3.MqttException e) {
-            listener.onFailure(topic, new MqttException(e));
-        }
+    public void publish(String topic, byte[] payload, MqttActionCallback listener) {
```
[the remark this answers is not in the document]
J-N-K: done

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
davidgraeff: I think it had to do keep alive subscriptions. But not sure anymore, I should have documented those methods better. Did you encounter any usages?

J-N-K: Not on openhab2-addons and not in openhab-core

davidgraeff: Deprecating or better removing should be fine then. I also do not see any benefit in keeping this API.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -438,6 +445,7 @@ public void setLastWill(@Nullable MqttWillAndTestament lastWill, boolean applyIm
      *
      * @param lastWill The last will object or null.
      */
+    @Deprecated
```
davidgraeff: Is is the LW handled now instead? Could you maybe if time permits add a documentation line

J-N-K: Added errorneously.

J-N-K:
@openhab/core-maintainers If we postpone this again, we‘ll be too close to the next milestone to merge it. And I would feel uncomfortable to add this to a release without having it tested in a milestone before.

kaikreuzer:
I'm ok with merging, but I would like to see @davidgraeff's final approval on it.
Plus, it seems you need to rebase :-)

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -513,17 +582,19 @@ public boolean hasSubscribers() {
...
-            MqttAsyncClient client = this.client;
-            if (client != null && client.isConnected()) {
-                client.subscribe(topic, qos, future, actionCallback);
-            } else {
-                future.complete(false);
-            }
-        } catch (org.eclipse.paho.client.mqttv3.MqttException e) {
-            logger.info("Error subscribing to topic {}", topic, e);
-            future.completeExceptionally(e);
+        CompletableFuture<Boolean> future = new CompletableFuture<Boolean>();
```
cweitkamp: Diamond operator.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -513,17 +582,19 @@ public boolean hasSubscribers() {
...
+        final MqttAsyncClientWrapper client = this.client;
```
cweitkamp: Can you rename the local variable?

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -513,17 +582,19 @@ public boolean hasSubscribers() {
...
+        if (client != null && client.getState().isConnected()) {
+            client.subscribe(topic, qos, clientCallback).whenComplete((s, t) -> {
+                if (t == null) {
+                    future.complete(true);
+                } else {
+                    logger.info("Error subscribing to topic {}", topic, t);
```
cweitkamp: If this is an error, why not log an error?

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -550,7 +621,7 @@ public boolean hasSubscribers() {
             // Remove from subscriber list
             subscribers.remove(topic);
             // No more subscribers to this topic. Unsubscribe topic on the broker
-            MqttAsyncClient client = this.client;
+            MqttAsyncClientWrapper client = this.client;
```
cweitkamp: See above: rename local variable.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/ssl/CustomTrustManagerFactory.java
```diff
@@ -0,0 +1,83 @@
...
+            // get trustmanager
+            Field trustManagerField = sslContextImplClass.getDeclaredField("trustManager");
+            trustManagerField.setAccessible(true);
+            Object trustManagerObj = trustManagerField.get(sslContextImpl);
+
+            tm = new TrustManager[] { (X509TrustManager) trustManagerObj };
+        } catch (IllegalAccessException | NoSuchFieldException | ConfigurationException e) {
+            logger.warn("using default insecure trustmanager, could not extract trustmanager from SSL context:", e);
+        }
+        tm = InsecureTrustManagerFactory.INSTANCE.getTrustManagers();
```
cweitkamp: Will this override the previous value assigned in line 63?

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/ssl/CustomTrustManagerFactory.java
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * The {@link CustomTrustManagerFactory} is a TrustManagerFactory that provides a custom {@link TrustManager}
+ *
+ * @author Jan N. Klug - Initial contribution
+ */
+@NonNullByDefault
+public class CustomTrustManagerFactory extends SimpleTrustManagerFactory {
+    private final Logger logger = LoggerFactory.getLogger(CustomTrustManagerFactory.class);
+    private TrustManager[] tm;
```
cweitkamp: Can this be final?

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/client/MqttAsyncClientWrapper.java
```diff
@@ -0,0 +1,58 @@
...
+    protected MqttQos getMqttQosFromInt(int qos) {
+        switch (qos) {
+            case 0:
+                return MqttQos.AT_LEAST_ONCE;
+            case 1:
+                return MqttQos.AT_MOST_ONCE;
+            case 2:
+                return MqttQos.EXACTLY_ONCE;
+            default:
+                throw new IllegalArgumentException();
```
cweitkamp: Does it make sense to use a specific message for this exception?

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/client/Mqtt5AsyncClientWrapper.java
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * The {@link Mqtt5AsyncClientWrapper} provides the wrapper for Mqttv3 async clients
+ *
+ * @author Jan N. Klug - Initial contribution
+ */
+
+@NonNullByDefault
+public class Mqtt5AsyncClientWrapper extends MqttAsyncClientWrapper {
+    private Mqtt5AsyncClient client;
```
cweitkamp: Final?

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/client/Mqtt5AsyncClientWrapper.java
```diff
@@ -0,0 +1,110 @@
...
+import com.hivemq.client.mqtt.mqtt5.message.publish.Mqtt5PublishResult;
+import com.hivemq.client.mqtt.mqtt5.message.subscribe.Mqtt5Subscribe;
+import com.hivemq.client.mqtt.mqtt5.message.unsubscribe.Mqtt5Unsubscribe;
+
+/**
+ * The {@link Mqtt5AsyncClientWrapper} provides the wrapper for Mqttv3 async clients
+ *
+ * @author Jan N. Klug - Initial contribution
+ */
+
```
cweitkamp: Please remove this empty line.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/client/Mqtt3AsyncClientWrapper.java
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * The {@link Mqtt3AsyncClientWrapper} provides the wrapper for Mqttv3 async clients
+ *
+ * @author Jan N. Klug - Initial contribution
+ */
+
+@NonNullByDefault
+public class Mqtt3AsyncClientWrapper extends MqttAsyncClientWrapper {
+    private Mqtt3AsyncClient client;
```
cweitkamp: Final?

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/client/Mqtt3AsyncClientWrapper.java
```diff
@@ -0,0 +1,109 @@
...
+import com.hivemq.client.mqtt.mqtt3.message.publish.Mqtt3Publish;
+import com.hivemq.client.mqtt.mqtt3.message.subscribe.Mqtt3Subscribe;
+import com.hivemq.client.mqtt.mqtt3.message.unsubscribe.Mqtt3Unsubscribe;
+
+/**
+ * The {@link Mqtt3AsyncClientWrapper} provides the wrapper for Mqttv3 async clients
+ *
+ * @author Jan N. Klug - Initial contribution
+ */
+
```
cweitkamp: Please remove this empty line.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
J-N-K: I changed the wording and upgraded to warn. error is too much IMO, this does not affect the funtion of the other topics or the connection.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/ssl/CustomTrustManagerFactory.java
J-N-K: Yes. Closing brace on wrong line.

J-N-K:
Please note that https://github.com/openhab/openhab2-addons/pull/6245 needs to be merged to keep the tests in addons running.

Issue #1113, opened 2019-10-10
SafeCallerImplTest unstable


wborn:
This test failed on [Jenkins](https://ci.openhab.org/job/PR-openHAB-Core/1719/console) (https://github.com/openhab/openhab-core/pull/1110#issuecomment-540485847) with the following stack trace:

```
00:15:44 TEST testAsyncExceptionHandler(org.eclipse.smarthome.core.internal.common.SafeCallerImplTest) <<< ERROR: 
00:15:44 No interactions wanted here:
00:15:44 -> at org.eclipse.smarthome.core.internal.common.SafeCallerImplTest.testAsyncExceptionHandler(SafeCallerImplTest.java:407)
00:15:44 But found this interaction on mock 'mockTimeoutHandler':
00:15:44 -> at org.eclipse.smarthome.core.internal.common.AbstractInvocationHandler.handleTimeout(AbstractInvocationHandler.java:121)
00:15:44 Actually, above is the only interaction with this mock.
00:15:44 org.mockito.exceptions.verification.NoInteractionsWanted: 
00:15:44 No interactions wanted here:
00:15:44 -> at org.eclipse.smarthome.core.internal.common.SafeCallerImplTest.testAsyncExceptionHandler(SafeCallerImplTest.java:407)
00:15:44 But found this interaction on mock 'mockTimeoutHandler':
00:15:44 -> at org.eclipse.smarthome.core.internal.common.AbstractInvocationHandler.handleTimeout(AbstractInvocationHandler.java:121)
00:15:44 Actually, above is the only interaction with this mock.
00:15:44 	at org.eclipse.smarthome.core.internal.common.SafeCallerImplTest.testAsyncExceptionHandler(SafeCallerImplTest.java:407)
00:15:44 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
00:15:44 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
00:15:44 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
00:15:44 	at java.lang.reflect.Method.invoke(Method.java:498)
00:15:44 	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
00:15:44 	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
[...]

PR #1126, opened 2019-10-13
add a feature for netty

J-N-K:
This adds a feature `openhab.tp-netty` with Netty 4.1.34.Final andthe most common bundles. 

Can/Should we define a `netty.version` which can be used by core- or addons-bundles too add additional bundles and keep them in sync?

### DIFF
--- a/features/karaf/openhab-tp/src/main/feature/feature.xml
+++ b/features/karaf/openhab-tp/src/main/feature/feature.xml
@@ -121,6 +121,20 @@
     <bundle dependency="true">mvn:org.jvnet.mimepull/mimepull/1.9.6</bundle>
   </feature>
 
+  <feature name="openhab.tp-netty" description="Netty bundles" version="${project.version}">
+    <capability>openhab.tp;feature=netty;version=4.1.34.Final</capability>
+    <bundle dependency="true">mvn:io.netty/netty-buffer/4.1.34.Final</bundle>
+    <bundle dependency="true">mvn:io.netty/netty-common/4.1.34.Final</bundle>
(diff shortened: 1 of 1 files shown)


cweitkamp:
> Can/Should we define a netty.version

From my point of view it makes sense.

J-N-K:
What would be the place? `jetty.version` is set in compile/runtime-bom. That feels wrong, because we don't add bundles for the boms.

FTR: Jenkins fails in some core-test, unrelated.

cweitkamp:
The right place will be the openhab-tp [POM](https://github.com/openhab/openhab-core/blob/91617d8af81dfa36cfd9514a379c7893b1de5ebd/features/karaf/openhab-tp/pom.xml). Isn't it?

Yes, we are aware of the issue with core-test (see #1113).

J-N-K:
But will this variable be available e.g. in openhab2-addons?

cweitkamp:
Hm, good question...

kaikreuzer:
Afaik, it won't be available in openhab2-addons, so the version property would have to be added to multiple poms (which is usually a pita, so better refrain from it).

Not sure which additional netty bundles could be needed and how diverse those requirements are.
Imho the best way is either to add them to the one tp feature or add some additional netty tp features that provide further capabilities.

J-N-K:
@kaikreuzer If we would like to resolve projects in the demo-app we need the dependencies. We can either add them to compile/runtime-bom or leave it to the developer to add them to the individual project-pom (like we do now, e.g. in homekit). This requires the version and thus needs to be kept in sync with the feature if we update that and want to ensure that we have the same versions at compile and runtime.

PR #1133, opened 2019-10-14
Add null annotations to semantics

wborn:
This PR adds null annotations to the semantics bundle. I also updated the Groovy script used for generating these classes based on the CSV file.

Apparently there is also a `Sensor` equipment interface which isn't defined in the CSV file. Therefor it's currently not added to the generated `Equipments` class or `tags.properties` file.

Do you know if should be there or deleted @kaikreuzer, @ghys? I see there was some discussion about using Measurement vs. Sensor in the [original PR](https://github.com/eclipse/smarthome/pull/6288#issuecomment-428056224).


ghys:
Ah! Good catch on Sensor missing from the CSV... I wondered what was up with it the other day (and didn't investigate 😅)

wborn:
That depends on the opinions of @ghys and @kaikreuzer.

Removal is functionally equivalent with the current behavior.  I think users following the [HABot Walkthrough](https://community.openhab.org/t/habot-walkthrough-2-n-semantic-tagging-item-resolving/55651) might think it's a bug when using the diagram as reference:

424749375f6d51214475ac0d2b9000a957d718a7


If the `Sensor`  is added we might also want to use it as parent for the `MotionDetector` and `SmokeDetector` .

kaikreuzer:
I don't have any strong opinion. While "sensor" is pretty generic and thus not really helpful, we might have to deal with devices that are pure sensors and which don't fall into any better category.

> If the Sensor is added we might also want to use it as parent for the MotionDetector and SmokeDetector .

Sounds sensible to me.

PR #1134, opened 2019-10-16
Add null annotations to registries

wborn:
Adds null annotations to all registries, the interfaces they implement and a few other classes.

---

There's also a PR to update the add-ons for this: https://github.com/openhab/openhab2-addons/pull/6233

### DIFF
--- a/bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/ModuleTypeResource.java
+++ b/bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/ModuleTypeResource.java
@@ -28,6 +28,8 @@ import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
 import javax.ws.rs.core.UriInfo;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.smarthome.io.rest.LocaleService;
 import org.eclipse.smarthome.io.rest.RESTResource;
 import org.openhab.core.automation.dto.ActionTypeDTOMapper;
@@ -62,14 +64,15 @@ import io.swagger.annotations.ApiResponses;
  */
 @Path("module-types")
 @Api("module-types")
+@NonNullByDefault
 @Component
 public class ModuleTypeResource implements RESTResource {
 
-    private ModuleTypeRegistry moduleTypeRegistry;
-    private LocaleService localeService;
+    private @NonNullByDefault({}) ModuleTypeRegistry moduleTypeRegistry;
+    private @NonNullByDefault({}) LocaleService localeService;
 
     @Context
-    private UriInfo uriInfo;
+    private @NonNullByDefault({}) UriInfo uriInfo;
 
     @Reference(cardinality = ReferenceCardinality.OPTIONAL, policy = ReferencePolicy.DYNAMIC)
     protected void setModuleTypeRegistry(ModuleTypeRegistry moduleTypeRegistry) {
@@ -94,11 +97,11 @@ public class ModuleTypeResource implements RESTResource {
     @ApiOperation(value = "Get all available module types.", response = ModuleTypeDTO.class, responseContainer = "List")
     @ApiResponses(value = {
             @ApiResponse(code = 200, message = "OK", response = ModuleTypeDTO.class, responseContainer = "List") })
-    public Response getAll(@HeaderParam("Accept-Language") @ApiParam(value = "language") String language,
-            @QueryParam("tags") @ApiParam(value = "tags for filtering", required = false) String tagList,
-            @QueryParam("type") @ApiParam(value = "filtering by action, condition or trigger", required = false) String type) {
+    public Response getAll(@HeaderParam("Accept-Language") @ApiParam(value = "language") @Nullable String language,
+            @QueryParam("tags") @ApiParam(value = "tags for filtering", required = false) @Nullable String tagList,
+            @QueryParam("type") @ApiParam(value = "filtering by action, condition or trigger", required = false) @Nullable String type) {
         final Locale locale = localeService.getLocale(language);
-        final String[] tags = tagList != null ? tagList.split(",") : null;
+        final String[] tags = tagList != null ? tagList.split(",") : new String[0];
         final List<ModuleTypeDTO> modules = new ArrayList<>();
 
         if (type == null || type.equals("trigger")) {
@@ -119,7 +122,7 @@ public class ModuleTypeResource implements RESTResource {
     @ApiOperation(value = "Gets a module type corresponding to the given UID.", response = ModuleTypeDTO.class)
     @ApiResponses(value = { @ApiResponse(code = 200, message = "OK", response = ModuleTypeDTO.class),
             @ApiResponse(code = 404, message = "Module Type corresponding to the given UID does not found.") })
-    public Response getByUID(@HeaderParam("Accept-Language") @ApiParam(value = "language") String language,
+    public Response getByUID(@HeaderParam("Accept-Language") @ApiParam(value = "language") @Nullable String language,
             @PathParam("moduleTypeUID") @ApiParam(value = "moduleTypeUID", required = true) String moduleTypeUID) {
         Locale locale = localeService.getLocale(language);
         final ModuleType moduleType = moduleTypeRegistry.get(moduleTypeUID, locale);
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/provider/AbstractResourceBundleProvider.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/provider/AbstractResourceBundleProvider.java
@@ -21,6 +21,7 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Enumeration;
 import java.util.LinkedList;
 import java.util.List;
@@ -30,6 +31,8 @@ import java.util.Map.Entry;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.smarthome.config.core.ConfigDescriptionParameter;
 import org.eclipse.smarthome.config.core.ConfigDescriptionParameterBuilder;
 import org.eclipse.smarthome.config.core.ParameterOption;
@@ -67,14 +70,11 @@ import org.slf4j.LoggerFactory;
  * @author Kai Kreuzer - refactored (managed) provider and registry implementation
  */
 @SuppressWarnings("rawtypes")
+@NonNullByDefault
 public abstract class AbstractResourceBundleProvider<E> {
 
-    public AbstractResourceBundleProvider() {
-        providedObjectsHolder = new ConcurrentHashMap<>();
-        providerPortfolio = new ConcurrentHashMap<>();
-        queue = new AutomationResourceBundlesEventQueue<>(this);
-        parsers = new ConcurrentHashMap<>();
-        waitingProviders = new ConcurrentHashMap<>();
+    public AbstractResourceBundleProvider(String path) {
+        this.path = path;
     }
 
     /**
@@ -86,7 +86,7 @@ public abstract class AbstractResourceBundleProvider<E> {
     /**
      * This field holds a reference to the service instance for internationalization support within the platform.
      */
-    protected TranslationProvider i18nProvider;
+    protected @NonNullByDefault({}) TranslationProvider i18nProvider;
 
     /**
      * This field keeps instance of {@link Logger} that is used for logging.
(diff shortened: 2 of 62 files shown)


bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/module/provider/i18n/ModuleTypeI18nService.java
```diff
@@ -32,6 +35,8 @@
      * @param bundle the bundle containing the localization files
      * @return the localized ModuleType
      */
-    ModuleType getModuleTypePerLocale(ModuleType defModuleType, Locale locale, Bundle bundle);
+    @Nullable
+    ModuleType getModuleTypePerLocale(@Nullable ModuleType defModuleType, @Nullable Locale locale,
```
cweitkamp: Are you sure the `Bundle` should be annotated with `@Nullable`? Because if we do not have a bundle we will not get a translation. Thus I think is does not make sense to call such a method in that case. We e.g. did not annotated it with `@Nullable` in the `ConfigI18nLocalizationService` or `ProfileI18nUtil`.

wborn: Looks like I used `@Nullable` because the `bundle` in `org.openhab.core.automation.internal.provider.ModuleTypeResourceBundleProvider.getPerLocale` can be `null` :

```java
        String uid = defModuleType.getUID();
        Bundle bundle = getBundle(uid);
```

We can also let `getPerLocale` returns `null` when it can't find the bundle.

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/provider/i18n/ModuleI18nUtil.java
```diff
@@ -31,22 +33,28 @@
...
         List<T> lmodules = new ArrayList<>();
         for (T module : modules) {
             String label = getModuleLabel(i18nProvider, bundle, uid, module.getId(), module.getLabel(), prefix, locale);
             String description = getModuleDescription(i18nProvider, bundle, uid, prefix, module.getId(),
                     module.getDescription(), locale);
-            lmodules.add(createLocalizedModule(module, label, description));
+            @Nullable
+            T lmodule = createLocalizedModule(module, label, description);
+            if (lmodule != null) {
+                lmodules.add(lmodule);
```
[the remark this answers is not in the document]
wborn: That might make sense. I sometimes had a hard time to understand what some code is about without any docs and all the generic terminology. :smile:

bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/ModuleTypeResource.java
```diff
@@ -62,14 +64,15 @@
  */
 @Path("module-types")
 @Api("module-types")
+@NonNullByDefault
 @Component
 public class ModuleTypeResource implements RESTResource {
 
-    private ModuleTypeRegistry moduleTypeRegistry;
-    private LocaleService localeService;
+    private @NonNullByDefault({}) ModuleTypeRegistry moduleTypeRegistry;
```
cweitkamp: One general thing: I would prefer to use constructor injection instead of this annotation. But I would be fine to do it in a separate PR after this one has been merged.

wborn: Yes this PR has become too big already so I'd like to wrap it up before I spent the rest of my time on resolving merge conflicts. :wink: 

I hope with having annotated all registries, any subsequent PR that adds null annotations will be easier.

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/module/provider/i18n/ModuleTypeI18nService.java
cweitkamp: I tried to look into this. At the first glance I second your suggestion to return `null` if the `bundle` is `null`.

At the second view I would tend to throw away the `getBundle()` method it somehow looks strange to me. In other bundles we use the `FrameworkUtil.getBundle()` method in such cases. But to be honest I do not now if it will work here.

wborn: Until I fully understand what is going here I've gone for the conservative approach by adding a null check and not throwing the method away. I think the idea is that these ModuleTypes can be provided by different bundles.

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/provider/i18n/ModuleI18nUtil.java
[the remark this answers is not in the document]
wborn: I think this was fixed by you. :+1:

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/provider/AbstractResourceBundleProvider.java
```diff
@@ -395,7 +387,8 @@ protected Bundle getBundle(String uid) {
     }
 
     protected List<ConfigDescriptionParameter> getLocalizedConfigurationDescription(TranslationProvider i18nProvider,
-            List<ConfigDescriptionParameter> config, Bundle bundle, String uid, String prefix, Locale locale) {
+            @Nullable List<ConfigDescriptionParameter> config, @Nullable Bundle bundle, String uid, String prefix,
+            Locale locale) {
```
cweitkamp:
```suggestion
            @Nullable Locale locale) {
```

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/type/ModuleTypeRegistryImpl.java
```diff
@@ -73,22 +76,23 @@ public ModuleType get(String typeUID) {
...
         Collection<T> result = new ArrayList<>(20);
         forEach((provider, mType) -> {
             ModuleType mt = locale == null ? mType
                     : ((ModuleTypeProvider) provider).getModuleType(mType.getUID(), locale);
-            Collection<String> tags = mt.getTags();
-            if (moduleTypeTag == null) {
-                result.add((T) createCopy(mt));
-            } else if (tags.contains(moduleTypeTag)) {
-                result.add((T) createCopy(mt));
+            if ((mt != null) && (moduleTypeTag == null || mt.getTags().contains(moduleTypeTag))) {
```
cweitkamp:
```suggestion
            if (mt != null && (moduleTypeTag == null || mt.getTags().contains(moduleTypeTag))) {
```

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/type/ModuleTypeRegistryImpl.java
```diff
@@ -101,23 +105,25 @@ public ModuleType get(String typeUID) {
...
+        Set<String> tagSet = new HashSet<>(Arrays.asList(tags));
         Collection<T> result = new ArrayList<>(20);
         forEach((provider, mType) -> {
             ModuleType mt = locale == null ? mType
                     : ((ModuleTypeProvider) provider).getModuleType(mType.getUID(), locale);
-            if (tagSet == null) {
-                result.add((T) createCopy(mt));
-            } else if (mt.getTags().containsAll(tagSet)) {
-                result.add((T) createCopy(mt));
+            if ((mt != null) && (mt.getTags().containsAll(tagSet))) {
```
cweitkamp:
```suggestion
            if (mt != null && (mt.getTags().containsAll(tagSet))) {
```

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/template/TemplateRegistry.java
```diff
@@ -47,25 +50,25 @@
...
+     *            {@code null} then the result will contain all available templates.
      * @param locale specifies the desired {@link Locale} to be used for localization of the returned elements. If
-     *               localization resources for this locale are not available or the passed locale is {@code null} the
-     *               elements are returned with the default localization.
+     *            localization resources for this locale are not available or the passed locale is {@code null} the
+     *            elements are returned with the default localization.
      * @return a collection of localized templates, which correspond to the specified tag.
      */
-    public Collection<E> getByTag(String tag, Locale locale);
+    public Collection<E> getByTag(@Nullable String tag, Locale locale);
```
cweitkamp:
```suggestion
    public Collection<E> getByTag(@Nullable String tag, @Nullable Locale locale);
```

bundles/org.openhab.core.ui/src/main/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImpl.java
```diff
@@ -800,7 +804,7 @@ private int getNbOptions(String itemName) {
         return 0;
     }
 
-    private Class<? extends Item> getItemType(@NonNull String itemName) {
+    private @Nullable Class<? extends Item> getItemType(@NonNull String itemName) {
```
cweitkamp:
```suggestion
    private @Nullable Class<? extends Item> getItemType(String itemName) {
```

bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/i18n/ConfigDescriptionI18nUtil.java
```diff
@@ -40,29 +43,30 @@ public ConfigDescriptionI18nUtil(TranslationProvider i18nProvider) {
         this.i18nProvider = i18nProvider;
     }
 
-    public String getParameterPattern(Bundle bundle, URI configDescriptionURI, String parameterName,
-            String defaultPattern, Locale locale) {
+    public @Nullable String getParameterPattern(@Nullable Bundle bundle, URI configDescriptionURI, String parameterName,
```
cweitkamp: Same comment like in the `ModuleTypeI18nUtil` regarding the bundle:

```suggestion
    public @Nullable String getParameterPattern(Bundle bundle, URI configDescriptionURI, String parameterName,
```

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/provider/AbstractResourceBundleProvider.java
wborn: Many methods got annotated this way but I seemed to have missed that one. :+1:

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/type/ModuleTypeRegistryImpl.java
wborn: On other projects my Eclipse is configured to automatically remove redundant parentheses as save cleanup action. :wink:

bundles/org.openhab.core.config.core/src/main/java/org/eclipse/smarthome/config/core/i18n/ConfigDescriptionI18nUtil.java
wborn: I'll any resolve any null analysis errors this might cause!

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/provider/i18n/ModuleTypeI18nServiceImpl.java
```diff
@@ -76,35 +86,35 @@ public ModuleType getModuleTypePerLocale(ModuleType defModuleType, Locale locale
...
+            @Nullable Locale locale) {
         try {
-            uri = new URI(prefix + ":" + uid + ".name");
+            return localizationService
+                    .getLocalizedConfigDescription(bundle,
+                            new ConfigDescription(new URI(prefix + ":" + uid + ".name"), parameters), locale)
+                    .getParameters();
         } catch (URISyntaxException e) {
             logger.error("Constructed invalid uri '{}:{}.name'", prefix, uid, e);
+            return null;
```
cweitkamp: We should return the default values here. I fixed it in https://github.com/openhab/openhab-core/pull/1161/files#diff-d90b601b2b68bbe8017ba644c98c6da1R387.

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/provider/AbstractResourceBundleProvider.java
```diff
@@ -406,6 +399,7 @@ protected Bundle getBundle(String uid) {
                     uri = new URI(prefix + ":" + uid + ".name");
                 } catch (URISyntaxException e) {
                     logger.error("Constructed invalid uri '{}:{}.name'", prefix, uid, e);
+                    return Collections.emptyList();
```
cweitkamp: We should return the default values here. I fixed it in https://github.com/openhab/openhab-core/pull/1161/files#diff-d90b601b2b68bbe8017ba644c98c6da1R387.

PR #1150, opened 2019-10-19
Fix ItemResourceOSGiTest on Windows

### LINKED ISSUE #1002: ItemResourceOSGiTest fails on Windows
```
09:18:19.524 [main] DEBUG org.eclipse.smarthome.core.common.registry.AbstractRegistry - Provider "ItemProvider$MockitoMock$2094127727" has been removed.
TEST shouldReturnUnicodeItems(org.eclipse.smarthome.io.rest.core.internal.item.ItemResourceOSGiTest) <<< ERROR:
Expected: (a collection containing "Test ▒▒▒▒")
     but: a collection containing "Test ▒▒▒▒" was "Test äöü▒?"
java.lang.AssertionError:
Expected: (a collection containing "Test ▒▒▒▒")
     but: a collection containing "Test ▒▒▒▒" was "Test äöü▒?"
        at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
        at org.junit.Assert.assertThat(Assert.java:956)
        at org.junit.Assert.assertThat(Assert.java:923)
        at org.eclipse.smarthome.io.rest.core.internal.item.ItemResourceOSGiTest.shouldReturnUnicodeItems(ItemResourceOSGiTest.java:103)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
        at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
        at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
        at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
        at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
        at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
        at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
        at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
        at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
        at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
        at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
        at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
        at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
        at junit.framework.JUnit4TestAdapter.run(JUnit4TestAdapter.java:38)
        at junit.framework.TestSuite.runTest(TestSuite.java:252)
        at junit.framework.TestSuite.run(TestSuite.java:247)
        at aQute.junit.Activator.test(Activator.java:340)
        at aQute.junit.Activator.automatic(Activator.java:236)
        at aQute.junit.Activator.run(Activator.java:177)
        at aQute.launcher.Launcher.lambda$serviceChanged$0(Launcher.java:1385)
        at aQute.launcher.Launcher.run(Launcher.java:352)
        at aQute.launcher.Launcher.main(Launcher.java:152)
```

wborn:
Fixes #1002

---

With this PR and #1148, a full build with tests on Windows builds successfully for me @lolodomo.

### DIFF
--- a/bundles/org.openhab.core.io.rest/src/main/java/org/eclipse/smarthome/io/rest/Stream2JSONInputStream.java
+++ b/bundles/org.openhab.core.io.rest/src/main/java/org/eclipse/smarthome/io/rest/Stream2JSONInputStream.java
@@ -14,7 +14,6 @@ package org.eclipse.smarthome.io.rest;
 
 import java.io.IOException;
 import java.io.InputStream;
-import java.nio.charset.StandardCharsets;
 import java.util.Iterator;
 import java.util.stream.Stream;
 
@@ -96,11 +95,7 @@ public class Stream2JSONInputStream extends InputStream {
         }
 
         IOUtils.closeQuietly(jsonElementStream);
-        try {
-            jsonElementStream = IOUtils.toInputStream(prefix + entity + postfix, StandardCharsets.UTF_8.name());
-        } catch (IOException e) {
-            // IOException is thrown for invalid encoding. This will never happen with StandardCharsets.UTF_8.
-        }
+        jsonElementStream = IOUtils.toInputStream(prefix + entity + postfix);
     }
 
     private boolean finished() {


bundles/org.openhab.core.io.rest/src/main/java/org/eclipse/smarthome/io/rest/Stream2JSONInputStream.java
```diff
@@ -96,11 +95,7 @@ private void fillBuffer() {
         }
 
         IOUtils.closeQuietly(jsonElementStream);
-        try {
-            jsonElementStream = IOUtils.toInputStream(prefix + entity + postfix, StandardCharsets.UTF_8.name());
```
kaikreuzer: Hm, I would have thought that there was a good reason to pin UTF-8 encoding here. After all, we are talking about the REST API, which should use a strictly defined encoding.
Are you sure that it isn't rather the test that might need to be modified?

lolodomo:
I have no opinion if this fix is what we need, but yes with this fix, the test is then ok.

wborn:
I will do some testing with a real openHAB instance with/without this change running on
 both Linux/Windows and a UTF-8 encoded items file and see if these characters properly show on both Linux/Windows.

By coincidence I [just read](https://community.openhab.org/t/feedback-from-a-new-user/83893/3?u=wborn) in OH 2.4 there were issues with this so it would be nice to know if this is currently resolved and if this change would thus be either a fix or causing a regression. :wink:

wborn:
After some testing it still seems to work properly with this change when running openHAB on both Linux/Windows and consuming the various `/rest/items` resources via the REST API using both Linux/Windows. So it seems the UTF-8 encoding is still taken care of at another level.

I also tested 2.4.0 and saw that a `GET` on `/rest/items/Item_Name` was returning ????? when running openHAB on Windows and using Greek/Russian symbols with the REST API  on both Linux/Windows. It also showed these question marks with `items:list`:

### openHAB 2.4.0 (Windows)

```
openhab> items list
German_Item (Type=StringItem, State=German state äßöüÄÖÜ, Label=German label äßöüÄÖÜ, Category=null)
Dutch_Item (Type=StringItem, State=Dutch state abcDEF, Label=Dutch label abcDEF, Category=null)
Greek_Item (Type=StringItem, State=Greek state ?????, Label=Greek label ?????, Category=null)
French_Item (Type=StringItem, State=French state àêîÎç€, Label=French label àêîÎç€, Category=null)
Russian_Item (Type=StringItem, State=Russian state ??????, Label=Russian label ??????, Category=null)
```

### openHAB 2.5.0.M4 (Linux/Windows)

```
openhab> items list
German_Item (Type=StringItem, State=NULL, Label=German label äßöüÄÖÜ, Category=null)
Dutch_Item (Type=StringItem, State=NULL, Label=Dutch label abcDEF, Category=null)
Greek_Item (Type=StringItem, State=NULL, Label=Greek label αβγΔΘ, Category=null)
French_Item (Type=StringItem, State=NULL, Label=French label àêîÎç€, Category=null)
Russian_Item (Type=StringItem, State=NULL, Label=Russian label АВГДЄЧ, Category=null)
```

kaikreuzer:
> So it seems the UTF-8 encoding is still taken care of at another level.

Hm, I would really prefer to understand why it might still work.
The method without the encoding parameter states in the JavaDoc "using the default character encoding of the platform". So clearly this should make a difference, if the platform is not using UTF-8. What is set on your Windows machine as encoding, @wborn? And which encoding uses the command line window?

@htreu Do you as the original author happen to remember any details on this matter?

wborn:
This is what PowerShell reports on my Windows 10 VM:

```
Windows PowerShell
Copyright (C) Microsoft Corporation. All rights reserved.

PS C:\Users\Wouter> [System.Text.Encoding]::Default


IsSingleByte      : True
BodyName          : iso-8859-1
EncodingName      : Western European (Windows)
HeaderName        : Windows-1252
WebName           : Windows-1252
WindowsCodePage   : 1252
IsBrowserDisplay  : True
IsBrowserSave     : True
IsMailNewsDisplay : True
IsMailNewsSave    : True
EncoderFallback   : System.Text.InternalEncoderBestFitFallback
DecoderFallback   : System.Text.InternalDecoderBestFitFallback
IsReadOnly        : True
CodePage          : 1252
```

wborn:
After following lots method calls I may have found the other level in the [org.eclipse.smarthome.io.rest.core.internal.GsonProvider](https://github.com/openhab/openhab-core/blob/be69d2254e74ee7a12b72559963990d046af2855/bundles/org.openhab.core.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/GsonProvider.java) which seems to take care of all JSON serialization/deserialization and always uses UTF-8 for this:

https://github.com/openhab/openhab-core/blob/be69d2254e74ee7a12b72559963990d046af2855/bundles/org.openhab.core.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/GsonProvider.java#L66-L74

https://github.com/openhab/openhab-core/blob/be69d2254e74ee7a12b72559963990d046af2855/bundles/org.openhab.core.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/GsonProvider.java#L81-L88

wborn:
It might also be better if the `Stream2JSONInputStream` doesn't convert between encodings. There is a  `charset` directive in the `Content-Type` header (see [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type)) which we may want to support one day. The conversion in the `Stream2JSONInputStream` to UTF-8 would then result in unnecessary processing on some platforms. 

The name of the `Stream2JSONInputStream` also suggests the entity wil always be JSON which may also be problematic if we add an additional provider for supporting `text/xml`.

kaikreuzer:
Thanks for your elaborate analysis, just for convincing me. But I can clearly say that you've succeeded 😉.

 > The name of the Stream2JSONInputStream also suggests the entity wil always be JSON

Right, I think this also cause part of my concern: I somehow was under the impression, that it directly produces JSON from the input stream, but you are right that the deserialisation is done elsewhere.

wborn:
Great! I also like knowing the true reason why it all keeps working and it was also nice to see it all working with some real openHAB instances cross OS-es. :-)

PR #1155, opened 2019-10-20
Use !isEmpty() instead of "size() > 0" or "size() != 0"

wborn:
`!isEmpty()` expresses the intent more clearly and is therefore preferred.
Counting the number of elements can also be an expensive operation e.g. when using linked lists.

Relates to: #1151

### DIFF
--- a/bundles/org.openhab.core.id/src/main/java/org/eclipse/smarthome/core/id/InstanceUUID.java
+++ b/bundles/org.openhab.core.id/src/main/java/org/eclipse/smarthome/core/id/InstanceUUID.java
@@ -88,7 +88,7 @@ public class InstanceUUID {
         } catch (IOException ioe) {
             LOGGER.warn("Failed reading the UUID file '{}': {}", file.getAbsolutePath(), ioe.getMessage());
         }
-        return lines != null && lines.size() > 0 ? lines.get(0) : "";
+        return lines != null && !lines.isEmpty() ? lines.get(0) : "";
     }
 
 }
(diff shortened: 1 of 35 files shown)


bundles/org.openhab.core.voice/src/main/java/org/eclipse/smarthome/core/voice/text/ExpressionCardinality.java
```diff
@@ -62,7 +62,7 @@ ASTNode parse(ResourceBundle language, TokenList tokenList) {
             node.setChildren(nodes.toArray(new ASTNode[0]));
             node.setRemainingTokens(list);
             node.setSuccess(true);
-            node.setValue(atMostOne ? (values.size() > 0 ? values.get(0) : null) : values.toArray());
+            node.setValue(atMostOne ? (!values.isEmpty() ? values.get(0) : null) : values.toArray());
```
cweitkamp: This tenary operator can be inverted:
```suggestion
            node.setValue(atMostOne ? (values.isEmpty() ? null : values.get(0)) : values.toArray());
```

wborn: I also prefer getting rid of the exclamation marks where possible because they are easily overlooked (https://github.com/openhab/openhab-core/issues/1022#issuecomment-538877251) :exclamation:

bundles/org.openhab.core.id/src/main/java/org/eclipse/smarthome/core/id/InstanceUUID.java
```diff
@@ -88,7 +88,7 @@ private static String readFirstLine(File file) {
         } catch (IOException ioe) {
             LOGGER.warn("Failed reading the UUID file '{}': {}", file.getAbsolutePath(), ioe.getMessage());
         }
-        return lines != null && lines.size() > 0 ? lines.get(0) : "";
+        return lines != null && !lines.isEmpty() ? lines.get(0) : "";
```
cweitkamp: Okay. Then you might want to check this one too. :wink:

Issue #1157, opened 2019-10-21
DateTime Items: Parse unix time


mueller-ma:
If a DateTime Item receives an updated, it parses a String and converts it into a date and time.
Can you extend this parsing to accept a unix time stamp as well?

The openHAB app for Android sends the next alarm clock time to the server as unix time. This fails if the receiver item is DateTime. While the app can be changed to send a DateTime string, it wouldn't require any migration of the client if the DateTime item can accept unix time.

For further discussion see: https://community.openhab.org/t/android-app-sends-alarm-time-as-long-not-datetime/83758/10

wborn:
It makes sense to add it. But it might not be the best solution because users of the Android App may still be using an older openHAB instance (< 2.5)  which would than not support this functionality.

mueller-ma:
> But it might not be the best solution because users of the Android App may still be using an older openHAB instance (< 2.5) which would then not support this functionality.

They can still use a `Number` item and cast it to `DateTime`, like in the docs: https://www.openhab.org/docs/apps/android.html#alarm-clock

CrazyIvan359:
I probably should have read those docs, but still from a user perspective I don't think it should be necessary to have a rule to handle integration between the mobile app and primary application.

clinique:
@mueller-ma : I think, from a general standpoint that automatic casting of epoch to DateTimeType is valid, why I did this PR. On the other hand, I think that Android App should evolve to directly send properly formatted DateTimeType, thus avoiding this rule needed to handle integration as said by @CrazyIvan359

mueller-ma:
> I probably should have read those docs, but still from a user perspective I don't think it should be necessary to have a rule to handle integration between the mobile app and primary application.

How should the app know what should happen when the alarm time is reached?
I don't like the idea to configure this in the app. Imagine you set your alarm to 7:00 and your phone looses connection between 6:59 and 7:20 => You'll get woken up 20 minutes too late.
With the current implementation (sending the alarm time and having a rule executing this) and one alarm clock per day, the client has 24h time to send the new alarm time.

> On the other hand, I think that Android App should evolve to directly send properly formatted DateTimeType, thus avoiding this rule needed to handle integration as said by @CrazyIvan359

The app can be updated to send a DateTime string, but a rule is still required to set up a timer and execute the alarm, e.g. turn lights on.

I cannot see how a rule can be avoided.

mueller-ma:
[...]

PR #1158, opened 2019-10-22
Issue #1157 Enhance DecimalType to accept epoch seconds and milliseconds.

clinique:


clinique:
Targets Issue #1157

bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/library/items/DateTimeItem.java
```diff
@@ -38,10 +40,14 @@
     private static List<Class<? extends Command>> acceptedCommandTypes = new ArrayList<>();
 
     static {
-        acceptedDataTypes.add((DateTimeType.class));
```
kaikreuzer: I don't think the accepted datatypes should be changed - after all, it expects data time values and nothing else.
Note that `DateTimeType` has a constructor that also takes Strings, so this is probably rather the way you'd go for?

clinique: Yes, I was not sure of my shot on this one. Thanks to point it.

PR #1164, opened 2019-10-24
[automation] Fix typos in EphemerisConditions.json

### LINKED ISSUE #1163: [automation] During start the Automation Provider Processing Queue throws an IAE
I can see this in my IDE and the unit / integration tests too.

```
15:35:09.731 [Automation Provider Processing Queue] ERROR org.openhab.core.automation.internal.provider.AbstractResourceBundleProvider - [Module Type] java.lang.IllegalArgumentException: The type must not be null!

org.openhab.core.automation.parser.ParsingException: [Module Type] java.lang.IllegalArgumentException: The type must not be null!

	at org.openhab.core.automation.internal.parser.gson.ModuleTypeGSONParser.parse(ModuleTypeGSONParser.java:61)
	at org.openhab.core.automation.internal.provider.AbstractResourceBundleProvider.parseData(AbstractResourceBundleProvider.java:410)
	at org.openhab.core.automation.internal.provider.AbstractResourceBundleProvider.processAutomationProvider(AbstractResourceBundleProvider.java:255)
	at org.openhab.core.automation.internal.provider.AutomationResourceBundlesEventQueue.processBundleChanged(AutomationResourceBundlesEventQueue.java:218)
	at org.openhab.core.automation.internal.provider.AutomationResourceBundlesEventQueue.run(AutomationResourceBundlesEventQueue.java:120)
	at java.lang.Thread.run(Thread.java:748)
```

Hilbrand:
Closes #1163

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/resources/ESH-INF/automation/moduletypes/EphemerisConditions.json
+++ b/bundles/org.openhab.core.automation/src/main/resources/ESH-INF/automation/moduletypes/EphemerisConditions.json
@@ -7,11 +7,11 @@
 			"configDescriptions": [
 				{
 					"name": "offset",
-					"type": "integer",
+					"type": "INTEGER",
 					"label": "Offset",
 					"description": "Today +/- offset days (+1 = tomorrow, -1 = yesterday).",
 					"default": 0,
-					"step": 1,
+					"stepsize": 1,
 					"required": false
 				}
 			]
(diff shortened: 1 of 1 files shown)


Hilbrand:
Or rename the field to `step` and use `stepsize` as alternative? as a backward compatible alternative. It seems more logical to keep XML and DTO in sync. Or does that have other consequences.

cweitkamp:
Yes, it will have consequences. The DTO is used by the storage and the REST API too. Internal renaming of the DTO parameter will work. I think for storage it will be also no problem. But for REST ... UIs and other services will use on the current parameter `stepsize`.

cweitkamp:
If we do such a change I would prefer a separate PR for it.

I will merge this now. Test failures are not related.

PR #1165, opened 2019-10-25
[binding] Added nullness annotations

cweitkamp:
- Added nullness annotations
- Use constructor injection
- Moved i18n util class into internal package, introduction of a `BindingI18nLocalizationService`

### DIFF
--- a/bundles/org.openhab.core.binding.xml/src/main/java/org/eclipse/smarthome/core/binding/xml/internal/XmlBindingInfoProvider.java
+++ b/bundles/org.openhab.core.binding.xml/src/main/java/org/eclipse/smarthome/core/binding/xml/internal/XmlBindingInfoProvider.java
@@ -57,7 +57,7 @@ public class XmlBindingInfoProvider extends AbstractXmlBasedProvider<String, Bin
 
     private final BindingI18nLocalizationService bindingI18nService;
     private AbstractXmlConfigDescriptionProvider configDescriptionProvider;
-    private XmlDocumentBundleTracker<BindingInfoXmlResult> bindingInfoTracker;
+    private @Nullable XmlDocumentBundleTracker<BindingInfoXmlResult> bindingInfoTracker;
     private final ReadyService readyService;
     private final ScheduledExecutorService scheduler = ThreadPoolManager
             .getScheduledPool(XmlDocumentBundleTracker.THREAD_POOL_NAME);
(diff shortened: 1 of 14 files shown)


bundles/org.openhab.core.binding.xml/src/main/java/org/eclipse/smarthome/core/binding/xml/internal/XmlBindingInfoProvider.java
```diff
@@ -86,24 +92,15 @@ public void deactivate(ComponentContext componentContext) {
     }
 
     @Override
-    public synchronized BindingInfo getBindingInfo(String id, Locale locale) {
-        return get(id, locale);
+    public synchronized @Nullable BindingInfo getBindingInfo(@Nullable String id, @Nullable Locale locale) {
```
cweitkamp: The JavaDoc in the `BindingInfoProvider` interface says that it can be null or empty but I think that is a mistake.

https://github.com/openhab/openhab-core/blob/d39b6bd35ba231a0633bdf6f5ec6a0922f63db47/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/binding/BindingInfoProvider.java#L33

bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/binding/BindingInfo.java
```diff
@@ -121,14 +124,13 @@ public String getServiceId() {
...
-    public URI getConfigDescriptionURI() {
-        return this.configDescriptionURI;
+    public @Nullable URI getConfigDescriptionURI() {
+        return configDescriptionURI;
     }
 
     @Override
     public String toString() {
-        return "BindingInfoImpl [id=" + id + ", name=" + name + ", description=" + description + ", author=" + author
+        return "BindingInfoI [id=" + id + ", name=" + name + ", description=" + description + ", author=" + author
```
cweitkamp:
```suggestion
        return "BindingInfo [id=" + id + ", name=" + name + ", description=" + description + ", author=" + author
```

wborn:
Do you also have any other plans for adding null annotations @cweitkamp? It saves time if we don't end up adding null annotations to the same files.

I was thinking about having a look myself if all classes that add-ons use have proper null annotations. That way there is no excuse for contributors for not adding them to add-ons. :wink:

cweitkamp:
@wborn Yes, I have. I already started working on `org.openhab.core.config.core`. Not only for improving stability of add-ons but the core itself.

I absolutely second your opinion. It will make sense to coordinate our work to avoid dublicate work.

bundles/org.openhab.core.binding.xml/src/main/java/org/eclipse/smarthome/core/binding/xml/internal/XmlBindingInfoProvider.java
kaikreuzer: I think this is in order to be "forgiving" to callers - if null is passed in, it won't break, but simply also return null.
But I think we can certainly change it here to expect a non-null id to be passed.

Issue #1089, opened 2019-10-03
HostFragmentSupportTest unstable


wborn:
This test failed on Travis with the following stack trace:

```
TEST assertThatTheUpdateOfTheFragmentHostProvidesTheResourcesCorrectly(org.openhab.core.automation.integration.test.HostFragmentSupportTest) <<< ERROR: 
Expected: is <1>
     but: was <32>
java.lang.AssertionError: 
Expected: is <1>
     but: was <32>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.junit.Assert.assertThat(Assert.java:956)
	at org.junit.Assert.assertThat(Assert.java:923)
	at org.openhab.core.automation.integration.test.HostFragmentSupportTest.assertThatTheUpdateOfTheFragmentHostProvidesTheResourcesCorrectly(HostFragmentSupportTest.java:397)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
[...]

PR #1166, opened 2019-10-27
Improve HostFragmentSupportTest stability with null assertions

wborn:
waitForAssert doesn't catch NPEs so this PR fixes that by also checking the localized modules for null.

Related to #1089


kaikreuzer:
@wborn FTR, I removed the "bug" label as we will use those labels for compiling our release notes (which should not cover test related changes).

PR #1168, opened 2019-10-27
[xml] Added constructor injection to config xml

cweitkamp:
- Added constructor injection to config xml


kaikreuzer:
Jenkins does not like it:

```
14:36:22 [ERROR] /home/jenkins/jenkins-agent1/workspace/PR-openHAB-Core/bundles/org.openhab.core.config.xml/src/main/java/org/eclipse/smarthome/config/xml/ConfigXmlConfigDescriptionProvider.java:[96,4208] Illegal redefinition of parameter bundle, inherited method from XmlDocumentProviderFactory<List<ConfigDescription>> does not constrain this parameter
```

cweitkamp:
Thanks for the hint. I think Jenkins failed because the changes are bases on some annotations added with #1165.

PR #1172, opened 2019-10-30
add console command extension for serial inspection

maggu2810:
I found that change in a branch of mine.
I don't know if you are interested in as this is more or less about "diagnostic" information about the internal handling of serial ports.
For me it has been useful (the time I wrote it) to analyse some serial port "problems".

### DIFF
--- a/bundles/org.openhab.core.io.transport.serial/pom.xml
+++ b/bundles/org.openhab.core.io.transport.serial/pom.xml
@@ -12,4 +12,12 @@
 
   <name>openHAB Core :: Bundles :: Serial Transport</name>
 
+  <dependencies>
+    <dependency>
+      <groupId>org.openhab.core.bundles</groupId>
+      <artifactId>org.openhab.core.io.console</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+  </dependencies>
+
 </project>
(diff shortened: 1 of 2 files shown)


kaikreuzer:
Looks useful, thanks!

Only question I would have is about the command names: From what I know, we never use dashes in commands, so this might feel a bit uncommon here.
How about changing it to
```
serial identifiers
serial identifier <name>
serial creators
```
or something similar?

maggu2810:
A question about the help:

As a Java developer I try to follow recommendations like [How to Write Doc Comments for the Javadoc Tool](https://www.oracle.com/technetwork/articles/java/index-137868.html)

This article recommends to document Java methods using the 3rd person:

```
Use 3rd person (descriptive) not 2nd person (prescriptive).
The description is in 3rd person declarative rather than 2nd person imperative.

Gets the label. (preferred)
Get the label. (avoid)
```

While changing the naming of the sub commands I also looked at the help texts I provide by this PR.
They have been inconsistent and now I need to decide which one should be changed.

Currently it looks like:

```java
buildCommandUsage(SUBCMD_IDENTIFIER_ALL, "lists all identifiers"), //
buildCommandUsage(SUBCMD_IDENTIFIER_NAME, "lists a specific identifier"), //
buildCommandUsage(SUBCMD_PORT_CREATORS, "get details about the port creators") //
```

So, I use the 3rd person ("lists") twice and the 2nd person ("get") once.

Which one do you prefer.
I assume we don't need to be strict here, but a common guideline would be nice.

kaikreuzer:
> I assume we don't need to be strict here, but a common guideline would be nice.

I agree and the option you chose (using 3rd person) is definitely fine, so let's stick to it.

PR #1181, opened 2019-11-03
[core] Added nullness annotations and unit test for 'CommandDescriptionBuilder'

cweitkamp:
- Added nullness annotations to `CommandDescriptionBuilder`
- Added unit test for `CommandDescriptionBuilder`

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/types/CommandDescriptionBuilder.java
+++ b/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/types/CommandDescriptionBuilder.java
@@ -15,6 +15,7 @@ package org.eclipse.smarthome.core.types;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.smarthome.core.internal.types.CommandDescriptionImpl;
 
 /**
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/types/CommandDescriptionBuilder.java
```diff
@@ -68,7 +77,7 @@ public CommandDescriptionBuilder withCommandOption(CommandOption commandOption)
      * @return this builder.
      */
     public CommandDescriptionBuilder withCommandOptions(List<CommandOption> commandOptions) {
-        commandOptions.forEach(co -> this.commandOptions.add(co));
```
cweitkamp: Currently all of our builder implementations replace the complete entity list instead of adding them to the existing list:

https://github.com/openhab/openhab-core/blob/f002a21da522b6c42a93b3cf65dce9961bdf10e9/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/types/StateDescriptionFragmentBuilder.java#L143-L146

https://github.com/openhab/openhab-core/blob/f002a21da522b6c42a93b3cf65dce9961bdf10e9/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/builder/ThingBuilder.java#L85-L91

kaikreuzer: Indeed, and that's on purpose - thanks for fixing it here.

PR #1186, opened 2019-11-04
Added links for (available) docs of all Add-Ons

### LINKED ISSUE #269: [Paper UI] Add links to (available) docs for all Add-Ons
On the "Add-ons" page the add-ons on tabs "BINDINGS" and "PERSISTENCE" have links to the corresponding documentation pages on docs.openhab.org.
For the add-ons on the other tabs (e.g. "TRANSFORMATIONS", "USER INTERFACES") the links are missing although the documentation exists.

BTW: the existing links in (https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/KarafExtensionService.java#L109-L115) should better use https instead of http.

cweitkamp:
- Added links for (available) docs of all Add-Ons
- Use constructor injection to simplify lifecycle
- Minor code refactoring

Closes #269 

### DIFF
--- a/bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/FeatureInstaller.java
+++ b/bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/FeatureInstaller.java
@@ -39,6 +39,7 @@ import org.apache.commons.lang.StringUtils;
 import org.apache.karaf.features.Feature;
 import org.apache.karaf.features.FeaturesService;
 import org.eclipse.smarthome.config.core.ConfigConstants;
+import org.eclipse.smarthome.config.core.ConfigurableService;
 import org.eclipse.smarthome.core.events.Event;
 import org.eclipse.smarthome.core.events.EventPublisher;
 import org.eclipse.smarthome.core.extension.ExtensionEventFactory;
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/FeatureInstaller.java
```diff
@@ -56,19 +57,27 @@
...
 
 /**
- * This service reads addons.cfg and installs listed addons (= Karaf features) and the selected package.
+ * This service reads addons.cfg and installs listed add-ons (= Karaf features) and the selected package.
  * It furthermore allows configuration of the base package through the Paper UI as well as administrating Karaf to
  * access remote repos and certain feature repos like for legacy or experimental features.
  *
  * @author Kai Kreuzer - Initial contribution
  */
 @Component(name = "org.openhab.addons", service = { FeatureInstaller.class, ConfigurationListener.class }, property = {
```
cweitkamp: One question: Do we need to define the `configurationPid` for this service?

kaikreuzer: It seems to work without, so probably not.
If I am not mistaken, if there is no dedicated configurationPid, the service name property is used.

cweitkamp: Yes, that is right.

PR #1197, opened 2019-11-10
Upgrade Karaf tooling and openHAB TP to 4.2.7

wborn:
For Karaf 4.2.7 release notes, see:

https://issues.apache.org/jira/secure/ReleaseNote.jspa?projectId=12311140&version=12345539

Also updates the following openHAB TP dependencies:

* Jetty 9.4.20.v20190813
* JNA 5.4.0 (used by most add-ons)

---

Should be merged together with:

* https://github.com/openhab/openhab2-addons/pull/6368
* https://github.com/openhab/openhab-webui/pull/139
* https://github.com/openhab/openhab-distro/pull/996

See also : https://github.com/openhab/openhab-distro/issues/995

### DIFF
--- a/features/karaf/openhab-core/pom.xml
+++ b/features/karaf/openhab-core/pom.xml
@@ -15,8 +15,8 @@
   <description>openHAB Core Features</description>
 
   <properties>
-    <jetty.version>9.4.18.v20190429</jetty.version>
-    <jna.version>4.5.2</jna.version>
+    <jetty.version>9.4.20.v20190813</jetty.version>
+    <jna.version>5.4.0</jna.version>
   </properties>
 
   <build>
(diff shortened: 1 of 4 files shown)


maggu2810:
@wborn Wouldn't it make sense to change the versions of Pax Web and Jetty for the runtime dependencies used in the openHAB demo application, too?

https://github.com/wborn/openhab-core/blob/25bc513c08987959abedcf040fc05623be02137f/bom/runtime/pom.xml#L18-L19

IIRC we tried to use similar bundles in the IDE demo app and the Karaf setup.

wborn:
Yes it would make sense indeed. Can you create a PR for that?

maggu2810:
done

PR #1204, opened 2019-11-11
Feature: system.round profile for rounding numbers before reaching an item

seime:
Similar to the `system.offset` profile, just with number rounding.

Use case: Avoid getting spammed with item updates for nearly identical values. A typical example is anything that measures AC amps/volt as it fluctuates slightly on every read.

Examples: 
`{channel=":volt" [profile="round", decimals="0"]}` will round to nearest int ie 123.32 -> 123
`{channel=":volt" [profile="round", decimals="1"]}` will have 1 decimal place, similar to `DecimalFormat("#.#")` ie 123.32 -> 123.3
`{channel=":volt" [profile="round", decimals="-1"]}` will round to nearest 10 ie 123.32 -> 120
`{channel=":volt" [profile="round", decimals="-2"]}` will round to nearest 100 ie 123.32 -> 100

Any number between 10 and -10 accepted as `decimals` profile parameter.
Handles `DecimalType` and `QuantityType`.

Signed-off-by: Arne Seime <arne.seime@gmail.com>


cweitkamp:
@seime Thanks you very much for your contribution. We already discussed this topic in https://github.com/openhab/openhab-core/pull/779.

PR #1218, opened 2019-11-15
update runtime dependencies to sync Jetty and Pax Web versions

maggu2810:
Use the same Pax Web and Jetty version for the runtime dependencies (used by the demo application) as is used in the Karaf distribution.

This PR needs to be merged synchron to the updated bndrun in the openHAB demo repository.

Related to: https://github.com/openhab/openhab-core/pull/1197#issuecomment-553029014


kaikreuzer:
Thanks!

Looks as if this also requires changes in the itests to succeed again.

itests/org.openhab.core.config.discovery.usbserial.tests/itest.bndrun
```diff
@@ -50,4 +44,11 @@ Fragment-Host: org.openhab.core.config.discovery.usbserial
...
 	net.bytebuddy.byte-buddy-agent;version='[1.9.10,1.9.11)',\
-	org.mockito.mockito-core;version='[3.1.0,3.1.1)'
+	org.mockito.mockito-core;version='[3.1.0,3.1.1)',\
+	org.eclipse.jetty.http;version='[9.4.20,9.4.21)',\
+	org.eclipse.jetty.io;version='[9.4.20,9.4.21)',\
+	org.eclipse.jetty.security;version='[9.4.20,9.4.21)',\
+	org.eclipse.jetty.server;version='[9.4.20,9.4.21)',\
+	org.eclipse.jetty.servlet;version='[9.4.20,9.4.21)',\
+	org.eclipse.jetty.util;version='[9.4.20,9.4.21)',\
+	org.openhab.core.config.discovery.usbserial.linuxsysfs;version='[2.5.0,2.5.1)'
```
wborn: The build fails because this runbundle needs to be manually removed, see the comment above:

```
# We would like to use a mock instead of the "PollingUsbSerialScanner" in tests
# So make sure 'org.openhab.core.config.discovery.usbserial.linuxsysfs' is not in runbundles
```
bnd will fail with another error when it's blacklisted. :-( Maybe you know of a better way?

maggu2810: > Maybe you know of a better way?

Yes, I have one. Will create a PR after this one is included (to avoid merge conflicts).

PR #1221, opened 2019-11-17
use mock instead of real service implementation

maggu2810:
Related to: https://github.com/openhab/openhab-core/pull/1218#pullrequestreview-317995622


maggu2810:
In OSGi it is mostly about capabilities and requirements.
If you annotate an implementation as a component, bnd creates the respective provide capability entry in the manifest.
As we are not using a component we need to add the information manually.

This itest bnd already defines e.g. the symbolic name, the fragment host etc.

PR #1232, opened 2019-11-20
upgrade to commons-collections4 version 4.4

Delanee9:


J-N-K:
Can you explain why this upgrade should be done? IMO the usages should be refactored to get rid of commons-collections in total.

cweitkamp:
I agree. We should get rid of it in OHC. Looking into this PR there seem to be only a few lines of code using `apache-common-collections` which can be changed to standard libraries / methods.

Additionally: Updating the `openhab.tp-base` feature means we are introducing an APIBreaking change here as one of the other bundles (openHAB 2 Add-ons, openHAB Web-UI, Zigbee, Zwave, etc.) outside of this repository might depend on the old version and need to be changed in advance.

kaikreuzer:
I agree with @J-N-K and @cweitkamp, so let's close this PR.

cweitkamp:
My draft for it: https://github.com/openhab/openhab-core/pull/1236

PR #1245, opened 2019-11-28
Added profile for inverting rollershutter / blinds

cweitkamp:
- Added default system wide channel type for rollershutter / blinds 
- Added profile for inverting rollershutter / blinds

For discussion: Currently we are mapping `UpDownType.UP` to `PercentType.ZERO` and `UpDownType.DOWN` to `PercentType.HUNDRED`. In some cases people / bindings / vendors want to have / require the opposite direction and a lot of bindings provide a specific channel configuration for inverting the state / commands for rollershutter / blinds.

See https://community.openhab.org/t/item-profile-to-invert-blinds-orientation/83626
See https://github.com/openhab/openhab2-addons/pull/6389#discussion_r350985672

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemInvertRollershutterProfile.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemInvertRollershutterProfile.java
@@ -1,0 +1,91 @@
+/**
+ * Copyright (c) 2010-2019 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.core.thing.internal.profiles;
+
+import java.math.BigDecimal;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.smarthome.core.library.types.PercentType;
+import org.eclipse.smarthome.core.library.types.UpDownType;
+import org.eclipse.smarthome.core.thing.profiles.ProfileCallback;
+import org.eclipse.smarthome.core.thing.profiles.ProfileTypeUID;
+import org.eclipse.smarthome.core.thing.profiles.StateProfile;
+import org.eclipse.smarthome.core.thing.profiles.SystemProfiles;
+import org.eclipse.smarthome.core.types.Command;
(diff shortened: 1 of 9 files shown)


bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemInvertRollershutterProfile.java
```diff
@@ -0,0 +1,91 @@
...
+        State invertedState = (State) invertType(state);
+        logger.info("Inverting state '{}' on update from handler to '{}'", state, invertedState);
+        callback.sendUpdate(invertedState);
+    }
+
+    private Type invertType(Type type) {
+        if (type instanceof PercentType) {
+            return new PercentType(HUNDRED.subtract(((PercentType) type).toBigDecimal()));
+        } else if (type == UpDownType.UP) {
+            return UpDownType.DOWN;
```
cweitkamp: Maybe convert to `PercentType.HUNDRED` instead?

bundles/org.openhab.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemInvertRollershutterProfile.java
```diff
@@ -0,0 +1,91 @@
...
+        } else if (type == UpDownType.DOWN) {
+            return UpDownType.UP;
```
cweitkamp: Maybe convert to `PercentType.ZERO` instead?

kaikreuzer:
I am very much against such a feature. In openHAB, there is a single semantic that is true for a Rollershutter. This is also expected by icons etc.
Bindings simply must map to these semantics. And yes, in case it isn't clear for a binding, which way the external device works, it must give the user a chance to configure it correctly as part of the channel/thing configuration. But as a result, the situation for the Rollershutter item should always match the openHAB semantics.

kaikreuzer:
As nobody seems to challenge my opinion above, let me take the liberty to close this PR.

IcyT:
> As nobody seems to challenge my opinion above, let me take the liberty to close this PR.

Hi @kaikreuzer
who defined the current semantics about 0 being opened and 100 being closed? When I asked my wife what 20% would mean for her, she had the same answer like my thoughts: 20% opened instead of 20% closed.

I know I'm a bit late now, but I'm challenging the semantics and would like to know the reasons for that decision.

kaikreuzer:
It is the value of a shutter, i.e. it answers "how much is it shut". It is also used e.g. for awnings, where it also makes sense that 0% means it is not there, and with 100% it is fully there. Same e.g. for screens (I want to see 90% of it) or other things.

IcyT:
Ok, that explanation makes sense for me and I already convinced my wife, too 😀
Thanks, Kai!

Issue #1252, opened 2019-12-01
Recent incompatibility with Jython 2.7.2b2


openhab-5iver:
I've been using [Jython 2.7.2b2](https://mvnrepository.com/artifact/org.python/jython-standalone/2.7.2b2) without issues since it came out earlier this month, but only in a test environment that was on S1749. After updating OH to S1760, I'm getting the following errors. In production, I use a build of Jython 2.7.2a1+ from earlier in the year and it works without issues on S1764, so it looks to be something that changed in OH.

```
2019-12-01 04:43:47.973 [ERROR] [org.openhab.core.automation.module.script] - bundle org.openhab.core.automation.module.script:2.5.0.201912010653 (229)[org.openhab.core.automation.module.script.internal.ScriptEngineManagerImpl(273)] : The addScriptEngineFactory method has thrown an exception
java.lang.ExceptionInInitializerError: null
	at org.python.core.PyJavaType.isAccessibleClass(PyJavaType.java:656) ~[?:?]
	at org.python.core.PyJavaType.init(PyJavaType.java:450) ~[?:?]
	at org.python.core.PyType$Registry.createType(PyType.java:430) ~[?:?]
	at org.python.core.PyType$Registry.addFromClass(PyType.java:379) ~[?:?]
	at org.python.core.PyType$Registry.resolveType(PyType.java:296) ~[?:?]
	at org.python.core.PyType$Registry$1.computeValue(PyType.java:210) ~[?:?]
	at org.python.core.PyType$Registry$1.computeValue(PyType.java:204) ~[?:?]
	at java.lang.ClassValue.getFromHashMap(ClassValue.java:227) ~[?:1.8.0_222]
	at java.lang.ClassValue.getFromBackup(ClassValue.java:209) ~[?:1.8.0_222]
	at java.lang.ClassValue.get(ClassValue.java:115) ~[?:1.8.0_222]
	at org.python.core.PyType.fromClass(PyType.java:2056) ~[?:?]
	at org.python.core.PyJavaType$Constant.<clinit>(PyJavaType.java:96) ~[?:?]
	at org.python.core.PyType$Registry.createType(PyType.java:417) ~[?:?]
	at org.python.core.PyType$Registry.addFromClass(PyType.java:379) ~[?:?]
	at org.python.core.PyType$Registry.resolveType(PyType.java:296) ~[?:?]
	at org.python.core.PyType$Registry$1.computeValue(PyType.java:210) ~[?:?]
[...]

PR #1253, opened 2019-12-01
[WIP][automation] Add JythonScriptEngineFactory

openhab-5iver:
... w/ core and custom helper libraries! Requires #1251. Fixes https://github.com/openhab/openhab2-addons/issues/4801. Fixes https://github.com/openhab-scripters/openhab-helper-libraries/issues/48. This PR dramatically simplifies the installation of Jython and the Jython core and community helper libraries.

Questions:

* Should this project go into OHC or openhab2-addons? My preference is to keep all automation in OHC and eventually split it out into another repo, as mentioned [here](https://github.com/openhab/openhab2-addons/pull/6399#issuecomment-559953076).
* Was the copyOnWriteArray needed in ScriptModuleTyeProvider? I don't think so, since concurrent writes should never happen. WDYT?
* I've unrolled the Jython jar. Is there an issue with the LICENSE file?
* I would like to have this and other custom ScriptEngineFactories display in Paper UI as optional installations. I don't know how to do this but made some changes to the feature.xml that may have this one installing along with the new rule engine. I need help with this (please).

I first used Jython 2.7.2b2, but it no longer works with recent builds of OH (works with S1749 though), so I used Jython 2.7.1. See #1252. I have also frequently seen exceptions when using 2.7.2b2 [caused by this issue](https://bugs.jython.org/issue2487), but never with manual installs.

I have another PR that makes a custom NashornScriptEngineFactory, but I'll wait on that one until the decisions have been made for this one. If custom ScriptEngineFactories or to go into openhab2-addons, then NashornScriptEngineFqactory should be moved there too, which will be difficult to have it load by default.

I marked this as WIP so that I can update the helper libraries in the PR before it is merged.

@openhab/core-maintainers, is it possible to fasttrack this one so that it can be included in the 2.5 stable release? I realize it is short notice, but since it is new functionality, if it doesn't work (it does!), it is not breaking anything... and could be easily just hidden. Having this in the release will be a major plus!


bundles/pom.xml
```diff
@@ -22,6 +22,8 @@
     <module>org.openhab.core.automation.module.media</module>
     <module>org.openhab.core.automation.module.script</module>
     <module>org.openhab.core.automation.module.script.rulesupport</module>
+    <module>org.openhab.core.automation.module.script.scriptenginefactory.jython</module>
+    <module>org.openhab.core.automation.module.script.scriptenginefactory.nashorn</module>
```
tavalin: Should this be included in this PR? Or the separate nashorn one?

jpg0:
@openhab-5iver I'm happy to bundle library scripts into the [JS PR]((https://github.com/openhab/openhab2-addons/pull/6399#issuecomment-559953076)) if you think this is the way we'll be going?

jpg0:
@openhab-5iver I have a question about paths for Jython. Is the assumption that the personal/community/core folders are for OH-specific scripts only, and that 3rd party existing scripts should be installed directly in `$OPENHAB_CONF/lib/python`? (I ask because the JS module support in PR does ***not*** currently work that way.)

bundles/pom.xml
openhab-5iver: Thank you... yes, this is for another PR and will be removed.

openhab-5iver:
> I'm happy to bundle library scripts into the JS PR if you think this is the way we'll be going?

Let's get to that in your graaljs PR, which I should be able to get reviewed today. Hopefully that one can also get into 2.5 stable!

> Is the assumption that the personal/community/core folders are for OH-specific scripts only, and that 3rd party existing scripts should be installed directly in $OPENHAB_CONF/lib/python?

You may be confusing scripts and modules. Scripts go to /automation/jsr223/python/. Modules go to /automation/lib/python/, which is where python.path points to. I recommend 3rd party modules be put in /automation/lib/python/personal/.

jpg0:
I ask this about paths because in the JS code, if current searches 3 paths, in order:
- `/automation/lib/personal`
- `/automation/lib/community`
- `/automation/lib/core`

I originally did it this way thinking that Jython was doing this too, until I saw the `core.___` imports. The reason that I did it that way was:
- script writers can choose to override core or community modules by dropping their versions into the personal directory (I have used this, for example)
- likewise for the community improving on core
- scripts _themselves_ don't need to care about the classification (personal/community/core) of modules
- but the main reason is that 3rd party scripts can exist in these folders unmodified. If the folder name (personal in your example) is required in import statements, then any 3rd party module with a transitive, absolute dependency will fail (as they will also have to import from a 'personal' prefix).

I just wanted to raise this as I noticed it as a divergence between the languages, and I cannot see how a single root can work (at least with JS), unless 3rd party scripts go directly into that root.

bundles/org.openhab.core.automation.module.script.scriptenginefactory.jython/src/main/java/org/openhab/core/automation/module/script/scriptenginefactory/jython/JythonScriptEngineFactory.java
```diff
@@ -0,0 +1,79 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(JythonScriptEngineFactory.class);
+    private static final String SCRIPT_TYPE = "py";
+    private static javax.script.ScriptEngineManager ENGINE_MANAGER = new javax.script.ScriptEngineManager();
+
+    static {
+        String home = JythonScriptEngineFactory.class.getProtectionDomain().getCodeSource().getLocation().toString()
+                .replace("file:", "");
+        String openhabConf = System.getenv("OPENHAB_CONF");
+        System.setProperty("python.home", home);
+        System.setProperty("python.path", openhabConf + "/automation/lib/python");
```
jpg0: Currently, afaik, the `.../lib/<language>` path is a convention, rather than required in OH. This change switches it to make it required.
I have no problem with this, but wanted to raise it because I also embedded it in the JS plugin, but made it configurable. We should probably adopt the same approach; I'm happy to go with the simpler code of forcing it to a predefined value if you think this is the superior option.

jpg0:
Hey, I noticed that you moved the import style further towards the standard python import syntax! Great! This is nearer to the JS implementation, with the only difference being that I've sub-namespaced the 'scope' lib to 'RuleSupport' etc, so the equivalent in python would be:
` from core.jsr223.scope.RuleSupport import TriggerBuilder`
(rather than first running importPreset, then getting from the single set of scopes).

I've submitted a PR which simply returns the scopes from the importPreset calls (as well as requesting the factory import them like before, where it can choose to do what it likes), which would probably mean that it would be a pretty trivial change to switch to this style so that python scripts use a completely standard import style.

bundles/org.openhab.core.automation.module.script.scriptenginefactory.jython/src/main/java/org/openhab/core/automation/module/script/scriptenginefactory/jython/JythonScriptEngineFactory.java
openhab-5iver: This is the directory structure convention suggested in the HL documentation, but for the JS HLs, it is required. However, I plan to add configuration options. Yesterday, I changed this to append the python.path, in case one was previously defined. This allows you to add third party libs in other locations. I also plan to add an option to set python.home, so that you can use this feature to get the HLs and use a custom Jython install (different version or full install).

openhab-5iver:
> Hey, I noticed that you moved the import style further towards the standard python import syntax! Great! 

Sorry, I don't follow what you mean by that. I saw the new PR, but got sucked into identifying the name of a UI rule when reviewing #1196 yesterday! I plan to shelve that today and get to the graaljs PR review.

We really need a maintainer to clarify whether these should be in OHC or openhab2-addons. @kaikreuzer, since you will ultimately make the decision, could you please move this forward (first bullet in OP)?

jpg0:
> What changes are you referring to?

That you can directly: `from core.jsr223.scope import ON, OFF, OPEN, CLOSED`
(rather than `importPreset` and rely on things being dropped in the global namespace)

bundles/org.openhab.core.automation.module.script.scriptenginefactory.jython/src/main/java/org/openhab/core/automation/module/script/scriptenginefactory/jython/JythonScriptEngineFactory.java
jpg0: Something to note: in my JS implementation the module loader takes care of locating and loading the libs. This means that there is no more JS code attempting to do it, so it ceases to be required any more than it is in python. It basically brings it inline with (a) standard JS dev envs and (b) the way Python works. The equivalent of:
`from foo import bar`
is:
`const { bar} = require('foo')`

openhab-5iver:
@cweitkamp, could you please help with this?

cweitkamp:
@openhab-5iver Yes, I can do that. But let say some things in advance:

How should I know that this  PR is ready as long as it is marked as WIP?

I am not a Python expert. Thus it would make sense if anyone else could have a look into this one too. Do you know a volunteer? @openhab/core-maintainers Some one of you?

I think you know that we are already in code freeze phase for the upcoming release - apart from critical fixes. So I am afraid this could not make it into 2.5 release.

openhab-5iver:
THANK YOU... I was getting worried!

> How should I know that this PR is ready as long as it is marked as WIP?

I will put in another commit here shortly that will append the python.path to one that already exists. I will also be downgrading Jython to 2.7.0, since I am seeing the same errors in 2.7.1 that I reported in the OP about 2.7.2b2.  The only other changes are to update the helper libraries with the latest when the review is complete. I have a number of PRs from Rich to merge, but that shouldn't hold this up.

The 51 files in this PR are mostly [helper libraries](https://github.com/openhab-scripters/openhab-helper-libraries), which are well proven! Other than the pom and bnd files, there is just the JythonScriptEngineFactory to review. I did not expect that the python files would be part of the review. If this is an issue, I could pull them out, or make a separate bundle for just vanilla Jython. I thought it wold be more confusing for the user to have multiple bundles for Jython, since the user can just choose not to use the helper libraries. The biggest area that I could use help with is to have this bundle show as a feature in Paper UI along side the NGRE. I think I may have gotten this done, but I did not see any way to test it without building the distribution, which is a bit of black magic to me.

> I think you know that we are already in code freeze phase for the upcoming release - apart from critical fixes. So I am afraid this could not make it into 2.5 release.

I realize we are close to a release, which is why I was trying to get attention over here for the last week! My understanding was that [the code freeze is December 8](https://community.openhab.org/t/2-5-release-planning/85672). This bundle is new and does not effect other functionality, so I would think of it as extremely low risk. It has also been [tested by several in the community](https://community.openhab.org/t/beta-testers-wanted-jython-addon-w-helper-libraries/86633). 

The decision still remains as to where this should live. As I have communicated, my preference is to put all automation in core, with the expectation that it would be split out for OH3. If OHC is now completely frozen, then I could potentially move this to openhab2-addons, but I would need help with where to put it and what type of addon it should be. I really don't like that option, but having this in 2.5 will be HUGE for scripted automation!

Either way, thank you. This will make it in eventually!

wborn:
> Should this project go into OHC or openhab2-addons?

If the result is that jython-standalone (40MB) is packaged into each openHAB runtime download I wouldn't add it to OHC for now unless we find some way to make this optional.

> I am not a Python expert.

Me neither and I think it would be better to manage the code of such scripting libraries in another repo. Then the library artifacts can be packaged into the distro and perhaps it can also be made configurable for users what scripting libraries (if any) they want to use.

> I think you know that we are already in code freeze phase for the upcoming release - apart from critical fixes. So I am afraid this could not make it into 2.5 release.

I fully agree with that testing/polishing the 2.5 release currently has the highest priority.

kaikreuzer:
Sorry guys, I have to admit that this issue went unnoticed for me, most likely as I didn't expect any major stuff around scripting still going into 2.5 and this PR being marked as WIP, I considered it to be relevant for after release only (didn't read the description, which would have made it clear to me...).

> I fully agree with that testing/polishing the 2.5 release currently has the highest priority.

Same for me - we only have one last day to review & merge feature changes. This really does not leave us any time for discussions on how to best do the implementation.

I understand that this is frustrating for you, but maybe we should see it from the positive side: Once 2.5 is out, our main branch is 3.x and one of its main new features will be about rules&scripting - so PRs like this fit perfectly into that theme, don't they?

openhab-5iver:
> Me neither and I think it would be better to manage the code of such scripting libraries in another repo.

Wouter, this confuses me a bit, because the helper libraries _are_ [maintained in another repo](https://github.com/openhab-scripters/openhab-helper-libraries). Or are you meaning an OH repo... like what I'd mentioned about splitting automation from OHC?

I've included the Core and Community libraries in this bundle, with the intent that rule templates can be created with Scripted Conditions and Actions that can utilize these libraries. The HL repo has script examples that can be distributed through rule templates. At some point, we need to discuss the setup of an ExtensionService for rule templates, module types, etc.

The installation of this bundle is much easier than [what is currently needed](https://openhab-scripters.github.io/openhab-helper-libraries/Getting%20Started/Installation.html) to set things up. The HLs are so small (~400KB), that it made more sense to me to include all of them rather than package them separately. The Core and Community libraries should not be modified, which is another plus to having them in the bundle. 

Going forward, I'd like to create releases in the HL repo that correspond with OH releases. The helper libraries are updated frequently, so the bundle would require frequent updates. So it may make more sense to package the HLs separately, I just do not know yet if that will be possible due to the tricks that I used to get it working in the first place.

 > I understand that this is frustrating for you

I don't really have words after hearing that it won't make it into 2.5... at least none that would be appropriate! After working on this for a year, it's much more than frustration for me, but moving on... 

If for some reason there is a delay of the 2.5 release, then hopefully this decision can be revisited. If at least #1251 could be merged, it will make it easier for this to be distributed for people to manually install. I have made the changes requested in the review for that PR.

wborn:
> Or are you meaning an OH repo... like what I'd mentioned about splitting automation from OHC?

Yes I'm aware that these files already have their own GitHub repo. But it would be better if we could just use them using some released artifact (bundle/archive). That way won't have to duplicate the code in this repo and the only PR content we would have to review are version number changes. It would also prevent unknowing contributors from creating PRs to the wrong repo.

If we create a separate OH repo for all script libraries we might not even see such PRs and we could integrate them into the builds and release pipeline.

This all needs to be thought well through especially if it impacts openhab-core which we consider to be a framework that can also be used for building other solutions on top.

jpg0:
I would also encourage considering using the native packaging/distribution mechanism specific to the scripting languages. This has both drawbacks (the user needs the tool - npm/pip/whatever installed), and benefits (dependencies to and from core are checked and resolved, brings script development closer to the environment expected by the scripting language, etc).

openhab-5iver:
I've updated this to OH 3.0.

> But it would be better if we could just use them using some released artifact (bundle/archive).

Thank you... that makes a lot of sense. I'll try this out... https://help.github.com/en/github/managing-packages-with-github-packages/configuring-apache-maven-for-use-with-github-packages#installing-a-package. 

I would appreciate some more input on the questions in the OP.

> I would also encourage considering using the native packaging/distribution mechanism specific to the scripting languages. 

I disagree. I've thought about creating a whl for the Jython helper libraries, but IMO OH users would prefer to not have to use a separate installation mechanism.

jpg0:
> I disagree. I've thought about creating a whl for the Jython helper libraries, but IMO OH users would prefer to not have to use a separate installation mechanism.

I agree with you that by default, we should not require users to use another installation mechanism. If we always tie the version of helper libraries with the engine then this naturally means that we don't need to support another installation mechanism for them at all. The reason I still think it's worth looking at language-native solutions is for distribution of shared functionality written in scripting languages - either specifically for OH, or not (for example I'm thinking of how to handle transitive dependencies). Whether core helper-libraries' updates are supported via this mechanism or not I don't think matters too much (as long as they never require explicit installation, as you suggest).

(I have not yet included JS helper libraries with the JS bundle I submitted a PR for, but this would be fairly straightforward and satisfy all of the above: no default installation required for core, but would allow independent updates to it via `npm`, and also allows standard installation of arbitrary 3rd party code via `npm` the same way.)

openhab-5iver:
Replaced by https://github.com/openhab/openhab-addons/pull/7208

PR #1256, opened 2019-12-02
[automation] Added API for ScriptEngineFactory implementations to pull presets

jpg0:
Existing API only allows engine factory implementors to request presets are pushed into existing scopes, rather than returning them directory, so that they can be bound into module or library systems. This change allows implementors to capture the presets requested (and potentially ignore any pushed), and therefore expose them via idiomatic import mechanisms for the language being implemented.

Specifically, I am using this for Javascript CommonJS module support. Without this change it's possible, but ugly (using reflection) and tricky (using weakrefs to ensure engines can be collected). This change is minimal and backwards compatible with the current API.

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptExtensionManagerWrapper.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptExtensionManagerWrapper.java
@@ -17,6 +17,7 @@ import java.util.Map;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.openhab.core.automation.module.script.ScriptEngineContainer;
+import org.openhab.core.automation.module.script.ScriptEngineFactory;
 import org.openhab.core.automation.module.script.ScriptExtensionProvider;
 
 /**
@@ -57,6 +58,21 @@ public class ScriptExtensionManagerWrapper {
         return manager.getDefaultPresets();
     }
 
+    /**
+     * Imports a collection of named host objects/classes into a script engine instance. Sets of objects are provided
+     * under their object name, and categorized by preset name. This method will import all named objects for a specific
+     * preset name.
+     *
+     * @implNote This call both returns the imported objects, and requests that the {@link ScriptEngineFactory} import them.
+     * The mechanism of how they are imported by the ScriptEngineFactory, or whether they are imported at all (aside from
+     * being returned by this call) is dependent of the implementation of the ScriptEngineFactory.
+     *
+     * @apiNote Objects may appear in multiple named presets.
+     * @see ScriptExtensionManager
+     *
+     * @param preset the name of the preset to import
+     * @return a map of host object names to objects
+     */
     public Map<String, Object> importPreset(String preset) {
         return manager.importPreset(preset, container.getFactory(), container.getScriptEngine(), container.getIdentifier());
     }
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptExtensionManagerWrapper.java
```diff
@@ -56,7 +57,7 @@ public Object get(String type) {
         return manager.getDefaultPresets();
     }
 
-    public void importPreset(String preset) {
-        manager.importPreset(preset, container.getFactory(), container.getScriptEngine(), container.getIdentifier());
+    public Map<String, Object> importPreset(String preset) {
```
cweitkamp: I know that you are not the initial contributor of these classes but may I ask you to add some JavaDoc to these methods?

jpg0: I've added docs to the this method. TBH I'm not so sure about the other methods as I haven't used them. I know that most of the actual implementation is in `ScriptExtensionManager`, which unfortunately is also missing docs.

openhab-5iver:
Please do not merge this! It requires extensive testing and I very much question the need for these changes. I would very much like to have the time to review this PR myself, but my time is very limited. I am still reviewing the GraalJS addon.

jpg0:
@openhab-5iver have you looked at the code changes in this PR? It's got to be one of the most benign changes there is. Why do you say that it needs extensive testing? All it does is return a collection of items that is also pushed elsewhere. There is next to no logic involved, and nothing (yet) actually uses the returned objects.

I don't agree that it means the docs would require changes; existing code can operate exactly as before. I do agree that it introduces an extra feature which we could choose to document (`importPreset` returns a map of objects), but personally I don't think that we should, as I don't think that we have agreed that it should be used that way, nor do either of us have any plans to use it that way.

cweitkamp:
@openhab-5iver I did not test anything. Just code review.

openhab-5iver:
> It's got to be one of the most benign changes there is. 

There is nothing benign in automation, especially scripted automation! Small changes can have huge effects. If the maintainers don't merge them before I get a chance, I have been doing extensive functional testing on all automation PRs.

>  but personally I don't think that we should, as I don't think that we have agreed that it should be used that way, nor do either of us have any plans to use it that way.

And that is exactly my hesitation about this PR, until I have a better understanding of what it does, what it will be used for, and what impacts there are to existing functionality. 

If new features are being added, the documentation should definitely be updated.

cweitkamp:
@jpg0 @openhab-5iver What is the status of your automation PRs. We would like to start working on #1290 which might result in some merge conflicts. Can we finalize some of them before?

openhab-5iver:
#1253 and https://github.com/openhab/openhab-addons/pull/6399 (which I have other concerns with) are hung up until a decision is made about where ScriptEngineFactories should live, but there is a deeper question about splitting automation from core. I plan to submit an issue for that over the holidays. 

I made a comment in #1196. 

#1256 and #1264 are tailored for the graaljs ScripEngineFactory and require extensive testing. Jonathan has not reported what, if anything, has been tested. I have not tested these for any language. This code is fragile and I believe it is not well known by any maintainers, so I advice being very cautious with any changes to this code. These PRs are only introducing functionality that would be useful with Jonathan's implementation of the graaljs ScriptEngineFactory. It would be best to address that PR first before these. If I can keep out of the eggnog, I hope to get to them this year.

jpg0:
> #1256 and #1264 are tailored for the graaljs ScripEngineFactory

> These PRs are only introducing functionality that would be useful with Jonathan's implementation of the graaljs ScriptEngineFactory.

These were explicitly submitted independently because they target any language, not only graaljs. The first allows any script to _get_ a preset (rather than requesting it be injected) and the second provides an API to listen for scripts being unloaded. They are in no way related to graal or javascript.

PR #1259, opened 2019-12-07
fix multiple MQTT subscriptions

J-N-K:
Paho and HiveMQ have differences in handling multiple subscriptions to the same topic. This ensures that only one subscription within the client is made even if more than one callback is subscribing to a given topic.

See https://github.com/openhab/openhab2-addons/issues/6488

### DIFF
--- a/bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
+++ b/bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
@@ -72,16 +72,14 @@ public class MqttBrokerConnection {
      * MQTT transport protocols
      */
     public enum Protocol {
-        TCP,
-        WEBSOCKETS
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -561,14 +560,21 @@ public boolean hasSubscribers() {
...
-                if (t == null) {
-                    logger.trace("Subscribed {} to topic {}", subscriber, topic);
-                    future.complete(true);
-                } else {
-                    future.completeExceptionally(new MqttException(t));
-                }
-            });
+            if (!client.isSubscribed(topic)) {
+                // "real" subscription for first subscriber
+                client.subscribe(topic, qos, clientCallback).whenComplete((s, t) -> {
```
jochen314: This is not threadsafe, or is it?
Two threads could both do the check for isSubscribed and then do the subscription.
The checking and the adding to the list need to be atomic.

PR #1261, opened 2019-12-08
make MQTT subscription tracking thread-safe

J-N-K:
Following  https://github.com/openhab/openhab-core/pull/1259#discussion_r355141114

### DIFF
--- a/bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/client/MqttAsyncClientWrapper.java
+++ b/bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/client/MqttAsyncClientWrapper.java
@@ -12,9 +12,9 @@
  */
 package org.eclipse.smarthome.io.transport.mqtt.internal.client;
 
-import java.util.ArrayList;
-import java.util.List;
+import java.util.*;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CopyOnWriteArraySet;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
@@ -35,7 +35,7 @@ import org.slf4j.LoggerFactory;
 @NonNullByDefault
 public abstract class MqttAsyncClientWrapper {
     private final Logger logger = LoggerFactory.getLogger(MqttAsyncClientWrapper.class);
-    private final List<String> subscribedTopics = new ArrayList<>();
+    private final Set<ClientCallback> subscriptions = new CopyOnWriteArraySet<>();
 
     /**
      * connect this client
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/client/MqttAsyncClientWrapper.java
```diff
@@ -35,7 +35,7 @@
 @NonNullByDefault
 public abstract class MqttAsyncClientWrapper {
     private final Logger logger = LoggerFactory.getLogger(MqttAsyncClientWrapper.class);
-    private final List<String> subscribedTopics = new ArrayList<>();
+    private final Set<ClientCallback> subscriptions = new CopyOnWriteArraySet<>();
```
jochen314: I am not sure, if CopyOnWriteArraySet is the best implementation choice here.
All the operations we are doing are add and remove, which are the expensive ones.
I would have gone for <code>ConcurrentHashMap.newKeySet()</code>.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/client/MqttAsyncClientWrapper.java
```diff
@@ -72,23 +72,29 @@
...
+        boolean needsSubscription = subscriptions.add(clientCallback);
+        if (needsSubscription) {
+            logger.trace("Trying to subscribe {} to topic {}", this, topic);
+            return internalSubscribe(topic, qos, clientCallback).thenApply(s -> {
+                logger.trace("Successfully subscribed {} to topic {}", this, topic);
+                return true;
+            });
+        } else {
+            logger.trace("{} already subscribed to topic {}", this, topic);
+            return CompletableFuture.completedFuture(true);
```
jochen314: We could use the result value of the future to indicate, that a real subscription was not needed.
Then we could still do the debug messages in the connection.
Not sure, id that will give us much benefit.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/client/MqttAsyncClientWrapper.java
J-N-K: good point.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/client/MqttAsyncClientWrapper.java
J-N-K: I don't think this gives much benefit and currently either true or exceptionally is expected.

PR #1262, opened 2019-12-08
added Californium to compile bom

kaikreuzer:

### DIFF
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -88,6 +88,23 @@
       <scope>compile</scope>
     </dependency>
 
+    <!-- Californium -->
+    <dependency>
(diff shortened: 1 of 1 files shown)


J-N-K:
I don‘t think this is needed (at least if it is to prevent embedding). If we want to force a single version, ok for me.

kaikreuzer:
> I don‘t think this is needed

It brings the CoAP support on the same level as UPnP - the libs are made available at compile time to the bindings, so that all they need to do is to declare the feature dependency (but no further Maven dependencies, which in turn would require a "no.embed" declaration", which seemingly is often forgotten, see https://github.com/openhab/openhab2-addons/pull/6544).

PR #1264, opened 2019-12-09
[automation] Add script extension to allow scripts to track their disposal

### LINKED ISSUE #1260: [automation] Provide unload hook for script modules
Currently scripts have 3 defined entry points (other than any they create themselves):
- The top of the script, executed first
- A 'scriptLoaded' function, which is executed immediately afterward (I'm not sure what this provides over convenience - the script can presumably execute this function itself as it's last statement?)
- A 'scriptUnloaded' function, which is executed once the script is disposed.

Whilst this works for scripts, any modules / libraries have essentially only the first option. I don't believe that lack of the 2nd matters, but the 3rd does. If there is a module which needs to cleanup afterwards then it can only do so by being notified by consumers from their 'scriptUnloaded' function. This can be tough to do properly and will regularly be missed by script writers.

A better option would be allow modules to register for a disposal hook, similar to Java shutdown hooks. This could be provided via a host object / preset where code in the module could register their own functions as listeners.

Note that this is currently impossible to achieve from an additional bundle providing a language factory because the factory has no visibility over the script engines' lifecycle and is not notified upon their disposal.

jpg0:
Add a default preset so that scripts, and more importantly modules, can call `lifecycleTracker.addDisposalHook(<disposal function>)`. This will be called when the engine is disposed.

Fixes #1260 

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/defaultscope/LifecycleScriptExtensionProvider.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/defaultscope/LifecycleScriptExtensionProvider.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2010-2019 Contributors to the openHAB project
+ * Copyright (c) 2010-2020 Contributors to the openHAB project
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information.
@@ -92,4 +92,3 @@ public class LifecycleScriptExtensionProvider implements ScriptExtensionProvider
         void dispose();
     }
 }
-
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/defaultscope/LifecycleScriptExtensionProvider.java
```diff
@@ -0,0 +1,95 @@
+/**
+ * Copyright (c) 2010-2019 Contributors to the openHAB project
```
kaikreuzer:
```suggestion
 * Copyright (c) 2010-2020 Contributors to the openHAB project
```

PR #1274, opened 2019-12-15
eliminate explicit indexes & re-resolve

rotty3000:


rotty3000:
Hey folks, I sent this because I wanted to show how it might be possible to eliminate the need for explicit OSGi indexes. Not that they are a bad thing, but they do cause some friction with IDEs and with developer knowledge and expectations of with how maven works (as can be seen in the IDE use case, or adding new Maveb repos for instance which should work completely seamlessly now). The latest and greatest bnd approach is to simply rely on plain old Maven BOMs.

I don't really expect this PR to be merged. It's more of a proof of concept. And it could be cleaned up even more than this. So, please let me know what you think and if you have any questions don't hesitate to ask.

Do note that I did make all the integration tests pass locally, but we'll see what CI says.

Anyway, I'm quite impressed with OpenHABs adoption of bnd maven plugins and general process and if I can be of any more assistance please let me know.

maggu2810:
Hi @rotty3000,

the idea of the different artifacts in the `bom` (perhaps the naming has been wrong as this are not a BOM similar to the Maven naming convention but that is another story) directory has been:

* compile: the dependency section contains all dependencies of non openHAB bundles that are used for compilation
* compile-model: as openHAB is using Xtext and the relevant bundles needs some additional stuff that should not be available for the remaining bundles, use an additional dependency section that will be used by the openHAB Xtext bundles only
* openhab-core: the dependency section contains all openHAB Core bundles. So it can be used by other ones (e.g. openHAB 2 Addons, integration tests, ...) to get all the openHAB Core bundles available
* runtime: the dependency section contains all dependencies of non openHAB bundles that are used at runtime (e.g. for the integration tests, the demo application that is started inside the IDE, ...)
* test: the dependency section contains additional bundles for the test execution only

The directory also contains some `index` projects. A index project depend on the project using the same name without the "-index" postfix and creates an OSGi repository index file. That's all for the index artifacts (okay, to be more precise the test-index also replaces some test artifacts with the respective OSGi aware bundles).

A comment about the Maven `scope`:

The dependencies in that POM files are using the scope `compile`.
If the bundles of this directory are used by consumers they depend on the `bom` artifacts and set the respective scope e.g. `provided` for the "compile" one, `runtime` for the "runtime" one etc. As soon as a Maven artifact depends on e.g. artifact "runtime" using scope `runtime` all the transitive dependencies are changed to `runtime`, too.
Similar if you use the "compile" one and set `provided` all the bundles of compile are not used for app resolution as this has been set to use the `runtime` ones only.

---

So, why are the "-index" artifacts used at all?

We differ between the normal unit tests -- the ones that does not rely on an OSGi environment -- and integration tests that are based on an OSGi environment.

To get the integration tests resolved and running, we need an repository (or more) that contains all the bundles necessary at runtime.

For the tests set of bundles consists of "openhab-core", "runtime" and "test" (with the replaced artifacts done explained above).

For sure we could use the implicit Maven index for every integration test only but that involves that the implicit Maven repository using "openhab-core", "runtime" and "test" needs to be generated all the time for every integration test.

The idea has been that the index used by all integration tests is generated only once at the time the "-index" artifacts have been created.
After that every integration test can re-use the already created OSGi repository XML files and do not need to generate its implicit repository with the big number of bundles all the time itself.

kaikreuzer:
Hey @rotty3000, many thanks for your suggestions to improve our use of bnd.
I must admit that I am a bit lost, though, as your PR shuffles many things around and I am not fully clear on what that means for our usage - especially taking @maggu2810's comments into consideration that he had put quite some thoughts into the current structure.
Does our use of the indexes as explained by @maggu2810 make sense to you? Are there nonetheless improvements you'd suggest we should do? If so, it would be awesome if you could follow up on it here!

PR #1277, opened 2019-12-15
Run resolver on integration tests to fix the build

cweitkamp:
- Run resolver on integration tests to fix the build

@pfink Do you think it is possible to add the resolver - Maven flag `-DwithResolver` for `itest/` folder -  to the unleash step of the build pipeline?


pfink:
> @pfink Do you think it is possible to add the resolver - Maven flag `-DwithResolver` for `itest/` folder - to the unleash step of the build pipeline?

Sure, [this ](https://github.com/openhab/openhab-distro/blob/master/launch/Jenkinsfile#L94) would be the right place.

cweitkamp:
That was a quick response. Thank.

Will those flags be applied in all repositories? I think we need them in openHAB 2 Add-ons and more.

I am not sure if we use `-DskipTests`, will the resolver be applied on the itests? Do you know that?

pfink:
Some time ago, I did a PR to ensure itests are always part of the reactor during release, even if skipTests is enabled, because it lead to issues that they were missing: https://github.com/openhab/openhab-core/pull/673

I don't know more about it.

Flag will be applied to all repos.

PR #1291, opened 2019-12-22
[mqtt] Keep track of retained messages

### LINKED ISSUE #1279: [MQTT] Second subscription to topic does not receive retained message
I call `MqttBrokerConnection.subscribe()` for the same topic for different subscribers.
The first subscriber receives the retained message.
The second subscriber does not receive the retained message.

This is causing homeassistent components not to initialize, as multiple subscriptions to the same topic can be requested by a single device (`availibility_topic`) and only one sees the message.

Do we need to keep track of retained messages in the `MqttAsyncClientWrapper`?

jochen314:
Should fix #1279

@J-N-K : What do you think  about it?


bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -114,7 +112,6 @@
      * The callback will interact with the {@link AbstractReconnectStrategy} as well as inform registered
      * {@link MqttConnectionObserver}s.
      */
-    @NonNullByDefault
```
J-N-K: Why did you remove that?

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -604,30 +591,33 @@ public boolean hasSubscribers() {
...
+                needsUnsubscribe = true;
+                subscribers.remove(topic);
+            } else {
+                needsUnsubscribe = false;
             }
-            // Remove from subscriber list
-            subscribers.remove(topic);
-            // No more subscribers to this topic. Unsubscribe topic on the broker
+        }
+        if (needsUnsubscribe) {
```
J-N-K:
```suggestion
        if (subscribers.compute(topic, (t,v) -> v.isEmpty() ? null : v ) == null) {
```

and remove everything above?

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/Subscription.java
```diff
@@ -0,0 +1,91 @@
...
+ * This class keeps track of all the subscribers to a specific topic.
+ * <p>
+ * <b>Retained</b> messages for the topic are stored so they can be replayed to new subscribers.
+ *
+ * @author Jochen Klein - Initial contribution
+ */
+@NonNullByDefault
+public class Subscription {
+
+    private Map<String, byte[]> retainedMessages = new ConcurrentHashMap<>();
```
J-N-K: If we have one `Subscription` per topic it should be enough to store the message, the topic should be the same for all messages.

jochen314: No, the topic can be a wildcard topic `homeassistant/#` so there can by multiple messages.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
jochen314: My Eclipse told me, that it was unnecessary as the outer class has it already:
`Nullness default is redundant with a default specified for the enclosing type MqttBrokerConnection`

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
jochen314: it is more complicated than that....
```
       if (subscribers.compute(topic, (t, v) -> {
            if (v != null) {
                v.remove(subscriber);
                if (v.isEmpty()) {
                    return null; // returning null here not allowed: required '@NonNull Subscription'
                }
            }
            return v;
        }) == null) {
```
And then it could be, that we try to unsubscribe, if we were never subscribed....

The question is actually: which is better to understand in a year....

jochen314:
I was thinking about making the `MqttBrokerConnection` an interface too.
The biggest problem is, that others use the constructor and they will break.
Should we introduce a `MqttBrokerConnectionFactory` now and deprecate the constructors? (in another PR)

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/Subscription.java
J-N-K: Good point

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
J-N-K: I think I added it because SAT complains about a missing annotation. Maybe the SAT check is wrong.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
J-N-K: the remove should be before, I was only thinking of the removal from the map.

jochen314: The manipulation of the subscribers map needs to be atomic. Therefor put it intoa sync-block.
But the actucal call to the `subscribeRaw` I wanted outside the sync-block to keep the block as short as posssible.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
jochen314: I think, SAT is wrong.
But I checked SAT and removed all warnings.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -189,9 +195,9 @@ public void onDisconnected(Throwable t) {
      * @param port A port or null to select the default port for a secure or insecure connection
      * @param secure A secure connection
      * @param clientId Client id. Each client on a MQTT server has a unique client id. Sometimes client ids are
-     * used for access restriction implementations.
-     * If none is specified, a default is generated. The client id cannot be longer than 65535
-     * characters.
+     *            used for access restriction implementations.
```
cweitkamp: Something looks wrong with the format of the JavaDocs. Can you apply the [OH formatter](https://www.openhab.org/docs/developer/guidelines.html#java-code) or run 'mvn spotless:apply`?

jochen314: I am a bit confused. OH formatter and spotless:apply give me the indented version...

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/openhab/core/io/transport/mqtt/internal/Subscription.java
```diff
@@ -0,0 +1,91 @@
...
+ * This class keeps track of all the subscribers to a specific topic.
+ * <p>
+ * <b>Retained</b> messages for the topic are stored so they can be replayed to new subscribers.
+ *
+ * @author Jochen Klein - Initial contribution
+ */
+@NonNullByDefault
+public class Subscription {
+
+    private Map<String, byte[]> retainedMessages = new ConcurrentHashMap<>();
```
cweitkamp: Can this property be `final`?

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/openhab/core/io/transport/mqtt/internal/Subscription.java
```diff
@@ -0,0 +1,91 @@
...
+    private Collection<MqttMessageSubscriber> subscribers = ConcurrentHashMap.newKeySet();
```
cweitkamp: Can this property be `final`?

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/openhab/core/io/transport/mqtt/sslcontext/AcceptAllCertificatesSSLContext.java
```diff
@@ -31,6 +33,7 @@
  * @author David Graeff - Initial contribution
  */
 @Deprecated
+@NonNullByDefault
```
cweitkamp: Would it make more sense to remove deprecated classes now instead of adding annotations?

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/openhab/core/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -575,13 +569,14 @@ public boolean hasSubscribers() {
...
         CompletableFuture<Boolean> future = new CompletableFuture<>();
         final MqttAsyncClientWrapper mqttClient = this.client;
         if (mqttClient != null && mqttClient.getState().isConnected()) {
-            mqttClient.subscribe(topic, qos, clientCallback).whenComplete((s, t) -> {
+            mqttClient.subscribe(topic, qos, subscription).whenComplete((s, t) -> {
                 if (t == null) {
+                    logger.trace("Successfully subscribed to topic {}", topic);
                     future.complete(true);
                 } else {
                     logger.warn("Failed subscribing to topic {}", topic, t);
```
cweitkamp: This line of code has not been touched by it looks like there is a placeholder missing in the logger message or we are passing an argument too much.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/openhab/core/io/transport/mqtt/internal/Subscription.java
```diff
@@ -0,0 +1,91 @@
...
+
+    public Subscription() {
```
cweitkamp: Remove empty constructor.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/openhab/core/io/transport/mqtt/sslcontext/AcceptAllCertificatesSSLContext.java
jochen314: The deprecated annotation was added in the process when we moved to the hivemq library.
Currently they are still in use in the public interface of the `MqttBrokerConnection`.
I think we agree, that we should refactor the `MqttBrokerConnection` into an interface and a factory. In this step we should definitly remove the deprecated functions /classes.

For this PR I was getting to use the SAT tools in depth for the first time and i wanted to see, if I was able to understand, what needs to be done for each warning. So, not needed here, but it was a side product from looking at all sat warnings for the project.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/openhab/core/io/transport/mqtt/MqttBrokerConnection.java
jochen314: Actually no. If the last argument to any slf4j logger is a Throwable, then the throwable will be logged even if there is not placeholder for the it.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/openhab/core/io/transport/mqtt/internal/Subscription.java
```diff
@@ -0,0 +1,87 @@
+/**
+ * Copyright (c) 2010-2019 Contributors to the openHAB project
```
cweitkamp:
```suggestion
 * Copyright (c) 2010-2020 Contributors to the openHAB project
```

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/openhab/core/io/transport/mqtt/MqttBrokerConnection.java
cweitkamp: Thanks for clarification. I was not aware that `t` is a `Throwable`.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/openhab/core/io/transport/mqtt/sslcontext/AcceptAllCertificatesSSLContext.java
cweitkamp: Okay. I am fine with that.

jochen314:
Can I have a status, what would need to be changed?
I am not sure I see the open points.

PR #1294, opened 2019-12-23
Renamed namespace "org.eclipse.smarthome" to "org.openhab.core"

### LINKED ISSUE #1290: Remove usage of org.eclipse namespace
We are legally obliged to change our namespace from `org.eclipse.smarthome` to `org.openhab.core` as Eclipse is a registered trademark. openHAB 3 must not have any remainders, so this refactoring must be done on master now as a very first action.

cweitkamp:
- Renamed namespace "org.eclipse.smarthome" to "org.openhab.core"

Closes #1290 

No changes in files except renaming of namespace and `QualifiedName` definitions in [bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/scoping/ScriptImportSectionNamespaceScopeProvider.java](https://github.com/openhab/openhab-core/pull/1294/files#diff-7b47306d522b547bb2add4566a2e52afR23).

~~One disabled integration test: `org.openhab.core.model.script.tests`.~~


Hilbrand:
What version of the script did you use? I was working on it too and had the same tests fail. However I first had a problem with the sed replace (earlier version had 3, now 4 sed calls) The earlier version missed namespace in the xtend among other things. I've also added a specific patch for quartz-test.properties also in same problematic test.

I was looking at why it didn't run the tests. It doesn't seem to be able to find `DecimalType` and `QuantityType` classes in rules.

cweitkamp:
I think I used revision 2 of your script.

Yes, I saw an error related to `QuantityType` too.

Hilbrand:
I clicked your link to scriptimport but it took some time to reload. That maybe what I was missing. I've updated your changes locally and it's now running again. Also I use a separate clean local maven repo as it conflicts with current other builds.

cweitkamp:
I submitted my WIP branch for IDE https://github.com/openhab/openhab-distro/pull/1030.

cweitkamp:
Integration test for `org.openhab.core.model.script.tests` is working now.

Hilbrand:
image
With version 5 of my script.

cweitkamp:
:+1:

One question regarding the naming:

We now renamed a lot of packages from `org.eclipse.smarthome.*` to `org.openhab.*` (e.g. `org.eclipse.smarthome.config.discovery.internal.console` to `org.openhab.config.discovery.internal.console`). But the Maven artifact is `org.openhab.core.config.discovery`. Shouldn't we move / rename the packages to `org.openhab.core.*` (e.g. `org.openhab.core.config.discovery.internal.console`) instead?

kaikreuzer:
> Shouldn't we move / rename the packages to `org.openhab.core.*`

Yes, absolutely! Just as I said [here](https://github.com/orgs/openhab/teams/maintainers/discussions/2).

Hilbrand:
That does seem to make sense. I does require some more work because some classes are already in a core package and simply adding core to my script causes them to go into `core.core`. Also it is theoretical possible that classes end up in the same namespace if one is moved to core and the other was already in core. But we'll need to analyze that.

kaikreuzer:
I guess the only exception has to be made for the https://github.com/openhab/openhab-core/tree/master/bundles/org.openhab.core bundle.
Afaik, there should be no clashes - if there are, we probably have a bad API, which would need to be fixed anyhow.

Hilbrand:
I've updated the script it now puts all in core. I've pushed a branch with the changes here: https://github.com/Hilbrand/openhab-core/tree/1290/
It still compiles with tests. But I haven't run it yet.

cweitkamp:
Do you like to commit your latest changes against my branch of should we replace this one.

Hilbrand:
You can simple `git reset --hard` your branch against master and rerun the script. Would be a good test if the script not only works on my machine.

cweitkamp:
I applied your script again. The replacement part in `corePatches` did not work. After changing it manually I got a green build.

kaikreuzer:
One observation: This PR now has all files deleted and new ones created - which means that we will lose the complete history of the changes on the files (ok, not technically, but practically as something like "git blame" won't work anymore).

Would it be possible to enhance the script to do a 2-step approach? First moving the files to the new location (while keeping their content untouched), commit the changes and then change the contents of the file with a second commit? This way, I would expect git to recognize that the files were moved and not newly created. Wdyt?

cweitkamp:
@kaikreuzer I tried your proposed split.

The files are still shown as deleted in the full changes list. But the separate commits shows them as moved. Maybe GitHub cannot handle such a huge list of changes.

Hilbrand:
GitHub will show a broken history, even if you only move the file. Just browse any binding on the addons repo in relation to the move to bundle. You can still see the history via blame. That will show the original changes line by line.

wborn:
Let's go ahead with this and many thanks for working on this! I'm not that worried about blame/praise history being lost because there isn't that many to begin with since the ESH migration. I sometimes still go back there to see why code was written in a certain way.

openhab-5iver:
Where is the script located? I was curious to see if it was using `git mv`.

cweitkamp:
I would be fine with this part too. There are some minor other changes necessary to completely get rid of "ESH" (e.g. naming of folders for XML definitions `ESH-INF/`) or some properties like these:

https://github.com/openhab/openhab-core/blob/8351a2f8d41a1afa61b9da44693e2b697ebf0a20/bundles/org.openhab.core.binding.xml/src/main/java/org/eclipse/smarthome/core/binding/xml/internal/BindingXmlConfigDescriptionProvider.java#L29

But I think that can be done in a follow-up PR.

PR #1301, opened 2019-12-28
Use DTOs of Thing interface to store and restore the data

silamon:
Change the ManagedThingProvider to use ThingDTO instead ThingImpl (#731). 

<s>Breaking change: Existing thing storages are not compatible without a migration from ThingImpl to ThingDTO. This migration is not part of this PR.</s>


silamon:
What kind of unit / integration tests do you expect? 

I don't think that a simple rename would work. The structure of the persisted UIDs, either in the thing or the channels, is different and should be converted too. 
I've already started working on a 'migration' bundle that can convert existing persisted ThingImpl to ThingDTO. It's based on the [code](https://github.com/maggu2810/openhab-core/commit/b5c224e) mentioned in the parent issue and adjusted so it works once at startup and for any storage.

cweitkamp:
> What kind of unit / integration tests do you expect?

First of all my biggest concern is about data loss. I want to make sure that the conversion from `Thing` -> `ThingDTO` -> `Thing` will result in an equal object - including all previous existing properties (channels, configurations, properties, and so on). Second I like to proof that `ThingDTO` mirrors current `ThingImpl` after persisting it.

The migration must not be part of this PR. We are already discussing a migration tool which will be executed during the upgrade process of OH. I will link our discussion to keep track of it later. Your are welcome to contribute your work for such a tool.

//CC @openhab/core-maintainers

cweitkamp:
I played with your PR, did some tests and implemented a few basic unit tests in https://github.com/cweitkamp/openhab-core/commit/017a2fcb66b4c109e800a6191da727e336a31663.

You already identified a bug in `ChannelDTOMapper` as the `ChannelTypeUID` can be `null`:

https://github.com/openhab/openhab-core/blob/e63db5f5c56bdb370707a395b78ee4d7b215a9d3/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/dto/ChannelDTOMapper.java#L52

I figured a second one as the current `ChannelDTO` does not yet consider the `autoUpdatePolicy` property of the original `Channel` - I suggest to add it to the DTO - maybe in a separate PR:

https://github.com/openhab/openhab-core/blob/e63db5f5c56bdb370707a395b78ee4d7b215a9d3/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/Channel.java#L65

silamon:
@cweitkamp, thanks for testing out. I've done some tests already as well, but I'm guessing that there's a problem. The ManagedThingProvider was keeping a collection of concrete classes derived from Thing. Getting a thing multiple times would deliver the same thing from the ThingRegistry. You can set a ThingHandler from like anywhere in the code as long as you have the Thing object.  But since I'm mapping values now from ThingDTO to ThingImpl/BridgeImpl, it can be that you're working on a new copy instead. I'm not entirely sure if that might cause something somewhere.

splatch:
I was looking at this topic couple of days ago - I think it would be good to also port `ManagedItemChannelLinkProvider` and `ManagedItemProvider` to do same - they all depend on serializing a concrete class.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/ManagedThingProvider.java
```diff
@@ -31,11 +40,21 @@
  */
 @NonNullByDefault
 @Component(immediate = true, service = { ThingProvider.class, ManagedThingProvider.class })
-public class ManagedThingProvider extends DefaultAbstractManagedProvider<Thing, ThingUID> implements ThingProvider {
+public class ManagedThingProvider extends AbstractManagedProvider<Thing, ThingUID, ThingDTO> implements ThingProvider {
+
+    private final ThingTypeRegistry thingTypeRegistry;
+    private final ThingManager thingManager;
+
+    private final Map<String, Thing> things;
```
cweitkamp, line 48: Does it really make sense to store a reference for all managed things here? Wouldn't it be better to move the check for duplicates to the `ThingManager` or use the `ThingRegistry`.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/ManagedThingProvider.java
```diff
@@ -47,4 +66,37 @@ protected String getStorageName() {
...
+            thing = this.things.get(key);
+            return thing;
+        } else {
+            ThingTypeUID thingTypeUID = new ThingTypeUID(persistableElement.thingTypeUID);
+            boolean isBridge = thingTypeRegistry.getThingType(thingTypeUID) instanceof BridgeType;
+
+            Thing thing = ThingDTOMapper.map(persistableElement, isBridge);
+
+            // Add the thing handler and thing status
+            ThingHandler thingHandler = thingManager.getHandler(thing.getUID());
```
cweitkamp, line 83: Could you briefly explain why we need to get and set the `ThingHandler` and the `ThingStatusInfo` after restoring a thing? IIRC this was not necessary before because the `DefaultAbstractManagedProvider` just returns the plain element from its storage.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1249,6 +1249,18 @@ public boolean isEnabled(ThingUID thingUID) {
         return !isDisabledByStorage(thingUID);
     }
 
+    @Override
+    public @Nullable ThingStatusInfo getStatusInfo(ThingUID thingUID) {
+        Thing thing = getThing(thingUID);
```
cweitkamp, line 1254: Just as a reminder for myself or future improvement: The `ThingManagerImpl` stores things in a `Set` and iterates the elements to get the thing with given `ThingUID` in the `getThing()` method. We maybe want to replace the `Set` by a `Map`.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/ManagedThingProvider.java
silamon: I wasn't aware of the references in the <code>ThingRegistry</code>. I thought that the <code>ThingRegistry</code> was fetching from their providers on every request. So therefore the <code>ManagedThingProvider</code> was providing the complete <code>Thing</code>. As a consequence, the <code>ManagedThingProvider</code> now provides the part available in storage of the <code>Thing</code> and <code>ThingHandler</code> and <code>ThingStatusInfor</code> would be <code>null</code> if fetched directly from the <code>ManagedThingProvider</code>.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/ThingManager.java
```diff
@@ -42,4 +44,20 @@
...
      */
     public void setEnabled(ThingUID thingUID, boolean isEnabled);
+
+    /**
+     * This method returns the {@link ThingStatusInfo}.
+     *
+     * @param thingUID UID of the {@link Thing}.
+     * @return the {@link ThingStatusInfo} of the thing
+     */
+    public @Nullable ThingStatusInfo getStatusInfo(ThingUID thingUID);
```
cweitkamp, line 54: So we do not need these additional methods anymore. Do we?

kaikreuzer:
@silamon & @cweitkamp This issue is tagged for openHAB 3 release - is it the plan to get that in? If so, it would be great if it could happen within the next 7 days. Thanks!

cweitkamp:
I think this PR is ready to be merged code wise. The real problem is the structure of an existing storage because the way how Things are stored changes heavily. After this PR it will not be possible to load Things from an existing storage. As stated in https://github.com/openhab/openhab-core/issues/731#issue-433337785 and discussed above we need a solution in advance.

silamon:
> The migration must not be part of this PR. We are already discussing a migration tool which will be executed during the upgrade process of OH. I will link our discussion to keep track of it later. Your are welcome to contribute your work for such a tool.
> 
> //CC @openhab/core-maintainers

This is what keeps the PR open for now. It's a breaking change you don't want to take at the moment.

kaikreuzer:
> This is what keeps the PR open for now. It's currently a breaking change you don't want to take at the moment.

Thank you all, understood.
As we won't be able to find a solution for the database migration in the last remaining days, I guess we will have to postpone it for a 3.x release, but definitely should be looking into some migration logic then.

PR #1311, opened 2019-12-31
[archetype] new binding feature should use ohc.version

Hilbrand:
When generating a new binding with the maven archetype it should refer to openhab core via the variable `ohc.version` In the feature.xml.


Hilbrand:
Oh darn :smile: We also might need to come up with some solution to get this working with the 2.5.x branch. Because there this fix is actual relevant as it doesn't work without this fix. Because core isn't getting a  2.5.1 update anymore it's not trivial. One fix could be to add patches in the skeleton scripts of 2.5.x branch. For Linux this is fairly trivial, on Windows I'm less certain it's trivial. I'm also researching why the script doesn't work correctly on windows (it's not updating the bundle pom and CODEOWNERS file).

cweitkamp:
I see what you mean.

Additionally to get it working on the openhab-addons 2.5.x branch we have to separate the `-Dversion` from the `-DarchetypeVersion`.

PR #1317, opened 2020-01-06
Added nullness annotations to 'ThingActions' and 'ThingHandlerService'

cweitkamp:
- Added nullness annotations to 'ThingActions' and 'ThingHandlerService'
- Minor code improvements and reordering of imports

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/ThingHandlerService.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/ThingHandlerService.java
@@ -12,12 +12,16 @@
  */
 package org.openhab.core.thing.binding;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+
 /**
  *
- * Interface for a service that provides access to a {@link ThingHandler}
+ * Interface for a service that provides access to a {@link ThingHandler}.
  *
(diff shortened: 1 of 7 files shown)


bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/ThingHandlerService.java
```diff
@@ -32,6 +36,7 @@
      *
      * @return the {@link ThingHandler}
      */
+    @Nullable
```
cweitkamp: A `ThingHandlerService` will be initialized directly after a `ThingHandler` will have been initialized. Thus I am not sure if the return value of this method really could be `null`.

Current [implementations](https://github.com/openhab/openhab-addons/search?q=ThingHandlerService&unscoped_q=ThingHandlerService) are using ways - Either `@Nullable` or `@NonNullByDefault({})` to annotate a `thingHandler` property. Which one is preferred?

https://github.com/openhab/openhab-core/blob/3811cb95604f69f6a3d77da2caf391d81b05458e/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/BaseThingHandlerFactory.java#L150-L153

kaikreuzer: Note sure. Going for `@Nullable` leaves us more flexibility wrt lifecyle - who knows, if we come across a situation, where it is valid to have no thingHandler available.

PR #1331, opened 2020-01-17
switched to SAT 0.9.0

kaikreuzer:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -660,6 +660,13 @@ Import-Package: \\
         <artifactId>sortpom-maven-plugin</artifactId>
       </plugin>
     </plugins>
+    <extensions>
+       <extension>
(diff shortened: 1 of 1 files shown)


kaikreuzer:
I've added the extension, thanks.

Wrt the formats, it seems to me as if [they are already in place](https://github.com/openhab/openhab-core/blob/master/pom.xml#L565-L599)?

wborn:
There isn't a format similar to that of the changes in https://github.com/openhab/openhab-addons/pull/6681.

Maybe @Hilbrand knows if we also need this change in openhab-core repo. Perhaps it's only required to prevent issues when merging all the feature.xml files?

J-N-K:
I think it‘s related to merging. Because that is done with regex matching for complete lines iirc.

Issue #1333, opened 2020-01-18
[persistence] averageSince shouldn't return null


5iver:
The JavaDoc says that the Items state will be returned if it has not changed since the DateTime specified, [but it returns null](https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.model.persistence/src/org/openhab/core/model/persistence/extensions/PersistenceExtensions.java#L468). This should be changed. Changing the docs would be easy, but I think this would be better functionality.

cweitkamp:
To find an answer why `null` is returned you have to dive into this discussion: https://github.com/eclipse/smarthome/pull/2405

As the calculation considers the current value - if known - imo it would make sense to return the current value instead. Maybe @J-N-K can elaborate on this topic.

I think we have to change the JavaDocs anyway because the other method says to return `null` if no default persistence service is available which is not the case either:
https://github.com/openhab/openhab-core/blob/2c8de4923138f31e1cccf6cd6996ed54a990903d/bundles/org.openhab.core.model.persistence/src/org/openhab/core/model/persistence/extensions/PersistenceExtensions.java#L400-L403

J-N-K:
The problem with returning the current value is that this might be wrong. The calculation is essentially the integral of values over time `\int_{t0}^{now} y(t) dt` divided by the timespan `t0-now`. If we don't have at least two values (one stored and the current value), we can't calculate the average. I'm not convinced that returning the current state is the correct thing to do in that case.

J-N-K:
I re-read the old discussion on ESH and I think the consensus reached is still valid. I'll adjust the JavaDoc.

PR #1338, opened 2020-01-22
[persistence] Fixed JavaDoc for persistence extension

cweitkamp:
- Fixed JavaDoc for persistence extension

Related to #1333 

Question is: How should proceed with the return value of `avarageSince()` method?

### DIFF
--- a/bundles/org.openhab.core.model.persistence/src/org/openhab/core/model/persistence/extensions/PersistenceExtensions.java
+++ b/bundles/org.openhab.core.model.persistence/src/org/openhab/core/model/persistence/extensions/PersistenceExtensions.java
@@ -20,6 +20,8 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.Iterator;
 
+import org.joda.time.DateTime;
+import org.joda.time.base.AbstractInstant;
 import org.openhab.core.i18n.TimeZoneProvider;
 import org.openhab.core.items.Item;
 import org.openhab.core.library.types.DecimalType;
@@ -30,8 +32,6 @@ import org.openhab.core.persistence.PersistenceService;
 import org.openhab.core.persistence.PersistenceServiceRegistry;
 import org.openhab.core.persistence.QueryablePersistenceService;
 import org.openhab.core.types.State;
-import org.joda.time.DateTime;
-import org.joda.time.base.AbstractInstant;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Reference;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


mhilbush:
> JavaDoc

Speaking of Javadoc... This is a somewhat related question, but it appears that the [api docs](https://www.eclipse.org/smarthome/documentation/javadoc/index.html) are not longer produced for OHC (at least I can't find it). I'm not sure when this ended, either in ESH or OHC.

I tweaked my OHC pom.xml to produce it for me as I find it occasionally useful to reference (versus looking through the code). Is there interest in turning this back on? Should I open an issue for it? WDYT?

bundles/org.openhab.core.model.persistence/src/org/openhab/core/model/persistence/extensions/PersistenceExtensions.java
```diff
@@ -182,9 +182,9 @@ public static HistoricItem historicState(Item item, AbstractInstant timestamp, S
      *
      * @param item the item to check for state changes
      * @param timestamp the point in time to start the check
-     * @return <code>true</code> if item state had changed, <code>false</code> if it hasn't or if the default
-     *         persistence service does not refer to a {@link QueryablePersistenceService}, or <code>null</code> if the
-     *         default persistence service is not available
+     * @return <code>true</code> if item state has changed, <code>false</code> if it did not change or if the default
```
kaikreuzer: Isn't "has changed or has not changed" the correct English?

openhab-5iver: It looks/reads better using the same verb in both cases.

cweitkamp: Absolutely. "has changed" is preferred.

kaikreuzer:
> How should proceed with the return value of avarageSince() method?

Hm, to be discussed on the issue, I guess. Didn't yet look at the referenced discussion at ESH, which was rather long as it seems...

>  Is there interest in turning this back on? Should I open an issue for it? WDYT?

Yes, this was done on ESH, but never for openHAB. Would imho be nice to have the JavaDoc published for openhab-core.

> Should I open an issue for it? 

If you have changed your pom.xml already to build it, why not directly create a PR :-) We will then just have to see how to publish it to a suitable place.

mhilbush:
> If you have changed your pom.xml already to build it, why not directly create a PR :-) We will then just have to see how to publish it to a suitable place.

Thanks. Happy to create the PR. I cobbled together my pom.xml changes based on what was being done in ESH. But, my pom-writing skills are, how should I say, weak. I'll submit my changes in a PR. Then I can adjust as necessary.

PR #1340, opened 2020-01-24
Add filters to some REST getAll endpoints

ghys:
These filters help reduce payloads for UIs which
retrieve ad-hoc information rather than maintain
a global state.


ghys:
note that this PR is benign in that it only expands the API without breaking it 🙂

PR #1341, opened 2020-01-25
added hashCode and equals method to StringListType to fix object comparison

kaikreuzer:

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/StringListType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/StringListType.java
@@ -33,7 +33,7 @@ import org.openhab.core.types.State;
 @NonNullByDefault
 public class StringListType implements Command, State {
 
-    protected List<String> typeDetails;
+    final protected List<String> typeDetails;
 
     // constants
     public static final String DELIMITER = ",";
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/StringListType.java
```diff
@@ -100,4 +101,30 @@ public static StringListType valueOf(String value) {
...
+            return false;
+        }
+        if (getClass() != obj.getClass()) {
+            return false;
+        }
+        StringListType other = (StringListType) obj;
+        if (!typeDetails.equals(other.typeDetails)) {
+            return false;
+        }
+        return true;
```
cweitkamp: We can return the result of the comparison directly:

```suggestion
        return typeDetails.equals(other.typeDetails);
```

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/StringListType.java
```diff
@@ -100,4 +101,30 @@ public static StringListType valueOf(String value) {
         return new StringListType(value);
     }
 
+    @Override
+    public int hashCode() {
+        final int prime = 31;
+        int result = 1;
+        result = prime * result + ((typeDetails == null) ? 0 : typeDetails.hashCode());
```
cweitkamp: I think the property `typeDetails` never will be `null` as every constructor assigns some value to it. Furthermore it would be possible to add the `final` keyword to the `typeDetails` definition in [line 36](https://github.com/openhab/openhab-core/pull/1341/files#diff-4bad14d81fc15d69e4381ecc79c0f340R36).

kaikreuzer:
The code was generated by the Eclipse IDE and I didn't touch it, that's why there were still room for improvement ;-)

PR #1342, opened 2020-01-25
replaced Joda Time by Java Time

### LINKED ISSUE #1250: Replace org.joda.time with java.time and drop Joda-Time support
Since Java 8 (introduced in 2014) there is a java.time API inspired by Joda-Time. So it would make a lot of sense to use java.time everywhere instead of org.joda.time. Then we can drop Joda-Time support in OH3.

kaikreuzer:
This PR finally gets rid off Joda Time.
As discussed in the past, this will break rules from users, but as the final rule format is anyhow still under discussion, it is clear that we will need some way or another for rule migration for the 3.0 release. This PR should hence not really have much additional weight.

Fixes #1250

### DIFF
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -178,14 +178,6 @@
       <scope>compile</scope>
     </dependency>
 
-    <!-- Joda Time -->
-    <dependency>
-      <groupId>joda-time</groupId>
-      <artifactId>joda-time</artifactId>
-      <version>2.9.2</version>
-      <scope>compile</scope>
-    </dependency>
-
     <!-- jUPnP -->
     <dependency>
       <groupId>org.jupnp</groupId>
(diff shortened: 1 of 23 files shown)


kaikreuzer:
It wasn't on mine, I only was reminded [here](https://github.com/openhab/openhab-core/pull/1338/files) that Joda is still around and I thought it is really time to do something about it 😉.

kaikreuzer:
I just have to get the tests working again...

kaikreuzer:
Allo green 😅 .

meingraham:
😬 I agree that Joda time is a pain... but I have it coded in many, many rules. I can't bear to think about troubleshooting all my rules to get the new syntax right. I can't say I disagree with this move... but I'm not looking forward to migrating.

openhab-5iver:
Are you planning another PR for the documentation or should that wait until the direction for automation has been finalized?

kaikreuzer:
No, not planning a docs PR as there'll be many more changes to rules and we anyhow first need to address https://github.com/openhab/openhab-docs/issues/1133.

PR #1347, opened 2020-01-30
Units for reactive and apparent power

mrbig:
Dear  openHAB developers,

this PR tries to fix some minor issues related to the units of reactive and apparent power.

kvar existed already but I believe it was incorrectly defined. Also var (1000 var = 1 kvar) was missing from the list of units.

Unit for apparent power (VA) and it's common derivates were missing entirely.

These units are used sometimes by smart electricity meters and solar inverters.

I hope you will like these changes.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/SmartHomeUnits.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/SmartHomeUnits.java
@@ -122,9 +122,11 @@ public final class SmartHomeUnits extends CustomUnits {
     public static final Unit<Energy> WATT_HOUR = addUnit(new ProductUnit<>(Units.WATT.multiply(Units.HOUR)));
     public static final Unit<Energy> KILOWATT_HOUR = addUnit(MetricPrefix.KILO(WATT_HOUR));
     public static final Unit<Energy> MEGAWATT_HOUR = addUnit(MetricPrefix.MEGA(WATT_HOUR));
-    public static final Unit<Power> KILOVAR = addUnit(MetricPrefix.KILO(new BaseUnit<Power>("var")));
-    public static final Unit<Energy> KILOVAR_HOUR = addUnit(new ProductUnit<>(KILOVAR.divide(Units.HOUR)),
+    public static final Unit<Power> VAR = addUnit(new AlternateUnit<Power>(Units.WATT, "var"));
+    public static final Unit<Power> KILOVAR = addUnit(MetricPrefix.KILO(VAR));
+    public static final Unit<Energy> VAR_HOUR = addUnit(new ProductUnit<>(VAR.multiply(Units.HOUR)),
             Energy.class);
+    public static final Unit<Energy> KILOVAR_HOUR = addUnit(MetricPrefix.KILO(VAR_HOUR));
     public static final Unit<Force> NEWTON = addUnit(Units.NEWTON);
     public static final Unit<Frequency> HERTZ = addUnit(Units.HERTZ);
     public static final Unit<Intensity> IRRADIANCE = addUnit(
@@ -238,6 +240,8 @@ public final class SmartHomeUnits extends CustomUnits {
         SimpleUnitFormat.getInstance().label(SIEMENS_PER_METRE, "S/m");
         SimpleUnitFormat.getInstance().label(TERABIT, "Tbit");
         SimpleUnitFormat.getInstance().label(TERABIT_PER_SECOND, "Tbit/s");
+        SimpleUnitFormat.getInstance().label(VAR, "var");
+        SimpleUnitFormat.getInstance().label(VAR_HOUR, "varh");
         SimpleUnitFormat.getInstance().label(WATT_HOUR, "Wh");
         SimpleUnitFormat.getInstance().label(WATT_SECOND, "Ws");
     }
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/SmartHomeUnits.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/SmartHomeUnits.java
@@ -127,6 +127,9 @@ public final class SmartHomeUnits extends CustomUnits {
     public static final Unit<Energy> VAR_HOUR = addUnit(new ProductUnit<>(VAR.multiply(Units.HOUR)),
             Energy.class);
     public static final Unit<Energy> KILOVAR_HOUR = addUnit(MetricPrefix.KILO(VAR_HOUR));
+    public static final Unit<Power> VOLT_AMPERE = addUnit(new AlternateUnit<Power>(Units.WATT, "VA"));
+    public static final Unit<Energy> VOLT_AMPERE_HOUR = addUnit(new ProductUnit<>(VOLT_AMPERE.multiply(Units.HOUR)),
+            Energy.class);
     public static final Unit<Force> NEWTON = addUnit(Units.NEWTON);
     public static final Unit<Frequency> HERTZ = addUnit(Units.HERTZ);
     public static final Unit<Intensity> IRRADIANCE = addUnit(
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/SmartHomeUnits.java
```diff
@@ -122,9 +122,16 @@
...
     public static final Unit<Energy> MEGAWATT_HOUR = addUnit(MetricPrefix.MEGA(WATT_HOUR));
-    public static final Unit<Power> KILOVAR = addUnit(MetricPrefix.KILO(new BaseUnit<Power>("var")));
-    public static final Unit<Energy> KILOVAR_HOUR = addUnit(new ProductUnit<>(KILOVAR.divide(Units.HOUR)),
+    public static final Unit<Power> VAR = addUnit(new AlternateUnit<Power>(Units.WATT, "var"));
+    public static final Unit<Power> KILOVAR = addUnit(MetricPrefix.KILO(VAR));
+    public static final Unit<Energy> VAR_HOUR = addUnit(new ProductUnit<>(VAR.multiply(Units.HOUR)),
             Energy.class);
+    public static final Unit<Energy> KILOVAR_HOUR = addUnit(MetricPrefix.KILO(VAR_HOUR));
+    public static final Unit<Power> VOLT_AMPERE = addUnit(new AlternateUnit<Power>(Units.WATT, "VA"));
+    public static final Unit<Power> KILOVOLT_AMPERE = addUnit(MetricPrefix.KILO(VOLT_AMPERE));
```
cweitkamp: I am still wondering why we need to add units with a metric prefix. Doesn't `KILO(VOLT_AMPERE)` in a rule or binding implementation do the job at all?

Okay, `KILOVAR` and `KILOVAR_HOUR` where available before, e should keep them.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/SmartHomeUnits.java
```diff
@@ -122,9 +122,16 @@
...
+    public static final Unit<Energy> VOLT_AMPERE_HOUR = addUnit(new ProductUnit<>(VOLT_AMPERE.multiply(Units.HOUR)),
+            Energy.class);
+    public static final Unit<Energy> KILOVOLT_AMPERE_HOUR = addUnit(MetricPrefix.KILO(VOLT_AMPERE_HOUR));
```
cweitkamp: see above

bundles/org.openhab.core/src/test/java/org/openhab/core/library/unit/SmartHomeUnitsTest.java
```diff
@@ -168,6 +168,40 @@ public void testKnotUnitSymbol() {
         assertThat(SmartHomeUnits.KNOT.getSymbol(), is("kn"));
         assertThat(SmartHomeUnits.KNOT.toString(), is("kn"));
     }
+    
+    @Test
+    public void testVarUnitSymbol() {
+        assertThat(SmartHomeUnits.VAR.getSymbol(), is("var"));
+        assertThat(SmartHomeUnits.VAR_HOUR.toString(), is("varh"));
```
cweitkamp: Can you align the unit test to the others? One assertion for `getSymbol()` and one assertion for `toString()` for each unit.

```
        assertThat(SmartHomeUnits.VAR.getSymbol(), is("var"));
        assertThat(SmartHomeUnits.VAR.toString(), is("var"));
        assertThat(SmartHomeUnits.VAR_HOUR.getSymbol(), is("varh"));
        assertThat(SmartHomeUnits.VAR_HOUR.toString(), is("varh"));
```

Additionally dose it make sense to add a test for the comparison to other units of the dimension power (e.g. (`1 var = 1 W`)?

bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/SmartHomeUnits.java
mrbig: Here I did not want to break any already existing bundles, so I've kept them here.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/SmartHomeUnits.java
mrbig: I will remove the kilo, I'm absolutely fine with that.

bundles/org.openhab.core/src/test/java/org/openhab/core/library/unit/SmartHomeUnitsTest.java
mrbig: There is a problem with this check:
`assertThat(SmartHomeUnits.VAR_HOUR.getSymbol(), is("varh"));`

Namely, that ProductUnit does not calculate the symbol propert when there are multiple units. I'm not sure why is it so, but there's a fixme in the sourcecode: https://github.com/unitsofmeasurement/uom-se/blob/master/src/main/java/tec/uom/se/unit/ProductUnit.java#L111

> Additionally dose it make sense to add a test for the comparison to other units of the dimension power (e.g. (1 var = 1 W)?
I'm uncertain about that. Wikipedia sais: "The volt-ampere is _dimensionally_ _equivalent_ to the watt (in SI units, 1 VA = 1 N m A-1 s -1 A = 1 N m s -1 = 1 J s -1 = 1 W)." But as I understand there's no such thing that we convert from var to watts because you will measure different things in W/VA/var. See: https://en.wikipedia.org/wiki/Volt-ampere

How should such a test look like in your opinion?

bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/SmartHomeUnits.java
mrbig: removed.

bundles/org.openhab.core/src/test/java/org/openhab/core/library/unit/SmartHomeUnitsTest.java
mrbig: note: I've pushed a version where I've entered the getSymbol() checks, but commented them out, because they now break the tests.

mrbig:
> May I ask you to amend the new units to the [documentation](https://www.openhab.org/docs/concepts/units-of-measurement.html#list-of-units) afterwards?

No problem, I've created a PR for that as well: https://github.com/openhab/openhab-docs/pull/1146

bundles/org.openhab.core/src/test/java/org/openhab/core/library/unit/SmartHomeUnitsTest.java
cweitkamp: I did my research on German [Wikipedia](https://de.m.wikipedia.org/wiki/Var_(Einheit)). I have the feeling there is a mistake about the formular on it compared to your link. Thus it does not make sense to introduce further tests.

mrbig: > > ProductUnit does not calculate the symbol
> 
> We can open an issue in https://github.com/unitsofmeasurement/uom-se/issues and ask them to fix it .

Did some research and it looks like the do not have the intention to fix this: https://github.com/unitsofmeasurement/uom-se/issues/184
The preferred way is to use the toString() unless dealing with a specific BaseUnit.

mrbig: As I understand now the symbol property means that a unit has a symbol directly attached to it. For derived units this should be null, and toString should implement the display of the derived symbol.
The test complying with this convertion should look like this:

`assertNull(SmartHomeUnits.VAR_HOUR.getSymbol());

@cweitkamp  please let me know if you would like me to change the tests this way.

cweitkamp: Okay, I see. Thanks for the research. We should keep the `toString()` tests and omit the `getSymbol()` tests for those units.

mrbig: Thank you, updated the code.

PR #1349, opened 2020-02-01
Binding events concept

cdjackson:
This introduces a binding event notification concept that was originally discussed [here](https://github.com/openhab/openhab-addons/pull/977). At the time there was discussion how this fits in with the imminent generic notification system (that then never appeared) and it was then suggested to move this to ESH, and then dropped as ESH maintainers did not require this.

I would like to introduce this again with a view to providing a standard way for bindings to provide information to users. This provides an event factory and standard event system that could be used within a UI to present feedback to users. This has been used in the ZWave binding since OH2.0 to provide feedback about things like exclusion or errors during inclusion.

The events provide a single Type (eg severity) and text information field that can present user feedback.

Currently this is based against the (now deleted) ```2.5.x``` branch, but before I do too much work I want to quickly see if this would now be accepted here, so would welcome any feedback.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/binding/events/BindingEvent.java
+++ b/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/binding/events/BindingEvent.java
@@ -1,0 +1,34 @@
+/**
+ * Copyright (c) 2010-2019 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.core.binding.events;
+
+import org.eclipse.smarthome.core.events.AbstractEvent;
+
+/**
+ * Event for binding level user notifications
+ *
+ * @author Chris Jackson - Initial Contribution
+ *
+ */
+public class BindingEvent extends AbstractEvent {
+    public final static String TYPE = BindingEvent.class.getSimpleName();
+
+    public BindingEvent(String topic, String binding, String payload) {
+        super(topic, payload, binding);
+    }
+
+    @Override
+    public String getType() {
+        return TYPE;
+    }
+}
--- a/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/binding/events/BindingEventDTO.java
+++ b/bundles/org.openhab.core/src/main/java/org/eclipse/smarthome/core/binding/events/BindingEventDTO.java
@@ -1,0 +1,38 @@
+/**
+ * Copyright (c) 2010-2019 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.core.binding.events;
+
+/**
+ * Base BindingEvent data transfer object
+ *
+ * This can be extended by bindings to add binding specific information, but provides basic information to allow the UI
+ * to render a message the user without having to implement any binding specific processor.
+ *
+ * @author Chris Jackson - Initial Contribution
+ *
+ */
+public class BindingEventDTO {
+    public BindingEventType type;
+    public String message;
+
+    /**
+     * Constructs a binding event with the minimal required information to be sent to the user
+     *
+     * @param type the {@link BindingEventType} providing general context and severity
+     * @param message the message to be provided to the user
+     */
+    public BindingEventDTO(BindingEventType type, String message) {
+        this.type = type;
+        this.message = message;
+    }
+}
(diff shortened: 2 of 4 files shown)


cdjackson:
Apologies - I was asked why I closed this. It has been 9 months since this was opened and despite hundreds of PRs being processed, this has had no feedback from the maintainers at all! OH3 milestone is now released and the concept of "quick" feedback seems to be lost.

kaikreuzer:
@cdjackson Please don't take anything personally here - I guess you are yourself well aware how difficult it is to juggle with priorities as a maintainer and how easily issues drown in inboxes and are left unanswered.

I think I was already heavily involved in the previous, linked discussion, so it isn't right that there has not been any feedback. I already stated my opinion and the my main argument was that we should have more than a single binding requiring such a feature. As you might have seen, I even referenced your idea very recently in https://github.com/openhab/openhab-webui/issues/374 to help finding supporters for it.

You also have to admit that the PR itself is worthless as long as there is no concept on how those events should be handled by the UI - and that is the major point, which brings a huge complexity. All the other "hundreds of PRs" where likely of a much smaller scope, less complexity and less open related questions and thus were easier to process and merge.

Feel free to re-open it, but honestly I think it would be even better as an RfC issue (where you will find it in a list of 163 other issues with ideas and suggestions that haven't been answered and discussed in detail either...) - typical management with limited resources, sorry...

cdjackson:
>  we should have more than a single binding requiring such a feature.

I'm not sure what you mean when you say this? The system allows any number of bindings to generate notifications.

> You also have to admit that the PR itself is worthless as long as there is no concept on how those events should be handled by the UI 

This is very simple though - this was done in HABmin way back in OH2.0 - the UI can simply provide a toast type notification, or it could provide an historical list, or...

Unfortunately I need to move on now so I'll introduce this into a fork. This is required for me to move forward with ZWave certification - ZWave alliance requires such a system to provide feedback to users.

ghys:
FWIW, I think a central notification infrastructure is still missing in openHAB, we have:
- the need for bindings to send notifications (this PR);
- different ways to push notifications, mostly from rules or as thing actions - e.g. openHAB Cloud, HABot webpush, Telegram, etc. and they all have their own API;
- the need to store ongoing notifications until they're seen/dismissed, and display them

I think a somewhat generic pub-sub pattern would be appropriate in this case, with some facilities to push arbitrary objects to allow rich notifications, eventually tailored to one subscriber in particular - for instance, some clients could display widgets along with the message, or an image, or action buttons...

kaikreuzer:
@ghys Sounds pretty much like what was discussed in https://github.com/eclipse-archived/smarthome/pull/459 - and you might already see the inherent complexity when mentioning different notification channels, images, action buttons, etc.

I am not at all against it though and I was already [deeply involved in the early discussions](https://www.eclipse.org/forums/index.php/t/1070058/) on that.
I'd say that this is foremost a UX discussion which we would have to do and where we need to be clear on what kind of notifications (prio, properties, back-channels) should be sent through which channels (push, UI notification inbox, part of a wizard flow,...) in which situation. A technical PR like this one here would then be an outcome from that.
Imho, the best place for such a discussion would be the AC as it will have an impact on all the different components of openHAB.

kaikreuzer:
> ZWave alliance requires such a system to provide feedback to users.

@cdjackson That could probably be valuable input to the discussion. What have they identified (and thus specified) when+what+how should be notified to the user?

I understand that doing a fork is undesired additional work for you, so I'll do my best to help to align any of the concepts, so that there's a good chance to bring the forks together again. Btw, what UI are you using for the certification work? Are you using some custom UI or did you also do a fork of the openHAB UI for that purpose?

cdjackson:
> That could probably be valuable input to the discussion.

I tried to make this PR unrelated to ZWave and not get driven by what one binding wants. This notification system, as was the case with the PR I submitted to ESH, was to allow ALL bindings to send simple notifications to the user. Therefore I did not previously list ZWave specific notifications that I want to send as I don't really think that level of detail is needed here. I know you closed the ESH PR with a statement that this was only applicable for ZWave - that was never the intention and I therefore try and keep ZWave out of this. If you think that this is only a ZWave issue still, then let's close this and I will implement my own notification system as I did previously in ZWave and HABmin.

> What have they identified (and thus specified) when+what+how should be notified to the user?

There are simply a number of things that the requirements state "the user must be notified" - it doesn't say how this is done, and I don't really think we need to get into detail of the ZWave binding. However, since you ask, here are a couple of lines from the requirements. Eg -:

* MUST issue a warning message to the user informing that the node has not been included with the highest security.
* MUST inform the end user that the node will stay in the network and requires to be excluded manually

For me, a toast, or something similar is fine - this is what HABmin did with this sort of thing using exactly this notification system. Most of these functions are to provide feedback to the user during a "manual" function where the user will be watching the UI - for example during discovery. This could be extended with some sort of caching, and possibly user acknowledgement to cancel old notifications, and it could be further extended to add support for actions. However I need to get a basic functioning notification system implemented sooner rather than later so spending another 6 months discussing extended requirements doesn't really work.

> I understand that doing a fork is undesired additional work for you, so I'll do my best to help to align any of the concepts, so that there's a good chance to bring the forks together again.

The fork is undesirable, but is unfortunately unavoidable for a number of reasons now. However I would like to contribute features into OH where possible for the benefit of everyone. However I'm also in a position where I can't wait months for feedback, and sadly that may drive things directly into the fork.

> what UI are you using for the certification work?

Currently I personally use the OH3 UI (largely unmodified) for testing, however the end user for this has their own UI as part of a much broader system so that will be used (I assume) for the actual certification.

> Are you using some custom UI or did you also do a fork of the openHAB UI for that purpose?

As above - both :)

PR #1352, opened 2020-02-02
[ThingManagerImpl] Prevent multiple handler inits on thing enabling

mdicke2s:
When a thing is enabled, the `ThingManagerImpl` takes care of calling `ThingHandler#initialize()`.
That makes totally sense if the thing was disabled before. However, if `setEnabled(thingUID, true)` is called for a thing which was already enabled, it causes an additional initialization.
Since concurrent hanlder initializations may cause trouble, it is safer to prevent this. Therefore enabling a thing (including starting its thing handler) should only be performed if the thing is disabled.

Signed-off-by: Michael Reitler <github@madpage.de>

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
@@ -1201,8 +1201,8 @@ public class ThingManagerImpl
 
         if (enabled) {
             // Enable a thing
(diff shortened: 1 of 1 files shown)


cdjackson:
@mdicke2s Did you mean to close this?

mdicke2s:
Hello @cdjackson 
I had an issue with the DCO check. Therefore I created another PR and fixed the DCO issue: https://github.com/openhab/openhab-core/pull/1353

cdjackson:
Ok, no problem. I know from experience that sometimes it’s easy to click the wrong button in GH and hadn’t seen the new PR.

Thanks.

PR #1353, opened 2020-02-02
[ThingManagerImpl] Prevent multiple handler inits on thing enabling

mdicke2s:
When a thing is enabled, the `ThingManagerImpl` takes care of calling `ThingHandler#initialize()`.
That makes totally sense if the thing was disabled before. However, if `setEnabled(thingUID, true)` is called for a thing which was already enabled, it causes an additional initialization.
Since concurrent hanlder initializations may cause trouble, it is safer to prevent this. Therefore enabling a thing (including starting its thing handler) should only be performed if the thing is disabled.

Signed-off-by: Michael Reitler <michael.dicke@smail.inf.h-brs.de>

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
@@ -1201,8 +1201,8 @@ public class ThingManagerImpl
 
         if (enabled) {
             // Enable a thing
-            if (thing.getStatus().equals(ThingStatus.ONLINE)) {
-                logger.debug("Thing {} is already in the required state.", thingUID);
+            if (thing.isEnabled()) {
+                logger.debug("Thing {} is already enabled.", thingUID);
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1201,8 +1201,8 @@ public void setEnabled(ThingUID thingUID, boolean enabled) {
 
         if (enabled) {
             // Enable a thing
-            if (thing.getStatus().equals(ThingStatus.ONLINE)) {
-                logger.debug("Thing {} is already in the required state.", thingUID);
+            if (thing.isEnabled()) {
```
cweitkamp: Actually this check is a lot better than before. But it will not prevent multiple initializations for sure. There will be still some milliseconds left before the `ThingStatus` will be set to `INITIALIZING`. The initialization process itself is covered by a [mutex lock](https://github.com/mdicke2s/openhab-core/blob/c318a96f065d68e937c37c1bed178ba1718ae019/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java#L632). We maybe have to introduce a similar lock for enabling / disabling process too.

Issue #1355, opened 2020-02-05
RFC/new concept: UI component registries


ghys:
This is a proposal for a new API dedicated to generic, frontend-managed UI components, with a possible extension to sitemaps.

Full disclosure: this is presented as a much simpler alternative to https://github.com/openhab/openhab-core/issues/594 and https://github.com/eclipse/smarthome/issues/5337, out of maintenability concerns due to the very complex nature of some parts of that proposal (both conceptually and technically), and the fact that it's been stale for months/years also sheds doubts. Some of the basic concepts are nonetheless quite similar, while some others are not in this proposal, which is more generic and less server-centric. Maybe if it sees the light of day, we could eventually find a way to plug it to this concept.

I'm ready to give a shot at implementing but as per the guidelines, I open this issue in hopes to discuss the concept itself and the proposed design first.

The fundamental idea is that component composition is now widely accepted as the standard and correct way of building complex pieces of UI (examples include Vue.js, React, Svelte, the web components spec etc.), so it could be very valuable for UI developers to be able to describe and store these trees of components on the server, and have a flexible and easy way to manage them.

![image](https://user-images.githubusercontent.com/2004147/73841423-e36ad380-481a-11ea-8b5b-4de758a555a3.png)
<sub>Source: https://vuejs.org/v2/guide/#Composing-with-Components</sub>

### Components & Slots

The main concept of this proposal revolves around _components_ and _slots_ and is already central to the HABot card system ([`org.openhab.ui.habot.card`](https://github.com/openhab/openhab-webui/tree/master/bundles/org.openhab.ui.habot/src/main/java/org/openhab/ui/habot/card), see also [here](https://community.openhab.org/t/habot-walkthrough-5-n-the-card-designer/55759)), which in turn takes these notions from many modern web frameworks.

A _Component_ consists in:
1. A _Type UID_ - the client uses the type UID to render the component accordingly. It is not planned at this stage to implement hard server-side checks on these IDs, or whether they are valid, or even describe these component types. We'll see later that components reside in a namespace often targeting a specific frontend which defines them implicitely, and it's assumed that this frontend knows what it's doing - these checks might for instance be performed client-side, for instance in a designer interface;
2. A _Configuration map of key/value pairs_ which depend on the component type - this could eventually be described in a standard config description (its URI being something like `uicomponent:<namespace>:<typeUID>`) but it's not required at this stage for the same reasons as above;
[...]

PR #1356, opened 2020-02-05
UI component registries initial implementation

ghys:
This is an initial implementation of #1355.
It was simple enough to make to be proposed as
a PR already without waiting for remarks on the RFC.

<strike>The SitemapProvider for the `system:sitemap`
namespace as described in #1355 is not part of
this PR.</strike> edit: now it is.

### DIFF
--- a/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
+++ b/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
@@ -1,0 +1,306 @@
+/**
+ * Copyright (c) 2010-2020 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.ui.internal.components;
+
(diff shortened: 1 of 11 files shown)


bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
```diff
@@ -0,0 +1,306 @@
...
+    private void setWidgetPropertyFromComponentConfig(Widget widget, UIComponent component, String configParamName,
+            int feature) {
+        if (component == null || component.getConfig() == null) {
+            return;
+        }
+        Object value = component.getConfig().get(configParamName);
+        if (value == null) {
+            return;
+        }
+        if (value instanceof Double) {
```
kaikreuzer: If you do a `ConfigUtil.normalizeTypes()` on the config, you won't have to deal with Double values and other exceptions.

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
```diff
@@ -0,0 +1,306 @@
...
+                        mapping.setLabel(label);
+                        mappings.add(mapping);
+                    }
+                }
+            }
+        }
+    }
+
+    @Override
+    public void addModelChangeListener(@NonNull ModelRepositoryChangeListener listener) {
```
kaikreuzer: This class here should registered as a listener on `UIComponentRegistry`, then it could also inform any `ModelRepositoryChangeListener` that is passed in here about changes in the sitemap.

PR #1358, opened 2020-02-07
Use 'TimeZone.getDefault().toZoneId()' instead of 'ZoneId.systemDefault()'

cweitkamp:
- Use `TimeZone.getDefault().toZoneId()` instead of `ZoneId.systemDefault()`

The later is only a wrapper for the other method (see [ZoneId.systemDefault()](https://docs.oracle.com/javase/8/docs/api/java/time/ZoneId.html#systemDefault--)).


kaikreuzer:
> The later is only a wrapper for the other method

In other words it is a convenience method to write less code. Why changing it?

wborn:
I'm fine with using the convenience method as is to improve code readability unless it results in some measurable performance decrease.

cweitkamp:
Yes, less code. And - I guess - no really measurable performance improvements. I will close this PR.

PR #1359, opened 2020-02-07
Removed dependency on 'org.apache.commons.exec'

cweitkamp:
- Removed dependency on `org.apache.commons.exec`

### DIFF
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -59,12 +59,6 @@
 
     <!-- Apache Commons -->
     <dependency>
-      <groupId>org.apache.commons</groupId>
-      <artifactId>commons-exec</artifactId>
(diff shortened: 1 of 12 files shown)


kaikreuzer:
@cweitkamp As you have added the API breaking label: Will this make the exec binding/transformation in openhab2-addons fail with compile errors, if it is merged?

cweitkamp:
No, exec binding and transformation are not affected. IIRC the binding uses an own but similar implementation, the transformation uses the `ExecUtil`.

Currently the itests will break in openhab-addons because the *.bndrun files will refer to a dependency removed from OHC features.

kaikreuzer:
Ok, thanks, then it should be fine - the itests can be removed from the 3.x build.

Issue #1185, opened 2019-11-04
[Q] User-defined state/command descriptions with metadata


ghys:
AFAIK (correct me if I'm wrong) there is no way for an user to set/override the state description or command description of a managed item, for instance specifying a custom format pattern between brackets in the label like DSL items.
The problem is that sometimes, the state/command descriptions provided by the channel might have to be altered, or the item might not be linked at all and managed solely with rules.

Would it make sense to have a `StateDescriptionProvider` resp. `CommandDescriptionProvider` reading the configuration parameters in well-known metadata namespaces - say, _stateDescription_ & _commandDescription_?

It could maybe even work for DSL items as well e.g.
```
Number EnergyCost { stateDescription="???" [ pattern="%.1f €", readOnly=true ] }
Switch VideoInput { commandDescription="???" [ options="HDMI1,HDMI2" ] }
```

openhab-5iver:
> AFAIK (correct me if I'm wrong) there is no way for an user to set/override the state description or command description of a managed item

I played around with this for a while using scriupted automation and the only way I saw to do it was to create a Thing with a Channel that the Item was linked to in order to manipulate the Item's StateDescription. I was planning to make a helper library for this, but it was getting so involved that I stopped, since OHC would be the right place to do it.

This functionality would be really nice to have. Maybe something for the OH 3.0 backlog, @cweitkamp?

J-N-K:
Solved in #1362

PR #1362, opened 2020-02-10
Metadata-based state & command description providers

ghys:
Implements #1185.

These providers will look into item metadata, which
can be managed by UIs with the API, to set or override
the item's state description (pattern, options, read
only...) or command description.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/MetadataCommandDescriptionProvider.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/MetadataCommandDescriptionProvider.java
@@ -49,7 +49,7 @@ public class MetadataCommandDescriptionProvider implements CommandDescriptionPro
     private MetadataRegistry metadataRegistry;
 
     @Activate
-    protected MetadataCommandDescriptionProvider(final @Reference MetadataRegistry metadataRegistry,
+    public MetadataCommandDescriptionProvider(final @Reference MetadataRegistry metadataRegistry,
             Map<String, Object> properties) {
         this.metadataRegistry = metadataRegistry;
     }
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/MetadataCommandDescriptionProvider.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/MetadataCommandDescriptionProvider.java
@@ -1,0 +1,83 @@
+/**
+ * Copyright (c) 2010-2020 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.internal.items;
+
+import java.util.Locale;
+import java.util.stream.Stream;
+
+import org.eclipse.jdt.annotation.NonNull;
+import org.eclipse.jdt.annotation.Nullable;
+import org.openhab.core.internal.types.CommandDescriptionImpl;
+import org.openhab.core.items.Metadata;
+import org.openhab.core.items.MetadataKey;
+import org.openhab.core.items.MetadataRegistry;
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/MetadataCommandDescriptionProvider.java
```diff
@@ -0,0 +1,83 @@
...
+            } catch (Exception e) {
+                logger.error("Unable to parse the commandDescription from metadata for item {}", itemName, e);
+                return null;
+            }
+        }
+
+        return null;
+    }
+
+    @Reference
```
kaikreuzer: You can add the `@Reference` directly in line 45 and get rid off the getter and setter methods.
Or pass it in via constructor as you do in `MetadataStateDescriptionFragmentProvider`.

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/MetadataStateDescriptionFragmentProvider.java
```diff
@@ -0,0 +1,151 @@
...
+                                return (o.contains("="))
+                                        ? new StateOption(o.split("=")[0].trim(), o.split("=")[1].trim())
+                                        : new StateOption(o.trim(), null);
+                            }).collect(Collectors.toList());
+                    builder.withOptions(stateOptions);
+                }
+
+                return builder.build();
+            } catch (Exception e) {
+                logger.error("Unable to parse the stateDescription from metadata for item {}", itemName, e);
```
kaikreuzer: This should imho be a WARN for the user, telling him that the value is ignored. No stacktrace must be added here as we are not talking about a bug in the software.

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/MetadataCommandDescriptionProvider.java
```diff
@@ -0,0 +1,83 @@
...
+                                    new CommandOption(o.split("=")[0].trim(), o.split("=")[1].trim()));
+                        } else {
+                            commandDescription.addCommandOption(new CommandOption(o.trim(), null));
+                        }
+                    });
+                }
+
+                return commandDescription;
+            } catch (Exception e) {
+                logger.error("Unable to parse the commandDescription from metadata for item {}", itemName, e);
```
kaikreuzer: This should imho be a WARN for the user, telling him that the value is ignored. No stacktrace must be added here as we are not talking about a bug in the software.

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/MetadataStateDescriptionFragmentProvider.java
```diff
@@ -83,10 +83,10 @@ public MetadataStateDescriptionFragmentProvider(final @Reference MetadataRegistr
                     builder.withMinimum(getBigDecimal(metadata.getConfiguration().get("min")));
                 }
                 if (metadata.getConfiguration().containsKey("max")) {
-                    builder.withMaximum(getBigDecimal(metadata.getConfiguration().get("min")));
+                    builder.withMaximum(getBigDecimal(metadata.getConfiguration().get("max")));
```
kaikreuzer: haha, it seems the unit tests have already helped ;-)

kaikreuzer:
Now you only have to convince Jenkins:
```
07:44:24 [ERROR] /home/jenkins/jenkins-agent1/workspace/PR-openHAB-Core/bundles/org.openhab.core/pom.xml [0:0]: Constructors must be public access.
07:44:24 [ERROR] /home/jenkins/jenkins-agent1/workspace/PR-openHAB-Core/bundles/org.openhab.core/pom.xml [0:0]: In component org.openhab.core.internal.items.MetadataCommandDescriptionProvider, @Reference can only be used for parameters on the constructor annotated @Activate
```

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/MetadataStateDescriptionFragmentProvider.java
ghys: My thoughts exactly but I wanted to keep that under the radar ;)

Issue #1363, opened 2020-02-12
RFC: flexible API/SSE endpoint for dynamic item state updates


ghys:
This is a suggestion of a new API to help UIs solve a common and specific problem with the current `/rest/events` SSE endpoint exposing the "raw" event bus: retrieving targeted item state updates. It'll help with the concern expressed here: https://github.com/openhab/openhab-core/issues/1355#issuecomment-584913593

The issues this proposal tries to address are, among others:
- `/rest/events` only offers a topic filter which is not flexible enough to retrieve state updates for a specific set of items only;
- SSE connections are initiated by a GET request so parameters are limited to URL query strings, which limits large sets of items to track and advanced querying;
- events are relatively costly to parse, while only (item, new state) tuples are usually enough;
- if the state has to be transformed for display, the new state value in the `ItemStateChangedEvent` doesn't have the transformation applied so the client has to call `/rest/items/...` to retrieve the transformed state;
- whenever the needs change, the connection has to be closed and opened again with new parameters

### How It Works

The proposed extension consists in a new SSE endpoint and a regular HTTP POST endpoint, and works as follows:

1. A client opens a EventSource connection to a new SSE endpoint `/rest/events/states`.
```js
const eventSource = new EventSource('/rest/events/states')
```

2. The server assigns an unique _connectionId_ to the new connection (stored in the `EventOutput` instance) then immediately sends a first event of type `ready` with the id `0` and the connectionId as data. No new events are sent for now (1):
```
> curl http://localhost:8080/rest/events/states
event: ready
id: 0
data: d7f63062-3a5a-453e-b824-3e462587a68d

(...connection stays open...)
```

3. The client catches this "ready" event and retains the connectionId:
```js
eventSource.addEventListener('ready', (e) => {
  console.log('the connectionId is: ' + e.data)
  // process and continue
})
```

4. The client makes a regular HTTP POST request to `/rest/events/states/{connectionId}` specifying in the body which item(s) it wants to track:
```js
// Assumed Content-type: application/json
fetch('/rest/events/states/d7f63062-3a5a-453e-b824-3e462587a68d',
   { method: 'POST', body: JSON.stringify(["DemoSwitch", "DemoSwitchGroup"]) })
.then(...)
```
Here a simple JSON array of item names is expected in the body but the request details can be discussed further. The client usually has all the information on the items it needs to track (likely as a result of a prior call to `/rest/items`), so a simple array of names could be enough, but we could for instance expand groups to include their members etc.

[...]

PR #1364, opened 2020-02-12
Simple dynamic item state updates SSE endpoint & API.

ghys:
Implements #1363.


bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/SseStatesResource.java
```diff
@@ -0,0 +1,214 @@
...
+    public void broadcastEvent(final ItemStateChangedEvent stateChangeEvent) {
+        executorService.execute(new Runnable() {
+            @Override
+            public void run() {
+                OutboundEvent.Builder builder = new OutboundEvent.Builder();
+                Map<String, String> payload = new HashMap<>();
+                Item item = itemRegistry.get(stateChangeEvent.getItemName());
+                if (item != null) {
+                    String transformedState = considerTransformation(item.getState().toString(), item, null);
+                    payload.put(item.getName(), transformedState);
```
kaikreuzer: As mentioned in the issue, I am pretty sure that you will additionally require the raw state as well in the UI, e.g. for being able to request the correct item for the state, apply ui-based logic on visibility or colors, etc. So the current item->state map might be too simplistic after all.
Also, we might add further metadata to the `ItemStateChangedEvent` in future, such as the reason why it changed (manual command, rule triggered, or whatever). Serializing the whole content of ItemStateChangedEvent would somehow feel more future-proof, wdyt?

ghys: > I am pretty sure that you will additionally require the raw state as well in the UI

True.

> Serializing the whole content of ItemStateChangedEvent would somehow feel more future-proof

The idea with this is not actually to expose events from the event bus, but only to make it as easy as possible for clients to keep track of the state of a predetermined set of items. So: 1) you open the SSE connection, 2) you send the list of items whose state you want to track, 3) you receive the current states of all your tracking list over the SSE connection, 4) you also receive any individual updates until you close the connection or change the tracking list.

The benefit with item-state maps is that for 3) ___and___ 4) the client-side code is pretty simple and actually the same in both cases: basically, just merge the map you received with the one you currently have. 3) is also useful because with this feature you might not even need to call `/rest/items` to initialize your client-side map if all you need are the current states. However, 3) doesn't correspond to any event on the event bus. That's why I wanted to abstract the event bus away from this API. Maybe the state could be an 2-element array or simple object (`{"state": "CLOSED", "transformedState": "zu"}`) in case there's a transformation involved.

Moreover it was asked a few years ago (IIRC by me) over in ESH to include the transformed state in `ItemStateChangedEvent`s but it wasn't accepted, because the transformation is possibly costly to compute (e.g. exec or JS) and blocking so it's not a good idea to do it on every change if it's not useful. Therefore, there's indeed a problem with this code here - it should first determine whether the item is being tracked or not.

kaikreuzer: If we'd drop transformation support and thus would only have patterns, which you can render on client side, we could merely pass the raw state in your item state map. Would that solve all our issues?

ghys: > If we'd drop transformation support and thus would only have patterns, which you can render on client side, we could merely pass the raw state in your item state map. Would that solve all our issues?

Turns out, even formatting the state using a pattern is no trivial problem (dates, quantity types unit conversions...), and it requires the client to know about the state description, which might not be always the case... so I figured making the server do the heavy lifting was the best option after all, and introduced in https://github.com/openhab/openhab-core/pull/1364/commits/1d11fa4e8076410e9dde8e3eb3243b944c041d09 a "displayState" in the map value.
So with items like this:
```
DateTime DemoDateTime { stateDescription="N/A" [ pattern="%1$tA, %1$td.%1$tm.%1$tY" ] }
Number:Temperature DemoTemperature "Info [%.5f %unit%]"
Number DemoFormat "Info [%.5f kWh]"
Contact DemoOptions "Options" { stateDescription="N/A" [ options="OPEN=unlocked,CLOSED=locked" ] }
```
It gives (line breaks added for clarity):
```
data: {
"DemoFormat":{"state":"500","displayState":"500.00000 kWh"},
"DemoContact":{"state":"OPEN"},
"DemoOptions":{"state":"CLOSED","displayState":"locked"},
"MetadataString":{"state":"1","displayState":"HDMI1"},
"DemoTemperature":{"state":"23.00 °F","displayState":"-5.00000 °C"},
"DemoDateTime":{"state":"2020-02-18T11:37:29.782672+0100","displayState":"Tuesday, 18.02.2020"}
}
```
So the UI can happily use the displayState for display purposes if it exists (or fall back to the actual state) and also have the raw version at hand.

kaikreuzer: Sounds sensible then, thanks.

bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/SseStatesResource.java
```diff
@@ -0,0 +1,170 @@
...
+
+    private static final String X_ACCEL_BUFFERING_HEADER = "X-Accel-Buffering";
+
+    private final ItemStateChangesSseBroadcaster broadcaster;
+
+    private final ExecutorService executorService;
+
+    private ItemRegistry itemRegistry;
+
+    @Reference
```
kaikreuzer: again, this annotation can be put on the field and the setter and getter removed.

bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/SseStatesResource.java
```diff
@@ -0,0 +1,170 @@
...
+        return Response.ok().build();
+    }
+
+    /**
+     * Broadcasts a state event to all currently listening clients, after transforming it to a simple map.
+     *
+     * @param stateChangeEvent the {@link ItemStateChangedEvent} containing the new state
+     */
+    public void broadcastEvent(final ItemStateChangedEvent stateChangeEvent) {
+        executorService.execute(new Runnable() {
```
kaikreuzer: How about using `() -> {}` syntax instead of the Runnable?

bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/internal/ItemStateChangesSseBroadcaster.java
```diff
@@ -0,0 +1,234 @@
...
+                            // This also handles IllegalFormatConversionException, which is a subclass of
+                            // IllegalArgument.
+                            try {
+                                displayState = state.format(pattern);
+                            } catch (IllegalArgumentException e) {
+                                logger.warn("Exception while formatting value '{}' of item {} with format '{}': {}",
+                                        state, item.getName(), pattern, e.getMessage());
+                                displayState = new String("Err");
+                            }
+
```
kaikreuzer: remove empty line

bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/SseStatesResource.java
```diff
@@ -0,0 +1,170 @@
...
+    private ItemRegistry itemRegistry;
+
+    @Reference
+    protected void setItemRegistry(ItemRegistry itemRegistry) {
+        this.itemRegistry = itemRegistry;
+        this.broadcaster.setItemRegistry(itemRegistry);
+    }
+
+    protected void unsetItemRegistry(ItemRegistry itemRegistry) {
+        this.itemRegistry = itemRegistry;
```
kaikreuzer: for symmetry reasons it would be good to unset the item registry in the broadcaster here as well.

bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/SseStatesResource.java
```diff
@@ -0,0 +1,170 @@
...
+    private UriInfo uriInfo;
+
+    @Context
+    private HttpServletResponse response;
+
+    @Context
+    private HttpServletRequest request;
+
+    public SseStatesResource() {
+        this.executorService = Executors.newSingleThreadExecutor();
```
kaikreuzer: Better to have named threads, so that we can identify them when debugging:
```suggestion
        this.executorService = Executors.newSingleThreadExecutor(new NamedThreadFactory("SseStatesResource"));
```

kaikreuzer: The executor creation should be move to an activate method and it should be terminated in deactivated again to ensure we do not have any zombie threads when the bundle is shut down.

ghys: I copied and adapted the code from the original SSE resource:
https://github.com/openhab/openhab-core/blob/730a8e1e6104d966584b823819bb730f2695dde8/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/SseResource.java#L77-L82

Do you want me to make the same changes there as well? Or maybe merge the two resources (even if it's a different broadcaster though).

bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/SseStatesResource.java
ghys: Same comment as above
https://github.com/openhab/openhab-core/blob/730a8e1e6104d966584b823819bb730f2695dde8/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/SseResource.java#L126-L129

ghys:
@kaikreuzer I have finally removed the SseStatesResource and merged it into the main SseResource to avoid duplicating code and share the executor service. I verified against regressions, all endpoints work.

PR #1367, opened 2020-02-16
[automation] Corrected issue where scripts were not starting when OH restarted

### LINKED ISSUE #841: [automation] Scripts do not load at startup on Windows
Observed on S1597, but very likely occurs on earlier post ESH reintegration snapshots too. With OH running on a Windows machine, scripts do not load at startup. After saving a script, the script is loaded. I've reproduced this in a test environment while troubleshooting the issue someone had [reported in the forum](https://community.openhab.org/t/jsr223-jython-simplified-rule-definition-similar-to-rules-dsl-and-universal-decorator/53252/172?u=5iver). If the issue does not occur on a build prior to one that includes #724, then the changes to ScriptFileWatcher could be the cause, but I don't see that PR causing an issue for Windows. I plan to investigate further.

openhab-5iver:
Fixes #841. I also added a few file extensions to the list that are not loaded as scripts.I've tested on Windows 10 and Fedora 31.

### DIFF
--- a/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/loader/ScriptFileWatcher.java
+++ b/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/loader/ScriptFileWatcher.java
@@ -19,6 +19,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.net.MalformedURLException;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -155,7 +156,7 @@ public class ScriptFileWatcher extends AbstractWatchService {
     }
 
     private synchronized void importFile(URL url) {
-        String fileName = getFileName(url);
+        String fileName = url.getFile();
         if (loaded.contains(url)) {
             this.removeFile(url); // if already loaded, remove first
         }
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/loader/ScriptFileWatcher.java
```diff
@@ -238,7 +229,9 @@ private String getScriptType(URL url) {
         String fileExtension = fileName.substring(idx + 1);
 
         // ignore known file extensions for "temp" files
-        if (fileExtension.equals("txt") || fileExtension.endsWith("~") || fileExtension.endsWith("swp")) {
+        if (fileExtension.equals("txt") || fileExtension.equals("old") || fileExtension.equals("example")
```
cweitkamp: Instead of blacklisting wouldn't it be safer to define a whitelist of allowed file extensions?

Otherwise you should put string litterals first in comparisons or may introduce a `Set` of Strings and check if the set contains your blacklisted extension.

openhab-5iver: It would be very difficult to use a whitelist, since any compatible ScriptEngineFactory could be used and there's no way to tell what file extensions might be used.

I'll remove this for now and add a separate PR with a better way to blacklist. I really shouldn't have included this here anyhow, but I've wanted to exclude these extensions for a while.

openhab-5iver: I have reverted this change and will submit another PR for it.

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/loader/ScriptFileWatcher.java
```diff
@@ -238,7 +229,7 @@ private String getScriptType(URL url) {
         String fileExtension = fileName.substring(idx + 1);
 
         // ignore known file extensions for "temp" files
-        if (fileExtension.equals("txt") || fileExtension.endsWith("~") || fileExtension.endsWith("swp")) {
+         if (fileExtension.equals("txt") || fileExtension.endsWith("~") || fileExtension.endsWith("swp")) {
```
cweitkamp: Can you remove the whitespace?

openhab-5iver: Oops.. sorry about that... done!

PR #1370, opened 2020-02-17
[automation] Improve file extension exclusion

openhab-5iver:
This change improves the way file extensions are excluded from ScriptFileWatcher. Discussed [here](https://github.com/openhab/openhab-core/pull/1367#discussion_r379914147).


wborn:
How about also excluding .bak and .tmp files?

openhab-5iver:
I added tmp and bak. In rethinking this, a whitelist could be generated by gathering up the extensions supported by the installed ScriptEngineFactories. This would not be a trivial effort, so I'll put it at the bottom of my backlog.

PR #1387, opened 2020-03-06
Allow persistence services to define default strategies

kaikreuzer:
Also refactored Simple... classes to Persistence.... classes (that's actually 95% of the changes).

### DIFF
--- a/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/PersistenceFilter.java
+++ b/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/PersistenceFilter.java
@@ -10,18 +10,13 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  */
-package org.openhab.core.persistence.config;
+package org.openhab.core.persistence;
 
 /**
- * This class represents the configuration that stand for "using all items".
+ *
  *
  * @author Markus Rathgeb - Initial contribution
  */
-public class SimpleAllConfig extends SimpleConfig {
-
-    @Override
-    public String toString() {
-        return String.format("%s []", getClass().getSimpleName());
-    }
+public class PersistenceFilter {
 
 }
--- a/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/PersistenceFilter.java
+++ b/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/PersistenceFilter.java
@@ -12,11 +12,14 @@
  */
 package org.openhab.core.persistence;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
 /**
  *
  *
  * @author Markus Rathgeb - Initial contribution
  */
+@NonNullByDefault
 public class PersistenceFilter {
 
 }
--- a/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/PersistenceService.java
+++ b/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/PersistenceService.java
@@ -12,11 +12,13 @@
  */
 package org.openhab.core.persistence;
 
+import java.util.List;
 import java.util.Locale;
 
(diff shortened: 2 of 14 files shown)


bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/PersistenceService.java
```diff
@@ -70,4 +72,11 @@
      * @param alias the alias under which the item should be persisted.
      */
     void store(Item item, String alias);
+
+    /**
+     * Provides default persistence strategies that are used for all items if no user defined configuration is found.
+     *
+     * @return The default persistence strategies
+     */
+    List<PersistenceStrategy> getDefaultStrategies();
```
cweitkamp: I know that this PR is APIBreaking anyways but you also can add the default method implementation here:
```suggestion
    default List<PersistenceStrategy> getDefaultStrategies() {
        return Collections.emptyList();
    }
```
Be aware of an eventually missing import.

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/PersistenceFilter.java
```diff
@@ -17,6 +17,6 @@
  *
  * @author Markus Rathgeb - Initial contribution
  */
-public class SimpleFilter {
+public class PersistenceFilter {
```
cweitkamp: Should we add nullness annotations and other touched interfaces / classes?

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/internal/PersistenceManagerImpl.java
```diff
@@ -256,46 +261,50 @@ private boolean appliesToItem(SimpleItemConfiguration config, Item item) {
...
+                                                                .format(historicItem.getTimestamp()),
+                                                        item.getName(), historicItem.getState());
+                                            }
+                                            return;
                                         }
-                                        return;
                                     }
+                                } else if (service != null) {
+                                    logger.warn(
+                                            "Failed to restore item states as persistence service '{}' can not be queried.",
```
cweitkamp: Minor spelling improvement:

```suggestion
                                            "Failed to restore item states as persistence service '{}' cannot be queried.",
```

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/internal/PersistenceManagerImpl.java
```diff
@@ -107,6 +108,9 @@ protected void deactivate() {
     protected void addPersistenceService(PersistenceService persistenceService) {
         logger.debug("Initializing {} persistence service.", persistenceService.getId());
         persistenceServices.put(persistenceService.getId(), persistenceService);
+        if (!persistenceServiceConfigs.containsKey(persistenceService.getId())) {
+            persistenceServiceConfigs.put(persistenceService.getId(), getDefaultConfig(persistenceService));
+        }
```
cweitkamp: One liner?

```suggestion
        persistenceServiceConfigs.putIfAbsent(persistenceService.getId(), getDefaultConfig(persistenceService));
```

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/PersistenceService.java
kaikreuzer: I thought about this, but I decided against it. Problem with the default methods is that no method skeleton is created in implementing classes, so that developers are not aware that they should actually think about implementing the method.

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/PersistenceFilter.java
kaikreuzer: Added.

PR #1395, opened 2020-03-22
utility for removing orphan links

J-N-K:
Do you think that something like this is a useful addition?

Edit: Merge conflicts because it's based on an old 2.5 branch to make it work in my production setup.


mherwege:
Yes, makes a lot of sense.

kaikreuzer:
Yes, I also think that this is useful!

cdjackson:
As a matter of interest, why not just avoid the problem in the first place by removing links when removing items rather than provide a utility to remove links later?

https://github.com/eclipse/smarthome/issues/1018

J-N-K:
Who should decide if an item is permanently removed? Imagine I have all items in one big item file. I then decide that I would like to split that. I open an editor, cut ten items out of that file, save and then paste the same ten items on another file. Why should all channel links from these items be removed?

And that is only half of the problem: we need to delete and re-add files things when the thing structure changes. Why should I need to re-link all channels that did not change?

PR #1399, opened 2020-03-31
[thing] Dynamic state/command provider should not return original description 

cweitkamp:
- Dynamic state/command provider should not return original description
- Added integration test for `CommandDescriptionProvider`

See https://community.openhab.org/t/dynamic-state-options-not-showing-up/95625

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ChannelStateDescriptionProvider.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ChannelStateDescriptionProvider.java
@@ -132,8 +132,14 @@ public class ChannelStateDescriptionProvider implements StateDescriptionFragment
             @Nullable StateDescription originalStateDescription, @Nullable Locale locale) {
         for (DynamicStateDescriptionProvider provider : dynamicStateDescriptionProviders) {
             StateDescription stateDescription = provider.getStateDescription(channel, originalStateDescription, locale);
-            if (stateDescription != null && !stateDescription.equals(originalStateDescription)) {
-                return stateDescription;
+            if (stateDescription != null) {
+                if (stateDescription.equals(originalStateDescription)) {
+                    logger.error(
+                            "Dynamic state description matches original state description. DynamicStateDescriptionProvider implementations must never return the original state description. {} has to be fixed.",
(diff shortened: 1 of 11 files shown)


bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ChannelStateDescriptionProvider.java
```diff
@@ -132,7 +132,7 @@ public Integer getRank() {
             @Nullable StateDescription originalStateDescription, @Nullable Locale locale) {
         for (DynamicStateDescriptionProvider provider : dynamicStateDescriptionProviders) {
             StateDescription stateDescription = provider.getStateDescription(channel, originalStateDescription, locale);
-            if (stateDescription != null) {
+            if (stateDescription != null && !stateDescription.equals(originalStateDescription)) {
```
cweitkamp: Question is if we should just ignore these wrong implementations in bindings or if we should throw an exception here and force the developer to fix it.

mhilbush: Being completely silent about it doesn't feel right. Would the exception be thrown every time the state options are accessed? If so, that could result in a lot of exceptions. Maybe just log it instead?

cweitkamp: I added a ERROR log message.

wborn: Can you also update the JavaDocs of these providers for the fact that if the result is not null it should not be the original object?

cweitkamp: I already did (see e.g. [here](https://github.com/openhab/openhab-core/pull/1399/files#diff-af4592feefd1fdeb2cb4cfcc431d75c0R38)).

wborn: Great! Also nice to see some unit tests for this. :+1:

PR #1404, opened 2020-04-04
Replace unit placeholder in patterns if updated by DecimalType states

### LINKED ISSUE #986: Units Of Measurement (Illuminance) format error
I build an item like this:
`Number:Illuminance  GF_Hallway_Brightness      		"Helligkeit [%d lx]"    					<motion>       		 (GF_Hallway, gMotion)			 ["MotionDetector"]                 {channel="<working_entry>"}`
on my sitemap. This works like a charm. (Btw.: The whole openHAB works like that, I'm deeply impressed!)

Now I wanted to work it out using "Units Of Measurement" like this:
`Number:Illuminance  GF_Hallway_Brightness      		"Helligkeit [%d %unit%]"    					<motion>       		 (GF_Hallway, gMotion)			 ["MotionDetector"]                 {channel="<working_entry>"}`
...which leads to the following message(s) in the log:
`[WARN ] [ui.internal.items.ItemUIRegistryImpl] - Exception while formatting value '11856.0' of item GF_Hallway_Brightness with format '%.0f %unit%': Conversion = 'u'`
This items value cannot be display anymore, instead I get an "Err" string on screen. I'd expect the same output as before because Illuminance's symbol should evaluate to 'lx'

Why does it show a float format string with no position after decimal point? It should be `%d` instead, shouldn't it?

Version: openHAB 2.4.0-1 (Release Build)

kind regards
Christian

[Edit] Bold styling (**) within code segments removed.

cweitkamp:
- Replace unit placeholder in patterns if updated by `DecimalType` states

Fixes #986

### DIFF
--- a/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java
+++ b/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java
@@ -400,7 +400,7 @@ public class ItemUIRegistryImpl implements ItemUIRegistry {
                         // for DecimalTypes we don't want to risk format exceptions, if pattern contains unit
                         // placeholder
                         if (formatPattern.contains(UnitUtils.UNIT_PLACEHOLDER)) {
-                            formatPattern = formatPattern.replaceAll(UnitUtils.UNIT_PLACEHOLDER, "");
+                            formatPattern = formatPattern.replaceAll(UnitUtils.UNIT_PLACEHOLDER, "").stripTrailing();
                         }
                     } else if (state instanceof QuantityType) {
                         QuantityType<?> quantityState = (QuantityType<?>) state;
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java
```diff
@@ -396,7 +396,13 @@ private Switch createPlayerButtons() {
                         }
                     }
 
-                    if (state instanceof QuantityType) {
+                    if (state instanceof DecimalType) {
+                        // for DecimalTypes we don't want to risk format exceptions, if pattern contains unit
+                        // placeholder
+                        if (formatPattern.contains(UnitUtils.UNIT_PLACEHOLDER)) {
+                            formatPattern = formatPattern.replaceAll(UnitUtils.UNIT_PLACEHOLDER, "");
```
kaikreuzer, line 403: Should we maybe also trim it, so that the pattern does not have a trailing space?

cweitkamp: Yes, I thought about that too. Will add it.

PR #1407, opened 2020-04-05
Add and fix more null annotations

wborn:
* Add more `@NonNullByDefault` and `@Nullable` annotations
* Remove unnecessary `@NonNull` annotations
* Fix a few other trivial SAT issues

### DIFF
--- a/bundles/org.openhab.core.config.discovery.mdns/src/main/java/org/openhab/core/config/discovery/mdns/internal/MDNSDiscoveryService.java
+++ b/bundles/org.openhab.core.config.discovery.mdns/src/main/java/org/openhab/core/config/discovery/mdns/internal/MDNSDiscoveryService.java
@@ -32,7 +32,9 @@ import org.openhab.core.config.discovery.mdns.MDNSDiscoveryParticipant;
 import org.openhab.core.io.transport.mdns.MDNSClient;
 import org.openhab.core.thing.ThingTypeUID;
 import org.openhab.core.thing.ThingUID;
+import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Deactivate;
 import org.osgi.service.component.annotations.Modified;
 import org.osgi.service.component.annotations.Reference;
 import org.osgi.service.component.annotations.ReferenceCardinality;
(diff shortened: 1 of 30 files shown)


itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/internal/ThingRegistryOSGiTest.java
```diff
@@ -237,7 +237,11 @@ public boolean supportsThingType(@NonNull ThingTypeUID thingTypeUID) {
         waitForAssert(() -> {
             assertTrue(thingResultWrapper.get() != null);
         });
-        assertThat(thing, is(thingResultWrapper.get()));
+
+        assertThat(this, is(notNullValue()));
```
cweitkamp, line 241: Can you shortly explain this assertion?

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
```diff
@@ -261,8 +273,8 @@ protected Widget buildWidget(UIComponent component) {
         return widget;
     }
 
-    private void setWidgetPropertyFromComponentConfig(Widget widget, UIComponent component, String configParamName,
-            int feature) {
+    private void setWidgetPropertyFromComponentConfig(Widget widget, @Nullable UIComponent component,
```
cweitkamp, line 276: Something beyond scope of this PR but passing a `@Nullable UIComponent component` here seems to be wrong. Could be captured by calling methods.

bundles/org.openhab.core.config.discovery.mdns/src/main/java/org/openhab/core/config/discovery/mdns/internal/MDNSDiscoveryService.java
```diff
@@ -59,7 +60,7 @@
 
     private final Set<MDNSDiscoveryParticipant> participants = new CopyOnWriteArraySet<>();
 
-    private MDNSClient mdnsClient;
+    private @NonNullByDefault({}) MDNSClient mdnsClient;
```
cweitkamp, line 63: Can we use constructor injection for this reference?

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
wborn: I was still unsure, after a short exploration of the call tree, if it would actually happen because of the null check in this method. So I opted for the conservative route of adding the annotation.

itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/internal/ThingRegistryOSGiTest.java
wborn: I'll fix that because it should have checked the `thing` instead of `this` so the compiler will know it is null safe but will still fail if null . :+1:

bundles/org.openhab.core.config.discovery.mdns/src/main/java/org/openhab/core/config/discovery/mdns/internal/MDNSDiscoveryService.java
wborn: I've pushed some changes that I think will work to accomplish this.

PR #1410, opened 2020-04-06
Removed 'advanced' attribute from 'ChannelGroupType'

### LINKED ISSUE #1402: Channel group 'advanced' property not forwarded to UIs
When creating channel group types, ether through XML files or in the binding, you can choose to mark is as advanced, e.g:
```xml
<channel-group-type id="default" advanced="true">
```
However, this property is not forwarded to UIs via the REST API where the thing-type model looks like:

```json
  "channelGroups": [
    {
      "id": "string",
      "description": "string",
      "label": "string",
      "channels": [
      ]
    }
  ],
```
I don't know if any UIs currently uses the property, but I think it should be available for any new admin UI in OH3.

Is it just in the `ChannelGroupDefinitionDTO` it needs to be added? (I noticed there aren't any `ChannelGroupTypeDTO` like there is for the other types, should it be added as well?) Or are there any other parts where it needs to be added? If I can get some pointers I might come up with a PR myself, unless there are many other parts that gets affected.

Edit: Found it's use in org.openhab.core.io.rest.core.internal.thing.ThingTypeResource, which would have to be adapted as well.

cweitkamp:
- Removed 'advanced' attribute from `ChannelGroupType`

Resolves #1402 

@kaikreuzer What would be an appropriate versioning for the XSD file? 3.0.0? matching OHC version? If we change it we have to migrate all XML files in this repo.

### DIFF
--- a/bundles/org.openhab.core.thing.xml/src/main/java/org/openhab/core/thing/xml/internal/ChannelGroupTypeConverter.java
+++ b/bundles/org.openhab.core.thing.xml/src/main/java/org/openhab/core/thing/xml/internal/ChannelGroupTypeConverter.java
@@ -34,24 +34,14 @@ import com.thoughtworks.xstream.io.HierarchicalStreamReader;
  *
  * @author Michael Grammling - Initial contribution
  * @author Chris Jackson - Modified to support channel properties
+ * @author Christoph Weitkamp - Removed "advanced" attribute
  */
 public class ChannelGroupTypeConverter extends AbstractDescriptionTypeConverter<ChannelGroupTypeXmlResult> {
 
     public ChannelGroupTypeConverter() {
         super(ChannelGroupTypeXmlResult.class, "channel-group-type");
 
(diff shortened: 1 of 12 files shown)


kaikreuzer:
> matching OHC version?

No, this would mean that we would have to change the version all the time...
I would say we update it according to semver, so the next would simply be 2.0.0.

> If we change it we have to migrate all XML files in this repo.

I'm not too keen on it either. Alternatively, we could check if any add-on uses this attribute and remove it manually, so that we can do the change without breaking anything.

cweitkamp:
As you merged it I conclude we do not change the versioning but remove the `advanced` attribute in all add-ons.

kaikreuzer:
> I conclude we do not change the versioning but remove the advanced attribute in all add-ons.

That's what I felt is probably easiest - and I assumed this was your preference as well. Let me know if I was too quick here.

cweitkamp:
I agree on that. Prepared a MR for the add-ons. I did not find much of them.

https://github.com/openhab/openhab-addons/pull/7325

Issue #888, opened 2019-06-20
Nullness EEA


maggu2810:
<a alt="bounty" href="https://www.bountysource.com/issues/75826293"><img src="https://api.bountysource.com/badge/issue?issue_id=75826293"></a>

I would like to resume the discussion about EEA for the nullness information of third party libraries (and the JRE files).

As we moved to pure Maven projects the situation changed a little bit.

First we need something that provides the EEAs we would like to consume (the approach here will not work for openHAB Core only but for all projects we would like to use it).

This is in my example a simple Maven project that contains the EEAs in the resources, so they are part of the created JAR file.

For the normal Maven build (command line) of our openHAB projects it is enough to change a simple setting of the used compiler that instruct the compiler to use EEAs that are part of the current classpath. Additionally we need to add a the EEAs to the classpath that is fulfilled by adding an artifact that contains the EEAs to the dependencies.
See https://github.com/maggu2810/openhab-core/commit/6e26af1c8c201e1f9bd26fe7fc94af3371fbfbaa#diff-600376dffeb79835ede4a0b285078036R303 and ignore the additional property (not part of that integration).

For the command line all is done now.

So, what needs to be done to get the support in the IDE?
We need to inform the IDE about the EEAs we want to use.
I don't know about a way to state "use the classpath", but we can refer to a project that is part of the workspace. This would be the preffered solution for me.
If the project containing the EEAs will be part of every workspace the remaining task will be easy. Would this be acceptable if we add it to the setup file and e.g. place the JAR containing the EEAs to openhab-distro or another separate repo that will be included for every IDE setup?
To add the respective annotationpath to the classpath for Eclipse, we can use the property above (in the POM file) and a simple Maven plugin of LastNPE (it will use the value of the specific property and add it to the projects settings).

I have already done all the necessary changed for openHAB Core.

WDYT?

maggu2810:
As there seems to be no interest of the core maintainers I close this one.

wborn:
I'd love to see EEAs in place to get rid of all those false positive warnings for `Map`s etc. I just didn't have time to further investigate this subject so I could add a meaningful comment here. Personally I'd rather first see the IDE/M2 woes resolved.

kaikreuzer:
Same for me. Interest is definitely there, but it's not on my top priority list atm. I wouldn't mind to keep this issue open nonetheless as it is a valid discussion.

wborn:
So let's reopen the issue because we do have interest in using EEA.

[...]

PR #1412, opened 2020-04-07
Add null annotations to providers and ThingManager

wborn:
It's almost complete so I'll already create a WIP PR for visibility so others don't start annotating these classes.

There are still failing tests and I want to give it a test to see that it really works.

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
@@ -361,7 +361,7 @@ public class ThingManagerImpl
             final @Reference ReadyService readyService, //
             final @Reference SafeCaller safeCaller, //
             final @Reference StorageService storageService, //
-            final @Reference ThingRegistryImpl thingRegistry,
+            final @Reference ThingRegistry thingRegistry,
             final @Reference ThingStatusInfoI18nLocalizationService thingStatusInfoI18nLocalizationService,
             final @Reference ThingTypeRegistry thingTypeRegistry) {
         this.bundleResolver = bundleResolver;
@@ -374,12 +374,12 @@ public class ThingManagerImpl
         this.itemChannelLinkRegistry = itemChannelLinkRegistry;
         this.readyService = readyService;
         this.safeCaller = safeCaller;
-        this.thingRegistry = thingRegistry;
+        this.thingRegistry = (ThingRegistryImpl) thingRegistry;
         this.thingStatusInfoI18nLocalizationService = thingStatusInfoI18nLocalizationService;
         this.thingTypeRegistry = thingTypeRegistry;
 
         readyService.registerTracker(this, new ReadyMarkerFilter().withType(XML_THING_TYPE));
-        thingRegistry.addThingTracker(this);
+        this.thingRegistry.addThingTracker(this);
         storage = storageService.getStorage(THING_STATUS_STORAGE_NAME, this.getClass().getClassLoader());
     }
 
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
@@ -34,6 +34,7 @@ import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Function;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.common.SafeCaller;
 import org.openhab.core.common.ThreadPoolManager;
@@ -120,6 +121,7 @@ import org.slf4j.LoggerFactory;
  * @author Christoph Weitkamp - Added preconfigured ChannelGroupBuilder
  * @author Yordan Zhelev - Added thing disabling mechanism
  */
+@NonNullByDefault
 @Component(immediate = true, service = { ThingTypeMigrationService.class, ThingManager.class })
 public class ThingManagerImpl
         implements ThingManager, ThingTracker, ThingTypeMigrationService, ReadyService.ReadyTracker {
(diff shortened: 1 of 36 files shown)


bundles/org.openhab.core/src/main/java/org/openhab/core/common/registry/ManagedProvider.java
```diff
@@ -63,6 +65,6 @@
      * @return returns element or null, if no element for the given key exists
      */
     @Nullable
-    E get(K key);
+    E get(@NonNull K key);
```
wborn, line 68: It seems that `@NonNullByDefault` is not applied to generic type parameters. So it would be nice if SAT will not generate warnings for these.

bundles/org.openhab.core.thing/src/test/java/org/openhab/core/thing/internal/ThingManagerImplTest.java
```diff
@@ -17,103 +17,129 @@
...
+    private @Mock @NonNullByDefault({}) ThingTypeRegistry thingTypeRegistryMock;
+    private @Mock @NonNullByDefault({}) ReadyService readyServiceMock;
+    private @Mock @NonNullByDefault({}) SafeCaller safeCallerMock;
+    private @Mock @NonNullByDefault({}) Storage<Object> storageMock;
+    private @Mock @NonNullByDefault({}) StorageService storageServiceMock;
+    private @Mock @NonNullByDefault({}) Thing thingMock;
+    private @Mock @NonNullByDefault({}) ThingRegistryImpl thingRegistryMock;
+
+    // This class is final so it cannot be mocked
+    private final ThingStatusInfoI18nLocalizationService thingStatusInfoI18nLocalizationService = new ThingStatusInfoI18nLocalizationService();
```
wborn, line 67: There are 57 classes defined as "public final class".
If they cannot be easily mocked in tests, it will either result in custom build Mocks or less tests.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
```diff
@@ -347,20 +348,53 @@ public boolean isChannelLinked(ChannelUID channelUID) {
...
+            final @Reference ChannelGroupTypeRegistry channelGroupTypeRegistry,
+            final @Reference ChannelTypeRegistry channelTypeRegistry,
+            final @Reference CommunicationManager communicationManager,
+            final @Reference ConfigDescriptionRegistry configDescriptionRegistry,
+            final @Reference ConfigDescriptionValidator configDescriptionValidator,
+            final @Reference EventPublisher eventPublisher,
+            final @Reference ItemChannelLinkRegistry itemChannelLinkRegistry,
+            final @Reference ReadyService readyService, //
+            final @Reference SafeCaller safeCaller, //
+            final @Reference StorageService storageService, //
```
wborn, line 363: I always use this trick to force the auto formatter to start the next item on a new line. :wink:
That way you don't have to use formatter:on/off for this particular case.

bundles/org.openhab.core.thing/src/test/java/org/openhab/core/thing/internal/ThingManagerImplTest.java
wborn: I think the root cause with this particular case is that the class doesn't implement an interface which would otherwise be perfectly mockable. :wink:

bundles/org.openhab.core/src/main/java/org/openhab/core/common/registry/ManagedProvider.java
wborn: I tried adding `@NonNull` to some generic type parameters but it doesn't seem to be used at all this way. :neutral_face:

maggu2810: Have you tried to state that K in general extends a non null object?

wborn: It seems to work and you can also override the annotations at method level. But it has a domino effect so I have to add/fix more annotations to get this working. :wink:

wborn: Apparently Xtend doesn't support type annotations 

https://bugs.eclipse.org/bugs/show_bug.cgi?id=506374

So there is no way to satisfy the nullness annotations of the `GenericThingProvider` causing the compiler to error on this project:

https://github.com/openhab/openhab-core/blob/aae207bfe95da3b0b7eaaf08e4beda546a1c228d/bundles/org.openhab.core.model.thing/src/org/openhab/core/model/thing/internal/GenericThingProvider.xtend#L76

For now I've reconfigured the compiler to only log warnings for such issues on this project.

wborn:
The PR builds successfully now and the it all still seemed to work well on a local distro build after some brief testing. 

I also created some PRs for other repos so their builds remain successful when this PR gets merged:
* https://github.com/openhab/openhab-addons/pull/7319
* https://github.com/openhab/openhab-webui/pull/214

bundles/org.openhab.core/src/main/java/org/openhab/core/items/ManagedItemProvider.java
```diff
@@ -78,7 +78,7 @@ public PersistedItem(String itemType) {
     public static class PersistedItemInstanceCreator implements InstanceCreator<PersistedItem> {
 
         @Override
-        public PersistedItem createInstance(Type type) {
+        public PersistedItem createInstance(@NonNullByDefault({}) Type type) {
```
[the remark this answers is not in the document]
wborn: It's an overridden method from a Gson class which isn't annotated. It's not null but if you add `@NonNull` or don't annotate the compiler errors:

`Illegal redefinition of parameter type, inherited method from InstanceCreator<ManagedItemProvider.PersistedItem> does not constrain this parameter`

Adding `@Nullable` works but isn't correct. Adding `@NonNullByDefault({})` makes the compiler happy too. The best way would of course be resolving this using  EEAs (https://github.com/openhab/openhab-core/issues/888).

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
```diff
@@ -379,8 +413,12 @@ public void run() {
                     final ThingHandlerFactory oldThingHandlerFactory = findThingHandlerFactory(thing.getThingTypeUID());
                     if (oldThingHandlerFactory != null) {
                         ThingHandler thingHandler = thing.getHandler();
-                        unregisterAndDisposeHandler(oldThingHandlerFactory, thing, thingHandler);
-                        waitUntilHandlerUnregistered(thing, 60 * 1000);
+                        if (thingHandler != null) {
+                            unregisterAndDisposeHandler(oldThingHandlerFactory, thing, thingHandler);
+                            waitUntilHandlerUnregistered(thing, 60 * 1000);
+                        } else {
+                            logger.debug("No ThingHandler to dispose for {}", thing.getUID());
```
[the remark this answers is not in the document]
wborn: You can derive the type from the UID and I think you want to know which specific handler is null in this case because there can be many instances but only this one is null for a reason:

```java
                    // Remove the ThingHandler, if any
                    final ThingHandlerFactory oldThingHandlerFactory = findThingHandlerFactory(thing.getThingTypeUID());
                    if (oldThingHandlerFactory != null) {
                        ThingHandler thingHandler = thing.getHandler();
                        if (thingHandler != null) {
                            unregisterAndDisposeHandler(oldThingHandlerFactory, thing, thingHandler);
                            waitUntilHandlerUnregistered(thing, 60 * 1000);
                        } else {
                            logger.debug("No ThingHandler to dispose for {}", thing.getUID());
                        }
                    } else {
                        logger.debug("No ThingHandlerFactory available that can handle {}", thing.getThingTypeUID());
                    }

```

bundles/org.openhab.core/src/main/java/org/openhab/core/items/ManagedItemProvider.java
```diff
@@ -195,12 +195,13 @@ protected void removeItemFactory(ItemFactory itemFactory) {
     }
 
     @Override
-    protected Item toElement(String itemName, PersistedItem persistedItem) {
+    protected @Nullable Item toElement(String itemName, PersistedItem persistedItem) {
         Item item = null;
 
         if (persistedItem.itemType.equals(GroupItem.TYPE)) {
```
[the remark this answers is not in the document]
wborn: I might do a PR to improve this everywhere in the code base. :wink:

Issue #1415, opened 2020-04-09
[transport.serial] No exception to indicate that a port is not found.


gitMiguel:
Hi all

Core package `org.openhab.core.io.transport.serial`  has no dedicated exception to throw when a serial port is not found. [Javax.comm](https://docs.oracle.com/cd/E17802_01/products/products/javacomm/reference/api/javax/comm/NoSuchPortException.html) package has one but without proper constructors to include a message. This kind of exception can be thrown like:
```
SerialPortIdentifier portIdentifier = portManager.getIdentifier(config.serialPort);
    if (portIdentifier == null) {
        throw new NoSuchPortException("No such port: " + config.serialPort);
    }
```

I'll submit a PR

PR #1416, opened 2020-04-09
[transport.serial] New exception class for serial transport bundle

gitMiguel:
Introduce new exception class for `transport.serial` package. Please see issue #1415 .

Signed-off-by: Miika Jukka <github@lantee.eu>

### DIFF
--- a/bundles/org.openhab.core.io.transport.serial/src/main/java/org/openhab/core/io/transport/serial/NoSuchPortException.java
+++ b/bundles/org.openhab.core.io.transport.serial/src/main/java/org/openhab/core/io/transport/serial/NoSuchPortException.java
@@ -1,0 +1,42 @@
+/**
+ * Copyright (c) 2010-2020 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.io.transport.serial;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
+/**
+ * Exception that marks that a port does not exist.
+ *
+ * @author Miika Jukka - Initial contribution
+ */
+@NonNullByDefault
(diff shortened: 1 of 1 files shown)


cweitkamp:
Wait ... What about `gnu.io.NoSuchPortException`?

wborn:
I don't really see a use for it if it's not thrown by the transport itself. If this is about throwing exceptions in an add-on you can create your own exception for that which can also be reused for throwing other initialization/configuration issues. Usually such exceptions are all handled the same and you can use the message in the exception to update the Thing status details. That's what I did in this [PlugwiseCommunicationContext](
https://github.com/openhab/openhab-addons/blob/5d01bb711bbbb804274050551c3e7ccef3ec09fb/bundles/org.openhab.binding.plugwise/src/main/java/org/openhab/binding/plugwise/internal/PlugwiseCommunicationContext.java#L111-L130).

gitMiguel:
> Wait ... What about `gnu.io.NoSuchPortException`?

[Here it is](https://www.docjava.com/book/cgij/jdoc/gnu/io/NoSuchPortException.html) but it's also missing constructors.

> If this is about throwing exceptions in an add-on you can create your own exception for that which can also be reused for throwing other initialization/configuration related exceptions.

I'm fine with that. I just thought that others might find this useful while developing something serial related.

wborn:
It's probably not a good idea to use any class from `gnu.io` in an add-on. The whole point of the serial transport is to be independent of any serial implementation so we can easily swap it out when necessary without breaking any add-on.

gitMiguel:
I assume that this will not be merged?

wborn:
IMHO it still only makes sense to add it if it is thrown by the transport itself. If it is added in this form bindings will start duplicating the code for throwing it. So you might as well change or add a method to the transport so it is thrown automatically. Changing the existing methods would cause a breaking change for which existing add-ons need to be updated. So maybe adding  an alternative method could make sense. But it will make the transport code/interface more complex which may ultimately not be worth it.

robnielsen:
Shouldn't this be named something like  `InvalidPortException `or `PortDoesNotExistException`?

kaikreuzer:
@wborn Do I read [your comment](https://github.com/openhab/openhab-core/pull/1416#issuecomment-612586516) correctly that you suggest to close this PR?

wborn:
Yes let's close this.

PR #1420, opened 2020-04-11
Use .equals() on constants and literals to prevent NPEs

wborn:
This should replace all such cases where possible so we don't have to spend time on this anymore when creating/reviewing other PRs.

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/provider/AbstractResourceBundleProvider.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/provider/AbstractResourceBundleProvider.java
@@ -306,15 +306,9 @@ public abstract class AbstractResourceBundleProvider<@NonNull E> {
      */
     protected String getParserType(URL url) {
         String fileName = url.getPath();
-        int fileExtensionStartIndex = fileName.lastIndexOf(".") + 1;
-        if (fileExtensionStartIndex == -1) {
-            return Parser.FORMAT_JSON;
-        }
-        String fileExtension = fileName.substring(fileExtensionStartIndex);
-        if ("txt".equals(fileExtension)) {
-            return Parser.FORMAT_JSON;
-        }
-        return fileExtension;
+        int index = fileName.lastIndexOf(".");
+        String extension = index != -1 ? fileName.substring(index + 1) : "";
+        return extension.isEmpty() || "txt".equals(extension) ? Parser.FORMAT_JSON : extension;
     }
 
     /**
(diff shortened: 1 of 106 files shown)


J-N-K:
IMO the other way is more readable (closer to natural speech) and I don‘t see the benefit if we have null-notations in place.

wborn:
That is true but lots of code isn't annotated and code and annotations can change over time causing NPEs eventually.

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/provider/AbstractResourceBundleProvider.java
```diff
@@ -306,15 +306,15 @@ protected void putNewPortfolio(Vendor vendor, List<String> portfolio) {
      */
     protected String getParserType(URL url) {
         String fileName = url.getPath();
-        int fileExtesionStartIndex = fileName.lastIndexOf(".") + 1;
-        if (fileExtesionStartIndex == -1) {
+        int fileExtensionStartIndex = fileName.lastIndexOf(".") + 1;
+        if (fileExtensionStartIndex == -1) {
```
cweitkamp, line 310: This looks like a bug. How can `fileExtensionStartIndex` equals to -1 when we add 1 in the line before?

We should either do this:

```suggestion
        if (fileName.lastIndexOf(".") == -1) {
```

Or add one later in `fileName.substring(fileExtensionStartIndex + 1);

Maybe do it similar to `AbstractFileProvider#getParserType()`.

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/provider/file/AbstractFileProvider.java
```diff
@@ -317,7 +317,7 @@ private String getParserType(URL url) {
             return Parser.FORMAT_JSON;
         }
         String fileExtension = fileName.substring(fileName.lastIndexOf(".") + 1);
```
cweitkamp, line 319: Store `fileName.lastIndexOf(".")` in a local variable earlier and reuse it here.

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/provider/AbstractResourceBundleProvider.java
wborn: I think you're right so it's fixed now. :+1:

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/provider/file/AbstractFileProvider.java
wborn: Done! And I threw the Elvis operator in the mix. ?:-)

wborn:
It was quite a bit of a grind this whole PR but it was totally self inflicted. :wink:

PR #1421, opened 2020-04-12
Add and fix more null annotations

wborn:
There's also https://github.com/openhab/openhab-addons/pull/7348 to keep the add-ons port working when these annotations get added.

### DIFF
--- a/bundles/org.openhab.core.binding.xml/src/main/java/org/openhab/core/binding/xml/internal/XmlBindingInfoProvider.java
+++ b/bundles/org.openhab.core.binding.xml/src/main/java/org/openhab/core/binding/xml/internal/XmlBindingInfoProvider.java
@@ -18,6 +18,7 @@ import java.util.Set;
 import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.binding.BindingInfo;
 import org.openhab.core.binding.BindingInfoProvider;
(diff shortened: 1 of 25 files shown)


bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/ConfigDescription.java
```diff
@@ -90,8 +93,8 @@ public ConfigDescription(URI uri, List<ConfigDescriptionParameter> parameters) {
      * @throws IllegalArgumentException if the URI is null or invalid
      */
     @Deprecated
-    public ConfigDescription(URI uri, List<ConfigDescriptionParameter> parameters,
-            List<ConfigDescriptionParameterGroup> groups) {
+    public ConfigDescription(@Nullable URI uri, @Nullable List<ConfigDescriptionParameter> parameters,
+            @Nullable List<ConfigDescriptionParameterGroup> groups) {
```
wborn, line 97: There are some deprecated constructors (see above) passing null values.

bundles/org.openhab.core.config.xml/src/main/java/org/openhab/core/config/xml/AbstractXmlBasedProvider.java
```diff
@@ -173,9 +177,12 @@ private T_OBJECT acquireLocalizedObject(Bundle bundle, T_OBJECT object, @Nullabl
             return cacheEntry;
         }
 
+        @Nullable
         final T_OBJECT localizedObject = localize(bundle, object, locale);
         if (localizedObject != null) {
-            localizedObjectCache.put(localizedKey, localizedObject);
+            @NonNull
+            T_OBJECT nonNullLocalizedObject = (@NonNull T_OBJECT) localizedObject;
```
wborn, line 184: The compiler failed without explicitly casting `localizedObject`. If it's not assigned here and casted as parameter on the line below, the cast would be removed by the auto save actions, making the compiler error again. Using `@format:on/off` didn't help. :slightly_frowning_face:

bundles/org.openhab.core.binding.xml/src/main/java/org/openhab/core/binding/xml/internal/XmlBindingInfoProvider.java
```diff
@@ -65,8 +67,10 @@
 
     @Activate
     public XmlBindingInfoProvider(final @Reference BindingI18nLocalizationService bindingI18nService,
+            final @Reference(target = "(esh.scope=core.xml.binding)") ConfigDescriptionProvider configDescriptionProvider,
```
cweitkamp, line 70: Nice trick.

PR #1422, opened 2020-04-14
[ui] Switch widget as default widget for Number or String items with command options

### LINKED ISSUE #1016: Rendering of command options
Command options is a new feature that is not yet supported neither in Basic UI nor in Classic UI. And I doubt it is in Android App too.

Before adding the support of this feature in Basic UI (and in few bindings), I would like to be sure to understand what is expected.

Command options can be different from state options.

I see 3 different cases:

1. the channel has only command options but no state (for example the clear queue channel of the Sonos binding)
2. the channel has command options and a state without state options
3. the channel has command options and a state with state options

I imagine the command options have to be considered when the user uses the switch or the selection sitemap element.

For the cases 1 and 2, we should render the command options as buttons for the switch element and as a selection for the selection element, the same as we do currently when having state options.

For the case 3, does it mean we have to consider command options + state options, meaning building one button for each command option + one button for each state option ? In this case, we have to consider only one item when one command option is the same as one item option ?

Last question is relative (again) to switch item. If I come back to my example, the clear queue channel is associated to a switch item. This channel should have only one command option: CLEAR. Is it acceptable for a switch element having command options to render it with buttons or a selection list ? I ask the question because it was recently rejected to render like that a switch item with state options. Or should we update the item type of this channel to String and only consider String and Number items for command options ?

@maniac103 @mueller-ma for Android App

cweitkamp:
- Switch widget as default widget for Number or String items with command options

Closes #1016


cweitkamp:
Together with https://github.com/openhab/openhab-webui/pull/217 the rendering looks like this:

https://user-images.githubusercontent.com/5131747/79213268-2c598c80-7e49-11ea-9553-d7e1004389b5.png

lolodomo:
I think that the default should remain a selection. The buttons are not adapted when you have 50 options !

cweitkamp:
I considered checking number of command options against `MAX_BUTTON = 4` (see https://github.com/openhab/openhab-webui/blob/909226e2725caa6aa4aed59a61f6b57d8110f464/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/SwitchRenderer.java#L55) and use Switch element if less or equals and Selection element else. Wdyt?

lolodomo:
Yes, that would be ok IMHO to consider a max number of options to choose between  switch or selection. 4 looks ok to me as a reasonable value.

cweitkamp:
I added the check for maximum number of options.

PR #1423, opened 2020-04-14
Model refactoring, Memory Leak prevention, UTF-8 encoding

cweitkamp:
- Model refactoring, Memory Leak prevention, UTF-8 encoding

Beside our basic improvements (nullness annotations and constructor injection) this draft tries to address two other issues - I will link them, when I find them - eventually they are placed in old ESH repository. First one is a potential memory leak in the `ModelRepositoryÌmpl` where we did not close the copied `InputStream`s (e.g. for validation). Second one is a potential encoding issue - I added `XtextResource.OPTION_ENCODING = "UTF-8"` in all places where the resources are loaded.

What do you think about that? I am fine with splitting this PR into smaller pieces.

But there might be a bigger problem related to the internal `resourceSet` (`SynchronizedXtextResourceSet`). Today I faced a system crash and luckily I could save the heap dump. @wborn I would like to share it with you for deeper analysis.

grafik

### DIFF
--- a/bundles/org.openhab.core.model.core/src/main/java/org/openhab/core/model/core/internal/ModelRepositoryImpl.java
+++ b/bundles/org.openhab.core.model.core/src/main/java/org/openhab/core/model/core/internal/ModelRepositoryImpl.java
@@ -19,7 +19,6 @@ import java.nio.charset.StandardCharsets;
 import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.core.model.core/src/main/java/org/openhab/core/model/core/internal/ModelRepositoryImpl.java
```diff
@@ -96,16 +100,18 @@ public EObject getModel(String name) {
...
-            InputStream inputStream = null;
             if (originalInputStream != null) {
                 byte[] bytes = IOUtils.toByteArray(originalInputStream);
-                String validationResult = validateModel(name, new ByteArrayInputStream(bytes));
-                if (validationResult != null) {
-                    logger.warn("Configuration model '{}' has errors, therefore ignoring it: {}", name,
-                            validationResult);
-                    removeModel(name);
-                    return false;
+                try (InputStream validationStream = new ByteArrayInputStream(bytes)) {
```
wborn, line 107: I'm OK with the changes, but this change will probably not fix memory leaks since the close method of a `ByteArrayInputStream` doesn't do anything:

```java
    /**
     * Closing a {@code ByteArrayInputStream} has no effect. The methods in
     * this class can be called after the stream has been closed without
     * generating an {@code IOException}.
     */
    public void close() throws IOException {
    }
```

It's just an `InputStream` backed by a byte array that doesn't hold any other resources besides the array and a few variables. It doesn't hurt to always use try-with-resources to prevent issues, but it's only really required for a few Stream implementations.

Issue #1408, opened 2020-04-05
Remove deprecated classes/methods from OH3


wborn:
There are quite a few classes and methods marked as `@Deprecated` that can be removed from the OH3 code base.

Classes marked as `@Deprecated` are:

* [x] org.openhab.core.config.discovery.DiscoveryServiceCallback (see #1424)
* [x] org.openhab.core.config.discovery.ExtendedDiscoveryService (see #1424)
* [x] org.openhab.core.config.discovery.UpnpDiscoveryParticipant (see #1418)
* [x] org.openhab.core.config.discovery.inbox.InboxFilterCriteria (see #1417)
* [x] org.openhab.core.types.StateDescriptionProvider (see #1413)
* [x] org.openhab.core.io.net.http.TrustManagerProvider (see #1425)
* [x] org.openhab.core.io.transport.mdns.discovery.MDNSDiscoveryParticipant (see #1418)
* [x] org.openhab.core.io.transport.mqtt.sslcontext.CustomSSLContextProvider (see #1668)
* [x] org.openhab.core.io.transport.mqtt.sslcontext.AcceptAllCertificatesSSLContext (see #1668)
* [x] org.openhab.core.io.transport.mqtt.sslcontext.SSLContextProvider (see #1668)

There are even more methods marked as `@Deprecated`, some examples are:

* [x] org.openhab.core.items.ItemRegistry.newItemBuilder(Item) (see #1429)
* [x] org.openhab.core.items.ItemRegistry.newItemBuilder(String, String) (see #1429)
* [x] org.openhab.core.net.NetUtil.getLocalIpv4HostAddress() (see #1622)
* [x] org.openhab.core.config.core.ConfigOptionProvider.getParameterOptions(URI, String, Locale) (see #1541, #1667)
* [x] org.openhab.core.thing.i18n.ChannelTypeI18nLocalizationService.createLocalizedStateDescription​ (see #1918)
* [x] org.openhab.core.thing.type.StateChannelTypeBuilder.withStateDescription​ (see #1918)

Classes with constructors marked as `@Deprecated`:

* [x] org.openhab.core.config.core.ConfigDescriptionParameterGroup​ (see #1918)
* [x] org.openhab.core.thing.Channel​ (see #1918)
* [x] org.openhab.core.thing.type.ChannelType​ (see #1918)
* [x] org.openhab.core.types.StateDescription​ (see #1918)

[...]

PR #1425, opened 2020-04-14
Remove deprecated TrustManagerProvider

wborn:
It was deprecated since the introduction of the ExtensibleTrustManager (https://github.com/eclipse/smarthome/pull/6281).

Related to #1408


bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/internal/WebClientFactoryImpl.java
```diff
@@ -373,7 +350,9 @@ private SslContextFactory createSslContextFactory(@Nullable String endpoint) {
         if (endpoint == null) {
             return createSslContextFactoryFromExtensibleTrustManager();
         } else {
-            return createSslContextFactoryFromTrustManagerProvider(endpoint);
+            SslContextFactory sslContextFactory = new SslContextFactory();
```
cweitkamp, line 353: Even if we are talking about three lines of code. They are duplicate. Should we merge this method with `createSslContextFactoryFromExtensibleTrustManager()` and initialize the `SSLContext` only if `endpoint == null`? 

Or why have we to distinguish here at all.

wborn: That change makes sense. I'm not that familiar with all this code though.

From what I understand, because of the deprecation, there wouldn't be any instance of a `TrustManagerProvider` anymore. So all code guarded by `trustmanagerProvider != null`, would effectively be dead code.

Maybe @martinvw can also comment on if these deprecated methods are cleaned up the right way?

Even if we don't fully understand how this works... now is the time to clean it up because OH3 will go through some rigorous testing.

wborn: If the `endpoint` parameter is only used to differentiate between the deprecated  `TrustManagerProvider` and `ExtensibleTrustManager` we might just want to remove the parameter and always use the `ExtensibleTrustManager`?

wborn: Looks like because of the removal of the other methods, the `endpoint` would now always be `null`.

cweitkamp: The latest change looks nice. And it should cause no problems because everything was internal.

martinvw:
> Maybe @martinvw can also comment on if these deprecated methods are cleaned up the right way?

It looks correct from what I can see without spending too long on it :-)

cweitkamp:
Thank you both.

PR #1426, opened 2020-04-14
Update nrjavaserial to 3.20.0

### LINKED ISSUE #1384: nrjavaserial crashes Java 11 JVM on Windows
The whole JVM crashes on Windows when using nrjavaserial as RXTX implementation with Java 11.

We must not forget to integrate a fix for this or OH3 releases might not be very usable on Windows.

See also:
* https://github.com/NeuronRobotics/nrjavaserial/issues/131
* https://github.com/openhab/openhab-distro/issues/768

wborn:
Updates nrjavaserial to the official 3.20.0 release (compatible with Java 8/11).
With the upgraded version the library will no longer crash the Java 11 VM on Windows.
The native libraries now use the built in RXTX lockfiles instead of the previously used (OS dependent) liblockdev whereas the OH build didn't use any lockfiles at all.
It also has a fix for errors being printed when scanning for serial ports whenever a serial port is locked (NeuronRobotics/nrjavaserial#166).
With the new version RXTX threads also have proper names and there is a fix for an IllegalMonitorStateException that might occur when closing ports.

Fixes #1384

### DIFF
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -240,7 +240,7 @@
     <dependency>
       <groupId>com.neuronrobotics</groupId>
       <artifactId>nrjavaserial</artifactId>
-      <version>3.14.0</version>
+      <version>3.20.0</version>
       <scope>compile</scope>
     </dependency>
 
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -466,14 +466,9 @@
 
     <!-- All serial transports -->
     <dependency>
-      <groupId>org.openhab</groupId>
+      <groupId>com.neuronrobotics</groupId>
       <artifactId>nrjavaserial</artifactId>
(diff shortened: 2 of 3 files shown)


kaikreuzer:
Hooray, thanks @wborn for driving this!

RafalLukawiecki:
Will this make into 2.5.x? I wonder if the locking issue is what is causing Java abort traps under FreeBSD, see https://community.openhab.org/t/instructions-for-openhab-on-freenas/37958/32?u=rafal

wborn:
That is not likely, see also this [community discussion](https://community.openhab.org/t/is-nrjavaserial-still-maintained/93885/4?u=wborn).

Also if you're using FreeBSD, there will not be any updated native libraries because these haven't been recompiled. Maybe you can help with that? See https://github.com/NeuronRobotics/nrjavaserial/issues/164.

PR #1450, opened 2020-04-28
[timer] Add Timer::isActive()

jimtng:
The following is referring to org.eclipse.smarthome.model.script.actions.Timer

There seems to be no way to programmatically tell whether a timer is no longer "active" e.g. because it was cancelled.

There's Timer::hasTerminated() but that will only return true if the timer has finished executing. It will not return true if timer.cancel() was called before the scheduled execution occurred. 

Then there's Timer::isRunning() but that will only return true if the timer has reached its time and is actually still executing and not yet finished.

Example code to illustrate:
```
@rule("test1")
@when("System started")
def test(event):  
    timer = ScriptExecution.createTimer(DateTime.now().plusSeconds(5), lambda: test.log.info("Timer executing. isRunning(): {}".format(timer.isRunning())))
    test.log.info("timer.hasTerminated(): {}, isRunning(): {}".format(timer.hasTerminated(), timer.isRunning())) // will log as "timer.hasTerminated(): 0, isRunning(): 0"
    timer.cancel()
    test.log.info("timer.hasTerminated(): {}, isRunning(): {}".format(timer.hasTerminated(), timer.isRunning())) // will log as "timer.hasTerminated(): 0, isRunning(): 0"
```

As a result of this, people have been resorting to assigning the timer instance to null when the timer got cancelled and also when it has terminated following the completion. Then they would check whether the timer variable is null or not to know whether the timer is still active. People have gotten so used to doing it this way it seems. However, this complicates the code unnecessarily. 

I understand that in some, and perhaps most cases, knowing whether a timer has been cancelled or terminated is not required by the code, but in other cases, it is.

I would propose adding Timer::isActive() to the core to return true when a timer is still in progress, i.e. it has not been cancelled nor finished.

This would greatly simply scripting logic involving timers.

### DIFF
--- a/bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/actions/Timer.java
+++ b/bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/actions/Timer.java
@@ -31,6 +31,13 @@ public interface Timer {
     public boolean cancel();
 
     /**
+     * Determines whether the scheduled execution is yet to happen.
+     *
+     * @return true, if the code is still scheduled to execute, false otherwise
+     */
+    public boolean isActive();
+
+    /**
      * Determines whether the scheduled code is currently executed.
      *
      * @return true, if the code is being executed, false otherwise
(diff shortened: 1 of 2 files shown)


jimtng:
If this is merged, we'd need to update the documentation too.

kaikreuzer:
I have created https://github.com/openhab/openhab-docs/issues/1207, so that it isn't forgotten.
Please note that this change will only become available in openHAB 3, so no need to update the current 2.5.x docs.

PR #1451, opened 2020-04-28
Replaced "classic" rule engine by a DSLRuleProvider for the NGRE

kaikreuzer:
This PR removes the "classic" rule engine completely, so that the NGRE now becomes the one and only rule engine of openHAB (and I'd suggest we can call it from now on simply "rule engine").

While the "classic" rule engine is removed, the support for DSL rules is kept, including their parsing through Xtext. This is imho tremendously important, since we would not want to break millions of lines of rules that already exist out there.

The newly introduced DSLRuleProvider serves this purpose nicely. It takes the parsed dsl rule files, creates a rule instance for each contained rule in the file, which then consists out of a list of triggers and always a single script action with scripts of mime-type `application/vnd.openhab.dsl.rule`.

Those rules are nicely shown in the new rule editor in the UI:
Bildschirmfoto 2020-04-28 um 22 52 06

And also their YAML representation is quite well readable:
Bildschirmfoto 2020-04-28 um 22 53 25

Some further changes that had to be done along the way:
- All traces of Quartz have been completely removed. The `Timer` class now works with our own scheduler and should be completely backward compatible. The "core.scheduler" bundle has thus been completely removed.
- The `now` keyword in rules now resolves to `java.time.ZonedDateTime.now()` instead of `java.time.Instant.now()`. This should bring us convenience methods like `now.plusHours(1)` back.
- ["Member of" triggers](https://www.openhab.org/docs/configuration/rules-dsl.html#member-of-triggers) were added to the rule engine, which will definitely continue to be asked for, so the new rule engine has them now as well.
- A "system trigger" type has been introduced with a rudimentary "startlevel" implementation. This is a first step towards https://github.com/eclipse/smarthome/issues/1896 and should be the appropriate successor for the "system started" rule trigger. At the moment, it only waits some time, before such "startup" rules are executed, but I hope we can further work on the startlevel service so that our rule lifecycle becomes much more reliable. 
- "system shutdown" triggers are no longer supported - I doubt that anybody is using them at all and if there's the strong which to keep them, it can possibly also be mapped to the new startlevel trigger once it also supports direction.

While it all already works quite well, I marked this as WIP as I will still have to do some polishing, extensive manual testing as well as adding unit tests for it.
I just wanted to make the work already visible to you for early testing and discussions.


bundles/org.openhab.core.automation/src/main/resources/OH-INF/automation/moduletypes/ItemTriggers.json
```diff
@@ -231,6 +231,256 @@
...
+                            "value": "DOWN"
+                        }
+                    ]
+                }
+            ],
+            "outputs": [
+                {
+                    "name": "triggeringItem",
+                    "type": "org.openhab.core.items.Item",
+                    "description": "the member of the group that changed it state",
```
ghys, line 461:
```suggestion
                    "description": "the member of the group that changed its state",
```

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/SystemTriggerHandler.java
```diff
@@ -0,0 +1,120 @@
...
+import org.openhab.core.events.EventFilter;
+import org.openhab.core.events.EventSubscriber;
+import org.openhab.core.events.system.StartlevelEvent;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.ServiceRegistration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This is a ModuleHandler implementation for Triggers which trigger the rule if a certain system events occurs.
```
ghys, line 39:
```suggestion
 * This is a ModuleHandler implementation for Triggers which trigger the rule if a certain system event occurs.
```

ghys:
Are you still planning on adding an `editable: false` flag in the API for rules provided by the DSLRuleProvider like we discussed, so that the UI could let users know the rule cannot be edited with it (and ideally `editable: true` otherwise)?

image
Minor comment on why you're getting this `|-2`:
(the multiline handling in YAML is so convoluted you have a whole website dedicated to them: https://yaml-multiline.info/)
I suppose the 2-character indent as seen above is due to the fact that the "then" block in DSL rule files is generally indented, maybe there could be a way to strip the indentation when processing the rule (perhaps use the first line to determine the number of spaces to trim).

digitaldan:
Amazing @kaikreuzer !  Since 3.0 should not be used yet for any production system,I vote to  merge this sooner than later so we can test with the snapshot builds (i'm lazy) .

kaikreuzer:
> Are you still planning on adding an editable: false flag in the API for rules 

Yes, absolutely! Still on my to-do list and I wondered if I should include it here or do a separate PR upfront.

> maybe there could be a way to strip the indentation when processing the rule

Yes, I thought about that as well - will check what I can do.

> I vote to merge this sooner than later so we can test with the snapshot builds

Yes, I'd like this as well. But give me maybe another week to bring it in a mergeable state :-)

kaikreuzer:
>> maybe there could be a way to strip the indentation when processing the rule

> Yes, I thought about that as well - will check what I can do.

Implemented!

bundles/org.openhab.core.model.rule.runtime/src/org/openhab/core/model/rule/runtime/internal/DSLRuleProvider.java
```diff
@@ -264,14 +265,26 @@ private Rule toRule(String modelName, boolean hasContext, org.openhab.core.model
...
                 .withTypeUID("script.ScriptAction").withConfiguration(cfg).build());
 
         return RuleBuilder.create(uid).withName(name).withTriggers(triggers).withActions(actions).build();
     }
 
+    private String removeIndentation(String script) {
+        String s = script;
+        if (s.startsWith("\n")) {
+            s = s.substring(1);
+        }
```
ghys, line 280: This doesn't work on Windows, you'll want to add this as well:
```java
        if (s.startsWith("\r\n")) {
            s = s.substring(2);
        }
```
Maybe also add a final newline if it doesn't exist (that would make the YAML multiline control a simple `|` instead of `|-`)

kaikreuzer: These Windows users ;-)
Done!

kaikreuzer:
> Are you still planning on adding an editable: false flag in the API for rules

Implemented!

bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/dto/EnrichedRuleDTOMapper.java
```diff
@@ -20,13 +21,16 @@
...
  */
 public class EnrichedRuleDTOMapper extends RuleDTOMapper {
 
-    public static EnrichedRuleDTO map(final Rule rule, final RuleManager ruleEngine) {
+    public static EnrichedRuleDTO map(final Rule rule, final RuleManager ruleEngine,
+            final ManagedRuleProvider managedRuleProvider) {
         final EnrichedRuleDTO enrichedRuleDto = new EnrichedRuleDTO();
         fillProperties(rule, enrichedRuleDto);
         enrichedRuleDto.status = ruleEngine.getStatusInfo(rule.getUID());
+        enrichedRuleDto.editable = managedRuleProvider.get(enrichedRuleDto.name) != null;
```
ghys, line 33: Isn't that supposed to be
```suggestion
        enrichedRuleDto.editable = managedRuleProvider.get(enrichedRuleDto.uid) != null;
```

kaikreuzer: I should have tested it myself, sorry. Fixed it.

ghys:
@kaikreuzer I've noticed that the UI didn't update the status of a non-managed DSL rule in the details page, turns out the topic filter for `/rest/events` doesn't like the dot:
```
$ http http://localhost:8080/rest/events?topics=smarthome/rules/demo.rules-2/*
HTTP/1.1 400 Bad Request
Cache-Control: must-revalidate,no-cache,no-store
Content-Length: 332
Content-Type: text/html;charset=iso-8859-1
Date: Sun, 03 May 2020 09:52:38 GMT
Server: Jetty(9.4.20.v20190813)
...

$ http http://localhost:8080/rest/events?topics=smarthome/rules/demo_rules-2/*
HTTP/1.1 200 OK
Content-Encoding: identity
Content-Type: text/event-stream
Date: Sun, 03 May 2020 09:54:37 GMT
Server: Jetty(9.4.20.v20190813)
Transfer-Encoding: chunked
X-Accel-Buffering: no
```

We could perform some escaping but you might also want to consider to revisit the automatic UIDs of these rules? (replace with an underscore maybe)

kaikreuzer:
@openhab/core-maintainers To finish this, I'll need to work on the itests, but I am currently struggling to get them running within the IDE. The jvm starts up, does something and then hangs without executing any tests. Before I dive too deep into the analysis: Should this be working straight away or is there currently anything special to be aware of when running itests in the IDE? I might not be up to date with the latest situation, so please help me to get up to speed :-)

kaikreuzer:
Forget my question - I've got it all running!

kaikreuzer:
> We could perform some escaping but you might also want to consider to revisit the automatic UIDs of these rules?

@ghys I have changed this to simply remove the ".rules" extension from the UID. I initially thought it is safer to have it to make it clear that those rules come from files, but I think that people should be able to do the match in their head also without the extension.

ghys:
Cool, that should leave no ambiguity, yes. And with the number at the end still present the risk of an eventual conflict with a managed rule's UID is low too.

kaikreuzer:
I'll keep adding tests and do further testing myself on this PR, but I'd like to invite you to start reviewing it now, so that it becomes in general ready to be merged!

bundles/org.openhab.core.model.script.runtime/src/org/openhab/core/model/script/runtime/internal/engine/DSLScriptEngine.java
```diff
@@ -0,0 +1,243 @@
...
+            }
+
+            @Override
+            public String getLanguageVersion() {
+                return "v1";
+            }
+
+            @Override
+            public String getLanguageName() {
+                return "openHAB Rule DSL";
```
ghys, line 223: By the way @kaikreuzer, have you noticed the "o" gets capitalized between here and the UI? I don't think it's done client-side, so not sure where it happens.

lolodomo:
This PR is a very good news, thank you, it will avoid a massive disagreement when OH will be released. 

In addition to individual rules, we have global constants and variables.
Are they considered by the new rule provider?

kaikreuzer:
>In addition to individual rules, we have global constants and variables.
> Are they considered by the new rule provider?

They are not global, they are shared for all rules within the same file. And yes, I have kept this supported. You will find a line like
```
// context: demo-1
```
as a first line in the script action. This is evaluated by the DSLScriptEngine and the associated context of its file gets automatically injected for execution.

bundles/org.openhab.core.model.script.runtime/src/org/openhab/core/model/script/runtime/internal/engine/DSLScriptEngine.java
kaikreuzer: Yes, I had noticed this as well. I assume it must be the javax.script implementation that capitalizes it. I've now simply changed the name to "Rule DSL", so we don't have to bother anymore ;-)

openhab-5iver:
Is there a good reason for all of these changes to be in a single PR? Unless I've overlooked something,  the triggers and handlers, system events, and possibly Quartz/scheduler could all be in separate PRs. This makes me chuckle a bit, since it is usually Kai who asks this question! :laughing:

digitaldan:
@kaikreuzer I finally got around to building the distro based on this PR,  and then tried loading all the rules from my production 2.x system.  Everything loaded without a single exception, which is impressive considering how many rules of i have accumulated over the years (which I now realize i need to go back and clean up!). 

I was not able to test each and every one as i don't have some bindings yet running (like z-wave), but i did trigger many of them and they did appear to work.  So very nice job!

Let me know if there is anything in particular you would like to see tested.

kaikreuzer:
@openhab-5iver Main reason is that there are strong dependencies between the code changes and all are required for testing the feature. Once I am done, I can split it into separate parts, if the reviewers prefer it. Main part that could be treated independently are probably the new automation modules.

Thanks for testing, @digitaldan - your report sounds encouraging!
I am still experimenting with good ways for the startup behavior and reloading upon item/thing changes to make it all a bit more robust than the old rule engine.

kaikreuzer:
> "Member of" triggers were added to the rule engine, which will definitely continue to be asked for, so the new rule engine has them now as well.
> A "system trigger" type has been introduced with a rudimentary "startlevel" implementation. This is a first step towards eclipse-archived/smarthome#1896 and should be the appropriate successor for the "system started" rule trigger. At the moment, it only waits some time, before such "startup" rules are executed, but I hope we can further work on the startlevel service so that our rule lifecycle becomes much more reliable.

This has been moved to https://github.com/openhab/openhab-core/pull/1509.

kaikreuzer:
Note, this PR depends on #1514 (which was extracted from here), so it might show build errors as long as #1514 is not merged.

@openhab/core-maintainers I am through with my additional testing and am quite confident that it should work well. Please feel free to jump into the code review now!

bundles/org.openhab.core.boot/bnd.bnd
```diff
@@ -1,2 +1 @@
 Bundle-SymbolicName: ${project.artifactId}
```
wborn, line 1: This file only existed for adding the activator to the manifest. So it should be removed now.

bundles/org.openhab.core.model.rule.runtime/src/org/openhab/core/model/rule/runtime/internal/DSLRuleProvider.java
```diff
@@ -0,0 +1,432 @@
...
+    private final Map<String, IEvaluationContext> contexts = new HashMap<>();
+    private final Map<String, XExpression> xExpressions = new HashMap<>();
+    private int triggerId = 0;
+    private Set<String> markers = new HashSet<>();
+
+    @Reference
+    private @NonNullByDefault({}) ModelRepository modelRepository;
+
+    @Reference
+    private @NonNullByDefault({}) ReadyService readyService;
```
wborn, line 102: Can you inject these fields using constructor injection instead? That allows for making them final and also makes it easier to write unit tests for this class without having to resort to reflection tricks.

bundles/org.openhab.core.model.rule/src/org/openhab/core/model/rule/jvmmodel/RulesRefresher.java
```diff
@@ -29,51 +49,133 @@
...
+    @Reference
+    protected ThingRegistry thingRegistry;
 
-    public void unsetModelRepository(ModelRepository modelRepository) {
-        this.modelRepository = null;
+    @Reference
+    protected EventPublisher eventPublisher;
+
+    @Reference
+    protected ReadyService readyService;
```
wborn, line 84: Personally I never use `@Reference` on fields because it complicates writing unit tests or perhaps you know an easy way to inject these fields in tests?

bundles/org.openhab.core.model.rule/src/org/openhab/core/model/rule/jvmmodel/RulesRefresher.java
```diff
@@ -29,51 +49,133 @@
...
             }
+            readyService.markReady(marker);
+        }, REFRESH_DELAY, TimeUnit.MILLISECONDS);
+        readyService.unmarkReady(marker);
+    }
+
+    private void setStartLevel() {
+        if (!started) {
+            started = true;
+            // TODO: This is still a very dirty hack in the absence of a proper system start level management.
```
wborn, line 122: Can this TODO be done in this PR?

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/jvmmodel/ScriptJvmModelInferrer.xtend
```diff
@@ -76,19 +94,29 @@ class ScriptJvmModelInferrer extends AbstractModelInferrer {
...
-		 	val name = item.name
-				if (fieldNames.add(name)) {
-		 		members += 	script.toField(item.name, script.newTypeRef(item.class)) [
-			 		static = true
-			 	]
-		 	} else {
-		 		logger.warn("Duplicate field: '{}'. Ignoring '{}'.", item.name, item.class.name)
-		 	}
-		 ]
+        itemRegistry?.items?.forEach[ item |
```
wborn, line 97: Please use the same indentation everywhere in this file instead of mixing tabs and spaces.

bundles/org.openhab.core.model.script.runtime/src/org/openhab/core/model/script/runtime/internal/engine/DSLScriptEngineFactory.java
```diff
@@ -0,0 +1,66 @@
...
+@NonNullByDefault
+@Component(service = ScriptEngineFactory.class)
+public class DSLScriptEngineFactory implements ScriptEngineFactory {
+
+    private static final String SCRIPT_TYPE = "dsl";
+
+    private @NonNullByDefault({}) DSLScriptEngine dslScriptEngine;
+
+    @Reference
+    protected @NonNullByDefault({}) ScriptEngine scriptEngine;
```
wborn, line 42: Use constructor injection here instead?

bundles/org.openhab.core.model.rule/src/org/openhab/core/model/rule/jvmmodel/RulesRefresher.java
kaikreuzer: No, definitely not - this is about https://github.com/eclipse-archived/smarthome/issues/1896, which will be a huge new PR in itself.

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/jvmmodel/ScriptJvmModelInferrer.xtend
kaikreuzer: I now applied the formatter on the full file, which touched most lines.

kaikreuzer:
Many thanks @wborn, I have addressed your comments.

cweitkamp:
:+1: Great job. Nice to have it in OHC now.

PR #1453, opened 2020-05-01
[automation] Fixed creation of 'EphemerisConditionHandler'

### LINKED ISSUE #1452: NGRE missing DaySet handler
https://community.openhab.org/t/ngre-dayset-missing-handler/97657

I am unable to use the daySet without an offset, in the NGRE.

cweitkamp:
- Fixed creation of `EphemerisConditionHandler`

Closes #1452

The `offset` is not required it should not throw an exception at all.

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/handler/BaseModuleHandler.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/handler/BaseModuleHandler.java
@@ -16,6 +16,7 @@ import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.automation.Module;
 import org.openhab.core.automation.ModuleHandlerCallback;
+import org.openhab.core.config.core.Configuration;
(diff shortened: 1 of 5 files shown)


bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/handler/BaseModuleHandler.java
```diff
@@ -41,4 +43,32 @@ public void setCallback(ModuleHandlerCallback callback) {
     public void dispose() {
         this.callback = null;
     }
+
+    /**
+     * Returns the {@link Module}, which belongs to the handler.
+     *
+     * @return {@link Module}, which belongs to the handler
+     */
+    public T getModule() {
```
kaikreuzer, line 52: This becomes public API, so I do not understand why that is necessary for fixing an NPE?
It also does not seem to be used at all.

cweitkamp: Yes, you are right. this method is not needed at all. I removed it.

Issue #1275, opened 2019-12-15
Jetty.version different between runtime and compile. Why?


kgoderis:
Just a question, 

bom/compile/pom.xml contains  <jetty.version>9.3.25.v20180904</jetty.version> but bom/runtime/pom.xml contains <jetty.version>9.4.20.v20190813</jetty.version> 

Is there a specific reason for this?

wborn:
@maggu2810 explained the thought behind this in https://github.com/openhab/openhab-core/pull/467#issuecomment-455773562:

> That's a very important point, the framework must not dictate the exact version of Karaf and Jetty that needs to be used by a product.
I don't see any change that e.g. openHAB and the products I am maintaining (and perhaps other consumers of the framework) will use the same versions of Karaf and Jetty all the time.

kgoderis:
@maggu2810 @wborn There are breaking changes between 9.3 and 9.4, and at least for me it created issues during development of a new binding, with ClassNotFound errors and alike. Why isn't the compile version not bumped to 9.4? I find this confusing to be honest.

kaikreuzer:
Imho it should be fine to bump the Jetty prerequisite to 9.4 (possibly an older patch level). Current Karaf versions use 9.4 and there are indeed some incompatibilities that can cause issues and confusion.

maggu2810:
> I find this confusing to be honest.

I am sorry about that.

Perhaps you missed, but I am not involved in the openHAB development anymore.

PR #1464, opened 2020-05-08
Update nrjavaserial in TP to 5.0.2 and define import version range

wborn:
The first version of nrjavaserial with OSGi manifest entries is 3.12.0.
The interface has not changed and is still the same with nrjavaserial 5.x.
So using 3.12.0 as compile time dependency and defining a [3.12,6) version range using bnd.bnd files allows for maximum flexibility.

### DIFF
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -223,8 +223,6 @@
     </dependency>
 
     <!-- Serial Communication -->
-    <!-- <dependency> <groupId>org.eclipse.kura</groupId> <artifactId>org.eclipse.soda.dk.comm</artifactId> <version>1.2.201</version>
-      <scope>compile</scope> </dependency> -->
     <dependency>
       <groupId>org.vesalainen.comm</groupId>
       <artifactId>javaxcomm</artifactId>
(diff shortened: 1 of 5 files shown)


wborn:
There may still be OH add-ons not using the transport and depending directly on gnu.io. When 3.12.0 is used for compilation bnd automatically adds a [3.12,4) version range for those add-ons. That may cause issues when only a 5.x bundle is deployed.

So we could:

1. add a compile BOM in the add-ons project, where we override specific dependency versions for OH only, that may also solve other add-on issues, e.g. when we know that a newer version of Jetty is available e.g. because of the Karaf dependency (https://github.com/openhab/openhab-core/issues/1275)
1. update the bnd imports for those add-ons
1. migrate them to use the serial transport
1. change the compile time dependency here, though it is not a core dependency issue, we would also no longer see when we do depend on a newer version

wborn:
I've created https://github.com/openhab/openhab-addons/issues/7573 since we want 4. to be addressed anyway.

wborn:
I've implemented option 3. with https://github.com/openhab/openhab-addons/pull/7589 so we have the same flexibility with all add-ons depending on gnu.io directly as well.

PR #1465, opened 2020-05-08
[thing] Removed 'getThingByUID()' method from 'BaseBridgeHandler'

cweitkamp:
- Removed `getThingByUID()` method from `BaseBridgeHandler`

The code for `BaseBridgeHandler#getThingByUID(ThingUID)` and `BridgeImpl#getThing(ThingUID)` are identical thus I marked it as deprecated.
Another option will be to remove it right now. 

---

Related to https://github.com/openhab/openhab-addons/pull/7665

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/BaseBridgeHandler.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/BaseBridgeHandler.java
@@ -12,8 +12,6 @@
  */
 package org.openhab.core.thing.binding;
 
-import java.util.List;
-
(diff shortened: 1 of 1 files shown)


cpmeister:
Why deprecate it when you can leave it as a convenience function?

cweitkamp:
We can keep it as convenience function. It at least increases maintainability a little bit.

kaikreuzer:
As we rather wanted to completely move away from methods in the base handler (and have the callback instead), I'd even say that we could remove them instead of deprecating them.

cweitkamp:
Done. I removed it. PR for OH add-ons: https://github.com/openhab/openhab-addons/pull/7665

PR #1470, opened 2020-05-11
[uom] Allow DateTimeFormat for labels of Number:Time items

cweitkamp:
- Allow DateTimeFormat for labels of Number:Time items

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityType.java
@@ -15,6 +15,9 @@ package org.openhab.core.library.types;
 import static org.eclipse.jdt.annotation.DefaultLocation.*;
 
 import java.math.BigDecimal;
+import java.time.Instant;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


cpmeister:
What about duration formats? Can you add support for those as well?

cweitkamp:
Supporting a `Duration` format would be a possible enhancement. But to be honest I am not aware of a nice way to support user-defined `Duration` format patterns. I know that `toString()` method prints a ISO-8601 seconds based representation, such as PT8H6M12.345S. Do you have a hint?

cpmeister:
Found a bit of code on stacktrace that might help:
```
public static String formatDuration(Duration duration) {
    long seconds = duration.getSeconds();
    long absSeconds = Math.abs(seconds);
    String positive = String.format(
        "%d:%02d:%02d",
        absSeconds / 3600,
        (absSeconds % 3600) / 60,
        absSeconds % 60);
    return seconds < 0 ? "-" + positive : positive;
}
```
Unlike jodatime, java's time framework doesn't apparently come with a prebuilt formatter.

cweitkamp:
@cpmeister I do not think an algorithm is the problem. I am looking for a pattern which users can add e.g. to their item labels. Similar to what we can do with other Java types (see https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Formatter.html#syntax).

Issue #1462, opened 2020-05-05
RFC2217 URI does not work in OH 2.5 (on FreeBSD?)


RafalLukawiecki:
Unsure if this is related to FreeBSD or is it a more generic issue, but using the ZWave binding 2.5.5 latest snapshot and OH 2.5.4, which in turn is using nrjavaserial-3.15.0.OH2, I cannot get the binding to recognise a ser2net remote serial by specifying it as "rfc2217://192.168.1.1:3333" etc. The binding reports offline with error:
```
Status: OFFLINE - COMMUNICATION_ERROR Serial Error: Port rfc2217://192.168.1.1:3333 does not exist
```
However, that port works when I use socat to link to a local "real" port, although OH2 crashes later due to another issue related to serial port lockfiles, see #1461  

I have discussed it to some extent with @cdjackson, [see this thread](https://community.openhab.org/t/share-z-wave-dongle-over-ip-usb-over-ip-using-ser2net-socat-guide/34895/235?u=rafal). @cdjackson seems adamant that the issue is not with the binding but with the underlying serial.

Could this issue be related to the 3.15.OH2 version of nrjavaserial? From a cursory glance at their repo I can see that version 3.16 introduced some changes to the user-side of defining the serial port in rfc2217 terms, even if the key support for it came in version 3.10.

Could it be related to the extra Java option `-Dgnu.io.rxtx.SerialPorts=/dev/ttyUSB0"` that needs to know what serial ports there are? If so, the syntax of that option does not allow for colons, so providing the URI in it does not work, either.

Happy to test any other suggestions.

cdjackson:
As I mentioned, I don’t think nrjavaserial is related - I think the rfc2217 is handled separately through a different abstraction. However I’m not really familiar with where this is done so hopefully someone else can help with that. You might also want to link the ESH issue I referenced earlier as well?

RafalLukawiecki:
Apologies if I have misquoted you, @cdjackson. Please see these two commits in nrjavaserial that mention RFC2217: https://github.com/NeuronRobotics/nrjavaserial/commit/4d1a0351d4a128db63de810c6c31f9fe71174bed and https://github.com/NeuronRobotics/nrjavaserial/commit/8f54ed981ac8aeb509cef79ccf400f08f9d51c70

[...]

PR #1471, opened 2020-05-12
Improve SerialPortManager.getIdentifiers() JavaDoc

wborn:
The current JavaDocs do not clearly state that it does not always return all identifiers.
E.g. it will not return any RFC2217 identifiers or undiscovered RXTX port identifiers.
So it should not be used to search for an identifier when the serial port name is known.
This method is useful for listing available port options in UIs and logging.

---

Related to:
* https://github.com/openhab/openhab-core/issues/1462
* https://github.com/openhab/openhab-addons/pull/7609

### DIFF
--- a/bundles/org.openhab.core.io.transport.serial/src/main/java/org/openhab/core/io/transport/serial/SerialPortManager.java
+++ b/bundles/org.openhab.core.io.transport.serial/src/main/java/org/openhab/core/io/transport/serial/SerialPortManager.java
@@ -44,7 +44,7 @@ public interface SerialPortManager {
      * Gets the discovered serial port identifiers.
      *
      * {@link SerialPortProvider}s may not be able to discover any or all identifiers.
-     * When the port name is known, the preferred way to get an identifier is by using {@link #getIdentifier(String).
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.io.transport.serial/src/main/java/org/openhab/core/io/transport/serial/SerialPortManager.java
```diff
@@ -41,9 +41,12 @@
     }
 
     /**
-     * Gets the serial port identifiers.
+     * Gets the discovered serial port identifiers.
      *
-     * @return stream of serial port identifiers
+     * {@link SerialPortProvider}s may not be able to discover any or all identifiers.
+     * When the port name is known, the preferred way to get an identifier is by using {@link #getIdentifier(String).
```
cweitkamp, line 47: A missing curly bracket? Isn't it?

```suggestion
     * When the port name is known, the preferred way to get an identifier is by using {@link #getIdentifier(String)}.
```

wborn: Well spotted! The curly has returned from its abscence. :-)

PR #1476, opened 2020-05-16
Give the missing channel group type

silamon:
A nullpointer exception is thrown when it cannot find the channel group type. This happens during development of bindings when you made a typo. This add a helpful message to the logs to help the binding developer.

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/thing/ThingTypeResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/thing/ThingTypeResource.java
@@ -202,6 +202,10 @@ public class ThingTypeResource implements RESTResource {
             String id = channelGroupDefinition.getId();
             ChannelGroupType channelGroupType = channelGroupTypeRegistry
                     .getChannelGroupType(channelGroupDefinition.getTypeUID(), locale);
+            if (channelGroupType == null) {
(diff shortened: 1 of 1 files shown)


wborn:
I think it's already fixed by https://github.com/openhab/openhab-core/pull/1475

See: https://github.com/openhab/openhab-core/pull/1475/files#diff-0f142a94933b2d1278ecf15a3d5e54faR211

I found the same issue after adding null annotations.

silamon:
@wborn, thanks for the review. You have fixed it in your PR indeed. Still, I would find it more helpful if a message was given anyway. I've changed my PR based on that. It will silently skip the channel group right now and log a warning message. This will now return an incomplete thing type whatsoever instead of a nullpointer exception. I've changed the channel type method to have the same behavior.

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/thing/ThingTypeResource.java
```diff
@@ -184,9 +184,6 @@ public Response getByUID(@PathParam("thingTypeUID") @ApiParam(value = "thingType
 
         final List<ChannelDefinitionDTO> channelDefinitions = convertToChannelDefinitionDTOs(
                 thingType.getChannelDefinitions(), locale);
-        if (channelDefinitions == null) {
-            return null;
```
wborn, line 188: Since it won't return `null` now `@Nullable` should be removed from the method result type.

PR #1481, opened 2020-05-17
[discovery] Added validation for relation between ThingUID and BridgeUID

cweitkamp:
- Added validation for relation between `ThingUID` and `BridgeUID`

See discussion in https://github.com/openhab/openhab-addons/pull/7496#discussion_r425425821

### DIFF
--- a/bundles/org.openhab.core.config.discovery/src/main/java/org/openhab/core/config/discovery/DiscoveryResultBuilder.java
+++ b/bundles/org.openhab.core.config.discovery/src/main/java/org/openhab/core/config/discovery/DiscoveryResultBuilder.java
@@ -36,7 +36,7 @@ public class DiscoveryResultBuilder {
     private final ThingUID thingUID;
 
     private @Nullable ThingUID bridgeUID;
-    private @Nullable Map<String, Object> properties;
+    private final Map<String, Object> properties = new HashMap<>();
     private @Nullable String representationProperty;
     private @Nullable String label;
     private long ttl = DiscoveryResult.TTL_UNLIMITED;
@@ -75,7 +75,9 @@ public class DiscoveryResultBuilder {
      * @return the updated builder
      */
     public DiscoveryResultBuilder withProperties(@Nullable Map<String, Object> properties) {
-        this.properties = properties;
+        if (properties != null) {
+            this.properties.putAll(properties);
+        }
         return this;
     }
 
@@ -85,11 +87,7 @@ public class DiscoveryResultBuilder {
      * @param property of the desired result
      * @return the updated builder
      */
-    @SuppressWarnings("null")
     public DiscoveryResultBuilder withProperty(String key, Object value) {
-        if (properties == null) {
-            properties = new HashMap<>();
-        }
         properties.put(key, value);
         return this;
     }
--- a/bundles/org.openhab.core.config.discovery/src/main/java/org/openhab/core/config/discovery/DiscoveryResultBuilder.java
+++ b/bundles/org.openhab.core.config.discovery/src/main/java/org/openhab/core/config/discovery/DiscoveryResultBuilder.java
@@ -36,7 +36,7 @@ public class DiscoveryResultBuilder {
     private final ThingUID thingUID;
 
     private @Nullable ThingUID bridgeUID;
-    private final Map<String, Object> properties = new HashMap<>();
+    private @Nullable Map<String, Object> properties;
     private @Nullable String representationProperty;
     private @Nullable String label;
     private long ttl = DiscoveryResult.TTL_UNLIMITED;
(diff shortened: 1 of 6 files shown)


bundles/org.openhab.core.config.discovery/src/main/java/org/openhab/core/config/discovery/DiscoveryResultBuilder.java
```diff
@@ -36,7 +36,7 @@
     private final ThingUID thingUID;
 
     private @Nullable ThingUID bridgeUID;
-    private final Map<String, Object> properties = new HashMap<>();
+    private @Nullable Map<String, Object> properties;
```
splatch, line 39: I would be careful here since it changes behavior of the builder. I know that 3.0.x addons must be recompiled to new packages, however such small changes remain undocumented and can blow up only when running a discovery with (most often) real things.

cweitkamp: You are right. I reverted it for the moment. Even if my change streamlined this builder with other OHC builders.

bundles/org.openhab.core.config.discovery/src/test/java/org/openhab/core/config/discovery/DiscoveryResultBuilderTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testDiscoveryResultBuilderWithBridge() {
+        @SuppressWarnings("unused")
+        DiscoveryResult otherDiscoveryResult = DiscoveryResultBuilder
+                .create(new ThingUID(THING_TYPE_UID, "otherThingId")).withBridge(BRIDGE_UID).build();
+    }
+
+    @Test
+    @Ignore
```
wborn, line 96: Why is this test ignored?

cweitkamp: To streamline our builders (following https://github.com/eclipse-archived/smarthome/issues/5929) I first changes the behavior of the `withProperties(Map<String, Object>)` method to replace existing properties map by the one set. This has been reverted after https://github.com/openhab/openhab-core/pull/1481#discussion_r429679569 and thus the test is failing now. Will come up with a follow-up PR if the change should be applied at all to separate it from this feature.

cweitkamp: Have a look [here](https://github.com/openhab/openhab-core/commit/59a59b4a0a127a741054beb9ee48bd47735cb5cb) for the changes I am talking about.

kaikreuzer (approved):
Thanks!

cweitkamp:
Thanks for merging. The unit test were not yet migrated to JUnit 5. I fixed them in https://github.com/openhab/openhab-core/pull/1619.

kaikreuzer:
Thanks for #1619 and sorry that I didn't notice that.

splatch:
FYI I spotted this one while migrating my own [bindings](https://github.com/connectorio/connectorio-addons). Now it did hit bacnet integration. This validation rule is completely unclear to me. It was not enforced in 2.5 hence it impacts migration and makes maintenance of 2.5 and 3.0 versions troublesome.

I believe this check is unnecessary since each and every binding declares bridge-thing matches. The `ThingUID` in my understanding serves an ID and should be used in such a way. Looking at fragment of @kaikreuzer [comment](https://github.com/openhab/openhab-addons/pull/7496#discussion_r425759542) in related addon: 
> ThingUID: It is recommended to use it (especially when textually defining a thing), but there must not be any semantics being assumed within a ThingUID.

This change does introduce assumption that ThingUID include bridge identifier. I believe that this change might be also problematic when binding use multiple nesting levels. For example when there is more than just one bridge above the thing.

Please also update developer/migration docs accordingly so other people do not fall into same trap all over again.

cdjackson:
It does feel that this is a bad idea. If we think back a few years, we had these links (ie the thing type, bridge were encoded in the thing uid). This was removed as it caused problems when the thing type or bridge changed. A UID should be universally unique - it should NEVER change through the life of a system (in theory at least). Now, by imposing a hard link to the bridge UID, it is no longer possible to move a thing between bridges (if I understand this correctly) without changing the thing UID (which IMHO is a no-no and will have side effects through persistence etc).

cweitkamp:
I understand your concerns. But as stated in https://github.com/openhab/openhab-core/issues/1800#issuecomment-723420138 I do not seen a problem here. This validation is restricted for building discovery results and will not prevent switching bridges during runtime. AFAIK the UI uses a similar approach to create random ThingUIDs for Things related to Bridges. And it is the recommended syntax to be used for textual configuration.

Both of you are long time contributors and probably remember a lot more historical changes than I do so if it really is a problem or a step backwards we can revert it before final OH 3 release.

PR #1484, opened 2020-05-19
[automation] Use UTF-8 to read files

cweitkamp:
- Use UTF-8 to read files

Otherwise I am seeing e.g. such log messages:

```
2020-05-19 14:37:56.639 [INFO ] [ipse.smarthome.model.script.presence] - Trigger fÃ¼r presenceChristophPresent changed from OFF to ON ...
```

Or wrong tags:

```JSON
        "tags": [
            "Anwesenheit",
            "KÃ¼che"
        ],
```

### DIFF
--- a/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/loader/ScriptFileWatcher.java
+++ b/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/loader/ScriptFileWatcher.java
@@ -21,6 +21,7 @@ import java.io.InputStreamReader;
 import java.net.MalformedURLException;
 import java.net.URISyntaxException;
 import java.net.URL;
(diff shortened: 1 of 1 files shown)


cweitkamp:
I am using Python combined with the helper libraries.

openhab-5iver:
Are you using unicode strings?

cweitkamp:
```
automation/jsr223/python/personal/presence.py: Python script, UTF-8 Unicode text executable
```

openhab-5iver:
@cweitkamp, I have tested this and I did not see any difference in functionality. Do you have a before/after test to show that this change is doing something?

cweitkamp:
IIRC `InputStreamReader` defaults to system default charset - which might not always be UTF-8. Specifying it does not hurt as we try to keep it aligned throughout the framework.

I learned of [this community post](https://community.openhab.org/t/no-international-characters-unicode-with-jsr223-jython/99567) that OHC is not the real problem but encoding used by current Python / Jython version. For me masking a String via `u"..."` in my rules and saving the files with UTF-8 encoding is a working solution.

PR #1491, opened 2020-05-21
Use ConfigurableService as ComponentPropertyType

wborn:
It is easier to use the `ConfigurableService` properties when it's possible to annotate classes using a `ComponentPropertyType`.

---

I gave it a test and it is backwards compatible with the ported add-ons that still use the properties.


wborn:
The Whiteboard annotations are based on the same functionality.

kaikreuzer:
Hm, at least the JavaDoc build does not seem to like it, though: https://ci.openhab.org/view/Documentation%20(3.x)/job/openHAB-JavaDoc/45/console

wborn:
I've triggered an openHAB-Core build which might help since openHAB-JavaDoc only generates docs and doesn't compile code.

PR #1345, opened 2020-01-27
Reenable javadoc generation

mhilbush:
@kaikreuzer This turned out to be a bit more involved than I expected.

I started with what was done in ESH, but immediately ran into problems trying to build the javadocs. It turns out there's an issue with javadoc in Java 11 that results in a nasty (and silent) ClassCastException. I traced it down to this construct `{@link SomeType[]}`. Fortunately, there were only 6 occurrences in the codebase, all of them in `org.openhab.core.voice`. Here's an example:
https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/text/ASTNode.java#L66

In this PR, I simply removed the @ in those 6 occurrences until I get some direction on how best to handle these 6 occurrences, as well as how to prevent the offending construct from being reintroduced in the future. In hindsight, I suppose it might've been better to change it to something such as `{@link String array}`.

There's also an issue with the Java 11 javadoc that inserts "undefined" in the path name to the classes. This was breaking the search feature. I eventually found the parameter `--no-module-directories` option, which fixed that.

Finally, the "Frames" view of the javadoc is deprecated, and disabled by default. I reenabled it using the parameter `--frames`. Not sure if you want to keep the frames view (the search feature is supposed to be the better way to find things).

In any event, I've tested with the following 2 commands. The results look correct to me.
```
mvn javadoc:aggregate
mvn javadoc:aggregate-jar
```

Signed-off-by: Mark Hilbush <mark@hilbush.com>

PR #1494, opened 2020-05-22
Update maven-javadoc-plugin to 3.2.0

wborn:
See: https://github.com/openhab/openhab-core/pull/1345#issuecomment-632639071

It seems to produce something useful when I run:

`mvn javadoc:javadoc javadoc:aggregate -B -Dfeatures.verify.skip=true -Dspotless.check.skip=true -T 1C`

Can you check the result @mhilbush?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -319,7 +319,7 @@ Import-Package: \\
             <doclint>none</doclint>
             <!-- Prevents the "undefined" module in path name, which causes a file not found when searching -->
             <additionalJOption>--no-module-directories</additionalJOption>
-            <excludePackageNames>*.internal,*.internal.*,nl.*,org.openhab.core.*.test,org.openhab.core.*.test.*</excludePackageNames>
+            <excludePackageNames>*.internal,*.internal.*</excludePackageNames>
           </configuration>
         </plugin>
 
--- a/pom.xml
+++ b/pom.xml
@@ -315,11 +315,20 @@ Import-Package: \\
           <version>3.2.0</version>
           <configuration>
             <failOnError>!${quality.skip}</failOnError>
-            <!-- Turn off doclint -->
             <doclint>none</doclint>
-            <!-- Prevents the "undefined" module in path name, which causes a file not found when searching -->
-            <additionalJOption>--no-module-directories</additionalJOption>
             <excludePackageNames>*.internal,*.internal.*</excludePackageNames>
+            <!-- The search function is broken without the workaround below -->
+            <!-- See: https://stackoverflow.com/questions/52326318/maven-javadoc-search-redirects-to-undefined-url -->
+            <additionalJOption>--allow-script-in-comments</additionalJOption>
+            <bottom>
+              <![CDATA[
+                <script>
+                if (typeof useModuleDirectories !== 'undefined') {
+                  useModuleDirectories = false;
+                }
+                </script>
+              ]]>
+            </bottom>
           </configuration>
         </plugin>
 
--- a/pom.xml
+++ b/pom.xml
@@ -312,13 +312,11 @@ Import-Package: \\
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-javadoc-plugin</artifactId>
-          <version>2.10.3</version>
+          <version>3.2.0</version>
           <configuration>
             <failOnError>!${quality.skip}</failOnError>
             <!-- Turn off doclint -->
-            <additionalparam>-Xdoclint:none</additionalparam>
-            <!-- Frames are deprecated and will disappear at some point -->
-            <additionalparam>--frames</additionalparam>
+            <doclint>none</doclint>
             <!-- Prevents the "undefined" module in path name, which causes a file not found when searching -->
             <additionalJOption>--no-module-directories</additionalJOption>
             <excludePackageNames>*.internal.*,nl.*</excludePackageNames>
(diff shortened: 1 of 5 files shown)


wborn:
Excluding test packages also makes sense because these won't be available in the distro

E.g.:

```
org.openhab.core.*.test
org.openhab.core.*.test.*
```

That would still allow for showing some packages that are available when writing tests:

```
org.openhab.core.test
org.openhab.core.test.java
org.openhab.core.test.storage
```

pom.xml
```diff
@@ -312,16 +312,14 @@ Import-Package: \\
...
+          <version>3.2.0</version>
           <configuration>
             <failOnError>!${quality.skip}</failOnError>
             <!-- Turn off doclint -->
-            <additionalparam>-Xdoclint:none</additionalparam>
-            <!-- Frames are deprecated and will disappear at some point -->
-            <additionalparam>--frames</additionalparam>
+            <doclint>none</doclint>
             <!-- Prevents the "undefined" module in path name, which causes a file not found when searching -->
             <additionalJOption>--no-module-directories</additionalJOption>
```
wborn, line 321: It may fix the searching, but is causes another issue, i.e. all links to Oracle no longer work for standard Java classes such as Object, String etc.

mhilbush:
> mvn javadoc:javadoc javadoc:aggregate -B -Dfeatures.verify.skip=true -Dspotless.check.skip=true -T 1C

I had to remove the `-T 1C` for it to run on my Mac. When I did that, it ran successfully. IIRC, `javadoc:aggregate` has always worked.

However, the build runs `javadoc:aggregate-jar`. And this is fails for me with the same `cannot find symbol` error.

> mvn javadoc:javadoc javadoc:aggregate-jar -B -Dfeatures.verify.skip=true -Dspotless.check.skip=true

```
/Users/mark/Development/openhab-master/git/openhab-core/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/SseResource.java:151: error: cannot find symbol
    @Produces(MediaType.SERVER_SENT_EVENTS)
                       ^
  symbol:   variable SERVER_SENT_EVENTS
  location: class MediaType
/Users/mark/Development/openhab-master/git/openhab-core/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/SseResource.java:182: error: cannot find symbol
    @Produces(MediaType.SERVER_SENT_EVENTS)
                       ^
  symbol:   variable SERVER_SENT_EVENTS
  location: class MediaType

Command line was: /Library/Java/JavaVirtualMachines/zulu-11.jdk/Contents/Home/bin/javadoc --no-module-directories @options @packages
```

pom.xml
mhilbush: Yes, at the time it was the lesser of two evils. I see you found a better workaround. Excellent!

mhilbush:
I'll try running with your changes using the updated plugin.

wborn:
OK that part I missed, but with these changes it generated a nice big `org.openhab.core.reactor-3.0.0-SNAPSHOT-javadoc.jar` with all docs in just 5 minutes in a parallel build. :slightly_smiling_face:

mhilbush:
Just pulled the PR. Running it now.

mhilbush:
This worked perfectly for me. So it must've been the feature verification that was tripping me up when I was trying to use the 3.1.1 plugin version.
```
mvn javadoc:aggregate-jar -Dfeatures.verify.skip=true
```

And
- you found a better workaround for the missing directory issue
- it looks like the issue with arrays in the Javadoc (e.g. `{@link SomeType[]}`) has been fixed

So overall this puts us in a better place.

wborn:
Great! Yes the feature verification doesn't make sense when code isn't build and only docs are generated. The new workaround may not have been an option for the older plugin version.

wborn:
I've also created JavaDocs for 2.5.0 based on the same changes.
See this [org.openhab.core.reactor-2.5.0-javadoc.jar](https://github.com/wborn/openhab-core/releases/download/javadoc-2.5.0/org.openhab.core.reactor-2.5.0-javadoc.jar) and [javadoc-2.5.0](https://github.com/wborn/openhab-core/tree/javadoc-2.5.0) branch.

For standard Java classes it will go to the Java 11 docs. From what I read that was the first version that allowed for external links. That's better than nothing and usually there are also `@since` tags to indicate in which version functionality was first added.

So perhaps @ghys can statically host the 2.5.0 docs and dynamically update the 3.0.0-SNAPSHOT docs? The Jenkins build should also generate a JAR containing the docs now (when it works again).

ghys:
Thanks all!

> dynamically update the 3.0.0-SNAPSHOT

I modified the [Jenkins job](https://ci.openhab.org/job/openHAB-JavaDoc/) so that it produces easy-to-download artifacts in a tar file, then added a [couple lines](https://github.com/openhab/website/blob/5fdf18ed9fb47523483602b0c6c330226f8f12c9/prepare-docs.rb#L609) to the next website build script.
It works! The latest Javadoc should be kept current at:

https://next.openhab.org/javadoc/latest/ 🥳 

(every time that site is redeployed that is, but you can do it manually from Jenkins by launching: https://ci.openhab.org/job/redeploy-website-v3/, we can also configure it as a downstream project)

wborn:
Thank you too @ghys! :+1:

ghys:
Javadoc for v2.5 is now available at:
https://www.openhab.org/javadoc/v2.5/
and won't change (it was simply copied from @wborn's jar above)

Any idea why the v3 javadoc doesn't include `org.openhab.core.io.rest.auth` e.g. `TokenResource`?

wborn:
That's because those classes are in the excluded `.internal` packages.

kaikreuzer:
Great, thanks @ghys!

mhilbush:
Thanks @wborn and @ghys. Your help on this is very much appreciated. Questions about the API docs come up periodically on the forum, so this will be a big help, especially to new binding developers.

PR #1483, opened 2020-05-18
Add OAuth2 information to Swagger API description

ghys:
As discussed in https://github.com/openhab/openhab-core/pull/1482#issuecomment-630458832.

This allows the REST docs client to perform the OAuth2 flow
to let the user try operations requiring an authorization.

Requires a small change on the client-side (Swagger UI)
to set the client ID properly.

Required authorization scopes should be added to all
API operations requiring authorization using annotations.

Signed-off-by: Yannick Schaus <github@schaus.net>

PR #1496, opened 2020-05-24
Swagger: support injection of ReaderListener, add OAuth2 annotations

ghys:
This is another attempt at #1483 without any blatant (but inadvertent!) third-party code borrowing & relicensing this time.

This introduces the ability to [customize the Swagger Definition](https://github.com/swagger-api/swagger-core/wiki/Annotations-1.5.X#customising-the-swagger-definition) in a way supported directly by swagger-jaxrs by providing `ReaderListener` implementations with DI. Two of them are implemented in this PR: one to add general information and the other to add the OAuth2 security definition (which will only be injected if the auth bundle is installed).

Additionally the enumeration of resources has been simplified for openHAB's purposes as they all implement `RESTResource` and there is only one `Application`, so they can be looked up and fed to the Swagger `Reader` easily.

It seems to work, with no change in functionality, therefore the dependency to  `jaxrswb-swagger1-gen` can be removed, in line with the current push to remove unneeded dependencies, as well as mitigating the risk of relying on third-party libraries when the functionality they offer can be easily replaced with first-party code (nothing personal against @maggu2810! And thanks for the [valuable insight here](https://github.com/maggu2810/jaxrswb/blob/6c57f9a7de310a12213ed3b2fdcab37bebc845a4/bundles/jaxrswb-utils/src/main/java/de/maggu2810/jaxrswb/gen/JaxRsWhiteboardBaseSpecialGenerator.java#L61)).

There was also a problem with a conflict between Guava versions (as discussed [here](https://github.com/openhab/openhab-distro/issues/1045#issuecomment-596255184)), solved by fixing the `org.reflections:reflections` dependency to version 0.9.12 (`swagger-jaxrs` depends on version 0.9.11) because this [recent release](https://github.com/ronmamo/reflections/releases/tag/0.9.12) removes the dependency to an outdated release of Guava. This seems to do the trick without side effects and apparently solves resolutions problems in both the demo app and Karaf. This allows in turn to remove the personal fork of this lib (`de.maggu2810.thirdparty.modified.org.reflections/reflections/0.9.10.v20160429-1435`), in favor of an official version.

The OAuth2 flow in Swagger UI depends on https://github.com/openhab/openhab-webui/pull/257 to work correctly.

image

### DIFF
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -230,9 +230,10 @@
       <scope>compile</scope>
     </dependency>
     <dependency>
-      <groupId>de.maggu2810.jaxrswb.bundles</groupId>
-      <artifactId>jaxrswb-swagger1-gen</artifactId>
-      <version>0.0.4</version>
+      <groupId>io.swagger</groupId>
+      <artifactId>swagger-jaxrs</artifactId>
+      <version>1.5.5</version>
+      <scope>compile</scope>
     </dependency>
 
     <!-- XStream -->
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -226,13 +226,13 @@
     <dependency>
       <groupId>io.swagger</groupId>
       <artifactId>swagger-annotations</artifactId>
-      <version>1.6.1</version>
+      <version>1.5.5</version>
       <scope>compile</scope>
     </dependency>
     <dependency>
       <groupId>io.swagger</groupId>
       <artifactId>swagger-jaxrs</artifactId>
-      <version>1.6.1</version>
+      <version>1.5.5</version>
       <scope>compile</scope>
     </dependency>
 
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -226,13 +226,13 @@
     <dependency>
       <groupId>io.swagger</groupId>
       <artifactId>swagger-annotations</artifactId>
-      <version>1.5.5</version>
+      <version>1.6.1</version>
       <scope>compile</scope>
     </dependency>
     <dependency>
       <groupId>io.swagger</groupId>
       <artifactId>swagger-jaxrs</artifactId>
-      <version>1.5.5</version>
+      <version>1.6.1</version>
       <scope>compile</scope>
     </dependency>
 
--- a/features/karaf/openhab-core/src/main/feature/feature.xml
+++ b/features/karaf/openhab-core/src/main/feature/feature.xml
@@ -151,7 +151,9 @@
 		<feature>openhab-core-auth-jaas</feature>
 		<bundle>mvn:org.openhab.core.bundles/org.openhab.core.io.rest.auth/${project.version}</bundle>
 		<requirement>openhab.tp;filter:="(feature=jose4j)"</requirement>
+		<requirement>openhab.tp;filter:="(feature=swagger-jaxrs)"</requirement>
 		<feature dependency="true">openhab.tp-jose4j</feature>
+		<feature dependency="true">openhab.tp-swagger-jaxrs</feature>
 	</feature>
 
 	<feature name="openhab-core-io-rest-log" version="${project.version}">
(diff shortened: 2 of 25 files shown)


bom/compile/pom.xml
```diff
@@ -238,13 +238,14 @@
     <dependency>
       <groupId>io.swagger</groupId>
       <artifactId>swagger-annotations</artifactId>
-      <version>1.5.5</version>
+      <version>1.6.1</version>
```
wborn, line 241: Do you know if it also compiles with the older version? If that's the case it's better to use older versions here. That way a broader version range is generated by Bnd and the framework will be more compatible with older Swagger versions. That way you don't force framework users to upgrade dependencies or use multiple versions, i.e. it could result in similar issues as we often have with guava and the org.reflections artifact.

ghys: Ah, thanks, it compiled fine before, the version change was done at the last minute (what I thought was) some regular housekeeping, I'll change it back then.

features/karaf/openhab-core/src/main/feature/feature.xml
```diff
@@ -155,7 +155,9 @@
 		<feature>openhab-core-auth-jaas</feature>
 		<bundle>mvn:org.openhab.core.bundles/org.openhab.core.io.rest.auth/${project.version}</bundle>
 		<requirement>openhab.tp;filter:="(feature=jose4j)"</requirement>
+		<requirement>openhab.tp;filter:="(feature=swagger-jaxrs)"</requirement>
 		<feature dependency="true">openhab.tp-jose4j</feature>
+		<feature dependency="true">openhab.tp-swagger-jaxrs</feature>
```
ghys, line 160: I'm not too happy with openhab-core-io-rest-auth depending on openhab.tp-swagger-jaxrs after all, if there's a better suggestion (make it optional? bring the ReaderListener, back to org.openhab.core.io.rest.swagger...) I'll gladly take it.

maggu2810:
@ghys Just for you information:
For me the approach Swagger is using to find the classes that are relevant to scan does not work. I used multiple different Whiteboard applications and relied a lot of the dynamic approach and the class loader separation of OSGi.
Swagger's JAX-RS scanner is using e.g. classgraph to find relevant classes. This could work for your OSGi runtime at the moment but I don't think you can expect that every class is accessible in OSGi without explicitly import it.
If you ever feel that some classes are not part of the documentation, remember :wink: (but perhaps I did something wrong).
Because of "finding all classes" and "use the correct path for the resources of different applications" I created the project.
I also posted about it on the Aries mailing list some time: https://lists.apache.org/thread.html/106af9ca8125e98327bd998b02f3f0a059cd179bdcb008805c71ffe4%40%3Cuser.aries.apache.org%3E
I have been told that there is some upstream (Aires JAX-RS Whiteboard) work for simpler Swagger integration, too. But I don't know its current progress.

maggu2810:
FYI: https://github.com/apache/aries-jax-rs-whiteboard/pull/23

features/karaf/openhab-core/src/main/feature/feature.xml
wborn: Dependency wise I agree it would be better to move `ReaderListener` back to `org.openhab.core.io.rest.swagger`.

wborn:
Since there is a Whiteboard 1.0.9 release with Swagger support, I think in the long term it's best to see if we can also use the Whiteboard integration in OH3 to generate the .json . We will probably need to upgrade swagger-core for that and I can help with that. I'm OK with merging this PR in the meantime since it will make the REST Docs a lot more functional.

While testing this PR again with the new Swagger UI it no longer seems to work for me. When I click Authorize, select the scope I get to the openHAB OAuth2 authorization page. But after I then enter my credentials it shows:

oauth2

and logs:

`[WARN ] [tp.auth.internal.AuthorizePageServlet] - Authentication failed: CSRF check failed`

It may also have to do that in the mean time I've upgraded all my OS-es to Ubuntu 20.04 and upgraded Eclipse / Java in the proces as well. :wink:

features/karaf/openhab-core/src/main/feature/feature.xml
ghys: The intent was to prevent the OAuth2 info injection to occur if auth isn't active i.e. the instance doesn't have the io-rest-auth bundle active. Surely there's a clean way of detecting that differently (and reliably/cleanly) somehow?

ghys:
> While testing this PR again with the new Swagger UI it no longer seems to work

I fixed my broken Eclipse and tested that again and you're right - the redirect URL was missing in the Swagger UI config. https://github.com/openhab/openhab-webui/pull/296 fixes it.

features/karaf/openhab-core/src/main/feature/feature.xml
wborn: I'm fine with moving it and always injecting the OAuth2 info for now. We could also create a service that list the available authentication methods so you could query that.

wborn:
> openhab/openhab-webui#296 fixes it

Yes it seem to work well now! Can you also fix the merge conflict @ghys?

ghys:
@wborn it should be good now, I also addressed the comment [above ](https://github.com/openhab/openhab-core/pull/1496#discussion_r459909846) and removed the REST auth dependency to Swagger.

PR #1503, opened 2020-05-27
Update nrjavaserial in TP to 5.2.1

wborn:
With this version it will no longer completely exit :exclamation:  the JVM in certain scenarios when ports are closed/reopened.

Related to:

* openhab/org.openhab.binding.zigbee#577
* openhab/org.openhab.binding.zwave#1332

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -486,7 +486,7 @@
     <dependency>
       <groupId>com.neuronrobotics</groupId>
       <artifactId>nrjavaserial</artifactId>
-      <version>5.0.2</version>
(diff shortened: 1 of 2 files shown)


wborn:
It should not be updated because the API is still compatible with 3.12. If it would be updated Bnd will raise the lower bound of the version range in package dependencies. So the framework will be compatible with more versions if we use older compile dependencies. This way you can still easily downgrade the nrjavaserial bundle in your openHAB instance with an older version if you run into some blocking issue.

cweitkamp:
Thanks for explaining again. Now I can remember that you told it before.

PR #1505, opened 2020-05-31
Added 'ChannelDescriptionChangedEvent'

### LINKED ISSUE #1753: What event is published when a binding is changing dynamically a list of state options
When a binding is changing dynamically a list of state options (using a DynamicStateDescriptionProvider), the result of my tests (listening to topic `openhab/items/*/*`) is that there is no item event published on the event bus to inform of this change.

Is there at least an event published ,maybe on another topic ?

I would have expected an event ItemUpdatedEvent with the new state description.

@cweitkamp : I am sure you are aware of that.

cweitkamp:
- Added `ChannelDescriptionChangedEvent`

Problem statement: 

OHC framework supports dynamic `StateDescription`s and dynamic `CommandDescription`s which are set by bindings and may change during runtime. When the state options / command options are changed the UI doesn't update automatically. You need to do a page refresh to pick up the changes. This approach introduces a new event (`ChannelDescriptionChangedEvent`) which UIs can subscribe to and reload related widgets. The event payload contains the `ChannelUID`, a field identifier to give a hint which field has changed and a list of linked item names.

See e.g. https://github.com/openhab/openhab-addons/pull/7396#issuecomment-630773749

Closes #1753 

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/AbstractDynamicDescriptionProvider.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/AbstractDynamicDescriptionProvider.java
@@ -12,16 +12,11 @@
  */
 package org.openhab.core.thing.binding;
 
-import java.util.Set;
-
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.events.Event;
 import org.openhab.core.events.EventPublisher;
-import org.openhab.core.thing.ChannelUID;
 import org.openhab.core.thing.events.ChannelDescriptionChangedEvent;
-import org.openhab.core.thing.events.ChannelDescriptionChangedEvent.CommonChannelDescriptionField;
-import org.openhab.core.thing.events.ThingEventFactory;
 import org.openhab.core.thing.i18n.ChannelTypeI18nLocalizationService;
 import org.openhab.core.thing.link.ItemChannelLinkRegistry;
 import org.osgi.framework.BundleContext;
@@ -52,22 +47,6 @@ public abstract class AbstractDynamicDescriptionProvider {
     protected @Nullable EventPublisher eventPublisher;
     protected @Nullable ChannelTypeI18nLocalizationService channelTypeI18nLocalizationService;
     protected @Nullable ItemChannelLinkRegistry itemChannelLinkRegistry;
-
-    /**
-     * This method can be used in a subclass in order to create a {@link ChannelDescriptionChangedEvent} and post it
-     * through the openHAB events bus.
-     *
-     * @param field the changed field
-     * @param channelUID the {@link ChannelUID}
-     * @param value the new value
-     * @param oldValue the old value
-     */
-    protected void postChannelDescriptionChangedEvent(CommonChannelDescriptionField field, ChannelUID channelUID,
-            Object value, @Nullable Object oldValue) {
-        postEvent(ThingEventFactory.createChannelDescriptionChangedEvent(field, channelUID,
-                itemChannelLinkRegistry != null ? itemChannelLinkRegistry.getLinkedItemNames(channelUID) : Set.of(),
-                value, oldValue));
-    }
 
     /**
      * This method can be used in a subclass in order to post events through the openHAB events bus. A common use case
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/AbstractDynamicDescriptionProvider.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/AbstractDynamicDescriptionProvider.java
@@ -20,6 +20,7 @@ import org.openhab.core.events.Event;
 import org.openhab.core.events.EventPublisher;
 import org.openhab.core.thing.ChannelUID;
 import org.openhab.core.thing.events.ChannelDescriptionChangedEvent;
+import org.openhab.core.thing.events.ChannelDescriptionChangedEvent.CommonChannelDescriptionField;
 import org.openhab.core.thing.events.ThingEventFactory;
 import org.openhab.core.thing.i18n.ChannelTypeI18nLocalizationService;
 import org.openhab.core.thing.link.ItemChannelLinkRegistry;
@@ -59,9 +60,11 @@ public abstract class AbstractDynamicDescriptionProvider {
      * @param field the changed field
      * @param channelUID the {@link ChannelUID}
      */
-    protected void postChannelDescriptionChangedEvent(String field, ChannelUID channelUID) {
+    protected void postChannelDescriptionChangedEvent(CommonChannelDescriptionField field, ChannelUID channelUID,
+            Object value, @Nullable Object oldValue) {
         postEvent(ThingEventFactory.createChannelDescriptionChangedEvent(field, channelUID,
-                itemChannelLinkRegistry != null ? itemChannelLinkRegistry.getLinkedItemNames(channelUID) : Set.of()));
+                itemChannelLinkRegistry != null ? itemChannelLinkRegistry.getLinkedItemNames(channelUID) : Set.of(),
+                value, oldValue));
     }
 
     /**
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/AbstractDynamicDescriptionProvider.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/AbstractDynamicDescriptionProvider.java
@@ -48,8 +48,8 @@ public abstract class AbstractDynamicDescriptionProvider {
     private final Logger logger = LoggerFactory.getLogger(getClass());
 
     protected @NonNullByDefault({}) BundleContext bundleContext;
-    protected @NonNullByDefault({}) EventPublisher eventPublisher;
-    protected @NonNullByDefault({}) ChannelTypeI18nLocalizationService channelTypeI18nLocalizationService;
+    protected @Nullable EventPublisher eventPublisher;
+    protected @Nullable ChannelTypeI18nLocalizationService channelTypeI18nLocalizationService;
     protected @Nullable ItemChannelLinkRegistry itemChannelLinkRegistry;
 
     /**
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/AbstractDynamicDescriptionProvider.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/AbstractDynamicDescriptionProvider.java
@@ -59,6 +59,8 @@ public abstract class AbstractDynamicDescriptionProvider {
      *
      * @param field the changed field
      * @param channelUID the {@link ChannelUID}
+     * @param value the new value
+     * @param oldValue the old value
      */
     protected void postChannelDescriptionChangedEvent(CommonChannelDescriptionField field, ChannelUID channelUID,
             Object value, @Nullable Object oldValue) {
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/AbstractDynamicDescriptionProvider.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/AbstractDynamicDescriptionProvider.java
@@ -1,0 +1,94 @@
+/**
+ * Copyright (c) 2010-2021 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.thing.binding;
+
+import java.util.Set;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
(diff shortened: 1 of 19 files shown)


bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/AbstractDynamicDescriptionProvider.java
```diff
@@ -0,0 +1,92 @@
...
+import org.openhab.core.thing.link.ItemChannelLinkRegistry;
+import org.osgi.framework.BundleContext;
+import org.osgi.service.component.ComponentContext;
+import org.osgi.service.component.annotations.Activate;
+import org.osgi.service.component.annotations.Deactivate;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The {@link AbstractDynamicDescriptionProvider} provides a base implementation for dynamic description provider.
```
cweitkamp, line 33:
```suggestion
 * The {@link AbstractDynamicDescriptionProvider} provides a base implementation for dynamic description provider.s
```

kaikreuzer:
I so far always considered those descriptions rather static and the "dynamic" providers mainly being a means to have the list defined at runtime and not at compile time (as with the xml files) - thus I never expected UIs to react upon changes.

If you think this is worthwhile to have, it should be fine to add it. You should just be clear of the implications: You should convince all UI devs to react to those events sooner or later and this includes Basic UI, the new Default UI as well as Android, iOS and Windows apps...

ghys:
I agree with @cdjackson that it's good to have the info no matter what's done with it.

A few notes however:
- Sitemap rendering UIs (i.e. mobile apps, Windows, Basic UI, etc.) listen to events on a dedicated SSE connection (`/rest/sitemaps/{sitemapname}/{pageid}`), not the "regular" event bus (`/rest/events`), Would this event appear in those?
- End user UIs usually deal with items, and don't care about channels, so even if the event contains the linked items for the channel, maybe it would be more natural to issue events for the affected item's themselves rather than the channel;
- The new "main" UI tries not to listen to the event bus or maintain a local copy of item descriptions for end-user pages (for performance reasons), but only tracks select items' states with `/rest/events/states`. Command options are currently displayed with "action sheets" like this by default:



In this case, since the options are not displayed until the user clicks on the widget, it would be safer just to make a quick call to `/rest/items/{itemName}` to get the up-to-date command options before displaying the actions. I suppose there will be a need for alternative rendering of these command options (e.g. with buttons when their number is limited), then of course it's another story. Nonetheless, maybe it would be worth looking into adding state descriptions/command descriptions and their updates to the [StateDTO](https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/internal/dto/StateDTO.java) the UI receives whenever an item tracked with `/rest/events/states` changes.

cweitkamp:
Thanks for you useful input. The pointer to SSE events was a good hint. It obviously makes a lot of sense to emit the `ChannelDescriptionChangedEvent` via them too. I will look into that and try to enhance my current implementation to add support for it.

mueller-ma:
> Sitemap rendering UIs (i.e. mobile apps, Windows, Basic UI, etc.) listen to events on a dedicated SSE connection (/rest/sitemaps/{sitemapname}/{pageid}), not the "regular" event bus (/rest/events), Would this event appear in those?

Ideally a SSE will be sent to all affected sitemap elements as well.

/cc @maniac103

lolodomo:
> Thanks for you useful input. The pointer to SSE events was a good hint. It obviously makes a lot of sense to emit the `ChannelDescriptionChangedEvent` via them too. I will look into that and try to enhance my current implementation to add support for it.

For the remote openHAB binding, an event on the topic openhab/items/*/xxx would be ideal for me. But if I have to listen to another topic, why not.
I have imagined it could have be done through the existing event ItemUpdatedEvent.

Do you plan to rework and finish this enhancement ? Before OH3 is released ?

cweitkamp:
> Do you plan to rework and finish this enhancement ? Before OH3 is released ?

I added it to the [openHAB 3 issue tracking](https://github.com/orgs/openhab/projects/3). So yes, I am planning to do further work on it.

lolodomo:
Any progress on this enhancement?

cweitkamp:
I rebased the branch onto latest main. From my POV the PR is in a final state and ready for review.

grafik

I have to admit that I was not able to look deeper into SSE connections. Feel free to give me a hand on that.

lolodomo:
Does it mean that any binding state options provider should then be updated to set the reference to ItemChannelLinkRegistry?

lolodomo:
I think we should publish an event to the item topic too.

cweitkamp:
> Does it mean that any binding state options provider should then be updated to set the references to ItemChannelLinkRegistry and EventPublisher?

AFAIR yes. At least when we want to use constructor injection. I did not test if field injection is working when using abstract base classes. Method injection did not work for it in the past.

lolodomo:
Isn't it possible to provide the new state description in the new event? This would avoid additional request(s) to get it.

cweitkamp:
Sounds like a reasonable idea.

lolodomo:
I think there is no reason to force 
 channelTypeI18nLocalizationService and eventPublisher as not null by default in the abstract class. In fact, they will be null if a binding provider does not set them.

cweitkamp:
@lolodomo Thanks for placing your ideas. I tested them one by one.

I am afraid we cannot use field injection or method injection in abstract classes. It looks like OSGi framework only considers `@Reference` annotations in the final class annotated with `@Component`. We have to adopt all existing implementations to support this new event.

> Isn't it possible to provide the new state description in the new event?

No, that is not possible. It is not available and we do not have access to the full state / command description in the provider classes. We only can add a partial state description which contains only the changed field.

lolodomo:
You could at least provide the new options in the event (or the new pattern).

cweitkamp:
That is possible. Looking into other events for updating an entity (e.g. Thing or Item) we publish the old entity beside the new one.

lolodomo:
Yes, that's true. Is there any advantage to provide the old list of options in addition to the new list of options, I am not sure. But if you want to provide the new AND old list of options, that's fine for me.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/events/ChannelDescriptionChangedEvent.java
```diff
@@ -0,0 +1,147 @@
...
+ * {@link CommandDescription} or {@link StateDescription} of a channel has changed. Instances must be created with the
+ * {@link ThingEventFactory}.
+ *
+ * @author Christoph Weitkamp - Initial contribution
+ */
+@NonNullByDefault
+public class ChannelDescriptionChangedEvent extends AbstractEvent {
+
+    public enum CommonChannelDescriptionField {
+        ALL,
```
lolodomo, line 36: Is ALL necessary (not used I believe) ?

bundles/org.openhab.core.thing/src/test/java/org/openhab/core/thing/binding/BaseDynamicCommandDescriptionProviderTest.java
```diff
@@ -0,0 +1,94 @@
...
+        subject.setCommandOptions(CHANNEL_UID, List.of(new CommandOption("reboot", "Reboot")));
+
+        ArgumentCaptor<Event> capture = ArgumentCaptor.forClass(Event.class);
+        verify(mockEventPublisher, times(1)).post(capture.capture());
+
+        Event event = capture.getValue();
+        assertTrue(event instanceof ChannelDescriptionChangedEvent);
+        ChannelDescriptionChangedEvent cdce = (ChannelDescriptionChangedEvent) event;
+        assertEquals(CommonChannelDescriptionField.COMMAND_OPTIONS, cdce.getField());
+
```
lolodomo, line 88: Suggestion: you could additionally checked the linked items and the list of options in the event.

bundles/org.openhab.core.thing/src/test/java/org/openhab/core/thing/binding/BaseDynamicStateDescriptionProviderTest.java
```diff
@@ -0,0 +1,112 @@
...
+        subject.setStateOptions(CHANNEL_UID, List.of(new StateOption("offline", "Offline")));
+
+        ArgumentCaptor<Event> capture = ArgumentCaptor.forClass(Event.class);
+        verify(mockEventPublisher, times(1)).post(capture.capture());
+
+        Event event = capture.getValue();
+        assertTrue(event instanceof ChannelDescriptionChangedEvent);
+        ChannelDescriptionChangedEvent cdce = (ChannelDescriptionChangedEvent) event;
+        assertEquals(CommonChannelDescriptionField.STATE_OPTIONS, cdce.getField());
+
```
lolodomo, line 106: Suggestion: you could additionally checked the linked items and the list of options in the event.

bundles/org.openhab.core.thing/src/test/java/org/openhab/core/thing/binding/BaseDynamicStateDescriptionProviderTest.java
```diff
@@ -0,0 +1,112 @@
...
+        subject.setStatePattern(CHANNEL_UID, "%s");
+
+        ArgumentCaptor<Event> capture = ArgumentCaptor.forClass(Event.class);
+        verify(mockEventPublisher, times(1)).post(capture.capture());
+
+        Event event = capture.getValue();
+        assertTrue(event instanceof ChannelDescriptionChangedEvent);
+        ChannelDescriptionChangedEvent cdce = (ChannelDescriptionChangedEvent) event;
+        assertEquals(CommonChannelDescriptionField.PATTERN, cdce.getField());
+
```
lolodomo, line 88: Suggestion: you could additionally checked the linked items and the pattern in the event.

lolodomo (changes requested):
Few remarks and suggestions

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/AbstractDynamicDescriptionProvider.java
```diff
@@ -0,0 +1,97 @@
...
+    protected @Nullable EventPublisher eventPublisher;
+    protected @Nullable ChannelTypeI18nLocalizationService channelTypeI18nLocalizationService;
+    protected @Nullable ItemChannelLinkRegistry itemChannelLinkRegistry;
+
+    /**
+     * This method can be used in a subclass in order to create a {@link ChannelDescriptionChangedEvent} and post it
+     * through the openHAB events bus.
+     *
+     * @param field the changed field
+     * @param channelUID the {@link ChannelUID}
```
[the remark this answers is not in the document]
cweitkamp: Done.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/events/ChannelDescriptionChangedEvent.java
cweitkamp: It will be needed if a binding publishes the whole `StateDEscription` (e.g. https://github.com/openhab/openhab-addons/pull/10718). Note: I have to change PRs in openHAB add-ons according the new API.

bundles/org.openhab.core.thing/src/test/java/org/openhab/core/thing/binding/BaseDynamicCommandDescriptionProviderTest.java
cweitkamp: The `ThingEventFactoryTest` class is responsible for testing the correct payloads.

lolodomo (approved):
LGTW

lolodomo:
Delay is now short before 3.1 is released. Can this PR be merged ASAP. We have then to enhance 33 bindings (some PRs are already submitted waiting for the merge of this one).

lolodomo (approved):

kaikreuzer (approved):
lgtm, thanks!

lolodomo:
Very good.
Can you please trigger a new core build so that we can have this enhancement in the next snapshot ? I need it ot test my changes in the remoteopenhab binding.

Regarding impact on UI (sitemap), I know how to handle that. First a little change is needed in the IO REST sitemap bundle: my idea is to add a boolean field to the widget SSE event to indicate if the item description was changed. This will have no impact on UIs not reading this field. In Basic UI, I will read this field and when set to true, I will trigger the reload of the current page. Nothing very difficult I believe, I just need to validate that it works as I imagine.

kaikreuzer:
@lolodomo I had triggered a build right after merging this PR: https://ci.openhab.org/job/openHAB-Core/793/

lolodomo:
Perfect. 

Note that we have already 3 PRs for addons ready to be merged that enhance bindings with this new feature. 
@cweitkamp is working on the PR for the hue  binding. 
I will try to create another PR covering the remaining bindings this evening.

lolodomo:
@cweitkamp : what is smarthome/j  ?

cweitkamp:
A third party repository for OHC based add-ons.

lolodomo:
I identified 3 bindings implementing a state description provider extending BaseDynamicStateDescriptionProvider in an unusual way: spotify, bticinosmarther and heos. These bindings require a special look. All the other (except hue) are now updated.

And for the 3 bindings implementing a command description provider extending BaseDynamicCommandDescriptionProvider, only the nanoleaf binding is not yet enhanced.

lolodomo:
@cweitkamp : I just discovered we have many bindings directly implementing DynamicStateDescriptionProvider. Those ones will not benefit this enhancement?

cweitkamp:
Yes, that is correct. They have to take care on their own.

PR #1506, opened 2020-05-31
Ensure managed provider is unset by registered provider

cdjackson:
This ensures that the managed provider is unregistered by the same provider that is registered. Otherwise it's possible for the provider to be inadvertently removed.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/common/registry/AbstractRegistry.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/common/registry/AbstractRegistry.java
@@ -48,8 +48,11 @@ import org.slf4j.LoggerFactory;
  * @author Kai Kreuzer - switched to parameterized logging
  * @author Hilbrand Bouwkamp - Made protected fields private and added new methods to give access.
  * @author Markus Rathgeb - Use separate collections to improve performance
+ * @author Chris Jackson - Ensure managed provider is only unset by current provider
  *
  * @param <E> type of the element
+ * @param <K> type of the key
+ * @param <P> type of the provider
  */
 @NonNullByDefault
 public abstract class AbstractRegistry<@NonNull E extends Identifiable<K>, @NonNull K, @NonNull P extends Provider<E>>
@@ -530,7 +533,9 @@ public abstract class AbstractRegistry<@NonNull E extends Identifiable<K>, @NonN
     }
 
     protected void unsetManagedProvider(ManagedProvider<E, K> provider) {
-        managedProvider = Optional.empty();
+        if (managedProvider.equals(provider)) {
+            managedProvider = Optional.empty();
+        }
     }
 
     /**


kaikreuzer:
Could you give a situation where this could possibly happen?
Afaik, we never do such a check in any unset method we have throughout the codebase, so why here?

cdjackson:
If there were to be two managed providers in a system (and I agree that would not be a normal case) then it would be possible for one to be registered, and the removal of a second one to remove the first - thus breaking the system.

It can certainly happen.

kaikreuzer:
A million things would break if there were ever two managed providers for the same registry available in the system. This is simple nothing that is supported in openHAB. The same holds true for ANY 1..1 relation between services that we have in place.

cdjackson:
But this is a base implementation for a registry - surely it should protect against users creating two managed providers?

Just because OH doesn't protect against this sort of thing elsewhere doesn't mean it shouldn't, and doesn't mean that in such an abstract base class, which can be used for additional functions it's not a good idea.

The system here allows multiple providers, but assumes there is just one that is "managed". In my opinion, there should be some consistency check here, but ok, if you don't think this is a useful check for openHAB, then please feel free to close this.

kaikreuzer:
> surely it should protect against users creating two managed providers?

If this is the intention, we should check whether multiple managed providers are registered and log a big fat warning to the log.

My concern with this PR here is that it might give the impression that having multiple managed providers is a supported and accepted situation. But as mentioned, I am very sure that it would lead to a lot of weird, buggy and unexpected behavior and should thus be avoided in any case.

So unless you have a real world situation where you came across an issue that would be fixed by this PR, I would indeed prefer to just close it.

cdjackson:
Yes, I have a real world situation that is fixed by this - I wasn’t just looking through the code looking for things to change ;)

I’m happy to add a warning as well if you like, or I could look to fix it in the extended class - it just seemed like a good idea to fix it at source and I personally don’t think it hurts.

kaikreuzer:
> Yes, I have a real world situation that is fixed by this

Which brings me back to my very first question above: Can you explain how such a situation looks like?
Our aim must be to prevent such situations from occurring, so understanding them will help to define the best countermeasure.

cdjackson:
Sure. As I said, if someone implements a provider based on managed provider, and another similar provider is added, then the system can't cope with this. Sure - I know it's not ideal anyway, but don't you think it still makes sense to add the simple check to ensure that we make the best of the situation?  The system would work reasonably well otherwise.

openHAB is meant to be a framework on which systems can be built. This is a base class on which other registries can be built, so IMHO it should be able to cope with situations that are not standard in OH.

It's a small change and probably not worth spending too much time on. If you don't agree its useful, then feel free to close this and I'll take care of this elsewhere (in the extended class).

kaikreuzer:
> If you don't agree its useful, then feel free to close this

I still try to understand why you think it is useful, as I have the impression that I am not yet getting the point.

> The system would work reasonably well otherwise.

You seem to imply that there are any legit situations to have multiple managed providers.
Is the situation that you would want to switch one managed provider against another one during runtime? In that case, I'd see how it would indeed fix a misbehavior.

I also agree that it isn't worth to spend too much time on the discussion, so let's merge.

cdjackson:
Thanks Kai. No - I don't at all mean to imply this is a legit situation - I said earlier that it's not a normal condition, and I certainly consider it an error. However it still seems reasonable for the framework to try and act appropriately where it can (to me at least).

PR #1508, opened 2020-05-31
Added "editable" info to rules in REST API

kaikreuzer:
This was part of https://github.com/openhab/openhab-core/pull/1451 and has been separated as an independent PR that can already been reviewed and merged independently.

### DIFF
--- a/bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/RuleResource.java
+++ b/bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/RuleResource.java
@@ -71,8 +71,6 @@ import org.openhab.core.io.rest.RESTResource;
 import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Reference;
-import org.osgi.service.component.annotations.ReferenceCardinality;
-import org.osgi.service.component.annotations.ReferencePolicy;
 import org.osgi.service.jaxrs.whiteboard.JaxrsWhiteboardConstants;
 import org.osgi.service.jaxrs.whiteboard.propertytypes.JSONRequired;
 import org.osgi.service.jaxrs.whiteboard.propertytypes.JaxrsApplicationSelect;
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/RuleResource.java
```diff
@@ -121,6 +126,15 @@ public RuleResource( //
         this.ruleRegistry = ruleRegistry;
     }
 
+    @Reference(cardinality = ReferenceCardinality.OPTIONAL, policy = ReferencePolicy.DYNAMIC)
```
cweitkamp, line 129: Is this really an optional reference? Isn't it mandatory and we can pass it via constructor injection?

bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/dto/EnrichedRuleDTOMapper.java
```diff
@@ -20,13 +21,16 @@
...
  */
 public class EnrichedRuleDTOMapper extends RuleDTOMapper {
 
-    public static EnrichedRuleDTO map(final Rule rule, final RuleManager ruleEngine) {
+    public static EnrichedRuleDTO map(final Rule rule, final RuleManager ruleEngine,
+            final ManagedRuleProvider managedRuleProvider) {
         final EnrichedRuleDTO enrichedRuleDto = new EnrichedRuleDTO();
         fillProperties(rule, enrichedRuleDto);
         enrichedRuleDto.status = ruleEngine.getStatusInfo(rule.getUID());
+        enrichedRuleDto.editable = managedRuleProvider.get(enrichedRuleDto.uid) != null;
```
cweitkamp, line 33: If the reference stays optional I prefer to see a `null`-check on the `managedRuleProvider` here.

```suggestion
        enrichedRuleDto.editable = managedRuleProvider.get(rule.getUID()) != null;
```

bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/RuleResource.java
kaikreuzer: Agree, can be mandatory.

kaikreuzer:

Issue #1472, opened 2020-05-13
mDNS records contain wrong hostname


HeMeZ:
I had some issues with the resolution of the hostname "openhab" on different machines.
After a lot of testing I found out that the mDNS records contain `JmDNS-/192-168-0-123` instead of the hostname.
I could narrow it down to this line of code:
https://github.com/openhab/openhab-core/blob/3811cb95604f69f6a3d77da2caf391d81b05458e/bundles/org.openhab.core.io.transport.mdns/src/main/java/org/openhab/core/io/transport/mdns/internal/MDNSClientImpl.java#L269
I would propose to replace `"JmDNS-" + address.toString()` with either `address.getHostName()` or with the value of mdnsName from `custom.system.properties`.
Which is a concept decision I cannot make.

kaikreuzer:
Alternatively, we could simply set this parameter to `null`, which means that JmDNS applies the following logic according to its JavaDoc:
```
If name parameter is null will use the hostname. The hostname is determined by the following algorithm:
- Get the hostname from the InetAdress obtained before.
- If the hostname is a reverse lookup default to JmDNS name or computer if null.
- If the name contains '.' replace them by '-'
- Add .local. at the end of the name.
```

PR #1510, opened 2020-06-01
Use hostname as JmDNS client name

kaikreuzer:
Fixes https://github.com/openhab/openhab-core/issues/1472


kaikreuzer:
Yes, this is backwards compatible - nobody outside should rely on a specific hostname for the mDNS announcements.

Issue #865, opened 2019-06-10
Items not found when displaying sitemap or executing rules


mhilbush:
I just installed build 1607 on a system that had been running 1502 for quite a while.

After an openHAB restart, sometimes, when displaying a sitemap, I see errors in my log indicating that items cannot be found, even though the items exist. 

I've found each of the following things will clear the issue:
- sometimes a restart will clear the issue
- modifying an items file will clear the issue for the items defined in that file
- stopping/starting `org.openhab.core.model.item` also resolves the issue

Could there be a timing issue where a dependent bundle might not be available at the right time during startup?

Example of errors.
```
2019-06-10 13:44:57.912 [ERROR] [ui.internal.items.ItemUIRegistryImpl] - Cannot retrieve item for widget org.eclipse.smarthome.model.sitemap.Text
2019-06-10 13:44:57.914 [ERROR] [ui.internal.items.ItemUIRegistryImpl] - Cannot retrieve visibility item OH_Status for widget org.eclipse.smarthome.model.sitemap.Text
2019-06-10 13:44:57.915 [ERROR] [ui.internal.items.ItemUIRegistryImpl] - Cannot retrieve item 'OH_Status' for widget org.eclipse.smarthome.model.sitemap.Text
2019-06-10 13:44:57.916 [ERROR] [ui.internal.items.ItemUIRegistryImpl] - Cannot retrieve item 'OH_Status' for widget org.eclipse.smarthome.model.sitemap.Text
2019-06-10 13:44:57.917 [ERROR] [ui.internal.items.ItemUIRegistryImpl] - Cannot retrieve item 'OH_Status' for widget org.eclipse.smarthome.model.sitemap.Text
2019-06-10 13:45:50.191 [ERROR] [ui.internal.items.ItemUIRegistryImpl] - Cannot retrieve item for widget org.eclipse.smarthome.model.sitemap.Text
2019-06-10 13:45:50.193 [ERROR] [ui.internal.items.ItemUIRegistryImpl] - Cannot retrieve item 'Uptime_Mark_Formatted' for widget org.eclipse.smarthome.model.sitemap.Text
2019-06-10 13:45:50.194 [ERROR] [ui.internal.items.ItemUIRegistryImpl] - Cannot retrieve item 'Uptime_Mark_Formatted' for widget org.eclipse.smarthome.model.sitemap.Text
[...]

PR #1514, opened 2020-06-07
[dsl] Model loading improvements

kaikreuzer:
Some further changes extracted from my work in #1451:
- removed superfluous null checks
- improved logging
- fixed bug that cause "_tmp" resources to be loaded (which are only internally temporarily created during model validation)
- added ready marker after finishing loading a model type
- let GenericItemProvider only start once at least one ItemFactory is available (I often had the effect in the IDE that items were not available as the file was loaded before the CoreItemFactory was available).

### DIFF
--- a/bundles/org.openhab.core.model.core/src/main/java/org/openhab/core/model/core/internal/ModelRepositoryImpl.java
+++ b/bundles/org.openhab.core.model.core/src/main/java/org/openhab/core/model/core/internal/ModelRepositoryImpl.java
@@ -98,20 +98,18 @@ public class ModelRepositoryImpl implements ModelRepository {
 
     @Override
     public boolean addOrRefreshModel(String name, final InputStream originalInputStream) {
+        logger.info("Loading model '{}'", name);
         Resource resource = null;
         InputStream inputStream = null;
         try {
-            if (originalInputStream != null) {
-                byte[] bytes = originalInputStream.readAllBytes();
-                String validationResult = validateModel(name, new ByteArrayInputStream(bytes));
-                if (validationResult != null) {
-                    logger.warn("Configuration model '{}' has errors, therefore ignoring it: {}", name,
-                            validationResult);
-                    removeModel(name);
-                    return false;
-                }
-                inputStream = new ByteArrayInputStream(bytes);
+            byte[] bytes = originalInputStream.readAllBytes();
+            String validationResult = validateModel(name, new ByteArrayInputStream(bytes));
+            if (validationResult != null) {
+                logger.warn("Configuration model '{}' has errors, therefore ignoring it: {}", name, validationResult);
+                removeModel(name);
+                return false;
             }
+            inputStream = new ByteArrayInputStream(bytes);
             resource = getResource(name);
             if (resource == null) {
                 synchronized (resourceSet) {
@@ -123,13 +121,6 @@ public class ModelRepositoryImpl implements ModelRepository {
                         Resource.Factory.Registry.INSTANCE.getExtensionToFactoryMap().remove("*");
                         resource = resourceSet.createResource(URI.createURI(name));
                         if (resource != null) {
-                            logger.info("Loading model '{}'", name);
-                            if (inputStream == null) {
-                                logger.warn(
-                                        "Resource '{}' not found. You have to pass an inputStream to create the resource.",
-                                        name);
-                                return false;
-                            }
                             resource.load(inputStream, resourceOptions);
                             notifyListeners(name, EventType.ADDED);
                             return true;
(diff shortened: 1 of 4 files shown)


mhilbush:
> let GenericItemProvider only start once at least one ItemFactory is available (I often had the effect in the IDE that items were not available as the file was loaded before the CoreItemFactory was available).

Hmm. This is interesting. There have been numerous cases of items not being loaded at startup. There's an open issue for it; I'll try to find it.

Do you think this might be the cause of that?

mhilbush:
Here's the issue. WDYT?

https://github.com/openhab/openhab-core/issues/865

kaikreuzer:
@mhilbush Might well be possible, but it is difficult to tell if it is (just this) cause.

bundles/org.openhab.core.model.core/src/main/java/org/openhab/core/model/core/internal/folder/FolderObserver.java
```diff
@@ -60,6 +62,12 @@
 
     /* the model repository is provided as a service */
     private final ModelRepository modelRepository;
+    private static final String READYMARKER_TYPE = "dsl";
+
+    @Reference
+    /* default */ ReadyService readyService;
```
cweitkamp, line 68: The `FolderObserver` already uses constructor injection? Can we use it for the `ReadyService` too?

bundles/org.openhab.core.model.core/src/main/java/org/openhab/core/model/core/internal/ModelRepositoryImpl.java
```diff
@@ -203,8 +189,9 @@ public void reloadAllModelsOfType(final String modelType) {
             // Make a copy to avoid ConcurrentModificationException
             List<Resource> resourceListCopy = new ArrayList<>(resourceSet.getResources());
             for (Resource resource : resourceListCopy) {
-                if (resource != null && resource.getURI().lastSegment().contains(".") && resource.isLoaded()) {
-                    if (modelType.equalsIgnoreCase(resource.getURI().fileExtension())) {
+                if (resource.getURI().lastSegment().contains(".") && resource.isLoaded()) {
+                    if (modelType.equalsIgnoreCase(resource.getURI().fileExtension())
+                            && !resource.getURI().lastSegment().startsWith("tmp_")) {
```
cweitkamp, line 194: Combine nested `if`s and consider new condition in method `removeAllModelsOfType()`.

bundles/org.openhab.core.model.core/src/main/java/org/openhab/core/model/core/internal/ModelRepositoryImpl.java
```diff
@@ -98,20 +98,18 @@ public ModelRepositoryImpl(final @Reference SafeEMF safeEmf) {
...
-                }
-                inputStream = new ByteArrayInputStream(bytes);
+            byte[] bytes = originalInputStream.readAllBytes();
+            String validationResult = validateModel(name, new ByteArrayInputStream(bytes));
+            if (validationResult != null) {
+                logger.warn("Configuration model '{}' has errors, therefore ignoring it: {}", name, validationResult);
+                removeModel(name);
+                return false;
             }
+            inputStream = new ByteArrayInputStream(bytes);
```
cweitkamp, line 112: This is a nice improvement. With absence of these ugly nestings we can use a try-with-resources statement on the `inputStream` declaration and get rid of the `finally` introduced in #1423.

kaikreuzer:
Thanks @cweitkamp, I have addressed all your comments.

bundles/org.openhab.core.model.core/src/main/java/org/openhab/core/model/core/internal/ModelRepositoryImpl.java
```diff
@@ -98,20 +98,22 @@ public ModelRepositoryImpl(final @Reference SafeEMF safeEmf) {
...
-                String validationResult = validateModel(name, new ByteArrayInputStream(bytes));
-                if (validationResult != null) {
-                    logger.warn("Configuration model '{}' has errors, therefore ignoring it: {}", name,
-                            validationResult);
-                    removeModel(name);
-                    return false;
-                }
-                inputStream = new ByteArrayInputStream(bytes);
+        byte[] bytes = null;
+        try (InputStream inputStream = originalInputStream) {
```
cweitkamp, line 104: ftr: This is probably not necessary as the `FolderObserver` handles the passed `InputStream` (`originalInputStream`), but I do not think that it will hurt:

https://github.com/openhab/openhab-core/blob/b7ab8070789079a9a9cbd1091ea7b9152f1a04e1/bundles/org.openhab.core.model.core/src/main/java/org/openhab/core/model/core/internal/folder/FolderObserver.java#L233-L238

kaikreuzer: True, I didn't see that. But imho it is better to explicitly make sure that the stream is closed within the method. One should not rely on the caller to do it.

PR #1517, opened 2020-06-08
[dsl] Allow negative numbers in metadata for Items

### LINKED ISSUE #1501: 2.5.3 Negative numbers not supported in items configuration
OpenHAB 2.5.3
Homekit binding

i would like to make min & max temperature configurable for a thermostat item (homekit binding)
following line works

`Number 	temperature  {homekit="TemperatureSensor" [minValue=10, maxValue=27] }
`
but if i want to use negative numbers i get parsing error. 
this line

`Number 	temperature  {homekit="TemperatureSensor" [minValue=-10, maxValue=27] }
`

leads to this exception

`2020-05-25 14:03:13.852 [WARN ] [el.core.internal.ModelRepositoryImpl] - Configuration model 'test.items' has errors, therefore ignoring it: [8,106]: no viable alternative at input '-'`

is it known limitation? or is there a way to provide negative numbers

cweitkamp:
- Allow negative numbers in metadata for Items

What is the best solution:

> If we compare the grammar to the one for Things there is exactly the "-" sign missing.
> 
> https://github.com/openhab/openhab-core/blob/1f4e374ece9d8d72f5bc77d647152b9430ca7524/bundles/org.openhab.core.model.thing/src/org/openhab/core/model/thing/Thing.xtext#L96
> 
> We have options to solve it.
> 
> 1. Add the "-" sign to the terminal `ID` in the Items grammar. This could lead to an impact on the Item names because the terminal `ID` is used to define allowed characters for it too.
> 2. Add an optional "-" sign for the `NUMBER` definition like this
> 
> ```
> NUMBER returns ecore::EBigDecimal:
>     ('-')? ID ('.' ID )?
> ;
> ```

Closes #1501

### DIFF
--- a/bundles/org.openhab.core.model.item/src/org/openhab/core/model/Items.xtext
+++ b/bundles/org.openhab.core.model.item/src/org/openhab/core/model/Items.xtext
@@ -9,11 +9,11 @@ import "http://www.eclipse.org/emf/2002/Ecore" as ecore
 
 generate items "https://openhab.org/model/Items"
 
-ItemModel :
+ItemModel:
 	{ItemModel} (items+=ModelItem)*
 ;
 
-ModelItem :
+ModelItem:
 	(ModelNormalItem | ModelGroupItem) name=ID
 	(label=STRING)?
 	('<' icon=(ID|STRING) '>')?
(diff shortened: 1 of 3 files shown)


kaikreuzer:
Do I see it correctly that you have now applied BOTH mentioned options in this PR? Could you briefly explain the reason for this?

cweitkamp:
Yes, you do. As I am not sure which is the best option I applied both to be open for discussions. I would prefer option 2 (optional "-" sign for the `NUMBER` definition). Maybe not only for items but for things too.

cweitkamp:
I changed it now according to my preferences.

PR #1541, opened 2020-07-09
Remove deprecated method of 'ConfigOptionProvider'

cweitkamp:
- Remove deprecated method of `ConfigOptionProvider`

Related to #1408

---

Depends on https://github.com/openhab/openhab-addons/pull/8093

### DIFF
--- a/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/ConfigOptionProvider.java
+++ b/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/ConfigOptionProvider.java
@@ -29,19 +29,11 @@ import org.eclipse.jdt.annotation.Nullable;
 @NonNullByDefault
 public interface ConfigOptionProvider {
 
-    /**
-     * Provides a collection of {@link ParameterOptions}s.
-     *
-     * @deprecated Use {@link getParameterOptions} with context instead.
-     *
-     * @param uri the uri of the config description
-     * @param param the parameter name for which the requested options shall be returned
-     * @param locale the locale in which the result is expected
-     * @return the configuration options provided by this provider if any or {@code null} otherwise
-     */
-    @Deprecated
-    @Nullable
-    Collection<ParameterOption> getParameterOptions(URI uri, String param, @Nullable Locale locale);
+    // @Deprecated
+    // @Nullable
+    // default Collection<ParameterOption> getParameterOptions(URI uri, String param, @Nullable Locale locale) {
+    // return getParameterOptions(uri, param, null, locale);
+    // }
 
     /**
      * Provides a collection of {@link ParameterOptions}s.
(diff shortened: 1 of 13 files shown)


bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/ConfigOptionProvider.java
```diff
@@ -29,19 +29,11 @@
...
-     * @param param the parameter name for which the requested options shall be returned
-     * @param locale the locale in which the result is expected
-     * @return the configuration options provided by this provider if any or {@code null} otherwise
-     */
-    @Deprecated
-    @Nullable
-    Collection<ParameterOption> getParameterOptions(URI uri, String param, @Nullable Locale locale);
+    // @Deprecated
+    // @Nullable
+    // default Collection<ParameterOption> getParameterOptions(URI uri, String param, @Nullable Locale locale) {
```
cweitkamp, line 34: This method needs a little attention. If we remove it now it will break the openHAB 3 Add-ons build.

kaikreuzer: So if I get it right, your plan now is to keep it as a deprecated "default" method in place and remove it later?
Which add-ons will break?

cweitkamp: Yes, that was the initial idea. We are talking about three bindings:

- Amazon Dash Button
- GPSTracker
- SmartMeter

There are two other options:

1. We could exclude them from openHAB 3 Add-ons build.

or

2. We could move them to "master"-branch and migrate them right now.

kaikreuzer: > We are talking about three bindings

Shouldn't it be fairly easy to adapt them?

cweitkamp: > Shouldn't it be fairly easy to adapt them?

Yes, it is. But we cannot use the same code to compile against OHC 2.5 and OHC 3 (after this PR got merged). If we remove the deprecated method in the bindings it will fail on OH 2.5.x Add-ons build and if we remove the deprecated method here it will fail on OH 3 Add-ons build.

kaikreuzer: Ok, I see. I had missed the fact that it is a "provider" interface and not a "consumer" interface.
So yeah, I think your approach is just fine - let's leave it in as a default method and remove it once we move all add-ons to master. Do you keep track of those open changes somewhere?

cweitkamp: We keep track of it in https://github.com/openhab/openhab-core/issues/1408. I added a note.

PR #1543, opened 2020-07-11
Use the same regex as the install and uninstall method

silamon:
The getById method is missing the ":". See the install and uninstall method.

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/extensions/ExtensionResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/extensions/ExtensionResource.java
@@ -138,7 +138,7 @@ public class ExtensionResource implements RESTResource {
     }
 
     @GET
-    @Path("/{extensionId: [a-zA-Z_0-9-]+}")
+    @Path("/{extensionId: [a-zA-Z_0-9-:]+}")
     @Produces(MediaType.APPLICATION_JSON)
     @ApiOperation(value = "Get extension with given ID.")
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/extensions/ExtensionResource.java
```diff
@@ -138,7 +138,7 @@ public Response getTypes(@HeaderParam("Accept-Language") @ApiParam(value = "lang
     }
 
     @GET
-    @Path("/{extensionId: [a-zA-Z_0-9-]+}")
+    @Path("/{extensionId: [a-zA-Z_0-9-:]+}")
```
cweitkamp, line 141: I am a fan of streamlining it but not sure if we should add the colon here or rather remove it from install/uninstall methods.

@kaikreuzer You introduced the colon in https://github.com/eclipse-archived/smarthome/pull/3076 because the marketplace adds `marcetplace:` prefix. As the Marketplace itself is EOL. What is your opinion on this?

kaikreuzer: > As the Marketplace itself is EOL

I would not say so. What we will drop is the [Eclipse IoT Marketplace extension service](https://github.com/openhab/openhab-addons/tree/2.5.x/bundles/org.openhab.extensionservice.marketplace), but not the general marketplace extension mechanism (see https://github.com/openhab/openhab-addons/issues/6673).
We will therefore need a way to differentiate between "distro" add-ons and stuff that comes from other (external) sources. And using colons for such namespaces imho still makes sense, so I would keep it in place.

PR #1546, opened 2020-07-14
[rest] introduce nickname attribute in ConfigDescriptionResource to make swaggers operationId unique

pavog:
By default the Swagger plugin uses the method name as operationId.
Since the operationId must be unique across all methods and some method names are named the same, we should use the nickname element in the ApiOperation annotation and thus set the operationId manually.

The `getAll()` method in `ConfigDescriptionResource.java` does not have a nickname set in its `@ApiOperation` annotation.
Swagger uses the method name (getAll) as default operationId which is not unique.

I will create more prs for some other classes that need to be fixed.

Reference:
 * [Nickname attribute](https://docs.swagger.io/swagger-core/v1.5.X/apidocs/io/swagger/annotations/ApiOperation.html#nickname--) in swagger annotations documentation
 * [Operation Object](https://swagger.io/specification/#operation-object) in swagger specification


pavog:
I don't know why the tests failed, but I'm pretty sure it's not my fault since all tests in my local test setup were successful.
Maybe someone can restart the tests / build?

PR #1559, opened 2020-07-21
[cleanup] removed boot bundle

kaikreuzer:
After the removal of the old rule engine, there isn't any need to keep the core.boot bundle in place.
It merely contains the icon forwarder, which is a 1.x compat feature as well as the default sitemap provider, which isn't really of any use.

As a side effect, I have deprecated the ConfigConstants class and moved its features to the OpenHAB class in the core bundle, which seems to be a better place.


ghys:
> As a side effect, I have deprecated the ConfigConstants class

Please also mind the references in ui and addons code:
https://github.com/openhab/openhab-webui/search?q=ConfigConstants&unscoped_q=ConfigConstants
https://github.com/openhab/openhab-addons/search?q=ConfigConstants&unscoped_q=ConfigConstants

After the OpenHAB class holds the config & version information, I guess we can probably start discussing about exposing it (or a serialized version of it) in the API's RootBean, so the UI can display the runtime's exact version & build number and maybe the location of the configuration directories as well - those might be considered a security threat, but on the other hand it could valuable when supporting users in figuring out where things are.

kaikreuzer:
@ghys We can only adapt the addons once they are on the master branch. That's why I only deprecated the class, but didn't remove it yet.

>  I guess we can probably start discussing about exposing it (or a serialized version of it) in the API's RootBean

Feel free to create a PR for those. We already showed the version+build in the old dashboard, so having it accessible in the new ui definitely makes sense.

PR #1560, opened 2020-07-21
[cleanup] refactored Extension to Addon

kaikreuzer:
The name "extension" was used in ESH and it precisely corresponds to an openHAB add-on.
To clean up with the naming, I have refactored the ExtensionService to become an AddonService instead.


kaikreuzer:
I missed the ExtensionResource class - will add a commit for it later today.

kaikreuzer:
Thanks - rebase done.

splatch:
I've did a basic scan of Java sources to determine if this change was really necessary. My scanner is basic shell command `find ~/projects/opnhab-core -name "*.java" -exec grep -i "addon" "{}" \;| wc -l`.

Here are results:
|revision | extension | addon |
| --- | --- | --- |
|before this PR| 1234 | 70 |
|master| 1141 | 238 |
|change| -93 | 168 |

Given above scan result this single commit introduced "addon" term in 168 linces and removed extension from 93. Because "addon" was much less popular term in sources it had just 70 occurrences. 

To the point - the "addon" term benefited from this change for no reason. Given that "extension" was used more frequently and it still outnumber addon any aligning should rather go in opposite direction than this PR. I know that "extension" is used in multiple contexts while addon is used just in specific one, but still **we had to introduce addon in 168 lines to remove extension from 93**. This means that addon was rather planted than aligned.

kaikreuzer:
Yeah, thanks @splatch, this is a very strong argument for my refactoring. The term "extension" is so overloaded that it is used for a million of things. "add-ons" are a very specific openHAB notion and so it is much better to have it cleanly separated from the rest to avoid and confusion.

PR #1563, opened 2020-07-21
[persistence] Allow max, min and avg calculations on other types

cweitkamp:
- Allow max, min and avg calculations on other types (e.g. `OnOffType`) which can be converted to `DecimalType`

Why I am thinking this is useful: Imagine you want to calculate the time someone is at home (or a light bulb was on). You are already using a [Generic Presence Detection](https://community.openhab.org/t/generic-presence-detection/17878) based on a master presence Switch - which state has to be persisted. It represents whether someone is at home or not. Currently it is very difficult to calculate the exact time of how long someone is present. By this feature you are able to calculate the average state of your Switch in a certain amount of time. The resulting value will be limited from [0...1] (0 = `OFF`, 1 = `ON`) and represents the percentage of of how long the Switch was `ON`.


kaikreuzer (approved):
lgtm, thanks!
Sorry, I had somehow missed the moment when you removed the WIP from the PR...

PR #1558, opened 2020-07-21
Use more java 8 features: Replace Collections.sort() with List.sort()

pavog:
Uses List.sort() instead of Collections.sort() to sort lists.
This function was introduced with Java 8 and might speed up sorting a bit.

See [javadocs](https://docs.oracle.com/javase/8/docs/api/java/util/List.html#sort-java.util.Comparator-) for reference.

PR #1564, opened 2020-07-21
Use more java 8 features: Replace Collections.sort() with List.sort() (second try)

pavog:
Uses List.sort() instead of Collections.sort() to sort lists.
This function was introduced with Java 8 and might speed up sorting a bit.

See [javadocs](https://docs.oracle.com/javase/8/docs/api/java/util/List.html#sort-java.util.Comparator-) for reference.

Second try, see #1558


kaikreuzer:
@cweitkamp And I leave it to you, which PR you want to merge first as the other one will have to be rebased then :-) (p.s.: I am fine to do so. First come, first serve!)

pavog:
I'd appreciate it if you could do the work.
I'm not that experienced with GitHub pull requests and rebasing.

PR #1573, opened 2020-07-29
[discovery] Added warning if 'representationProperty' is missing in keys of 'properties' map for 'DiscoveryResult'

cweitkamp:
- Added warning if `representationProperty` is missing in keys of `properties` map for `DiscoveryResult`

If one defines a `representationProperty` it should be present in the keys of the `properties` map. Otherwise the discovery feature will not work properly. In the past I saw some situations where the developer adds the value instead the key for it. Wdyt?


kaikreuzer:
@lolodomo The PR is still in draft status, that's why there isn't yet any feedback.

bundles/org.openhab.core.config.discovery/src/main/java/org/openhab/core/config/discovery/DiscoveryResultBuilder.java
```diff
@@ -143,6 +146,11 @@ public DiscoveryResultBuilder withTTL(long ttl) {
      */
     @SuppressWarnings("deprecation")
     public DiscoveryResult build() {
+        if (representationProperty != null && !properties.containsKey(representationProperty)) {
+            logger.warn(
+                    "Representation property '{}' of discovery result for thing '{}' is missing in properties map. It should be added.",
```
kaikreuzer, line 151: This message sounds as if the user should do something about it, but in fact you want to tell him that he just found a bug that needs to be reported, right? I think this should clearly be stated here, best by also including the stacktrace.

cweitkamp: It now looks like this:

```
[main] WARN org.openhab.core.config.discovery.DiscoveryResultBuilder - Representation property 'test' of discovery result for thing 'test:test:test' is missing in properties map. It has to be fixed by the bindings developer.
	at java.base/java.lang.Thread.getStackTrace(Thread.java:1607)
	at org.openhab.core.config.discovery.DiscoveryResultBuilder$1.run(DiscoveryResultBuilder.java:166)
	at org.openhab.core.config.discovery.DiscoveryResultBuilder$1.run(DiscoveryResultBuilder.java:1)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at org.openhab.core.config.discovery.DiscoveryResultBuilder.getStackTrace(DiscoveryResultBuilder.java:163)
	at org.openhab.core.config.discovery.DiscoveryResultBuilder.build(DiscoveryResultBuilder.java:156)
	at org.openhab.core.config.discovery.DiscoveryResultBuilderTest.testDiscoveryResultBuilder(DiscoveryResultBuilderTest.java:48)
```

kaikreuzer: Looks good to me. 👍

kaikreuzer (approved):

kaikreuzer:
@cweitkamp Could you please resolve the merge conflicts? Thanks!

cweitkamp:
Done. Depends on #1619.

PR #1575, opened 2020-07-31
Fix QuantityType.format timezone bug

wborn:
The following asserts failed for me when building in CEST:

```
        assertThat(seconds.format("%1$tH:%1$tM:%1$tS"), is("00:01:20"));
        assertThat(millis.format("%1$tHh %1$tMm %1$tSs"), is("00h 01m 20s"));
```

In the original code it would convert the quantity to my timezone by adding 1h and thus formatting the 80s quantity as `01:01:20` and `01h 01m 20s` respectively.

Caused by #1470

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityType.java
@@ -249,7 +249,7 @@ public class QuantityType<T extends Quantity<T>> extends Number
             if (millis != null) {
                 try {
                     return String.format(formatPattern,
-                            ZonedDateTime.ofInstant(Instant.ofEpochMilli(millis.longValue()), ZoneId.systemDefault()));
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityType.java
```diff
@@ -249,7 +249,7 @@ public String format(String pattern) {
             if (millis != null) {
                 try {
                     return String.format(formatPattern,
-                            ZonedDateTime.ofInstant(Instant.ofEpochMilli(millis.longValue()), ZoneId.systemDefault()));
+                            ZonedDateTime.ofInstant(Instant.ofEpochMilli(millis.longValue()), ZoneId.of("UTC")));
```
cweitkamp, line 252: Can we make the UTC ZoneId static?

wborn: There isn't a constants for it in the `ZoneId` class. But we can use `ZoneOffset.UTC` which is a constant and sub-class of `ZoneId`.

PR #1576, opened 2020-08-01
Migrate to OpenAPI 3 annotations

wborn:
* Upgrade to Swagger 2.1.0
* Add missing root and events resources
* Fix wrong and add missing response implementations


bundles/org.openhab.core.io.rest.ui/src/main/java/org/openhab/core/io/rest/ui/internal/UIResource.java
```diff
@@ -125,10 +127,9 @@ public Response getComponentByUID(@PathParam("namespace") String namespace,
     @RolesAllowed({ Role.ADMIN })
     @Path("/components/{namespace}")
     @Produces({ MediaType.APPLICATION_JSON })
-    @ApiOperation(value = "Add an UI component in the specified namespace.", authorizations = {
-            @Authorization(value = "oauth2", scopes = {
-                    @AuthorizationScope(scope = "admin", description = "Admin operations") }) })
-    @ApiResponses(value = { @ApiResponse(code = 200, message = "OK", response = Tile.class) })
```
ghys, line 131: Thanks for fixing the response class 😊 I've been meaning to take care of it, but 🤷

wborn: These copy/paste errors and wrong responses probably only become a major issue when using the spec with codegen for generating your own model classes. I think most users wouldn't really notice if they only use the Swagger UI for exploring the API. :-)

kaikreuzer:
@wborn Might this have broken our addons port https://ci.openhab.org/job/openHAB-Addons-port/?
From the timeline, this seems to be the culprit. I am not really sure what has changed about the availablility of `javax.activation`here, do you have any idea?

kaikreuzer:
I have created https://github.com/openhab/openhab-addons/pull/8272, which I tested in my IDE - I hope that no other bundles have similar problems.

kaikreuzer:
Hm, it wasn't the only one, the next one is mail that (might) need javax.activation: https://ci.openhab.org/job/openHAB-Addons-port/210/

wborn:
> do you have any idea?

Yes it should be fixed by https://github.com/openhab/openhab-core/pull/1582.

After that there is still an issue with the Netatmo Binding [generating code](https://github.com/openhab/openhab-addons/blob/6e32b2f1071311257fb84183c541c101b3317373/bundles/org.openhab.binding.netatmo/pom.xml#L60-L83) depending on the old Swagger version.

kaikreuzer:
Thanks @wborn, I have created https://github.com/openhab/openhab-addons/issues/8274 for Netatmo to keep track of it.

wborn:
Great! It seems to build again now the itests module is also excluded.
There aren't that many big dependencies left to upgrade now. :smile:

kaikreuzer:
I am just sitting here and upgrading my production system! 😄 
Most stuff is already working.

wborn:
That's certainly a big leap! :crossed_fingers:
Last time I did some major testing there were still some issues with linking channels.

PR #1586, opened 2020-08-11
Use "openhab" console scope

### LINKED ISSUE #1566: Scope/prefix for console commands
Current prefix for console commands is "smarthome".
Maybe it is something to change in OH3, replacing "smarthome" by "openhab" for example.
Hopefully, I discovered recently that this prefix is just optional and you can use for example "things list" as well as "smarthome:things list".

wborn:
Fixes #1566


wborn:
Yes I was about to take look at the next. It's better to break it now than later on.

PR #1551, opened 2020-07-16
[persistence] Added methods to calculate variance and standard deviation for persisted items

cweitkamp:
- Added methods to calculate variance and standard deviation for persisted items

Signed-off-by: Christoph Weitkamp <github@christophweitkamp.de>

PR #1588, opened 2020-08-11
switched PersistenceExtensions to ZonedDateTime

kaikreuzer:
In https://github.com/openhab/openhab-core/pull/1451 I changed the rules from `Instant` to `ZonedDateTime`, but I had missed the `PersistenceExtensions`. Fixing this, I also did some further clean up and also moved it to `core.persistence` as it does not have any dependency on the persist DSL itself.


kaikreuzer:
@openhab/core-maintainers Tests are finally green, so ready for review and hopefully soon to merge!

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/scoping/ScriptImplicitlyImportedTypes.java
```diff
@@ -90,15 +82,11 @@
...
+        result.add(ScriptExecution.class);
         result.add(BusEvent.class);
         result.add(Exec.class);
         result.add(HTTP.class);
         result.add(Ping.class);
-        result.add(ScriptExecution.class);
-        result.add(LogAction.class);
-        result.add(Audio.class);
-        result.add(Voice.class);
-        result.add(Things.class);
```
wborn, line 101: Are the removals in this class because the rule engine already gets this information provided by `ActionService`s?

itests/pom.xml
```diff
@@ -35,11 +35,11 @@
     <module>org.openhab.core.io.rest.core.tests</module>
     <module>org.openhab.core.model.core.tests</module>
     <module>org.openhab.core.model.item.tests</module>
-    <module>org.openhab.core.model.persistence.tests</module>
     <module>org.openhab.core.model.rule.tests</module>
     <module>org.openhab.core.model.script.tests</module>
     <module>org.openhab.core.model.thing.testsupport</module>
     <module>org.openhab.core.model.thing.tests</module>
+    <module>org.openhab.core.persistence.tests</module>
```
wborn, line 42: I saw these tests were moved to another project in #1551.
Does it even require OSGi for running the tests?

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/scoping/ScriptImplicitlyImportedTypes.java
kaikreuzer: Yes, exactly. This seemed duplicated and I checked that it works without it.

itests/pom.xml
kaikreuzer: At least in their current form, yes.

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/scoping/ScriptImplicitlyImportedTypes.java
wborn: Than we can safely merge this. :+1:

wborn:
Looks like the webui's also need to be updated for these changes @kaikreuzer.

kaikreuzer:
Yeah, indeed, the CometVisu seems to have issues now...
Will have a look at it later.

kaikreuzer:
Done: https://github.com/openhab/openhab-webui/pull/307

itests/pom.xml
cweitkamp: I think it is safe to move all the related tests because from my understanding they do not need OSGi.

PR #1590, opened 2020-08-13
re-added log action to script scope and cleaned up imported extensions

kaikreuzer:
The log feature in scripts got accidentially lost in https://github.com/openhab/openhab-core/pull/1588.

Additionally, I noticed that those classes only need to be added as static imports, but not as extensions. Extensions are merely used for operator overloading and stuff like that.


wborn:
The build now fails because there was a suppression for LogAction:

https://github.com/openhab/openhab-core/blob/3c83e8a8eba1945204e6f25feda70ea44095c075/tools/static-code-analysis/spotbugs/suppressions.xml#L29

PR #1591, opened 2020-08-14
[rest] fixed item links in rest api

kaikreuzer:
Fixes https://github.com/openhab/openhab-android/issues/2139

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
@@ -188,7 +188,6 @@ public class ItemResource implements RESTResource {
     private UriBuilder uriBuilder(final UriInfo uriInfo, final HttpHeaders httpHeaders) {
         final UriBuilder uriBuilder = uriInfo.getAbsolutePathBuilder();
         respectForwarded(uriBuilder, httpHeaders);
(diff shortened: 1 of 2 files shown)


wborn:
Do you know if there are more issues with the item linking?
There was also a discussion about a regression introduced by #1385.

kaikreuzer:
This is not at all related to item links (the openHAB concept) and thus not related to #1385.
It is about the "link" field within the REST API responses, see openhab/openhab-android#2139 for details.

wborn:
Yes I understand that. When I did my most recent testing the linking appeared to be buggy but I had no time to further investigate the root cause. So I don't know if it was caused by #1385 or some other issue like this one. Therefor I postponed updating my secondary OH instance to OH3. :wink:

kaikreuzer:
> Therefor I postponed updating my secondary OH instance to OH3

This issue here is hopefully my last "major" bug that I have encountered.
It really helped my to upgrade my (one and only) production system to really have the urge to debug and fix stuff - I can highly recommend doing so and I plan to push all of you to do the same very soon, so be warned 😎 .

wborn:
I don't mind running it on my secondary instance when the linking works. But for my main instance I will first implement a WoL action (https://github.com/openhab/openhab-addons/issues/3799) or I wouldn't be able to turn on my desktop PC using voice commands. :smile:

ghys:
> I will first implement a WoL action

Cool ;) I'm using the exec binding to run `wakeonlan <mac>` but would definitely prefer an action ;)

PR #1611, opened 2020-08-26
Allow property value to be updated to null

wborn:
With the changes in #1598 the BaseThingHandler could no longer update property values to null.

---

In the unchanged Nest tests the following stracktraces were logged:

```
[pool-80-thread-1] WARN  o.o.b.n.i.r.NestStreamingRestClient - An exception occurred while processing the inbound event
	at java.base/java.util.Objects.requireNonNull(Objects.java:221)
	at java.base/java.util.ImmutableCollections$Map1.<init>(ImmutableCollections.java:749)
	at java.base/java.util.Map.of(Map.java:1307)
	at org.openhab.core.thing.binding.BaseThingHandler.updateProperty(BaseThingHandler.java:512)
	at org.openhab.binding.nest.internal.handler.NestThermostatHandler.update(NestThermostatHandler.java:211)
	at org.openhab.binding.nest.internal.handler.NestThermostatHandler.update(NestThermostatHandler.java:1)
	at org.openhab.binding.nest.internal.handler.NestBaseHandler.onUpdatedData(NestBaseHandler.java:194)
	at org.openhab.binding.nest.internal.update.NestUpdateHandler.lambda$3(NestUpdateHandler.java:99)
```

See also https://github.com/openhab/openhab-addons/pull/8334#issuecomment-680696040.

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/BaseThingHandler.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/BaseThingHandler.java
@@ -12,6 +12,7 @@
  */
 package org.openhab.core.thing.binding;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -509,7 +510,7 @@ public abstract class BaseThingHandler implements ThingHandler {
      * @param value the value of the property
      */
     protected void updateProperty(String name, String value) {
-        updateProperties(Map.of(name, value));
+        updateProperties(Collections.singletonMap(name, value));
     }
 
     /**


kaikreuzer (approved):
At least you broke it yourself ;-)
Can we have a similar issue at other occasions as well? `Map.of` seems to have a different semantic than `Collections.singletonMap` and thus is not a drop-in replacement as we had thought...

wborn:
Yes these immutable objects sometimes have different semantics. I already learned that the hard way before creating the PR because with some initial changes several core unit/i-tests failed which took hours of debugging. So I reverted some changes while going through all changes again. But apparently this change slipped through. :neutral_face:

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/BaseThingHandler.java
```diff
@@ -509,7 +510,7 @@ protected void updateProperties(Map<String, String> properties) {
      * @param value the value of the property
      */
     protected void updateProperty(String name, String value) {
```
cweitkamp, line 512: Hm, ... shouldn't the `value` be annotated by `@Nullable` then?

cweitkamp (commented):
It looks like we didn't cover this feature - unset the property if value is `null` - by a unit test at all. We can add it but one cannot pass a `null` value to the `updateProperty(...)` method directly because the way it is annotated does not allow it ...

cweitkamp:
> We can add it

I submitted a PR (#1615) including the test.

wborn:
Thanks for adding the test @cweitkamp! :+1:

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/BaseThingHandler.java
wborn: That makes sense and if we keep allowing null values the Map value type in `protected void updateProperties(Map<String, String> properties)` should also be properly annotated. I don't know how common it is for properties to be nullable. Another way could be to add a `removeProperty` method so we can rid of keys with `null` values. But `updateProperties` would also have to be updated for that and I don't know how breaking all these API changes would be. :wink:

Issue #979, opened 2019-08-15
[rest] get request to /items/itemname does return values with the wrong encoding


spacemanspiff2007:
A get request to the items command still doesn't return properly encoded (utf-8) strings! [Original issue](https://github.com/eclipse/smarthome/issues/4041)

To reproduce create this in an items file
```
String	TestUmlaut  "Label Umlaute: äöüß [%s]"
```

Response:
```
{
  "link": "http://localhost:8080/rest/items/TestUmlaut",
  "state": "NULL",
  "stateDescription": {
    "pattern": "%s",
    "readOnly": false,
    "options": []
  },
  "editable": false,
  "type": "String",
  "name": "TestUmlaut",
  "label": "Label Umlaute: ����",
  "tags": [],
  "groupNames": []
}
```
It seems to ascii or something else:
```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe4 in position 212
```

cweitkamp:
Can you check the file encoding of your items file?

spacemanspiff2007:
It's utf-8

spacemanspiff2007:
If I request all items (```/rest/items```) it works properly:

```
  {
    "link": "http://localhost:8080/rest/items/TestUmlaut",
    "state": "NULL",
    "stateDescription": {
      "pattern": "%s",
      "readOnly": false,
      "options": []
    },
    "editable": false,
    "type": "String",
    "name": "TestUmlaut",
    "label": "Label Umlaute: äöüß",
    "tags": [],
    "groupNames": []
  },
```

cweitkamp:
My findings after a quick search: The two REST endpoints indeed using different ways to create the resulting response.

https://github.com/openhab/openhab-core/blob/608507e451ca7b71171f273cfc9f3c1c7831df61/bundles/org.openhab.core.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/item/ItemResource.java#L218-L235

https://github.com/openhab/openhab-core/blob/608507e451ca7b71171f273cfc9f3c1c7831df61/bundles/org.openhab.core.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/item/ItemResource.java#L244-L265

I guess the issue is somewhere in there.

spacemanspiff2007:
Could you restructure the code so that the mapping the same function?

cweitkamp:
[...]

PR #1616, opened 2020-08-31
Removes the incorrect use of the encoding header in each JSON response

pavog:
The ResponseBuilder's `encoding(String encoding)` function sets the http header "Content-Encoding", which is used to compress the media type and not to specify the character encoding.

See [ResponseBuilderImpl](https://github.com/apache/cxf/blob/master/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseBuilderImpl.java#L286):
```java
public ResponseBuilder encoding(String encoding) {
    return this.setHeader("Content-Encoding", encoding);
}
```

The content-encoding header should be used to compress the media type.
See [MDN Web Docs Content-Encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding).

If you want to specify the charset you should use the content-type header.
See [MDN Web Docs Content-Type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type).

### DIFF
--- a/bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/JSONResponse.java
+++ b/bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/JSONResponse.java
@@ -94,8 +94,7 @@ public class JSONResponse {
      * @return ResponseBuilder configured for "Content-Type" MediaType.APPLICATION_JSON
      */
     private ResponseBuilder responseBuilder(Response.StatusType status) {
-        return Response.status(status).header("Content-Type", MediaType.APPLICATION_JSON)
-                .encoding(StandardCharsets.UTF_8.name());
+        return Response.status(status).header("Content-Type", MediaType.APPLICATION_JSON);
     }
 
     /**


wborn:
It seems to have been added to fix encoding issues on Windows (#979, #982).
Would this change again cause these issues?

pavog:
I hope it does not. Because using "UTF-8" in the content-encoding header is actually wrong.
At least the test [ItemResourceOSGiTest](https://github.com/openhab/openhab-core/pull/982/commits/a675e34698cd5d54aad6c41bde4a8e6d598de099#diff-ac88f1606e05d638b19985354dfb6148) (which checks the encoding for special characters) does not fail [on jenkins](https://travis-ci.com/github/openhab/openhab-core/builds/182046112).

cweitkamp:
I cannot test it on Windows. But it looks like my changes in #982 were not the real fix. Maybe the real fix was #1150. We in general should consider refactoring respo6 creation at all because as described in https://github.com/openhab/openhab-core/issues/979#issuecomment-521677133 there are multiple ways how we create them internally.

I guess the tests are still working because in the meantime we moved away from `org.apache.commons.io` - which was used to extract the values from mocked response - to native Java including UTF-8 support.

pavog:
If you could tell me how I can test it on Windows, I would be happy to do so.

I believe that all variants internally call the above mentioned function.

wborn (approved):
I did some cross OS testing with this change (similar to #1150) and this change didn't seem to impact responses.

Hilbrand:
Just something I noticed for @pavog All your contributions are signed with an invalid e-mail address. You should sign off your commits with a real e-mail address. See https://www.openhab.org/docs/developer/contributing.html#sign-your-work

Issue #1354, opened 2020-02-04
Some addons restart after uninstalling an unrelated addon


openhab-5iver:
I believe this is a symptom of issues reported in the forum, where zigbee or zwave are no longer working after upgrading to 2.5. Logs show cyclic binding uninstalls, which is what I am seeing too. When uninstalling an addon, there are other addons being stopped and started. Included in these are zigbee and zwave. Zigbee does not behave well to binding restarts and usually requires an OH restart to get it working properly. Zwave handles binding restarts OK, but it is a resource hog and responsiveness in the zwave network will be much reduced. If this happens repeatedly, then I can understand why people think 2.5 is broken. 

I was recently looking into the new icalendar binding and needed to remove the 1.x caldav-command to test something. I noticed in the log and figured it was time to report this. I suspect that zwave and zigbee are restarting due to org.openhab.core.config.serial being restarted. Here is the log...

```
2020-02-04 13:11:54.775 [INFO ] [org.apache.sshd.server.session.ServerUserAuthService] - Session openhab@/127.0.0.1:39180 authenticated
2020-02-04 13:23:53.647 [INFO ] [org.apache.karaf.features.internal.service.FeaturesServiceImpl] - Removing features: openhab-binding-caldav-command1/[1.14.0,1.14.0]
2020-02-04 13:23:55.665 [INFO ] [org.apache.karaf.features.internal.service.FeaturesServiceImpl] - Changes to perform:
2020-02-04 13:23:55.665 [INFO ] [org.apache.karaf.features.internal.service.FeaturesServiceImpl] -   Region: root
2020-02-04 13:23:55.665 [INFO ] [org.apache.karaf.features.internal.service.FeaturesServiceImpl] -     Bundles to uninstall:
2020-02-04 13:23:55.665 [INFO ] [org.apache.karaf.features.internal.service.FeaturesServiceImpl] -       org.openhab.binding.caldav-command/1.14.0
2020-02-04 13:23:55.666 [INFO ] [org.apache.karaf.features.internal.service.FeaturesServiceImpl] - Stopping bundles:
2020-02-04 13:23:55.666 [INFO ] [org.apache.karaf.features.internal.service.FeaturesServiceImpl] -   org.openhab.binding.zwave/2.5.2.202001250516
2020-02-04 13:23:57.392 [INFO ] [org.apache.karaf.features.internal.service.FeaturesServiceImpl] -   org.openhab.binding.zigbee.telegesis/2.5.2.202001250519
[...]

Issue #1322, opened 2020-01-08
Generic bug installing bindings


kandersen01:
This issue is very hard to decripe, so please be patient....

Fresh install of the officiel Rasibian Buster on a Rpi4. 
Having Google TTS, Habmin or Rest Docs installed will result in issue installing another addon will cause some others bindings to stop, and after a few seconds they will start again.
This happens having openhab cloud, netamo, expire and zwave binding installed. They will all stop during the installation of another binding, and restart again after a few seconds.

Here is an example where I install Google TTS:

```2020-01-08 21:52:20.664 [INFO ] [io.openhabcloud.internal.CloudClient] - Shutting down openHAB Cloud service connection

==> /var/log/openhab2/events.log <==

2020-01-08 21:52:21.238 [hingStatusInfoChangedEvent] - 'netatmo:NAMain:61f15dc2:70ee502727e6' changed from ONLINE to UNINITIALIZED
2020-01-08 21:52:21.250 [hingStatusInfoChangedEvent] - 'netatmo:NAMain:61f15dc2:70ee502727e6' changed from UNINITIALIZED to UNINITIALIZED (HANDLER_MISSING_ERROR)
2020-01-08 21:52:21.255 [hingStatusInfoChangedEvent] - 'netatmo:NAModule1:61f15dc2:0200002741f4' changed from ONLINE to UNINITIALIZED
2020-01-08 21:52:21.267 [hingStatusInfoChangedEvent] - 'netatmo:NAModule1:61f15dc2:0200002741f4' changed from UNINITIALIZED to UNINITIALIZED (HANDLER_MISSING_ERROR)
2020-01-08 21:52:21.272 [hingStatusInfoChangedEvent] - 'netatmo:netatmoapi:61f15dc2' changed from ONLINE to UNINITIALIZED
2020-01-08 21:52:21.291 [hingStatusInfoChangedEvent] - 'netatmo:netatmoapi:61f15dc2' changed from UNINITIALIZED to UNINITIALIZED (HANDLER_MISSING_ERROR)

==> /var/log/openhab2/openhab.log <==

2020-01-08 21:52:21.858 [INFO ] [b.core.service.AbstractActiveService] - Expire Refresh Service has been shut down

==> /var/log/openhab2/events.log <==

[...]

Issue #286, opened 2018-02-02
Invalid binding name in addons.cfg causes memory usage to increase continuously


mhilbush:
@kaikreuzer This is the issue I mentioned here https://github.com/eclipse/smarthome/issues/4871#issuecomment-362714353

In the course of analyzing https://github.com/eclipse/smarthome/issues/4871, I noticed an unrelated condition where memory utilization would increase continuously. The condition was caused by entering an invalid binding name in `services/addons.cfg`. Since an error is not logged due to an invalid binding name in addons.cfg, it's possible that a user would not easily be able to determine the cause.

Here's the base memory data collected while I had sar running. 
[openhab-pidstat.txt](https://github.com/openhab/openhab-core/files/1690954/openhab-pidstat.txt)

I used the following command, where `12299` was the pid of the openhab java process.
```
sudo pidstat -r 6 -p 12299
```

Here's a graph of minflt/s and RSS plotted over time. 
- The memory began to increase every minute (presumably when the feature installer runs) as soon as I added an invalid binding name to addons.cfg.
- The memory consumption flattened out as soon as I removed the invalid binding name from addons.cfg 

![sar-minflt-rss](https://user-images.githubusercontent.com/19286354/35756490-953b5c8e-0839-11e8-9a69-f3a8bdff3a9b.jpg)

This was a short duration test to illustrate the issue. I had other tests that ran much longer.

Edit: Note that I was running build 1194.

kaikreuzer:
Thanks, this indeed looks like some memory leak.
It is ok that the feature installer kicks in once a minute - this is actually what it does, it checks whether all add-ons are installed as requested and if not, it tries to do so. But this must clearly not result in any memory issues.

abedwardsw:
I am not 100% sure if this is related, but I was having issues where I was seeing the message “Offline: waiting for connection to become available” quite often after I upgraded to 2.4, I happened to stumble on this issue due to the related thread https://github.com/eclipse/smarthome/issues/4871#issuecomment-362714353

I eventually figured out that I had the mqqtitude binding listed in addons.cfg which was only available in the legacy bindings repo, therefore it didn't auto install after the upgrade.   I removed the binding from addons.cfg and my issues with "Offline: waiting for connection.." in paperui went away.   

[...]

PR #1617, opened 2020-09-02
Improved add-on installation logic

kaikreuzer:
- Checking the existence of an add-on before trying to install it and otherwise logging a warning.
- Additionally checking for real changes in the config, which stops the regular re-installation mechanism that otherwise runs every minute. In case of a failure, this means people have to manually restart or to change their configuration once, but they anyhow see an error in the log.
- Applied thread naming convention and removed one (hopefully?) unnecessary scheduler thread.

### DIFF
--- a/bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/FeatureInstaller.java
+++ b/bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/FeatureInstaller.java
@@ -421,6 +421,7 @@ public class FeatureInstaller implements ConfigurationListener {
                 }
                 if (!failed.isEmpty()) {
                     logger.error("Failed installing '{}'", failed.stream().collect(Collectors.joining(", ")));
+                    configMapCache = null; // make sure we retry the installation
                 }
 
                 for (String addon : installed) {
@@ -428,10 +429,12 @@ public class FeatureInstaller implements ConfigurationListener {
                 }
             } catch (Exception e) {
                 logger.error("Failed retrieving features: {}", e.getMessage());
+                configMapCache = null; // make sure we retry the installation
             }
         } catch (Exception e) {
             logger.error("Failed installing '{}': {}", addons.stream().collect(Collectors.joining(", ")),
                     e.getMessage());
+            configMapCache = null; // make sure we retry the installation
         }
     }
 
@@ -455,6 +458,7 @@ public class FeatureInstaller implements ConfigurationListener {
             }
         } catch (Exception e) {
             logger.error("Failed installing '{}': {}", name, e.getMessage());
+            configMapCache = null; // make sure we retry the installation
         }
         return false;
     }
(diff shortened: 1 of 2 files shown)


wborn:
Do you think this will still work well on low-spec machines where it can take minutes before all bundles are installed and downloaded? I once experimented with removing the 1 minute install delay and ran into the issue that everything wasn't properly initialized. That was on a fast machine so I can imagine it can take minutes on a Raspberry Pi 3 before everything is in place and add-ons can be properly installed.

kaikreuzer:
I would expect Karaf to install the features as requested, even on slow machines during startup. But yeah, my expectation might not match reality...
I have pushed a commit that enables the retry whenever a feature installation has failed - so I hope this should nicely cover any slow startup issues while it still makes sure that the feature installer is not called anymore, if all installations have succeeded.

wborn (approved):
Looks like a good improvement, thanks!

bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/FeatureInstaller.java
```diff
@@ -336,15 +339,19 @@ private synchronized void installAddons(final FeaturesService service, final Map
                     Feature[] features = service.listInstalledFeatures();
                     for (String addon : entries) {
                         if (addon != null && !addon.isEmpty()) {
```
cweitkamp, line 341: At the time we use `String.strip()` it might be better to check for `isBlank()` here.

cweitkamp (commented):
Thanks for these improvements. It might be a solution for #1322 and #1354 as well as #286. Wdyt?

bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/FeatureInstaller.java
```diff
@@ -336,15 +339,19 @@ private synchronized void installAddons(final FeaturesService service, final Map
...
+                            } else {
+                                logger.warn("The {} add-on '{}' does not exist - ignoring it.", type, addon.strip());
                             }
                         }
                     }
 
-                    // we collect all installed addons
+                    // we collect all installed add-ons
                     for (String addon : getAllAddonsOfType(service, type)) {
                         if (addon != null && !addon.isEmpty()) {
```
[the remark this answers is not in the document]
wborn: I think it would be better to move any such checks to `getAllAddonsOfType` so it doesn't return unusable add-ons.

bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/FeatureInstaller.java
wborn: It would prevent errors when there is only whitespace between commas: "addon1, ,addon2"

wborn: Though the whitespace would probably cause the new check `(service.getFeature(id) != null)` to return `null` :wink:

wborn: But it will still trigger the warning logic:

                                logger.warn("The {} add-on '{}' does not exist - ignoring it.", type, addon.strip());

kaikreuzer:
Feel free to create further PRs for improvements :-)

PR #1626, opened 2020-09-07
Fix null analysis errors/warnings

wborn:
The changes in #1614 caused null analysis errors in `CommunicationManager.createCallback` and caused unnecessary null check warnings in `ProfileCallbackImpl`.


kaikreuzer (approved):

cweitkamp:
Great. Thanks. I was looking into this too but you won the race.

I'd now suggest to add an additional `null` check on the `item` after line

https://github.com/wborn/openhab-core/blob/c95f176cf9be9355068614c5a88702328b10c1a6/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/profiles/ProfileCallbackImpl.java#L139

PR #1631, opened 2020-09-09
[REST/Auth] Always accept tokens in the alt header

ghys:
Currently the AuthFilter will try to find a token in the
X-OPENHAB-TOKEN HTTP header - only when it finds a cookie
named X-OPENHAB-AUTH-HEADER. It can cause problems because
browsers or proxies might block the cookie from being sent
for various reasons (for instance if there's a path set
for it).

There is no downside IMHO to always try to fallback to
checking the X-OPENHAB-TOKEN header for a token, if and
only if it's not already provided in the Authorization
header. It is the responsibility of the client to decide
how it wants to authorize the request among the available
options - by checking a cookie, or something else entirely.

Also removed the '?api_key=' option because Swagger UI
doesn't provide tokens that way anymore.


digitaldan:
Fwiw , openHAB cloud strips off cookies before proxing requests back to a users openHAB instance

kaikreuzer (approved):
Sounds fine for me, thanks.

PR #1634, opened 2020-09-10
Fix item channel links not properly initialized

### LINKED ISSUE #1596: Item channel links not properly initialized
A regression was introduced by https://github.com/openhab/openhab-core/pull/1385 causing ThingHandlers not to receive channel link/unlink events. As a result items do not always get proper state.

It's reproducible by installing the Astro Binding. Creating a Sun Thing then add some items to the model using the UI. If you then check the items in the UI or run `items list` the state remains null. As workaround you can restart the binding bundle resulting in the items getting proper state.

wborn:
I've tested the `ChannelLinkNotifier` by @splatch (https://github.com/openhab/openhab-core/pull/1385#issuecomment-633454655) and it solves the channel linking issues for me. 

I made a few small changes:
* add `@NonNullByDefault`
* tune logging
* addRegistryChangeListener before notifying handlers to prevent events getting lost (if channels can be unlinked during this call synchronization may be necessary too)
* removeRegistryChangeListener on `@Deactivate`

With the channels being properly linked, adding Things via the inbox and adding them to the model works a lot better. :-)

Fixes #1596

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ChannelLinkNotifier.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ChannelLinkNotifier.java
@@ -1,0 +1,99 @@
+/**
+ * Copyright (c) 2010-2020 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.thing.internal;
+
+import java.util.function.Consumer;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.openhab.core.common.registry.RegistryChangeListener;
+import org.openhab.core.thing.ChannelUID;
+import org.openhab.core.thing.Thing;
+import org.openhab.core.thing.ThingRegistry;
+import org.openhab.core.thing.ThingUID;
+import org.openhab.core.thing.binding.ThingHandler;
+import org.openhab.core.thing.link.ItemChannelLink;
+import org.openhab.core.thing.link.ItemChannelLinkRegistry;
+import org.openhab.core.thing.util.ThingHandlerHelper;
+import org.osgi.service.component.annotations.Activate;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Deactivate;
+import org.osgi.service.component.annotations.Reference;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The {@link ChannelLinkNotifier} notifies initialized thing handlers of channels being linked or unlinked.
+ *
+ * @author Łukasz Dywicki - Initial contribution
+ * @author Wouter Born - Initial contribution
+ */
+@Component(immediate = true)
+@NonNullByDefault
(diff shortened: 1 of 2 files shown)


cweitkamp:
Code Looks great. Thanks to both of you. To avoid future regressions do you think it is possible to add an integration test - in a simple way - for this feature?

wborn:
I haven't added any tests because I first wanted to know from the other @openhab/core-maintainers if we still think this is the way to go (and it was getting late :wink:).

cweitkamp:
As I said, I like it and I agree on it :wink:.

kaikreuzer:
And I already stated in https://github.com/openhab/openhab-core/pull/1385#issuecomment-641536209 that I like it 😎 .

wborn:
There may be issues with this implementation when you link multiple items to one thing channel. If you remove one of the links it will send a "channelUnlinked" event to the handler. Which may use the event to stop a channel update mechanism (polling).

I think the handler should only be informed of these changes if:

* the number of linked items to the channel changed from **0** to **at least 1** -> channelLinked
* the number of linked items to the channel changed from **at least 1** to **0** -> channelUnlinked

It looks like this may already have been an issue with the old implementation.

Similarly it will call "channelLinked" multiple times when the `ChannelLinkNotifier` is activated and there are several item channel links for the same thing channel.

Another improvement may be to use a scheduler for these events to reduce the impact of thing handlers hijacking threads for long running tasks. It was also used in the old implementation.

splatch:
> Another improvement may be to use a scheduler for these events to reduce the impact of thing handlers hijacking threads for long running tasks. It was also used in the old implementation.

Looking at issue and how it looks a like I believe that we should rather focus on keeping link identifier as part of notification. If we have channel linked which is related to link which have been just created or removed the divider is link and not channel.

There might be cases when binding wants to keep distinct subscriptions for each and every link due to let say latency or better concurrency. Letting binding developer decide how to handle above scenario is better than building an extra logic in framework notifier itself.

Best,
Łukasz

wborn:
I've now added some pretty decent integration test coverage as well.

@splatch adding such details would be breaking API changes and I was under the impression the details of links and items were hidden from handlers on purpose. But perhaps the other @openhab/core-maintainers can also comment on that.

splatch:
> adding such details would be breaking API changes and I was under the impression the details of links and items were hidden from handlers on purpose.

@wborn Given that this is one of less popular framework features coming next major release is great opportunity to introduce breaking change. If a whole intention is to keep handler unaware of links then idea of having link notifications is wrong in first place.
Given that `Link` bridges `Item` together with `Channel` it seems to play keep role in keeping both separated, yet it can't and shouldn't stay hidden or belong just to an item.

kaikreuzer:
> I think the handler should only be informed of these changes if:
> the number of linked items to the channel changed from 0 to at least 1 -> channelLinked
> the number of linked items to the channel changed from at least 1 to 0 -> channelUnlinked

Yes, I fully agree.

>  the details of links and items were hidden from handlers on purpose

Correct. The purpose of these methods was only for the handler to know, whether a certain channel is at all in use as it might otherwise skip expensive internal operations or unnecessary communication.

wborn:
After some more thoughts I think it might still be necessary to send "channelLinked" events for already linked channels. That may be the only way for newly linked items to receive the current channel state.

robnielsen:
The Insteon binding relies on `channelLinked `and `channelUnlinked ` methods to build the bridge between OH and the Insteon network. It was reported in the OH community at https://community.openhab.org/t/insteon-binding-on-3-0-0-snapshot/105406. I was able to reproduce the same behavior with things defined in .things files and items defined in .items files.

kaikreuzer (approved):
lgtm, many thanks!

PR #1635, opened 2020-09-11
[automation] Add script extension helpers to track loading/unloading refs

jpg0:
Script extension module to allow scripts to:
* Register for callbacks when they are unloaded (in case they need to clean up)
* Provide access to the OSGi bundle context to get hold of host services

Example usage: https://github.com/jpg0/ohj/blob/master/osgi.js

Some questions:
* Should this be in 'rulesupport'? I put it here because there is nowhere else obvious, but it's a questionable location. However I also don't want to end up with many tiny bundles.
* Currently script lifecycle and osgi are together. It grew this way because I used it this way, but maybe it shouldn't be together. Again, I'm concerned about fragmentation as splitting would create tiny extension modules.

### DIFF
--- a/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/shared/osgi/Lifecycle.java
+++ b/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/shared/osgi/Lifecycle.java
@@ -1,0 +1,64 @@
+/**
+ * Copyright (c) 2010-2020 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.automation.module.script.rulesupport.shared.osgi;
+
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.function.Consumer;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Allows scripts to register for lifecycle events
+ *
+ * @author Jonathan Gilbert
+ */
+public class Lifecycle implements ScriptDisposalAware {
+    private static final Logger logger = LoggerFactory.getLogger(Lifecycle.class);
+    public static final int DEFAULT_PRIORITY = 50;
+    private List<Hook> listeners = new ArrayList<>();
+
+    public void addDisposeHook(Consumer<Object> listener, int priority) {
+        addListener(listener, priority);
+    }
+
+    public void addDisposeHook(Consumer<Object> listener) {
+        addDisposeHook(listener, DEFAULT_PRIORITY);
+    }
+
+    private void addListener(Consumer<Object> listener, int priority) {
+        listeners.add(new Hook(priority, listener));
+    }
+
+    @Override
+    public void unload(String scriptIdentifier) {
+        try {
+            listeners.stream().sorted(Comparator.comparingInt(h -> h.priority))
+                    .forEach(h -> h.fn.accept(scriptIdentifier));
+        } catch (RuntimeException ex) {
+            logger.warn("Script unloading halted due to exception in disposal: {}: {}", ex.getClass(), ex.getMessage());
+        }
+    }
+
+    private static class Hook {
+        public Hook(int priority, Consumer<Object> fn) {
+            this.priority = priority;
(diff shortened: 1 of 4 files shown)


kaikreuzer:
> Should this be in 'rulesupport'?
> Currently script lifecycle and osgi are together.

Good questions. What is the goal of these helpers in general? It feels as if it is indeed not related to writing scripting rules, but rather allow coding "stuff" for openHAB in JS instead of Java.
Users are not supposed to interact with bundleContexts and other inner workings of openHAB, so we are clearly leaving the scope of "user scripts" here. We should hence be careful not to smoothly move from user scripting to openHAB coding in JS since for both we have very different scopes of what their API is.

jpg0:
So I just took at a look at all the places where this is used (directly & indirectly). There are plenty of places in the helper library: https://github.com/jpg0/ohj/search?q=osgi

The typical pattern here is to get a reference to a service to use in the JS helper. The helper I would indeed class as 'coding stuff', although it's purpose is to allow scripts use this stuff for rules. Simple examples would be to allow access to openHAB actions, or to provide access to read and update Item metadata.

However, it would certainly be possible to push this code into Java, directly into classes that are exposed like the other things in rulesupport. My feeling is:
- This is the right way to go in general, as it allows use from other languages too
- I do think that there is benefit in providing access for development and prototyping of these things in the scripting language (I accept this could be in a specific plugin to support it)
- I expect that this would not remove the need to wrap these things in the language-specific code, as there are too many oddities in the scripting-language-to-Java translation provided by different script runtimes (as they all differ and need to be completely generic).

Ultimately, I guess the question is how much support we want to offer to allow the creation of helper libraries (in script, possibly as a path to port them to Java). It is possible today without this, via static references to osgi classes, but these libraries make it easier. If we do want to support this, then I'd suggest that we do include things like this. If we don't, then I'd suggest that I:
- Create a new plugin with these extensions, I'll keep it in my repo, not sure that it needs to be in openHAB anywhere
- I'll put guards into my JS helper library to fail if an 'experimental' feature is accessed which needs this plugin
- I'll gradually move these features into Java code, and rewrap (if necessary) in JS, removing the guards

Either approach works for me.

jpg0:
@kaikreuzer a question related to my response above:

What should we expect scripts to be able to do? Right now really everything is in 'rulesupport' which suggests that they are for creating rules. Looking at my current config, I use scripting to:

* Create Rules (plus abstract & reuse common functionality across them etc)
* Create Things and Items (again, to allow reuse that I cannot achieve otherwise)
* Create Sitemaps (based on existing Items, rather than hard-coded, with some dynamic bits etc)

This is a little different from the question above on how much support we want to offer to allow the creation of helper libraries, but means that even if we don't, the 'rulesupport' question still stands.

kaikreuzer:
Hey @jpg0,

Sorry for not responding any earlier - you have very good thoughts and questions and that makes it hard to answer in an equally profound manner, so answering is often kept for later ;-) .

But even thinking about it, I cannot give you any good answer.
Yes, there are mainly 2 distinct things to look at here:
1. The helper libraries, which more or less can be seen as an API for "user scripts". As such APIs, they should be considered part of the core framework. They need to be maintained, versioned, documented, backward compatibility should be kept as much as possible, etc. Being part of the core framework, they would have access to "inner services" as when something is refactored in the core framework, such refactorings would then also cover the helper libraries to make sure that nothing breaks. With those considerations, I'd personally prefer your third option from above, i.e. moving as much of it to Java code as possible, since that's the only way that the very small team of openhab core maintainers can help maintaining it and not breaking it when doing changes in the core.
2. What should we expect scripts to be able to do? That's now the user perspective, i.e. which APIs do we offer (and maintain and document) through (1) above. You are right, at the moment the focus is clearly on writing rules, hence "rulesupport". I see the value in being able to also define items, things, sitemaps and maybe even write complete bindings in scripts. But at the same time I fear the complexity that comes with having to provide a stable, backward-compatible API for such things and the risk of not being able to maintain and support a polyglot setup in the long run. I'd hence prefer a step-wise expansion of the API and see how well it is picked up and where we feel comfortable to further extend it to.

But these are just my 2 cents, I'd also very much love to hear other @openhab/core-maintainers opinions!

jpg0:
Hi @kaikreuzer ,

1. So I completely agree with pushing as much to Java as possible; the question is whether that completely eliminates the need for more 'internal' access. I don't really think of user scripts on one side and helper libraries on the other, but rather a spectrum: for example community provided libraries sit in between these, and even libraries that users may create to reuse functionality. How will we suggest that people develop a new script that uses something not yet exposed in Java APIs? Having them write Java first is a big hurdle, however we would really want the contributions to be primarily Java so that they can be maintained and reused across languages.

2. That's a reasonable approach; to watch and learn. It will mean that I need to separate the libraries that I currently use into rules-specific (to contribute) and otherwise (to keep to myself for the moment), but I don't think that should be a problem.

jpg0:
As this isn't directly required for getting GraalJS in, I'm going to close it. I'll focus on getting the core language in, and other 'helpers' can come later.

PR #1641, opened 2020-09-13
Add null annotations to ConfigDescriptionRegistry and ConfigUtil

wborn:
Also added "org.eclipse.jdt.annotation" to the test BOM so we can use "org.eclipse.jdt.annotation.Checks" in itests.
That class has many useful methods that help with writing more readable test code when using the Eclipse JDT null analysis annotations.

After running the resolver on the itests  a lot of bundles were removed from the itest.bndrun files.


cweitkamp (approved):
LGTM. Thanks.

There are many changes in the `itest.bndrun` files. I did not expect that.

wborn:
> There are many changes in the itest.bndrun files. 

I wasn't happy to see them being added by the resolver in https://github.com/openhab/openhab-core/pull/1580.
So I'm happy to see them being removed again.
It may improve build times because all those bundles won't have to be started.

Issue #1637, opened 2020-09-12
DSL rules with "System started" condition not executed


wborn:
When using the [demo.rules](https://github.com/openhab/openhab-distro/blob/0d9e3273e42bbd907c18bb99d22acda56ce805b5/features/distro-resources/src/main/resources/rules/demo.rules#L75-L129) file the rules with the "System started" don't seem to be executed at startup.
They do get executed when I manually trigger the rules from the UI.

It's most likely caused by https://github.com/openhab/openhab-core/pull/1451

J-N-K:
„System started“ is not available in JavaScript either. Not sure about Jython.

openhab-5iver:
This was moved to #1509 and should make the 'System started' trigger available for all languages which has always been available when using the Jython helper libraries.

wborn:
It's probably due to the disclaimer:

> disclaimer: No start level events are defined in this PR yet, so the trigger itself is still rather useless

kaikreuzer:
@wborn I have tried the demo setup (changed logging to INFO for rule events) and I see at least some rules being triggered at startup:
```
2020-09-12 15:29:32.462 [INFO ] [smarthome.event.StartlevelEvent     ] - Startlevel '20' reached.
2020-09-12 15:29:35.749 [INFO ] [smarthome.event.RuleAddedEvent      ] - Rule 'demo-1' has been added.
2020-09-12 15:29:35.749 [INFO ] [smarthome.event.RuleStatusInfoEvent ] - demo-1 updated: UNINITIALIZED
2020-09-12 15:29:35.750 [INFO ] [smarthome.event.RuleStatusInfoEvent ] - demo-1 updated: INITIALIZING
2020-09-12 15:29:35.754 [INFO ] [smarthome.event.RuleStatusInfoEvent ] - demo-1 updated: IDLE
2020-09-12 15:29:35.755 [INFO ] [smarthome.event.RuleAddedEvent      ] - Rule 'demo-2' has been added.
2020-09-12 15:29:35.755 [INFO ] [smarthome.event.RuleStatusInfoEvent ] - demo-2 updated: UNINITIALIZED
2020-09-12 15:29:35.755 [INFO ] [smarthome.event.RuleStatusInfoEvent ] - demo-2 updated: INITIALIZING
2020-09-12 15:29:35.756 [INFO ] [smarthome.event.RuleStatusInfoEvent ] - demo-2 updated: IDLE
[...]

PR #1649, opened 2020-09-17
Increase StartlevelEvent delay to 15 seconds

wborn:
The rules are often not instantiated within 5 seconds.
On a Raspberry Pi 3B it can take 12 seconds before rules are instantiated.
The code has also been improved with proper null annotations.

Related to #1637

### DIFF
--- a/bundles/org.openhab.core.model.rule/src/org/openhab/core/model/rule/jvmmodel/RulesRefresher.java
+++ b/bundles/org.openhab.core.model.rule/src/org/openhab/core/model/rule/jvmmodel/RulesRefresher.java
@@ -17,7 +17,8 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 
-import org.eclipse.jdt.annotation.NonNull;
+import org.eclipse.jdt.annotation.NonNullByDefault;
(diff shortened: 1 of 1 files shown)


cpmeister (approved):
LGTM

bundles/org.openhab.core.model.rule/src/org/openhab/core/model/rule/jvmmodel/RulesRefresher.java
```diff
@@ -50,30 +51,65 @@
...
-    private ThingRegistryChangeListener thingRegistryChangeListener;
+    private final ReadyMarker marker = new ReadyMarker("dsl", RULES_REFRESH);
 
     private final ModelRepository modelRepository;
     private final ItemRegistry itemRegistry;
     private final ThingRegistry thingRegistry;
     private final EventPublisher eventPublisher;
     private final ReadyService readyService;
 
+    private final ItemRegistryChangeListener itemRegistryChangeListener = new ItemRegistryChangeListener() {
```
kaikreuzer, line 76: Oh, that's a worthwhile improvement - creating a new listener on every marker event was indeed a very bad idea. I'd bet that this change alone will make the RPi to only need 5 secs to be ready for rules now ;-)

kaikreuzer (approved):
Thanks!

PR #1652, opened 2020-09-18
Make logger transient

### LINKED ISSUE #1651: QuantityType not Serializable because of member variable 'logger'
The Class QuantityType is obviously supposed to be Serializable, because it contains a SerialVersionUID. Since it extends Number, it implicitly inherits Serializable interface from this class.
BUT: it cannot be serialized because it may contain a member-reference to non-serializable org.ops4j.pax.logging.slf4j.Slf4jLogger at runtime.

`...`
`public class QuantityType<T extends Quantity<T>> extends Number`
`    implements PrimitiveType, State, Command, Comparable<QuantityType<T>> {`
`private final Logger logger = LoggerFactory.getLogger(QuantityType.class);`
` `
`private static final long serialVersionUID = 8828949721938234629L;`
`...`

Please make the variable ‘logger’ static or transient

I am using stable Version 2.5.8-1

kaikreuzer:
This makes QuantityType serialisable and thus fixes #1651


cweitkamp (approved):
LGTM. Thanks.

PR #1664, opened 2020-09-22
Use newer javax.measure library

cpmeister:
I'm currently working on adding proper GATT characteristic support for the official bluetooth binding and found that the bluetooth specification uses a ton of units that openHAB doesn't currently support. It seems that the `uom-se` library we are using is no longer maintained so I'd like to switch over to using the newer `indriya` library. It contains a lot of the missing units and dimensions that the bluetooth feature would require.

This change includes some of the missing dimensions such as `AngularVelocity` and `MassFlowRate`.

There were some small hiccups from the swapping out the libraries that I found:
1. Quantity not longer implements an `equals` method, so the unit tests had to be adjusted. I don't think that this should have any impact outside of the unit tests though since the rest of openHAB should be using QuantityType instead.
2. RationalConverter is not package private, so instances of it much be created through the MultiplyConverter static functions.
3. Quantities.getQuantity can no longer parse values without units like `100`. I've had to implement a workaround.
4. The unicode greek `mu` letter is now returned for unit prefixes instead of the unicode `micro` character. These characters are visually identical but the unit tests had to be adjusted. The new library seems to parse both types just fine.

This PR doesn't include all of the units and quantity types that bluetooth will eventually use, but I'll add those in as a followup PR.


cpmeister:
I'm still getting a `Feature resolution failed` message when trying to do a full build, so if someone can point out what else I need to change that would be great. Thanks!

cpmeister:
ok, I think I've gotten a better understanding of how stupidly hard it is to change any dependencies in openhab-core. Just to change the version of 'javax.measure' would require modifying 21 different files. That doesn't even solve why my build can't find the new dependencies as part of building `org.openhab.core.features.karaf.openhab-core`. I think I'm stuck at this point and frankly I'm considering changing this PR to just copy all of the source code from the libraries directly into this repository because it would be a lot faster and easier at this point!

cpmeister:
Wrapping the bundles doesn't seem to do any good on my local. I still end up with missing package errors.

wborn:
Hi @cpmeister! I've created #2319 based on your nice work in this PR.
Since there is still no new si-units release that exports the package, I've created one that does with openhab/openhab-osgiify#19.

PR #1671, opened 2020-09-24
Remove remaining deprecated MQTT transport code

wborn:
This removes the remaining deprecated MQTT transport code which was not part of #1668.

The `retainMessages` configuration parameter in the `MqttBrokerConnectionConfig` is already unused by MQTT Broker Things (see [config description](https://github.com/openhab/openhab-addons/blob/cc042dabff6cb4bc3fabc3f3264d2e520d7cd46b/bundles/org.openhab.binding.mqtt/src/main/resources/OH-INF/thing/thing-types.xml#L11)). So keeping track of it and using it serves no purpose because the value would always be `false`.

Related to #1408


kaikreuzer (approved):
thanks!

PR #1673, opened 2020-09-25
[archetype] Use project.version iso ohc.version in skeleton feature.xml

Hilbrand:


kaikreuzer:
Why this change? `ohc.version` [is defined](https://github.com/openhab/openhab-addons/blob/main/pom.xml#L71) in openhab-addons and this place references a core artifact - so using `ohc.version`is imho correct.

Hilbrand:
Hmm. I don't know. Somehow I had in mind this was a temporary thing. But I can't find or remember how I came to that conclusion (There needs to be a workaround removed from the skeleton scripts in addons repo related to this, I just found out). I added this change to the guide on the forum too. But I'll change that. I do actually think having the option to set ohc.version is indeed convenient so not having to change this is great.

kaikreuzer:
Atm we are back to releasing ohc and addons in sync, but who knows how long we keep this. I'd therefore think it makes sense to have the option to decouple both in place, if needed.

PR #1679, opened 2020-10-01
Fix sitemap subscription cleanup

### LINKED ISSUE #1674: [rest] SSE subscription clean-up not working
Stale SSE subscriptions are not removed, which causes openHAB after a while to not accept any new subscriptions - see https://github.com/openhab/openhab-android/issues/2215 for details.

wborn:
* Add listener to SseBroadcaster so sseEventSinkRemoved events are handled
* Use Instant instead of long for tracking subscription creation times
* Run cleanup every 2 minutes instead of every 5 minutes

Fixes #1674

### DIFF
--- a/bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/SitemapSubscriptionService.java
+++ b/bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/SitemapSubscriptionService.java
@@ -12,6 +12,8 @@
  */
 package org.openhab.core.io.rest.sitemap;
 
+import java.time.Duration;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


kaikreuzer (approved):
You're the best, thanks!
I'll update my prod system right away and should be able to tell you the result by tomorrow :-)

wborn:
Great! Then we still have a bit of time to do something about it before the M1 is released. :gift: 
I still don't run OH3 in production, if this is the last major bug impacting functionality I'd be tempted. :wink:

wborn:
Did it fix the issue for you @kaikreuzer?

kaikreuzer:
It indeed did! After more than 48 hours running, I just hit it with my Android app and see
```
15:56:55.354 [DEBUG] [st.sitemap.SitemapSubscriptionService] - Created new subscription with id 39c9b8e3-8e69-4148-8afc-a5c0249a4125 (3 active subscriptions for a max of 50)
```
-> unused subscriptions are nicely cleaned up now! Thanks again 👍

wborn:
Thanks for the positive feedback! :partying_face:

Issue #1662, opened 2020-09-21
Issues in Regional Settings


MHerbst:
In my first tests of the new UI I have found some issues (tested in Chrome Browser on Windows 10 with german language):

Language selection contains two empty entries at the beginning:
![image](https://user-images.githubusercontent.com/616792/93811897-22865180-fc51-11ea-9e42-d161a512caca.png)

BTW: is it really necessary to have languages like "Altpreußisch", "Kölsch" etc. in the list?

Country/Region also contains two empty entries:
![image](https://user-images.githubusercontent.com/616792/93812028-5c575800-fc51-11ea-9d40-7b9b81b7054c.png)

Below the Country/Region is this text:
![image](https://user-images.githubusercontent.com/616792/93812091-709b5500-fc51-11ea-93e1-06a1eaeef486.png)
Seems to be copied from the old Paper UI, but in my opinion it does not make sense here.

"Variant" has also two empty entries at the beginning of the list.

Use "Encoding" instead of "Script":
The name "Script" seems unusual to me. I have never seen it in this context. In my opinion "Encoding" would be better.

kaikreuzer:
Done.

J-N-K:
As far as I can see the double-empty entry is no longer present. Since the empty entry is added by UI, labelling it must be added there (if necessary).

PR #1680, opened 2020-10-02
Filter empty i18n config parameter options

wborn:
There are many locales that do not have a country resulting in these empty parameter options.

Related to #1662

### DIFF
--- a/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/i18n/I18nConfigOptionsProvider.java
+++ b/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/i18n/I18nConfigOptionsProvider.java
@@ -96,7 +96,11 @@ public class I18nConfigOptionsProvider implements ConfigOptionProvider {
 
     private Collection<ParameterOption> getAvailable(@Nullable Locale locale,
             Function<Locale, ParameterOption> mapFunction) {
-        return Arrays.stream(Locale.getAvailableLocales()).map(l -> mapFunction.apply(l)).distinct()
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/i18n/I18nConfigOptionsProvider.java
```diff
@@ -96,7 +96,11 @@ private static String getTimeZoneRepresentation(TimeZone tz) {
 
     private Collection<ParameterOption> getAvailable(@Nullable Locale locale,
             Function<Locale, ParameterOption> mapFunction) {
-        return Arrays.stream(Locale.getAvailableLocales()).map(l -> mapFunction.apply(l)).distinct()
-                .sorted(Comparator.comparing(a -> a.getLabel())).collect(Collectors.toList());
+        return Arrays.stream(Locale.getAvailableLocales()) //
+                .map(l -> mapFunction.apply(l)) //
```
cpmeister, line 100:
```suggestion
                .map(mapFunction) //
```

cpmeister (changes requested):

bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/i18n/I18nConfigOptionsProvider.java
wborn: That looks a bit better indeed!

kaikreuzer (approved):

PR #1681, opened 2020-10-02
Sort Ephemeris config parameter options

### LINKED ISSUE #1663: Issues in Ephemeric Settings
Tested with Chrome Browser on Windows 10 with german language):

Strange order of the country list:
image

At first I thought the list was sorted backwards, but it looks like a random order.

The same happens in the "Region" selection list:
image

BTW: both list have on empty entry at the beginning.

wborn:
Fixes #1663

### DIFF
--- a/bundles/org.openhab.core.ephemeris/src/main/java/org/openhab/core/ephemeris/internal/EphemerisManagerImpl.java
+++ b/bundles/org.openhab.core.ephemeris/src/main/java/org/openhab/core/ephemeris/internal/EphemerisManagerImpl.java
@@ -27,6 +27,7 @@ import java.time.format.TextStyle;
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.ephemeris/src/main/java/org/openhab/core/ephemeris/internal/EphemerisManagerImpl.java
```diff
@@ -112,12 +113,21 @@ public EphemerisManagerImpl(final @Reference LocaleProvider localeProvider) {
             final Properties properties = new Properties();
             properties.load(stream);
             properties.forEach(this::parseProperty);
+
+            sortByLabel(countries);
```
kaikreuzer, line 117: I had difficulties figuring out that the line above actually fills the Lists of the Maps with values by calling `parseProperty` and that those Lists are `ArrayList` instances that are sortable. Imho it is a pretty unexpected side-effect of a method with the name `parseProperty`...

But that's not the fault of your PR, so all fine for merging this!

kaikreuzer (approved):
lgtm, thanks!

bundles/org.openhab.core.ephemeris/src/main/java/org/openhab/core/ephemeris/internal/EphemerisManagerImpl.java
wborn: I would also not expect these side effects as the current name suggests it's some kind of trivial operation. ;-)

PR #1682, opened 2020-10-02
Synced guava version with openhab-tp usages

Hilbrand:
I did a fresh install of openHAB 3.0 and ended up 2 activated guava bundles. While 1 is already to much :wink: 
```
 28 │ Active │  80 │ 18.0.0                  │ Guava: Google Core Libraries for Java
 29 │ Active │  80 │ 27.1.0.jre              │ Guava: Google Core Libraries for Java
```
So I tracked down to this dependencies.

My only question. Shouldn't it be removed completely here actually?

### DIFF
--- a/features/karaf/openhab-core/src/main/feature/feature.xml
+++ b/features/karaf/openhab-core/src/main/feature/feature.xml
@@ -420,7 +420,6 @@
 	</feature>
 
 	<feature name="openhab-runtime-base" description="openHAB Runtime Base" version="${project.version}">
-		<bundle>mvn:com.google.guava/guava/18.0</bundle>
 		<requirement>openhab.tp;filter:="(feature=commons-net)"</requirement>
 		<feature dependency="true">openhab.tp-commons-net</feature>
 		<feature>openhab-core-base</feature>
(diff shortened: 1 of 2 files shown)


wborn:
It used to be a dependency of the old Swagger version we used, see https://github.com/openhab/openhab-core/issues/892#issuecomment-504784241.
But that should no longer be an issue since it got upgraded in https://github.com/openhab/openhab-core/pull/1576.

wborn:
> My only question. Shouldn't it be removed completely here actually?

I had a look and Karaf feature validation still succeeds when it's removed so I think it can be safely removed.

kaikreuzer:
Wasn't it also required by Xtext in some way? I don't remember, but we can try to completely remove it and see what happens 😎 .

wborn:
Yes it's still required for Xtext, but it is part of that feature:

https://github.com/openhab/openhab-core/blob/3b16f2ce7c9ef73377b0094e1673c1360bc5e45a/features/karaf/openhab-tp/src/main/feature/feature.xml#L216-L217

It looks like it can also be removed from the Swagger feature:

https://github.com/openhab/openhab-core/blob/3b16f2ce7c9ef73377b0094e1673c1360bc5e45a/features/karaf/openhab-tp/src/main/feature/feature.xml#L255-L256

wborn (approved):
Thanks!

PR #1683, opened 2020-10-02
Upgrade Maven compiler and fix/upgrade some TP dependencies

wborn:
Upgrades the compiler and its dependencies so the compiler results of Maven builds are more similar to those generated in recent Eclipse versions.

To fix compilation issues in Eclipse for add-ons using classes from javax.xml.stream several dependencies were upgraded/excluded.

---

Upgrading the compiler and its dependencies may also help solving/debugging errors like in this [build](https://ci.openhab.org/job/openHAB-Addons/73/console):

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile (default-compile) on project org.openhab.binding.alarmdecoder: Compilation failure: Compilation failure:
[ERROR] [ecj] The compiler reported an error but has not written it to its logging
[ERROR] [ecj] The following line(s) might indicate the issue:
[ERROR] at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
[ERROR] at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
[ERROR] at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
[ERROR] at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
[ERROR] at java.base/java.lang.Thread.run(Thread.java:834)
[ERROR] -> [Help 1]
````

### DIFF
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -252,10 +252,6 @@
           <groupId>jakarta.activation</groupId>
           <artifactId>jakarta.activation-api</artifactId>
         </exclusion>
-        <exclusion>
-          <groupId>jakarta.xml.bind</groupId>
-          <artifactId>jakarta.xml.bind-api</artifactId>
-        </exclusion>
       </exclusions>
     </dependency>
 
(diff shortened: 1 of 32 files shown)


kaikreuzer (approved):

kaikreuzer:
Seems this made Lametric not compile anymore: https://ci.openhab.org/view/Sandbox/job/sandbox-openhab3-release/8/execution/node/117/log/

`javax.activation' is no longer part of the tp, so at least for compilation we'd have to add it as a dependency to this binding, is that correct?

wborn:
Yes that why I added that dependency in https://github.com/openhab/openhab-addons/pull/8636 .

I've excluded the activation dependency from jaxb 2.3 because due to transitivity it would be pushed to add-ons that still depend on activation 1.1. E.g. the mail  binding does not work with 1.2 but it does work with 1.1. There were also no newer versions of their dependencies to get rid of 1.1. Excluding activation 1.2 from transitive dependencies in the add-ons POMs did not work well either.

kaikreuzer:
Ah, great - sorry, I had completely missed that you had also created https://github.com/openhab/openhab-addons/pull/8636 :-) .

Hilbrand:
Is there any relation here on why I need to add `org.apache.servicemix.specs.activation-api-1.1` and `org.apache.servicemix.specs.jaxb-api-2.2` to the demo pom.xml and add it to the app.bndrun to avoid resolve errors? Or I'm just running behind?

wborn:
Updating the demo app with openhab/openhab-distro#1160 was still on my todo list.

PR #1684, opened 2020-10-02
Upgrade to Jollyday 0.5.10

wborn:
With some excludes and the upgraded JAXB I got the most recent Jollyday version working.
The integration tests also succeed again on Java 11.


kaikreuzer (approved):
Nice!

PR #1690, opened 2020-10-04
Remove actions from available add-on types

ghys:
Since there are no more action add-ons in the distribution, I'd suggest to remove these from the `KarafAddonService` so they won't appear in the UI menus.


kaikreuzer (approved):
Makes a lot of sense, thanks!

PR #1700, opened 2020-10-07
Make ExecUtil more robust

### LINKED ISSUE #1677: ExecUtil.executeCommandLine() is broken
I assume that the semantics of e.g. `ExecUtil.executeCommandLine("ls .")` is to execute the command "ls" with a parameter ".".

(There is some sort of "extra safety" variant where I can pass "ls@@." and it will tokenize it etc. to workaround some "issues" on Mac, as it says. Looks a bit smelly, but anyways.)

The problem is when the caller does NOT use the "@@" trick. Passing "ls ." (i.e. where regular whitespace is used) *will not do the right thing anymore*, because someone has switched the implementation to ProcessBuilder on master several months ago.  ProcessBuilder (unlike Runtime.exec() which was seemingly used in 2.5) *does not* accept a single command-line string where the command and its arguments are separated by whitespace. 

Calling new `ProcessBuilder(cmdLine)` deceptively seems to work because the class does have a constructor with `(String... commands)` signature so it takes a single `String` parameter, but that parameter is supposed to be the command ONLY, not the whole command-line.

At least one binding in 3.0 (SamsungTV) seems to be broken because of this issue.

cpmeister:
Ever since I first read this code I've been waiting for an excuse to rewrite it.
I went ahead and got rid of the `@@` delimiter usage since it seems silly to not just take the arguments as an additional parameter.

Fixes #1677 

### DIFF
--- a/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/exec/ExecUtil.java
+++ b/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/exec/ExecUtil.java
@@ -35,7 +35,7 @@ import org.slf4j.LoggerFactory;
  *
  * @author Pauli Anttila - Initial contribution
  * @author Kai Kreuzer - added exception logging
- * @author Connor Petty - rewrote
+ * @author Connor Petty - replaced delimiter usage with argument array
  */
 @NonNullByDefault
 public class ExecUtil {
@@ -68,9 +68,9 @@ public class ExecUtil {
      * <p>
      * A possible {@link IOException} gets logged but no further processing is done.
      *
-     * @param timeout timeout for execution in milliseconds, 0 to wait indefinitely
+     * @param timeout the max time to wait for a process to finish, 0 to wait indefinitely
      * @param commandLine the command line to execute
-     * @return response data from executed command line or <code>null</code> if an error occurred
+     * @return response data from executed command line or <code>null</code> if a timeout or error occurred
      */
     public static @Nullable String executeCommandLineAndWaitResponse(@Nullable Duration timeout,
             String... commandLine) {
--- a/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/exec/ExecUtil.java
+++ b/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/exec/ExecUtil.java
@@ -18,13 +18,11 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.StringWriter;
 import java.lang.ProcessBuilder.Redirect;
-import java.util.List;
+import java.time.Duration;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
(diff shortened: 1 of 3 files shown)


wborn:
I also never like the usage of `@@` and think it is OK to fix (and break) this with OH3.

The  `org.openhab.core.model.script.actions.Exec` class would also need to be updated as scripts will no longer have a way to pass parameters and users will need to adapt their rules for this.

There are also several add-ons that need to be updated for this as well as documentation.

I think it would also be better when the command and parameters always follow each other. That means when specifying a timeout it should be specified as first parameter instead of being in between the command and parameters.

wborn:
Being a big fan of java.time I think it would also be more clear to specify the timeout using a `Duration` so nobody has to look up in docs what the unit is and convert from seconds to milliseconds. :wink:

cpmeister:
I've gone ahead and made the recommended changes, the only reason I didn't do it the first time is because I didn't want to break existing code. But if you are ok breaking some code then I'd be more than happy to.

From a quick search it looks like only 4 addons are using the ExecUtil right now so fixing those be very easy.

bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/exec/ExecUtil.java
```diff
@@ -28,87 +35,101 @@
...
-     * Executes <code>commandLine</code> and return its result. Sometimes (especially observed on MacOS) the commandLine
-     * isn't executed properly. In that cases another exec-method is to be used. To accomplish this please use the
-     * special delimiter '<code>@@</code>'. If <code>commandLine</code> contains this delimiter it is split into a
-     * String list and the special exec-method is used.
+     * Executes <code>commandLine</code> and return its result.
      *
      * <p>
      * A possible {@link IOException} gets logged but no further processing is done.
      *
+     * @param timeout timeout for execution in milliseconds, 0 to wait indefinitely
```
wborn, line 71: This JavaDoc still needs to be updated for the latest changes.

bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/exec/ExecUtil.java
```diff
@@ -28,87 +35,101 @@
  *
  * @author Pauli Anttila - Initial contribution
  * @author Kai Kreuzer - added exception logging
+ * @author Connor Petty - rewrote
```
wborn, line 38: Can you  be a bit more specific?

wborn (commented):
Great! Can you also create PR for the add-ons so we can limit the time builds are broken?

paulianttila:
> I also never like the usage of @@

Me either, but as an excuse, it was selected to be exec1 binding syntax compatible ;)

https://github.com/openhab/openhab1-addons/tree/master/bundles/binding/org.openhab.binding.exec

bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/exec/ExecUtil.java
cpmeister: done

bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/exec/ExecUtil.java
cpmeister: done

cpmeister:
> Great! Can you also create PR for the add-ons so we can limit the time builds are broken?

Done: https://github.com/openhab/openhab-addons/pull/8690

bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/exec/ExecUtil.java
```diff
@@ -28,87 +35,101 @@
...
-     * Executes <code>commandLine</code> and return its result. Sometimes (especially observed on MacOS) the commandLine
-     * isn't executed properly. In that cases another exec-method is to be used. To accomplish this please use the
-     * special delimiter '<code>@@</code>'. If <code>commandLine</code> contains this delimiter it is split into a
-     * String list and the special exec-method is used.
+     * Executes <code>commandLine</code> and return its result.
      *
      * <p>
      * A possible {@link IOException} gets logged but no further processing is done.
      *
+     * @param timeout the max time to wait for a process to finish, 0 to wait indefinitely
```
wborn, line 71: The timeout is only used when it is not null:

```suggestion
     * @param timeout the max time to wait for a process to finish, null to wait indefinitely
```

cpmeister: oof, forgot that one.
Fixed

wborn (approved):
Thanks!

PR #1704, opened 2020-10-08
[REST] Added validation for relation between ThingUID and BridgeUID

cweitkamp:
- Check for valid `ThingUID` when creating a `Thing` which is related to a `Bridge`

Similar to https://github.com/openhab/openhab-core/pull/1481

My first idea was to add the check inside [ThingRegistryImpl.createThingOfType(ThingTypeUID, ThingUID, ThingUID, String, Configuration)](https://github.com/openhab/openhab-core/blob/c4b76a0ad11d5098f0f48bb3f86b0f943afc23cf/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingRegistryImpl.java#L245) but I was not sure if it is the best idea to throw an IAE if the given `ThingUID` does not match as it will be API breaking. 

Maybe we should not merge it until we have a valid solution for https://github.com/openhab/openhab-webui/issues/380.


kaikreuzer (approved):
Thanks!

PR #1713, opened 2020-10-11
Allow basic authentication to authorize API access

### LINKED ISSUE #1699: OAuth2 makes it difficult to consume the REST API
OAuth2 workflow was introduced in #1388.

Whilst it might make sense in an untrusted environment or in an environment with lots of users both are not the scope of openhab.
The scope is **home** automation which means a trusted environment.
I'd even argue that 99.99% of all openhab instance and all of its devices are fully managed by just one person so there is nothing to gain by expiring tokens.
I am not against authentication but Oauth2 is way overblown and adds nothing that can't be achieved with basic auth (for a trusted home environment!). It also adds a huge overhead (only 1% of a message is payload, the rest is auth) making the API even slower.

There are countless people that are already struggling with the Rest API (just look at the forum) and OAuth2 will effectively kill Rest for the majority of users.
The easy integration into scripts and other already existing things through the API always has been a strength of OH and will be mitigated if not removed by OAuth2.

I'd argue that OAuth2 should be removed again but you guys seem to really like doing authentication so please - for the sake of the normal users - add an configuration option which can disable OAuth2.

ghys:
Closes #1699.

Note, this opens a minor security issue that allows an attacker
to brute force passwords by making calls to the API - contrary to
the authorization page, the credentials parsing for the REST API
is stateless & doesn't have a lock mechanism to lock user accounts
after too many failed login attempts.

### DIFF
--- a/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
+++ b/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
@@ -67,10 +67,10 @@ public class AuthFilter implements ContainerRequestFilter {
 
     protected static final String CONFIG_URI = "system:restauth";
     private static final String CONFIG_ALLOW_BASIC_AUTH = "allowBasicAuth";
-    private static final String CONFIG_NO_IMPLICIT_USER_ROLE = "noImplicitUserRole";
+    private static final String CONFIG_IMPLICIT_USER_ROLE = "implicitUserRole";
 
     private boolean allowBasicAuth = false;
-    private boolean noUnauthenticatedUserRole = false;
+    private boolean implicitUserRole = true;
 
     @Reference
     private JwtHelper jwtHelper;
@@ -88,8 +88,8 @@ public class AuthFilter implements ContainerRequestFilter {
         if (properties != null) {
             Object value = properties.get(CONFIG_ALLOW_BASIC_AUTH);
             allowBasicAuth = value != null && "true".equals(value.toString());
-            value = properties.get(CONFIG_NO_IMPLICIT_USER_ROLE);
-            noUnauthenticatedUserRole = value != null && "true".equals(value.toString());
+            value = properties.get(CONFIG_IMPLICIT_USER_ROLE);
+            implicitUserRole = value != null && "false".equals(value.toString());
         }
     }
 
@@ -137,7 +137,7 @@ public class AuthFilter implements ContainerRequestFilter {
                 return;
             }
 
-            if (!noUnauthenticatedUserRole) {
+            if (implicitUserRole) {
                 requestContext.setSecurityContext(new AnonymousUserSecurityContext());
             }
 
--- a/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
+++ b/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
@@ -14,6 +14,7 @@ package org.openhab.core.io.rest.auth.internal;
 
 import java.io.IOException;
 import java.util.Base64;
+import java.util.Map;
 
 import javax.annotation.Priority;
 import javax.security.sasl.AuthenticationException;
@@ -26,17 +27,24 @@ import javax.ws.rs.core.Response.Status;
 import javax.ws.rs.core.SecurityContext;
 import javax.ws.rs.ext.Provider;
 
+import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.auth.Authentication;
 import org.openhab.core.auth.User;
 import org.openhab.core.auth.UserRegistry;
 import org.openhab.core.auth.UsernamePasswordCredentials;
+import org.openhab.core.config.core.ConfigurableService;
 import org.openhab.core.io.rest.JSONResponse;
 import org.openhab.core.io.rest.RESTConstants;
+import org.osgi.framework.Constants;
+import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Modified;
 import org.osgi.service.component.annotations.Reference;
 import org.osgi.service.jaxrs.whiteboard.JaxrsWhiteboardConstants;
 import org.osgi.service.jaxrs.whiteboard.propertytypes.JaxrsApplicationSelect;
 import org.osgi.service.jaxrs.whiteboard.propertytypes.JaxrsExtension;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * This filter is responsible for parsing credentials provided with a request, and hydrating a {@link SecurityContext}
@@ -46,19 +54,44 @@ import org.osgi.service.jaxrs.whiteboard.propertytypes.JaxrsExtension;
  * @author Yannick Schaus - Allow basic authentication
  */
 @PreMatching
-@Component
+@Component(configurationPid = "org.openhab.restauth", property = Constants.SERVICE_PID + "=org.openhab.restauth")
+@ConfigurableService(category = "system", label = "API Security", description_uri = AuthFilter.CONFIG_URI)
 @JaxrsExtension
 @JaxrsApplicationSelect("(" + JaxrsWhiteboardConstants.JAX_RS_NAME + "=" + RESTConstants.JAX_RS_NAME + ")")
 @Priority(Priorities.AUTHENTICATION)
 @Provider
 public class AuthFilter implements ContainerRequestFilter {
+    private final Logger logger = LoggerFactory.getLogger(AuthFilter.class);
+
     private static final String ALT_AUTH_HEADER = "X-OPENHAB-TOKEN";
+
+    protected static final String CONFIG_URI = "system:restauth";
+    private static final String CONFIG_ALLOW_BASIC_AUTH = "allowBasicAuth";
+    private static final String CONFIG_NO_UNAUTHENTICATED_USER_ROLE = "noUnauthenticatedUserRole";
+
+    private boolean allowBasicAuth = false;
+    private boolean noUnauthenticatedUserRole = false;
 
     @Reference
     private JwtHelper jwtHelper;
 
     @Reference
     private UserRegistry userRegistry;
+
+    @Activate
+    protected void activate(Map<String, Object> config) {
+        modified(config);
+    }
+
+    @Modified
+    protected void modified(@Nullable Map<String, @Nullable Object> properties) {
+        if (properties != null) {
+            Object value = properties.get(CONFIG_ALLOW_BASIC_AUTH);
+            allowBasicAuth = value != null && "true".equals(value.toString());
+            value = properties.get(CONFIG_NO_UNAUTHENTICATED_USER_ROLE);
+            noUnauthenticatedUserRole = value != null && "true".equals(value.toString());
+        }
+    }
 
     @Override
     public void filter(ContainerRequestContext requestContext) throws IOException {
@@ -72,6 +105,9 @@ public class AuthFilter implements ContainerRequestFilter {
                         requestContext.setSecurityContext(new JwtSecurityContext(auth));
                         return;
                     } else if ("Basic".equals(authParts[0])) {
+                        if (!allowBasicAuth) {
+                            throw new AuthenticationException("Basic authentication is not allowed");
+                        }
                         try {
                             String[] decodedCredentials = new String(Base64.getDecoder().decode(authParts[1]), "UTF-8")
                                     .split(":");
@@ -88,7 +124,7 @@ public class AuthFilter implements ContainerRequestFilter {
                             requestContext.setSecurityContext(new UserSecurityContext(user, "Basic"));
                             return;
                         } catch (org.openhab.core.auth.AuthenticationException e) {
-                            throw new AuthenticationException("Invalid user name or password");
+                            throw new AuthenticationException("Invalid Basic authentication credentials", e);
                         }
                     }
                 }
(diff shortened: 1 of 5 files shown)


kaikreuzer:
I like how you do not go into lengthy discussions, but rather produce code ;-)
Will have to look at what implications this PR actually has before merging - let's get M1 first out of the door!

ghys:
Sure, this one was more to prove a point than anything else, really... :)

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -63,6 +71,25 @@ public void filter(ContainerRequestContext requestContext) throws IOException {
                         Authentication auth = jwtHelper.verifyAndParseJwtAccessToken(authParts[1]);
                         requestContext.setSecurityContext(new JwtSecurityContext(auth));
                         return;
+                    } else if ("Basic".equals(authParts[0])) {
```
splatch, line 74: A small suggestion here - it might be worth to use `equalsIgnoreCase` as some tools can send auth type lowercase.

ghys:
I was thinking, maybe this could be made optional with a configurable service "REST API Authentication > Allow Basic Authentication", and disabled by default. Most people wouldn't need it, unless they want to perform admin operations from external tools without doing the OAuth2 dance.

Another option could be "Protect User Operations", also disabled by default. Currently these operations aren't protected at all: getting items, sending commands, the reason being that would probably break most end-user UIs. But if we enable it and make it advanced, the UI developers can enable it to help developing support. I have to check, but I believe the main UI can work with this option enabled, so you can't lock yourself out.

sikksakk:
How about a "host" whitelist? I.e. I suspect most people who would need to
run scripts against the REST api, would be from localhost, or other
"things" in the local area network that could be allowed to bypass the
authentication.

/Chris :-)

man. 12. okt. 2020 kl. 12:01 skrev Yannick Schaus <notifications@github.com

spacemanspiff2007:
I'd rather see the rules (e.g. admin, user) implemented consistently so they work independent of auth type (Oauth, BasicAuth).
With basic auth available using it really is easy and whitelisting hosts would prevent role based operations.

kaikreuzer:
> maybe this could be made optional with a configurable service "REST API Authentication > Allow Basic Authentication", and disabled by default.

I think an "Authentication" (or "Security"?) System Services entry will be helpful in any case.
And yes, I'd agree that basic auth should be disabled by default, since from a security perspective everything that is not required for the standard operation of the system should not be active as it is a potential vulnerability.

> Another option could be "Protect User Operations", also disabled by default.

Yes, it would be at least a first step towards a fully secured setup. 👍 

Adding whitelisting etc. can then also be future options for this new configurable service.

Would you want to add such a configurable service (with a switch for basic auth) as part of this PR?

9037568:
>I know it's annoying that I stress this so much because it's already implemented but I firmly believe OAuth will only create problems for the users of openhab whilst providing hardly any benefits.

Absolutely.

ghys:
@kaikreuzer I added the options.

image

I also contemplated adding the support for API tokens in this PR (they would prevent external tools from having to perform the OAuth2 flow to get a token, while not requiring to enable Basic auth) and Karaf commands as mentioned here: https://github.com/openhab/openhab-webui/issues/332#issuecomment-708545356
But perhaps it would be best to open another PR for that, I think I can do it before the next milestone.

Btw, not particularly happy to mention this, but this would bring us on par with [what another popular home automation platform](https://developers.home-assistant.io/docs/auth_api/) has implemented and somehow didn't get nearly as much backlash as we see here...

spacemanspiff2007:
> Btw, not particularly happy to mention this, but this would bring us on par with [what another popular home automation platform](https://developers.home-assistant.io/docs/auth_api/) has implemented and somehow didn't get nearly as much backlash as we see here...

We can go into a lengthy discussion what Home Assistant does better (or worse) than Openhab but this will not bring any new factual arguments to the table. Imho the goal should not be to implement Home Assistant in Java but rather to develop well thought and consistent concepts and to build a better smart home system.
Home Assistant does generate 10 year long living tokens by default and you even can name the integration - so it's basically a static user/pw combination.

And while one could say it doesn't matter if you generate a basic auth header or a header with a token it really does.
Many devices allow pushing their state with simple requests to a destination. A basic auth header can be generated with ``user:password`` directly in the url. There is no possibility to modify the header, just the url.
E.g. almost all IP cameras that push the movement detection this way and polling doesn't make sense there.
If you enforce oauth you break integrations without providing any more security and the question still stands why one would do that.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -25,33 +27,72 @@
...
+        modified(config);
+    }
+
+    @Modified
+    protected void modified(@Nullable Map<String, @Nullable Object> properties) {
+        if (properties != null) {
+            Object value = properties.get(CONFIG_ALLOW_BASIC_AUTH);
+            allowBasicAuth = value != null && "true".equals(value.toString());
+            value = properties.get(CONFIG_NO_UNAUTHENTICATED_USER_ROLE);
+            noUnauthenticatedUserRole = value != null && "true".equals(value.toString());
```
kaikreuzer, line 92: Argh, I am really getting a knot in my brain here (German proverb) by this double negation. How about changing this parameter to something like "enableAuthForUserResources"?

kaikreuzer (changes requested):
Awesome, thanks! Just some minor comments.

kaikreuzer:
> But perhaps it would be best to open another PR for that, I think I can do it before the next milestone.

Fully agree, that should be a separate PR.

> We can go into a lengthy discussion what Home Assistant does better (or worse) 

I think @ghys main point here was that there are also very clever guys at HA, who have put a lot of thought into shaping their Auth API and it isn't by accident that they have chosen something different than Basic Auth.

> A basic auth header can be generated with user:password directly in the url.

That's deprecated and should not be used. See https://www.ietf.org/rfc/rfc3986.txt sections 3.2.1 and 7.5.

> almost all IP cameras that push the movement detection this way and polling doesn't make sense there.

IP cams are the best example how to NOT do any secure software. They are hacked together in no time and I have hardly seen more buggy IoT devices out there. They are the backbone of Mirai and other botnets for a reason...

>  I firmly believe OAuth will only create problems for the users of openhab whilst providing hardly any benefits.

This opinion is pretty much against any security best practice of the last decade. Again, Basic Auth has some major flaws that a token based authentication prevents:
- It sends the password as clear text over the wire
- It continuously sends the password with every single request
- If the password is compromised, it has to be changed
- It does not allow any advanced protection against replay attacks, such as TTL, timestamps etc

But never mind - this PR introduces Basic Auth for the ones that really want it despite its shortcomings, so everyone should be served.

spacemanspiff2007:
> That's deprecated and should not be used. See https://www.ietf.org/rfc/rfc3986.txt sections 3.2.1 and 7.5.

I was not aware that it should not be used, yet it still is widely supported.


> This opinion is pretty much against any security best practice of the last decade. Again, Basic Auth has some major flaws that a token based authentication prevents:

I would have really like a constructive discussion like this in the original issue, but I'll comment anyway.
When using a long living token:
- It sends the token as clear text over the wire
- It continuously sends the token with every single request
- If the token is compromised, it has to be changed
- It does not allow any advanced protection against replay attacks, such as TTL, timestamps etc

I still don't see any difference to Basic Auth!

OAuth has become the standard because it allows to delegate access and allows separation of access granting and api infrastructure - a separation we don't have in the case of openhab.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
ghys: It was more to the tune of "prevent unauthenticated requests to get an implicit user role in the security context", so I tried to rename it to "noImplicitUserRole" / "No implicit user role for unauthenticated requests", that's perhaps clearer. Still a double negation, so if you still disagree I'll change it to "enableAuthForUserResources" :)

kaikreuzer: And if we change it to "ImplicitUserRole" and have default=true, knowing that on the long run our wish is to get the setting to OFF? It would be in line with the basic auth setting, i.e. enabling something means your system becomes less secure.

ghys:
> OAuth has become the standard because it allows to delegate access and allows separation of access granting and api infrastructure - a separation we don't have in the case of openhab.

Which is something that one day we might want to do, for instance there's a Synology SSO server (https://global.download.synology.com/download/Document/Software/DeveloperGuide/Package/SSOServer/All/enu/Synology_SSO_API_Guide.pdf) which is also OAuth2; openHAB runs on Synology so it's not that far fetched that we might offer delegating to it for the identity & credentials management.
Some admins might also want to enable a multi-factor authentication provider to open a session, and the sign in/authorize page might offer a way to plug-in such providers.

> Imho the goal should not be to implement Home Assistant in Java but rather to develop well thought and consistent concepts and to build a better smart home system.

Agreed, my remark was not merely whataboutism or emulation at all costs (though there's no shame in borrowing good ideas), but simply pointing out the fact that they've implemented it like this for 2 years, so we might profit from this hindsight and AFAIK there were no major complaints.
So the assessment that it would be a complete nightmare for openHAB users is perhaps more of an opinion than a fact. My guess is that 80% of users won't care about this at all, perhaps 10% will be fine with using non-expiring, special-purpose, scoped tokens to integrate services and the remaining 10% can now enable Basic auth for their instances if they're fine with sharing passwords around.
I'm just getting another idea: allowing Basic auth could be a per-user setting as well, so regular user accounts can have an authentication experience potentially including MFA or SSO, and service accounts can access the API with Basic auth.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
ghys: Good point, I insisted on having false as the default but that's not really a requirement.

kaikreuzer:
> allowing Basic auth could be a per-user setting

That would imho make the configuration pretty complex - I do not see much value in this.

> service accounts can access the API with Basic auth.

Service accounts should use individual tokens (api keys), which are created for them and which can anytime be revoked without requiring the user to change his password.

bundles/org.openhab.core.io.rest.auth/src/main/resources/OH-INF/config/config.xml
```diff
@@ -0,0 +1,25 @@
...
+				and API tokens.</description>
+		</parameter>
+		<parameter name="implicitUserRole" type="boolean" required="false">
+			<advanced>true</advanced>
+			<label>Implicit user role for unauthenticated requests</label>
+			<default>true</default>
+			<description>By default, operations requiring the "user" role are available when unauthenticated. Disabling this
+				option will enforce authorization for these operations. Warning: this will cause clients which don't
+				support
+				authorization to break.</description>
```
kaikreuzer, line 21:
```suggestion
				option will enforce authorization for these operations. Warning: This causes clients that do not
				support authentication to break.</description>
```

kaikreuzer (approved):

bwosborne2:
>Note, this opens a minor security issue that allows an attacker to brute force passwords by making calls to the API - contrary to the authorization page,`

Why should Basic AUTHENTICATION ( identifying WHO you are) 
affect AUTHORIZATION ( what an authenticated entity is PERMITTED to do)?
If you are concerned, add the third part of AAA, ACCOUNTING ( what an entity actually DID)

As part of my daily job I administer RADIUS AAA servers.

bwosborne2:
@ghys  needs to claim a bounty for this.
https://www.bountysource.com/issues/64726789-introduce-basic-auth-on-http-server

ghys:
> Note, this opens a minor security issue that allows an attacker to brute force passwords by making calls to the API - contrary to the authorization page,`

What I was trying to say is, formerly the only place where you could submit your password is in the form on the /auth page, where there are mechanisms to lock out for an increasing number of seconds after a unsuccessful attempt, so trying to guess the password with e.g. a dictionary attack become very unpractical very fast.
Now you can transmit your password in a Basic auth credential, where there's no such mechanism in place, so you can keep hammering the API several times a second until you've guessed the correct password, which can happen very fast if that password is too obvious.
With tokens the risk of that happening is minimal because tokens are not easy to guess.

bwosborne2:
From what I understood, the OpenHAB UIs only access OH through the REST API. The REST Tokens are used for authorization, not authentication. 

Some systems use a complex shared secret for authentication between trusted systems but still require authorization tokens for each use of the API. Others use username / password ( the lockout should be a server function anyway, independent of the client) and then require an authorization token.

You are the architect here, though.

ghys:
Yes, the tokens are generated by already authenticated users (which might have used a much more secure way of doing so, like multi-factor) to _authorize_ the bearer of this token to access the API on their behalf, eventually for a very limited scope.
If you pass the username/password to the API directly, like in Basic auth, you authenticate and authorize the user at the same time.

CrazyIvan359:
Just my 2c, I agree with @ghys and @spacemanspiff2007 on the pros/cons of tokens and basic auth, but I think tokens fit better for service access. All over the web you see mounting use of SSO+2FA to authenticate users, but as Yannick says, services get a token to access the API on behalf of the user. Both are transmitted in the clear (unless you are running https, which I am via nginx but still have local access on http) but with basic auth you have a username and password, both of which users (myself included tisk tisk) are very likely to reuse. 

Add together that and the lack of throttling on auth attempts via basic auth, inside your average home network. Seems like a not ideal setup given the scope of access some installations have to hardware in people's homes. I'm no security expert, but I can't see the firewall on a $50 router being rock solid, never mind the cheap modem/router you get from most ISPs.

I'd rather provide a token that is in the URL or a header than user/pass for basic auth. For ease of use, tokens could default to full admin access but be restricted from there if desired.

PR #1714, opened 2020-10-11
Update ActionService and ThingActions classes in Xtext cache

### LINKED ISSUE #1265: [thing actions] Actions fail to execute after a binding restart
Reference my comment [here](https://github.com/openhab/openhab2-addons/pull/6494#issuecomment-561128456).

Most (if not all) thing action implementations contain code that looks like this.
```
    public static void restart(@Nullable ThingActions actions) {
        if (actions instanceof DoorbirdActions) {
            ((DoorbirdActions) actions).restart();
        } else {
            throw new IllegalArgumentException("Instance is not a DoorbirdActionsService class");
        }
    }
```

Normally, the `instanceof` test succeeds. However, after installing a new version of the binding, this test fails, which causes the code to throw `IllegalArgumentException` when invoking the action from a rule. This condition can only be cleared by restarting openHAB. There are several ways to reproduce this, one of which is to drop a new version of a binding into `addons` that overwrites an older version that's already there.

After installing the new version of the binding, it appears that the classloader instances are different. I suspect this is what's causing the `instanceof` test to fail.

Before installing new binding version:
```
2019-12-09 09:23:02.018 [INFO ] [ding.doorbird.action.DoorbirdActions] - DoorbirdActions classloader org.eclipse.osgi.internal.loader.EquinoxClassLoader@50578098[org.openhab.binding.doorbird:2.5.0.201912091420(id=245)]
2019-12-09 09:23:02.019 [INFO ] [ding.doorbird.action.DoorbirdActions] - ThingActions actions classloader org.eclipse.osgi.internal.loader.EquinoxClassLoader@50578098[org.openhab.binding.doorbird:2.5.0.201912091420(id=245)]
```
After installing new binding version:
```
2019-12-09 09:24:39.813 [INFO ] [ding.doorbird.action.DoorbirdActions] - DoorbirdActions classloader org.eclipse.osgi.internal.loader.EquinoxClassLoader@50578098[org.openhab.binding.doorbird:2.5.0.201912091423(id=245)]
2019-12-09 09:24:39.814 [INFO ] [ding.doorbird.action.DoorbirdActions] - ThingActions actions classloader org.eclipse.osgi.internal.loader.EquinoxClassLoader@1c14e241[org.openhab.binding.doorbird:2.5.0.201912091423(id=245)]
```

### LINKED ISSUE #1694: sendNotification fails to resolve on initial start-up
On openhab 2.5.9 and prior 2.5.Y releases, sendNotification rules fail to resolve on initial start-up of openhab2 services (ie. when the system boots up and automatically starts services at boot up, or if you perform an upgrade from 2.5.8 -> 2.5.9 and the services stop and start the new version).

Error message in the openhab.log:
2020-10-05 08:42:22.840 [ERROR] [ntime.internal.engine.RuleEngineImpl] - Rule 'Send txt message when door is open for longer than 5 minutes': The name 'sendNotification' cannot be resolved to an item or type; line 10, column 8, length 155

However, if you perform a systemctl stop openhab2 and systemctl start openhab2 (without performing a full reboot), the sendNotification rules do resolve and begin to work correctly. The config of the rules do not need to be changed for them to function, just a manual stopping and starting of the services.

wborn:
Xtext uses a cache for looking up classes when rules are run.
It also adds a null class value to this cache when a class is not found.

Once a value has entered the cache it will not be updated.
This causes the cache to return the wrong class (or the null value) when
calling static methods on ActionService and ThingActions classes that
were added/updated.

With the changes in this PR Xtext will be configured to use a custom cache
that updates the ActionService and ThingActions class references.

The PR also has a fix for the AnnotatedThingActionModuleTypeProvider not
properly sending ModuleType removed events when all ThingActions
registrations have been removed.

Fixes #1265
Fixes #1694

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/thingsupport/AnnotatedThingActionModuleTypeProvider.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/thingsupport/AnnotatedThingActionModuleTypeProvider.java
@@ -184,11 +184,10 @@ public class AnnotatedThingActionModuleTypeProvider extends BaseModuleHandlerFac
             for (ModuleInformation mi : moduleInformations) {
                 mi.setThingUID(thingUID);
 
-                ModuleType oldType = null;
                 Set<ModuleInformation> availableModuleConfigs = moduleInformation.get(mi.getUID());
                 if (availableModuleConfigs != null) {
+                    ModuleType oldType = helper.buildModuleType(mi.getUID(), moduleInformation);
                     if (availableModuleConfigs.size() > 1) {
-                        oldType = helper.buildModuleType(mi.getUID(), moduleInformation);
                         availableModuleConfigs.remove(mi);
                     } else {
                         moduleInformation.remove(mi.getUID());
@@ -196,12 +195,12 @@ public class AnnotatedThingActionModuleTypeProvider extends BaseModuleHandlerFac
 
                     ModuleType mt = helper.buildModuleType(mi.getUID(), moduleInformation);
                     // localize moduletype -> remove from map
-                    if (mt != null) {
+                    if (oldType != null) {
                         for (ProviderChangeListener<ModuleType> l : changeListeners) {
-                            if (oldType != null) {
+                            if (mt != null) {
                                 l.updated(this, oldType, mt);
                             } else {
-                                l.removed(this, mt);
+                                l.removed(this, oldType);
                             }
                         }
                     }
--- a/bundles/org.openhab.core.model.rule/src/org/openhab/core/model/rule/RulesRuntimeModule.xtend
+++ b/bundles/org.openhab.core.model.rule/src/org/openhab/core/model/rule/RulesRuntimeModule.xtend
@@ -19,6 +19,7 @@ org.openhab.core.model.rule
 import com.google.inject.Binder
 import com.google.inject.name.Names
 import org.openhab.core.model.rule.scoping.RulesImplicitlyImportedTypes
+import org.openhab.core.model.rule.scoping.RulesJavaReflectAccess
 import org.openhab.core.model.script.interpreter.ScriptInterpreter
 import org.openhab.core.model.script.jvmmodel.ScriptTypeComputer
 import org.openhab.core.model.script.scoping.ActionClassLoader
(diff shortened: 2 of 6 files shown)


kaikreuzer (approved):
Wow, looks like pure magic and I do not claim to understand all details what you did there and even less how you have figure out to do it - but the result is fantastic 🎉 .

cweitkamp:
@wborn Thanks again for this implementation. But I am afraid is is not the final solution. I still see an exception when updating a binding manually in the `addons/` folder:

```
2020-10-30 12:04:52.830 [ERROR] [internal.handler.ScriptActionHandler] - Script execution of rule with UID 'b5c106f5b5' failed: class org.openhab.binding.pushover.internal.actions.PushoverActions cannot be cast to class org.openhab.binding.pushover.internal.actions.PushoverActions (org.openhab.binding.pushover.internal.actions.PushoverActions is in unnamed module of loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @1979a84; org.openhab.binding.pushover.internal.actions.PushoverActions is in unnamed module of loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @610f03)
```

wborn:
I think there is still one other issue unrelated to this cache. I've also sometimes seen that if you would install a binding the `ThingActions` service registration is sometimes not created. As a result it would then also not be available when creating a rule using the UI.

So when that occurs the cache entry is probably also not updated. You can enable debug logging on `org.openhab.core.model.rule.scoping.RulesClassCache` to see if the ServiceRegistration for the binding ThingActions is created and causes the [cache entry to be updated](https://github.com/openhab/openhab-core/blob/25683471e813f1250ef6d4ca5f906ad58c074211/bundles/org.openhab.core.model.rule/src/org/openhab/core/model/rule/scoping/RulesClassCache.java#L71).

cweitkamp:
I get the updated message - but still the same error:

```
2020-10-30 14:20:58.280 [DEBUG] [e.model.rule.scoping.RulesClassCache] - Updated cache entry: org.openhab.binding.pushover.internal.actions.PushoverActions
...
2020-10-30 14:21:04.675 [ERROR] [internal.handler.ScriptActionHandler] - Script execution of rule with UID 'b5c106f5b5' failed: class org.openhab.binding.pushover.internal.actions.PushoverActions cannot be cast to class org.openhab.binding.pushover.internal.actions.PushoverActions (org.openhab.binding.pushover.internal.actions.PushoverActions is in unnamed module of loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @b20289; org.openhab.binding.pushover.internal.actions.PushoverActions is in unnamed module of loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @1666cf7)
```

cweitkamp:
These test were done with DSL rules. Created via UI.

PR #1716, opened 2020-10-12
Fix ScriptEngine parameter option removal

wborn:
When one of the engines is unset the ScriptModuleTypeProvider clears all parameter options instead of only those that apply to that engine.
This fixes the Nashorn engine missing from the parameter options on the first openHAB startup.

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/provider/ScriptModuleTypeProvider.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/provider/ScriptModuleTypeProvider.java
@@ -145,26 +145,34 @@ public class ScriptModuleTypeProvider implements ModuleTypeProvider {
     public void setScriptEngineFactory(ScriptEngineFactory engineFactory) {
         ScriptEngine scriptEngine = engineFactory.createScriptEngine(engineFactory.getScriptTypes().get(0));
         if (scriptEngine != null) {
-            List<String> mimeTypes = new ArrayList<>();
-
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/provider/ScriptModuleTypeProvider.java
```diff
@@ -145,26 +145,34 @@ public void removeProviderChangeListener(ProviderChangeListener<ModuleType> list
...
             logger.trace("ParameterOptions: {}", parameterOptions);
         } else {
             logger.trace("setScriptEngineFactory: engine was null");
         }
     }
 
-    public void unsetScriptEngineFactory(ScriptEngineFactory scriptEngineFactory) {
-        parameterOptions.clear();
+    public void unsetScriptEngineFactory(ScriptEngineFactory engineFactory) {
+        ScriptEngine scriptEngine = engineFactory.createScriptEngine(engineFactory.getScriptTypes().get(0));
```
kaikreuzer, line 157: Not too nice to call "createScriptEngine" here, but it seems to be the only way to get hold of the mimeTypes to remove...

kaikreuzer (approved):
Thanks!

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/provider/ScriptModuleTypeProvider.java
wborn: Another way might have been to introduce another mapping to keep track of it but this also worked when I tested it.

PR #1721, opened 2020-10-15
upgraded JmDNS to version 3.5.6

kaikreuzer:


wborn (changes requested):
I think some itest runbundles need to be reresolved.

wborn (approved):

PR #1723, opened 2020-10-16
[cache] Added 'ByteArrayFileCache'

cweitkamp:
- Added `ByteArrayFileCache`

Related to https://github.com/openhab/openhab-addons/issues/4736

---

See also https://github.com/openhab/openhab-addons/pull/8774


bundles/org.openhab.core/src/main/java/org/openhab/core/cache/ByteArrayFileCache.java
```diff
@@ -0,0 +1,305 @@
...
+    /**
+     * Creates a unique file name from the key with which the file is to be associated.
+     *
+     * @param key the key with which the file is to be associated
+     * @return unique file name for the file associated with the given key
+     */
+    String getUniqueFileName(String key) {
+        try {
+            final MessageDigest md = MessageDigest.getInstance(MD5_ALGORITHM);
+            return String.format("%032x", new BigInteger(1, md.digest(key.getBytes(StandardCharsets.UTF_8))));
```
cpmeister, line 298: I think the MD5 algorithm is a bit overkill here. Why not just use `key.hashCode()` instead? It is a lot faster and 4 billion different values should be unique enough right?

cpmeister (changes requested):
Why are you adding an actual file for testing? Just generate a temporary file during testing instead.

cweitkamp:
Thanks for your review. I incorporated your suggestions.

cweitkamp:
I have forgotten to remove the actual file ... Will do it later.

bundles/org.openhab.core/src/main/java/org/openhab/core/cache/ByteArrayFileCache.java
```diff
@@ -0,0 +1,304 @@
...
+
+    private final Logger logger = LoggerFactory.getLogger(ByteArrayFileCache.class);
+
+    static final String CACHE_FOLDER_NAME = "cache";
+    private static final char EXTENSION_SEPARATOR = '.';
+
+    private final File cacheFolder;
+
+    static final long ONE_DAY_IN_MILLIS = TimeUnit.DAYS.toMillis(1);
+    private long expiry;
```
jochen314, line 48: Could this be `final` if we rearrange the contructors and do all the work in the one with String and Duration?

bundles/org.openhab.core/src/main/java/org/openhab/core/cache/ByteArrayFileCache.java
```diff
@@ -0,0 +1,304 @@
...
+     * @param servicePID PID of the service
+     * @param Duration the days for how long the files stay in the cache valid. Must be positive. 0 to
+     *            disable this functionality.
+     */
+    public ByteArrayFileCache(String servicePID, Duration expiry) {
+        this(servicePID);
+        if (expiry.isNegative()) {
+            throw new IllegalArgumentException("Cache expiration time must be greater than or equal to 0");
+        }
+        this.expiry = expiry.toDays();
```
jochen314, line 94: If we do a `expiry.toMillis()`, then we do not need to do the multiplication each time we check the expiry of a file.

And `duration` could be anything, not just days.

bundles/org.openhab.core/src/main/java/org/openhab/core/cache/ByteArrayFileCache.java
```diff
@@ -0,0 +1,304 @@
...
+     */
+    public void clearExpired() {
+        // exit if expiry is set to 0 (disabled)
+        if (expiry <= 0) {
+            return;
+        }
+        File[] filesInCache = cacheFolder.listFiles();
+        if (filesInCache != null && filesInCache.length > 0) {
+            logger.debug("Deleting expired files from cache");
+            Arrays.stream(filesInCache).filter(file -> isExpired(file)).forEach(File::delete);
```
jochen314, line 208: ...`filter(this::isExpired).`....

bundles/org.openhab.core/src/test/java/org/openhab/core/cache/ByteArrayFileCacheTest.java
```diff
@@ -0,0 +1,203 @@
...
+        assertThat(subject.getFileExtension("/var/log/openhab2/"), is(nullValue()));
+        assertThat(subject.getFileExtension("/var/log/foo.bar/"), is(nullValue()));
+        assertThat(subject.getFileExtension("doorbell.mp3"), is(equalTo("mp3")));
+        assertThat(subject.getFileExtension("/tmp/doorbell.mp3"), is(equalTo("mp3")));
+        assertThat(subject.getFileExtension(MP3_FILE_NAME), is(equalTo("mp3")));
+        assertThat(subject.getFileExtension(TXT_FILE_NAME), is(equalTo("txt")));
+        assertThat(subject.getFileExtension("/var/log/openhab2/.."), is(""));
+        assertThat(subject.getFileExtension(".hidden"), is(equalTo("hidden")));
+        assertThat(subject.getFileExtension("C:\\Program Files (x86)\\java\\bin\\javaw.exe"), is(equalTo("exe")));
+        assertThat(subject.getFileExtension("https://www.youtube.com/watch?v=qYrpPrLY868"), is(nullValue()));
```
jochen314, line 94: Is this testing, what is implemented? None of the tests go into the branch to remove the `?` at the end of the extension.
What about:

- https://www.youtube.com/watch?v=Test.With.Dots
- https://host/test.xlsx?cache=false

bundles/org.openhab.core/src/test/java/org/openhab/core/cache/ByteArrayFileCacheTest.java
```diff
@@ -0,0 +1,203 @@
...
+    @Test
+    public void testPut() throws IOException {
+        byte[] buffer = readTempFile();
+        subject.put(TXT_FILE_NAME, buffer);
+
+        assertThat(subject.get(TXT_FILE_NAME), is(equalTo(buffer)));
+    }
+
+    @Test
+    public void testPutIfAbsent() throws IOException {
```
jochen314, line 122: Should there be a test, that `putIfAbsent ` does not overwrite an existing file?
So

- do a put with contentA
- do a putIfAbsent with contentB
- do a get and assert contentA

bundles/org.openhab.core/src/main/java/org/openhab/core/cache/ByteArrayFileCache.java
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    /**
+     * Adds a file to the cache and returns the content of the file.
+     *
+     * @param key the key with which the file is to be associated
+     * @param content the content for the file to be associated with the specified key
+     * @return the content of the file associated with the given key
+     */
+    public byte[] putIfAbsentAndGet(String key, byte[] content) {
```
jochen314, line 132: This is returning the wrong value.
If `putIfAbsent` does not change the value, then this function should return the old value, not the parameter.

bundles/org.openhab.core/src/main/java/org/openhab/core/cache/ByteArrayFileCache.java
```diff
@@ -0,0 +1,304 @@
...
+        writeFile(getUniqueFile(key), content);
+    }
+
+    /**
+     * Adds a file to the cache.
+     *
+     * @param key the key with which the file is to be associated
+     * @param content the content for the file to be associated with the specified key
+     */
+    public void putIfAbsent(String key, byte[] content) {
```
jochen314, line 114: Maybe, this function should return a boolean:
- return true, iff the file was changed
That would make implementing `putIfAbsentAndGet` easier.

cweitkamp: The idea behind of these method follows the `ExpiringCacheMap` implementation. That is why we have a `void putIfAbsent()` and a `byte[] putIfAbsentAndGet(...)` method.

bundles/org.openhab.core/src/test/java/org/openhab/core/cache/ByteArrayFileCacheTest.java
cweitkamp: :+1: Good point.

First one fails:
```
assertThat(subject.getFileExtension("https://www.youtube.com/watch?v=Test.With.Dots"), is(nullValue()));
```
To got it sorted I think I have to strip the `?` and the content after it from the file name first.

Second one is working:
```
assertThat(subject.getFileExtension("https://host/test.xlsx?cache=false"), is("xlsx"));
```

I can remember a situation where someone who uses Kodi (player) combined with Plex (media server) in a local network sent me an error in which an url like the following got interpreted wrong because of the local ip address:

- http://127.0.0.1:8080/image/image%3A%2F%2Fhttp%253a%252f%252f127.0.0.1%253a32400%252fphoto%252f%253a%252ftranscode%253fwidth%253d1920%2526height%253d1920%2526minSize%253d1%2526upscale%253d0%2526url%253d%252flibrary%252fmetadata%252f1896%252fthumb%252f1569782004%2526X-Plex-Token%253dXScJLJbUdcybNXFyHLuv

I never found a solution for it. The extension will result in `1%253a32400%252fphoto%252f%253a%252ftranscode%253fwidth%253d1920%2526height%253d1920%2526minSize%253d1%2526upscale%253d0%2526url%253d%252flibrary%252fmetadata%252f1896%252fthumb%252f1569782004%2526X-Plex-Token%253dXScJLJbUdcybNXFyHLuv`. But I think it is a very rare situation.

jochen314: I saw a problem with the first one.
But we should add the tests to be sure, that it keeps working even after firther development :-)

bundles/org.openhab.core/src/main/java/org/openhab/core/cache/ByteArrayFileCache.java
jochen314: But then the signature should be:
```
byte[] putIfAbsentAndGet(String key, Supplier<byte[]> action)
```
and the action should only be called, if the value is absent.

bundles/org.openhab.core/src/test/java/org/openhab/core/cache/ByteArrayFileCacheTest.java
cweitkamp: Already done in https://github.com/openhab/openhab-core/pull/1723/commits/9769174086818da34834ad4d8a575d2a0622f689.

bundles/org.openhab.core/src/main/java/org/openhab/core/cache/ByteArrayFileCache.java
jochen314:
```
	public void putIfAbsent(String key, byte[] content) {
		putIfAbsentAndGet(key, () -> content);
	}
	public byte[] putIfAbsentAndGet(String key, Supplier<byte[]> action) {
		File fileInCache = getUniqueFile(key);
		if (fileInCache.exists()) {
			logger.debug("File '{}' present in cache", fileInCache.getName());
			// update time of last use
			fileInCache.setLastModified(System.currentTimeMillis());
		} else {
			writeFile(fileInCache, action.get());
		}
	}
````

Comment (jochen314) [reply bundles/org.openhab.core/src/main/java/org/openhab/core/cache/ByteArrayFileCache.java:115]: sorry, getting thing mixed up ....

Comment (cweitkamp) [reply bundles/org.openhab.core/src/main/java/org/openhab/core/cache/ByteArrayFileCache.java:112]: Alright. Then we are done here?

**Review by jochen314 — APPROVED** [review]
LGTM

File: bundles/org.openhab.core/src/main/java/org/openhab/core/cache/ByteArrayFileCache.java [hunk 9]
```diff
@@ -0,0 +1,304 @@
...
+
+    /**
+     * Creates a new {@link ByteArrayFileCache} instance for a service. Creates a <code>cache</code> folder under
+     * <code>$OPENHAB_USERDATA/cache/$servicePID/</code>.
+     *
+     * @param servicePID PID of the service
+     * @param expiry the duration for how long the files stay valid in the cache. Must be positive. 0 to
+     *            disable this functionality.
+     */
+    public ByteArrayFileCache(String servicePID, long expiry) {
```
wborn, line 67: I would remove this method and only support a method taking a `Duration` parameter (i.e. the method below). The issue with long/int values is always that you do not know the unit so you have to document it in either the variable name or JavaDoc. If you don't do any (like this method) you have to inspect the code. With the `Duration` type you never have to worry about this.

bundles/org.openhab.core/src/main/java/org/openhab/core/cache/ByteArrayFileCache.java
```diff
@@ -0,0 +1,304 @@
...
+public class ByteArrayFileCache {
+
+    private final Logger logger = LoggerFactory.getLogger(ByteArrayFileCache.class);
+
+    static final String CACHE_FOLDER_NAME = "cache";
+    private static final char EXTENSION_SEPARATOR = '.';
+
+    private final File cacheFolder;
+
+    private final long expiry;
```
wborn, line 46: Is there a good reason for not making this a `Duration`?

```suggestion
    private final Duration expiry;
```

bundles/org.openhab.core/src/main/java/org/openhab/core/cache/ByteArrayFileCache.java
```diff
@@ -0,0 +1,304 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+import org.openhab.core.OpenHAB;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This is a simple file based cache implementation.
```
wborn, line 32: The class doesn't look thread safe. So you may want to mention this here.

bundles/org.openhab.core/src/main/java/org/openhab/core/cache/ByteArrayFileCache.java
```diff
@@ -0,0 +1,304 @@
...
+     *
+     * @param fileInCache the {@link File}
+     * @return <code>true</code> if the file is expired, <code>false</code> otherwise
+     */
+    private boolean isExpired(File fileInCache) {
+        // exit if expiry is set to 0 (disabled)
+        if (expiry <= 0) {
+            return false;
+        }
+        return expiry < System.currentTimeMillis() - fileInCache.lastModified();
```
wborn, line 222:
```suggestion
        return 0 < expiry && expiry < System.currentTimeMillis() - fileInCache.lastModified();
```

bundles/org.openhab.core/src/test/java/org/openhab/core/cache/ByteArrayFileCacheTest.java
```diff
@@ -0,0 +1,238 @@
...
+    private static final String SERVICE_PID = "org.openhab.core.test";
+
+    private static final File USERDATA_FOLDER = new File(OpenHAB.getUserDataFolder());
+    private static final File CACHE_FOLDER = new File(USERDATA_FOLDER, ByteArrayFileCache.CACHE_FOLDER_NAME);
+    private static final File SERVICE_CACHE_FOLDER = new File(CACHE_FOLDER, SERVICE_PID);
+
+    private static final String MP3_FILE_NAME = SERVICE_CACHE_FOLDER.getAbsolutePath() + "doorbell.mp3";
+    private static final String TXT_FILE_NAME = SERVICE_CACHE_FOLDER.getAbsolutePath() + "doorbell.txt";
+
+    private static @Nullable File txt_file;
```
wborn, line 53:
```suggestion
    private static @Nullable File txtFile;
```

wborn (commented):
It's nice this we can remove the duplicated code from the add-ons with this. :+1:

wborn (approved):
Thanks!

PR #1724, opened 2020-10-17
transport/mqtt: Introduce setUnsubscribeOnStop() function

Sonic-Amiga:
Some MQTT servers can be quirky, they do not handle Unsubscribe request properly.
In this case we have to omit sending it. Introduce a boolean flag, telling
whether the request should be sent or not, and add a public function to set it.

iRobot built-in MQTT server is known to suffer from this problem.

Signed-off-by: Pavel Fedin <pavel_fedin@mail.ru>


kaikreuzer (approved):
lgtm, thanks!

Sonic-Amiga:
@kaikreuzer Hello! How long should i wait before the change appears in Maven downloadable bundles? Currently the timestamp says "23 hours ago", Maven did pull packages when i attempted to rebuild the iRobot binding, but the new function is still not available.

kaikreuzer:
Maven checks once a day (unless you use the `-U´ option). The change was included in the build yesterday at 3am (https://ci.openhab.org/job/openHAB-Core/567/), so your Maven might only pick it up today.

Sonic-Amiga:
Thank you, it's all fine now, got the update yesterday's evening. Will test my updated binding on weekend

PR #1726, opened 2020-10-17
[profiles] Added 'Hysteresis' profile

cweitkamp:
- Added 'Battery Low' profile -> Changed to 'Hysteresis' Profile

~~This profile is useful to implement a user-defined threshold for battery low warning. Sometimes a vendors create a battery low  warning themselves based on a fixed value - which may not be useful from users pov - beside delivering a battery level for devices. In other cases the bindings developers add a estimation in their code to deliver battery low warning (like e.g. Hue or TRADFRI).~~

How does it work:

The profile configuration needs a mandatory lower bound (`lower`) config parameter and an optional upper bound (`upper`) config parameter - will be set to lower bound, if missing. The resulting Switch will change to `ON` if the value drops below the lower bound. It will change to `OFF` if the value exceeds upper bound.

Configuration:
```java
Switch hysteresisResult0 {
    channel="..." [profile="system:hysteresis", lower=10, inverted=true]
}
Switch hysteresisResult1 {
    channel="..." [profile="system:hysteresis", lower=10, upper=40]
}
Switch hysteresisResult2 {
    channel="..." [profile="system:hysteresis", lower="10 cm", upper="2 m"]
}
```


kaikreuzer (commented):
Thanks for this PR, @cweitkamp.
Code looks good and seems to make sense at a first glance.

But on a second thought, I'd like to challenge it:
Our other profiles are "technically" changing commands/states with certain logic to other commands/states. The new "Battery Low" profile instead seems to hold semantics (it seems to apply only to number channels for battery levels).

We specifically have a system "battery low" channel type - any channel that will use this profile here, won't be recognized to have the battery low semantic, since it is not the system channel type (and we don't have a feature to claim it were). This seems to make it imperfect, and it would especially annoy if we one day add special support for that system channel type in the UI (to e.g. directly show battery level next to a Thing view - @ghys ever thought about such a feature? Same would be neat for signal_strength).

With this in mind, why should it at all be restricted to a "battery" semantic? It essentially is a "threshold" profile that could work for ANY number channel. In that way, it would be very similar to the OFFSET profile. 

Thinking even further, it could even become a "hysteresis" profile, which I'd think would be really cool. It could be used for triggering all kinds of alarms and notifications and would avoid flip-flopping.

Wdyt?

ghys:
> @ghys ever thought about such a feature?

Yes :)

In general I think it would be a good thing to have the semantic tag ontology and the (system) channel types match - and try to stick to those as much as possible instead of redefining channel types. For instance I recently noticed there's a lot of channel types related to media players (title, artist, album, mute audio etc.), that's great but 1) some bindings still define their own when perhaps the system one would have been fine, and 2) you don't easily have the information by looking at the Item alone that it's linked to a channel of that type.

Also most if not all of these channel types don't have semantic tags attached so that the item could inherit these tags when it's created along with a new link, and as I said above you often don't have an equivalent in the ontology - there's no "album art" or "thermostat mode" tag.

In the media player example, if we had all this information available from the channel type it's linked to (which item controls the volume, which item is the artist name, etc...) then the UI could somehow detect "media-player-like" structures of items and start offering a tailored control (player controls + mute/volume controls + title/artist/album + cover art, and so on).

cweitkamp:
@kaikreuzer That is all true and changing it to a "Hyteresis" profile sounds indeed very nice. I will rework my PR. A "Battery Low" can be a specific use case for it and we later can give users a hint by an example in the docs.

kaikreuzer:
> Yes :)

I was actually referring to two very specific features:
Using the low-battery/battery-level and signal-strength channels as something directly related to the Thing and showing it in the Thing (list/detail) view - similar like a smartphone shows battery level and signal strength in the top bar. I.e. considering it as administrative information more than user information, so that admins can easily browse through their things and see if something isn't ok (battery to be replaced or signal issues), just like he is seeing there if Things go OFFLINE (battery+signal are some kind of "early stage" of being OFFLINE).

But wrt your points: Yes, for media and other channels, it will make sense to have proper semantic tags as well. The reason for having these system channel types is to make sure that every binding that implements them does it in the same way (i.e. same item type, same states and commands, same ranges, same labels, etc.). This is hence a much stronger semantic than the tags alone, which you can potentially put on any kind of item.
Looking at the tags and identifying if the UI can render control groups for them, is fine as long as the UI can flexibly adapt to all the situations it could come across. If it has a way to identify the underlying channel type (and see that they are system channel types), it would probably help to have even more specialised controls being rendered. But you are probably the much better expert to come up with good solutions for that :-) .

@cweitkamp Sorry for going off-topic here...

cweitkamp:
> @cweitkamp Sorry for going off-topic here...

No problem, I hope you will find your discussion again if you are looking for it in the future ... :wink:.

In the meantime I tried to implement the new "Hyteresis" profile. You can have a look if you like.

kaikreuzer:
Awesome, you are incredibly quick, @cweitkamp! Code looks excellent, as always.

Just one further remark, if you allow:

> The resulting Switch will change to ON if the value drops below the lower bound. It will change to OFF if the value exceeds upper bound.

This seems logical coming with the low-bat use case in mind. In general, I'd feel that a mapping to ON above the upper bound and to OFF below the lower bound might be more natural, though (think e.g. of alarms for temperature, Co2 concentration, etc.). There are clearly use cases for both variants. So how about making low=OFF, high=ON the default and add another config param "negate" or so to it?

cweitkamp:
Sounds reasonable too. I will change the ON/OFF calculation and provide a configuration for inverting it.

Some questions from my side:

- Do we want to allow `QuantityType` values for lower and upper bound - similar to Offset Profile or are we happy with `BigDecimals`
- Should we address other target Items like `Contact` too?

Beside this new "Hysteresis" Profile I cold imagine some use cases where users like to trigger an alert if values are outside a specified range (lower or higher). E.g. I have a BBQ thermometer and want to receive an alert if the temperature in my oven drops below 110°C or is higher than 180°C. Or if one would want to control some smart gardening for rare and delicate plants which need a specific humidity. I would like to add a "Boundary" Profile for it too.

kaikreuzer:
> Do we want to allow QuantityType values for lower and upper bound - similar to Offset Profile

Might be a nice addition. If it is as easy to do as in the Offset profile, why not.

> Should we address other target Items like Contact too?

Not sure if that brings much value - it might rather cause confusion (what is mapped to OPEN vs. CLOSED?). I guess with ON/OFF, we are fine here.

> I would like to add a "Boundary" Profile for it too.

Yeah, that might be another neat feature indeed. Other maybe suitable names to consider might be "Corridor" or "Variance".

bundles/org.openhab.core.thing/src/main/resources/OH-INF/config/hyteresisProfile.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<config-description:config-descriptions
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:config-description="https://openhab.org/schemas/config-description/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/config-description/v1.0.0 https://openhab.org/schemas/config-description-1.0.0.xsd">
+
+	<config-description uri="profile:system:hysteresis">
+		<parameter name="lower" type="decimal" required="true">
+			<label>Lower Bound</label>
+			<description>Maps to ON if value is below lower bound.</description>
```
cweitkamp, line 10:
```suggestion
			<description>Maps to OFF if value is below lower bound.</description>
```

bundles/org.openhab.core.thing/src/main/resources/OH-INF/config/hyteresisProfile.xml
```diff
@@ -0,0 +1,17 @@
...
+		</parameter>
+		<parameter name="upper" type="decimal">
+			<label>Upper Bound</label>
+			<description>Maps to OFF if value is above upper bound.</description>
```
cweitkamp, line 14:
```suggestion
			<description>Maps to ON if value is above upper bound.</description>
```

kaikreuzer (approved):
Thanks, lgtm.

May I ask you to look into identifying a good place in the documentation to add some user docs about this profile? I think we are currently lacking this completely, but I think we have reached a point where those profiles are not self-explanatory anymore.
For the hysteresis, it probably makes sense to also clearly define whether the bounds are "value less than boundary" or "value less or equal boundary". Also it should be explained to the user that it is not just a mapping (as the parameter description suggests), but that it is about toggling a state within this profile (otherwise it would be undefined what value is delivered for a value between lower and upper bound).

cweitkamp:
We have a [Profiles section](https://www.openhab.org/docs/configuration/items.html#profiles) in Configuration Guide -> Items. I will add documentation and examples over there unless you want to extract this whole part into a new subcategory.

kaikreuzer:
Ah, great, I had missed that. Yeah, would be great if you could add the new profile there. Extracting profiles to a separate page might make sense, but I think we have more urgent tasks on updating the documentation for 3.0, so restructuring can imho wait until later.

PR #1727, opened 2020-10-18
Fix issues with configuration PIDs

wborn:
* Fixes warnings about "using different service PIDs"
* Fixes missing default values

---

These warnings would show up:

```
[WARN ] [.service.ConfigurableServiceResource] - The component "org.openhab.core.net.NetUtil" is using different service PIDs ([org.openhab.core.network, org.openhab.network]). Different service PIDs are not supported, the first one (org.openhab.core.network) is used.
[WARN ] [.service.ConfigurableServiceResource] - The component "org.openhab.core.ui.internal.chart.ChartServlet" is using different service PIDs ([org.openhab.core.chart, org.openhab.chart]). Different service PIDs are not supported, the first one (org.openhab.core.chart) is used.
```


kaikreuzer (approved):
Interesting that we had missed those previously. Thanks for fixing!

PR #1729, opened 2020-10-19
Fix build by wrapping lines

wborn:
Caused by #1713

### DIFF
--- a/bundles/org.openhab.core.io.rest.auth/src/main/resources/OH-INF/config/config.xml
+++ b/bundles/org.openhab.core.io.rest.auth/src/main/resources/OH-INF/config/config.xml
@@ -16,8 +16,8 @@
 			<label>Implicit user role for unauthenticated requests</label>
 			<default>true</default>
 			<description>By default, operations requiring the "user" role are available when unauthenticated. Disabling this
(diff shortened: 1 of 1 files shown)


kaikreuzer (approved):
Thanks, you were one minute faster than me ;-)

wborn:
Because the Spotless XML wrapping is buggy, it will likely have a strange opinion about last minute [changes](https://github.com/openhab/openhab-core/pull/1713/commits/0c9d04494645b41b45b09048d439c35c182f7fae). :wink: 
I still need to check if there is already an issue for it in the Spotless issue tracker (and create one if missing). This buggy behavior can do serious damage (https://github.com/openhab/openhab-addons/pull/8413/) and is a source of unnecessary work, waiting and discussions. Maybe @nedtwigg knows if there is an existing issue for this?

nedtwigg:
I maintain the Spotless core, but I haven't used the XML formatter myself.  I see that lines are being rewrapped, but that seems like expected behavior.  Is the bug that you don't want Spotless to care about line wrapping?  Or is Spotless being inconsistent about it, where it passes in one place but fails in another?

wborn:
I've created https://github.com/diffplug/spotless/issues/723 for this @nedtwigg. :-)

PR #1735, opened 2020-10-19
[REST Auth] API tokens & openhab:users console command

ghys:
This adds API tokens as a new credential type. Their format is:
`oh.<name>.<random chars>`

They are stored hashed in the user's profile, and can be listed, added
or removed managed with the new `openhab:users` console command.

Currently the scopes are still not checked, but ultimately they could
be, for instance a scope of e.g. `user admin.items` would mean that the
API token can be used to perform user operations like retrieving info
or sending a command, _and_ managing the items, but nothing else -
even if the user has more permissions because of their role (which
will of course still be checked).

Tokens are normally passed in the Authorization header with the Bearer
scheme, or the X-OPENHAB-TOKEN header, like access tokens.
As a special exception, API tokens can also be used with the Basic
authorization scheme, **even if the allowBasicAuth** option is not
enabled in the "API Security" service, because there's no additional
security risk in allowing that. In that case, the token should be
passed as the username and the password MUST be empty.

In short, this means that all these curl commands will work:
- `curl -H 'Authorization: Bearer <token>' http://localhost:8080/rest/inbox`
- `curl -H 'X-OPENHAB-TOKEN: <token>' http://localhost:8080/rest/inbox`
- `curl -u '<token>' http://localhost:8080/rest/inbox` (will prompt for a password, just hit Return);
- `curl -u '<token>:' http://localhost:8080/rest/inbox` (will not prompt for a password)
- `curl http://<token>@localhost:8080/rest/inbox`

2 REST API operations were adding to the AuthResource, to allow
authenticated users to list their tokens or remove (revoke) one.

Self-service for creating a token or changing the password is more
sensitive so these are handled with a servlet and pages devoid
of any JavaScript instead of REST API calls.

Closes https://github.com/openhab/openhab-webui/issues/332


ghys:
Examples:

```
openhab> openhab:users
Usage: openhab:users list - lists all users
Usage: openhab:users add <userId> <password> <role> - adds a new user with the specified role
Usage: openhab:users remove <userId> - removes the given user
Usage: openhab:users changePassword <userId> <newPassword> - changes the password of a user
Usage: openhab:users listApiTokens - lists the API keys for all users
Usage: openhab:users addApiToken <userId> <tokenName> <scope> - adds a new API token on behalf of the specified user for the specified scope
Usage: openhab:users rmApiToken <userId> <tokenName> - removes (revokes) the specified API token
Usage: openhab:users clearSessions <userId> - clear the refresh tokens associated with the user (will sign the user out of all sessions)
openhab> openhab:users add ysc P@ssword1 administrator
ysc (administrator)
User created.
openhab> openhab:users list
ysc (administrator)
openhab> openhab:users changePassword ysc P@ssword2
Password changed.
openhab> openhab:users addApiToken ysc service1 admin
oh.service1.IR71uwmszu0KRBEHo6GUC1BIJh7dHhv9yobyPCqwPgiOHFCT7sRjEcE3R6XfT9y6LsnR9ntk9KDGY5kchy1uBw
openhab> openhab:users addApiToken ysc service2 admin
oh.service2.GXOeP2weDRpHLXWirACcZfaz9FHh7PJhodMwHnt8NHCJWYSaSqNidvCdztHguCSQtBG885DgrWV8F3aaQQwPg
openhab> openhab:users addApiToken ysc service3 admin
oh.service3.XCIhrBGOQiNBrVdiFt2VopNgoVmQk75q9jZeOjJl9PEJpiK9V8pT9XvyKfSpoGtfmUCja9FwilHdQUL5eWDMPw
openhab> openhab:users listApiTokens
user=ysc (administrator), name=service1, createdTime=Mon Oct 19 20:17:40 CEST 2020, scope=admin
user=ysc (administrator), name=service2, createdTime=Mon Oct 19 20:17:46 CEST 2020, scope=admin
user=ysc (administrator), name=service3, createdTime=Mon Oct 19 20:17:50 CEST 2020, scope=admin
openhab> openhab:users rmApiToken ysc service1
API token revoked.
openhab> openhab:users rmApiToken ysc service2
API token revoked.
openhab> openhab:users rmApiToken ysc service3
API token revoked.
openhab> openhab:users listApiTokens
openhab> 
```
Note that the tokens are only printed on the console after they've been created, and there should be copy-pasted immediately, because they won't be retrievable afterwards.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -93,6 +96,32 @@ protected void modified(@Nullable Map<String, @Nullable Object> properties) {
         }
     }
 
+    private SecurityContext authenticateBearerToken(String token) throws AuthenticationException {
+        if (token.startsWith(API_TOKEN_PREFIX)) {
+            UserApiTokenCredentials credentials = new UserApiTokenCredentials(token);
+            Authentication auth = userRegistry.authenticate(credentials);
+            User user = userRegistry.get(auth.getUsername());
+            if (user == null) {
+                throw new org.openhab.core.auth.AuthenticationException("User not found in registry");
```
wborn, line 105:
```suggestion
                throw new AuthenticationException("User not found in registry");
```

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -93,6 +96,32 @@ protected void modified(@Nullable Map<String, @Nullable Object> properties) {
...
+        }
+    }
+
+    private SecurityContext authenticateUsernamePassword(String username, String password)
+            throws AuthenticationException {
+        UsernamePasswordCredentials credentials = new UsernamePasswordCredentials(username, password);
+        Authentication auth = userRegistry.authenticate(credentials);
+        User user = userRegistry.get(auth.getUsername());
+        if (user == null) {
+            throw new org.openhab.core.auth.AuthenticationException("User not found in registry");
```
wborn, line 120:
```suggestion
            throw new AuthenticationException("User not found in registry");
```

wborn (commented):
It's nice to have some commands for this. :+1: 
I quickly scrolled through the code and saw a few minor improvements.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
ghys: Thanks, I rebased and addressed it in the process.

ghys:
FTR, I'm not too happy with passwords being provided in clear on the command line (they are stored in the history).
In Karaf can set censor/mask properties to command arguments to hide them from view (https://github.com/apache/karaf/commit/5228a23fc4113922cc9936c8d921a97cc2ea434d), but the openHAB console wrapper doesn't seem to allow that.

ghys:
Finally went ahead and implemented the self-service pages for changing passwords and creating API tokens, at `/changePassword` and `/createApiToken` respectfully.

localhost_8080_changePassword(iPad Mini)
localhost_8080_changePassword(iPad Mini) (1)
localhost_8080_createApiToken(iPad Mini)
localhost_8080_createApiToken(iPad Mini) (1)

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/TokenResource.java
```diff
@@ -154,6 +158,49 @@ public Response getSessions(@Context SecurityContext securityContext) {
...
     }
 
+    @GET
+    @Path("/apitokens")
+    @Operation(summary = "List the API tokens associated to the authenticated user.", responses = {
+            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = UserApiTokenDTO.class))) })
+    @Produces({ MediaType.APPLICATION_JSON })
+    public Response getApiTokens(@Context SecurityContext securityContext) {
+        if (securityContext.getUserPrincipal() == null) {
+            return JSONResponse.createErrorResponse(Status.UNAUTHORIZED, "User is not authenticated");
```
wborn, line 168: This response seems to be missing from the OpenAPI responses of the `@Operation`.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/TokenResource.java
```diff
@@ -154,6 +158,49 @@ public Response getSessions(@Context SecurityContext securityContext) {
...
+        }
+
+        ManagedUser user = (ManagedUser) userRegistry.get(securityContext.getUserPrincipal().getName());
+        if (user == null) {
+            return JSONResponse.createErrorResponse(Status.NOT_FOUND, "User not found");
```
wborn, line 173: This response seems to be missing from the OpenAPI responses of the `@Operation`.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/TokenResource.java
```diff
@@ -154,6 +158,49 @@ public Response getSessions(@Context SecurityContext securityContext) {
...
+        return Response.ok(new Stream2JSONInputStream(sessions)).build();
+    }
+
+    @DELETE
+    @Path("/apitokens/{name}")
+    @Operation(summary = "Revoke a specified API token associated to the authenticated user.", responses = {
+            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = UserApiTokenDTO.class))) })
+    public Response removeApiToken(@Context SecurityContext securityContext, @PathParam("name") String name) {
+        if (securityContext.getUserPrincipal() == null) {
+            return JSONResponse.createErrorResponse(Status.UNAUTHORIZED, "User is not authenticated");
```
wborn, line 186: This response seems to be missing from the OpenAPI responses of the `@Operation`.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/TokenResource.java
```diff
@@ -154,6 +158,49 @@ public Response getSessions(@Context SecurityContext securityContext) {
...
+        }
+
+        ManagedUser user = (ManagedUser) userRegistry.get(securityContext.getUserPrincipal().getName());
+        if (user == null) {
+            return JSONResponse.createErrorResponse(Status.NOT_FOUND, "User not found");
```
wborn, line 191: This response seems to be missing from the OpenAPI responses of the `@Operation`.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/TokenResource.java
```diff
@@ -154,6 +158,49 @@ public Response getSessions(@Context SecurityContext securityContext) {
...
+        }
+
+        Optional<UserApiToken> userApiToken = user.getApiTokens().stream()
+                .filter(apiToken -> apiToken.getName().equals(name)).findAny();
+        if (userApiToken.isEmpty()) {
+            return JSONResponse.createErrorResponse(Status.NOT_FOUND, "No API token found with that name");
```
wborn, line 197: This response seems to be missing from the OpenAPI responses of the `@Operation`.

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/auth/UserRegistryImpl.java
```diff
@@ -128,23 +133,132 @@ public User register(String username, String password, Set<String> roles) {
 
     @Override
     public Authentication authenticate(Credentials credentials) throws AuthenticationException {
-        UsernamePasswordCredentials usernamePasswordCreds = (UsernamePasswordCredentials) credentials;
-        User user = this.get(usernamePasswordCreds.getUsername());
-        if (user == null) {
-            throw new AuthenticationException("User not found: " + usernamePasswordCreds.getUsername());
+        if (credentials instanceof UsernamePasswordCredentials) {
+            UsernamePasswordCredentials usernamePasswordCreds = (UsernamePasswordCredentials) credentials;
+            User user = this.get(usernamePasswordCreds.getUsername());
```
wborn, line 138:
```suggestion
            User user = get(usernamePasswordCreds.getUsername());
```

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/auth/UserRegistryImpl.java
```diff
@@ -128,23 +133,132 @@ public User register(String username, String password, Set<String> roles) {
...
+        if (!(user instanceof ManagedUser)) {
+            throw new IllegalArgumentException("User is not managed: " + user.getName());
+        }
+
+        ManagedUser managedUser = (ManagedUser) user;
+        String passwordSalt = generateSalt(KEY_LENGTH / 8).get();
+        String passwordHash = hash(newPassword, passwordSalt, PASSWORD_ITERATIONS).get();
+        managedUser.setPasswordSalt(passwordSalt);
+        managedUser.setPasswordHash(passwordHash);
+        this.update(user);
```
wborn, line 192:
```suggestion
        update(user);
```

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/auth/UserRegistryImpl.java
```diff
@@ -128,23 +133,132 @@ public User register(String username, String password, Set<String> roles) {
...
+
+    @Override
+    public void addUserSession(User user, UserSession session) {
+        if (!(user instanceof ManagedUser)) {
+            throw new IllegalArgumentException("User is not managed: " + user.getName());
+        }
+
+        ManagedUser managedUser = (ManagedUser) user;
+        managedUser.getSessions().add(session);
+        this.update(user);
```
wborn, line 203:
```suggestion
        update(user);
```

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/auth/UserRegistryImpl.java
```diff
@@ -128,23 +133,132 @@ public User register(String username, String password, Set<String> roles) {
...
+
+    @Override
+    public void removeUserSession(User user, UserSession session) {
+        if (!(user instanceof ManagedUser)) {
+            throw new IllegalArgumentException("User is not managed: " + user.getName());
         }
+
+        ManagedUser managedUser = (ManagedUser) user;
+        managedUser.getSessions().add(session);
+        this.update(user);
```
wborn, line 214:
```suggestion
        update(user);
```

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/auth/UserRegistryImpl.java
```diff
@@ -128,23 +133,132 @@ public User register(String username, String password, Set<String> roles) {
...
-            throw new AuthenticationException("User is not managed: " + usernamePasswordCreds.getUsername());
+            throw new IllegalArgumentException("User is not managed: " + user.getName());
         }
 
         ManagedUser managedUser = (ManagedUser) user;
-        String hashedPassword = hashPassword(usernamePasswordCreds.getPassword(), managedUser.getPasswordSalt()).get();
-        if (!hashedPassword.equals(managedUser.getPasswordHash())) {
-            throw new AuthenticationException("Wrong password for user " + usernamePasswordCreds.getUsername());
+        managedUser.getSessions().clear();
+        this.update(user);
```
wborn, line 225:
```suggestion
        update(user);
```

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/auth/UserRegistryImpl.java
```diff
@@ -128,23 +133,132 @@ public User register(String username, String password, Set<String> roles) {
...
+        byte[] rnd = new byte[64];
+        RAND.nextBytes(rnd);
+        String token = APITOKEN_PREFIX + "." + name + "."
+                + Base64.getEncoder().encodeToString(rnd).replaceAll("(\\+|/|=)", "");
+        String tokenHash = hash(token, tokenSalt, APITOKEN_ITERATIONS).get();
+
+        UserApiToken userApiToken = new UserApiToken(name, tokenHash + ":" + tokenSalt, scope);
+
+        managedUser.getApiTokens().add(userApiToken);
+        this.update(user);
```
wborn, line 248:
```suggestion
        update(user);
```

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/auth/UserRegistryImpl.java
```diff
@@ -128,23 +133,132 @@ public User register(String username, String password, Set<String> roles) {
...
+
+    @Override
+    public void removeUserApiToken(User user, UserApiToken userApiToken) {
+        if (!(user instanceof ManagedUser)) {
+            throw new IllegalArgumentException("User is not managed: " + user.getName());
+        }
+
+        ManagedUser managedUser = (ManagedUser) user;
+        managedUser.getApiTokens().remove(userApiToken);
+        this.update(user);
```
wborn, line 261:
```suggestion
        update(user);
```

wborn (changes requested):
Thanks a lot for your awesome work!
I've added a few more (minor) comments below.

bundles/org.openhab.core.io.http.auth/src/main/java/org/openhab/core/io/http/auth/internal/CreateAPITokenPageServlet.java
```diff
@@ -0,0 +1,186 @@
...
+            } catch (Exception e) {
+                resp.setContentType("text/plain;charset=UTF-8");
+                resp.getWriter().append(e.getMessage());
+                resp.getWriter().close();
+            }
+        }
+    }
+
+    @Override
+    protected void doPost(@Nullable HttpServletRequest req, @Nullable HttpServletResponse resp)
```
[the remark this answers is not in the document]
J-N-K: This is against what was decided before. Otherwise using @NonNullByDefault({}) would also be possible for mandatory config fields or the config itself of set in the thing handler’s initialize(). The decision was to only use the @NonNullByDefault({})-annotation for injected ressources of the OSGi-framework.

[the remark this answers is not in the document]
wborn: The agreed upon [guidelines](https://www.openhab.org/docs/developer/guidelines.html#null-annotations) don't mention anything about how to deal with annotating classes that implement methods from unannotated classes.

These issues also did not occur previously because we did not annotate classes with `@NonNullByDefault({})` if they were overriding methods from unannotated classes.

If there were EEAs for this dependency, these parameters would probably be analyzed to be `@NonNull` so in that case it makes more sense to use `@NonNullByDefault({})` which would prevent these unnecessary null checks.

Another case not mentioned in the guidelines is how to deal with generics. The `@NonNullByDefault({})` defined on classes is not used by the analysis for generic types. So in that case the only way to make them non-null is by annotating the generic types with `@NonNull`, see for example the [AbstractRegistry](https://github.com/openhab/openhab-core/blob/5d75bce5539aee321f263202f071304ffe2908b0/bundles/org.openhab.core/src/main/java/org/openhab/core/common/registry/AbstractRegistry.java#L58-L59).

wborn:
It would be nice if another maintainer can also have a look since this deals with security.

bundles/org.openhab.core.io.console/src/main/java/org/openhab/core/io/console/internal/extension/UserConsoleCommandExtension.java
```diff
@@ -0,0 +1,197 @@
...
+                    break;
+                case SUBCMD_ADDAPITOKEN:
+                    if (args.length == 4) {
+                        ManagedUser user = (ManagedUser) userRegistry.get(args[1]);
+                        if (user != null) {
+                            Optional<UserApiToken> userApiToken = user.getApiTokens().stream()
+                                    .filter(t -> args[2].equals(t.getName())).findAny();
+                            if (userApiToken.isEmpty()) {
+                                String tokenString = userRegistry.addUserApiToken(user, args[2], args[3]);
+                                // inform the user that the token will not be printed again, and they should save it?
```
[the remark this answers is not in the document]
wborn: Is it even true the tokens cannot be retrieved?

I see there is a command to list them:

`openhab:users listApiTokens - lists the API keys for all users`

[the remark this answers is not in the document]
ghys: > I think it would be helpful to print it. The UI also informs users of this.

I don't know, in the end I figured it's best to leave the command ouputs as no-frills as possible, in case the output is redirected to a file/piped to another command. The admin who masters these commands is supposed to know that it's the only opportunity to save the token.

> I see there is a command to list them

You only see the metadata like below, but not the actual token (and you can't even retrieve it from the JSON DB since it's not stored in plain text):
```
openhab> openhab:users listApiTokens
user=ysc (administrator), name=service1, createdTime=Mon Oct 19 20:17:40 CEST 2020, scope=admin
user=ysc (administrator), name=service2, createdTime=Mon Oct 19 20:17:46 CEST 2020, scope=admin
user=ysc (administrator), name=service3, createdTime=Mon Oct 19 20:17:50 CEST 2020, scope=admin
```

bundles/org.openhab.core.io.console/src/main/java/org/openhab/core/io/console/internal/extension/UserConsoleCommandExtension.java
```diff
@@ -0,0 +1,197 @@
...
+
+    @Activate
+    public UserConsoleCommandExtension(final @Reference UserRegistry userRegistry) {
+        super("users", "Access the user registry.");
+        this.userRegistry = userRegistry;
+    }
+
+    @Override
+    public List<String> getUsages() {
+        return Arrays.asList(new String[] { buildCommandUsage(SUBCMD_LIST, "lists all users"),
```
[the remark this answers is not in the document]
ghys: I see there's inconsistencies in the other CommandExtensions:

https://github.com/openhab/openhab-core/blob/5d75bce5539aee321f263202f071304ffe2908b0/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceConsoleCommandExtension.java#L65-L68

https://github.com/openhab/openhab-core/blob/5d75bce5539aee321f263202f071304ffe2908b0/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/AudioConsoleCommandExtension.java#L66-L69

I happened to take the code from the AudioConsoleCommandExtension :) But I will change it to `List.of` nonetheless.

bundles/org.openhab.core.io.http.auth/src/main/java/org/openhab/core/io/http/auth/internal/AbstractAuthPageServlet.java
```diff
@@ -0,0 +1,129 @@
...
+        this.userRegistry = userRegistry;
+        this.authProvider = authProvider;
+
+        pageTemplate = "";
+        URL resource = bundleContext.getBundle().getResource("pages/authorize.html");
+        if (resource != null) {
+            try {
+                pageTemplate = new String(resource.openStream().readAllBytes(), StandardCharsets.UTF_8);
+            } catch (IOException e) {
+                throw new UncheckedIOException(e);
```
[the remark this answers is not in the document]
ghys: This is only when the page cannot be loaded from the bundle resources, I don't think it will ever happen. I didn't know how to handle this exception, any suggestion is welcome.
In general authentication failures are logged but not shown in detail to the user on the webpage (it only says "Please try again") because you don't want to divulge too much (i.e. does the account exist, and only the password is wrong, etc.).

bundles/org.openhab.core.io.console/src/main/java/org/openhab/core/io/console/internal/extension/UserConsoleCommandExtension.java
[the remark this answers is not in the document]
wborn: Great! :+1: I didn't have time to compile and execute the command myself or figure out if it was hashed or plain text when I made the comment. Otherwise the next comment would have been to hash it. :wink:

lolodomo:
Just for helping any user, would it be possible to mention in the help of the new commands what values can be used for the `<scope>` parameter ? I see in your examples that you use `admin` but are there other possible values ?

In the case of my Remote openHAB binding that uses the openHAB REST API, in case the remote OH3 server was setup to require authorization for normal REST API, I understand I will have to create a token using your new console command and then use this token in the binding by injecting it in the header of each REST API call. Is it correct ?

ghys:
The scope model is not clear yet, but I'll suggest something like this:
- Missing/empty scope, or `all` => no scope restrictions, authorize everything the user's role(s) normally allow (not sure it's a good idea)
- `user` => operations to interact with the system as a user: SSE events, view items, sitemaps, pages, persisted data; send item updates & commands; audio & voice; and their own profile (sessions & tokens);
- `admin` => operations to change the system: full access on things, inbox & discovery, items & links & metadata, UI components, rules, services, system info, addons & binding configuration...
- Eventually we'll have more granular scopes like `admin.things`, `admin.rules`...

> in case the remote OH3 server was setup to require authorization for normal REST API, I understand I will have to create a token using your new console command and then use this token in the binding by injecting it in the header of each REST API call. Is it correct ?

yes currently, the "user" operations are unsecured but in case they are (you can switch that on now), you'll have to provide a token to access even the most basic operations that expose data, if you don't need admin access to make changes to the configuration then a scope of `user` should be fine. Normally you'll be able to specify multiple scopes (space-separated).
You can generate a token with the console command or with the `/createApiToken` page, see above. There will be a list in the UI in the user's profile page for them to view their current tokens and allowing to create new ones.

lolodomo:
Ok thank you, I am waiting for the merge of this PR to test with my Remote openHAB binding.

bundles/org.openhab.core.io.http.auth/src/main/java/org/openhab/core/io/http/auth/internal/AbstractAuthPageServlet.java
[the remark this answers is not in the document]
wborn: You'll run into this whenever one day the file is deleted/renamed or moved to another directory.
So I would add a message to the exception that it cannot open "pages/authorize.html".

bundles/org.openhab.core.io.http.auth/src/main/java/org/openhab/core/io/http/auth/internal/ChangePasswordPageServlet.java
```diff
@@ -57,43 +56,41 @@ public ChangePasswordPageServlet(BundleContext bundleContext, @Reference HttpSer
...
+            resp.getWriter().append(e.getMessage());
+            resp.getWriter().close();
         }
     }
 
     @Override
-    protected void doPost(@Nullable HttpServletRequest req, @Nullable HttpServletResponse resp)
+    protected void doPost(@NonNullByDefault({}) HttpServletRequest req, @NonNullByDefault({}) HttpServletResponse resp)
             throws ServletException, IOException {
         if (req != null && resp != null) {
```
wborn, line 84: These checks can be removed now.

bundles/org.openhab.core.io.console/src/main/java/org/openhab/core/io/console/internal/extension/UserConsoleCommandExtension.java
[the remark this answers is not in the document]
wborn: > I don't know, in the end I figured it's best to leave the command ouputs as no-frills as possible, in case the output is redirected to a file/piped to another command. The admin who masters these commands is supposed to know that it's the only opportunity to save the token.

Then the question has been answered and we should remove the comment. :-)

bundles/org.openhab.core.io.http.auth/src/main/java/org/openhab/core/io/http/auth/internal/AuthorizePageServlet.java
```diff
@@ -64,210 +55,172 @@
...
-                String scope = params.get("scope")[0];
+            String baseRedirectUri = params.get("redirect_uri")[0];
+            String responseType = params.get("response_type")[0];
+            String clientId = params.get("redirect_uri")[0];
+            String scope = params.get("scope")[0];
 
-                if (!("code".equals(responseType))) {
-                    throw new AuthenticationException("unsupported_response_type");
-                }
+            if (!("code".equals(responseType))) {
```
wborn, line 140:
```suggestion
            if (!"code".equals(responseType)) {
```

bundles/org.openhab.core.io.http.auth/src/main/java/org/openhab/core/io/http/auth/internal/AuthorizePageServlet.java
```diff
@@ -64,210 +55,172 @@
...
+                user = userRegistry.register(username, password, Set.of(Role.ADMIN));
+                logger.info("First user account created: {}", username);
+            } else {
+                user = login(username, password);
+            }
+
+            String authorizationCode = UUID.randomUUID().toString().replace("-", "");
+
+            if (user instanceof ManagedUser) {
+                String codeChallenge = (params.containsKey("code_challenge")) ? params.get("code_challenge")[0] : null;
```
wborn, line 173:
```suggestion
                String codeChallenge = params.containsKey("code_challenge") ? params.get("code_challenge")[0] : null;
```

bundles/org.openhab.core.io.http.auth/src/main/java/org/openhab/core/io/http/auth/internal/AuthorizePageServlet.java
```diff
@@ -64,210 +55,172 @@
...
+                String codeChallengeMethod = (params.containsKey("code_challenge_method"))
```
wborn, line 174:
```suggestion
                String codeChallengeMethod = params.containsKey("code_challenge_method")
```

bundles/org.openhab.core.io.http.auth/src/main/java/org/openhab/core/io/http/auth/internal/AuthorizePageServlet.java
```diff
@@ -64,210 +55,172 @@
...
-                }
             }
         }
     }
 
-    private String getPageBody(Map<String, String[]> params, String message) {
+    @Override
+    protected String getPageBody(Map<String, String[]> params, String message, boolean hideForm) {
         String responseBody = pageTemplate.replace("{form_fields}", getFormFields(params));
         String repeatPasswordFieldType = (isSignupMode()) ? "password" : "hidden";
```
wborn, line 208:
```suggestion
        String repeatPasswordFieldType = isSignupMode() ? "password" : "hidden";
```

bundles/org.openhab.core.io.http.auth/src/main/java/org/openhab/core/io/http/auth/internal/AuthorizePageServlet.java
```diff
@@ -64,210 +55,172 @@
...
         String buttonLabel = (isSignupMode()) ? "Create Account" : "Sign In"; // TODO: i18n
```
wborn, line 209:
```suggestion
        String buttonLabel = isSignupMode() ? "Create Account" : "Sign In"; // TODO: i18n
```

bundles/org.openhab.core.io.http.auth/src/main/java/org/openhab/core/io/http/auth/internal/AuthorizePageServlet.java
```diff
@@ -64,210 +55,172 @@
...
+        responseBody = responseBody.replace("{resultClass}", "");
         return responseBody;
     }
 
-    private String getFormFields(Map<String, String[]> params) {
+    @Override
+    protected String getFormFields(Map<String, String[]> params) {
         String hiddenFormFields = "";
 
         if (!params.containsKey(("redirect_uri"))) {
```
wborn, line 226:
```suggestion
        if (!params.containsKey("redirect_uri")) {
```

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/JwtHelper.java
```diff
@@ -150,7 +149,8 @@ public Authentication verifyAndParseJwtAccessToken(String jwt) throws Authentica
             JwtClaims jwtClaims = jwtConsumer.processToClaims(jwt);
             String username = jwtClaims.getSubject();
             List<String> roles = jwtClaims.getStringListClaimValue("role");
-            Authentication auth = new Authentication(username, roles.toArray(new String[roles.size()]));
+            String scope = jwtClaims.getStringClaimValue("scope");
+            Authentication auth = new Authentication(username, roles.toArray(new String[roles.size()]), scope);
             return auth;
```
wborn, line 154:
```suggestion
            return new Authentication(username, roles.toArray(new String[roles.size()]), scope);
```

bundles/org.openhab.core/src/main/java/org/openhab/core/auth/ManagedUser.java
```diff
@@ -59,6 +59,24 @@ public String getPasswordHash() {
...
+     * @param passwordSalt the new password salt
+     */
+    public void setPasswordSalt(String passwordSalt) {
+        this.passwordSalt = passwordSalt;
+    }
+
+    /**
+     * Alters the password hash.
+     *
+     * @param the new password hash
```
wborn, line 74:
```suggestion
     * @param passwordHash the new password hash
```

bundles/org.openhab.core/src/main/java/org/openhab/core/auth/UserRegistry.java
```diff
@@ -38,4 +38,55 @@
      * @return the new registered {@link User} instance
      */
     public User register(String username, String password, Set<String> roles);
+
+    /**
+     * Change the password for an {@link User} in this registry. The implementation receives the new password and is
+     * responsible for their secure storage (for instance by hashing the password).
+     *
+     * @param username the username of the existing user
+     * @param password the new password
```
wborn, line 47:
```suggestion
     * @param newPassword the new password
```

bundles/org.openhab.core.io.console/src/main/java/org/openhab/core/io/console/internal/extension/UserConsoleCommandExtension.java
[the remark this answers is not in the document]
ghys: I still wanted to hear opinions about this ;) Done now.

wborn:
Many thanks for addressing all my comments @ghys.

I just did some testing on a local build and it seems that duplicate sessions end up in `users.json`, they also show up in the UI so the number of user sessions did not match what I expected, so I was triggered to investigate:

```json
  "johndoe": {
    "class": "org.openhab.core.auth.ManagedUser",
    "value": {
      "name": "johndoe",
      "passwordHash": "SZXFDGrobzJMLiFNQFvQJyhkpQePuIfPDrFM2wncIxCnNszAPYBnpd/vU0GqTDHgMAB9erA3wfTXttaXyx4+HQ\u003d\u003d",
      "passwordSalt": "/L85mMFRatxrpLgsM+9JlawfhSPbez2GGTe7y1emvfzlVZOsJlnP9RVxtDA7w+X8YX8wJbMz2EJHTfaCm+WM9w\u003d\u003d",
      "roles": [
        "administrator"
      ],
      "sessions": [
        {
          "sessionId": "ef5822a1-91bb-47f5-82e1-c6390b9e9fb6",
          "refreshToken": "da8fe178ab344076bf80438fd5f5f14f",
          "createdTime": "Oct 23, 2020, 11:52:06 AM",
          "lastRefreshTime": "Oct 23, 2020, 11:52:06 AM",
          "clientId": "http://localhost:8080",
          "redirectUri": "http://localhost:8080",
          "scope": "admin",
          "sessionCookie": true
        },
        {
          "sessionId": "ef5822a1-91bb-47f5-82e1-c6390b9e9fb6",
          "refreshToken": "da8fe178ab344076bf80438fd5f5f14f",
          "createdTime": "Oct 23, 2020, 11:52:06 AM",
          "lastRefreshTime": "Oct 23, 2020, 11:52:06 AM",
          "clientId": "http://localhost:8080",
          "redirectUri": "http://localhost:8080",
          "scope": "admin",
          "sessionCookie": true
        },
        {
          "sessionId": "ddb26127-3942-43e8-a519-ec5b46cee9dc",
          "refreshToken": "0298b62ce5f444f296684627b880ed21",
          "createdTime": "Oct 23, 2020, 11:52:17 AM",
          "lastRefreshTime": "Oct 23, 2020, 11:53:18 AM",
          "clientId": "http://localhost:8080",
          "redirectUri": "http://localhost:8080",
          "scope": "admin",
          "sessionCookie": true
        }
      ],
      "apiTokens": []
    }
  },

```

But using multiple users does work nice as do the commands I've tested :+1:

ghys:
Strange, I've never noticed these session duplications before, maybe it's due to a change in this PR...?

wborn:
It is reproducible when you sign out of your session using the UI.
Instead of removing the session from the file it is duplicated.

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/auth/UserRegistryImpl.java
```diff
@@ -128,23 +133,132 @@ public User register(String username, String password, Set<String> roles) {
...
+    }
+
+    @Override
+    public void removeUserSession(User user, UserSession session) {
+        if (!(user instanceof ManagedUser)) {
+            throw new IllegalArgumentException("User is not managed: " + user.getName());
         }
+
+        ManagedUser managedUser = (ManagedUser) user;
+        managedUser.getSessions().add(session);
```
ghys, line 213: That's probably because I am not very careful :) 🙄

ghys: (and, admittedly, not too big on unit tests...)

wborn: Yes that explains!

wborn:
Shouldn't there also be a DELETE operation for a specific session? If you click the delete button on whatever session in the UI it always navigates to `/logout` and thus deletes the current session.

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/auth/UserRegistryImpl.java
wborn: I also noticed the test coverage is lacking on such critical code. But I'm OK with that if you promise to make up for that in a follow up PR. :wink:

ghys:
> Shouldn't there also be a DELETE operation for a specific session in TokenResource? If you click the delete button on whatever session in the UI it always navigates to /logout and thus deletes the current session.

Are you sure? I initially thought of it as the counterpart to `/auth/token` so it accepts the `application/x-www-form-urlencoded` content type with an `id` parameter which is the first part of the session id.
But a DELETE operation could make sense as well...

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/auth/UserRegistryImpl.java
ghys: No time like the present, I added some test coverage to the UserRegistryImpl :)

wborn: That's even better. :+1: I sometimes run out of time myself too while very enthusiastically implementing lots of new functionalities. :smile:

wborn:
I didn't notice there was an `id` parameter in the `logout` operation to make this work. So that way there is indeed no need for an operation for deleting sessions . I'll retest deleting other sessions and see if I can make some sense of what happens.

While looking at the code in `TokenResource`, I also noticed some null analysis warnings. It would be nice to see those fixed too. :slightly_smiling_face:

wborn:
> I'll retest deleting other sessions and see if I can make some sense of what happens.

I see that it can properly delete other sessions but it looks like the frontend has some logic to also remove it own session data in the browser regardless of the session it deletes.

ghys:
> While looking at the code in TokenResource, I also noticed some null analysis warnings. It would be nice to see those fixed too. 🙂

Done.

Not sure what you mean by:

>  see that it can properly delete other sessions but it looks like the frontend has some logic to also remove its own session data in the browser regardless of the session it deletes.

It shouldn't be "regardless of the session it deletes", normally it only does that if you click "Sign out of this session".

wborn (approved):
It all seems to work well now. :+1: The deleting other sessions issue doesn't seem to be related to these changes so I'll create an issue for that.

lolodomo:
@ghys: I just tried the new mode where the user REST API is fully secured.
The remote openHAB binding is working when you provide the expected token. Good.

In this mode, I tried to launch Basic UI and I was surprised to see that it works, even if it is without SSE subscriptions.
Is it expected ? That would mean that some RES API are still unsecured in this mode ?

ghys:
> In this mode, I tried to launch Basic UI and I was surprised to see that it works, even if it is without SSE subscriptions.
Is it expected ? That would mean that some RES API are still unsecured in this mode ?

Basic UI renders its pages using servlets, so it's not unexpected that it "works" (displays something) because the auth layer only applies to the REST API. You won't be able to send commands or get updates since those parts are handled with API calls.

kaikreuzer:
@ghys How easy would it be to adapt our `HttpContextFactoryServiceImpl` (which is used by those servlets) to also consider the auth configuration?

lolodomo:
I will try again but I believe BasicUI was displaying the current items values.

ghys:
> How easy would it be to adapt our HttpContextFactoryServiceImpl (which is used by those servlets) to also consider the auth configuration?

I think we can forego the need for a token for the servlets and simply rely on the session cookie - i.e. the user would need to open a session with the main UI first and then the cookie will identify the session which could be reused by Basic UI and others.
However Basic UI still needs to make API calls and for that it will need a token. Basic UI can reuse the refresh token that the main UI puts in the local storage to get an access token (that's what HABPanel does).

PR #1739, opened 2020-10-19
Add missing roles checks

ghys:
(I included these fixes in #1735 but extracted them in a stanalone
PR because it's easier to review and a little more urgent.)

As a result of the refactoring in #1713, the operations annotated with
`@RolesAllowed` containing `Role.USER` are not anymore automatically
considered accessible to all users, regardless of their actual roles.

4 operations are therefore now denied to admins if they only have the
`Role.ADMIN` role, as the first admininistrator is created only with
that role the UI encounters unexpected access denied errors and breaks.
(See https://github.com/openhab/openhab-webui/issues/422).

Closes https://github.com/openhab/openhab-webui/issues/422.


kaikreuzer (approved):
Thanks!

Issue #1751, opened 2020-10-24
OH3: selection as default sitemap element even for read only item


lolodomo:
I discovered a change in OH3.
in OH2, the default sitemap element was a selection when the item has state options and was not read only.
![image](https://user-images.githubusercontent.com/10186704/97095756-d1a8b500-1663-11eb-839d-e0aebfa1d938.png)
In OH3, we have a selection whatever if the item is read only or not.
![image](https://user-images.githubusercontent.com/10186704/97095764-ef761a00-1663-11eb-86a0-70897527be1b.png)
I consider this change as a bug as it makes no sense to provide to the user a list of options for an item which is read only. In my example, the user will not be able to change the pressure tendance ;)

lolodomo:
@cweitkamp : I imagine you could be the source of this unexpected change ;)

cweitkamp:
At a first glance the code looks okay:

https://github.com/openhab/openhab-core/blob/dd92288e9708c40d07b298fa32d23eb7ceefb00d/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java#L242-L254

I will dig into it later - e.g. adding a unit test.

lolodomo:
@cweitkamp : wait a little and don't loose your time, it could be a bug in my remote openHAB binding which would not "propagate" the read only information. I will first check without using the remote openHAB binding.

cweitkamp:
Too late :wink: . You reminded me to add a proper unit test for read-only Items. And as we are now using `CommandOptions` for rendering I saw a minor improvement in the related code.

lolodomo:
I close the issue, the problem is clearly relative to my bindiong.

PR #1754, opened 2020-10-25
[ui] Added unit test for read-only Number- and String-Items to not return a Selection Element

cweitkamp:
- Added unit test for read-only Number- and String-Items to not return a Selection Element

Related to #1751


wborn (approved):
LGTM!

PR #1755, opened 2020-10-25
[core] Replaced static inline declarations by List.of() method

cweitkamp:
- Replaced static inline declarations by `List.of()` method
- Adds `RefreshType` to accepted commands of CallItem

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/items/CallItem.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/items/CallItem.java
@@ -12,8 +12,6 @@
  */
 package org.openhab.core.library.items;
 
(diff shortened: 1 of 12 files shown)


wborn (approved):
Great! The new constructs like `List.of()` help a lot with reducing the amount of code. :-) Having them as unmodifiable constants will also reduce the the amount of garbage collection taking place. :+1:

ssalonen:
I believe this has caused regression in some bindings: https://github.com/openhab/openhab-addons/issues/8880#issuecomment-717724038

In the commit the ordering of type, at least with string item, has changed.

cweitkamp:
See https://github.com/openhab/openhab-addons/issues/8880#issuecomment-717873990

PR #1758, opened 2020-10-25
[REST Auth] Clear session cookie only when deleting own session

ghys:
Fix https://github.com/openhab/openhab-webui/issues/441


ghys:
When deleting a session from the list:
image

When deleting its own session
(using the refresh_token instead of the id prefix - the UI doesn't know the ID of its own session by design)
the cookie is cleared:
image

wborn (approved):
Nice! This will make it a lot easier to clean up unused sessions using the UI. :-)

wborn:
Did you also consider adding some more session details like last known user agent and IP to the session details @ghys?
That would make it easier to identify the sessions.

ghys:
Yes I considered it months ago. Not sure if the IP is easily retrievable (and it wouldn't always be relevant i.e. when all requests come from a reverse proxy). The user agent is quite cryptic and a mess so I figured parsing it into a user-friendly string with a library like https://github.com/nielsbasjes/yauaa could be nice - another dependency though.

PR #1770, opened 2020-10-27
[automation] Provide triggering item name in the rule context

### LINKED ISSUE #1768: OH3: Implicit variables are not initialized in DSL rules
Platform information:
  Hardware: PC
  OS: Windows 10
  Java Runtime Environment: Zulu 11
  openHAB version: 3.0.0.0 M1

I just started playing around with OH3 on an old Windows laptop.
I try to run a few DSL rules on OH3 that worked fine under OH2.
Now I encountered a problem when I try to use [implicite variables](https://www.openhab.org/docs/configuration/rules-dsl.html#implicit-variables-inside-the-execution-block) like "triggeringItem" or "receivedEvent" under OH3.
It seems like this implicite variables are not initialized under OH3 as they are with OH2.

A rule like

```csv
rule "Test"
when
   Item Test1 received command or
   Item Test2 received command
then
   logInfo("triggeringImtem", triggeringItem.name)
   //logInfo("receivedCommand", receivedCommand)
end
```

ends with an error telling that triggeringItem or receivedCommand  is null.

```csv
20:56:13.896 [ERROR] [.internal.handler.ScriptActionHandler] - Script execution failed: An error occurred during the script execution: Could not invoke method: org.openhab.core.model.script.actions.Log.logInfo(java.lang.String,java.lang.String,java.lang.Object[]) on instance: null in Test
```

Is this an OH3 bug?

kaikreuzer:
~~I have created this PR for discussion wrt whether we should merge it or not.~~
It addressed #1768, but as mentioned in that issue, it isn't easily feasible to provide the item itself, but rather only the item name.
My assumption is that this should be ok for most rules as it is really the item name that the user is after, but it nonetheless means a(nother) breaking change.

Fixes #1768

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/items/events/ItemCommandEvent.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/items/events/ItemCommandEvent.java
@@ -12,7 +12,8 @@
  */
 package org.openhab.core.items.events;
 
-import org.openhab.core.events.AbstractEvent;
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.types.Command;
 
 /**
@@ -21,14 +22,13 @@ import org.openhab.core.types.Command;
  *
  * @author Stefan Bußweiler - Initial contribution
  */
-public class ItemCommandEvent extends AbstractEvent {
+@NonNullByDefault
+public class ItemCommandEvent extends ItemEvent {
 
     /**
      * The item command event type.
      */
     public static final String TYPE = ItemCommandEvent.class.getSimpleName();
-
-    private final String itemName;
 
     private final Command command;
 
@@ -41,9 +41,9 @@ public class ItemCommandEvent extends AbstractEvent {
      * @param command the command
      * @param source the source, can be null
      */
-    protected ItemCommandEvent(String topic, String payload, String itemName, Command command, String source) {
-        super(topic, payload, source);
-        this.itemName = itemName;
+    protected ItemCommandEvent(String topic, String payload, String itemName, Command command,
+            @Nullable String source) {
+        super(topic, payload, itemName, source);
         this.command = command;
     }
 
@@ -57,6 +57,7 @@ public class ItemCommandEvent extends AbstractEvent {
      *
      * @return the item name
      */
+    @Override
     public String getItemName() {
         return itemName;
     }
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/items/events/ItemCommandEvent.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/items/events/ItemCommandEvent.java
@@ -53,16 +53,6 @@ public class ItemCommandEvent extends ItemEvent {
     }
 
     /**
-     * Gets the item name.
-     *
-     * @return the item name
-     */
-    @Override
-    public String getItemName() {
-        return itemName;
-    }
-
-    /**
      * Gets the item command.
      *
      * @return the item command
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/items/events/ItemStateChangedEvent.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/items/events/ItemStateChangedEvent.java
@@ -12,7 +12,7 @@
  */
 package org.openhab.core.items.events;
 
-import org.openhab.core.events.AbstractEvent;
+import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.openhab.core.types.State;
 
 /**
@@ -22,14 +22,13 @@ import org.openhab.core.types.State;
  *
  * @author Dennis Nobel - Initial contribution
  */
-public class ItemStateChangedEvent extends AbstractEvent {
+@NonNullByDefault
+public class ItemStateChangedEvent extends ItemEvent {
 
     /**
      * The item state changed event type.
      */
     public static final String TYPE = ItemStateChangedEvent.class.getSimpleName();
-
-    protected final String itemName;
 
     protected final State itemState;
 
@@ -46,8 +45,7 @@ public class ItemStateChangedEvent extends AbstractEvent {
      */
     protected ItemStateChangedEvent(String topic, String payload, String itemName, State newItemState,
             State oldItemState) {
-        super(topic, payload, null);
-        this.itemName = itemName;
+        super(topic, payload, itemName, null);
         this.itemState = newItemState;
         this.oldItemState = oldItemState;
     }
@@ -62,6 +60,7 @@ public class ItemStateChangedEvent extends AbstractEvent {
      *
      * @return the item name
      */
+    @Override
     public String getItemName() {
         return itemName;
     }
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/items/events/ItemStateChangedEvent.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/items/events/ItemStateChangedEvent.java
@@ -56,16 +56,6 @@ public class ItemStateChangedEvent extends ItemEvent {
     }
 
     /**
-     * Gets the item name.
-     *
-     * @return the item name
-     */
-    @Override
-    public String getItemName() {
-        return itemName;
-    }
-
-    /**
      * Gets the item state.
      *
      * @return the item state
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/items/events/ItemStatePredictedEvent.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/items/events/ItemStatePredictedEvent.java
@@ -59,16 +59,6 @@ public class ItemStatePredictedEvent extends ItemEvent {
     }
 
     /**
-     * Gets the item name.
-     *
-     * @return the item name
-     */
-    @Override
-    public String getItemName() {
-        return itemName;
-    }
-
-    /**
      * Gets the predicted item state.
      *
      * @return the predicted item state
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/items/events/ItemStatePredictedEvent.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/items/events/ItemStatePredictedEvent.java
@@ -12,7 +12,7 @@
  */
 package org.openhab.core.items.events;
 
-import org.openhab.core.events.AbstractEvent;
+import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.openhab.core.types.State;
 
 /**
@@ -26,14 +26,14 @@ import org.openhab.core.types.State;
  *
  * @author Simon Kaufmann - Initial contribution
  */
-public class ItemStatePredictedEvent extends AbstractEvent {
+@NonNullByDefault
+public class ItemStatePredictedEvent extends ItemEvent {
 
     /**
      * The item state predicted event type.
      */
     public static final String TYPE = ItemStatePredictedEvent.class.getSimpleName();
 
-    protected final String itemName;
     protected final State predictedState;
     protected final boolean isConfirmation;
 
@@ -48,8 +48,7 @@ public class ItemStatePredictedEvent extends AbstractEvent {
      */
     public ItemStatePredictedEvent(String topic, String payload, String itemName, State predictedState,
             boolean isConfirmation) {
-        super(topic, payload, null);
-        this.itemName = itemName;
+        super(topic, payload, itemName, null);
         this.predictedState = predictedState;
         this.isConfirmation = isConfirmation;
     }
@@ -64,6 +63,7 @@ public class ItemStatePredictedEvent extends AbstractEvent {
      *
      * @return the item name
      */
+    @Override
     public String getItemName() {
         return itemName;
     }
--- a/bundles/org.openhab.core.model.rule/src/org/openhab/core/model/rule/jvmmodel/RulesJvmModelInferrer.xtend
+++ b/bundles/org.openhab.core.model.rule/src/org/openhab/core/model/rule/jvmmodel/RulesJvmModelInferrer.xtend
@@ -40,6 +40,7 @@ import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import org.eclipse.xtext.common.types.JvmFormalParameter
 import org.eclipse.emf.common.util.EList
+import org.openhab.core.events.Event
 
 /**
  * <p>Infers a JVM model from the source model.</p> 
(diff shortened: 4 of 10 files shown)


cweitkamp:
Hm 🤔 ... exposing the `triggeringItem` name does not make sense to me if it is a regression at all. The Item name is already available via exposed event. Isn't it?

```
logInfo("triggeringItem", event.getItemName())
```

Should we - to be safe - cast the from `Event` to `ItemCommandEvent` before adding it to the values?

kaikreuzer:
Having the "event" object available is so far an undocumented side-effect of moving DSLs to be executed in the new rule engine. It isn't part of the [DSL documentation](https://www.openhab.org/docs/configuration/rules-dsl.html#implicit-variables-inside-the-execution-block).
But yes, you are right: It should be added there for OH3 and then `triggeringItem` becomes more or less redundant. Likewise,  we should then do another breaking change and also remove the `receivedEvent` implicit variable, since the rule engine is doing nothing else than renaming the `event` for this.

kaikreuzer:
> Should we - to be safe - cast the from Event to ItemCommandEvent before adding it to the values?

Would that make any difference? The value map will contain this object - being casted or not should just be the same, no?

kaikreuzer:
> The Item name is already available via exposed event. Isn't it?

Trying this right now, I get
```
22:14:47.798 [ool-14-thread-1] ERROR o.o.c.a.m.s.i.h.ScriptActionHandler:65 - Script execution of rule with UID 'demo-test-1' failed: The name 'event' cannot be resolved to an item or type; line 5, column 30, length 5 in demo
```
Does your example work for you or was it just a guess?

Rossko57:
Comment only - I believe `triggeringItem` - as an actual Item - is fairly extensively used in OH2 DSL user rules.   Typically it'll be where a rule is triggered by Group and the first thing we do is look at `triggeringItem.state`

A _simple_ way to replicate that is desirable, if it has be be removed.

A trade off against a newly available general "event" feels like a good selling point.

kaikreuzer:
@Rossko57 Hm, this describes a use case, which wasn't at all clear for me from the documentation.
From the docs and the name of the variable, I had assumed that the triggeringItem is the GroupItem, but if I get you correctly, it is the item that triggered the group item to change, correct? This hence is only relevant for state change triggers on group items, I guess - in all other situations, the `event` and `newState` would just be fine.

rkoshak:
@kaikreuzer, it depends on the rule trigger. 

Trigger | `triggeringItem ==`
-|-
`Member of MyGroup ...` | the member of MyGroup that triggered the rule (probably the most used use case)
`Item MyGroup ...` | `MyGroup`
`Time cron` | not defined
`System started` | not defined

The most common use case is when a rule is triggered with a `Member of` trigger as `triggeringItem` will give us the actual Item that triggered the Rule. And it isn't necessary for the Group Item to change in response to the trigger. One can have

    Member of MyGroup received update

and the rule will trigger even when an member of MyGroup receives an update to the same state. And triggeringItem will be that Item that received the update.

My vague understanding is that Member of is really kind of a macro that gets unrolled by separate triggers for each member of the Group when the rule gets loaded. That's why one has to reload the rule after changing Group membership to have the rule triggers pick up the changes. 

Based on my experience with Jython and JavaScript, event.itemName works the same way when using a Member of trigger. It has the name of the member of the Group that triggered the rule.

Rossko57:
Fair comment, it is true that implicit `newState` - which was introduced after implicit `triggeringItem` - could tackle many rule applications as alternative.

I suppose as rule users, we can clarify -
* `triggeringItem` as it exists is genuinely useful, especially with Group/member of triggers.
* but also other cases, say an ad-hoc list of several Item triggers
* It would be a serious regression not to have some substitute made available, although that is mitigated somewhat by newer implicit `newState`.   It's important that `newState` is available in **both** update and changed state-related triggers (which I think it already is).
* `triggeringItemName` is a poor but acceptable substitute, so long as there is a simple/fast way to follow up by getting access to the real Item.

It probably doesn't help any technical workings, but to throw it out there - a "real" triggeringItem but only available from member-of triggers.  An implict `triggeringMember` perhaps.

kaikreuzer:
Thanks for the clarification @rkoshak!

> the member of MyGroup that triggered the rule (probably the most used use case)

Good news on this. I have created items
```
Group:Number TestGroup
Number TestItem1 (TestGroup)
Number TestItem2 (TestGroup)
```
and a rule 
```
rule "Test"
when
   Member of TestGroup changed
then
   logInfo("triggeringItem", triggeringItem.name)
end
```
and get the output:
```
openhab:update TestItem1 4
22:56:28.074 [ool-13-thread-1] INFO  o.o.core.model.script.triggeringItem:53 - TestItem1
```
So `triggeringItem` is actually correctly available for "Member of" triggers and it is the full item and not just the name, so that you can also easily get hold of its state.

If we can agree that this is what we need and that in the other places the `newState` and the `event` are the better options anyway, I'd suggest to close this PR and rather update the documentation to make it clear that `triggeringItem` is for `Member of` triggers only. Wdyt?

seimenb:
Hi @kaikreuzer 
Many thanks for tackling that triggeringItem problem.

I tried to follow your suggestion. 
But for my purpose I would like to work with the **received command** trigger instead of **changed**.

My items
```
Group:Switch TestGroup
Switch TestItem1 (TestGroup)
Switch TestItem2 (TestGroup)
```
and my rule
```
rule "Test"
when
    Member of TestGroup received command
then
    logInfo("triggeringItem", triggeringItem.name)
end
```
Then I get following error
```
22:43:27.033 [WARN ] [hab.core.internal.events.EventHandler] - Dispatching/filtering event for subscriber 'org.openhab.core.events.EventSubscriber' failed: null
        at org.openhab.core.automation.internal.module.handler.GroupCommandTriggerHandler.receive(GroupCommandTriggerHandler.java:102) ~[?:?]
        at org.openhab.core.internal.events.EventHandler.lambda$0(EventHandler.java:152) [bundleFile:?]
        at java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source) [?:?]
        at java.util.concurrent.FutureTask.run(Unknown Source) [?:?]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source) [?:?]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source) [?:?]
        at java.lang.Thread.run(Unknown Source) [?:?]
````

The combination of **received command** trigger and **triggeringItem.name** works well on OH2.X but not on OH3 M1

**Comment by rkoshak** [comment]
There is actually one other use case I think where triggeringItem is used that I didn't cover above (because I forgot about it).

Consider the following rule triggers

```
when
    Item MyItem1 changed or
    Item MyItem2 changed or
    Item MyItem3 changed
then
```

In this case too triggeringItem used to resolve to the Item that triggered the rule. Obviously a work around would be to just put the Items into a Group and use Member of triggers. But it this issue is closed, the fact that this use case no longer works would need to be documented as a breaking change I think.

**Comment by cweitkamp** [comment]
> Does your example work for you or was it just a guess?

I did not test my example. It was a guess. I thought it would working as we assign the event variable to the result list of the triggers. But it would be really great if that would be possible.

**Comment by lolodomo** [comment]
> There is actually one other use case I think where triggeringItem is used that I didn't cover above (because I forgot about it).
> 
> Consider the following rule triggers
> 
> ```
> when
>     Item MyItem1 changed or
>     Item MyItem2 changed or
>     Item MyItem3 changed
> then
> ```
> 
> In this case too triggeringItem used to resolve to the Item that triggered the rule

This is certainly a common case that 95% of users will encounter.
That is the use case that triggers the issue with `triggeringItem`.
We absolutely need a way to know what item triggered such a rule.

**Comment by lolodomo** [comment]
@kaikreuzer : I tried using `event` or `receivedEvent` in the rule but I got the same error as you got.
Any progress on your side ?

**Comment by kaikreuzer** [comment]
@seimenb The NPE in your example was an unrelated bug that got fixed with https://github.com/openhab/openhab-core/pull/1806.

@rkoshak Your example should be covered by having the `event` in place - if I don't miss anything, my [statement from above](https://github.com/openhab/openhab-core/pull/1770#issuecomment-719820267) still holds true and we can cover also the case you mention by it: "If we can agree that this is what we need and that in the other places the newState and the event are the better options anyway"

@lolodomo No, I couldn't figure out a solution yet, but we will have to find one 😠 .

**Comment by lolodomo** [comment]
Is your idea to use "event" finally not possible?

**Comment by kaikreuzer** [comment]
As I said, we will have to find a way. Everything is possible.
Afaics, the problem is that the scripts are currently resolving Types and Items, but cannot deal with the Java class "Event" yet. Too late to further dig into this today, though...

**Comment by kaikreuzer** [comment]
Just to summarize the plan for this PR (and to fix any issue people have with `triggeringItem`):

- ' triggeringItem' already works correctly for "Member of" triggers, so there's nothing to do here
- the documentation will be updated to clarify that `triggeringItem` is only supported for "Member of" triggers
- for all other situations, the `event`variable should be used, which allows to get the triggering item name through `event. getItemName()` and its state as `newState`.
- while `newState` is correctly available, `event` is still missing and needs to be fixed as part of this PR.

**Comment by rkoshak** [comment]
Just for my understanding, `event.getItemName()` should also still work for `Member of` triggers too, right?

**Comment by kaikreuzer** [comment]
@rkoshak Afaics, this should be there as well, yes.

**Comment by kaikreuzer** [comment]
> for all other situations, the `event` variable should be used, which allows to get the triggering item name through `event. getItemName()` and its state as newState.

Ok, small change in plan here. The `event` variable didn't work well in practice, since it would only have been available as the general `Event` interface, which does not allow calling any methods like `getItemName()` without prior casting (and the additional import of the specific types). This all made it pretty complicated and cumbersome to use.
I've hence instead reactivated the idea of an `triggeringItemName` implicit variable, which is now available for every rule that has any command or state item trigger (incl. the "member of" trigger).

This works quite well and I have updated the PR accordingly, so this is ready for review.

N.B.: I also noticed that the `receivedEvent` was the raw event instead of the channel event string - I have fixed this as well, which makes it much better to use. I'll cover it in a documentation PR update.

**Comment by kaikreuzer** [comment]
I've produced the according documentation PR as https://github.com/openhab/openhab-docs/pull/1294.
Additionally, I have explained the change in on https://github.com/openhab/openhab-distro/wiki/Breaking-Changes-in-openHAB-3#rules.

@openhab/core-maintainers Imho everything ready for review/merge!

**Comment by lolodomo** [comment]
Too late for the milestone 4. I hope we can test that soon.

**Comment by kaikreuzer** [comment]
Test it on a snapshot as soon as it is merged!

**Comment by lolodomo** [comment]
Of course I will ;)

File: bundles/org.openhab.core/src/main/java/org/openhab/core/items/events/ItemCommandEvent.java [hunk 1]
```diff
@@ -57,6 +57,7 @@ public String getType() {
      *
      * @return the item name
      */
+    @Override
     public String getItemName() {
```
cweitkamp, line 61: This can be removed.

bundles/org.openhab.core/src/main/java/org/openhab/core/items/events/ItemStateChangedEvent.java
```diff
@@ -62,6 +60,7 @@ public String getType() {
      *
      * @return the item name
      */
+    @Override
     public String getItemName() {
```
cweitkamp, line 64: This can be removed as it the same like defined in the abstract `ItemEvent` class.

bundles/org.openhab.core/src/main/java/org/openhab/core/items/events/ItemStatePredictedEvent.java
```diff
@@ -64,6 +63,7 @@ public String getType() {
      *
      * @return the item name
      */
+    @Override
     public String getItemName() {
```
cweitkamp, line 67: See above.

cweitkamp (commented):
Thanks. I think it is time to merge this. I left few minor comment.

kaikreuzer:

cweitkamp (approved):
LGTM. Thank you very much.

lolodomo:
I can confirm `triggeringItemName` is working well in the standard case when you have several triggers on different items.
One major issue fixed.

kaikreuzer:
Hooray 🎉

morph166955:
I can also confirm this works.  I've reenabled all of my rules that used triggeringItem.name to now use this and it seems to work just fine.  Thank you!!!

seimenb:
Tested with 3.0.0.M5.
Just replaced the ```triggeringItem.Name``` with ```triggeringItemName``` in all of my rule files.
Pam! And it works!

@kaikreuzer Thank you so much for the easy workaround.

image

image

ratza:
@kaikreuzer 
> N.B.: I also noticed that the `receivedEvent` was the raw event instead of the channel event string - I have fixed this as well, which makes it much better to use. I'll cover it in a documentation PR update.

I'm afraid the documented breaking changes don't mention that the `receivedEvent` variable is a String now. As a result the `receivedEvent.getEvent()` construction used quite often in documentation examples, would fail with a pretty cryptic message. An example taken out of the [rule documentation](https://next.openhab.org/docs/configuration/rules-dsl.html) is:
```
rule "Start wake up light on sunrise"
when
    Channel "astro:sun:home:rise#event" triggered
then
    switch(receivedEvent.getEvent()) {
        case "START": {
            Light.sendCommand(ON)
        }
    }
end
```

I've seen this suggested in the [Xiaomi Mi Smart Home Binding](https://next.openhab.org/addons/bindings/mihome/) as well.

This seems like a bigger change to me, since the `ChannelTriggeredEvent` object does have more methods that people could've used in their rules. From the documentation and forums, I don't see an explicit explanation of the supported fields/methods on this variable. I've seen `.getEvent()` and `.toString()` calls from the quick check I did.

I've only started evaluating OpenHAB for my own use, so I don't know much about the project or if this is intended or not. I can, of course, look through the documentation and correct the bits that I see, but since this also means going through each binding repo, wanted to make sure this is as planned.

morph166955:
@kaikreuzer I've had one issue with this.  I have some rules which are fired by item, cron, and system start.  When the rule is fired by the item, this works as expected.  Obviously no item fires a cron or system start.  This causes the output to the log file to be null logInfo("TEST","Trigger {}",triggeringItemName) comes out as Trigger {} in the log.  I would very much like to know that the rule was fired by the cron or the system startup.  Is there any way to account for this and set the string to "cron" or something like that?

ccutrer:
Sorry to comment directly on the code first without reading the thread. It sounds like it's on purpose that receivedEvent now returns a string with the value of the event, instead of the event object itself. But the documentation update didn't actually happen (https://github.com/openhab/openhab-distro/wiki/Breaking-Changes-in-openHAB-3#rules doesn't mention it), and I doubt the examples have been updated either. Critically, it is now no longer possible to access which channel triggered an event if a rule is set up to handle multiple triggers at the same time. Maybe a new implicit variable should be added for this case, such as `triggeringChannelName`?

kaikreuzer:
>  But the documentation update didn't actually happen

It is mentioned in the release notes, see https://github.com/openhab/openhab-distro/releases/tag/3.0.0#breaking-changes-that-require-manual-interaction-after-the-upgrade.
The examples in bindings might indeed not yet been updated.
Implicit variable might be an option, if that is a regular use case (I myself didn't come across such yet).

hamwong:
> I've seen this suggested in the [Xiaomi Mi Smart Home Binding](https://next.openhab.org/addons/bindings/mihome/) as well.
> 
> This seems like a bigger change to me, since the `ChannelTriggeredEvent` object does have more methods that people could've used in their rules. From the documentation and forums, I don't see an explicit explanation of the supported fields/methods on this variable. I've seen `.getEvent()` and `.toString()` calls from the quick check I did.

I still see it in Xiaomi page....

PR #1774, opened 2020-10-28
[core] Revert ordering of accepted data types for 'StringItem'

cweitkamp:
- Revert ordering of accepted data types for `StringItem`

See https://github.com/openhab/openhab-core/pull/1755 and https://github.com/openhab/openhab-addons/issues/8880#issuecomment-718200422

I tried to move `DateTimeType` too but this probably will result in more conflicts because it now can parse Unix timestamp and every positive integer will be parsed as `DateTimeType`.


kaikreuzer (approved):
Thanks for adding the tests as well!

PR #1775, opened 2020-10-30
[infrastructure] add external null-annotations

wborn:
Add EEAs and fix null analysis errors.

Related to:

* #888
* openhab/openhab-addons#8848

### DIFF
--- a/bundles/org.openhab.core.addon.sample/.classpath
+++ b/bundles/org.openhab.core.addon.sample/.classpath
@@ -9,11 +9,13 @@
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11">
 		<attributes>
 			<attribute name="maven.pomderived" value="true"/>
+			<attribute name="annotationpath" value="target/dependency"/>
 		</attributes>
 	</classpathentry>
 	<classpathentry kind="con" path="org.eclipse.m2e.MAVEN2_CLASSPATH_CONTAINER">
 		<attributes>
 			<attribute name="maven.pomderived" value="true"/>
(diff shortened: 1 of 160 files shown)


wborn:
Can we make some progress here @openhab/core-maintainers ?

cweitkamp (approved):
LGTM. Thanks.

cweitkamp:
Just as a side note. I recently discovered a NPE in [AbstractResourceBundleProvider#L243](https://github.com/openhab/openhab-core/pull/1775/files#diff-abe5ef1432d6cd35774add02be012ecff1cc38dd7314b82e36dab965b8a4a752R243). The method `getSymbolicName()` of `Bundle.class` can return a `null` value and we do not check on it. Do you know if there are null-annotations for OSGi too?

wborn:
You can add EEAs for any class. :-) There isn't a lot for OSGi yet:

https://github.com/lastnpe/eclipse-null-eea-augments/tree/master/libraries/osgi-core

I created a PR for adding more classes we use in core with   https://github.com/lastnpe/eclipse-null-eea-augments/pull/110

cweitkamp:
Thanks for the hint.

kaikreuzer:
Pulling in this change, my openab.core bundle does not compile anymore in Eclipse with these errors:

Screenshot 2020-11-08 at 20 56 20

What do I need to do to resolve this?

wborn:
It will unpack the EEAs to target/dependency when you run for instance `mvn test`

I seem to have the "m2e connector for the Maven Dependency Plugin" installed so I don't see those plugin execution errors. Can you test if it works well if you set the execution to ignore in the root POM?

kaikreuzer:
To verify, I did a plain new IDE install on a green field and can confirm that this PR introduced an issue, see https://github.com/openhab/openhab-core/issues/1850.

PR #1780, opened 2020-10-31
Remove Map null annotation workarounds

wborn:
These workarounds to prevent false positives can be removed now the EEAs allow for proper null analysis.

Depends on #1775


wborn:
Since #1775 just got merged, we can also make some progress here @openhab/core-maintainers. :wink:

cweitkamp (approved):
LGTM. Thanks.

PR #1794, opened 2020-11-03
Print 'ItemChannelLink' configuration in output of console commands

cweitkamp:
- Print `ItemChannelLink` configuration in output of console commands
- Renamed subcommands to `link` / `unlink`

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/console/LinkConsoleCommandExtension.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/console/LinkConsoleCommandExtension.java
@@ -49,8 +49,8 @@ import org.osgi.service.component.annotations.Reference;
 public class LinkConsoleCommandExtension extends AbstractConsoleCommandExtension {
 
     private static final String SUBCMD_LIST = "list";
-    private static final String SUBCMD_CL_ADD = "addChannelLink";
-    private static final String SUBCMD_CL_REMOVE = "removeChannelLink";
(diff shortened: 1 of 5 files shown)


cweitkamp:
Another way would be to overwrite the `toString()` method in `ItemChannelLink` but I cannot foresee the impact.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/console/LinkConsoleCommandExtension.java
```diff
@@ -89,7 +90,8 @@ public void execute(String[] args, Console console) {
                         ChannelUID channelUID = new ChannelUID(args[2]);
                         addChannelLink(console, itemName, channelUID);
                     } else {
-                        console.println("Specify item name and channel UID to link: link <itemName> <channelUID>");
+                        console.println(
+                                "Specify item name and channel UID to link: addChannelLink <itemName> <channelUID>");
```
kaikreuzer, line 94: `addChannelLink` comes from a time when there were also `ThingLinks`. Since they do not exist anymore, I wonder if we shouldn't rather simplify the command to be simply `link` and `unlink` again, just as the help text says. Wdyt?

cweitkamp: Yes, I agree. We are using it in the new UI too and it should be the new openHAB naming. I will change it.

kaikreuzer (approved):
thanks!

PR #1799, opened 2020-11-05
[automation] Create Nashorn script engines with the proper class loader

ghys:
This should fix the issue reported here:
https://community.openhab.org/t/openhab-3-0-milestone-2-discussion/107564/8

where the Nashorn script engine would be created with the
current thread's class loader, causing JS code like this:
```
var Log = Java.type("org.openhab.core.model.script.actions.Log");
Log.logError("Experiments", "This is an OH error log");
Log.logWarn("Experiments", "This is an OH warn log");
Log.logInfo("Experiments", "This is an OH info log");
Log.logDebug("Experiments", "This is an OH debug log");
```
to run fine when the rule was triggered but fail to find the Log
class when run from the REST API's `/rest/rules/{ruleUID}/runnow`,
because in that case the generic createScriptEngine implementation
would return script engines using the JAX-RS class loader as the
"app" class loader.

Note:
We also have an opportunity to restrict which classes are exposed
to the script with a ClassFilter to a specific set:
https://docs.oracle.com/javase/8/docs/jdk/api/nashorn/jdk/nashorn/api/scripting/NashornScriptEngineFactory.html#getScriptEngine-java.lang.String:A-java.lang.ClassLoader-jdk.nashorn.api.scripting.ClassFilter-
This could prove useful to mitigate code execution vulnerabilities,
as the script code is modifiable remotely.

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/NashornScriptEngineFactory.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/NashornScriptEngineFactory.java
@@ -24,6 +24,7 @@ import javax.script.ScriptEngine;
 import javax.script.ScriptException;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.automation.module.script.AbstractScriptEngineFactory;
 import org.openhab.core.automation.module.script.ScriptEngineFactory;
 import org.osgi.service.component.annotations.Component;
@@ -33,12 +34,21 @@ import org.osgi.service.component.annotations.Component;
  *
  * @author Simon Merschjohann - Initial contribution
  * @author Scott Rushworth - removed default methods provided by ScriptEngineFactory
+ * @author Yannick Schaus - create script engines with the bundle's class loader as "app" class loader
  */
 @NonNullByDefault
 @Component(service = ScriptEngineFactory.class)
 public class NashornScriptEngineFactory extends AbstractScriptEngineFactory {
 
     private static final String SCRIPT_TYPE = "js";
+
+    @SuppressWarnings("removal")
+    jdk.nashorn.api.scripting.NashornScriptEngineFactory nashornScriptEngineFactory;
+
+    @SuppressWarnings({ "deprecation", "removal" })
+    public NashornScriptEngineFactory() {
+        nashornScriptEngineFactory = new jdk.nashorn.api.scripting.NashornScriptEngineFactory();
+    }
 
     @Override
     public List<String> getScriptTypes() {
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/NashornScriptEngineFactory.java
```diff
@@ -72,4 +80,25 @@ public void scopeValues(ScriptEngine scriptEngine, Map<String, Object> scopeValu
...
+    @Override
+    public @Nullable ScriptEngine createScriptEngine(String scriptType) {
+        try {
+            for (javax.script.ScriptEngineFactory f : ENGINE_MANAGER.getEngineFactories()) {
+                List<String> mimeTypes = f.getMimeTypes();
+                List<String> extensions = f.getExtensions();
+
+                if (mimeTypes.contains(scriptType) || extensions.contains(scriptType)) {
+                    Method method = f.getClass().getMethod("getScriptEngine", ClassLoader.class);
+                    return (ScriptEngine) method.invoke(f, NashornScriptEngineFactory.class.getClassLoader());
```
wborn, line 93: If the classloader that creates the component is also usable, you might be able to get rid of this reflection and reuse the same `ScriptFactory` in `createScriptEngine` like I did with Groovy by just assigning the `ScriptFactory` to a field.

See: [GroovyScriptEngineFactory](https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.automation.groovyscripting/src/main/java/org/openhab/automation/groovyscripting/internal/GroovyScriptEngineFactory.java)

ghys: That's what I did initially, but I got resolution failures (https://ci.openhab.org/job/PR-openHAB-Core/2682/console):
`missing requirement [org.openhab.core.automation.module.script/3.0.0.202011051754] osgi.wiring.package; filter:="(osgi.wiring.package=jdk.nashorn.api.scripting)"]]`
and I didn't know how to solve them 😞

ghys:
I have finally replaced that ugly reflection: now it simply temporarily replaces the thread's context class loader, let the abstract factory create the script engine with its default parameter-less signature, and switches the context back to the original class loader - safe & easy.

And it still works:
```js
var log = Java.type("org.openhab.core.model.script.actions.Log");
log.logInfo("TestLog", "This is a test log");
```
when run directly from the UI with Ctrl-R:
```
12:48:43.508 [INFO ] [org.openhab.core.model.script.TestLog] - This is a test log
```

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/NashornScriptEngineFactory.java
wborn: It's probably because it is considered to be internal JDK package and it may not be available in each JVM implementation. Since it is also deprecated it's probably best not to depend on it also because Nashorn has been removed since Java 15.

wborn (approved):
Thanks for the fix!

PR #1807, opened 2020-11-11
[REST Auth] Remove user access to things

ghys:
I can't think of a good reason why listing things or querying their status should be allowed for users.
The things layer should only be of concern to admins IMHO.
As noted here: https://community.openhab.org/t/oh3-will-list-all-your-things-even-if-you-are-not-logged-in/108006/3
passwords and other sensible information in configuration could end up being exposed without auth required.

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/thing/ThingResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/thing/ThingResource.java
@@ -289,7 +289,7 @@ public class ThingResource implements RESTResource {
     }
 
     @GET
-    @RolesAllowed({ Role.USER, Role.ADMIN })
+    @RolesAllowed({ Role.ADMIN })
     @Produces(MediaType.APPLICATION_JSON)
     @Operation(summary = "Get all available things.", responses = {
             @ApiResponse(responseCode = "200", description = "OK", content = @Content(array = @ArraySchema(schema = @Schema(implementation = EnrichedThingDTO.class), uniqueItems = true))) })
@@ -510,7 +510,7 @@ public class ThingResource implements RESTResource {
     }
 
     @GET
-    @RolesAllowed({ Role.USER, Role.ADMIN })
+    @RolesAllowed({ Role.ADMIN })
     @Path("/{thingUID}/status")
     @Produces(MediaType.APPLICATION_JSON)
     @Operation(summary = "Gets thing status.", security = {


lolodomo:
I disagree.
I use it in the remote openHAB binding and I would not like to make the use of this binding very difficult.

Thing status is a user information, something that can be used in a rule, not an administration information.

And sensible information in thing configuration is hidden as soon as `<context>password</context>` is set.

kaikreuzer:
@lolodomo Things are meant by design for admins only.
For the remote openHAB binding, you should use admin tokens, if any thing related information is required.

lolodomo:
That is true that the "password" settings are hidden in the UI but not in the repsonse of the REST API.

lolodomo:
> For the remote openHAB binding, you should use admin tokens, if any thing related information is required.

But this is not something that the binding can create automatically ?
The user will have to create a token manually and then use it as parameter in the binding.
Of course, it will work (I already tested the use of token) but it will make unfortunately the access to the binding more difficult for a standard user.
In this case, I will have to make the token setting a required thing setting.

kaikreuzer:
> The user will have to create a token manually and then use it as parameter in the binding.

That's imho absolutely ok and correct.
I assumed though, that your "basic"/first version of the binding was only using the item level for integration - in that case there would be no need for admin access.

lolodomo:
> I assumed though, that your "basic"/first version of the binding was only using the item level for integration - in that case there would be no need for admin access.

Yes, that is correct.

By the way, the new version of the binding also uses API GET /things/{thingUID} which is already protected. Of course, it was working with OH2 remote server, but will not with OH3 server.
I will keep the token setting optional. If not set, some features of the binding will not be active. And the documentation has to be enhanced to explain all that.

lolodomo:
Please note that if I follow your logic, the SSE API should also be protected, depending on the topics you subscribe ?
If not, you can open a SSE link and get data from things when they are updated.
The difficulty is that it will depend on the topic(s) you set as parameter in your URL.

ghys:
> Please note that if I follow your logic, the SSE API should also be protected

That's very true, especially since events tend to dump the entire related object in the payload.
At least `openhab/things/*` and `openhab/rules/*` events should be filtered out unless the user is an admin.

lolodomo:
For the thing status API, I think we could keep the USER role, there is no sensible data returned and this is rather a user information.

lolodomo:
FYI, this is correctly handled in the remoteopenhab binding, I just checked.

Even if I don't use the thing status API in the remoteopenhab binding, I believe it should not be "protected".

kaikreuzer:
> At least openhab/things/* and openhab/rules/* events should be filtered out unless the user is an admin.

@ghys Do you think there is an easy way to accomplish this?

ghys:
> @ghys Do you think there is an easy way to accomplish this?

On the server side, maybe here:
https://github.com/openhab/openhab-core/blob/94d6a804938bb989bb3736a2621b0d893b51b00f/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/SseResource.java#L164
we could add the current user's role in the SseSinkTopicInfo constructor.
then implement a "isVisibleForRole" predicate so that https://github.com/openhab/openhab-core/blob/94d6a804938bb989bb3736a2621b0d893b51b00f/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/SseResource.java#L173
would become:
```
topicBroadcaster.sendIf(sseEvent, matchesTopic(eventDTO.topic).and(isVisibleToRole()));
```

However, on the client side the EventSource cannot send credentials (discussed [here](https://community.openhab.org/t/solved-oh3-new-ui-not-displaying-item-states/107906/5)) by default, it's only possible with a polyfill, so we'll want to make sure the credentials are indeed passed correctly before filtering because we still want admin to receive admin events when logged in.

> Even if I don't use the thing status API in the remoteopenhab binding, I believe it should not be "protected".

And I believe it still should. This seems illogical if the things themselves are restricted. Besides, IMHO the things are an admin-only concept as a whole, and users should completely oblivious of it, this includes their statuses.

kaikreuzer:
> we'll want to make sure the credentials are indeed passed correctly before filtering

Definitely. So if I get you correctly, you'd first add that in the UI. Do you plan to do so with the polyfill or are there any drawbacks with that?

> And I believe it still should. This seems illogical if the things themselves are restricted.

I agree with that - when not having access to Things, there isn't much need to see a status for the stuff one does not see.

ghys:
> Definitely. So if I get you correctly, you'd first add that in the UI. Do you plan to do so with the polyfill or are there any drawbacks with that?

I'll make some tests. The drawback is that for instance you cannot see the messages in real-time in the developer tools when you're using the polyfill, and of course that you're basically trusting a library's author to provide support for the API that you need (with extensions), instead of the (flawed) standard agreed upon by browser makers.

kaikreuzer (approved):

PR #1819, opened 2020-11-14
[automation] Create separate DSLScriptEngine instance for every handler

kaikreuzer:
Fixes https://github.com/openhab/openhab-core/issues/1818


lolodomo:
It may solve the issue with "null context" too ;) We will see...

wborn (changes requested):
Thanks for the fix! There seems to be a unit test failing now:

```
[ERROR] Failed to execute goal biz.aQute.bnd:bnd-testing-maven-plugin:5.1.2:testing (default) on project org.openhab.core.automation.module.script.tests: 1 errors found -> [Help 1]

TEST org.openhab.core.automation.module.script.ScriptRuleOSGiTest#testPredefinedRule() <<< ERROR: 
Expected: not null
     but: was null
java.lang.AssertionError: 
Expected: not null
     but: was null
```

kaikreuzer:
Seems the tests are right - the change didn't work for Javascript rules :-/.
I've now come up with a different (and hopefully better) solution.

wborn (approved):
LGTM now!

PR #1827, opened 2020-11-16
[rest] Add summary option to rules, things, UI components resources

ghys:
The /things, /rules, /ui/components endpoints retrieve all objects
in their entirety, which can become very big, i.e. channels, config
parameters, script rule modules or trees of UI components can
quickly add up to the size.

When the UI simply displays a list of those objects it retrieves all
this extra information but does nothing with it.

This introduces an optional ?summary=true query parameter for the
above resources to limit the output to pre-defined fields which are
deemed most relevant for displaying these lists, omitting the rest.

When the option is not set, the behavior remains unchanged so this
change is not API breaking. The API version has therefore not been
incremented. The client is responsible for adding the option to
retrieve summarized collections instead of the entire objects.


kaikreuzer (approved):
lgtm and is probably helpful for the performance of the UI.

PR #1833, opened 2020-11-17
Remove the groupname from the members if the group item is removed

### LINKED ISSUE #1392: Managed Items are not removed from deleted group
I noticed that when I deleted a managed group item using the paperui, its children are not removed automatically from the group. Shouldn't this be the correct behavior for managed items?

### LINKED ISSUE #1785: [group handling] deletion of parent group is not propagated to child groups
Running on build #1998.

While working with the semantic model i deleted my root group and found that the clid groups held still the root group as parent group. I check that in the json db too and there the attrubute groupNames still hold the parent/root group.

Maybe one option ist  not allow a deletion if a group is not empty.

silamon:
Remove the groupname from the members if the group item is removed. 
This is implemented in the ManagedItemProvider. 

Fixes #1785 
Fixes #1392

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/items/ManagedItemProvider.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/items/ManagedItemProvider.java
@@ -83,6 +83,16 @@ public class ManagedItemProvider extends AbstractManagedProvider<Item, String, P
         this.itemBuilderFactory = itemBuilderFactory;
     }
 
+    @Override
+    public @Nullable Item remove(String key) {
+        Item item = this.get(key);
+        if (item instanceof GroupItem) {
+            this.removeGroupNameFromMembers((GroupItem) item);
+        }
+
+        return super.remove(key);
+    }
+
     /**
      * Removes an item and it´s member if recursive flag is set to true.
      *
@@ -118,6 +128,16 @@ public class ManagedItemProvider extends AbstractManagedProvider<Item, String, P
         return memberNames;
     }
 
+    private Set<Item> getMembers(GroupItem groupItem, Collection<Item> allItems) {
+        Set<Item> members = new HashSet<>();
+        for (Item item : allItems) {
+            if (item.getGroupNames().contains(groupItem.getName())) {
+                members.add(item);
+            }
+        }
+        return members;
+    }
+
     private @Nullable Item createItem(String itemType, String itemName) {
         try {
             Item item = itemBuilderFactory.newItemBuilder(itemType, itemName).build();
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core/src/main/java/org/openhab/core/items/ManagedItemProvider.java
```diff
@@ -118,6 +128,16 @@ public ManagedItemProvider(final @Reference StorageService storageService,
         return memberNames;
     }
 
+    private Set<Item> getMembers(GroupItem groupItem, Collection<Item> allItems) {
+        Set<Item> members = new HashSet<>();
+        for (Item item : allItems) {
```
cweitkamp, line 133: Wouldn't it be easier to iterate over `group.getMembers()`?

silamon: If that would be possible, the other recursive function in the `ManagedItemProvider `could use that as well. But `group.getMembers()` is empty since you're fetching the Item at the start of `Remove(string key)` from the `StorageService`.

cweitkamp: I think digging into it and research why group members are empty needs too much time. Considering how often a group will be deleted it should be okay for now to accept the iteration over all items. Wdyt?

bundles/org.openhab.core/src/main/java/org/openhab/core/items/ManagedItemProvider.java
```diff
@@ -83,6 +83,16 @@ public ManagedItemProvider(final @Reference StorageService storageService,
         this.itemBuilderFactory = itemBuilderFactory;
     }
 
+    @Override
+    public @Nullable Item remove(String key) {
+        Item item = this.get(key);
```
cweitkamp, line 88: Please remove the `this` where possible.

cweitkamp (changes requested):

bundles/org.openhab.core/src/main/java/org/openhab/core/items/ManagedItemProvider.java
silamon: No need to research that, I think. There's an `initializeItem(Item item)` in the `ItemRegistry` where members get filled in. If I want to have an item filled in with members, I would need to get that item from the `ItemRegistry` on removal or the remove function would need to be adjusted to contain the item instead of the key. 

I can't use the `ItemRegistry` removal event since it gets called too whenever a provider get disposed (https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core/src/main/java/org/openhab/core/common/registry/AbstractRegistry.java#L640) . Making a dedicated service with a reference to the `ManagedItemProvider` and `ItemRegistry`, so I can listen on `ManagedItemProvider` removals and then fetch the item from `ItemRegistry` would only work if my service is being called before the listener of `ItemRegistry`, which cannot be guaranteed.

cweitkamp: Thank you for explanation. Let's go for the current solution.

cweitkamp (approved):
LGTM. Thanks.

PR #1837, opened 2020-11-19
[automation] Pass script context to script engines

jpg0:
Pass explicit script context to script engines. Context initially includes only engine identifier.

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
@@ -12,14 +12,18 @@
  */
 package org.openhab.core.automation.module.script.internal;
 
+import static org.openhab.core.automation.module.script.ScriptEngineFactory.CONTEXT_KEY_ENGINE_IDENTIFIER;
+
 import java.io.InputStreamReader;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
 import javax.script.Invocable;
+import javax.script.ScriptContext;
 import javax.script.ScriptEngine;
 import javax.script.ScriptException;
+import javax.script.SimpleScriptContext;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
```diff
@@ -129,6 +134,11 @@ private boolean isCustomFactory(ScriptEngineFactory engineFactory) {
                     loadedScriptEngineInstances.put(engineIdentifier, result);
                     logger.debug("Added ScriptEngine for language '{}' with identifier: {}", scriptType,
                             engineIdentifier);
+
+                    SimpleScriptContext scriptContext = new SimpleScriptContext();
+                    scriptContext.setAttribute(CONTEXT_KEY_ENGINE_IDENTIFIER, engineIdentifier, ScriptContext.ENGINE_SCOPE);
+                    engine.setContext(scriptContext);
```
kaikreuzer, line 140: Is it safe to set that here? Couldn't the script engine have already set up some initial context?
I'm especially concerned about the DSLScriptEngine, where I recently had a lot of trouble getting the handling of the context stable... Would be good if you could test that there's no issue by this on DSL rules.

jpg0: @kaikreuzer Good point; I was primarily thinking about the newer script engines which I don't believe set things up here. TBH my main goal here is passing context to script engine factories upon creation of script engines to allow them to be tailored.

The cleanest way to do this would be to modify the `engineFactory.createScriptEngine(scriptTypes.get(0));` call eariler in the method to add a custom `ScriptEngineContext` object as a second argument. However I chose not to do this because it would be a breaking change for existing implementors of the interface declaring this method (of course I could also introduce a 2nd interface and collect implementations of both, but that seems overkill). Or do you think this would be a better option? If not, I'll update the code to add context to whatever is there already (and create if not present) before testing on the DSLScriptEngine.

jpg0: (Updated code to only set context if it's not already been set)

kaikreuzer: I think the current solution is fine, thanks.

kaikreuzer:
Did you notice that the build fails due to spotless errors?

jpg0:
I had not, thanks for pointing it out. Build is now passing.

kaikreuzer (approved):
thanks!

PR #1843, opened 2020-11-23
[automation] Extracted accessor into interface and added to script engine context …

jpg0:
…to allow access to script extensions from script engine factories

This is to allow script engines the ability to reference extensions (via the script context). Note that I have called the methods `findX` rather than `getX` because `getDefaultPresets` already exists, yet returns the preset names, not the presets themselves.

(I will move out of draft once I've finished testing.)


jpg0:
I've now tested this with both existing (groovy) and new (graaljs) and it all looks good, so I'll take out of draft.

kaikreuzer (approved):
lgtm

PR #1844, opened 2020-11-23
removed deprecated author information from bindings

kaikreuzer:
This was discussed a long time ago and was planned for the start of openHAB 3 already - it fell of my plate somehow.

In short: Having a single author listed in this file does not make much sense for multiple reasons:
- Bindings in the official repo are considered to be part of the openHAB project and not by an individual.
- Many bindings have different contributors over the course of time, which isn't reflected here either.
- The attribution of contributors is done within the source code.
- We don't have this info for other types of add-on either.
- Naming an author makes sense for marketplace entries, but for those we will anyhow have other metadata files.
- The original purpose of this field was to give companies a possibility to "officially claim" a binding to be implemented and supported by them, but this actually never worked that way.

### DIFF
--- a/itests/org.openhab.core.binding.xml.tests/src/main/resources/test-bundle-pool/BundleInfoTest.bundle/OH-INF/binding/binding.xml
+++ b/itests/org.openhab.core.binding.xml.tests/src/main/resources/test-bundle-pool/BundleInfoTest.bundle/OH-INF/binding/binding.xml
@@ -6,8 +6,6 @@
 	<name>hue Binding</name>
 	<description>The hue Binding integrates the Philips hue system. It
 		allows to control hue lights.</description>
-	<author>Deutsche Telekom AG</author>
-
 
 	<!-- Dummy config -->
 	<config-description>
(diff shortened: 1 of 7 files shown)


cweitkamp:
I think we should go one step (or two steps) further. The OHC code does contain quite some [test bundles which uses author-tags](https://github.com/openhab/openhab-core/search?l=XML&q=author&type=) as well. They should be removed as a rolemodel. Furthermore we can think about removing it from the binding XSD. Another open question would be the `BundingInfo` and it's related DTO.

kaikreuzer:
Thanks - I fixed the itests.
Wrt the XSD and the related class/DTO: I didn't feel that this really justifies a breaking change in the schema (with all related efforts that come with that). I'd preferred a "smooth" phaseout by simply not using this "feature" anymore and thus considering it deprecated.
If you think it is cleaner to also adapt the XSD, let us do so as a second step.

cweitkamp:
Okay, just remove the tags from itest. As long as we do not touch the XSD we should keep the related test for parsing in place. this currently fails.

```
07:57:08.873 [main] DEBUG org.openhab.core.config.xml.osgi.XmlDocumentBundleTracker - Releasing the XmlDocumentProvider for module '@bundleId@0x2e'.
TEST org.openhab.core.binding.xml.test.BindingInfoTest#assertThatBindingInfoIsReadProperly() <<< ERROR: 
Expected: is "Deutsche Telekom AG"
     but: was null
java.lang.AssertionError: 
Expected: is "Deutsche Telekom AG"
     but: was null
```

kaikreuzer:
Stupid me to also remove the entry from the test... Should be fixed now.

cweitkamp (approved):
LGTM. Thanks.

PR #1854, opened 2020-11-28
[automation] Added Actions for Play and Say which sets the volume

cweitkamp:
- Added Actions for Play and Say which sets the volume

### DIFF
--- a/bundles/org.openhab.core.automation.module.media/src/main/java/org/openhab/core/automation/module/media/internal/MediaActionTypeProvider.java
+++ b/bundles/org.openhab.core.automation.module.media/src/main/java/org/openhab/core/automation/module/media/internal/MediaActionTypeProvider.java
@@ -46,6 +46,7 @@ import org.osgi.service.component.annotations.Reference;
  *
  * @author Kai Kreuzer - Initial contribution
  * @author Simon Kaufmann - added "say" action
+ * @author Christoph Weitkamp - Added parameter volume
  */
 @NonNullByDefault
 @Component(service = ModuleTypeProvider.class)
@@ -64,12 +65,8 @@ public class MediaActionTypeProvider implements ModuleTypeProvider {
         switch (UID) {
             case PlayActionHandler.TYPE_ID:
                 return getPlayActionType(locale);
-            case PlayActionHandler.VOLUME_TYPE_ID:
-                return getPlayWithVolumeActionType(locale);
             case SayActionHandler.TYPE_ID:
                 return getSayActionType(locale);
-            case SayActionHandler.VOLUME_TYPE_ID:
-                return getSayWithVolumeActionType(locale);
             default:
                 return null;
         }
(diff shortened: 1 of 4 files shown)


kaikreuzer:
Hm, I am a bit scared of action proliferation by having dedicated types for every parameter combination.
Can't the volume simply be added as an optional parameter to the existing action types?

cweitkamp:
In general I tend to agree. Do you know if it will be backwards compatible regarding JSONDB / DTOs? I assume yes, because configuration parameters are a list.

kaikreuzer:
I'd strongly think so as an optional parameter should not be required and thus it should be fine to not have it in the db for older entries. But it might be safer, if you do a quick test of that assumption.

cweitkamp:
I will do that. If it is working I will change this PR accordingly. And I will submit a minor patch for https://github.com/openhab/openhab-addons/pull/9145 too.

cweitkamp:
Works flawlessly. I updated this PR.

bundles/org.openhab.core.automation.module.media/src/main/java/org/openhab/core/automation/module/media/internal/MediaActionTypeProvider.java
```diff
@@ -62,43 +62,45 @@ public MediaActionTypeProvider(final @Reference AudioManager audioManager) {
...
     @Override
     public Collection<ModuleType> getModuleTypes(@Nullable Locale locale) {
-        return Stream.of(getPlayActionType(locale), getSayActionType(locale)).collect(Collectors.toList());
+        return List.of(getPlayActionType(locale), getSayActionType(locale));
     }
 
     private ModuleType getPlayActionType(@Nullable Locale locale) {
         return new ActionType(PlayActionHandler.TYPE_ID, getConfigPlayDesc(locale), "play a sound",
-                "Plays a sound file.", null, Visibility.VISIBLE, new ArrayList<>(), new ArrayList<>());
+                "Plays a sound file. Optionally set the volume.", null, Visibility.VISIBLE, null, null);
```
kaikreuzer, line 82:
```suggestion
                "Plays a sound file. Optionally sets the volume.", null, Visibility.VISIBLE, null, null);
```

bundles/org.openhab.core.automation.module.media/src/main/java/org/openhab/core/automation/module/media/internal/MediaActionTypeProvider.java
```diff
@@ -62,43 +62,45 @@ public MediaActionTypeProvider(final @Reference AudioManager audioManager) {
...
     }
 
     private ModuleType getSayActionType(@Nullable Locale locale) {
         return new ActionType(SayActionHandler.TYPE_ID, getConfigSayDesc(locale), "say something",
-                "Speaks a given text through a natural voice.", null, Visibility.VISIBLE, new ArrayList<>(),
-                new ArrayList<>());
+                "Speaks a given text through a natural voice. Optionally set the volume.", null, Visibility.VISIBLE,
```
kaikreuzer, line 87:
```suggestion
                "Speaks a given text through a natural voice. Optionally sets the volume.", null, Visibility.VISIBLE,
```

kaikreuzer (approved):
lgtm, thanks!

frest-de:
I think the Parameter "volume" is not working as it should. Using it with say and googletts in ECMA rules, the volume gets resetted immediately, before the message has finished.

cweitkamp:
Yes, I agree. It is not working like a user expects. But I am afraid the framework is not able to handle it in a better way. The problem is we are living in an asynchronous world. The `VoiceManager` stores the current volume, sets the given new, volume, sends the media file or audio stream to be played to the `AudioSink` and restores the old volume afterwards. Most of the sink implementations send an asynchronous request towards the target device or API and return immediately afterwards. At this time the old volume will already be restored but the actual playback of the media is not finished or even started yet.

https://github.com/openhab/openhab-core/blob/3c301772695b80f5c27a23af987bb2cc440b1387/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java#L229-258

PR #1860, opened 2020-11-29
Remove semantic metadata on item tag removal

### LINKED ISSUE #1841: [mainUI] Changing semantic class from "Indoor" to "None" does not work
*@TBail commented on Oct 25, 2020, 4:16 PM UTC:*

To see some locations i trieb to setup a semantoc model og my home and tried to use the limited classes from the model in plemented. In the communisty i read that it is not possible to use the class location and get some location in the location pane. So i changed the clases i used and recognized that if i change the class to None this has some errors.

[image](https://user-images.githubusercontent.com/40339145/97112628-b1b9d580-16e5-11eb-9830-5df1e0793dca.png)

As you in the picture in the tree the zones are still indoor in the right None is selected.

Currently i am on build #1985

*This issue was moved by [ghys](https://github.com/ghys) from [openhab/openhab-webui#443](https://github.com/openhab/openhab-webui/issues/443).*

silamon:
Fixes #1841

### DIFF
--- a/bundles/org.openhab.core.semantics/src/main/java/org/openhab/core/semantics/internal/SemanticsMetadataProvider.java
+++ b/bundles/org.openhab.core.semantics/src/main/java/org/openhab/core/semantics/internal/SemanticsMetadataProvider.java
@@ -104,7 +104,6 @@ public class SemanticsMetadataProvider extends AbstractProvider<Metadata>
      *
      * @param item the item to update the metadata for
      */
-    @SuppressWarnings({ "null", "unused" })
     private void processItem(Item item) {
         MetadataKey key = new MetadataKey(NAMESPACE, item.getName());
         Map<String, Object> configuration = new HashMap<>();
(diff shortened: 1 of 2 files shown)


cweitkamp (commented):
Thank you. May I ask you to verify it by adding a unit test? Would be appreciated.

bundles/org.openhab.core.semantics/src/main/java/org/openhab/core/semantics/internal/SemanticsMetadataProvider.java
```diff
@@ -242,7 +246,6 @@ public void added(Item item) {
         processItem(item);
     }
 
-    @SuppressWarnings("null")
     @Override
     public void removed(Item item) {
         Metadata removedMd = semantics.remove(item.getName());
```
cweitkamp, line 251: This part of code can be replaced by `processItem(item)`.

cweitkamp (commented):
Thanks you for adding the test case. This will help.to stabilize the framework. One final comment.

bundles/org.openhab.core.semantics/src/main/java/org/openhab/core/semantics/internal/SemanticsMetadataProvider.java
silamon: That would mean that I need to modify the item by removing the tags - otherwise the metadata is not removed since the removed item possibly still has a tag. I prefer to keep this untouched. :)

cweitkamp: Ah, got it. Thanks for clarification.

cweitkamp (approved):
LGTM. Thanks.

PR #1864, opened 2020-11-29
fix null type in model.thing generic item channel link provider

maggu2810:
This fixes the following problem:

* Description: Null type mismatch (type annotations): required '@NonNull Collection<?>' but this expression has type '@Nullable Set<@NonNull String>'	
* Resource: GenericItemChannelLinkProvider.java
* Path: /org.openhab.core.model.thing/src/org/openhab/core/model/thing/internal
* Location: line 133
* Type: Java Problem


wborn (approved):
Thanks!

Issue #1867, opened 2020-11-29
Rule Refresher Error


hmerk:
After upgrading to M4, I can see the following errors in the log file and my textual rules are not loaded. This error still occurs when no .rules file exists.
```
2020-11-29 22:08:45.342 [ERROR] [org.openhab.core.model.rule         ] - bundle org.openhab.core.model.rule:3.0.0.M4 (172)[org.openhab.core.model.rule.jvmmodel.RulesRefresher(122)] :  Error during instantiation of the implementation object
java.lang.reflect.InvocationTargetException: null
	at jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method) ~[?:?]
	at jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62) ~[?:?]
	at jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45) ~[?:?]
	at java.lang.reflect.Constructor.newInstance(Constructor.java:490) ~[?:?]
	at org.apache.felix.scr.impl.inject.ComponentConstructor.newInstance(ComponentConstructor.java:309) ~[bundleFile:?]
	at org.apache.felix.scr.impl.manager.SingleComponentManager.createImplementationObject(SingleComponentManager.java:277) [bundleFile:?]
	at org.apache.felix.scr.impl.manager.SingleComponentManager.createComponent(SingleComponentManager.java:114) [bundleFile:?]
	at org.apache.felix.scr.impl.manager.SingleComponentManager.getService(SingleComponentManager.java:982) [bundleFile:?]
	at org.apache.felix.scr.impl.manager.SingleComponentManager.getServiceInternal(SingleComponentManager.java:955) [bundleFile:?]
	at org.apache.felix.scr.impl.manager.AbstractComponentManager.activateInternal(AbstractComponentManager.java:765) [bundleFile:?]
	at org.apache.felix.scr.impl.manager.DependencyManager$SingleStaticCustomizer.addedService(DependencyManager.java:1045) [bundleFile:?]
	at org.apache.felix.scr.impl.manager.DependencyManager$SingleStaticCustomizer.addedService(DependencyManager.java:999) [bundleFile:?]
	at org.apache.felix.scr.impl.manager.ServiceTracker$Tracked.customizerAdded(ServiceTracker.java:1216) [bundleFile:?]
[...]

PR #1869, opened 2020-11-29
[automation] Request automation thread pool as a scheduled pool

kaikreuzer:
This is necessary since the RulesRefresher uses the same pool and needs to be able to schedule jobs.

Fixes https://github.com/openhab/openhab-core/issues/1867


wborn (approved):
Thanks!

Issue #1868, opened 2020-11-29
Rules fail to run with multiple triggers


MikeJMajor:
If a rule with multiple triggers triggers again while it's already running, the second trigger seems to be discarded with an error such as:

`2020-11-29 20:43:06.865 [ERROR] [e.automation.internal.RuleEngineImpl] - Failed to execute rule ‘cdf8ab1ce1' with status 'RUNNING'`

Prior to M4 the rule would run for each triggering condition. Probably related to #1857

kaikreuzer:
Yes, it seems to be related to #1857, but actually the change now only provoked to bring this behavior to the surface, as it existed before already: The rule engine does not permit starting a rule that is already running. My assumption is that before the rules were triggered so quickly that the rule engine actually didn’t notice that they were running in parallel. Logging an error and not doing anything is probably the worst thing to do.

We have imho two options:
1. Simply remove the check whether a rule is already running or not. This will most likely cause some issues with finding the right moment to set it back to "IDLE", since we would have to count how many parallel executions are running. I am also not sure how many issues there might turn up with having the same rule modules being executed by multiple threads in paralllel - I am pretty sure that there will be issues with rule contexts and concurrent changes of variables etc.
2. Wait with the next execution until the current one has finished. This might slow down the rule execution a bit and make actions more sequential, but I'd in general think that it's the much safer option for now.

@openhab/core-maintainers You're opinion would be much appreciated as well here.

kaikreuzer:
Ok, looking deeper in the code, option 2 was pretty easy to implement and I further thinking about it, I strongly believe that it is also the right solution in the long run.

PR #1870, opened 2020-11-29
[automation] Synchronize the executions of each rule

kaikreuzer:
Synchonisation happened before for all rules, but only up to checking for the RUNNING status - which means that subsequent executions were already started to enter the codeblock, before the previous execution finished and set the status back to IDLE.
Now the executions are correctly chained, while different rules are executed in parallel.

Fixes https://github.com/openhab/openhab-core/issues/1868


wborn (approved):
Thanks for fixing it!

PR #1872, opened 2020-11-30
fix null constraint mismatch in model.thing for generic thing provider

maggu2810:
This fixes the following problem:

* Description: Null constraint mismatch: The type 'Thing' is not a valid substitute for the type parameter '@NonNull E extends Object'
* Resource: GenericThingProvider.java
* Path: /org.openhab.core.model.thing/xtend-gen/org/openhab/core/model/thing/internal
* Location: line 99
* Type: Java Problem


wborn (approved):
Thanks!

PR #1874, opened 2020-11-30
[autoupdatemanager] Get only relevant Links of an Item instead of all

cweitkamp:
- Get only relevant Links of an Item instead of all by querying the `AbstractLinkRegistry` which holds a `itemNameToLink`


kaikreuzer (approved):
Lgtm, should at least slightly improve performance.

Issue #1852, opened 2020-11-26
[OH3] 3.0.0M3 has substantial lag in processing and increased CPU


morph166955:
Upgraded from M2 to M3.  On M2, the java process sits around 20-30% CPU usage as per "top".  M3 sits around 115-145%.  I backed out from M3 back to M2 but left the addons as M3 (confirmed with bundle:list -s showing M2 and M3 packages).  No issues with M3 addons on M2 from a usage perspective.  Running on Ubuntu 20.04 in a (beefy) VM.

kaikreuzer:
I am on M3 and the java process is at 3% CPU - I don't think there is any general issue in the core.
Best if you analyse with thread dumps what keeps your system so busy.

morph166955:
Happy to. I've never done that before, can you provide a link so I can get started?

kaikreuzer:
Here's all you need to now :-)
https://www.baeldung.com/java-thread-dump

morph166955:
Looks fun. I'll drop M3 back on probably Sunday and see what I can capture. I am intrigued that M2 with M3 addons works just fine where M3 has this issue. I looked through the changes for M3 and I didn't see much in terms of things that could impact this.

morph166955:
Not exactly sure what I'm looking at, but starting to dig through it now.  Uploading the thread dumps for someone who knows this better than me to look at if they have some time.  I updated to M3 via dpkg and immediately noticed the load on the process staying over 100%.  I gave the system 10 minutes to "calm down" (it didn't).  I ran 3 separate thread dumps during that 10 mins which I'm going through now.  I stopped the system and put M2 back in via dpkg.  System is running just fine.  There were no config changes made at all.  Just start/stop of process and dpkg.  No cache cleaning, nada.

[M3-Slow10Min.txt](https://github.com/openhab/openhab-core/files/5611210/M3-Slow10Min.txt)
[M2-NoIssues.txt](https://github.com/openhab/openhab-core/files/5611211/M2-NoIssues.txt)

morph166955:
@kaikreuzer M4 seems to have the same issue for me.  M2 with M4 addons works just fine.  Also interesting that it seems to identify AbstractLink as what it's waiting on.  I attempted to do the dump at exactly the 10 minute mark as a comparison point.  It looks like OH-eventexecutor-1 ran for 7+ minutes out of the 10 which seems excessive.

@cweitkamp As you did #1794 I thought you may be interested in this.

[...]

PR #1880, opened 2020-12-02
Cache safe call proxies for profiles instead of continuously recreating them

kaikreuzer:
I have spent significant time profiling threads, especially the "OH-eventexecutor-1" thread.
The major insight I had was that this thread spent around 90% of its active time creating proxies for the safecaller.
This PR caches the created proxies and thus avoids them being continuously recreated.
It massively reduces the CPU load that is caused by this thread and hardly has any memory impact (on my production system, it ran smoothly for 5 hours and only cached 60 proxies altogether).

Here's the figures from a thread dump of my production system after > 5 hours with that change:
Screenshot 2020-12-02 at 18 05 40

During the 18877 seconds that the thread existed, it only consumes 32.5 seconds CPU time, which shows that it easily handles the load of event dispatching.

This should hopefully also fix https://github.com/openhab/openhab-core/issues/1852.

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/CommunicationManager.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/CommunicationManager.java
@@ -12,9 +12,11 @@
  */
 package org.openhab.core.thing.internal;
 
+import java.time.Duration;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArraySet;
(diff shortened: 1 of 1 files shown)


wborn (approved):
Looks like a very good improvement indeed! Did you also have a look if other usages of the `SafeCaller` might benefit from similar caching?

kaikreuzer:
I actually first tried to implement a cache in the safecaller itself, but that didn't really work out well.
The communication manager was definitely the most relevant place to address, so I covered it in this PR.
Besides this, the `ProfileCallbackImpl` might be another candidate, but this is only used for outgoing commands that are much less frequent than incoming events.

wborn:
OK great! From the looks of it the other usages indeed seemed less frequent. If one day we keep creating caches for `SafeCaller`s it might make sense to invest some time in making these caches more reusable.

PR #1884, opened 2020-12-03
[automation] Script dependency tracking

jpg0:
Added ability for ScriptEngines to allow script dependencies to be tracked and changes to trigger script reloading.

This is via a supplier function included in the Script Context.

I have tested this with (a) Python that does not use it and (b) GraalJS which does.

### DIFF
--- a/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/loader/DependencyTracker.java
+++ b/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/loader/DependencyTracker.java
@@ -1,0 +1,70 @@
+/**
+ * Copyright (c) 2010-2020 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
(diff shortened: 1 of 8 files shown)


jpg0:
@kaikreuzer this PR is the last one before the main Graal one. This adds the ability for dependency tracking of script libraries, so that modifying a library will reload all scripts that depend on it (I had plenty of issues without this as library changes would break things that I didn't see until a restart). I had included this functionality in the script provider addon itself, but though that it would be beneficial to push it down in the framework so that other languages can use it. If you don't agree please let me know and I'll move it back to the script addon (or even out of that). I noticed that you had made changes in this area recently.

kaikreuzer:
Thanks @jpg0! Since we are this week in a feature freeze for 3.0, I'd suggest to keep that for after the release and plan it for 3.1, since GraalVM won't be earlier than that either.
I'll be very busy with the release the next days, so please allow some time before receiving review feedback (at least from me, maybe the other @openhab/core-maintainers are faster ;-)).

jpg0:
@kaikreuzer Thanks for the update. I've actually made a small change so that the future 3.1 GraalVM support should be backwards compatible with 3.0, just without the dependency tracking from this PR, also so that it can be reviewed independently. Good luck with 3.0 GA!

kaikreuzer (approved):
Thanks for your patience, @jpg0!
Looks good to me, so let's merge!

PR #1885, opened 2020-12-03
Provide Script Names to Script Engines

boc-tothefuture:
Per the [spec](https://download.oracle.com/otn-pub/jcp/java_scripting-1.0-fr-eval-oth-JSpec/java_scripting-1_0-fr-spec.pdf?AuthParam=1607021147_3d19a0e3655d0b824f9f078215210522)

```
For instance, the value of the key
javax.script.filename
is defined to be the name of the file or resource containing the source
of the script. Setting this value makes the name of the source
available to the underlying scripting implementation for use in
constructing error messages.
```

I have tested this in JRuby and it works correctly.

With the ruby code:
`puts __FILE__`

Before the change:
`<script>`

After the change:
`/src/openhab-scripting/tmp/openhab/conf/automation/jsr223/ruby/personal/test_1607028003.rb`


boc-tothefuture:
FYI - The commit is signed off, the name just doesn't match the commit. OpenHAB requires the full name, where as the commit was not done with the full name.

5iver:
Duplicate of #1196. It looks like you also [refer a full path rather than a file name. Maybe with your input, this can finally be merged.

kaikreuzer:
@boc-tothefuture Is this PR still relevant? I see it has been marked as duplicate, although I am not sure if it isn't rather a further enhancement.
So please either close this PR or rebase it in order to resolve the merge conflict. Thanks!

boc-tothefuture:
I have rebased this.
@jpg0 - Are we Ok with this simple method of providing the filename to the script to assist in troubleshooting?

jpg0:
Sure, fine by me. Personally I would prefer the setting to happen in the `ScriptEngineManager` as I don't think that it's the responsibility of the watcher to be concerned with javax.script stuff (and shouldn't tie itself to it), but it's not a big deal and could be fixed later. Same for the ability to use a filename rather than a filepath.

kaikreuzer (approved):
Thanks guys!

Issue #1888, opened 2020-12-04
[oauth] scope in AccessTokenResponse


clinique:
Hello,
I'm reviewing the netatmo binding and would like to remove all external dependencies. Thus, using OAuthFactory.
This works nearly well, except that Netatmo API response is formed this way : 
````
{
"access_token":"5xxxxx95e984",
"refresh_token":"51cyyyyy841a489a4a346eb3b",
"scope":["read_station"],
"expires_in":10800,
"expire_in":10800
}
````
While in the API documentation, the scope should not be part of the answer, even worth, its a string array instead of the expected string.
So OAuthConnector fails to deserialize the AccessTokenResponse.

While on one hand, I can ask Netatmo (that I did) to respect better the RFC, I'm quite realistic regarding my chance to succeed.
Does anybody have an idea on how I could make this foot fit in the shoe ?

clinique:
@cweitkamp  : any indea ?

clinique:
@cweitkamp  : hello, just trying to integrate the merged PR in my binding, I get this error : 
````
18:48:26.758 [main] ERROR o.o.c.a.o.internal.OAuthConnector:80 - Unable to construct custom deserializer 'org.openhab.binding.netatmo.NetatmoAccessTokenDeserializer'
java.lang.ClassNotFoundException: org.openhab.binding.netatmo.NetatmoAccessTokenDeserializer cannot be found by org.openhab.core.auth.oauth2client_3.1.0.202012270303
	at org.eclipse.osgi.internal.loader.BundleLoader.findClassInternal(BundleLoader.java:511)
	at org.eclipse.osgi.internal.loader.BundleLoader.findClass(BundleLoader.java:422)
	at org.eclipse.osgi.internal.loader.BundleLoader.findClass(BundleLoader.java:414)
	at org.eclipse.osgi.internal.loader.ModuleClassLoader.loadClass(ModuleClassLoader.java:153)
````

I suppose I've got a visibility issue from my class toward the core but can't figure how to solve it. Would you have an idea ?

cweitkamp:
Where did you put the new class? Into an internal package? Or do you expose it?

clinique:
No, it's not an internal package, it's org.openhab.binding.netatmo

clinique:
@cweitkamp : I'm wondering wether I should not be using ````Bundle.loadClass()```` instead of ````Class.forName()````

cweitkamp:
[...]

PR #1891, opened 2020-12-05
Adressing issue #1888

clinique:
Would this be acceptable as a solution to my problem faced with Netatmo presented in #1888 ?
Signed-off-by: clinique <gael@lhopital.org>


clinique:
@kaikreuzer or @cweitkamp : can somebody have a look at this ? Sorry to be pushy, the rewritting of the Netatmo binding is pending this. Thanks.

cweitkamp:
Thank you for this draft. To be honest I am not feeling comfortable with it right now even if it looks good. I am not convinced to add workarounds in OHC for bad vendor implementations. But I agree with you that we need a solution for it and a similar issue may occur in the future for other external APIs.

What about moving the code for the `AccessTokenDeserializer` into the add-on bundle and just open the OHC OAuth implementation to allow a bindings developer pass a customized deserializer if needed?

clinique:
@cweitkamp : I like your approach and will think on it.

clinique:
@cweitkamp : can you give me your feelings regarding this approach ?

clinique:
@cweitkamp : had a chance to take a look ?

cweitkamp:
@clinique Thanks for the reminder. I am afraid I was not able to look into it yet. It looks better than before but I have to admit that I am not feeling comfortable with this solution either. Sry to be nit-picking here.

Do you think we can design it like a builder pattern for the `OAuthClientService` without touching the factory itself? Similar to classes in Jetty library? Imo the method to create services are already overloaded by parameters. E.g. a developer will be able to do something like this:

```
OAuthClientService service = OAuthFactory.createOAuthClientService(String, String, String, String, String, String, Boolean).withDeserializer(Class<T>);
```

This might need a little bit more refactoring but will be a lot more flexible to add more features in the future.

bundles/org.openhab.core.auth.oauth2client/src/main/java/org/openhab/core/auth/oauth2client/internal/OAuthClientServiceImpl.java
```diff
@@ -150,7 +150,7 @@ public String getAuthorizationUrl(@Nullable String redirectURI, @Nullable String
             throw new OAuthException("Missing client ID");
         }
 
-        OAuthConnector connector = new OAuthConnector(httpClientFactory);
+        OAuthConnector connector = new OAuthConnector(httpClientFactory, persistedParams.deserializerClassName);
```
cweitkamp, line 153: A minor reminder for later refactoring: This looks like a potential candidate to be used as singleton.

clinique:
> Do you think we can design it like a builder pattern for the `OAuthClientService` without touching the factory itself? Similar to classes in Jetty library? Imo the method to create services are already overloaded by parameters. E.g. a developer will be able to do something like this:
> 
> ```
> OAuthClientService service = OAuthFactory.createOAuthClientService(String, String, String, String, String, String, Boolean).withDeserializer(Class<T>);
> ```
> 
> This might need a little bit more refactoring but will be a lot more flexible to add more features in the future.

Hello @cweitkamp , thanks for the suggestion but it is unclear to me how I could reach this with only a single method ````withDeserializer(Class <T>)````, I think I would need a ````build()```` also and this would be a breaking change.

Can you give me some insights on how you imagine this ?

clinique:
@cweitkamp : thinking of it twice, I think I found the solution. Can you tell me what you think of it ? Thanks

bundles/org.openhab.core.auth.oauth2client/src/main/java/org/openhab/core/auth/oauth2client/internal/OAuthClientServiceImpl.java
```diff
@@ -395,4 +397,17 @@ public boolean removeAccessTokenRefreshListener(AccessTokenRefreshListener liste
     private String createNewState() {
         return UUID.randomUUID().toString();
     }
+
+    @Override
+    public <T extends JsonDeserializer<?>> OAuthClientService withDeserializer(Class<T> deserializerClass) {
+        OAuthClientServiceImpl clientService = new OAuthClientServiceImpl(handle, persistedParams.tokenExpiresInSeconds,
```
cweitkamp, line 403: Yes, this looks nice. Do we have to create a new instance here or can we use `this`?

clinique: I did opt to create a new instance to avoid any side effect regarding other parameter changed, it was the most straitforward path

bundles/org.openhab.core.auth.oauth2client/src/main/java/org/openhab/core/auth/oauth2client/internal/PersistedParams.java
```diff
@@ -32,6 +32,7 @@
     String state;
     String redirectUri;
     int tokenExpiresInSeconds = 60;
+    String deserializerClassName;
```
cweitkamp, line 35:
```suggestion
    @Nullable String deserializerClassName;
```

bundles/org.openhab.core.auth.oauth2client/src/main/java/org/openhab/core/auth/oauth2client/internal/OAuthClientServiceImpl.java
cweitkamp: Understood. Sounds reasonable.

cweitkamp (commented):
Thanks for your patience. One final comment. After that I am fine with merging this nice and useful improvement.

bundles/org.openhab.core.auth.oauth2client/src/main/java/org/openhab/core/auth/oauth2client/internal/OAuthClientServiceImpl.java
clinique: > 
> 
> A minor reminder for later refactoring: This looks like a potential candidate to be used as singleton.

Sorry, I'm afraid I did not get your point here

cweitkamp (approved):
LGTM. Thanks.

Please run 'mvn spotless:apply'. Jenkins is not happy with the formatting.

cweitkamp (approved):

Issue #1435, opened 2020-04-19
Move org.openhab.io.transport.modbus to openhab-core


kaikreuzer:
See https://github.com/openhab/openhab-addons/pull/7424#discussion_r410993403.

All io.transport bundles reside in openhab-core with the exception of the modbus one.
When moving the active development of add-ons from the 2.5.x branch to master, we should align this and move this bundle to the correct repo.

ssalonen:
We (@mrbig ) were thinking some (backwards incompatible) changes to transport bundle api to simplify it 

What would be good moment to do these kind of changes? Before or after move to core?

kaikreuzer:
If the changes only break the api of the transport bundle, but not the functionality of the binding(s), it should be fine to do it on 2.5.x already.

PR #1892, opened 2020-12-05
Moved modbus transport from addons to core

kaikreuzer:
addresses https://github.com/openhab/openhab-core/issues/1435

Maintainers, please note: I only moved the code and didn't touch it apart from the namespace change - so please refrain from review comments that concern the code itself.


bundles/org.openhab.core.io.transport.modbus/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+          <bnd><![CDATA[Bundle-SymbolicName: ${project.artifactId}
+Automatic-Module-Name: ${def;bsn}
+Import-Package: \\
+  org.eclipse.jdt.annotation.*;resolution:=optional,\\
+  org.openhab.*;version=!,\\
+  !net.sf.cglib.proxy,\\
+  gnu.io;version="[3.12,6)",\\
+  *
+-sources: false
+-contract: *]]></bnd>
```
wborn, line 59: Would it be possible to move this to a `bnd.bnd` file? That's where we manage these customizations for the other bundles in openhab-core.

wborn (commented):
It looks good to me and indeed the code is unchanged except for the packages/imports etc. :+1:  I have one remark below:

bundles/org.openhab.core.io.transport.modbus/pom.xml
kaikreuzer: If you can help me figuring out, what content should be put in that bnd file...? Exactly what is here in the `<bnd>` element?

kaikreuzer: Seems to work - done.

wborn: Does it also work with one trailing `\`? That's what is used in all the other [bnd.bnd](https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.model.item/bnd.bnd) files. They also all use only one space for indentation.

kaikreuzer: sure, done.

wborn (approved):

PR #1901, opened 2020-12-07
Remove SmartHome leftovers

wborn:
The only "SmartHome" references that still remain are in:

* [EventLogger](https://github.com/openhab/openhab-core/blob/13c2d7bf2be051fe8c72046d637770a92f517781/bundles/org.openhab.core.io.monitor/src/main/java/org/openhab/core/io/monitor/internal/EventLogger.java#L56), could also be changed but it may need some logger configuration updates
* [MDNSAnnouncer](https://github.com/openhab/openhab-core/blob/13c2d7bf2be051fe8c72046d637770a92f517781/bundles/org.openhab.core.io.rest.mdns/src/main/java/org/openhab/core/io/rest/mdns/internal/MDNSAnnouncer.java#L65), changing this would break discoverability with iOS/Android Apps and the Remote openHAB binding.

### DIFF
--- a/bundles/org.openhab.core.auth.jaas/src/main/java/org/openhab/core/auth/jaas/internal/JaasAuthenticationProvider.java
+++ b/bundles/org.openhab.core.auth.jaas/src/main/java/org/openhab/core/auth/jaas/internal/JaasAuthenticationProvider.java
@@ -43,8 +43,7 @@ import org.osgi.service.component.annotations.Modified;
 /**
  * Implementation of authentication provider which is backed by JAAS realm.
  *
- * Real authentication logic is embedded in login modules implemented by 3rd party, this code is just for bridging it to
- * smarthome platform.
+ * Real authentication logic is embedded in login modules implemented by 3rd party, this code is just for bridging it.
  *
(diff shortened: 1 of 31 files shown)


cweitkamp:
If we want to get rid of those remaining occurrences we should do it now - before freezing the code. The logger configuration (at least in openhab-distro) shouldn't be too hard to adopt, or did I miss something? Coordinating the changes in MDNS will be harder.

kaikreuzer:
> Coordinating the changes in MDNS will be harder.

That's perfectly easy and safe to change, since we already overwrite it for openHAB, see https://github.com/openhab/openhab-distro/blob/master/distributions/openhab/src/main/resources/userdata/etc/custom.system.properties#L3

kaikreuzer:
When doing this nice cleaning, I have found some more places: https://github.com/openhab/openhab-core/search?q=esh

wborn:
Thanks! I had missed those but do remember running into them some time ago. :wink:

kaikreuzer:
Looks as if ESH was another good keyword to watch out for! Thanks for your efforts 👍

kaikreuzer (approved):
Many thanks!

PR #1907, opened 2020-12-09
HexUtils: fix empty string handling

### LINKED ISSUE #1903: HexUtils: empty string should result into byte array of length zero
IMHO `HexUtils.hexToBytes("")` should result into an empty byte array.
Instead an IAE is thrown.

I also think that hexToBytes should be the inverse of hexToBytes and calling the one function with the result of the other should return the argument given to the first one.

`HexUtils.bytesToHex(new byte[0]);` generates an empty string.

`HexUtils.hexToBytes(HexUtils.bytesToHex(new byte[0]));` results not into an empty byte array but into an IAE.

maggu2810:
Fixes #1903


kaikreuzer (approved):
Thanks!

PR #1909, opened 2020-12-10
Improve ChannelLinkNotifier.removed and AbstractUID.getAsString performance

### LINKED ISSUE #1908: Saving items file causes excessive CPU utilization
I described a problem in the community forum here.
https://community.openhab.org/t/openhab-3-0-milestone-5-discussion/109612/53

I'm opening an issue now, as I've been able to reproduce the problem on a fresh install of OH3 build 2050 using nothing other than one items file containing 405 item definitions.

I see sustained high cpu and memory utilization whenever I save changes to an items file. It’s especially severe (high CPU and memory for several minutes) when the items file contains a large number of items (e.g. > 400).

The problem manifests itself a couple different ways.

- If I try to shutdown openHAB immediately after I save an .items file, it takes 1.5 minutes for OH to shutdown. And, it only shuts down because the systemctl timeout causes systemctl to kill the openHAB process (the timeout is currently 90 seconds).

- After saving changes to one items file, if I immediately save changes to another items file, the normal “Loading model…” log message doesn’t show up in the log file right away. It shows up a few minutes later after OH finishes whatever it’s doing with the first file.

- Looking at cpu utilization in top or pidstat, I see elevated cpu utilization for anywhere from 30 seconds to several minutes, depending on the number of items in the file.

Here’s an example from VisualVM. You can see very stable/normal cpu and memory utilization. The red arrows represent when I saved the items file, and you can see where the process of loading the items is consuming 1 of the 4 cpus in my box for about 3 minutes.

items-cpu

items-memory

wborn:
Fixes #1908

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ChannelLinkNotifier.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ChannelLinkNotifier.java
@@ -77,10 +77,7 @@ public class ChannelLinkNotifier implements RegistryChangeListener<ItemChannelLi
         ChannelUID channelUID = element.getLinkedUID();
         ThingUID thingUID = channelUID.getThingUID();
 
-        boolean channelLinked = itemChannelLinkRegistry.stream()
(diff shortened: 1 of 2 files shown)


kaikreuzer (approved):
Excellent, many thanks for the analysis and the fix!

PR #1913, opened 2020-12-10
Auth pages i18n

ghys:
This implements localized messages for the authorize, change
password and create API token pages using a resource bundle.

Messages in English & French are included.

### DIFF
--- a/bundles/org.openhab.core.io.http.auth/.classpath
+++ b/bundles/org.openhab.core.io.http.auth/.classpath
@@ -6,6 +6,7 @@
 			<attribute name="maven.pomderived" value="true"/>
 		</attributes>
 	</classpathentry>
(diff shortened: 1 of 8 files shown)


ghys:
@kaikreuzer after this is approved & merged, thanks in advance to add these properties files in Crowdin as we discussed :)

ghys:
Would appreciate an expedited review of this so it can be inclusion in the release :)
Given that the UI was translated into 15 languages in less 72 hours by the community, it could well happen for these too, and it would add to a good experience.

image

And if the properties files are configured it's relatively harmless to let the translation campaign happen next week for these IMO. If there are e.g. file encoding issues this can be adjusted before merging.

kaikreuzer (approved):
Thanks!

kaikreuzer:
@ghys I have included these files in the crowdin config as discussed!

Issue #1829, opened 2020-11-16
DSL rules are deleted after reload of .things file


robnielsen:
I am using .things, .items, and .rules files that I copied over from OH2 to OH 3.0.0-SNAPSHOT, Build #2007. If the .things file is modified, any rule associated with things from the modified file are deleted. In my case the number of rules went from 22 to 2.

Here are the steps to reproduce:

1. Define things in a .things file
2. Add rules in .rules file for things in above .things file
3. Look in UI to make sure rules associated with things are displayed
4. Modify the .things file (I used the touch command)
5. Look in the UI and see the rules are deleted and don't show up again

lolodomo:
Can you explain what you mean by "rules associated with things". Are you talking about arule having as trigger a channel trigger for example ?
Can you provide an example of rule.
Are all rules based on items kept ?

robnielsen:
All rules associated with the items for thing things in the .things file. 

The UI looks like this with 3.0.0-SNAPSHOT build 2029:

![image](https://user-images.githubusercontent.com/838221/100517195-a3c8fb80-314e-11eb-9d75-c4d450b3df56.png)

Then I touch my insteon.things file and it looks like this:

![image](https://user-images.githubusercontent.com/838221/100517252-e559a680-314e-11eb-8c7f-6ef2eb9d9748.png)

One of the rules:
```
rule "Close Main Garage Door"
    when
        Item garageDoor1Close received command ON or
        Item garageDoor1ga received command DOWN
    then
        if (garageDoor1.state == OPEN) {
            sendCommand(garageDoor1Switch, ON)
        }
end
```

Associated items referenced in the rule:
```
Switch garageDoor1Close
Switch garageDoor1Open  
Switch garageDoor1Switch    {channel="insteon:device:home:316F4B:switch", autoupdate="false"}
Contact garageDoor1         {channel="insteon:device:home:316F4B:contact"}
Rollershutter garageDoor1ga {ga="Garage", synonyms="both garage"}
```

And the associated thing from the insteon.things file:
```
Thing device 316F4B "Main Garage Door" [address="31.6F.4B", productKey="0x00001A"]
```

robnielsen:
When I shut down, 8 of the missing rules reappear in the UI:

![image](https://user-images.githubusercontent.com/838221/100542800-02a37900-3212-11eb-96c6-9154343be8be.png)

robnielsen:
[...]

Issue #1823, opened 2020-11-14
[OH3] ScriptEngine for language 'application/vnd.openhab.dsl.rule' could not be found for identifier


morph166955:
Every couple of restarts I get the following error show up.  I get one error for every rule in every rules file (so yes I get about 150 of these).  Example below is of my roku rules file.  I have 5 rules defined, and 5 errors.

```
2020-11-14 20:28:23.789 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model 'roku.rules'
2020-11-14 20:28:23.862 [ERROR] [ipt.internal.ScriptEngineManagerImpl] - ScriptEngine for language 'application/vnd.openhab.dsl.rule' could not be found for identifier: 62f8e117-5217-4e30-ad47-754360b75caa
2020-11-14 20:28:24.105 [ERROR] [ipt.internal.ScriptEngineManagerImpl] - ScriptEngine for language 'application/vnd.openhab.dsl.rule' could not be found for identifier: 3b53ba8b-09bf-4898-a97f-0b08b726f348
2020-11-14 20:28:24.281 [ERROR] [ipt.internal.ScriptEngineManagerImpl] - ScriptEngine for language 'application/vnd.openhab.dsl.rule' could not be found for identifier: 34634ee4-001d-42e0-9a4f-f000d45318e6
2020-11-14 20:28:24.566 [ERROR] [ipt.internal.ScriptEngineManagerImpl] - ScriptEngine for language 'application/vnd.openhab.dsl.rule' could not be found for identifier: d1b25dd8-649b-4154-83c5-c0755964d745
2020-11-14 20:28:24.862 [ERROR] [ipt.internal.ScriptEngineManagerImpl] - ScriptEngine for language 'application/vnd.openhab.dsl.rule' could not be found for identifier: 62f8e117-5217-4e30-ad47-754360b75caa

```

lolodomo:
As @kaikreuzer made few changes in DSL rule management, could you confirm you still have this problem with a recent OH3 version ?
What vcersion were you using ?
Do you have this issue if you use only one rule file (my case and no such error) ? Just to understand if it occurs when having more than one rule file.

grzegorz914:
OH 3.0-2030

the problem still exist:

````
2020-11-28 13:06:37.396 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model 'rolety.rules'
[...]

Issue #1808, opened 2020-11-11
[OH3] DSL rules can appear in double


lolodomo:
In OH3, it can occur that my DSL rules appear all in double.
Suppose that my file is named maison.rules. In this case I see one rule for `maison-xx` and one rule for `tmp_maison-xx`.
The trigger for this issue could be when at OH startup a new bionding is installed. But I am not absolutely sure as I only got this issue 2 or 3 times.

If I restart OH, the rules are no more in double.

thewiep:
I also noticed this once yesterday after I upgraded from M2 to latest snapshot.
After restarting openHAB the tmp_ versions are no longer present.
It didn't seem to happen on M2 though.

lolodomo:
I am not able to find a way to reproduce. And you? Any idea what could be the trigger? 
I only notice that bug 2 or 3 times, that is not a lot considering how many times I started OH3.

seimenb:
I had the same problem today. 
On of my rules was doubled with an tmp_ prefix.
Restarting openHab has not resolved it for me. The doubled rule still showed up after restarting.
Finaly I stopped openHab, copyied the doubling rule file away and startet openHab without the rule file.
Then I adde the rule file again to the rules folder, while openHab was running. 
Since then the rule shows up only ones, as expected. 
And I could start my openHab several times witout doubling the rule again.

Issue #1777, opened 2020-10-31
[OH3] DSL rule failing sometimes at startup


lolodomo:
OH3 snapshot 1995

Here ius a part of my DSL file:
```
import java.util.regex.Pattern

var Pattern patternVoice

rule "Démarrage"
when
	System started
then
	logInfo("Rule","Démarrage début")
	patternVoice = Pattern::compile("(donner|passer en hors gel|passer en mode confort|passer en mode réduit) (le |la |les |l'|du )(.*) (du |de la |de l'|dans le |dans la |dans l')(.*)")
	logInfo("Rule","Démarrage fin")
end
```
And here is what happens at startup:
```
11:53:52.167 [INFO ] [org.openhab.core.model.script.Rule   ] - Démarrage début
11:53:52.177 [ERROR] [.internal.handler.ScriptActionHandler] - Script execution of rule with UID 'maison3-7' failed: An error occurred during the script execution: Cannot assign a value in null context. in maison3
```

**It does not occur at each restart, only sometimes.**

If I run later the rule from Main UI, it works as expected.
So there is probably something started too early at startup ?

wborn:
It doesn't seem to be something new according to these [community search results](https://community.openhab.org/search?q=Cannot%20assign%20a%20value%20in%20null%20context).

lolodomo:
Maybe this is new for the new rule engine ?
At least, now an issue is opened :) and I hope it can be solved ;)

lolodomo:
"Cannot assign a value in null context" : isn't it something that should help what's wrong here ?

Is it this method that is called with a null parameter ?
https://github.com/openhab/openhab-core/blob/584c85a07f760216b2c42f482a7800bfe064b0d3/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/ScriptActionHandler.java#L56
Could it be linked to the fact that my rule trigger is "System started" ?
@wborn @kaikreuzer : Maybe the rule is called too early when this "context" is still not yet set ?

lolodomo:
If the problem is linked to trigger "System started", could the framework waits for this "context" to be ready (not null) before triggering the rule ?

lolodomo:
https://github.com/openhab/openhab-core/blob/0d1a15ef3470bcf350ed88679447e0f7e7adeafe/bundles/org.openhab.core.model.rule/src/org/openhab/core/model/rule/jvmmodel/RulesRefresher.java#L177
[...]

PR #1914, opened 2020-12-11
Implemented start level service

kaikreuzer:
This finally implements https://github.com/eclipse-archived/smarthome/issues/1896, fixes a couple of issues that currently exist due to the unpredictable startup sequence and it should also improve the startup performance, since activities do not have to be done multiple times in order to deal with lazy loaded stuff.

This service combines different ReadyMarkers into a new start level ready marker and thus lets other services depend on those, without having to know about the single markers.
This brings an important decoupling, since the set of markers for a certain start level might depend on the individual set up-
The start level service is therefore configurable (see [here](https://github.com/openhab/openhab-distro/pull/1211/files#diff-6f5d3ba8289d20be508f0997f8b24e4b48962d8f5e56df3e793a566bec2469f1R16-R21), so that users have a chance to adapt the conditions upon a certain start level is reached.

Start levels are defined as values between 0 and 100. They carry the following semantics:

- 00 - OSGi framework has been started.
- 10 - OSGi application start level has been reached, i.e. bundles are activated.
- 20 - Model entities (items, things, links, persist config) have been loaded, both from db as well as files.
- 30 - Item states have been restored from persistence service, where applicable.
- 40 - Rules are loaded and parsed, both from db as well as dsl and script files.
- 50 - Rule engine has executed all "system started" rules and is active.
- 70 - User interface is up and running.
- 80 - All things have been initialized.
- 100 - Startup is fully complete.

Note: Levels 70 and 80 are not yet implemented in this PR, but should be added at a later stage.

While still being in a level <50, no events will be logged out anymore, which reduces the logs wrt "item added" and "link added" events tremendously.

I have also further improvements in mind like automatically registering what bundles are installed and for which markers needs to be waited, so that there is less necessity to adapt the configuration, if anybody does not use a "standard" setup of openHAB. Note that for anybody using the normal distro and not removing any essential parts of it, there should be no need to configure anything with the current implementation.

Fixes https://github.com/openhab/openhab-core/issues/1637
Fixes https://github.com/openhab/openhab-core/issues/1777
Fixes https://github.com/openhab/openhab-core/issues/1734
Fixes https://github.com/openhab/openhab-core/issues/1823
Fixes https://github.com/openhab/openhab-core/issues/1808

Might also fix https://github.com/openhab/openhab-core/issues/1365
Might also fix https://github.com/openhab/openhab-core/issues/1829


bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleEngineImpl.java
```diff
@@ -900,13 +895,13 @@ private void setStatus(String ruleUID, RuleStatusInfo newStatusInfo) {
...
         if (f == null || f.isDone()) {
-            scheduleTasks.put(rUID, getScheduledExecutor().schedule(() -> {
+            scheduleTasks.put(rUID, getScheduledExecutor().submit(() -> {
                 final WrappedRule managedRule = getManagedRule(rUID);
                 if (managedRule == null) {
                     return;
                 }
                 setRule(managedRule);
-            }, scheduleReinitializationDelay, TimeUnit.MILLISECONDS));
+            }, TimeUnit.MILLISECONDS));
```
wborn, line 904: This seems like a `schedule` parameter left over:

```suggestion
            }));
```

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ItemRegistryImpl.java
```diff
@@ -280,6 +281,17 @@ protected void unsetEventPublisher(EventPublisher eventPublisher) {
...
+    @Override
+    @Reference
+    protected void setReadyService(ReadyService readyService) {
+        super.setReadyService(readyService);
+    }
+
+    @Override
+    protected void unsetReadyService(ReadyService readyService) {
+        super.unsetReadyService(readyService);
+    }
```
wborn, line 293: Did you also look into using constructor injection for injecting the `ReadyService` into the registries?

wborn (commented):
Thanks for your awesome work on this! I'm very happy to see this being addressed just in time for OH3. :smiley:
I only have a few comments and think we should then proceed with merging this so we can start testing it and it makes it into the RC.
Can you also have a look at the failing tests?

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ItemRegistryImpl.java
kaikreuzer: No, I didn't want to mess with that in that PR as it can easily be a bigger refactoring. That's imho rather something for after the release.

kaikreuzer:
> Can you also have a look at the failing tests?

Yep, I'm on it.

lolodomo:
Might become one of the major improvements of OH2 over OH3. I like.

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ItemRegistryImpl.java
wborn: I can imagine that, but if you had already given it a try and it didn't work out well... then your insights on this might help if someone else looks into it one day. :slightly_smiling_face:

kaikreuzer:
@wborn Tests succeeded!

wborn (approved):
Thanks a lot! :+1:  Let's merge it so we can start testing it.

kaikreuzer:
Let's do testing (and fixing 🤔) until tomorrow evening...

ghys:
@kaikreuzer could you confirm the "system started" rules still have to run at start level 20? I just implemented some UI helper to abstract this when adding these sorts of triggers yesterday.

kaikreuzer:
@ghys Anything below 50 will do.
I was also thinking about adding an additional "rule loaded" trigger, to bring back the 2.x functionality that a rule is executed once whenever being modified. This is not part of this PR, though, but might actually then be a common choice for users. I wouldn't put a "system started" trigger into the UI, I consider this rather deprecated. The "start level" trigger is more flexible and will allow users in future to also choose start level 80, e.g. if they want to wait for their things to be initialized.

ghys:
Got it. So maybe offer these as options in a list? (or those above 50)

> 00 - OSGi framework has been started.
10 - OSGi application start level has been reached, i.e. bundles are activated.
20 - Model entities (items, things, links, persist config) have been loaded, both from db as well as files.
30 - Item states have been restored from persistence service, where applicable.
40 - Rules are loaded and parsed, both from db as well as dsl and script files.
50 - Rule engine has executed all "system started" rules and is active.
70 - User interface is up and running.
80 - All things have been initialized.
100 - Startup is fully complete.

(this could also go to the configParameterDescription of the SystemStartlevelTrigger)

kaikreuzer:
Options above 40 make sense, yes. And indeed, they should be provided as configParameterDescription, I didn't do this yet.

lolodomo:
After first startup of snapshot 2066, I noticed in UI that no rules are loaded. I also noticed that I have no item state changes logged.

kaikreuzer:
Same for me. On the second startup, everything went well, though.

lolodomo:
> Same for me. On the second startup, everything went well, though.

Yes, second startuplooks good, item/thing logging is back, I see no rule errors in log, my "system started" rule is run and without error, all rules are listed in UI.

kaikreuzer:
Sounds great - first sign that I didn't completely destroy the runtime ;-)

For me all further starts (2-4 by now) went smooth. I guess the first start is our usual problem that the system runs amok directly after doing an update.

lolodomo:
After few other restarts, I finally got errors of tyis kind:
```
17:48:33.113 [ERROR] [.internal.handler.ScriptActionHandler] - Script execution of rule with UID 'maison3-10' failed: The name 'NULL' cannot be resolved to an item or type; line 240, column 23, length 4 in maison3
17:48:33.114 [ERROR] [.internal.handler.ScriptActionHandler] - Script execution of rule with UID 'maison3-7' failed: The name 'NULL' cannot be resolved to an item or type; line 173, column 23, length 4 in maison3
```
This is failing on lines of this kind:
```
if (previousState == NULL
```
while the rule trigger is "item xxxx changed".

lolodomo:
Another example, this time with ON instead of NULL:
```
17:59:39.019 [ERROR] [.internal.handler.ScriptActionHandler] - Script execution of rule with UID 'maison3-3' failed: The name 'ON' cannot be resolved to an item or type; line 96, column 28, length 2 in maison3
```
Line 96 is:
```
	else if (previousState == ON && newState == OFF) {
```
It looks like errors are related to `previousState `.

lolodomo:
Errors are certainly occuring at first startup of the rule, while previous state is NULL.

lolodomo:
I even got this one:
```
18:00:02.577 [ERROR] [.internal.handler.ScriptActionHandler] - Script execution of rule with UID 'maison3-2' failed: The name 'previousState' cannot be resolved to an item or type; line 34, column 6, length 13 in maison3
```
One time again, it is related to previousState for a rule with a trigger "item xxx changed".


Regarding the run of the rule with "system started", I have always seen this rule run correctly. This case is now working perfectly.

kaikreuzer:
@lolodomo Are those problems with the `previousState` only temporary during startup, i.e. does the same rule start to work a bit later? Or is it broken until you restart the system? Feel free to create a new bug report for it, ideally some more logs.

lolodomo:
I will check again but I guess it is only temporary during startup.

kaikreuzer:
This would clearly reduce the severity of the issue!

lolodomo:
@kaikreuzer : as you provided several fixes since my tests yesterday, I will try again when a new snapshot is available and create an issue if I still see problems.

lolodomo:
@kaikreuzer : just another positive feedback. With snapshot 2085, I was able to restart openHAB server 5 times without any errors triggered by rules. Super. No need for me to open another issue ;)

kaikreuzer:
Excellent 🎉

maggu2810:
Hi @kaikreuzer, just a small question without digging deeper into the code changes myself:
Does this change add some dependencies between some modules?
Just an example that you also state above yourself (logging will enabled only at a certain time):
The EventLogger registers a tracker for the rules startlevel and logging will be enabled after that level has been reached.
If the rule engine is not part of the distribution anymore and this marker is not added to the ready service, the logging will not be activated at all. Correct? If that is correct I assume that the rule engine is a mandatory part of the distribution.
Are there other markers and trackers that have a strong dependency?

kaikreuzer:
@maggu2810 Yes, reaching certain start levels requires certain components to be available in the distro.
The markers that are expected for a certain level for the standard openHAB distro are [defined here](https://github.com/openhab/openhab-distro/blob/master/distributions/openhab/src/main/resources/runtime/services.cfg#L17-L22). If you want to build a custom distro that is lacking certain features, you simply have to remove entries from that list and that's it.
´

PR #1487, opened 2020-05-20
Use constructor injection and update null annotations

wborn:
There's also https://github.com/openhab/openhab-webui/pull/258 to update the Web UIs for these changes.

PR #1917, opened 2020-12-12
Register CommunicationManager to listen on ItemChannelLink changes

### LINKED ISSUE #1645: [profiles] configuration change not activated
Changes to profile configuration via UI are stored (checked that the new configuration is available in the REST API), but the used profile itself is not updated. Because of that the old configuration is used until the bundle providing the profile is restarted.

cweitkamp:
- Register `CommunicationManager` to listen on `ItemChannelLink` changes

We missed it in #1487.

Fixes #1645


kaikreuzer (approved):
Thanks!

PR #1920, opened 2020-12-12
increase the rule refresher delay

kaikreuzer:
Despite the better startup order, the rule refresher still kicks in several times during startup.
This is caused by thing actions being added at a later point (namely after thing handlers are instantiated). Since we do the refresh once all items and things are loaded and thus defined, we can imho easily reduce the frequency of any subsequent refreshes. Allowing 30 seconds should ensure that most thing handlers are instantiated and we thus only have a single further refresh triggered. Note: Once start level 80 is implemented, we should probably use markers to also exactly time this second required refresh.

### DIFF
--- a/bundles/org.openhab.core.model.rule/src/org/openhab/core/model/rule/jvmmodel/RulesRefresher.java
+++ b/bundles/org.openhab.core.model.rule/src/org/openhab/core/model/rule/jvmmodel/RulesRefresher.java
@@ -56,7 +56,7 @@ import org.slf4j.LoggerFactory;
 public class RulesRefresher implements ReadyTracker {
 
     // delay in seconds before rule resources are refreshed after items or services have changed
-    private static final long REFRESH_DELAY = 5;
+    private static final long REFRESH_DELAY = 30;
 
     public static final String RULES_REFRESH_MARKER_TYPE = "rules";
     public static final String RULES_REFRESH = "refresh";


bundles/org.openhab.core.model.rule/src/org/openhab/core/model/rule/jvmmodel/RulesRefresher.java
```diff
@@ -56,7 +56,7 @@
 public class RulesRefresher implements ReadyTracker {
 
     // delay in seconds before rule resources are refreshed after items or services have changed
-    private static final long REFRESH_DELAY = 5;
+    private static final long REFRESH_DELAY = 30;
```
wborn, line 59: Can it be changed such that it only uses this during startup? Once you start making changes to your instance I think it will be very annoying having to wait 30 seconds each time.

kaikreuzer: I was thinking about that as well, but I then really wondered what exactly would be the (annoying) use cases that would require a faster refresh. Note that the rule refresher only kicks in upon changes in items/things/addons to make sure that references in rules are correctly resolved.
If people work on rules, those changes are immediately reloaded upon saving the rule, so there is no delay.
If they add new items&things (after startup), it is likely that they do not reference those within the rules yet and will rather go and edit the rules later on (for which there won't be a delay).
So only if they first created rules with references to stuff that isn't yet defined/installed on their system, the 30 sec delay kicks in and I'd think this should be acceptable, especially when helping on the general startup behavior.

wborn: Thanks for the explanation! I agree that if it works the way you described it shouldn't be too annoying. Let's then merge and test it.

wborn (approved):

PR #1939, opened 2020-12-15
[persistence] Use 'getStateAs()' instead of type casting

cweitkamp:
- Use `getStateAs()` or `as()` instead of type casting

As we now allow the persistence services to return `QuantityType` values we should be a little bit more careful with type casting and rather use framework provided methods.

See https://community.openhab.org/t/openhab3-rule-with-quantitytype-items/110284


kaikreuzer (approved):
lgtm, thanks!

Issue #1745, opened 2020-10-21
[automation] Binding actions cannot be configured by UIs


cweitkamp:
This is a follow-up issue of https://github.com/eclipse-archived/smarthome/issues/6602 as no solution has been implemented yet and we need to track it for OH3 release.

While implementing and testing my port of the Pushover add-on (https://github.com/openhab/openhab-addons/pull/8586) I again stumbled upon this missing feature. Currently it is not possible to configure inputs for Binding actions based on `ThingActions`. The annotated inputs are not visible in UIs.

![image](https://user-images.githubusercontent.com/5131747/96733029-499f8280-13b9-11eb-89c9-ec4342cd8a54.png)

```java
    @RuleAction(label = "@text/sendHTMLMessageActionLabel", description = "@text/sendHTMLMessageActionDescription")
    public @ActionOutput(name = "sent", label = "@text/sendMessageActionOutputLabel", description = "@text/sendMessageActionOutputDescription", type = "java.lang.Boolean") Boolean sendHtmlMessage(
            @ActionInput(name = "message", label = "@text/sendMessageActionInputMessageLabel", description = "@text/sendMessageActionInputMessageDescription", type = "java.lang.String", required = true) String message,
            @ActionInput(name = "title", label = "@text/sendMessageActionInputTitleLabel", description = "@text/sendMessageActionInputTitleDescription", type = "java.lang.String") @Nullable String title) {
        logger.trace("ThingAction 'sendHtmlMessage' called with value(s): message='{}', title='{}'", message, title);
        return send(getDefaultPushoverMessageBuilder(message).withHtmlFormatting(), title);
    }
```

Internal automation actions like `say` or `playSound` define next to the inputs a list of configuration descriptions matching their inputs.

Conclusions of https://github.com/eclipse-archived/smarthome/issues/6602 are:

- No short term solution to add support for input in the UIs. Proposal is to hide all ThingActions.
- The is uncertainty about when to use inputs and when to use configuration descriptions (e.g. currently all existing core actions for NGRE defined them twice, an input and a related configuration).

Question is how to proceed? Should we hide all `ThingActions` in the near future? Or would it be an option to introduce a way to add configuration descriptions to `ThingActions` (e.g. by adding a new annotation for it)?

A previous private discussion between @kaikreuzer , @ghys , @openhab-5iver and myself resulted in the following brainstorming:

- Displaying inputs along with config parameters should be doable to configure an action instance in a rule.
[...]

PR #1947, opened 2020-12-16
[automation] Temporarily hide Binding ThingActions in UIs

cweitkamp:
- Temporarily hide Binding action in UIs

This PR will temporarily hide all Binding ThingActions in UIs. Please do not merge if you do not agree on this temporarily solution.

Related to #1745


kaikreuzer:
This will hide actions on existing rules, but not the possibility to select thing actions in the UI, right @ghys?

ghys:
They will be hidden from view with this filter:
https://github.com/openhab/openhab-webui/blob/e16bf67a3e6d5e0306ee1eeeafca6c387bdf113c/bundles/org.openhab.ui/web/src/pages/settings/rules/rule-module-popup.vue#L128

But IIRC the autocompletion in the code view will still show them. I have no problem with that, what about you?

kaikreuzer:
No, that should be just fine.

> Please do not merge if you do not agree on this temporarily solution.

We do agree to the solution, but we cannot merge, since it is still a draft :-)

cweitkamp:
Ready for review.

kaikreuzer (approved):

PR #1953, opened 2020-12-16
Introduce ready marker to correctly set start level 80

kaikreuzer:
We claim in the start level description that 80 is only reached once thing handlers are initialized - this is the code to actually fulfill this promise.


wborn (approved):
LGTM and it builds fine locally.

PR #1961, opened 2020-12-18
Introduced rule loaded trigger

kaikreuzer:
This brings full backward compatibility to DSL system started rules as it re-triggers rules upon modification.
For UI created rules, it adds another choice of trigger besides the start level trigger.

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleEngineImpl.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleEngineImpl.java
@@ -507,6 +507,7 @@ public class RuleEngineImpl implements RuleManager, RegistryChangeListener<Modul
                     f.cancel(true);
                 }
             }
+            executeRuleOnLoad(rule.unwrap());
         }
     }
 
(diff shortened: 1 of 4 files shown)


kaikreuzer:
Dropping that idea as it is too late (a too complex&risky change) for 3.0 and after the release, the backward compatibility is not that important anymore - and if people want to retrigger certain rules, this is now easily possible within the UI.

giig1967g:
Hi Kai,
I would like you to reconsider your decision. 
This "rule loaded" trigger could be essential in case of global variables.
Please consider this situation:

```
var int dummy=0

rule "test"
when
 System started or
 <other conditions>
then
 dummy=item1.state
end

rule "test2"
when
then
<use dummy variable>
end

```

As it is a global variable, it is supposed to maintain it's value during the whole life of the application.
But if you save the file, the global variable value is reset to 0
Until Version 2.5 the "System started" was triggered when the rule file was saved, and this allowed to restore the variable value, but starting from OH3 everytime the file is saved, the variable value becomes 0 messing up all the code that depends on that variable. The only way to solve the situation now is manually run the rule, but as all manual things anyone could forget to run the rule. So a "rule loaded" trigger could be very helpful.

I hope I have been clear.

Thanks for your time.

kaikreuzer:
@giig1967g Yes, I see and understand the use case of initializing variables through such rules.
Feel free to open a new issue (discussions on closed issues tend to get lost&forgotten very quickly).
My main issue I had when implementing it was to find a proper meaning of that trigger for rules created in the UI. What should "loaded" mean in such a context? Execute it upon every change through the UI? What about disabled rules that are enabled again? Is that "loading" or not? It wasn't easy for me to find answers...

PR #1963, opened 2020-12-19
Fix CORS header to allow preflight requests with authorization payload

philippwaller:
Due to the introduction of the oAuth authorization layer, the CORS filter for preflight requests needs to be adapted. Currently, pre flight requests with an authorization header are rejected by the API. 
image
This pull request corrects this behaviour by adding the value "Authorization" to the "Access-Control-Allow-Headers“.

@kaikreuzer: If it's not too late, this fix might be interesting for the stable release.


kaikreuzer (approved):
Too late for RC2, but it should be little risk to add it and if it solves an issue here, let's get still get it in.

PR #1969, opened 2020-12-20
[auth] Don't allow creating API tokens with no name

ghys:
Fixes https://github.com/openhab/openhab-webui/issues/674.

Remove unnecessary checks and fix exception message.
Add Norwegian messages for auth pages.


ghys:
FTR I quickly tested it with success and I think it's benign enough to be merged before the release :)

kaikreuzer:
Did you add norwegian translations in this PR on purpose?

ghys:
Yes - since I was building that bundle anyway, I figured I could throw in the latest translations and test them at the same time. Apologies if I shouldn't have.

kaikreuzer (approved):
Would be nicer, also for the release notes, to have dedicated PRs for different topics, yes.

But change looks good and if you have tested it, let's get it in as a last minute merge. Build failure is not related, so I'll ignore that.

PR #1971, opened 2020-12-20
Fixed spelling of german laundry tag synonyms

HolgerHees:


kaikreuzer (approved):
Thanks. I see no risk with that PR, so I merge it despite the code freeze.

ghys:
These files are also on Crowdin (but I only focused on the auth messages for now).
These synonyms are used only for HABot.

PR #1985, opened 2020-12-22
Fixed NPE in ScriptError if INode is empty

cweitkamp:
- Fixed NPE in `ScriptError` if `INode` is empty

This will at lest show the message without referencing line and column.

See https://community.openhab.org/t/openhab-3-0-release-discussion/110751/16


kaikreuzer (approved):
lgtm, thanks!

PR #1986, opened 2020-12-22
Fixed invalid day of week max check

### LINKED ISSUE #1813: "value too high" cron expression not accepted although expression is correct in OH2
using 3.0M3

I deployed a rule that was working fine under 2.5.10
Replacing "MON-SUN" with "*" fixes this so it's not critical but nonetheless this should be fixed in the cron code not in users' rules.

Related: #1516 

```
2020-11-13 11:01:32.452 [ERROR] [ore.common.registry.AbstractRegistry] - Cannot inform the listener "org.openhab.core.automation.internal.RuleEngineImpl$2@1ae05dd" about the "ADDED" event: Value too high in range in cron expression '0 30 16 ? * MON-SUN' in field 'DayOfWeek': value 8, minimum: 7
java.lang.IllegalArgumentException: Value too high in range in cron expression '0 30 16 ? * MON-SUN' in field 'DayOfWeek': value 8, minimum: 7
```

Hilbrand:
Closes #1813


kaikreuzer (approved):
Thanks for the fix!

PR #1918, opened 2020-12-12
Fix remaining deprecations

wborn:
Fixes #1408

---

Some builders are in different packages as the types they build. So for those I've removed the public constructors by creating some sub-classes that call a protected constructor. That will hopefully make contributors use the builders.  In newer Java versions you can also limit which classes are allowed to subclass a certain type.

PR #1995, opened 2020-12-23
Hide StateDescription constructor

wborn:
I forgot to hide this constructor in https://github.com/openhab/openhab-core/pull/1918!
The add-ons were already updated for this.


cweitkamp (approved):
LGTM. Even if it is after OH3 release it still is nice to see it being erased.

PR #1996, opened 2020-12-23
Add CODEOWNERS file

wborn:
With a CODEOWNERS file we don't have to manually request reviews, so it prevents some clicking and may help speed up reviews and prevent duplicate PRs.


cweitkamp (approved):
LGTM. Appreciated. 👍

PR #1997, opened 2020-12-23
Upgrade to XStream 1.4.15

wborn:
For changes, see: https://x-stream.github.io/changes.html

Replaces #1979, #1980


cweitkamp (approved):
LGTM. Makes sense to do it at this point.

PR #2003, opened 2020-12-25
Updated rule dsl filename to rule uid mapping

Legion2:
I migrated from openhab 2.x to 3.x and many of my rules managed in `.rules` dsl files where not shown in the UI. I found that the uid of the rules only contained the string before ".".  Examples of my rule filenames: `Bedroom.TV.rules`, `Bedroom.Light.rules`
This resulted in a collision of uids for the rules. With openhab 2.x this was not a problem.

To fix this problem, the full filename without extension should be used in the uid for rules provided by DSL files.


cweitkamp (approved):
LGTM. Thanks.

PR #2021, opened 2020-12-28
[semantics] Added properties Distance and Angle

Rosi2143:
Proposal is to add Distance and Angle as additional property
this allows to use these basic units for items.

Signed-off-by: Michael Roßner <Schrott.Micha@web.de>


cweitkamp:
Thank you very much for your first contribution. And sorry for our late response.

May I ask you to prepare a related PR in https://github.com/openhab/openhab-webui/ (similar to https://github.com/openhab/openhab-webui/pull/537)?

bundles/org.openhab.core.semantics/src/main/java/org/openhab/core/semantics/model/property/Angle.java
```diff
@@ -0,0 +1,27 @@
+/**
+ * Copyright (c) 2010-2020 Contributors to the openHAB project
```
kaikreuzer, line 2: This looks like an outdated license header (saying 2020 instead of 2021). Does our generator script need an update?

kaikreuzer:
@Rosi2143 & @cweitkamp: I didn't follow the process of the ontology expansion in detail, so I wonder if this PR is aligned with https://github.com/openhab/openhab-core/issues/1791, which seems to be the "coordinating issue" for such changes?

cweitkamp:
I didn't follow #1791 in detail either. Thus I do not know if the list in the OP is fix or final or just an intermediate wrap up. Both properties are not mentioned in the discussion but from my pov the properties (for Point -> Measurements) should be aligned - beside others - with the list of dimensions provided by OHC [UoM model](https://www.openhab.org/docs/concepts/units-of-measurement.html).

Rosi2143:
I follow up on it with proposed #1791

PR #2030, opened 2020-12-29
Add cache for auth

spacemanspiff2007:
Adds a 10 entry cache for basic auth to fix performance issues

### DIFF
--- a/bundles/org.openhab.core.io.http.auth.basic/src/main/java/org/openhab/core/io/http/auth/basic/internal/BasicCredentialsExtractor.java
+++ b/bundles/org.openhab.core.io.http.auth.basic/src/main/java/org/openhab/core/io/http/auth/basic/internal/BasicCredentialsExtractor.java
@@ -12,9 +12,12 @@
  */
 package org.openhab.core.io.http.auth.basic.internal;
 
+import java.security.MessageDigest;
+
 import java.util.Base64;
 import java.util.HashMap;
 import java.util.Optional;
+import java.util.UUID;
 
 import javax.servlet.http.HttpServletRequest;
 
@@ -30,7 +33,9 @@ import org.osgi.service.component.annotations.Component;
  */
 @Component(property = { "context=javax.servlet.http.HttpServletRequest" })
 public class BasicCredentialsExtractor implements CredentialsExtractor<HttpServletRequest> {
-    static HashMap<String, UsernamePasswordCredentials> authCache = new HashMap<String, UsernamePasswordCredentials>();
+    private static final String UUID_STRING = UUID.randomUUID().toString();
+    static MessageDigest msgDigest = MessageDigest.getInstance("SHA-256");
+    static HashMap<ByteBuffer, UsernamePasswordCredentials> authCache = new HashMap<String, UsernamePasswordCredentials>();
 
     @Override
     public Optional<Credentials> retrieveCredentials(HttpServletRequest request) {
@@ -42,23 +47,26 @@ public class BasicCredentialsExtractor implements CredentialsExtractor<HttpServl
 
         String[] tokens = authenticationHeader.split(" ");
         if (tokens.length == 2) {
-            String authType = tokens[0];
+            final String authType = tokens[0];
             if (HttpServletRequest.BASIC_AUTH.equalsIgnoreCase(authType)) {
-                String auth_value = tokens[1];
-                UsernamePasswordCredentials cached_value = BasicCredentialsExtractor.authCache.get(auth_value);
-                if (cached_value != null) {
-                    return Optional.of(cached_value);
+                final String authValue = tokens[1];
+                ByteBuffer authHash = ByteBuffer.wrap(msgDigest.digest((tokens[1] + UUID_STRING).getBytes()));
+                
+                UsernamePasswordCredentials cachedValue = BasicCredentialsExtractor.authCache.get(authHash);
+                if (cachedValue != null) {
+                    return Optional.of(cachedValue);
                 }
 
-                String usernameAndPassword = new String(Base64.getDecoder().decode(auth_value));
-
+                String usernameAndPassword = new String(Base64.getDecoder().decode(authValue));
                 tokens = usernameAndPassword.split(":");
                 if (tokens.length == 2) {
-                    String username = tokens[0];
-                    String password = tokens[1];
+                    final String username = tokens[0];
+                    final String password = tokens[1];
 
                     UsernamePasswordCredentials creds = new UsernamePasswordCredentials(username, password);
-                    if (BasicCredentialsExtractor.authCache.size() > 9) {
+                    
+                    // limit auth cache to a certain size
+                    if (BasicCredentialsExtractor.authCache.size() > 20) {
                         Object remName = null;
                         for (Object obj : BasicCredentialsExtractor.authCache.keySet()) {
                             remName = obj;
@@ -66,7 +74,8 @@ public class BasicCredentialsExtractor implements CredentialsExtractor<HttpServl
                         }
                         BasicCredentialsExtractor.authCache.remove(remName);
                     }
-                    BasicCredentialsExtractor.authCache.put(auth_value, creds);
+                    
+                    BasicCredentialsExtractor.authCache.put(authHash, creds);
                     return Optional.of(creds);
                 }
             }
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.io.http.auth.basic/src/main/java/org/openhab/core/io/http/auth/basic/internal/BasicCredentialsExtractor.java
```diff
@@ -29,6 +30,7 @@
  */
 @Component(property = { "context=javax.servlet.http.HttpServletRequest" })
 public class BasicCredentialsExtractor implements CredentialsExtractor<HttpServletRequest> {
+    static HashMap<String, UsernamePasswordCredentials> authCache = new HashMap<String, UsernamePasswordCredentials>();
```
wborn, line 33: It's very unsecure to cache plain text credentials. If you inspect a heap dump you can easily get hold of them. With a bit of creativity and reflection you might even be able to obtain them when you are able to run some malicious code/rules.

wborn (changes requested):

bundles/org.openhab.core.io.http.auth.basic/src/main/java/org/openhab/core/io/http/auth/basic/internal/BasicCredentialsExtractor.java
spacemanspiff2007: It's base64 encoded so at least it's not directly readable. The OAuth2 tokens are cached, too. So whilst they don't leak the password it's possible to obtain them in the same way.

What would be another way to implement a cache?

wborn: The best practices that I know of are to immediately hash it and compare it to the stored hash. After that you dispose the sensitive data ASAP to prevent it from leaking in any way. I guess the hashing is the part giving performance issues? The  tokens need to be refreshed every once in a while using the password. So if you have a token it is only valid until it expires which is more secure than having a password that can be used indefinitely. The data in users.json is also hashed so you cannot obtain the password from it.

Some users share heap dumps publicly to get some help without knowing that it contains this kind of sensitive data. You can get malicious code to run by sharing some .jar in the community forums or in the market place. If users use the same credentials with other services they are in even bigger troubles when someone gets hold of it.

kaikreuzer: > I guess the hashing is the part giving performance issues?

Calculating hashes should really be a very quick and easy operation. I also hence wonder what exactly is causing the performance issues, then it can be better discussed what to do about it.

spacemanspiff2007: > The tokens need to be refreshed every once in a while using the password or refresh token. So if you have a token it is only valid until it expires which is more secure than having a password that can be used indefinitely.

That is correct, but not for long living tokens because they never expire.

> If users use the same credentials with other services they are in even bigger troubles when someone gets hold of it.

Whilst that is correct that's imho a bit out of scope. Password reuse is very bad and everybody knows it.
I don't see the need to ensure this programatically.

kaikreuzer:
> Calculating hashes should really be a very quick and easy operation. I also hence wonder what exactly is causing the performance issues, then it can be better discussed what to do about it.

Just did a quick test with this class:

```java
import java.util.*;

public class Test {

    public static void main(String[] args) {
        Base64.Encoder encoder = Base64.getEncoder();
        byte[] code = encoder.encode("TestString:WithPassword".getBytes());
        long start = System.currentTimeMillis();
        for(int i=0; i<1000000;i++) {
            Base64.getDecoder().decode(code);
        }
        long test1 = System.currentTimeMillis() - start;

        Base64.Decoder decoder = Base64.getDecoder();
        for(int i=0; i<1000000;i++) {
            decoder.decode(code);
        }
        long test2 = System.currentTimeMillis() - start - test1;
        System.out.println("Test 1: " + test1 + "ms");
        System.out.println("Test 2: " + test2 + "ms");
    }
}
```
Result on my Mac:
```
Test 1: 58ms
Test 2: 52ms
```
I think this proves that decoding basic auth credentials cannot be the cause of the performance issue.

bundles/org.openhab.core.io.http.auth.basic/src/main/java/org/openhab/core/io/http/auth/basic/internal/BasicCredentialsExtractor.java
wborn: > I changed it so plaintext is no longer stored.

They are still stored unsecure because they are part of `UsernamePasswordCredentials`.

ghys:
> Calculating hashes should really be a very quick and easy operation.

When you deal with password storage, not so much. When you generate and verify secrets, the operation 
is in fact _supposed_ to be a tiny bit slow, so if someone gains access to the system and gets hold of the database with the hashed passwords they should not be able to crack them easily using brute-force. A few more milliseconds to check a password leads to exponentially more time and/or computational power needed to guess it with brute force.

The current implementation of the password hashing uses PBKDF2-SHA512 with a salt (to prevent rainbow tables attacks) with [respectfully 65535 iterations for passwords and 1024 iterations for API tokens](https://github.com/openhab/openhab-core/blob/c28090497da2678850056bcc7b3e9d46d3db5e9b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/auth/UserRegistryImpl.java#L62). I'm open to a debate on these figures, or to make them configurable (while making sure existing passwords doesn't get invalidated).

This stackexchange question has more info about these: https://security.stackexchange.com/questions/3959/recommended-of-iterations-when-using-pkbdf2-sha256. Specifically [this answer](https://security.stackexchange.com/a/211146) contains recent benchmarks with different algorithms and iteration counts. For instance [werkzeug.security](https://werkzeug.palletsprojects.com/en/0.16.x/utils/#werkzeug.security.check_password_hash) defaults to 150000 iterations and PBKDF2-SHA256. [In 2013](https://github.com/signalapp/Signal-Android/issues/184), the Signal app for Android increased it up to 15000. I think we owe it to the users to make sure we care about their passwords properly.

Now why the low iteration counts for API tokens and passwords? Well, passwords are chosen by the user, that means they may be weak, sensible to dictionary attacks or containing personal information, while API tokens _(not to be confusing with access tokens used in conjunction with OAuth2, which are signed JWTs)_ are long, opaque strings of random alphanumeric characters so we have a guarantee that they will have a high entropy, and can therefore afford to make their verification less expensive. In the benchmarks above we can see the huge difference between 1000 and 15000 iterations.

Also, since passwords were supposed to be verified infrequently, only when opening a new session of creating an API token, performance was not deemed an issue - even 100ms to verify a password is acceptable when you only do it a few times a day, at most. The recommended way of authenticating services is to create an API token, which will be more performant by an order of magnitude, and has only benefits: you can create them and revoke them with simple commands - `openhab:users addApiToken <username> <tokenname> <scope>` or with the self-service web UI, you can scope them to certain operations only in the future, and API tokens can act as usernames for Basic authentication. In fact if it were me, I'd remove the ability to use Basic auth with the real username/password now that they exist, or at least hide it behind the "show advanced" checkbox to discourage using it.

In other words, tl;dr: **use API Tokens**.

bundles/org.openhab.core.io.http.auth.basic/src/main/java/org/openhab/core/io/http/auth/basic/internal/BasicCredentialsExtractor.java
spacemanspiff2007: > 
> They are still stored unsecure because they are part of `UsernamePasswordCredentials`.

Imho they contain the hashed password (with 65k iterations), not the plaintext password.

spacemanspiff2007:
@kaikreuzer 
You are doing the base64 decode which is more of an encoding operation.
The expensive part is the hashing which is buried deeper.

---

> I think we owe it to the users to make sure we care about their passwords properly.

Definitely. Since it is easy to gain read access to the openhab configuration folder I am not debating the current implementation.
Stored passwords need to be hashed frequently and if we want to increase the numbers it good to me.

> Also, since passwords were supposed to be verified infrequently, only when opening a new session or creating an API token, performance was not deemed an issue - even 100ms to verify a password is acceptable when you only do it a few times a day, at most. 

I am sending at least 1 req/sec and sometimes up to 20 (or more) req/sec to the REST API.
It's unclear to me why you would think that subsequent verification times are not an issue when every rest integration (e.g. Node-Red, HABApp, GUIs, etc.) makes extensive use of the REST API.

So there are two things that need to be differentiated:
1. The storage and creation of passwords which is fine the way it is. I am not advocating to weaken security or anything else.
2. Verification time for subsequent requests:
While the first request may take long to perform the password validation this doesn't have to be done for each subsequent request.

The only argument against 2. is that it is possible to see the user/pw combination if a user shares a memory dump publicly.
And even then it would mean no compromise of security if the user uses a unique password.
So if we implement a (fast!) caching mechanism that mitigates the argument against 2, we can have both: speed and auth.

The implementation of auth is to secure the openhab installation against invalid access from the home network.
If I run a reverse proxy with ssl certificates and authentication why do I need to do additional auth from openhab to the reverse proxy on the same machine? It's just a hindrance and not necessary any and thus my requirement to have to possibility to turn auth off.

Another idea is to just display info to the user:
- don't reuse passwords
- If you share memory dumps passwords may be stored in plain text

spacemanspiff2007:
This does not work, correct fix is #2101

PR #2036, opened 2020-12-30
[Persistence Extensions] Fixed varianceSince and deviationSince

mhilbush:
Fixed `varianceSince` and `deviationSince`, which always returned null. Also enabled tests for `varianceSince` and `deviationSince`, which were not being executed (and therefore masking that those functions didn't work).

Signed-off-by: Mark Hilbush <mark@hilbush.com>


cweitkamp (approved):
LGTM. Thanks.

PR #2043, opened 2021-01-01
Added rule condition for not a holiday

### LINKED ISSUE #2035: Rule Engine: Condition for "it's NOT a holiday"
Hey,

i try to set up an easy rule within my openhab 3 installation. It should run every day at an specific time when i've to get up to work.
So it should run all days, that are not at the weekend and not an holiday.

Defining this rule is straight forward until the "not an holiday" condition. The ephemeris condition has an "it's a weekday" condition (which is actually a it't not the weekend) but there is no corresponding possibility for "it's not a holiday / its a working day" or something similar.

It's possible to create an rule to achive the desired behavior - when using an dsl script as an action which has the content
```
if(!(Ephemeris.isBankHoliday() || Ephemeris.isWeekend())) {
... do something
}
```
but if done so the new schedule feature is not correct any more.

So my question is, if there is a reason for this missing condition?

soenkekueper:
Fixes #2035
Signed-off-by: Sönke Küper <soenkekueper@gmx.de>

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/factory/EphemerisModuleHandlerFactory.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/factory/EphemerisModuleHandlerFactory.java
@@ -44,8 +44,8 @@ public class EphemerisModuleHandlerFactory extends BaseModuleHandlerFactory impl
     private final Logger logger = LoggerFactory.getLogger(EphemerisModuleHandlerFactory.class);
 
     private static final Collection<String> TYPES = List.of(EphemerisConditionHandler.HOLIDAY_MODULE_TYPE_ID,
-            EphemerisConditionHandler.WEEKEND_MODULE_TYPE_ID, EphemerisConditionHandler.WEEKDAY_MODULE_TYPE_ID,
-            EphemerisConditionHandler.DAYSET_MODULE_TYPE_ID);
+            EphemerisConditionHandler.NOT_HOLIDAY_MODULE_TYPE_ID, EphemerisConditionHandler.WEEKEND_MODULE_TYPE_ID,
+            EphemerisConditionHandler.WEEKDAY_MODULE_TYPE_ID, EphemerisConditionHandler.DAYSET_MODULE_TYPE_ID);
 
     private final EphemerisManager ephemerisManager;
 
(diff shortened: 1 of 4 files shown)


soenkekueper:
The jenkins build for this pull request was aborted after 120min, what could be the root cause of such a long running built?
I believe that this has nothing to do with my changes...
Is there anything i could do to get this ready to get merged?

wborn:
I've tuned the builds a bit recently, so let's see if it succeeds this time. :slightly_smiling_face:

soenkekueper:
Nice, thanks!

kaikreuzer:
Looks good in general, but what I dislike a bit is the inherent negation - all the other ephemeris conditions don't do this.
I wonder, if a "is working day" condition would also fulfill your requirement? While it has a more complex meaning (as it combines weekday and holiday check), it might be the condition that most users actually need in their rules. Wdyt?

soenkekueper:
Hey,
i don't have any "is working day" condition at all. There is only an "it is a weekday" condition which is just the negation of "it is a weekend" but don't care about holiday. So which condition do you mean? When looking into the EphemerisConditionHandler and EphemerisManager.isBankHoliday(ZonedDateTime) it's just used by the "it is a holiday" condition.

I think you're right for most users a "is working day" condition would match, but i think the given "not a holiday" condition is more flexible, as some users may have to work on the weekend days too.

kaikreuzer:
My issue with "not a holiday" is that it is merely a negation of an existing condition and I'd somewhat want to avoid a proliferation of conditions that way. I agree that this is already the case with "is weekday" and "is weekend", although the negation is not so directly visible.
Ideally, I'd say the holiday condition should just have another config parameter to allow the negation. E.g. for item state conditions we also have "==" and "!=" as configuration options. Technically, this is definitely my preferred solution. The hardest thing about it in this case is to find a proper name for the enhanced condition, but I bet we can work something out.
@clinique As the master of Ephemeris, what's your view on this?

clinique:
> @clinique As the master of Ephemeris, what's your view on this?

I think that "is working day" or any equivalent would intuitively means neither weekend, neither holiday, so perhaps not very elegant by "not an holiday" seems less ambiguous.

kaikreuzer (approved):
Ok, thanks for your feedback @clinique - then let's go for the less elegant, but also less ambiguous option. :-)

soenkekueper:
@kaikreuzer Thanks for merging.
There is also PR: openhab/openhab-webui#741 That adds the new Condition within the Ephemeris Quick Selection in the web-UI. Otherwise it's only available when choosing "show all". 
Sorry i've missed to link this on creation.

PR #2050, opened 2021-01-01
Extended serial port implementation

### LINKED ISSUE #2048: SerialPort interface missing methods implementation
SerialPort interface in package org.openhab.core.io.transport.serial does not implement all methods to control serial port. These are methods that provide port status such as port setting and control signal status.

docbender:
Serial port interface is extended by non implemented methods
Fixes #2048 
Signed-off-by: Vita Tucek


kaikreuzer (approved):
Lgtm, thanks!

I thought to vaguely remember that there was a reason for not implementing all of them - something like supporting RfC2217 ports as well, but checking the implementation, it seems that we do not have any constraint there.
@openhab/core-maintainers If I am missing anything, please scream ;-)

PR #2053, opened 2021-01-02
Improved error logging when adding rules

bergvandenp:
I stumbled upon the same issue as described here:

https://community.openhab.org/t/solved-oh3-issue-unable-to-create-rules-scripts-schedules/111325
https://community.openhab.org/t/openhab-3-error-creating-rules-or-scripts-using-blocky/111422

This makes the logging a bit more useful

### DIFF
--- a/bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/RuleResource.java
+++ b/bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/RuleResource.java
@@ -189,7 +189,7 @@ public class RuleResource implements RESTResource {
 
     private void logException(RuntimeException e, String errMessage) {
         if (logger.isDebugEnabled()) {
-            logger.warn(errMessage, e);
+            logger.warn("{}", errMessage, e);
         } else {
             logger.warn("{}", errMessage);
         }
--- a/bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/RuleResource.java
+++ b/bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/RuleResource.java
@@ -178,12 +178,20 @@ public class RuleResource implements RESTResource {
                     .header("Location", "rules/" + URLEncoder.encode(newRule.getUID(), StandardCharsets.UTF_8)).build();
         } catch (IllegalArgumentException e) {
             String errMessage = "Creation of the rule is refused: " + e.getMessage();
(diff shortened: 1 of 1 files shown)


kaikreuzer (changes requested):
We do not log stacktraces when there is no bug in openHAB itself.
What can be done is to log exceptions when debugging is enabled for a given class.

NorbertHD:
@bergvandenp can you please change it to:
````
                logger.warn(errMessage, e);
            } else {
                logger.warn("{}", errMessage);
            }
````

**Review by kaikreuzer — APPROVED** [review]
thanks!

**Review by kaikreuzer — CHANGES_REQUESTED** [review]
Now you only have to make SAT happy:
```
[ERROR] Code Analysis Tool has found: 
 1 error(s)! 
 4 warning(s) 
 1 info(s)
[ERROR] org.openhab.core.automation.rest.internal.RuleResource.java:[192]
Format should be constant. Use placeholder to reduce the needless cost of parameter construction. see http://www.slf4j.org/faq.html#logging_performance
```

File: bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/RuleResource.java [hunk 1]
```diff
@@ -178,15 +178,23 @@ public Response create(@Parameter(description = "rule data", required = true) Ru
...
             String errMessage = "Creation of the rule is refused: " + e.getMessage();
-            logger.warn("{}", errMessage);
+            logException(e, errMessage);
             return JSONResponse.createErrorResponse(Status.BAD_REQUEST, errMessage);
         }
     }
 
+    private void logException(RuntimeException e, String errMessage) {
+        if (logger.isDebugEnabled()) {
+            logger.warn(errMessage, e);
```
kaikreuzer, line 192:
```suggestion
            logger.warn("{}", errMessage, e);
```

NorbertHD: @kaikreuzer does this work?
AFAIK there is no warn​(String format, Object... arguments, Throwable t)

kaikreuzer: Yes, that works.

bergvandenp: whoops. my bad. thanks for the suggestion

kaikreuzer (approved):
Thanks!

Issue #2031, opened 2020-12-29
DSL-rules via UI cause "out of memory" errors (java heap space)


spy0r:
**System (Proxmox VM):**

> stef@openhab3:~$ uname -a
Linux openhab3 4.19.0-13-amd64 openhab/openhab-distro#1 SMP Debian 4.19.160-2 (2020-11-28) x86_64 GNU/Linux

**Openhab:**

> 3.0.0 - Release Build

**Java:**

> stef@openhab3:~$ java --version
openjdk 11.0.9.1 2020-11-04 LTS
OpenJDK Runtime Environment Zulu11.43+55-CA (build 11.0.9.1+1-LTS)
OpenJDK 64-Bit Server VM Zulu11.43+55-CA (build 11.0.9.1+1-LTS, mixed mode)

**Scenario**: Migrating DSL-rules from OH2 to OH3 via the UI cause "out of memory issues" (java heap space). One can see, that those (easy) rules run extremely long (up to many seconds - seen in the rule status indicator in the UI or via the java processes top -H -p yourOH3javaProcess)

**Solution**: To get a stable OH3 instance i needed to put those rules in files again, like in OH2. Same rules, same trigger, no problems. Those rules run that fast then, that you can not even see the "running" state in the rule indicater.

See also https://community.openhab.org/t/openhab-3-openhabian-runs-out-of-memory-java-heap-space-errors-cpu-100-after-a-few-hours

An example rule was this:
```
rule "Wohnzimmerlicht"
when
    Time cron "0 0/5 * * * ?" or
    Item presence_wohnbereich received update
then
    if (presence_wohnbereich.state == ON) {
        if (now.hour >= 9 && group_light_wohnzimmer.state != ON && (lumen_wohnzimmer.state < 5 || now.isAfter((LokaleSonnendaten_Sonnenuntergang.state as DateTimeType).getZonedDateTime))) {
            sendCommand(group_light_wohnzimmer, ON)
        }
        else if (group_light_wohnzimmer.state == ON && lumen_wohnzimmer.state > 10) {
            sendCommand(group_light_wohnzimmer, OFF)
        }
    }
    else if (group_light_wohnzimmer.state != OFF) sendCommand(group_light_wohnzimmer, OFF)
end
```

spy0r:
I will try to do that in another VM

spy0r:
I could set up a very small example, which got me a 100% cpu usage after half an hour:

- Fresh Debian install inside Proxmox VM
[...]

Issue #1877, opened 2020-11-30
[OH3] Defining Rules in Main UI increases CPU Load


gdolfen:
I have a problem with high cpu usage in M2 and M3.
I figuerd out that the reason are rules which are configure over main ui.
I started with 4 rules in a *.rules file with a cpu usage of 1%. In main ui they always stay on "IDLE" after migration them to main ui by copy the generated code by the code view everything was fine, but the logs was full with ERRORs "No pre-parsed script found for e3dc-2". So I deleted the line "// context: e3dc-2" from the code and the log disappears, but the cpu was increased up to 5% per rule and the state stays always on "RUNNING".
Here is my rule code:

```
triggers:
  - id: "0"
    configuration:
      itemName: E3DC_Extern
    type: core.ItemStateChangeTrigger
conditions: []
actions:
  - inputs: {}
    id: script
    configuration:
      type: application/vnd.openhab.dsl.rule
      script: |
        // context: e3dc-2
        var Number negativeValue = E3DC_Extern.state as Number
        var Number positiveValue = negativeValue * -1
        E3DC_String3.postUpdate(positiveValue)
    type: script.ScriptAction
```
It seemed that the rule is parsed every time, because every 5 seconds (5 seconds update period of item E3DC_Extern) I got the log error. Maybe preparsing the rule after saving it in main ui makes the trick.

Boby71:
+1; same for me on OH3.0.0 stable:
After defining many DSL rules on my system, I run into errors like "java.lang.OutOfMemoryError: Java heap space" which leaves the system completely unresponsive after a while.

It seems to happen when the rule(s) are triggered often, by a constantly changing value (e.g. a temp sensor). The rules seem to take too long to run (e.g. a few seconds for a simple postUpdate) and when there enough rules running and new ones started upon, the system runs out of memory.

JohannesPtaszyk:
I got this issue as well. My hue dimmer switch state is changed almost instantly, but the rule for it takes like 10 seconds to execute

JohannesPtaszyk:
I moved my setup to a Raspberry Pi 4 4GB and now everything is working fine.

kaikreuzer:
Thanks for the report. I have created https://github.com/openhab/openhab-core/pull/2057 for it, it would be great if you could test a snapshot, once that PR is merged and available in the distro!

gdolfen:
@kaikreuzer: I testet your change with the latest snapshot build an the problem is gone. Thanks a lot for your fix!

PR #2057, opened 2021-01-02
[automation] Cache parsed script in order to improve performance

kaikreuzer:
Since we have a dedicated DSLScriptEngine instance per rule script action, we can easily cache the parsed script.
Parsing seems to cause significant CPU and memory usage - with this PR, there is no noticeable CPU activity anymore and also the heap size does not increase over time anymore.

Fixes https://github.com/openhab/openhab-core/issues/1877
Fixes https://github.com/openhab/openhab-core/issues/2031


wborn (approved):
Thanks! Looks like a very nice performance improvement!

DanielMalmgren:
I can't seem to find this PR in the list of fixes for 3.0.1. Does that mean we will have to either switch to snapshots or wait for the next milestone to get it? It's kinda critical for my system.

kaikreuzer:
@DanielMalmgren Added. 3.0.1 is not yet announced, unfortunately, there were some issues during the build.

PR #2063, opened 2021-01-03
[skeleton] Remove license header from feature.xml

fwolter:
See openhab/openhab-addons#9620

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -435,6 +435,7 @@ Import-Package: \\
               <exclude>target/**</exclude>
               <exclude>**/pom.xml</exclude>
               <exclude>_*.java</exclude>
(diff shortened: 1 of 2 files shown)


wborn (commented):
Can you also add this file to the [excludes](https://github.com/openhab/openhab-core/blob/531a478a4567fc6c195cdc57bda7d8b7ce15ddd7/pom.xml#L434-L438) so the header isn't added the next time `mvn license:format` is ran?

Adding something like this seems to work:

```xml
<exclude>**/archetype/**/feature.xml</exclude>
```

fwolter:
Sure, the exclude pattern works.

wborn (approved):
Thank you!

PR #2075, opened 2021-01-05
Update Mockito, JUnit and use MockitoExtensions in itests

wborn:
We can finally use `MockitoExtensions` in itests now that https://github.com/mockito/mockito/issues/1997 is fixed in Mockito 3.7.0.

This saves some lines of code, makes the tests more consistent and prevents contributors from wasting their time if they tried to use `MockitoExtensions` in itests.

---

* Updates Mockito to 3.7.0
* Updates JUnit to 5.7.0
* Use MockitoExtensions where possible


kaikreuzer (approved):
Many thanks!

PR #2088, opened 2021-01-08
[semantics] Added ActionService to support using Semantics features in Rules

cweitkamp:
- Added ActionService to support using Semantics features in Rules
- Enhanced unit tests

I like to use Semantic features in Rules. E.g. determine the location of an item. Regardless if it is a Point, an Equipment or a Location itself. Therefore I propose to add a bunch methods to be used in Rules. I submitted this as a draft because I am not sure if this is the preferred way to add such features.

- `boolean isLocation(Item)` - checks if the given Item is is a Location
- `boolean isEquipment(Item)` - checks if the given Item is is an Equipment
- `boolean isPoint(Item)` - checks if the given Item is is a Point
- `Item getLocation(Item)` - gets the Location Item of the Item
- `Class<? extends Location> getLocationType(Item)` - gets the Location type of the Item
- `Item getEquipment(Item)` - gets the Equipment Item an Item belongs to
- `Class<? extends Equipment> getEquipmentType(Item)` - gets the Equipment type an Item belongs to
- `Class<? extends Point> getPointType(Item)` - gets the Point type of an Item
- `Class<? extends Property> getPropertyType(Item)` - gets the Property type an Item relates to
- `Class<? extends Tag> getSemanticType(Item)` - gets the semantic type of an Item (i.e. a sub-type of Location, Equipment or Point)

### DIFF
--- a/bundles/org.openhab.core.model.script/bnd.bnd
+++ b/bundles/org.openhab.core.model.script/bnd.bnd
@@ -30,6 +30,7 @@ Import-Package: \
  org.openhab.core.persistence,\
  org.openhab.core.persistence.extensions,\
  org.openhab.core.scheduler,\
+ org.openhab.core.semantics,\
  org.openhab.core.thing,\
  org.openhab.core.thing.binding,\
  org.openhab.core.thing.events,\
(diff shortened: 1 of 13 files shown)


kaikreuzer (approved):
Excellent addition, thanks!

Issue #2084, opened 2021-01-07
Javascript based rules lack the ability for global variables


FSeidinger:
As already mentioned in the community under [Experimental Next-Gen Rules Engine Documentation 1 of : Introduction](https://community.openhab.org/t/experimental-next-gen-rules-engine-documentation-1-of-introduction/54669), the OH3 rule engine lacks the ability to hold variables that have a persistent nature for ECMA Script rules. As also described in the article, the most common use case for such variables would be timer handles or stored state.

So not having the possibility to store the timer handles reduces the acceptance and usability of rules based on javascript significantly. In contrast, the rule DSL based rules had this feature for a long time and they still have.

To solve this issue I can think of two solutions:

1. Introduction of a global context that gets injected into the script engine environment when the script is created or updated.
2. Adding a shutdown hook to rules that can be implemented by the rule script to give the script the ability to do it's housekeeping before it gets destroyed.

kaikreuzer:
>  In contrast, the rule DSL based rules had this feature for a long time and they still have.

That's not true - there are no global variables and I am not convinced that global variables are a good concept in general.
What DSL rules (from rule files) offer is a shared context between rules within the same file. Files hence create a natural boundary for them, especially as they are also loaded and refreshed together.

As you mainly seem to be concerned about Javascript rules: Isn't it already possible to have shared variables in Javascript files (by creating scripted rules)? This would then be similar to what exists for DSL rules (from files) and probably be the simplest solution.

If we really want to introduce global scopes in rules in general, we should imho not only do it in ScriptActions, but in ALL rule modules, independent of whether they are scripted or Java. But that would be a bigger change and it would break the current isolation, potentially introduce concurrency issues and increase the complexity.

FSeidinger:
> That's not true - there are no global variables and I am not convinced that global variables are a good concept in general.
> What DSL rules (from rule files) offer is a shared context between rules within the same file. Files hence create a natural boundary for them, especially as they are also loaded and refreshed together.

Dear Kai,

thanks for the clarification. Maybe I was not very precise with my words. But there is a mismatch in the behaviour of the contexts between the DSL scripts and the script engine scripts. When you declare variable in a Rule DSL file outside the rules itself, you get, what most of the users would call a 'global variable'. It is true, that it is only global in the context of the file, but it survives changes and edits to the rule and the action script. So, if I store e.g. a timer instance in such a variable, it will be there next time the rule runs, even if it was updated or edited.

[...]

PR #2089, opened 2021-01-08
Adds global context variables to action scripts

FSeidinger:
Adds a global context to the ScriptEngineManager that can be used by action handlers to setup an execution context with a global scope. Currently only the ScriptActionHandler uses this feature.

It's purpose is to share global variables between scripts or letting scripts store variables like timers or item states.

It's one of the solutions to my issue #2084.


kaikreuzer (changes requested):
Since https://github.com/openhab/openhab-core/issues/2084 is under active discussion and it so far looks as if the suggested solution will look differently than the change of this PR, I'm adding a blocking review here, just as a marker to prevent an accidental merge.

kaikreuzer:
As we agreed in #2084 that the solution of this PR should not be the way to go, I hope it is ok to close it.

Issue #1899, opened 2020-12-07
[RestAPI] Very poor performance compared to OH2


spacemanspiff2007:
Test:
Create 300 Number items and measure how long it takes.
Post an update to 300 Number items and see how long it takes until all items report the new state through SSE events.

While doing the other tests:
Ping an item every 10 secs and measure how long it takes until the new state is reported back  through an SSE events.

Both tests were done with
```
openjdk version "11.0.8" 2020-07-14 LTS
OpenJDK Runtime Environment Zulu11.41+23-CA (build 11.0.8+10-LTS)
OpenJDK 64-Bit Server VM Zulu11.41+23-CA (build 11.0.8+10-LTS, mixed mode)
```

3.0.0.M5
```
Updated 300 item definitions in 45.037s --> 6.661 updates per sec

Benchmark duration: 7.151s --> 41.950 updates per sec
Pings (11): min: 144.0 max: 188.0 median: 182.0

Benchmark duration: 7.169s --> 41.844 updates per sec
Pings (5): min: 147.0 max: 4890.3 median: 176.0

Benchmark duration: 7.061s --> 42.484 updates per sec
Pings (6): min: 169.0 max: 201.0 median: 190.0
```
CPU spikes to 100% on all cores
![grafik](https://user-images.githubusercontent.com/10754716/101356828-f023e200-3898-11eb-8d82-59682407fa43.png)

2.5.3
```
Updated 300 item definitions in 2.313s --> 129.694 updates per sec

Benchmark duration: 1.965s --> 152.663 updates per sec
Pings (6): min: 11.0 max: 39.0 median: 25.0

Benchmark duration: 1.972s --> 152.121 updates per sec
Pings (6): min: 9.0 max: 26.0 median: 21.5

Benchmark duration: 1.933s --> 155.190 updates per sec
Pings (6): min: 19.0 max: 25.0 median: 22.5
```
Absolutely no CPU spike whatsoever
![grafik](https://user-images.githubusercontent.com/10754716/101357769-2f9efe00-389a-11eb-9a1f-e325421336b2.png)

Changes compared to OH2:

Test | speed change
---  | ---
Item update | 18x slower
Value update | 3.6x slower
Round trip time | 7x slower

kaikreuzer:
So do I get you correctly that you identified the REST API/SSE to be a bottleneck?
Or is it the item event processing itself? Did you test that independently?

spacemanspiff2007:
[...]

PR #2101, opened 2021-01-11
Add cache for Basic Authentication

spacemanspiff2007:
This PR adds a small cache for calculating the auth hashes which aims to fix the performance issues introduced with OH3.
The cache does **not** store plain text passwords and is limited to a certain size.

Fixes https://github.com/openhab/openhab-core/issues/1899

### DIFF
--- a/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
+++ b/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
@@ -13,8 +13,13 @@
 package org.openhab.core.io.rest.auth.internal;
 
 import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.util.Base64;
+import java.util.HashMap;
 import java.util.Map;
+import java.util.Random;
 
 import javax.annotation.Priority;
 import javax.ws.rs.Priorities;
@@ -75,6 +80,28 @@ public class AuthFilter implements ContainerRequestFilter {
     private boolean allowBasicAuth = false;
     private boolean implicitUserRole = true;
 
+    private static HashMap<ByteBuffer, UserSecurityContext> authCache = new HashMap<ByteBuffer, UserSecurityContext>();
+    private static byte[] rndBytes = getRndBytes();
+
+    private static byte[] getRndBytes() {
+        final Random rd = new Random();
+        byte[] rndBytes = new byte[32];
+        rd.nextBytes(rndBytes);
+        return rndBytes;
+    }
+
+    private static ByteBuffer getCacheKey(String creds) {
+        try {
+            final MessageDigest md = MessageDigest.getInstance("SHA-256");
+            md.update(AuthFilter.rndBytes);
+            return ByteBuffer.wrap(md.digest(creds.getBytes()));
+        } catch (NoSuchAlgorithmException e) {
+            // SHA-256 is available for all java distributions so this code will actually never run
+            // If it does we'll just flood the cache with random values
+            return ByteBuffer.wrap(getRndBytes());
+        }
+    }
+
     @Reference
     private JwtHelper jwtHelper;
 
@@ -93,6 +120,7 @@ public class AuthFilter implements ContainerRequestFilter {
             allowBasicAuth = value != null && "true".equals(value.toString());
             value = properties.get(CONFIG_IMPLICIT_USER_ROLE);
             implicitUserRole = value == null || !"false".equals(value.toString());
+            AuthFilter.authCache.clear();
         }
     }
 
@@ -111,15 +139,33 @@ public class AuthFilter implements ContainerRequestFilter {
         }
     }
 
-    private SecurityContext authenticateUsernamePassword(String username, String password)
+    private SecurityContext authenticateUsernamePassword(String username, String password, String authCacheKey)
             throws AuthenticationException {
+        final ByteBuffer cacheKey = getCacheKey(authCacheKey);
+        final UserSecurityContext cachedValue = AuthFilter.authCache.get(cacheKey);
+        if (cachedValue != null) {
+            return cachedValue;
+        }
+
         UsernamePasswordCredentials credentials = new UsernamePasswordCredentials(username, password);
         Authentication auth = userRegistry.authenticate(credentials);
         User user = userRegistry.get(auth.getUsername());
         if (user == null) {
             throw new AuthenticationException("User not found in registry");
         }
-        return new UserSecurityContext(user, auth, "Basic");
+
+        // limit cache to a certain size
+        if (AuthFilter.authCache.size() > 20) {
+            Object remName = null;
+            for (Object obj : AuthFilter.authCache.keySet()) {
+                remName = obj;
+                break;
+            }
+            AuthFilter.authCache.remove(remName);
+        }
+        UserSecurityContext context = new UserSecurityContext(user, auth, "Basic");
+        AuthFilter.authCache.put(cacheKey, context);
+        return context;
     }
 
     @Override
@@ -142,22 +188,17 @@ public class AuthFilter implements ContainerRequestFilter {
                         try {
                             String[] decodedCredentials = new String(Base64.getDecoder().decode(authParts[1]), "UTF-8")
                                     .split(":");
-                            if (decodedCredentials.length > 2) {
+                            if (decodedCredentials.length != 2) {
                                 throw new AuthenticationException("Invalid Basic authentication credential format");
                             }
-                            switch (decodedCredentials.length) {
-                                case 1:
-                                    requestContext.setSecurityContext(authenticateBearerToken(decodedCredentials[0]));
-                                    break;
-                                case 2:
-                                    if (!allowBasicAuth) {
-                                        throw new AuthenticationException(
-                                                "Basic authentication with username/password is not allowed");
-                                    }
-                                    requestContext.setSecurityContext(
-                                            authenticateUsernamePassword(decodedCredentials[0], decodedCredentials[1]));
+                            if (!allowBasicAuth) {
+                                throw new AuthenticationException(
+                                        "Basic authentication with username/password is not allowed");
                             }
-
+                            requestContext.setSecurityContext(
+                                    // use the base64 String as a key to the cache because it's long and already created
+                                    authenticateUsernamePassword(decodedCredentials[0], decodedCredentials[1],
+                                            authParts[1]));
                             return;
                         } catch (AuthenticationException e) {
                             throw new AuthenticationException("Invalid Basic authentication credentials", e);
--- a/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
+++ b/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
@@ -13,6 +13,7 @@
 package org.openhab.core.io.rest.auth.internal;
 
 import java.io.IOException;
+import java.io.UnsupportedEncodingException;
 import java.nio.ByteBuffer;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
@@ -38,6 +39,7 @@ import org.openhab.core.auth.User;
 import org.openhab.core.auth.UserApiTokenCredentials;
 import org.openhab.core.auth.UserRegistry;
 import org.openhab.core.auth.UsernamePasswordCredentials;
+import org.openhab.core.common.registry.RegistryChangeListener;
 import org.openhab.core.config.core.ConfigurableService;
 import org.openhab.core.io.rest.JSONResponse;
 import org.openhab.core.io.rest.RESTConstants;
@@ -51,6 +53,30 @@ import org.osgi.service.jaxrs.whiteboard.propertytypes.JaxrsApplicationSelect;
 import org.osgi.service.jaxrs.whiteboard.propertytypes.JaxrsExtension;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+class SimpleUserRegistryCallback implements RegistryChangeListener<User> {
+
+    private Runnable callback;
+
+    public SimpleUserRegistryCallback(Runnable callback) {
+        this.callback = callback;
+    }
+
+    @Override
+    public void added(User element) {
+        return;
+    }
+
+    @Override
+    public void removed(User element) {
+        callback.run();
+    }
+
+    @Override
+    public void updated(User oldElement, User element) {
+        callback.run();
+    }
+}
 
 /**
  * This filter is responsible for parsing credentials provided with a request, and hydrating a {@link SecurityContext}
@@ -77,11 +103,13 @@ public class AuthFilter implements ContainerRequestFilter {
     private static final String CONFIG_ALLOW_BASIC_AUTH = "allowBasicAuth";
     private static final String CONFIG_IMPLICIT_USER_ROLE = "implicitUserRole";
 
+    private static final int authCacheSize = 10;
+
     private boolean allowBasicAuth = false;
     private boolean implicitUserRole = true;
 
-    private static HashMap<ByteBuffer, UserSecurityContext> authCache = new HashMap<ByteBuffer, UserSecurityContext>();
-    private static byte[] rndBytes = getRndBytes();
+    private HashMap<ByteBuffer, UserSecurityContext> authCache = new HashMap<ByteBuffer, UserSecurityContext>();
+    private byte[] rndBytes = getRndBytes();
 
     private static byte[] getRndBytes() {
         final Random rd = new Random();
@@ -90,17 +118,37 @@ public class AuthFilter implements ContainerRequestFilter {
         return rndBytes;
     }
 
-    private static ByteBuffer getCacheKey(String creds) {
+    @Nullable
+    private ByteBuffer getCacheKey(String creds) {
         try {
             final MessageDigest md = MessageDigest.getInstance("SHA-256");
-            md.update(AuthFilter.rndBytes);
+            md.update(rndBytes);
             return ByteBuffer.wrap(md.digest(creds.getBytes()));
         } catch (NoSuchAlgorithmException e) {
             // SHA-256 is available for all java distributions so this code will actually never run
             // If it does we'll just flood the cache with random values
-            return ByteBuffer.wrap(getRndBytes());
+            logger.warn("SHA-256 is not available. Cache for basic auth disabled!");
+            return null;
         }
     }
+
+    private RegistryChangeListener<User> cacheCleanListener = new RegistryChangeListener<User>() {
+
+        @Override
+        public void added(User element) {
+            return;
+        }
+
+        @Override
+        public void removed(User element) {
+            authCache.clear();
+        }
+
+        @Override
+        public void updated(User oldElement, User element) {
+            authCache.clear();
+        }
+    };
 
     @Reference
     private JwtHelper jwtHelper;
@@ -111,6 +159,8 @@ public class AuthFilter implements ContainerRequestFilter {
     @Activate
     protected void activate(Map<String, Object> config) {
         modified(config);
+
+        userRegistry.addRegistryChangeListener(cacheCleanListener);
     }
 
     @Modified
@@ -120,7 +170,7 @@ public class AuthFilter implements ContainerRequestFilter {
             allowBasicAuth = value != null && "true".equals(value.toString());
             value = properties.get(CONFIG_IMPLICIT_USER_ROLE);
             implicitUserRole = value == null || !"false".equals(value.toString());
-            AuthFilter.authCache.clear();
+            authCache.clear();
         }
     }
 
@@ -139,32 +189,47 @@ public class AuthFilter implements ContainerRequestFilter {
         }
     }
 
-    private SecurityContext authenticateUsernamePassword(String username, String password, String authCacheKey)
-            throws AuthenticationException {
-        final ByteBuffer cacheKey = getCacheKey(authCacheKey);
-        final UserSecurityContext cachedValue = AuthFilter.authCache.get(cacheKey);
-        if (cachedValue != null) {
-            return cachedValue;
+    private SecurityContext authenticateBasicAuth(String credentialString) throws AuthenticationException {
+        if (!allowBasicAuth) {
+            throw new AuthenticationException("Basic authentication with username/password is not allowed");
         }
 
-        UsernamePasswordCredentials credentials = new UsernamePasswordCredentials(username, password);
+        final ByteBuffer cacheKey = getCacheKey(credentialString);
+        if (cacheKey != null) {
+            final UserSecurityContext cachedValue = authCache.get(cacheKey);
+            if (cachedValue != null) {
+                return cachedValue;
+            }
+        }
+
+        String[] decodedCredentials;
+        try {
+            decodedCredentials = new String(Base64.getDecoder().decode(credentialString), "UTF-8").split(":");
+        } catch (UnsupportedEncodingException e) {
+            throw new AuthenticationException("Invalid base64 credentials for basic auth");
+        }
+        if (decodedCredentials.length != 2) {
+            throw new AuthenticationException("Invalid Basic authentication credential format");
+        }
+
+        UsernamePasswordCredentials credentials = new UsernamePasswordCredentials(decodedCredentials[0],
+                decodedCredentials[1]);
         Authentication auth = userRegistry.authenticate(credentials);
         User user = userRegistry.get(auth.getUsername());
         if (user == null) {
             throw new AuthenticationException("User not found in registry");
         }
 
-        // limit cache to a certain size
-        if (AuthFilter.authCache.size() > 20) {
-            Object remName = null;
-            for (Object obj : AuthFilter.authCache.keySet()) {
-                remName = obj;
-                break;
-            }
-            AuthFilter.authCache.remove(remName);
-        }
         UserSecurityContext context = new UserSecurityContext(user, auth, "Basic");
-        AuthFilter.authCache.put(cacheKey, context);
+
+        if (cacheKey != null) {
+            // limit the auth cache to a certain size
+            if (authCache.size() >= authCacheSize) {
+                authCache.remove(authCache.keySet().iterator().next());
+            }
+            authCache.put(cacheKey, context);
+        }
+
         return context;
     }
 
@@ -181,28 +246,14 @@ public class AuthFilter implements ContainerRequestFilter {
             if (authHeader != null) {
                 String[] authParts = authHeader.split(" ");
                 if (authParts.length == 2) {
-                    if ("Bearer".equalsIgnoreCase(authParts[0])) {
-                        requestContext.setSecurityContext(authenticateBearerToken(authParts[1]));
+                    String authType = authParts[0];
+                    String authValue = authParts[1];
+
+                    if ("Bearer".equalsIgnoreCase(authType)) {
+                        requestContext.setSecurityContext(authenticateBearerToken(authValue));
                         return;
-                    } else if ("Basic".equalsIgnoreCase(authParts[0])) {
-                        try {
-                            String[] decodedCredentials = new String(Base64.getDecoder().decode(authParts[1]), "UTF-8")
-                                    .split(":");
-                            if (decodedCredentials.length != 2) {
-                                throw new AuthenticationException("Invalid Basic authentication credential format");
-                            }
-                            if (!allowBasicAuth) {
-                                throw new AuthenticationException(
-                                        "Basic authentication with username/password is not allowed");
-                            }
-                            requestContext.setSecurityContext(
-                                    // use the base64 String as a key to the cache because it's long and already created
-                                    authenticateUsernamePassword(decodedCredentials[0], decodedCredentials[1],
-                                            authParts[1]));
-                            return;
-                        } catch (AuthenticationException e) {
-                            throw new AuthenticationException("Invalid Basic authentication credentials", e);
-                        }
+                    } else if ("Basic".equalsIgnoreCase(authType)) {
+                        requestContext.setSecurityContext(authenticateBasicAuth(authValue));
                     }
                 }
             }
--- a/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
+++ b/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
@@ -14,11 +14,10 @@ package org.openhab.core.io.rest.auth.internal;
 
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
-import java.nio.ByteBuffer;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
+import java.time.Duration;
 import java.util.Base64;
-import java.util.HashMap;
 import java.util.Map;
 import java.util.Random;
 
@@ -46,6 +45,7 @@ import org.openhab.core.io.rest.RESTConstants;
 import org.osgi.framework.Constants;
 import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Deactivate;
 import org.osgi.service.component.annotations.Modified;
 import org.osgi.service.component.annotations.Reference;
 import org.osgi.service.jaxrs.whiteboard.JaxrsWhiteboardConstants;
@@ -53,30 +53,6 @@ import org.osgi.service.jaxrs.whiteboard.propertytypes.JaxrsApplicationSelect;
 import org.osgi.service.jaxrs.whiteboard.propertytypes.JaxrsExtension;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-class SimpleUserRegistryCallback implements RegistryChangeListener<User> {
-
-    private Runnable callback;
-
-    public SimpleUserRegistryCallback(Runnable callback) {
-        this.callback = callback;
-    }
-
-    @Override
-    public void added(User element) {
-        return;
-    }
-
-    @Override
-    public void removed(User element) {
-        callback.run();
-    }
-
-    @Override
-    public void updated(User oldElement, User element) {
-        callback.run();
-    }
-}
 
 /**
  * This filter is responsible for parsing credentials provided with a request, and hydrating a {@link SecurityContext}
@@ -102,37 +78,28 @@ public class AuthFilter implements ContainerRequestFilter {
     protected static final String CONFIG_URI = "system:restauth";
     private static final String CONFIG_ALLOW_BASIC_AUTH = "allowBasicAuth";
     private static final String CONFIG_IMPLICIT_USER_ROLE = "implicitUserRole";
-
-    private static final int authCacheSize = 10;
+    private static final String CONFIG_CACHE_EXPIRATION = "cacheExpiration";
 
     private boolean allowBasicAuth = false;
     private boolean implicitUserRole = true;
+    private Long cacheExpiration = 6L;
 
-    private HashMap<ByteBuffer, UserSecurityContext> authCache = new HashMap<ByteBuffer, UserSecurityContext>();
-    private byte[] rndBytes = getRndBytes();
+    private ExpiringUserSecurityContextCache authCache = new ExpiringUserSecurityContextCache(
+            Duration.ofHours(cacheExpiration).toMillis());
 
-    private static byte[] getRndBytes() {
-        final Random rd = new Random();
-        byte[] rndBytes = new byte[32];
-        rd.nextBytes(rndBytes);
-        return rndBytes;
+    static private byte[] RANDOM_BYTES = new byte[32];
+    static {
+        new Random().nextBytes(RANDOM_BYTES);
+        ;
     }
 
-    @Nullable
-    private ByteBuffer getCacheKey(String creds) {
-        try {
-            final MessageDigest md = MessageDigest.getInstance("SHA-256");
-            md.update(rndBytes);
-            return ByteBuffer.wrap(md.digest(creds.getBytes()));
-        } catch (NoSuchAlgorithmException e) {
-            // SHA-256 is available for all java distributions so this code will actually never run
-            // If it does we'll just flood the cache with random values
-            logger.warn("SHA-256 is not available. Cache for basic auth disabled!");
-            return null;
-        }
-    }
+    @Reference
+    private JwtHelper jwtHelper;
 
-    private RegistryChangeListener<User> cacheCleanListener = new RegistryChangeListener<User>() {
+    @Reference
+    private UserRegistry userRegistry;
+
+    private RegistryChangeListener<User> userRegistryListener = new RegistryChangeListener<User>() {
 
         @Override
         public void added(User element) {
@@ -150,17 +117,10 @@ public class AuthFilter implements ContainerRequestFilter {
         }
     };
 
-    @Reference
-    private JwtHelper jwtHelper;
-
-    @Reference
-    private UserRegistry userRegistry;
-
     @Activate
     protected void activate(Map<String, Object> config) {
         modified(config);
-
-        userRegistry.addRegistryChangeListener(cacheCleanListener);
+        userRegistry.addRegistryChangeListener(userRegistryListener);
     }
 
     @Modified
@@ -170,7 +130,34 @@ public class AuthFilter implements ContainerRequestFilter {
             allowBasicAuth = value != null && "true".equals(value.toString());
             value = properties.get(CONFIG_IMPLICIT_USER_ROLE);
             implicitUserRole = value == null || !"false".equals(value.toString());
+            value = properties.get(CONFIG_CACHE_EXPIRATION);
+            if (value != null) {
+                try {
+                    cacheExpiration = Long.valueOf(value.toString());
+                } catch (NumberFormatException e) {
+                    logger.warn("Ignoring invalid configuration value '{}' for cacheExpiration parameter.", value);
+                }
+            }
             authCache.clear();
+        }
+    }
+
+    @Deactivate
+    protected void deactivate() {
+        userRegistry.removeRegistryChangeListener(userRegistryListener);
+    }
+
+    @Nullable
+    private String getCacheKey(String credentials) {
+        try {
+            final MessageDigest md = MessageDigest.getInstance("SHA-256");
+            md.update(RANDOM_BYTES);
+            return new String(md.digest(credentials.getBytes()));
+        } catch (NoSuchAlgorithmException e) {
+            // SHA-256 is available for all java distributions so this code will actually never run
+            // If it does we'll just flood the cache with random values
+            logger.warn("SHA-256 is not available. Cache for basic auth disabled!");
+            return null;
         }
     }
 
@@ -190,11 +177,7 @@ public class AuthFilter implements ContainerRequestFilter {
     }
 
     private SecurityContext authenticateBasicAuth(String credentialString) throws AuthenticationException {
-        if (!allowBasicAuth) {
-            throw new AuthenticationException("Basic authentication with username/password is not allowed");
-        }
-
-        final ByteBuffer cacheKey = getCacheKey(credentialString);
+        final String cacheKey = getCacheKey(credentialString);
         if (cacheKey != null) {
             final UserSecurityContext cachedValue = authCache.get(cacheKey);
             if (cachedValue != null) {
@@ -223,10 +206,6 @@ public class AuthFilter implements ContainerRequestFilter {
         UserSecurityContext context = new UserSecurityContext(user, auth, "Basic");
 
         if (cacheKey != null) {
-            // limit the auth cache to a certain size
-            if (authCache.size() >= authCacheSize) {
-                authCache.remove(authCache.keySet().iterator().next());
-            }
             authCache.put(cacheKey, context);
         }
 
@@ -253,15 +232,16 @@ public class AuthFilter implements ContainerRequestFilter {
                         requestContext.setSecurityContext(authenticateBearerToken(authValue));
                         return;
                     } else if ("Basic".equalsIgnoreCase(authType)) {
+                        if (!allowBasicAuth) {
+                            throw new AuthenticationException(
+                                    "Basic authentication with username/password is not allowed");
+                        }
                         requestContext.setSecurityContext(authenticateBasicAuth(authValue));
                     }
                 }
-            }
-
-            if (implicitUserRole) {
+            } else if (implicitUserRole) {
                 requestContext.setSecurityContext(new AnonymousUserSecurityContext());
             }
-
         } catch (AuthenticationException e) {
             logger.warn("Unauthorized API request: {}", e.getMessage());
             requestContext.abortWith(JSONResponse.createErrorResponse(Status.UNAUTHORIZED, "Invalid credentials"));
--- a/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
+++ b/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
@@ -13,7 +13,7 @@
 package org.openhab.core.io.rest.auth.internal;
 
 import java.io.IOException;
-import java.io.UnsupportedEncodingException;
+import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.time.Duration;
@@ -31,6 +31,7 @@ import javax.ws.rs.core.Response.Status;
 import javax.ws.rs.core.SecurityContext;
 import javax.ws.rs.ext.Provider;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.auth.Authentication;
 import org.openhab.core.auth.AuthenticationException;
@@ -61,6 +62,8 @@ import org.slf4j.LoggerFactory;
  * @author Yannick Schaus - initial contribution
  * @author Yannick Schaus - Allow basic authentication
  * @author Yannick Schaus - Add support for API tokens
+ * @author Sebastian Gerber - Add basic auth caching
+ * @author Kai Kreuzer - Add null annotations, constructor initialization
  */
 @PreMatching
 @Component(configurationPid = "org.openhab.restauth", property = Constants.SERVICE_PID + "=org.openhab.restauth")
@@ -69,6 +72,7 @@ import org.slf4j.LoggerFactory;
 @JaxrsApplicationSelect("(" + JaxrsWhiteboardConstants.JAX_RS_NAME + "=" + RESTConstants.JAX_RS_NAME + ")")
 @Priority(Priorities.AUTHENTICATION)
 @Provider
+@NonNullByDefault
 public class AuthFilter implements ContainerRequestFilter {
     private final Logger logger = LoggerFactory.getLogger(AuthFilter.class);
 
@@ -87,17 +91,10 @@ public class AuthFilter implements ContainerRequestFilter {
     private ExpiringUserSecurityContextCache authCache = new ExpiringUserSecurityContextCache(
             Duration.ofHours(cacheExpiration).toMillis());
 
-    static private byte[] RANDOM_BYTES = new byte[32];
-    static {
-        new Random().nextBytes(RANDOM_BYTES);
-        ;
-    }
+    private final byte[] RANDOM_BYTES = new byte[32];
 
-    @Reference
-    private JwtHelper jwtHelper;
-
-    @Reference
-    private UserRegistry userRegistry;
+    private final JwtHelper jwtHelper;
+    private final UserRegistry userRegistry;
 
     private RegistryChangeListener<User> userRegistryListener = new RegistryChangeListener<User>() {
 
@@ -116,6 +113,13 @@ public class AuthFilter implements ContainerRequestFilter {
             authCache.clear();
         }
     };
+
+    @Activate
+    public AuthFilter(@Reference JwtHelper jwtHelper, @Reference UserRegistry userRegistry) {
+        this.jwtHelper = jwtHelper;
+        this.userRegistry = userRegistry;
+        new Random().nextBytes(RANDOM_BYTES);
+    }
 
     @Activate
     protected void activate(Map<String, Object> config) {
@@ -148,6 +152,10 @@ public class AuthFilter implements ContainerRequestFilter {
     }
 
     private @Nullable String getCacheKey(String credentials) {
+        if (cacheExpiration == 0) {
+            // caching is disabled
+            return null;
+        }
         try {
             final MessageDigest md = MessageDigest.getInstance("SHA-256");
             md.update(RANDOM_BYTES);
@@ -184,12 +192,8 @@ public class AuthFilter implements ContainerRequestFilter {
             }
         }
 
-        String[] decodedCredentials;
-        try {
-            decodedCredentials = new String(Base64.getDecoder().decode(credentialString), StandardCharsets.UTF_8).split(":");
-        } catch (UnsupportedEncodingException e) {
-            throw new AuthenticationException("Invalid base64 credentials for basic auth");
-        }
+        String[] decodedCredentials = new String(Base64.getDecoder().decode(credentialString), StandardCharsets.UTF_8)
+                .split(":");
         if (decodedCredentials.length != 2) {
             throw new AuthenticationException("Invalid Basic authentication credential format");
         }
@@ -212,38 +216,40 @@ public class AuthFilter implements ContainerRequestFilter {
     }
 
     @Override
-    public void filter(ContainerRequestContext requestContext) throws IOException {
-        try {
-            String altTokenHeader = requestContext.getHeaderString(ALT_AUTH_HEADER);
-            if (altTokenHeader != null) {
-                requestContext.setSecurityContext(authenticateBearerToken(altTokenHeader));
-                return;
-            }
-
-            String authHeader = requestContext.getHeaderString(HttpHeaders.AUTHORIZATION);
-            if (authHeader != null) {
-                String[] authParts = authHeader.split(" ");
-                if (authParts.length == 2) {
-                    String authType = authParts[0];
-                    String authValue = authParts[1];
-
-                    if ("Bearer".equalsIgnoreCase(authType)) {
-                        requestContext.setSecurityContext(authenticateBearerToken(authValue));
-                        return;
-                    } else if ("Basic".equalsIgnoreCase(authType)) {
-                        if (!allowBasicAuth) {
-                            throw new AuthenticationException(
-                                    "Basic authentication with username/password is not allowed");
-                        }
-                        requestContext.setSecurityContext(authenticateBasicAuth(authValue));
-                    }
+    public void filter(@Nullable ContainerRequestContext requestContext) throws IOException {
+        if (requestContext != null) {
+            try {
+                String altTokenHeader = requestContext.getHeaderString(ALT_AUTH_HEADER);
+                if (altTokenHeader != null) {
+                    requestContext.setSecurityContext(authenticateBearerToken(altTokenHeader));
+                    return;
                 }
-            } else if (implicitUserRole) {
-                requestContext.setSecurityContext(new AnonymousUserSecurityContext());
+
+                String authHeader = requestContext.getHeaderString(HttpHeaders.AUTHORIZATION);
+                if (authHeader != null) {
+                    String[] authParts = authHeader.split(" ");
+                    if (authParts.length == 2) {
+                        String authType = authParts[0];
+                        String authValue = authParts[1];
+
+                        if ("Bearer".equalsIgnoreCase(authType)) {
+                            requestContext.setSecurityContext(authenticateBearerToken(authValue));
+                            return;
+                        } else if ("Basic".equalsIgnoreCase(authType)) {
+                            if (!allowBasicAuth) {
+                                throw new AuthenticationException(
+                                        "Basic authentication with username/password is not allowed");
+                            }
+                            requestContext.setSecurityContext(authenticateBasicAuth(authValue));
+                        }
+                    }
+                } else if (implicitUserRole) {
+                    requestContext.setSecurityContext(new AnonymousUserSecurityContext());
+                }
+            } catch (AuthenticationException e) {
+                logger.warn("Unauthorized API request: {}", e.getMessage());
+                requestContext.abortWith(JSONResponse.createErrorResponse(Status.UNAUTHORIZED, "Invalid credentials"));
             }
-        } catch (AuthenticationException e) {
-            logger.warn("Unauthorized API request: {}", e.getMessage());
-            requestContext.abortWith(JSONResponse.createErrorResponse(Status.UNAUTHORIZED, "Invalid credentials"));
         }
     }
 }
--- a/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
+++ b/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
@@ -231,16 +231,26 @@ public class AuthFilter implements ContainerRequestFilter {
                     if (authParts.length == 2) {
                         String authType = authParts[0];
                         String authValue = authParts[1];
-
                         if ("Bearer".equalsIgnoreCase(authType)) {
                             requestContext.setSecurityContext(authenticateBearerToken(authValue));
                             return;
                         } else if ("Basic".equalsIgnoreCase(authType)) {
-                            if (!allowBasicAuth) {
-                                throw new AuthenticationException(
-                                        "Basic authentication with username/password is not allowed");
+                            String[] decodedCredentials = new String(Base64.getDecoder().decode(authValue), "UTF-8")
+                                    .split(":");
+                            if (decodedCredentials.length > 2) {
+                                throw new AuthenticationException("Invalid Basic authentication credential format");
                             }
-                            requestContext.setSecurityContext(authenticateBasicAuth(authValue));
+                            switch (decodedCredentials.length) {
+                                case 1:
+                                    requestContext.setSecurityContext(authenticateBearerToken(decodedCredentials[0]));
+                                    break;
+                                case 2:
+                                    if (!allowBasicAuth) {
+                                        throw new AuthenticationException(
+                                                "Basic authentication with username/password is not allowed");
+                                    }
+                                    requestContext.setSecurityContext(authenticateBasicAuth(authValue));
+                            }
                         }
                     }
                 } else if (implicitUserRole) {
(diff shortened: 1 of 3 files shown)


spacemanspiff2007:
This brings OH3 close to OH2 levels (~10%)

```
Benchmark item creation
Updated 300 item definitions in 2.598s --> 115.467 updates per sec

Benchmark value update creation
Wait: 0.000
Benchmark duration: 2.009s --> 149.319 updates per sec
Pings (13): min: 9.0 max: 26.0 median: 24.0
```

Pedals2Paddles:
So this will fix logging into OH3 main UI with a username/password after using `bundle:stop org.openhab.core.io.rest.auth`?

spacemanspiff2007:
No, this will make that you don't need ``bundle:stop org.openhab.core.io.rest.auth`` any more.
Which then will make the main UI work again.

Pedals2Paddles:
Oh that's even better!  Is this a workaround for the problem, or *the* cause and solution to the problem?  I'll take either as a victory at this point of course.

Pedals2Paddles:
@spacemanspiff2007 it looks like the commits need some work to pass the auto tests for the PR.  Also, you should do a squash so it is only one single commit with the title of the first one.

spacemanspiff2007:
Imho the PR tests breaks because of something else:
```
[INFO] openHAB Core :: Integration Tests :: IO REST Core Tests SUCCESS [ 18.996 s]
[INFO] openHAB Core :: Integration Tests :: Model Core Tests FAILURE [  5.271 s]
```

spacemanspiff2007:
@wborn
You reviewed the last pr. I modified it and moved it to the correct part.
Would you mind taking another look?
Thank you

Pedals2Paddles:
Bumping for maintainer attention. I truly feel this should be considered a critical fix. Openhab is **unusable** through NGINX with this auth problem.

splatch:
Isn't issue is caused by pbkf2 passwords which needs to be calculated for each and every request?

IMHO Place which gets fixed by this commit is where you see symptoms but not where root cause is.

spacemanspiff2007:
This commit adds a cache so the password hashes don't have to be calculated for every request any more.
The root cause for the performance issue is the long and multiple hash times but I don't know how and if this should be tackled,

@splatch 
I am not sure I understand correctly what you are trying to say. Could you elaborate a little bit further?

splatch:
> I am not sure I understand correctly what you are trying to say. Could you elaborate a little bit further?

My suggestion is to consider that functionality into ManagedUserProvider/Registry which does password checks. This is specific implementation which is affected by hashing algo.

spacemanspiff2007:
Yes - that would be even better. 
However maybe let's get this merged first so OH3 is usable again and then we can tackle the proper implementation.
Since I am not a Java developer it took me quite some time to find this implementation and I'd like to migrate to OH3, too so I can provide better support for the HABApp users.

Pedals2Paddles:
Well according to the maintainer for openhabion, there's already a solution for this. But he's currently refusing to tell anyone what it is out of spite. I hope someone else can give this some attention to resolve the problem

spacemanspiff2007:
@kaikreuzer 
Would you mind taking a look at this? It is a really small change which imho is good to get merged.

kaikreuzer:
@spacemanspiff2007 Not sure why you think that I would be the best to look into it - there are other (and imho better) experts that already expressed their opinion [here](https://github.com/openhab/openhab-core/pull/2030#discussion_r550916552), [here](https://github.com/openhab/openhab-core/pull/2030#issuecomment-757576201) and [here](https://github.com/openhab/openhab-core/pull/2101#issuecomment-762514051). Please don't expect me to overrule their verdict in any way.

Besides this, I usually only start looking at PRs once they have some content in their description and pass the automatic checks.

spacemanspiff2007:
@kaikreuzer 
I have addressed both of the first two concerns with this commit and they not longer apply.
These are not the same changes and I should have probably named this PR different than the first one.
I just want OH3 to work with basic auth again an my concern is that this PR somehow does not make it into OH3,1.


> Besides this, I usually only start looking at PRs once they have some content in their description and pass the automatic checks.

Fair enough - I added a small description what this does. 
However I have signed the DCO and all commits properly but the checks seem to fail because of the git privacy guard.
Also the automatic build failed (imho) because of another PR (``Update Mockito, JUnit and use MockitoExtensions in itests``) and not because of these changes. I do not know how I can trigger a build that only includes the changes from this PR.

kaikreuzer:
> I have addressed both of the first two concerns with this commit and they not longer apply.

Is this confirmed by @wborn and @ghys? Sorry, I didn't follow in all details, but I I saw @ghys summary "tl;dr: **use API Tokens.**"
Did you consider this advice? Is there anything wrong with it? Unless there's indeed an issue, I'd also think that API tokens are the proper way to have external systems authenticate themselves against openHAB. And in the case of an NGINX as reverse-proxy, it should also be simple to add an API token to the HTTP header when passing the request to the proxy instead of turning off auth for admin endpoints altogether.

kaikreuzer:
@spacemanspiff2007 And please don't get me wrong: I don't want to block you here and I am honestly looking for a solution - but I have learned to (a) listen to the experts, (b) don't mess with cryptography and (c) to really need to understand the real issue that is being tried to solve in order to think out of the box to find the best solution and not just the nearest.

Pedals2Paddles:
Hey guys.  Looks like a cause and solution has been identified.  [See this thread starting here.](https://community.openhab.org/t/oh3-with-nginx-reverse-proxy-and-authentication/106528/48?u=pedals2paddles)  There is a misplaced directive in the NGINX configuration file causing this mess.  Once you move the directive down to the location/ block where it belongs, everything works beautifully. 

My configuration had the `proxy_set_header Authorization ""` directive up in the top section, not under the location/ block. This is where the OpenHabian-config tool put it twice, and where it seems to have put it for everyone else having this problem.  It is also not in the documentation on the web (which I know needs to be updated) so unlikely anyone would know or notice something is out of place.

I moved the directive down to the location/ block as suggested by ysc and mstormi. Everything now seems to operate with lightning efficiency. The CPU is no longer melting down when trying to login. It actually seems just as fast as when connected locally not through NGINX. So this certainly seems to have been the fix we’ve all been looking for.

If this takes care of it for @spacemanspiff2007 and the others having this issue, I believe this PR and associated issues can be closed with thanks.

spacemanspiff2007:
@kaikreuzer 
> I'd also think that API tokens are the proper way to have external systems authenticate themselves against openHAB.

I have made mutual factual arguments (and linked 3rd party articles by experts that do this stuff for a living) that the assumption that API keys are more secure than basic is just wrong. Additionally the recommendation by the expert was that if you don't do delegation (e.g. letting a 3rd party do the authentication) API keys should ***not*** be used for authentication.

Also I think if openhab provides a method for authentication (e.g. basic auth) it should work adequately and not only if I send one request per day.

The argument against the first implementation was that the password/user is stored in plain text and thus leaked if someone shares a memory dump in the community (note that this concern does not target the implementation of a cache).
Imho this is a weak argument because in the things there are often credentials for 3rd party web services in plain text which are more likely and easier to be abused than some openhab credentials for a local installation.
But this is something that can easily be changed and so I did.


> And please don't get me wrong: I don't want to block you here and I am honestly looking for a solution - 

I know you mean well but this whole contribution experience has been nothing but extremely frustrating for me.
Sure my setup is not the most common one but all my issues can be measured objectively and reproducibly and I addressed them in a timely matter.  But let's move forward from that.

> but I have learned to (a) listen to the experts, (b) don't mess with cryptography and (c) to really need to understand the real issue that is being tried to solve in order to think out of the box to find the best solution and not just the nearest.

(a)
Maybe I am an expert, too and you should listen to me! :wink: 

(b)
This just adds a small cache so that the very expensive hashes for send in passwords don't have to be calculated every time. So for a request with a unknown password everything behaves like it was before. This just speeds up things for consequent requests with the same password (which will be allowed or rejected) so it changes nothing for the underlying implementation.

(c)
I hope you now have a better understanding what is going on

@Pedals2Paddles 
This does not solve the issue and the pull request can not be closed

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -75,6 +80,19 @@
     private boolean allowBasicAuth = false;
     private boolean implicitUserRole = true;
 
+    private static HashMap<ByteBuffer, UserSecurityContext> authCache = new HashMap<ByteBuffer, UserSecurityContext>();
+    private static final String UUID_STRING = UUID.randomUUID().toString();
```
spacemanspiff2007, line 84: one could remove the '-' from the UUID because they don't add anything to the entropy

spacemanspiff2007:
I have updated my benchmark. The benchmark can be run through HABApp if it is started with the ``--benchmark`` argument.
With this pr we see a > 20x improvement in performance and a >22x improvement in response time.

Current implementation
```text
Bench item operations ... done!

            |  dur   | per sec | median |  min   |  max   |  mean 
create item | 43.4s |   6.907 | 0.144s | 0.140s | 0.188s | 0.145s
update item | 43.3s |   6.931 | 0.144s | 0.140s | 0.198s | 0.144s
delete item | 42.7s |   7.029 | 0.141s | 0.138s | 0.195s | 0.142s

Bench item state update .... done!

                      |  dur   | per sec | median |  min   |  max   |  mean 
             rtt idle | 5.150s |   6.990 | 0.140s | 0.139s | 0.180s | 0.143s
       async rtt idle | 5.026s |   6.963 | 0.141s | 0.141s | 0.196s | 0.144s
      rtt load (+10x) | 5.022s |   4.181 | 0.236s | 0.213s | 0.284s | 0.239s
async rtt load (+10x) | 5.070s |   4.142 | 0.244s | 0.224s | 0.250s | 0.241s

Cleanup ... complete
```

With applied fix from this PR
```text
Bench item operations ... done!

            |  dur   | per sec | median |  min   |  max   |  mean 
create item | 2.557s | 117.318 | 8.00ms | 7.00ms | 20.0ms | 8.52ms
update item | 2.425s | 123.704 | 8.00ms | 6.00ms | 33.0ms | 8.08ms
delete item | 1.807s | 166.012 | 6.00ms | 5.00ms | 26.0ms | 6.02ms

Bench item state update .... done!

                      |  dur   | per sec | median |  min   |  max   |  mean 
             rtt idle | 4.994s | 149.971 | 7.00ms | 6.00ms | 12.0ms | 6.67ms
       async rtt idle | 5.003s | 133.712 | 7.00ms | 7.00ms | 14.0ms | 7.48ms
      rtt load (+10x) | 5.048s |  17.828 | 56.0ms | 46.0ms | 69.0ms | 56.1ms
async rtt load (+10x) | 5.021s |  12.148 | 82.0ms | 73.0ms | 95.0ms | 82.3ms

Cleanup ... complete
```

spacemanspiff2007:
@kaikreuzer 
Can we get this merged? This is a really small change that does not touch the underlying cryptographic implementation.
There are objectively no arguments not to merge this but very strong arguments to merge this (e.g. > 20x speed improvement, prevents rest api server from disconnecting, makes basic auth usable again).

I can't find any explanation why this has not yet been merged and imho any other pr with a 20x speed improvement would already have been included with flying colors.
It's really mind-boggling and the only explanation I've come up with so far is that basic auth is intentionally broken because it's not the preference of some of the devs. While I've strongly rejected this explanation so far it's really starting to creep on me since it's been a month since I opened the first PR and for the most part I've either gotten feedback that does not deal with the content of the PR or no feedback at all.

wborn:
> the only explanation I've come up with so far 

Usually there is a very simple explanation. Lack of time and different priorities. Each of the core maintainers have a big todo list and also maintains several repos. Besides that they also have a day job and non-openHAB obligations screaming for attention. :wink:

kaikreuzer:
@spacemanspiff2007 I am not sure why you expect ME to merge it - I am not the one to decide here, but in general, it is up to maintainers and in this specific case, where all the code that is changed by this PR was written by @ghys, I'd seek his approval as well as mentioned [above](https://github.com/openhab/openhab-core/pull/2101#issuecomment-767131743):

> Is this confirmed by @wborn and @ghys?

I don't see this answered and don't see their approval here, what is for me a clear signal that it isn't at all clear that there are "no arguments not to merge".

spacemanspiff2007:
> Besides that they also have a day job and non-openHAB obligations screaming for attention. wink

@wborn I feel you. We're all volunteers and pour our precious time and effort into this and many other projects.
I hope you did think this was directed at you because you reviewed my first pr and actually tried to help fix this issue.

> @spacemanspiff2007 I am not sure why you expect ME to merge it - I am not the one to decide here, but in general, it is up to maintainers and in this specific case, where all the code that is changed by this PR was written by @ghys, I'd seek his approval as well as mentioned [above](https://github.com/openhab/openhab-core/pull/2101#issuecomment-767131743):

@kaikreuzer 
I don't expect you to blindly merge it, I'm pining you because I just want some kind of progress and I'm hoping that you see the "low effort high reward" of this PR. This PR would fix (and would have fixed) the reverse proxy issues and make basic auth actually usable. All I wanted/want is that you put this (higher) on someones radar by e.g. "Hey @AwesomeOpenhabMaintainer - this looks promising! Would have have a look?".

> I don't see this answered and don't see their approval here, what is for me a clear signal that it isn't at all clear that there are "no arguments not to merge".

@ghs does not care about basic auth at all and will only say "use token based authentication".
If we have to wait for his approval to fix something that is clearly broken I might as well close this PR because it'll never happen since he obviously is more than biased.

spacemanspiff2007:
I disabled the github privacy guard especially for the commit & push, but somehow it redacted the mail again.
Could someone of the maintainers manually pass the DCO again.

Also I really hoped this would have made it into 3.1. 
*sigh*

spacemanspiff2007:
@wborn Would you mind starting a review? Imho this looks good to get merged.

spacemanspiff2007:
@kaikreuzer 
How shall we proceed here? 
I've created the fix and I've pinged all the maintainers. The change is 50 lines of codes and should be easy to review.
What more can I do?

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -111,15 +139,33 @@ private SecurityContext authenticateBearerToken(String token) throws Authenticat
...
         UsernamePasswordCredentials credentials = new UsernamePasswordCredentials(username, password);
         Authentication auth = userRegistry.authenticate(credentials);
         User user = userRegistry.get(auth.getUsername());
         if (user == null) {
             throw new AuthenticationException("User not found in registry");
         }
-        return new UserSecurityContext(user, auth, "Basic");
+
+        // limit cache to a certain size
+        if (AuthFilter.authCache.size() > 20) {
```
kaikreuzer, line 158: Magic numbers like "20" should be defined as a constant.
20 feels pretty high for me as well.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -111,15 +139,33 @@ private SecurityContext authenticateBearerToken(String token) throws Authenticat
...
+            Object remName = null;
+            for (Object obj : AuthFilter.authCache.keySet()) {
```
kaikreuzer, line 160: I don't understand the logic here. What's the use of a for loop that immediately breaks in the first iteration?

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -142,22 +188,17 @@ public void filter(ContainerRequestContext requestContext) throws IOException {
                         try {
                             String[] decodedCredentials = new String(Base64.getDecoder().decode(authParts[1]), "UTF-8")
                                     .split(":");
-                            if (decodedCredentials.length > 2) {
+                            if (decodedCredentials.length != 2) {
                                 throw new AuthenticationException("Invalid Basic authentication credential format");
                             }
-                            switch (decodedCredentials.length) {
-                                case 1:
```
kaikreuzer, line 149: Why has this case been removed? How is it covered now? I cannot see it anymore.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -111,15 +139,33 @@ private SecurityContext authenticateBearerToken(String token) throws Authenticat
         }
     }
 
-    private SecurityContext authenticateUsernamePassword(String username, String password)
+    private SecurityContext authenticateUsernamePassword(String username, String password, String authCacheKey)
```
kaikreuzer, line 142: `authCacheKey` isn't a cache key, it is the unencrypted credentials. The parameter name is hence dangerously misleading.

kaikreuzer: Besides this, I do not see any need for this third parameter, since it holds no additional information over the first two.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -111,15 +139,33 @@ private SecurityContext authenticateBearerToken(String token) throws Authenticat
...
             throws AuthenticationException {
+        final ByteBuffer cacheKey = getCacheKey(authCacheKey);
+        final UserSecurityContext cachedValue = AuthFilter.authCache.get(cacheKey);
+        if (cachedValue != null) {
+            return cachedValue;
+        }
+
         UsernamePasswordCredentials credentials = new UsernamePasswordCredentials(username, password);
         Authentication auth = userRegistry.authenticate(credentials);
```
kaikreuzer, line 151: So if there was some entry in the cache, we fully skip the authentication? Same question as above: How do you ensure that the cache entry is still valid? How are entries invalidated if changes to the userRegistry are done?

kaikreuzer (changes requested):
@spacemanspiff2007 As mentioned before, I don't feel to be the right person to review this as I haven't written the code and didn't find the time to dive into the topic in depths.

Since there wasn't any feedback from others, I tried to do a review nonetheless. I might be missing some underlying thoughts, but reviewing the code rather increased my doubts that this should be merged. But as I said, I'm not the expert here, so I am fully ok if others (like @ghys as the author of the original code or @openhab/core-maintainers) feel good about it and give a merge approval.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -75,6 +80,28 @@
...
+    private static byte[] rndBytes = getRndBytes();
+
+    private static byte[] getRndBytes() {
+        final Random rd = new Random();
+        byte[] rndBytes = new byte[32];
+        rd.nextBytes(rndBytes);
+        return rndBytes;
+    }
+
+    private static ByteBuffer getCacheKey(String creds) {
```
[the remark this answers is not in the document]
spacemanspiff2007: The digest of the hash function comes back as bytes, that's why I've wrapped a buffer object around it.
Imho it doesn't matter and I can cast to String but with ByteBuffer it's the native data type.
If you want to I can change to String

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -75,6 +80,28 @@
...
+        try {
+            final MessageDigest md = MessageDigest.getInstance("SHA-256");
+            md.update(AuthFilter.rndBytes);
+            return ByteBuffer.wrap(md.digest(creds.getBytes()));
+        } catch (NoSuchAlgorithmException e) {
+            // SHA-256 is available for all java distributions so this code will actually never run
+            // If it does we'll just flood the cache with random values
```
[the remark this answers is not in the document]
spacemanspiff2007: SHA-256 is part of the java spec so it's available everywhere but the IDE told me to catch the NoSuchAlgorithmException exception.
What would be a better way to resolve this?

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
spacemanspiff2007: It's limiting the cache size to 20 entries by removing an object from the cache

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
spacemanspiff2007: This was old legacy code that was not correct any more. It's trying to mix token based auth with basic auth.
It's either one or the other: If the header is basic auth it should be basic auth, if the header specifies token it should use token.
But specifying basic auth and trying to do token based auth is wrong and not necessary any more if basic auth works fine.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
spacemanspiff2007: But why split the String and then join it back together - this seems like unnecessary overhead?
If I pass it into the function I can just hash it the way it is.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
spacemanspiff2007: Okay - sure thing. 
About the size: keep in mind the cache can also contain invalid auth attempts.
What would be a better number? 10?

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -75,6 +80,28 @@
     private boolean allowBasicAuth = false;
     private boolean implicitUserRole = true;
 
+    private static HashMap<ByteBuffer, UserSecurityContext> authCache = new HashMap<ByteBuffer, UserSecurityContext>();
```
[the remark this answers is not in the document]
spacemanspiff2007: It doesn't. Would you prefer non static?

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -111,15 +139,33 @@ private SecurityContext authenticateBearerToken(String token) throws Authenticat
         }
     }
 
-    private SecurityContext authenticateUsernamePassword(String username, String password)
+    private SecurityContext authenticateUsernamePassword(String username, String password, String authCacheKey)
             throws AuthenticationException {
+        final ByteBuffer cacheKey = getCacheKey(authCacheKey);
+        final UserSecurityContext cachedValue = AuthFilter.authCache.get(cacheKey);
+        if (cachedValue != null) {
```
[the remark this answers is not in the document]
spacemanspiff2007: It's currently linked to the activation of basic auth in the settings.
Disabling basic auth will clear the cache.
Any user change should also just clear the cache, but I am not sure how to achieve that

spacemanspiff2007:
> Since there wasn't any feedback from others, I tried to do a review nonetheless. I might be missing some underlying thoughts, but reviewing the code rather increased my doubts that this should be merged. But as I said, I'm not the expert here, so I am fully ok if others (like @ghys as the author of the original code or @openhab/core-maintainers) feel good about it and give a merge approval.

Thank you for your effort and your review.
I am not a native Java coder so there are sure some things that can and have to be implemented in a better way.
The one thing that is missing from this PR is the link from a user change in the registry to a cache clear.
I am unsure how this can be achieved and it would be really nice if you or any one else could provide a hint.

If after this has been implemented and you'll still have doubts about merging this I'd really like some guidance how to proceed.
I don't want to put in lots of effort just for the PR not to be merged.
However my application is getting ~8k downloads/month and while only a fraction are real users they keep asking me when everything will work with OH3 properly again.
I'm using a library which doesn't support the token and the last change there took ~9 month to get merged so the developing pace is rather slow.

So what do I tell them? Don't migrate to OH3? That sure can't be the solution ...

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
spacemanspiff2007: What would be a better name for the parameter? It gets hashed and used as the key ...

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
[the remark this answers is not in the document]
kaikreuzer: Definitely so.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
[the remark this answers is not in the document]
kaikreuzer: As it is part of a method signature, I'd prefer String, yes.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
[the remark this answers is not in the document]
kaikreuzer: Return null, so that there is no cache key and hence no cache is used.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
kaikreuzer: Maybe the method should only receive the BASE64 credential string as a param - that would be simplest.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
kaikreuzer: Ok, good point, I didn't think of the invalid attempts. But yeah, 10 should be enough, I'd say.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
kaikreuzer: You do not need a for loop for that.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
kaikreuzer: How can I see that it is legacy code? Which class deals with the token header now? This isn't clear for me from the code. Anyhow, as it is unrelated to what you want to do in the PR, better leave the change out.

kaikreuzer:
> I am not a native Java coder so there are sure some things that can and have to be implemented in a better way.

Understood and much appreciated that you try to do it anyway. I'm willing to help improving the code, but unfortunately won't be able to do so in the coming days.

> The one thing that is missing from this PR is the link from a user change in the registry to a cache clear.
> I am unsure how this can be achieved and it would be really nice if you or any one else could provide a hint.

An easy option might be to simply use an expiring cache (as I had suggested [here](https://github.com/openhab/openhab-core/pull/2101#discussion_r584098693)) - if entries are e.g. only valid for a minute, it will soon enough pick up any changes that are done in the user registry, while still giving you the desired performance on high-frequent subsequent calls.

> I don't want to put in lots of effort just for the PR not to be merged.

Understood, but here are some good news. Diving into the topic, I came across https://github.com/traefik/traefik/issues/7897, which is a very same feature request on Traefik with a very similar suggested solution. The discussion there convinced me that it should be ok to introduce it - my concern that we might do something nasty security-wise in openHAB is addressed by this and I approve the feature. So let's just get the code straight and I think we can merge it. Please excuse my bad availability in the coming days, though.

spacemanspiff2007:
> An easy option might be to simply use an expiring cache (as I had suggested here) - if entries are e.g. only valid for a minute, it will soon enough pick up any changes that are done in the user registry, while still giving you the desired performance on high-frequent subsequent calls.

I understand, but I still think the performance impact is noticeable. On the raspberry pi processing the auth blocks for ~0.3 secs.
This is quite noticeable especially when doing an automation (e.g. turning a light on based on movement).
Is there any possibility to trigger on an event/register a callback that gets emitted when a user gets changed/added/removed.

> Please excuse my bad availability in the coming days, though.

No problem, as long as something is/will be happening it's fine with me.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -84,6 +159,8 @@
     @Activate
     protected void activate(Map<String, Object> config) {
         modified(config);
+
+        userRegistry.addRegistryChangeListener(cacheCleanListener);
```
spacemanspiff2007, line 163: @kaikreuzer am I doing this right? Does "activate" only get called once?

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -72,9 +103,53 @@
...
+        rd.nextBytes(rndBytes);
+        return rndBytes;
+    }
+
+    @Nullable
+    private ByteBuffer getCacheKey(String creds) {
+        try {
+            final MessageDigest md = MessageDigest.getInstance("SHA-256");
+            md.update(rndBytes);
+            return ByteBuffer.wrap(md.digest(creds.getBytes()));
```
spacemanspiff2007, line 126: If I want to cast to String I'll have do encode the returned ``byte[]`` into an hex string.
I'd have to use/create a byte -> hexstr function which only adds overhead so if it's possible I'd like to keep the byte structure

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -111,15 +189,48 @@ private SecurityContext authenticateBearerToken(String token) throws Authenticat
         }
     }
 
-    private SecurityContext authenticateUsernamePassword(String username, String password)
-            throws AuthenticationException {
-        UsernamePasswordCredentials credentials = new UsernamePasswordCredentials(username, password);
+    private SecurityContext authenticateBasicAuth(String credentialString) throws AuthenticationException {
```
spacemanspiff2007, line 192: I've moved all basic auth handling into this function as suggested

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -111,15 +189,48 @@ private SecurityContext authenticateBearerToken(String token) throws Authenticat
...
             throw new AuthenticationException("User not found in registry");
         }
-        return new UserSecurityContext(user, auth, "Basic");
+
+        UserSecurityContext context = new UserSecurityContext(user, auth, "Basic");
+
+        if (cacheKey != null) {
+            // limit the auth cache to a certain size
+            if (authCache.size() >= authCacheSize) {
+                authCache.remove(authCache.keySet().iterator().next());
```
spacemanspiff2007, line 228: Is this better than the for loop?

spacemanspiff2007:
This is a ping for @kaikreuzer so this issue is back on your radar and we can continue the review.

kaikreuzer (approved):

kaikreuzer:
@wborn I have extensively looked into this and also contributed code. I'm fine with the feature and implementation now, but as it contains code from myself, I'd like to ask you for a quick review as well. Thanks!

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
wborn: This change will probably break existing setups as the current docs also state you can use tokens with basic authentication:

https://www.openhab.org/docs/configuration/restdocs.html#authentication

It maybe wrong it seems convenient. This way you can probably also easily test/use these tokens in your browser or in code by adding it in the URL as https://token@host/...

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -71,19 +78,49 @@
...
     private boolean allowBasicAuth = false;
     private boolean implicitUserRole = true;
+    private Long cacheExpiration = 6L;
+
+    private ExpiringUserSecurityContextCache authCache = new ExpiringUserSecurityContextCache(
+            Duration.ofHours(cacheExpiration).toMillis());
+
+    static private byte[] RANDOM_BYTES = new byte[32];
+    static {
+        new Random().nextBytes(RANDOM_BYTES);
```
wborn, line 92: If you change `RANDOM_BYTES` in to a non-static variable it can be initialized in the constructor.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -93,6 +130,34 @@ protected void modified(@Nullable Map<String, Object> properties) {
...
+        }
+    }
+
+    @Deactivate
+    protected void deactivate() {
+        userRegistry.removeRegistryChangeListener(userRegistryListener);
+    }
+
+    @Nullable
+    private String getCacheKey(String credentials) {
```
wborn, line 151:
```suggestion
    private @Nullable String getCacheKey(String credentials) {
```

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -111,15 +176,40 @@ private SecurityContext authenticateBearerToken(String token) throws Authenticat
...
+        if (cacheKey != null) {
+            final UserSecurityContext cachedValue = authCache.get(cacheKey);
+            if (cachedValue != null) {
+                return cachedValue;
+            }
+        }
+
+        String[] decodedCredentials;
+        try {
+            decodedCredentials = new String(Base64.getDecoder().decode(credentialString), "UTF-8").split(":");
```
wborn, line 190:
```suggestion
            decodedCredentials = new String(Base64.getDecoder().decode(credentialString), StandardCharsets.UTF_8).split(":");
```

bundles/org.openhab.core.io.rest.auth/src/main/resources/OH-INF/config/config.xml
```diff
@@ -11,6 +11,17 @@
 			<description>Allow the use of Basic authentication to access protected API resources, in addition to access tokens
 				and API tokens.</description>
 		</parameter>
+		<parameter name="cacheExpiration" type="integer">
```
wborn, line 14: I think it should also have a minimum ` min="0" ` ? Is it also possible to disable this cache by using "0" ?

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
wborn: Another way to make it less wrong but keep the convenient way, is by allowing such tokens to be passed in a parameter instead of a header. E.g. while integrating [Keycloak](https://www.keycloak.org/) in another project, I've seen it also allows for passing tokens with `https://host?access_token=...` . But it would still be a breaking change.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
kaikreuzer: Right, by I preferred to initialize it statically. Is that an issue?

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
kaikreuzer: I'd claim that the docs are simply wrong or at least an ugly workaround. Passing a bearer token [should be done this way](https://stackoverflow.com/a/25036577):
```
curl -H "Authorization: Bearer <TOKEN>" http://www.example.com
````
I'm happy to provide a PR to update the documentation, if you agree with it.

File: bundles/org.openhab.core.io.rest.auth/src/main/resources/OH-INF/config/config.xml (hunk 18 above)
Comment (kaikreuzer) [reply bundles/org.openhab.core.io.rest.auth/src/main/resources/OH-INF/config/config.xml]: I've added `min="1"` (since it is not possible to deactivate the cache (on purpose)).

Comment (kaikreuzer) [reply bundles/org.openhab.core.io.rest.auth/src/main/resources/OH-INF/config/config.xml]: You think we should add a deactivation? Value 0 might be misleading as @spacemanspiff2007 interpreted it as "cache does not expire".

Comment (wborn) [reply bundles/org.openhab.core.io.rest.auth/src/main/resources/OH-INF/config/config.xml]: If it allows for disabling the cache without adding much other code it might be a nice addition for those concerned about security. The behavior should then be documented in the description.

Note that message digest algorithms like SHA-256 are also considered unsecure nowadays. That's because such hashes are relatively easy to compute and there is dedicated hardware that simplifies brute forcing them. E.g. classes to generate such hashes are deprecated in [Spring Security](https://github.com/spring-projects/spring-security/blob/aad76e68a2427ae525bc31dbb5ebe02905d6b6d3/crypto/src/main/java/org/springframework/security/crypto/password/MessageDigestPasswordEncoder.java#L78-L82).

File: bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java (hunk 4 above)
Comment (wborn) [reply bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java:149]: I'm fine with dropping support for this when the docs are updated and it is mentioned in the release notes.

File: bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java (hunk 15 above)
Comment (wborn) [reply bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java:92]: It will probably work, but I ran into a nasty classloader issue due to static variables yesterday, so I'm a currently a bit less enthusiastic about seeing them being used. :wink:

File: bundles/org.openhab.core.io.rest.auth/src/main/resources/OH-INF/config/config.xml (hunk 18 above)
Comment (spacemanspiff2007) [reply bundles/org.openhab.core.io.rest.auth/src/main/resources/OH-INF/config/config.xml]: I'm aware that SHA-256 is a deprecated algorithm to **store** passwords.
But this is an in memory cache and get access to the hashed pw an attacker already needs to have memory access and/or code execution privileges.
If an attacker already has that, then imho brute forcing a basic auth hash is the least of the worries.

File: bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java (hunk 4 above)
Comment (ghys) [reply bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java:149]: > I've seen it also allows for passing tokens with https://host?access_token=...

Strongly advise against this. Long-lived tokens should never be part of any URL because of the risk they might end in access logs.

And **please** do not remove the "passing an API token as basic auth creds" since it helps in situations where passing it in another way is impossible. See the Node-Red example in the docs. It is very much not "legacy", GitHub (and others) is doing it albeit slightly differently (https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token#using-a-token-on-the-command-line) and, by the way, they deprecated password authorization to the API (https://docs.github.com/en/rest/overview/other-authentication-methods#via-username-and-password), which was the right thing to do.

Comment (wborn) [reply bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java:149]: I agree using tokens is the way forward so it is quite useful to also have a way to use them with applications that cannot easily add add the right header. It's not an improvement if we would again force our users to create users with passwords and enable basic authentication to use the API with such applications.

File: bundles/org.openhab.core.io.rest.auth/src/main/resources/OH-INF/config/config.xml (hunk 18 above)
Comment (kaikreuzer) [reply bundles/org.openhab.core.io.rest.auth/src/main/resources/OH-INF/config/config.xml]: @wborn SHA-256 is secure, the link to Spring Security just says that it isn't recommended for password hashes, since it is too fast to calculate - but this is exactly what we try to avoid here: Doing expensive calculations for users that already authenticated successfully. In our situation, there is no risk of any dictionary attacks here. See also https://github.com/traefik/traefik/issues/7897#issuecomment-780755004, which does the very same (with SHA-512 instead of SHA-256, but with the same arguments).

File: bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java (hunk 15 above)
Comment (kaikreuzer) [reply bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java]: Ok, changed it!

File: bundles/org.openhab.core.io.rest.auth/src/main/resources/OH-INF/config/config.xml (hunk 18 above)
Comment (kaikreuzer) [reply bundles/org.openhab.core.io.rest.auth/src/main/resources/OH-INF/config/config.xml]: I've added `0` as a valid configuration value to disable the cache.

File: bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java (hunk 4 above)
Comment (kaikreuzer) [reply bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java:149]: @ghys I'd still say that the docs should be changed to clearly say that tokens are supposed to be passed as headers and to only keep the Basic Auth workaround as a last resort.
But also there, I don't like that it is `<token>:` for `<username>:<password>`. It would be much cleaner to require it the same way as Github, i.e. as `<username>:<token>`. Our tokens are always associated to users, aren't they?

Comment (ghys) [reply bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java:149]: > @ghys I'd still say that the docs should be changed to clearly say that tokens are supposed to be passed as headers and to only keep the "token passing as Basic Auth" workaround as a last resort.

Absolutely. I overlooked that part.

> But also there, I don't like that it is <token>: for <username>:<password>. It would be much cleaner to require it the same way as Github, i.e. as <username>:<token>. Our tokens are always associated to users, aren't they?

Agreed as well, this very point was discussed in the forums a while ago but I can't seem to find it anymore - in hindsight this should have been done like this. Maybe keep the old way too (pass the API token as username) at least temporarily for retrocompatibility's sake? So the recommended way to access the API would be, in order:
1. `Authorization: Bearer <JWT token>` (i.e. prefer OAuth is possible - web clients, native clients able to spawn a webview for OAuth authorization...)
2. `Authorization: Bearer <API token>`
3. `X-OPENHAB-TOKEN: <JWT token or API token>` (if the Authorization header is already taken for authentication to a reverse proxy)
4. `Authorization: Basic base64(<username>:<API token>)` (alternative to already implemented 5. below; if the client only supports Basic auth)
5. `Authorization: Basic base64(<API token>:)` (API token as username and blank password, the currently implemented option; maybe don't document it so usage will drop)
6. `Authorization: Basic base64(<username>:<password>`) **(DISCOURAGED and needs the option activated in Settings > API Security)**.

Note that the "Allow Basic authorization" option should perhaps be renamed to "Allow password-based Basic authentication", if you don't want to remove it, since it could be assumed that 4./5. require this option to work when they don't. Maybe clearly indicate that it's not recommended to use that option, since we've already seen what it does with improperly configured proxies forwarding their own Basic auth to openHAB - in some cases it will try to validate the proxy's password and this will slow the UI to a crawl.

Comment (spacemanspiff2007) [reply bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java:149]: > Maybe clearly indicate that it's not recommended to use that option, since we've already seen what it does with improperly configured proxies forwarding their own Basic auth to openHAB - in some cases it will try to validate the proxy's password and this will slow the UI to a crawl.

This PR fixes that issue.

> by the way, they deprecated password authorization to the API (https://docs.github.com/en/rest/overview/other-authentication-methods#via-username-and-password), which was the right thing to do.

A web facing service with 40 million users should definitely have increased security over something that runs locally in your network because it has a much higher attack surface. But since openhab doesn't enforce ssl certifactes what we're doing is all just smoke and mirrors anyway. Basic Auth is just as secure as long living tokens except for the password leakage and there have been no constructive arguments that prove otherwise.

> And please do not remove the "passing an API token as basic auth creds" since it helps in situations where passing it in another way is impossible.

Please do, this is nothing more than a very bad hack. If the header specifies basic auth, that's what it should be!
What's next - a POST request that deletes items, a GET request that sets values (the one from classic UI was removed with this exact argument). If we do this, then why use standards at all?

@ghys 
Why do you get to make the decision about API security and not the user?
If I have openhab running as a microservice with an authenticating reverse proxy with an ssl connection, why do I have to pass credentials to openhab which then validates it again. Why can't I just pass a user and oh will accept it without further performance penalty.
I'm the last person to say we don't need to secure things but it has to be in such a way that it
- works for many use cases
- works with good performance
- can be configured by the user


@kaikreuzer @ghys @wborn 
There is definitely a need to interact with openhab in a very fast way from other programs.
Maybe the REST API is not the right tool for that after all.
HomeAssistant and NodeRed also offer websockets as an in/output. Authentication is done **once** when the websocket connects and thus the performance impact is negligible (That also the reason there was no "drama" when HomeAssistant switched to token based auth because it did not break anything performance wise).
Maybe that way is worth investigating?

Comment (ghys) [reply bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java:149]: > Why do you get to make the decision about API security and not the user?
If I have openhab running as a microservice with an authenticating reverse proxy with an ssl connection, why do I have to pass credentials to openhab

As said before, if you want to run openhab as a microservice and roll your own auth, you're not in the "standard" supported use case and should modify the Karaf config or fork the distro to disable the built-in auth bundle. The UI was fixed to work in these cases. If you want the built-in auth, and most people do, then you play by its rules and avoid basic auth.

Comment (wborn) [reply bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java:149]: > if you don't want to remove it, since it could be assumed that 4./5. require this option to work when they don't. Maybe clearly indicate that it's not recommended to use that option, since we've already seen what it does with improperly configured proxies forwarding their own Basic auth to openHAB

I think we should also mention it is deprecated and will be removed eventually so users stay clear of it and migrate to tokens. With 4. and 5. in place what valid reasons are there to cling to 6.? It's a bad idea to recycle the same user credentials in an application using the API.

Comment (ghys) [reply bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java:149]: So to recap:

- You've come out of the blue [advocating](https://github.com/openhab/openhab-core/issues/1699) for the removal of my spare-time work on OAuth2, instead of finding solutions, because it didn't fit YOUR use case
- I recognized it was not ideal (despite its benefits) at the time and quickly added first basic auth (https://github.com/openhab/openhab-core/pull/1713) (which I immediately thought was a mistake) and then one week later API tokens (https://github.com/openhab/openhab-core/pull/1735) mostly (but of course not only) for YOU
- I researched how to disable auth from an existing instance (https://github.com/openhab/openhab-core/issues/2060#issuecomment-757392996), commented on why it shouldn't be easy to do especially remotely, and also fixed the UI (https://github.com/openhab/openhab-webui/pull/812) to allow it to work
- I tried my best to explain to you why passwords were deliberately slow to validate and tokens were not (https://github.com/openhab/openhab-core/pull/2030#issuecomment-757576201); I made the point that tokens were a better fit for API usage and using them is faster and keeps the password verifications to a minimum (not to mention storage/transport); you didn't even acknowledge that point and quoted me out of context (https://github.com/openhab/openhab-core/pull/2030#issuecomment-757651574) - I hope not deliberately
- you repeatedly ignored advice to even try using tokens and try benchmarking if it solves your perf issues (https://github.com/openhab/openhab-core/issues/1899#issuecomment-739996254, https://github.com/openhab/openhab-core/issues/1899#issuecomment-768097895)
- you implemented this PR anyways against multiple reservations and silently removed a documented feature that allows people to still enjoy the benefits of using a token when their clients don't allow it because it's "a bad hack" (that you will have to do soon if you want to continue Git using GitHub)
- you nagged the maintainers until you got your review (and refactoring)

At that point, let me ask, who's selfish, and who's constructive, and who has the attitude? You tell me.

No need to waste your keystrokes to answer. I'm done here, and I'm recusing myself from this conversation.

Comment (spacemanspiff2007) [reply bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java:149]: It's funny that you feel the same way as I do, but with switched personas:

- In this issue I've pointed out that auth will make problems and just asked for a config switch to disable it. And here we are, arguing back and forth still not having a feasible solution.
- I was glad that you jumped in and created the basic auth option
- I asked for help to disable auth and I got an answer that broke the UI so of course I felt betrayed, especially since my hint that it breaks the UI has been [left unanswerd](https://github.com/openhab/openhab-core/issues/2060#issuecomment-757419590). I haven't gotten the news that you fixed this issue and I was under the impression it'll stay that way. It would have been nice if you would have linked the pr to fix the issue.
- I understand why passwords are deliberately slow to validate and even acknowledged that this is a good thing in my [comment](https://github.com/openhab/openhab-core/pull/2030#issuecomment-757651574). However in my comment I also describe that an attacker gains no new information from sending the same auth twice so subsequent requests should be quick to validate. That point and the differentiation between passwords in memory and password storage/transport has been (I hope not deliberately) ignored by you.
- I couldn't switch to token based auth because I'm depending on sse-libraries that don't allow setting custom headers. I know that's bad, but there aren't many sse libraries in python so I'll need to jump in and help refactoring but that's something that still has to be done
- When it was clear that there would be no quick option that'll fix my application I decided to jump in and provide an implementation in java myself which is not the language of my choice. The implementations also happens to fix many of the reverse proxy issues apparent in the forum
- True

I tried my best to make a simple solution work but it sure felt that you tried every time to shoot it down.

Comment (wborn) [reply bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java:149]: > Why remove it - you loose nothing if it's opt in.

We remove unneccessary features because it takes time to maintain, document and the extra configuration options confuse users. I.e. the same reasons as why the compatibility layer and various UIs were removed in OH 3.

> I couldn't switch to token based auth because I'm depending on sse-libraries that don't allow setting custom headers. I know that's bad, but there aren't many sse libraries in python so I'll need to jump in and help refactoring but that's something that still has to be done

Did you try using a token as username and an empty password with the library? That should result in the same kind of Basic authentication header that should already work since 4. is already implemented. If it doesn't work because of the empty password, it should work when 5. is implemented.

Comment (kaikreuzer) [reply bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java:149]: Hey guys, please calm down again. I don't understand why the discussion suddenly escalates like this, I had the hope that we were almost done and all that is left is to add a "Github-like token passing with `Authorization` header" for tools that can't deal with adding proper HTTP Bearer headers. I'm volunteering to (re-)implement this according to @ghys [summary above](https://github.com/openhab/openhab-core/pull/2101#discussion_r598359939). I think with that we should all be ok to accept this PR and everyone should be served.

Comment (ghys) [reply bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java:149]: > Hey guys, please calm down again. I don't understand why the discussion suddenly escalates like this, 

Sorry Kai, I know I overreacted, with the hundreds of hours I pour into this project out of "love", really (I don't even use the auth layer, I just want it to be "right"), being accused of being "selfish"? This was just the last straw.

Comment (spacemanspiff2007) [reply bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java:149]: @ghys:
I'm sorry I called your behavior selfish and about the too rough tone. My intention was never to insult you.
I've - like you - poured hundreds of hours into openhab in the form of a 3rd party rule engine and all I could see was it breaking and my countless efforts being all for nothing.

File: bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/ExpiringUserSecurityContextCache.java [hunk 19]
```diff
@@ -0,0 +1,92 @@
...
+
+    synchronized @Nullable UserSecurityContext get(String key) {
+        calls++;
+        if (calls >= CLEANUP_FREQUENCY) {
+            entryMap.keySet().forEach(k -> getEntry(k));
+            calls = 0;
+        }
+        Entry entry = getEntry(key);
+        if (entry != null) {
+            return entry.value;
```
spacemanspiff2007, line 58: @kaikreuzer 
Shouldn't we reset the entry.timestamp when we have a cache hit?

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
spacemanspiff2007: > If it doesn't work because of the empty password, it should work when 5. is implemented.

Yep - I'll make sure to check again once 5. is implemented

@kaikreuzer 
Maybe we hash the [whole auth header](https://github.com/openhab/openhab-core/blob/ed42fd0a82762fcc1944a4d4e2d253ae37ad508e/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java#L228) and use it as a key for the cache? That way everything gets a speedup independent of the authentication method.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -111,68 +183,89 @@ private SecurityContext authenticateBearerToken(String token) throws Authenticat
...
-                requestContext.setSecurityContext(authenticateBearerToken(altTokenHeader));
-                return;
-            }
+    public void filter(@Nullable ContainerRequestContext requestContext) throws IOException {
+        if (requestContext != null) {
+            String cacheKey = null;
+            try {
+                String altTokenHeader = requestContext.getHeaderString(ALT_AUTH_HEADER);
+                if (altTokenHeader != null) {
+                    cacheKey = getCacheKey(altTokenHeader);
```
spacemanspiff2007, line 213: Imho it's not necessary to hash this token. Should this be replaced with
```
cacheKey = altTokenHeader;
```

wborn:
Looks like Spotless doesn't agree with your formatting @spacemanspiff2007.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/ExpiringUserSecurityContextCache.java
```diff
@@ -31,8 +31,8 @@
  */
 @NonNullByDefault
 public class ExpiringUserSecurityContextCache {
-    final static private int MAX_SIZE = 10;
-    final static private int CLEANUP_FREQUENCY = 10;
+    final static private int MAX_SIZE = 30; // includes failed auth attempts!
```
kaikreuzer, line 35: Does it? When are they added?

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
kaikreuzer: Why did you move the cache here?
I do not see any need to use the cache for tokens.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -207,47 +200,62 @@ private SecurityContext authenticateBasicAuth(String credentialString) throws Au
...
+                            requestContext.setSecurityContext(securityContext);
                         }
                     }
                 } else if (implicitUserRole) {
                     requestContext.setSecurityContext(new AnonymousUserSecurityContext());
                 }
             } catch (AuthenticationException e) {
-                logger.warn("Unauthorized API request: {}", e.getMessage());
+                String errorMessage = e.getMessage();
+                authCache.put(cacheKey, null, errorMessage);
```
kaikreuzer, line 265: Why is this useful?

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/ExpiringUserSecurityContextCache.java
```diff
@@ -47,46 +50,65 @@ protected boolean removeEldestEntry(Map.@Nullable Entry<String, Entry> eldest) {
         };
     }
 
-    synchronized @Nullable UserSecurityContext get(String key) {
+    synchronized @Nullable SecurityContext get(@Nullable String key) throws AuthenticationException {
```
kaikreuzer, line 54: Why allow null as a key?

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/ExpiringUserSecurityContextCache.java
```diff
@@ -47,46 +50,65 @@ protected boolean removeEldestEntry(Map.@Nullable Entry<String, Entry> eldest) {
...
+
         calls++;
         if (calls >= CLEANUP_FREQUENCY) {
             entryMap.keySet().forEach(k -> getEntry(k));
             calls = 0;
         }
         Entry entry = getEntry(key);
         if (entry != null) {
-            return entry.value;
+            if (entry.errorMessage != null) {
```
kaikreuzer, line 66: This change wouldn't be necessary if we didn't store those in the first place.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/ExpiringUserSecurityContextCache.java
```diff
@@ -47,46 +50,65 @@ protected boolean removeEldestEntry(Map.@Nullable Entry<String, Entry> eldest) {
...
+            return entry.context;
         }
+
         return null;
     }
 
-    synchronized void put(String key, UserSecurityContext value) {
-        entryMap.put(key, new Entry(System.currentTimeMillis(), value));
+    synchronized void put(@Nullable String key, @Nullable SecurityContext value, @Nullable String errorMessage) {
+        if (key != null) {
```
kaikreuzer, line 76: Why allow null parameters now?

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/ExpiringUserSecurityContextCache.java
```diff
@@ -47,46 +50,65 @@ protected boolean removeEldestEntry(Map.@Nullable Entry<String, Entry> eldest) {
...
+            entryMap.put(key, new Entry(System.currentTimeMillis(), value, errorMessage));
+        }
     }
 
     synchronized void clear() {
         entryMap.clear();
     }
 
-    private @Nullable Entry getEntry(String key) {
+    private @Nullable Entry getEntry(@Nullable String key) {
```
kaikreuzer, line 85: Why allow null parameters now?

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/ExpiringUserSecurityContextCache.java
```diff
@@ -47,46 +50,65 @@ protected boolean removeEldestEntry(Map.@Nullable Entry<String, Entry> eldest) {
...
         Entry entry = entryMap.get(key);
         if (entry != null) {
-            long entryAge = System.currentTimeMillis() - entry.timestamp;
+            final long curTimeMillis = System.currentTimeMillis();
+            long entryAge = curTimeMillis - entry.timestamp;
             if (entryAge < 0 || entryAge >= keepPeriod) {
                 entryMap.remove(key);
                 entry = null;
+            } else {
+                entry.timestamp = System.currentTimeMillis();
```
kaikreuzer, line 98: Better reuse `curTimeMillis` here.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/ExpiringUserSecurityContextCache.java
```diff
@@ -47,46 +50,65 @@ protected boolean removeEldestEntry(Map.@Nullable Entry<String, Entry> eldest) {
...
 
     static class Entry {
-        final public long timestamp;
-        final public UserSecurityContext value;
+        public long timestamp;
+        final @Nullable public SecurityContext context;
+        final @Nullable public String errorMessage;
 
-        Entry(long timestamp, UserSecurityContext value) {
+        Entry(long timestamp, @Nullable SecurityContext context, @Nullable String errorMessage) {
```
kaikreuzer, line 109: Why null values here?

bundles/org.openhab.core.io.rest.auth/src/main/resources/OH-INF/config/config.xml
```diff
@@ -16,12 +16,11 @@
 			<label>Cache Expiration Time</label>
 			<default>6</default>
 			<unitLabel>h</unitLabel>
-			<description>When basic authentication is activated, credentials are put in a cache in order to speed up request
```
kaikreuzer: I'd think it was just fine as it was before.

kaikreuzer (changes requested):
Not sure if I have missed anything, but your latest changes don't make much sense to me.
I would have preferred to keep the code as is and to merely add the lost feature of token authentication again (without applying the cache to it, since for tokens it does not really make any sense).

spacemanspiff2007 (commented):
If caching the negative responses is not desired I can revert the changes again.
But imho this makes sense since it speeds up cases where the user enters a wrong password.
The entries should get thrown out of the cache pretty quickly, so.

---

Just because tokens are ~15x quicker doesn't mean they don't need a speedup.
Whilst I haven't profiled it yet if I (roughly) calculate back the performance:
```
350ms (request for basic auth)
 15x  (speedup for tokens, based on 15k vs 1k iterations)
-------
~24ms per token request
```

That tops out at ~45 requests per seconds which is ~3x slower than the basic auth implementation with cache.

If the speedup or the caching of the rejected auth is not desired, I'll gladly revert it.
Just tell me what to do so this can finally be merged.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/ExpiringUserSecurityContextCache.java
spacemanspiff2007:
```
            } catch (AuthenticationException e) {
                logger.warn("Unauthorized API request: {}", e.getMessage());
                String errorMessage = e.getMessage();
                authCache.put(cacheKey, null, errorMessage);
```

kaikreuzer:
I haven't seen anybody asking for it and rather understood @ghys that he simply wishes that this feature stays untouched as it is. So I think it would be simplest to just address what we really require here and hence revert these changes.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/ExpiringUserSecurityContextCache.java
kaikreuzer: Afair, the previous code always made sure that the cacheKey had a value - so there should be no need for the check.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/ExpiringUserSecurityContextCache.java
kaikreuzer: The error message imho does not belong in here.

spacemanspiff2007:
There you go - I've reverted the changes I made.
These should be the changes from your branch.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
kaikreuzer: If unauthorized requests are done, it is imho good if they do not return quickly.
For any DoS-style attack, we should rather think about putting some throttling in place, but not to increase the response speed, which is rather counter-productive.
If a "friendly" app keeps firing requests even though it sees that they are unauthenticated, the logic should rather be changed within the app.

spacemanspiff2007:
I forgot that this reversed the time reset on cache hit and I've added it again.
So this should be basically your PR and good to go.

spacemanspiff2007:
There are errors in the build, but I am not sure why.

kaikreuzer:
All, to stay in scope of this PR, I have decided to only add again, what was there before wrt token-based auth, see https://github.com/openhab/openhab-core/pull/2101/commits/622e3a17b2f015fdd35efadb19c5395ef420868e.
Improving this by adding an `user:token` option should imho better be done in a follow-up PR.

kaikreuzer:
@wborn New try to ask for your review&approval :-)

wborn (approved):
Thank you all!

spacemanspiff2007:
Thank you guys, especially @kaikreuzer for reworking!

WiredLife:
hi there
is this already included in 3.0.2?
have long wait times when updating items with wget or curl.

SebastianRumohr:
It is included in 3.1.0.M3
https://github.com/openhab/openhab-distro/releases/tag/3.1.0.M3

Issue #2033, opened 2020-12-30
OH3.0.0: ExecUtil.executeCommandLine() is not returning result if command return code !=0


olialb:
Just migrate from OH2 to OH3 and I am in general very happy with the new version...

But I am porting my executeCommandLine commands in my rules from OH2 to OH3 and I observerd that some commands are not excecuted because the return value of executeCommandLine is null or "". But it seams they are actually exceuted, only the output is not catched in the result variable. 
After playing arround with python2 and python3 scripts I come to the following conclusion:

If the command line returns something !=0 the output is not catched and empty. Also there is no error in the log.

This can be reproduced with the following rule content

`var String ScriptPath = "/etc/openhab/scripts/"`
`var result = executeCommandLine(Duration.ofSeconds(50),"python", ScriptPath+"test.py")`

And and the script **test.py**:

`import sys`
`print("Hallo World")`
`sys.exit(1)`

the result is null or "".

If the script is modified like this:

`import sys`
`print("Hallo World")`
`sys.exit(0)`

result is "Hello World" as expected.
I would say this is a bug or is it a feature? For sure it is not nice that there is not any output in the log if the return value is != 0 and the result variable is empty. Also its diffrent undocumented behaviour to OH2.

See also here: https://community.openhab.org/t/oh3-openhabianpi-executecommandline-returns-always-with-empty-result-for-python2-commands/112056

Stefan300381:
Hi,

I face the same issue when I run some ping and ssh commands.
Here a quick example:

```
        //this ip address does not exist in my LAN
        var result = executeCommandLine(Duration.ofSeconds(5),"/usr/bin/ping", "-c", "1", "192.168.2.254")
        logInfo("ping", result) 

        //this ip address DOES exist in my LAN
        result = executeCommandLine(Duration.ofSeconds(5),"/usr/bin/ping", "-c", "1", "192.168.2.10")
        logInfo("ping", result) 
```
As you can see, first result is null:

```
2021-01-01 22:57:24.717 [INFO ] [org.openhab.core.model.script.ping  ] - null

2021-01-01 22:57:24.778 [INFO ] [org.openhab.core.model.script.ping  ] - PING 192.168.2.10 (192.168.2.10) 56(84) bytes of data.
64 bytes from 192.168.2.10: icmp_seq=1 ttl=64 time=0.535 ms
--- 192.168.2.10 ping statistics ---
1 packets transmitted, 1 received, 0% packet loss, time 0ms
[...]

PR #2104, opened 2021-01-11
Also return command line execution result in case of error return code

### LINKED ISSUE #2078: ExecUtil.executeCommandLineAndWaitResponse should return stderr response e.g. for Network binding
ExecUtil.executeCommandLineAndWaitResponse checks the return code of the program and if it is not 0, then the returned text will be discarded.

            if (exitCode == 0) {
                return outputFuture.get();
            } else {
                    logger.debug("exit code '{}', result '{}', errors '{}'", exitCode, outputFuture.get(),
                            errorFuture.get());
                }
                return null;
            }

## Expected Behavior
since stdout and stderr can be written with any exit code, the result of calling an executable should include the exit code, stdout and stderr. Especially when the stderr is read into errorFuture some lines above.

## Current Behavior
See description

## Possible Solution
Create an Class for the Result (How is this problem solved in other applications?)

Install the network binding and use iptools arping. It will call arping with argument --help. The arping does not know the argument and returns with code 2 and the argument list printed to stderr.

## Context
I'm trying to get the Network binding working and discovered this issue with arping. But it may be an issue with other bindings too, which run executables.

## Your Environment
current main & master of openHAB (problem seem to exist since long time OH 2.5.x with network binding.)
Ubuntu 20 on Odroid C2 with iputils arping
Linuxmint on Oracle VB with iputils arping

arping -V returns "arping from iputils s20190709"

kaikreuzer:
Fixes https://github.com/openhab/openhab-core/issues/2033
Fixes #2078 

### DIFF
--- a/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/exec/ExecUtil.java
+++ b/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/exec/ExecUtil.java
@@ -115,7 +115,7 @@ public class ExecUtil {
                     logger.debug("exit code '{}', result '{}', errors '{}'", exitCode, outputFuture.get(),
                             errorFuture.get());
                 }
-                return null;
+                return errorFuture.get();
             }
         } catch (ExecutionException e) {
             if (logger.isDebugEnabled()) {


Confectrician (approved):
👍

lolodomo:
@kaikreuzer : can you please check that it will not break any existing OH3 binding?

kaikreuzer:
@lolodomo I found 4 uses of the method, all of them should work nicely with that change.

lolodomo:
Thank you.

wborn (approved):
Thanks!

PR #2114, opened 2021-01-14
[executeCommandLine] should return STDERR if STDOUT is empty

andrewfg:
NOTE: this relates to Issue https://github.com/openhab/openhab-core/issues/2033#issuecomment-759781893

If ErrorCode == 0 then executeCommandLineAndWaitResponse() would always return the response from STDOUT. 

However on some Unix commands (e.g. wget, curl) if ErrorCode == 0 there is no response on STDOUT, but the commands return valid information on STDERR (even though there was not an error).

In this PR, if ErrorCode == 0, the return value of executeCommandLineAndWaitResponse() is `!"".equals(STDOUT) ? STDOUT : STDERR`

### DIFF
--- a/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/exec/ExecUtil.java
+++ b/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/exec/ExecUtil.java
@@ -90,12 +90,9 @@ public class ExecUtil {
                 }
             });
 
-            int exitCode;
             if (timeout == null) {
-                exitCode = process.waitFor();
-            } else if (process.waitFor(timeout.toMillis(), TimeUnit.MILLISECONDS)) {
-                exitCode = process.exitValue();
-            } else {
+                process.waitFor();
+            } else if (!process.waitFor(timeout.toMillis(), TimeUnit.MILLISECONDS)) {
                 logger.warn("Timeout occurred when executing commandLine '{}'", Arrays.toString(commandLine));
                 break cleanup;
             }
--- a/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/exec/ExecUtil.java
+++ b/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/exec/ExecUtil.java
@@ -78,9 +78,8 @@ public class ExecUtil {
 
         Process processTemp = null;
         Future<String> outputFuture = null;
-        Future<String> errorFuture = null;
         cleanup: try {
-            Process process = processTemp = new ProcessBuilder(commandLine).start();
+            Process process = processTemp = new ProcessBuilder(commandLine).redirectErrorStream(true).start();
 
             outputFuture = executor.submit(() -> {
                 try (InputStream inputStream = process.getInputStream();
@@ -91,14 +90,6 @@ public class ExecUtil {
                 }
             });
 
-            errorFuture = executor.submit(() -> {
-                try (InputStream inputStream = process.getErrorStream();
-                        BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream))) {
-                    StringWriter output = new StringWriter();
-                    reader.transferTo(output);
-                    return output.toString();
-                }
-            });
             int exitCode;
             if (timeout == null) {
                 exitCode = process.waitFor();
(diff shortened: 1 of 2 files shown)


andrewfg:
@kaikreuzer @cpmeister

cpmeister (approved):

cpmeister:
LGTM

andrewfg:
> how could the stderr output of wget be read?

I have no idea. All I can say is that on OH2 it returns the message, and on OH3 it does not. Maybe it is related to the change of Java version, or something like that??

> would prefere to set the ProcessBuilder redirectErrorStream property to true

Ok. Currently I have no idea how to do that though. But I will look into it tomorrow. However it might therefore require reverting the prior commit related to STDERR and STDOUT when ResultCode != 0. Since if both streams are merged in STDOUT then there's no need to look at STDERR anymore at all. Or??

NorbertHD:
> I have no idea. 

I  was hoping that someone with more java (ProcessBuilder) knowledge than me knows the answer.

> Ok. Currently I have no idea how to do that though

You have to remove the errorFuture code and change

    Process process = processTemp = new ProcessBuilder(commandLine).start();

to

    Process process = processTemp = new ProcessBuilder(commandLine);
    process.redirectErrorStream(true);
    process.start();

> then there's no need to look at STDERR anymore at all?

Yes

andrewfg:
@NorbertHD many thanks for the tips. I have made the changes that you suggest.
Note: it works perfectly now in my environment. And the JUnit tests all passed.

@cpmeister it is ready to go (again) so can you please have a look at it (again) :)

cpmeister (approved):
LGTM

andrewfg:
I think it would be useful if I add an extra JUnit test to confirm the STDERR redirection.

cweitkamp:
Unit tests are always appreciated. Can you do that quickly? Otherwise a follow-up PR may be okay too.

I would like to merge this into https://github.com/openhab/openhab-core/tree/3.0.x before building a 3.0.1 patch release for this weekend.

andrewfg:
I will try it today.

andrewfg:
> add an extra JUnit test to confirm the STDERR redirection

@cweitkamp done :)

kaikreuzer:
> I would like to merge this into https://github.com/openhab/openhab-core/tree/3.0.x before building a 3.0.1 patch release for this weekend.

@cweitkamp Afaik, this is an improvement, not a bug fix.
The regression in comparison to 2.5 was fixed by https://github.com/openhab/openhab-core/pull/2104 and is already cherry-picked to 3.0.1.

andrewfg:
> yhis is an improvement, not a bug fix.

@kaikreuzer hmm, not really. The prior PR only partly solved the regression from v2.5 (the case where errorCode != 0), but it did not resolve the case where errorCode == 0 and (yet) there was still data on STDERR. But I think we are talking of nuances here..

kaikreuzer:
Are you really sure that STDERR was considered in 2.5? I wouldn't understand why this would have been the case - [from the code](https://github.com/openhab/openhab-core/blob/2.5.0/bundles/org.openhab.core.io.net/src/main/java/org/eclipse/smarthome/io/net/exec/ExecUtil.java#L127) it looks as if clearly only STDOUT was returned.

kaikreuzer:
Ok, diving a bit deeper, it seems you are right - the variable name in 2.5 was chosen misleadingly. The [PumpStreamHandler](https://commons.apache.org/proper/commons-exec/apidocs/org/apache/commons/exec/PumpStreamHandler.html#PumpStreamHandler(java.io.OutputStream)) indeed combines the output of STDOUT and STDERR into a single stream.

andrewfg:
> indeed combines the output of STDOUT and STDERR into a single stream.

Yes. And the test case is the 'wget' command which just returns an empty string in STDOUT, and returns the full connection status trace with its HTTP response headers in STDERR. The 'curl' command is similar..

kaikreuzer:
> I would like to merge this into https://github.com/openhab/openhab-core/tree/3.0.x before building a 3.0.1 patch release for this weekend.

Go ahead with this then @cweitkamp - I leave it to you :-)

kaikreuzer (approved):
Seems you are busy, so I'll take care of it.

cweitkamp:
Yes, I was. Thanks for taking care.

PR #2120, opened 2021-01-15
Fix BOM dependency issues

wborn:
* Sync biz.aQute.tester.junit-platform version using bnd.version property
* Remove duplicate osgi-resource-locator dependency, causing warnings/unstability


cweitkamp (approved):
LGTM. Thanks.

PR #2123, opened 2021-01-16
Sort group item by string

bjoernbrings:
When using groups in the sitemap the order is quite randomly.
According [ticket](https://github.com/openhab/openhab-webui/issues/76) was closed in the ui repository as the solution needs to come from the core.

I saw @ghys coming up with some orderid property which I really liked but wasn't sure whether that is a general solution. Would love to implement that if I got some hint.
For now sorting by string (which I implemented with this little PR) is to me much better than a random order.

Signed-off-by: bjoernbrings <bjoernbrings@web.de>

### DIFF
--- a/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java
+++ b/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java
@@ -15,8 +15,10 @@ package org.openhab.core.ui.internal.items;
 import java.time.DateTimeException;
 import java.time.ZonedDateTime;
 import java.time.temporal.ChronoUnit;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


cweitkamp:
Thanks for this contribution. I really like the idea of ordering Group members. I agree with @ghys that we need a general solution based on metadata settings because it will maybe not only makes sense in UIs but in other places too. I assume users might want to decide individually if the ordering is ascending or descending and which property or properties should be ordered by (e.g. name, label, value, and so on).

bjoernbrings:
Sorry to pop up but any idea how to proceed?

cweitkamp:
If we want to introduce a default sorting I personally would prefer by Label instead of item name. Wdyt?

bjoernbrings:
Agree, did adapt accordingly.
Still I think it would be good to have a manual way to change the order as I like a lot to work with groups / semantic model to auto generate the sitemap.

cweitkamp (approved):
LGTM. Thanks.

cweitkamp:
Jenkins build errors do not look like they are related.

lolodomo:
This PR was a disaster (ok, disaster is a bit excessive ;) ) because it broke all the work done by users to order the items in a specific order in their item file.

I am suggesting to add a server configuration setting to define which filter to apply:

- No filter: historic behaviour which respects order of definition in item file
- item label: the current beahviour introduced by this PR

I will implement this.

PR #2126, opened 2021-01-17
Add current owner to PortInUseException

fwolter:
I don't know when the javacomm bundle is needed, so I applied the changes there, too.


wborn:
It would also be helpful to preserve the root cause exception as `cause` in thrown exceptions. The serial transport also does this when throwing an `UnsupportedCommOperationException` .

fwolter:
Like so?

wborn (approved):
Thank you!

Issue #2115, opened 2021-01-14
Removal of dependency on 'org.apache.commons.*' from o.o.c.io.transport.modbus


cweitkamp:
In openhab/openhab-core#1433 and openhab/openhab-core#1441 (and more) we removed dependency on `org.apache.commons.*` libraries. They will be removed from OH3 Core API and thus not available anymore for OH add-ons too.

```
./src/main/java/org/openhab/core/io/transport/modbus/endpoint/ModbusIPSlaveEndpoint.java:import org.apache.commons.lang.builder.EqualsBuilder;
./src/main/java/org/openhab/core/io/transport/modbus/endpoint/ModbusIPSlaveEndpoint.java:import org.apache.commons.lang.builder.HashCodeBuilder;
./src/main/java/org/openhab/core/io/transport/modbus/endpoint/ModbusIPSlaveEndpoint.java:import org.apache.commons.lang.builder.StandardToStringStyle;
./src/main/java/org/openhab/core/io/transport/modbus/endpoint/ModbusIPSlaveEndpoint.java:import org.apache.commons.lang.builder.ToStringBuilder;
./src/main/java/org/openhab/core/io/transport/modbus/endpoint/EndpointPoolConfiguration.java:import org.apache.commons.lang.builder.EqualsBuilder;
./src/main/java/org/openhab/core/io/transport/modbus/endpoint/EndpointPoolConfiguration.java:import org.apache.commons.lang.builder.HashCodeBuilder;
./src/main/java/org/openhab/core/io/transport/modbus/endpoint/EndpointPoolConfiguration.java:import org.apache.commons.lang.builder.StandardToStringStyle;
./src/main/java/org/openhab/core/io/transport/modbus/endpoint/EndpointPoolConfiguration.java:import org.apache.commons.lang.builder.ToStringBuilder;
./src/main/java/org/openhab/core/io/transport/modbus/endpoint/ModbusSerialSlaveEndpoint.java:import org.apache.commons.lang.builder.EqualsBuilder;
./src/main/java/org/openhab/core/io/transport/modbus/endpoint/ModbusSerialSlaveEndpoint.java:import org.apache.commons.lang.builder.StandardToStringStyle;
./src/main/java/org/openhab/core/io/transport/modbus/endpoint/ModbusSerialSlaveEndpoint.java:import org.apache.commons.lang.builder.ToStringBuilder;
[...]

Issue #2006, opened 2020-12-25
Modbus transport directly uses deprecated jetty ConcurrentHashSet


splatch:
The `ConcurrentHashSet` from org.eclipse.jetty.util package is deprecated since Java 8 and can be safely swapped with `ConcurrentHashMap.newKeySet`.

splatch:
@ssalonen I believe you're most suited to judge migration to other type. :-)

ssalonen:
I don't see a problem migrating to other type, looks like a drop-in replacement?

splatch:
Yes, jetty javadoc points sdk class as replacement.

kaikreuzer:
Fixed by https://github.com/openhab/openhab-core/pull/2127.

PR #2127, opened 2021-01-17
[transport.modbus] Remove dependencies on deprecated libraries

andrewfg:
Addressing the following Issues..
- https://github.com/openhab/openhab-core/issues/2006 (dependency org.eclipse.jetty.util.ConcurrentHashSet)
- https://github.com/openhab/openhab-core/issues/2115 (dependency org.apache.commons.*)

References..
- https://github.com/openhab/jamod/pull/9

This PR removes a few (but not all) of the deprecated libraries..
- org.eclipse.jetty.util.ConcurrentHashSet
- org.apache.commons.lang.builder.EqualsBuilder
- org.apache.commons.lang.builder.HashCodeBuilder
- org.apache.commons.lang.builder.ToStringBuilder
- org.apache.commons.lang.builder.StandardToStringStyle

**_Note: it does not remove all of the dependencies on 'org.apache.commons.*'_** (but it is a big first step..)

@ssalonen @splatch @cweitkamp

### DIFF
--- a/bundles/org.openhab.core.io.transport.modbus/src/main/java/org/openhab/core/io/transport/modbus/ModbusReadRequestBlueprint.java
+++ b/bundles/org.openhab.core.io.transport.modbus/src/main/java/org/openhab/core/io/transport/modbus/ModbusReadRequestBlueprint.java
@@ -14,8 +14,6 @@ package org.openhab.core.io.transport.modbus;
 
 import org.apache.commons.lang.builder.EqualsBuilder;
 import org.apache.commons.lang.builder.HashCodeBuilder;
-import org.apache.commons.lang.builder.StandardToStringStyle;
-import org.apache.commons.lang.builder.ToStringBuilder;
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 
@@ -33,10 +31,6 @@ import net.wimpi.modbus.Modbus;
  */
 @NonNullByDefault
 public class ModbusReadRequestBlueprint {
-    private static StandardToStringStyle toStringStyle = new StandardToStringStyle();
-    static {
-        toStringStyle.setUseShortClassName(true);
-    }
 
     private final int slaveId;
     private final ModbusReadFunctionCode functionCode;
@@ -110,8 +104,8 @@ public class ModbusReadRequestBlueprint {
 
     @Override
     public String toString() {
-        return new ToStringBuilder(this, toStringStyle).append("slaveId", slaveId).append("functionCode", functionCode)
-                .append("start", start).append("length", length).append("maxTries", maxTries).toString();
+        return "ModbusReadRequestBlueprint [slaveId=" + slaveId + ", functionCode=" + functionCode + ", start=" + start
+                + ", length=" + length + ", maxTries=" + maxTries + "]";
     }
 
     @Override
--- a/bundles/org.openhab.core.io.transport.modbus/src/main/java/org/openhab/core/io/transport/modbus/ModbusReadRequestBlueprint.java
+++ b/bundles/org.openhab.core.io.transport.modbus/src/main/java/org/openhab/core/io/transport/modbus/ModbusReadRequestBlueprint.java
@@ -119,7 +119,6 @@ public class ModbusReadRequestBlueprint {
             return false;
         }
         ModbusReadRequestBlueprint rhs = (ModbusReadRequestBlueprint) obj;
-        return functionCode == rhs.functionCode && length == rhs.length && maxTries == rhs.maxTries
-                && slaveId == rhs.slaveId && start == rhs.start;
+        return functionCode == rhs.functionCode && length == rhs.length && slaveId == rhs.slaveId && start == rhs.start;
     }
 }
--- a/bundles/org.openhab.core.io.transport.modbus/src/main/java/org/openhab/core/io/transport/modbus/ModbusReadRequestBlueprint.java
+++ b/bundles/org.openhab.core.io.transport.modbus/src/main/java/org/openhab/core/io/transport/modbus/ModbusReadRequestBlueprint.java
@@ -12,8 +12,8 @@
  */
 package org.openhab.core.io.transport.modbus;
 
-import org.apache.commons.lang.builder.EqualsBuilder;
-import org.apache.commons.lang.builder.HashCodeBuilder;
+import java.util.Objects;
+
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 
@@ -98,8 +98,7 @@ public class ModbusReadRequestBlueprint {
 
     @Override
     public int hashCode() {
-        return new HashCodeBuilder(81, 3).append(slaveId).append(functionCode).append(start).append(length)
-                .append(maxTries).toHashCode();
+        return Objects.hash(functionCode, length, maxTries, slaveId, start);
     }
 
     @Override
@@ -120,7 +119,7 @@ public class ModbusReadRequestBlueprint {
             return false;
         }
         ModbusReadRequestBlueprint rhs = (ModbusReadRequestBlueprint) obj;
-        return new EqualsBuilder().append(slaveId, rhs.slaveId).append(functionCode, rhs.functionCode)
-                .append(start, rhs.start).append(length, rhs.length).isEquals();
+        return functionCode == rhs.functionCode && length == rhs.length && maxTries == rhs.maxTries
+                && slaveId == rhs.slaveId && start == rhs.start;
     }
 }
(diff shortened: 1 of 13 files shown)


bundles/org.openhab.core.io.transport.modbus/src/main/java/org/openhab/core/io/transport/modbus/ModbusReadRequestBlueprint.java
```diff
@@ -126,7 +119,7 @@ public boolean equals(@Nullable Object obj) {
             return false;
         }
         ModbusReadRequestBlueprint rhs = (ModbusReadRequestBlueprint) obj;
-        return new EqualsBuilder().append(slaveId, rhs.slaveId).append(functionCode, rhs.functionCode)
-                .append(start, rhs.start).append(length, rhs.length).isEquals();
+        return functionCode == rhs.functionCode && length == rhs.length && maxTries == rhs.maxTries
```
ssalonen, line 122: Original implementation did not compare maxTries. I think it is simply a mistake, cannot see any harm with adding it to equals

ssalonen (approved):
Thanks you very much.

ssalonen:
Spotless fails in Jenkins, some simple formatting errors

andrewfg:
@ssalonen concerning the hashes: the new code generates different hashes than the prior code; can you confirm that there are no chances of the hashes being persisted and compared (wrongly) between old and new?

andrewfg:
> Spotless fails in Jenkins

Ok I will run it tomorrow..

bundles/org.openhab.core.io.transport.modbus/src/main/java/org/openhab/core/io/transport/modbus/ModbusReadRequestBlueprint.java
andrewfg: Ok. Will fo

andrewfg: I reverted this just in case.

andrewfg:
@ssalonen I am seeing some errors in the logger outputs from the JUnit tests (the actual tests are passing ok, but there are logger.error messages appearing in the console output). I am not sure is this is a regression, or if it is supposed to be like this. => Can you please advise?

@cweitkamp in https://github.com/openhab/openhab-core/issues/2115 in your first post, you specifically mentioned some `org.apache.commons.lang.builder.*` modules where you want to remove the dependencies. However I notice that there are also quite a lot of dependencies on `org.apache.commons.pool2.*`; so my question is whether you want to try to remove those as well?

cweitkamp:
@andrewfg Yes, I saw them too. The reason why I was asking to remove o.a.c.l only is because it is included by the openHAB feature POM. o.a.c.pool2 is only included by this bundle - which is okay as long as it does not depend on the earlier mentioned lib.

If you know a valid replacement and think you can refactor it without causing regression feel free to do so. It can be done in a follow-up PR.

andrewfg:
> asking to remove o.a.c.l only is because it is included by the openHAB feature POM

OK @cweitkamp I have finished removing all the dependencies on o.a.c.l so it is ready for you to review.

> o.a.c.pool2 is only included by this bundle

Understood. In that case, I will **_not_** attempt any refactoring of o.a.c.pool2 in this PR.

@cweitkamp I don't know how to handle the pom files between this PR and the openhab/jamod PR. The interface between the two modules did not change from their respective prior releases, (only the external third party dependencies have been cut), so the new or prior version of one should work with either new or prior version of the other. I just don't know how to code the 'not before version' resp. 'not after version' in the pom file. Also both are declared as 'SNAPSHOT' versions and this obviously needs to be changed before a release.

Furthermore: I think the openhab/jamod module is used solely by transport.modbus so I wonder if it is necessary to build and distribute its own separate jar (on jfrog/ bintray). Or perhaps it would be simpler if transport.modbus would simply pull in the openhab/jamod source code and compile it directly into its jar file. What do you think?

ssalonen:
> Furthermore: I think the openhab/jamod module is used solely by transport.modbus so I wonder if it is necessary to build and distribute its own separate jar (on jfrog/ bintray). Or perhaps it would be simpler if transport.modbus would simply pull in the openhab/jamod source code and compile it directly into its jar file. What do you think?

My 2 cents...

This is how it was in openhab1 days, before we splitted it out with @kaikreuzer . In my opinion it's more hygienic to keep it as a separate repo, avoiding mixing the two code bases. 

Then again I see the complexity with the packaging currently. Do we expect many changes to the jamod fork in the future?

andrewfg:
Is it not possible to keep and maintain the jamod code in its own repo, and when building the oh core transport.modbus module, it just integrates that into the compiled transport.modbus jar?

Reason why I suggest this is that @cpmeister recently did something similar with the Tado binding (and for similar reasons)..

ssalonen:
That would be ideal indeed!

andrewfg:
This is what was done with the Tado bindiing; it builds an included `dependency` by converting a swagger yaml file into java code, and then including the respective generated java files into the final build, so it requires no dependency any more..

https://github.com/openhab/openhab-addons/pull/9125/commits/d22e8103fcb951c1aed0f30f3239f00c17e779c1

cweitkamp:
Yes, I can remember, that we did such integrations for some add-ons. I think it might work here too.

@openhab/core-maintainers Are you okay with integrating [openhab/jamod](https://github.com/openhab/jamod) into OHC? Or do we have licensing issue?

splatch:
> Doesn't the license cover the usage of the source code, rather than specifying how the code shall be compiled (into a separate jar or not)? It doesn't sound to me that we would have a licensing issue here. Or??

It depends. While Apache license is fairly liberal (more liberal than Eclipse Public License v1 and v2) changing source code license requires consent of copyright owners. You can do a lot with Apache code as long as it is licensed in the same way. Re-licensing usually this falls close to "IP clearance" category which involves reaching all patch authors who ever contributed to project and asking them to consent for re-license of their work. Most of people have no objections with that, but there are known examples of situations where re-licensing of lib caused some.. complains. If all contributions are coming from OH community members then you should have no troubles with consent. :-)

Looking at [usage of jamod](https://mvnrepository.com/artifact/net.wimpi/jamod/usages) - there are not many and most of them is dated around 2012/2013. However there are some projects which got re-released last year (such [marid](https://mvnrepository.com/artifact/org.marid/marid-proto)) or are even more popular than jamod itself (ie. [robotiq-hand-drivers](https://mvnrepository.com/artifact/us.ihmc/robotiq-hand-drivers)). Are you sure that you wanna cut these off by moving library to openhab and hard-wiring it into openhab ecosystem?

Looking at amount of commits in jamod (three since move in 2017) I would say that from this perspective it requires a release and update every three years. Are you sure that you want to drag over this dependency with no changes just for sake of not doing one ocassional release happening every 3 years? The same issue is with jmdns and jupnp which get released every couple of years and get used in core.

J-N-K:
I would agree with @splatch. I don‘t think that a release every six month or so is too much work. We don‘t need to setup release builds, that can be done locally and uploaded to our own bintray-repo.

kaikreuzer:
Yeah, I can cope with doing 2 releases of jamod a year...

andrewfg:
The point is that @cweitkamp wanted to remove dependencies on Apache commons. But honestly if you don't want the change, I don't mind at all.

J-N-K:
@andrewfg I didn’t understand why it would not be possible to remove the dependencies when the jamod bundles stays in it‘s own repo. Can you comment on that?

andrewfg:
The problem only is the synchronization of version numbers of the pom files for a) the jamod lib, and b) the core module that depends on it.

splatch:
> The problem only is the synchronization of version numbers of the pom files for a) the jamod lib, and b) the core module that depends on it.

It is not a problem if you use maven dependency management in proper way cause then openhab versions take over precedence over jamod ones. I've pointed this flaw in project organization long time ago: https://community.openhab.org/t/maven-boms-without-dependencymanagement/86393

Technically the BOMs should specify `<dependencyManagement>` and not `<dependencies>` alone cause later one just pulls everything to classpath. This is leading to situations where you will have two versions of `commons-lang` instead of one.

ssalonen:
I need to correct one thing here, for the sake of posterity.

> Are you sure that you wanna cut these off by moving library to openhab and hard-wiring it into openhab ecosystem?

The dependency analysis above misleading IMO. The openHAB jamod is a fork, not enjoying great usage outside openHAB to my understanding.

When looking at dependencies in we can see they refer to the original jamod (v1.2), not related to discussion here.

andrewfg:
I just pushed a new commit that tweaks the dependencies on jamod

kaikreuzer (approved):
Thanks!

andrewfg:
Thank you @kaikreuzer

Issue #774, opened 2019-04-30
Introduce core framework metrics


kaikreuzer:
Follow up of [this discussion](https://github.com/openhab/openhab-core/pull/744#issuecomment-487681836).

We should make use of Eclipse MicroProfile metrics API and provide all kinds of core metrics from openHAB this way - like CPU usage, heap size, number of threads, job queue sizes, event throughput, etc.

Once such information is provided through the API, we can use it to e.g. provide a Prometheus publisher, to provide it through JMX or directly integrate it into the systeminfo binding.

kaikreuzer:
@pokerazor How much effort would it be to port your implementation to the Eclipse MicroProfile API? I only mean the provider side, not the consuming part.

pokerazor:
Hm. Very hard to say for me, as I don't have experience with the Eclipse MicroProfile API ... how could we approach this? What is the aim? What is the downside of using what is currently implemented in the bundle?

pravussum:
Following up on the discussion in #2065 here:
> > Prometheus exporter binding
> 
> Please note that it must not be done as a binding, though. Exposing openHAB for Prometheus probably best qualifies as an I/O add-on. Please also note #774, which would be the preferred way of implementation (i.e. by using the MicroProfile API).

My approach is to write the metrics to items in order to let the user choose how to proceed with them.
* either Prometheus can directly access them via the items state REST API endpoint
* or it can be further processed / transformed / whatever

Why do you write "must not"? What's the downside?

I'll try to push my WIP version tonight.

pravussum:
@kaikreuzer What do you think about using [Micrometer](https://micrometer.io) instead of MicroProfile Metrics? 
MP Metrics themselves plan to [move towards that direction](https://github.com/eclipse/microprofile-metrics/blob/master/spec/src/main/asciidoc/intro.adoc) (implying future effort to migrate it, when going from MP Metrics 3 to 4) and looking at Quarkus, Micronaut and Spring Boot that seems the way to go for me.
https://quarkus.io/blog/micrometer-metrics/

kaikreuzer:
@pravussum Sounds like a very good point. The situation has changed since the initial discussion a few years back and Micrometer seems indeed being the preferred API for all the frameworks out there - so yes, heading for Micrometer sounds perfectly reasonable!

kaikreuzer:
Implemented with https://github.com/openhab/openhab-core/pull/2133.

PR #2133, opened 2021-01-18
Integrated micrometer metering #774

pravussum:
This is the initial draft of the micrometer integration to address #774.

I figured the best place to put it would be the already existing core.io.monitoring bundle. I'll move it otherwise.

Some notes: 
- the bundle registers some "default" metrics (basically all JVM related) and some openhab specific metrics (ThreadManager pool statistics, event counts, bundle states, thing states, rule run counts)
- it uses the Micrometer [global meter registry](https://micrometer.io/docs/concepts#_global_registry), a [composite registry](https://micrometer.io/docs/concepts#_composite_registries) where other registries (e. g. the Prometheus one) can attach later to generate the actual metrics
- I'm developing an IO addon, which attaches to the global meter registry and provides the actual metrics as REST endpoint (PR https://github.com/openhab/openhab-addons/pull/9890)
- other new Metrics could easily be added by registering with the global meter registry at their own convenience within their respective bundle

Some questions:
- I had a hard time to get the Maven dependencies, bundle dependencies and Karaf features straight. Is this the right way to include a third party Maven dependency (non OSGI-ready) including its transitive dependencies? See https://community.openhab.org/t/3rd-party-maven-dependencies-in-openhab-core/113722 for a related forum post as well. I ended up using Karafs wrap feature instead of adding the dependency to the osgiify repo - although no one else seems to have used it that way. 
 
/cc @kaikreuzer

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -1024,7 +1024,6 @@
       <version>0.9.12</version>
       <scope>compile</scope>
     </dependency>
-
   </dependencies>
 
 </project>
--- a/bundles/org.openhab.core.io.monitor/bnd.bnd
+++ b/bundles/org.openhab.core.io.monitor/bnd.bnd
@@ -1,0 +1,9 @@
+Bundle-SymbolicName: ${project.artifactId}
+Import-Package: \
+ org.eclipse.jdt.annotation.*;resolution:=optional,\
+ org.openhab.*;version=!,\
+ org.osgi.framework,\
+ org.osgi.service.*,\
+ org.slf4j.*,\
+ com.google.gson.*;version="[2.8,3)"
+Export-Package: io.micrometer.core.*
--- a/bundles/org.openhab.core.io.monitor/pom.xml
+++ b/bundles/org.openhab.core.io.monitor/pom.xml
@@ -20,6 +20,44 @@
       <artifactId>org.openhab.core</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.openhab.core.bundles</groupId>
+      <artifactId>org.openhab.core.automation</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.micrometer</groupId>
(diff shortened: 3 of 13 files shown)


wborn:
I saw a Micrometer presentation once and ever since put it on my todo list to start integrating it into projects. So thanks for creating this PR. :+1:

features/karaf/openhab-tp/src/main/feature/feature.xml
```diff
@@ -179,6 +179,12 @@
 		<feature dependency="true">openhab.tp-xtext</feature>
 	</feature>
 
+	<feature name="openhab.tp-micrometer" description="Micrometer instrumentation library" version="${project.version}">
+		<capability>openhab.tp;feature=micrometer;version=1.6.2</capability>
+		<bundle>wrap:mvn:io.micrometer/micrometer-core/1.6.2$Bundle-SymbolicName=micrometer-core&amp;Bundle-Version=1.6.2</bundle>
```
wborn, line 184: Using the wrap protocol causes issues when for instance debugging the code in non-Karaf environments. E.g. it won't work when running the Demo App in Eclipse. To fix that you can either use bnd or Maven to embed a non-OSGi dependency into your bundle. See for instance the `org.openhab.core.audio` bundle for embedding a dependency in your bundle using Maven.

pravussum:
I think this is in a state where it can be reviewed. Not much action to be seen without the addon PR, though.

kaikreuzer (approved):
Many thanks @pravussum, this looks awesome to me!
From my perspective, it's ready to merge.
Since I have never worked with micrometer so far, I'd be happy for a second review by @wborn or @cweitkamp, though.

kaikreuzer:
@openhab/core-maintainers If you're not too familiar with it yourself and don't find the time for a review here, it is absolutely fine and I'd just go ahead and merge.

cweitkamp:
I am afraid I will have now time to test it now. Fell free to merge this PR. I will test it afterwards in my prod environment.

kaikreuzer:
Fair enough, let's merge then - thanks @pravussum, looking forward to your work on the add-on side!

PR #2141, opened 2021-01-21
Bump apache.commons.net to 3.7.2

### LINKED ISSUE #2140: Old Apache-commons-net version leads to NPEs
A disposal `Thinghandler` instance for a Smartmeter `Thing` connected via RFC2217 serial implementation leads to:

```
2021-01-16 20:47:01.832 [ERROR] [ding.smartmeter.internal.MeterDevice] - Fatal error occured
        at org.apache.commons.net.telnet.TelnetClient._closeOutputStream(TelnetClient.java:84) ~[?:?]
        at org.apache.commons.net.telnet.TelnetOutputStream.close(TelnetOutputStream.java:163) ~[?:?]
        at org.apache.commons.net.telnet.TelnetClient.disconnect(TelnetClient.java:129) ~[?:?]
        at gnu.io.rfc2217.TelnetSerialPort.close(TelnetSerialPort.java:321) ~[?:?]
        at org.openhab.core.io.transport.serial.rxtx.RxTxSerialPort.close(RxTxSerialPort.java:51) ~[?:?]
        at org.openmuc.jrxtx.JRxTxPort.close(JRxTxPort.java:159) ~[bundleFile:?]
        at org.openmuc.j62056.Iec21Port.close(Iec21Port.java:285) ~[bundleFile:?]
        at org.openhab.binding.smartmeter.internal.iec62056.Iec62056_21SerialConnector.closeConnection(Iec62056_21SerialConnector.java:132) ~[bundleFile:?]
        at io.reactivex.internal.operators.flowable.FlowableDoOnLifecycle$SubscriptionLambdaSubscriber.cancel(FlowableDoOnLifecycle.java:112) [bundleFile:?]
        at io.reactivex.internal.subscribers.BasicFuseableSubscriber.cancel(BasicFuseableSubscriber.java:158) [bundleFile:?]
        at io.reactivex.internal.subscriptions.SubscriptionHelper.cancel(SubscriptionHelper.java:189) [bundleFile:?]
        at io.reactivex.internal.operators.flowable.FlowablePublish$PublishSubscriber.dispose(FlowablePublish.java:183) [bundleFile:?]
        at io.reactivex.internal.disposables.DisposableHelper.dispose(DisposableHelper.java:124) [bundleFile:?]
        at io.reactivex.internal.operators.flowable.FlowableRefCount.timeout(FlowableRefCount.java:140) [bundleFile:?]
        at io.reactivex.internal.operators.flowable.FlowableRefCount.cancel(FlowableRefCount.java:107) [bundleFile:?]
        at io.reactivex.internal.operators.flowable.FlowableRefCount$RefCountSubscriber.cancel(FlowableRefCount.java:229) [bundleFile:?]
        at io.reactivex.internal.subscriptions.SubscriptionArbiter.drainLoop(SubscriptionArbiter.java:218) [bundleFile:?]
        at io.reactivex.internal.subscriptions.SubscriptionArbiter.drain(SubscriptionArbiter.java:187) [bundleFile:?]
        at io.reactivex.internal.subscriptions.SubscriptionArbiter.cancel(SubscriptionArbiter.java:179) [bundleFile:?]
        at io.reactivex.internal.operators.flowable.FlowableRepeatWhen$WhenSourceSubscriber.cancel(FlowableRepeatWhen.java:174) [bundleFile:?]
        at io.reactivex.internal.subscriptions.SubscriptionHelper.cancel(SubscriptionHelper.java:189) [bundleFile:?]
        at io.reactivex.internal.operators.flowable.FlowableSubscribeOn$SubscribeOnSubscriber.cancel(FlowableSubscribeOn.java:141) [bundleFile:?]
        at io.reactivex.internal.subscriptions.SubscriptionHelper.cancel(SubscriptionHelper.java:189) [bundleFile:?]
        at io.reactivex.internal.subscribers.LambdaSubscriber.cancel(LambdaSubscriber.java:119) [bundleFile:?]
        at io.reactivex.internal.subscribers.LambdaSubscriber.dispose(LambdaSubscriber.java:104) [bundleFile:?]
        at org.openhab.binding.smartmeter.internal.SmartMeterHandler.cancelRead(SmartMeterHandler.java:139) [bundleFile:?]
        at org.openhab.binding.smartmeter.internal.SmartMeterHandler.dispose(SmartMeterHandler.java:128) [bundleFile:?]
        at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:?]
        at jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[?:?]
        at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:?]
        at java.lang.reflect.Method.invoke(Method.java:566) ~[?:?]
        at org.openhab.core.internal.common.AbstractInvocationHandler.invokeDirect(AbstractInvocationHandler.java:154) [bundleFile:?]
        at org.openhab.core.internal.common.Invocation.call(Invocation.java:52) [bundleFile:?]
        at java.lang.Thread.run(Thread.java:834) [?:?]
```

Thus, this connection becomes a zombie connection that will eventually timeout.

There is a fix for this in apache commons net: https://github.com/apache/commons-net/commit/98d829eb6af5449ac27b66382952de37602ebe4e

and a related fix here:

https://github.com/apache/commons-net/commit/796a288e0c2e17ddf7dc44a4cd790a7eab4e158d

which are both > 3 years old.

The apache-commons-net library was added to openHAB 2 years ago: https://github.com/openhab/openhab-core/pull/694

But the decision was done to use a (at that time already) outdated version.

Is there a reason why we are not using the most recent (3.7.2) one, which is available here? https://mvnrepository.com/artifact/commons-net/commons-net/3.7.2

If there is none, please accept the PR to bump up the version to fix the issue.

t2000:
Fixes #2140

Signed-off-by: Stefan Triller <github@stefantriller.de>

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -413,7 +413,7 @@
     <dependency>
       <groupId>commons-net</groupId>
       <artifactId>commons-net</artifactId>
-      <version>3.3</version>
+      <version>3.7.2</version>
       <scope>compile</scope>
     </dependency>
 
--- a/features/karaf/openhab-tp/src/main/feature/feature.xml
+++ b/features/karaf/openhab-tp/src/main/feature/feature.xml
@@ -44,8 +44,8 @@
 	</feature>
 
 	<feature name="openhab.tp-commons-net" description="The Apache Commons Net library" version="${project.version}">
-		<capability>openhab.tp;feature=commons-net;version=3.3</capability>
-		<bundle dependency="true">mvn:commons-net/commons-net/3.3</bundle>
+		<capability>openhab.tp;feature=commons-net;version=3.7.2</capability>
+		<bundle dependency="true">mvn:commons-net/commons-net/3.7.2</bundle>
 	</feature>
 
 	<feature name="openhab.tp-gson" description="Gson" version="${project.version}">


cweitkamp:
I strongly vote for removing the `apache.commons.net` dependency from OHC feature file and include it in the add-ons POM and feature files.

There are already two bindings using a newer version (3.6): [network](https://github.com/openhab/openhab-addons/blob/7d1c6ee37124116f009dc00b9d93b49673c20c49/bundles/org.openhab.binding.network/src/main/feature/feature.xml) and [heos](https://github.com/openhab/openhab-addons/blob/7d1c6ee37124116f009dc00b9d93b49673c20c49/bundles/org.openhab.binding.heos/src/main/feature/feature.xml).

The library itself is not used by OHC anymore and we want to remove it soon (see https://github.com/openhab/openhab-addons/issues/7722).

t2000:
Well, what do we do then? The bundle `org.openhab.core.io.transport.serial.rxtx.rfc2217` in the core requires `apache.commons.net`.

You are saying to move the dependency into the dependencies from that bundle, correct?

But then you also say
>The library itself is not used by OHC anymore and we want to remove it soon (see openhab/openhab-addons#7722).

if you are going to remove it, then we also should not move the dependency into `org.openhab.core.io.transport.serial.rxtx.rfc2217`, right?

So what should we do instead? Come up with our own implementation for RFC2217?

cweitkamp:
> The bundle `org.openhab.core.io.transport.serial.rxtx.rfc2217` in the core requires `apache.commons.net`.

Are you sure? Can you point me to the lines of code where you found that?

cweitkamp:
Okay, I found it. It is a dependency of [com.neuronrobotics/nrjavaserial](https://github.com/NeuronRobotics/nrjavaserial/blob/6a97d715333cf73ffbf97a218438b916fb9dcc7c/build.gradle#L45) which needs at least version 3.3. In that case you are right and we probably cannot remove it.

t2000:
Yes, correct. nrjavaserial needs it. I have added the dependency to the serial bundle and left the dependency in the core bundle as it is (until it will be removed).
Since I don't know much about the dependency hell, especially in combination with karaf, please let me know if I forgot anything. Thanks.

bundles/org.openhab.core.io.transport.serial/pom.xml
```diff
@@ -19,6 +19,12 @@
       <artifactId>org.openhab.core.io.console</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>commons-net</groupId>
+      <artifactId>commons-net</artifactId>
+      <version>3.7.2</version>
+      <scope>compile</scope>
+    </dependency>
```
wborn, line 27: This will never work. You cannot manage dependencies of one bundle through another. The only way to make sure 3.7.2 is used by RXTX is by changing the Karaf feature files and making sure that only 3.7.2 of commons-net is installed. The RXTX bundle will use any available commons-net bundle that satisfies its specified version range. So it may even use 3.3 when it is installed eventhough 3.7.2 is available.

wborn (changes requested):

bundles/org.openhab.core.io.transport.serial/pom.xml
t2000: Thanks for the explanation, I already feared that I need to touch the karaf stuff as well. Since I dont know much about karaf, I have it a try but obviously I failed :P

So do we have to wait for https://github.com/NeuronRobotics/nrjavaserial/issues/207 that I have just opened? Because I think think might take years again until we get a new release (from experiences in the past). Or is there any other option? I.e. that we leave nrjavaserial as it is, like compiling against 3.3, but at runtime only have 3.7.2 available and thus the bug is fixed for us?

It is really annoying to work with serial ports in openHAB at the moment. I would like to concentrate on working on features and improving stuff, rather than always getting obstructed by randomly working/not working thinghandlers because there are port available or not.

wborn: > Because I think think might take years again until we get a new release (from experiences in the past).

Since the dependency change has already been merged now, I think there will be a release in time before OH 3.1.

> Or is there any other option?

If you update all core/add-on bundles/features to use commons-net 3.7+ you can guarantee that the current nrjavaserial version will also use the updated version. So that means you need to update this PR and probably create a PR for add-ons too.

t2000: >I think there will be a release in time before OH 3.1.
That would be great. Let's hope for it :)

>If you update all core/add-on bundles/features to use commons-net 3.7+ you can guarantee that the current nrjavaserial version will also use the updated version. So that means you need to update this PR and probably create a PR for add-ons too.

For now I gave this a try, since we found out that we cannot remove the commons.net dependency due to nrjavaserial. So I adusted this PR and also created https://github.com/openhab/openhab-addons/pull/9921 to update the addons.

I guess it doesn't hurt to have it set to the latest version and if nrjavaserial bumps up its version to get compiled against commons-net-3.7.2, we can bump up nrjavaserial in addition.

Please let me know if I forgot anything.

wborn (approved):
Looks good to me now! We probably also want to exclude commons-net from all core dependencies. That way we can prevent new usages being added in add-ons. But let's do that another time. :wink:

cweitkamp (approved):
Thanks for your work and bumping the version in nrjavaserial too.

@wborn Yes, I agreed. We should exclude it from other bundles. I additionally suggest to remove the optional dependency from openhab-base feature. Add-ons which depend on it should include it by themselves.

PR #2144, opened 2021-01-22
[discovery.upnp] Devices may apply a grace period for removal from the Inbox

andrewfg:
**BACKGROUND**

The JUPnP library strictly follows the UPnP specification insofar as if a device fails to send its next 'ssdp:alive' notification within its declared 'maxAge' period, it is immediately considered to be gone. Unfortunately some types of OpenHAB bindings have devices that can sometimes be a bit late in sending their 'ssdp:alive' notifications even though they have not really gone offline (e.g. the Philips Hue bridge), which means that such devices are repeatedly removed from, and (re)added to, the InBox. This leads to confusion in the UI, and repeated logger messages.

**SOLUTION**

To prevent this, when the OH core receives a remoteDeviceRemoved() notification from JUPnP it asks the binding if it should immediately remove the device from the Inbox, or alternatively wait for a further period of time ('grace period') before actually so doing. (And if the device 're-appears' within the grace period, its removal task is cancelled).

**REFERENCES**

Refers to following Issue..
https://github.com/openhab/openhab-addons/issues/9357

**NOTES**

NOTE: this PR extends the UpnpDiscoveryParticipant interface, by declaring a new 'default' method. This 'default' method means that there is no impact on bindings that have been compiled against on prior versions of the interface. Furthermore the return value of the 'default' implementation (0) is such that the behaviour of existing bindings will be unchanged versus prior implementations. Only those bindings that OPTIONALLY choose to override the default method return value will benefit from the enhanced performance of this PR.

### DIFF
--- a/bundles/org.openhab.core.config.discovery.upnp/src/main/java/org/openhab/core/config/discovery/upnp/UpnpDiscoveryParticipant.java
+++ b/bundles/org.openhab.core.config.discovery.upnp/src/main/java/org/openhab/core/config/discovery/upnp/UpnpDiscoveryParticipant.java
@@ -62,4 +62,21 @@ public interface UpnpDiscoveryParticipant {
      */
     @Nullable
     ThingUID getThingUID(RemoteDevice device);
+
+    /**
+     * The JUPnP library strictly follows the UPnP specification insofar as if a device fails to send its next
+     * 'ssdp:alive' notification within its declared 'maxAge' period, it is immediately considered to be gone. But
+     * unfortunately some types of OpenHAB bindings have devices that can sometimes be a bit late in sending their
+     * 'ssdp:alive' notifications even though they have not really gone offline, which means that such devices are
+     * repeatedly removed from, and (re)added to, the InBox.
+     *
+     * To prevent this, a binding that implements a UpnpDiscoveryParticipant may OPTIONALLY implement this method to
+     * specify an additional delay period (grace period) to wait before the device is removed from the Inbox.
+     *
+     * @param device the upnp device on the network
+     * @return the additional grace period delay in seconds before the device will be removed from the Inbox
+     */
+    default int getRemovalGracePeriodSeconds(RemoteDevice device) {
+        return 0;
+    }
 }
--- a/bundles/org.openhab.core.config.discovery.upnp/src/main/java/org/openhab/core/config/discovery/upnp/UpnpDiscoveryParticipant.java
+++ b/bundles/org.openhab.core.config.discovery.upnp/src/main/java/org/openhab/core/config/discovery/upnp/UpnpDiscoveryParticipant.java
@@ -66,14 +66,14 @@ public interface UpnpDiscoveryParticipant {
     /**
      * The JUPnP library strictly follows the UPnP specification insofar as if a device fails to send its next
      * 'ssdp:alive' notification within its declared 'maxAge' period, it is immediately considered to be gone. But
-     * unfortunately some types of OpenHAB bindings have devices that can sometimes be a bit late in sending their
-     * 'ssdp:alive' notifications even though they have not really gone offline, which means that such devices are
-     * repeatedly removed from, and (re)added to, the InBox.
+     * unfortunately some openHAB bindings handle devices that can sometimes be a bit late in sending their 'ssdp:alive'
+     * notifications even though they have not really gone offline, which means that such devices are repeatedly removed
+     * from, and (re)added to, the Inbox.
      *
      * To prevent this, a binding that implements a UpnpDiscoveryParticipant may OPTIONALLY implement this method to
      * specify an additional delay period (grace period) to wait before the device is removed from the Inbox.
      *
-     * @param device the upnp device on the network
+     * @param device the UPnP device on the network
      * @return the additional grace period delay in seconds before the device will be removed from the Inbox
      */
     default long getRemovalGracePeriodSeconds(RemoteDevice device) {
--- a/bundles/org.openhab.core.config.discovery.upnp/src/main/java/org/openhab/core/config/discovery/upnp/UpnpDiscoveryParticipant.java
+++ b/bundles/org.openhab.core.config.discovery.upnp/src/main/java/org/openhab/core/config/discovery/upnp/UpnpDiscoveryParticipant.java
@@ -76,7 +76,7 @@ public interface UpnpDiscoveryParticipant {
      * @param device the upnp device on the network
      * @return the additional grace period delay in seconds before the device will be removed from the Inbox
      */
-    default int getRemovalGracePeriodSeconds(RemoteDevice device) {
+    default long getRemovalGracePeriodSeconds(RemoteDevice device) {
         return 0;
     }
 }
--- a/bundles/org.openhab.core.config.discovery.upnp/src/main/java/org/openhab/core/config/discovery/upnp/internal/UpnpDiscoveryService.java
+++ b/bundles/org.openhab.core.config.discovery.upnp/src/main/java/org/openhab/core/config/discovery/upnp/internal/UpnpDiscoveryService.java
@@ -13,11 +13,11 @@
 package org.openhab.core.config.discovery.upnp.internal;
 
 import java.util.Collection;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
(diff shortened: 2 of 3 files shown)


lolodomo:
What is a typical value for max age? 
Is it set for all UPnP devices? 
Is it not better to define our delay, a small delay and the same for all devices? 

I am also asking myself if we should make this delay something optional and settable by any binding. By default, 0 will be considered. It wiould simply require a new method in the API + a change in the hue binding to override it.

lolodomo:
My idea is to not force a general change on the UPnP stuff just for one or two devices not respecting well the UPnP standard.

andrewfg:
> What is a typical value for max age?
> Is it set for all UPnP devices? 

Each UPNP device (type) defines its own (different) maxAge. The only thing fixed is that the UPnP specification defines the largest allowed (resp. default) value of maxAge as 1800 seconds. The Hue Bridge for example uses 100 seconds.

> I am also asking myself if we should make this delay something optional and settable by any binding.
> It would simply require a new method in the API + a change in the hue binding to override it.
> not force a general change on the UPnP stuff just for one or two devices not respecting well the UPnP standard.

Yes I agree that this would be a better solution. Later today I will make another commit with an alternate proposed solution.

andrewfg:
@lolodomo the grace period is now defined in the API interface via a default method whose default value is 0 seconds..

bundles/org.openhab.core.config.discovery.upnp/src/main/java/org/openhab/core/config/discovery/upnp/internal/UpnpDiscoveryService.java
```diff
@@ -65,6 +69,11 @@ public UpnpDiscoveryService() {
 
     private UpnpService upnpService;
 
+    /*
+     * Map of scheduled tasks to remove devices from the Inbox
+     */
+    private Map<UDN, Future<?>> deviceRemovalTasks = new HashMap<>();
```
lolodomo, line 75: Maybe rather a ConcurrentHashMap ?

andrewfg: Yes.

lolodomo (approved):
LGTM

andrewfg:
Just so you know, I have tested this code with the Hue binding over 24 hours now; with the following (positive) results..

- The bridge Thing remained fully 'online' and functional during this 24 hour test period.
- However it sent 'ssdp:alive' notifications late at a rate of ~2 times / hour (varied between 0..9 times/hr).
- Without a grace period, this results in 48 spurious Inbox removals, 48 spurious re-addings, and 48 spurious log messages.
- Applying a 50 second grace period ('maxAge' / 2), results in zero spurious Inbox removals, re-addings, and log messages.

andrewfg:
Adding @cweitkamp to the distribution since you are a Hue expert

andrewfg:
And (FYI) when/if this PR is approved, I would also submit the following corresponding PR for the Hue binding..

https://github.com/openhab/openhab-addons/compare/main...andrewfg:hue-grace_period

lolodomo:
50 seconds looks enormous. Could you try with 5, maybe 10 ?

andrewfg:
> 50 seconds looks enormous. Could you try with 5, maybe 10 ?

My logic is that if a device misses one notification '1 x maxAge' it should be overlooked. But if it misses '2 x maxAge' it has definitely gone. So setting a threshold at '1.5 x maxAge' seems totally reasonable to me. Thus the grace period of '0.5 x maxAge'..

bundles/org.openhab.core.config.discovery.upnp/src/main/java/org/openhab/core/config/discovery/upnp/UpnpDiscoveryParticipant.java
```diff
@@ -62,4 +62,21 @@
...
+     * 'ssdp:alive' notifications even though they have not really gone offline, which means that such devices are
+     * repeatedly removed from, and (re)added to, the InBox.
+     *
+     * To prevent this, a binding that implements a UpnpDiscoveryParticipant may OPTIONALLY implement this method to
+     * specify an additional delay period (grace period) to wait before the device is removed from the Inbox.
+     *
+     * @param device the upnp device on the network
+     * @return the additional grace period delay in seconds before the device will be removed from the Inbox
+     */
+    default int getRemovalGracePeriodSeconds(RemoteDevice device) {
```
cweitkamp, line 79: Two questions:

1. Why do we need the parameter `RemoteDevice`? You are neither using it here nor in https://github.com/openhab/openhab-addons/pull/9985.

2. Shouldn't we return a `long` value instead of an `int`? The `delay` parameter in `schedule(...)` is a `long` too.

cweitkamp (commented):
Thanks for you analysis and looking for a solution.

bundles/org.openhab.core.config.discovery.upnp/src/main/java/org/openhab/core/config/discovery/upnp/UpnpDiscoveryParticipant.java
andrewfg: @kaikreuzer thanks for your review. Concerning your two points..

1. Regarding the extra argument: I thought that there could be cases in future where the discovery service might be handling more than one class of device; so it would allow the binding to have different gracePeriods for different device types. But the Hue discovery only handles the Hue Bridge so the argument is not used..
2. Re long vs int: ok I will fix that.

cweitkamp: I am not @kaikreuzer , but I take it as a compliment. 😉

andrewfg: >I am not @kaikreuzer , but I take it as a compliment

Sorry for that :( 
But I edited the post to show your handle now :)

andrewfg:
I will need to add a paragraph to the developer guide [ReadMe](https://www.openhab.org/docs/developer/bindings/#upnp-discovery). -- How does one do this? It is on a different repo, so presumably it needs a different PR for that?

cweitkamp (commented):
> How does one do this? It is on a different repo, so presumably it needs a different PR for that? 

Yes, we have a dedicated repository for the documentation. Check on https://github.com/openhab/openhab-docs/blob/main/developers/bindings/index.md#upnp-discovery.

I afraid I missed some spellings in my first review.

bundles/org.openhab.core.config.discovery.upnp/src/main/java/org/openhab/core/config/discovery/upnp/UpnpDiscoveryParticipant.java
```diff
@@ -62,4 +62,21 @@
      */
     @Nullable
     ThingUID getThingUID(RemoteDevice device);
+
+    /**
+     * The JUPnP library strictly follows the UPnP specification insofar as if a device fails to send its next
+     * 'ssdp:alive' notification within its declared 'maxAge' period, it is immediately considered to be gone. But
+     * unfortunately some types of OpenHAB bindings have devices that can sometimes be a bit late in sending their
```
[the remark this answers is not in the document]
andrewfg: Done.

bundles/org.openhab.core.config.discovery.upnp/src/main/java/org/openhab/core/config/discovery/upnp/UpnpDiscoveryParticipant.java
```diff
@@ -62,4 +62,21 @@
...
+     * 'ssdp:alive' notifications even though they have not really gone offline, which means that such devices are
+     * repeatedly removed from, and (re)added to, the InBox.
```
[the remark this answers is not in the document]
andrewfg: Done.

bundles/org.openhab.core.config.discovery.upnp/src/main/java/org/openhab/core/config/discovery/upnp/UpnpDiscoveryParticipant.java
```diff
@@ -62,4 +62,21 @@
...
+     *
+     * To prevent this, a binding that implements a UpnpDiscoveryParticipant may OPTIONALLY implement this method to
+     * specify an additional delay period (grace period) to wait before the device is removed from the Inbox.
+     *
+     * @param device the upnp device on the network
```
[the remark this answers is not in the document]
andrewfg: Done.

andrewfg:
> we have a dedicated repository for the documentation

New PR added for the documentation https://github.com/openhab/openhab-docs/pull/1478

andrewfg:
NB the ReadMe PR has already been merged :)

cweitkamp (approved):
LGTM. Thank you.

PR #2147, opened 2021-01-23
[timer] Add Timer.getExecutionTime()

jimtng:
This will return a ZonedDateTime of the scheduled execution time. This will allow the timer logic to determine how much extension (via reschedule) a timer would need, based on the current timer schedule.

It will return null if the timer has been cancelled.

It will return a time in the past if the timer has terminated.


kaikreuzer (approved):
lgtm, indeed a useful addition 👍

PR #2151, opened 2021-01-24
Stop commons-net proliferation

wborn:
* Exclude commons-net from core dependencies so it is no longer automatically a transitive compile dependency
* Update nrjavaserial compile dependency to a version that no longer includes commons-net packages
* Define and use commons.net.version property
* Rework features so commons-net is only installed when required

---

This stops commons-net from being automatically available on the compile classpath and in features.
That way we can stop it from being used unintentionally.

Related to:

* https://github.com/openhab/openhab-core/pull/2141
* https://github.com/openhab/openhab-addons/pull/9921

### DIFF
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -187,12 +187,24 @@
       <artifactId>javaxcomm</artifactId>
       <version>1.0.1</version>
       <scope>compile</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>commons-net</groupId>
+          <artifactId>commons-net</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>com.neuronrobotics</groupId>
       <artifactId>nrjavaserial</artifactId>
-      <version>3.12.0</version>
+      <version>3.14.0</version>
       <scope>compile</scope>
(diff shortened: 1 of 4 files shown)


bom/compile/pom.xml
```diff
@@ -187,12 +187,24 @@
...
+          <groupId>commons-net</groupId>
+          <artifactId>commons-net</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>com.neuronrobotics</groupId>
       <artifactId>nrjavaserial</artifactId>
-      <version>3.12.0</version>
+      <version>3.14.0</version>
```
t2000, line 200: Why not nrjavaserial `5.2.1` ?

t2000 (commented):
One small question from my side.

bom/compile/pom.xml
wborn: There is a difference between compile and runtime dependencies. The compile time dependencies are used by bnd to compute version ranges for package dependencies. So you want these computed ranges to be as big as possible so you do not unnecessarily force users of the framework to upgrade their bundles.

cweitkamp (approved):
LGTM. Thanks.

PR #2153, opened 2021-01-24
[timer] Fix to allow timer reschedule after termination

### LINKED ISSUE #2131: [OH3] Rescheduling timer failed
Reschedule timer ends with warning log and no action:
`[WARN ] [el.script.internal.actions.TimerImpl] - Rescheduling failed as execution has already started!`

Documentation says this should be possible: 
`Reschedules a timer to a new starting time. This can also be called after a timer has terminated, which will result in another execution of the same code.`
source: https://openhab.org/javadoc/latest/org/openhab/core/model/script/actions/timer#reschedule(java.time.ZonedDateTime)

Example rule:
```javascript
var Timer testTimer

rule "Reschedule test"
when
    Item TestSwitch changed
then
    logInfo("RescheduleTest", "TestSwitch changed")

    if (testTimer !== null)
    {
        logInfo("RescheduleTest", "Timer state: " + testTimer.hasTerminated + testTimer.isRunning + testTimer.isActive)

        testTimer.reschedule(now.plusSeconds(10))
    }
    else
        testTimer = createTimer(now.plusSeconds(10), [|
            logInfo("RescheduleTest", "Test timer fired")])
end
```

Test switch:
```
Switch  TestSwitch
```

Related discussion: https://community.openhab.org/t/oh3-timer-reschedule-failed/112912
Related source code: https://github.com/openhab/openhab-core/blob/774b9d607b8277f809c2b2b1d94da1be9c23156d/bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/internal/actions/TimerImpl.java#L55

jimtng:
Fix #2131

### DIFF
--- a/bundles/org.openhab.core.model.script/src.moved/test/java/org/openhab/core/model/script/internal/actions/TimerImplTest.java
+++ b/bundles/org.openhab.core.model.script/src.moved/test/java/org/openhab/core/model/script/internal/actions/TimerImplTest.java
@@ -73,6 +73,21 @@ public class TimerImplTest {
     }
 
     @Test
+    public void testTimerHasTerminatedAndReschedule() throws InterruptedException {
+        Thread.sleep(TimeUnit.SECONDS.toMillis(DEFAULT_TIMEOUT_SECONDS + DEFAULT_RUNTIME_SECONDS + 1));
+        assertThat(subject.isActive(), is(false));
+        assertThat(subject.hasTerminated(), is(true));
+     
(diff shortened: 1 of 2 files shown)


Mywk (approved):
Seems to fix it for me.

kaikreuzer (approved):
Interesting that the JavaDoc and the implementation showed such a mismatch.
But I don't see any reason against this feature, so I agree to keep the doc and adapt the implementation. Thanks for the PR!

PR #2157, opened 2021-01-25
Added system default channel type to control color temperature in Kelvin

cweitkamp:
- Added system default channel type to control color temperature in Kelvin

See https://github.com/openhab/openhab-addons/pull/9939#discussion_r563328208


bundles/org.openhab.core.thing/src/main/resources/OH-INF/i18n/DefaultSystemChannels_fr.properties
```diff
@@ -16,7 +16,9 @@ channel-type.system.brightness.description = Contr
 channel-type.system.color.label = Couleur
 channel-type.system.color.description = Contr�le la couleur de la lumi�re.
 channel-type.system.color-temperature.label = Temp�rature de couleur
-channel-type.system.color-temperature.description = Contr�le la temp�rature de couleur de la lumi�re.
+channel-type.system.color-temperature.description = Contr�le la temp�rature de couleur de la lumi�re de 0 (froid) � 100 (chaud).
```
cweitkamp, line 19: //CC @lolodomo For approving my French skills 😉 .

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/DefaultSystemChannelTypeProvider.java
```diff
@@ -173,13 +173,23 @@
...
+            .withDescription("Controls the color temperature of the light from 0 (cold) to 100 (warm)")
+            .withCategory("ColorLight").withStateDescriptionFragment(StateDescriptionFragmentBuilder.create()
+                    .withMinimum(BigDecimal.ZERO).withMaximum(new BigDecimal(100)).withPattern("%d").build())
+            .build();
+
+    /**
+     * Color-temperature: default system wide {@link ChannelType} which allows changing the color temperature in Kelvin
+     */
+    public static final ChannelType SYSTEM_COLOR_TEMPERATURE_ABS = ChannelTypeBuilder
+            .state(new ChannelTypeUID(BINDING_ID, "color-temperature-abs"), "Color Temperature", "Number")
```
J-N-K, line 189: I‘m unsure if we should support UoM here. In general it is a quantity with a unit. OTOH I have never seen someone using another unit than K.

kaikreuzer: Nobody is sure here, my tendency goes to what @cweitkamp did here, see https://github.com/openhab/openhab-addons/pull/9939#discussion_r563350083.

J-N-K: I was thinking the same, that’s why it is implemented in deconz that way. The only other unit I know is mired (which is used in Zigbee internally), but I have never seen it user-facing. So a plain `Number` is fine with me, too.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/DefaultSystemChannelTypeProvider.java
```diff
@@ -173,13 +173,23 @@
...
+     * Color-temperature: default system wide {@link ChannelType} which allows changing the color temperature in percent
      */
     public static final ChannelType SYSTEM_COLOR_TEMPERATURE = ChannelTypeBuilder
             .state(new ChannelTypeUID(BINDING_ID, "color-temperature"), "Color Temperature", "Dimmer")
-            .withDescription("Controls the color temperature of the light").withCategory("ColorLight")
-            .withStateDescriptionFragment(StateDescriptionFragmentBuilder.create().withMinimum(BigDecimal.ZERO)
-                    .withMaximum(new BigDecimal(100)).withPattern("%d").build())
+            .withDescription("Controls the color temperature of the light from 0 (cold) to 100 (warm)")
+            .withCategory("ColorLight").withStateDescriptionFragment(StateDescriptionFragmentBuilder.create()
+                    .withMinimum(BigDecimal.ZERO).withMaximum(new BigDecimal(100)).withPattern("%d").build())
```
kaikreuzer, line 182: wouldn't it be safer to also use `%.0f` here, since PercentTypes are floating point numbers?

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/DefaultSystemChannelTypeProvider.java
```diff
@@ -173,13 +173,23 @@
...
+            .build();
+
+    /**
+     * Color-temperature: default system wide {@link ChannelType} which allows changing the color temperature in Kelvin
+     */
+    public static final ChannelType SYSTEM_COLOR_TEMPERATURE_ABS = ChannelTypeBuilder
+            .state(new ChannelTypeUID(BINDING_ID, "color-temperature-abs"), "Color Temperature", "Number")
+            .withDescription("Controls the color temperature of the light in Kelvin").withCategory("ColorLight")
+            .withStateDescriptionFragment(StateDescriptionFragmentBuilder.create().withMinimum(BigDecimal.ONE)
+                    .withMaximum(new BigDecimal(1000000)).withPattern("%.0f K").build())
```
kaikreuzer, line 192: Does 0-1.000.000 really make sense here?
Looking at https://www.lightingdesignlab.com/resources/articles/articles-lighting-fundamentals/color-temperature, I'd think a range from 1.000-10.000 might make sense, especially in case people want to attach a selection box or slider to it?

kaikreuzer (changes requested):

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/DefaultSystemChannelTypeProvider.java
cweitkamp: 1-1000000 is the possible range for it. But the visible spectrum for humans is from 1000-10000. I did not want to add to strict limits but it makes sense to only use what is most useful wrt to home automation.

bundles/org.openhab.core.thing/src/main/resources/OH-INF/i18n/DefaultSystemChannels_fr.properties
lolodomo: That's ok

kaikreuzer (approved):
thanks!

PR #2160, opened 2021-01-27
RFC: thing type versioning

J-N-K:
This is another approach on solving the issue with changing channel configurations on thing-types.

For a thing-type `myThing` the `myThing.update` in the resources folder would look like

```
1,ADD_CHANNEL,color,Color,system-channel.color,Color Control
1,REMOVE_CHANNEL,toberemovedchannelid
```

Since all code is in the BaseThingHandler, there is noneed for additional code within bindings, only the `*.update` files need to be managed by the developer.


J-N-K:
https://github.com/openhab/openhab-addons/pull/9856 Could be solved easily with this proposal.

Issue #2091, opened 2021-01-09
Changing temperature via Web BasicUI after restart fails unless done in App


chris922:
I migrated my whole setup to openHAB 3 and noticed a strange issue:

When I am using the Web BasicUI and try to change a Setpoint that is linked to a Number:Temperature Item it doesn't work.
But when I change this item via openHAB Android App then I am afterwards able to change it via Web BasicUI as well.

I checked the network logs from my browser and found out, that the first requests after a restart of OH via Web BasicUI fails with a status 400.
The request payload is: `20.5 %unit%`
After changing the value via Android App and then changing via Web BasicUI the request payload changed and is now: `21 °C`
So first change failed while using `%unit%`, afterwards it works when `°C` is used.
When I restart openHAB the issue is back again.

As mentioned I am using openHAB 3.0 via openHABian on a RPi4.

Item:
```
Number:Temperature EG_Bad_Heizung_Temperatur_Soll     "Komfort Soll-Temperatur [%.1f °C]"  <heating_temp> (agEG_Bad_Heizung_Thermostat) { channel="zwave:device:controller:node43:thermostat_setpoint_heating", alexa="ThermostatController.targetSetpoint" }
```
Sitemap:
```
Setpoint item=EG_Bad_Heizung_Temperatur_Soll     minValue=16 maxValue=25 step=0.5
```

chris922:
Seems to be the same issue as already reported here: https://github.com/openhab/openhab-webui/issues/765

I will close this one

chris922:
I will reopen this issue, because as mentioned in https://github.com/openhab/openhab-webui/issues/765 it looks like the root cause of the issue is in the core.ui lib. You can also find a PR for this: #2165

PR #2165, opened 2021-01-30
Fix changing temperature via BasicUI (#2091)

chris922:
As described in https://github.com/openhab/openhab-webui/issues/765 changing temperature via BasicUI fails as long as the widget in the sitemap doesn't have a label defined.

I debugged the issue on my openHAB instance (v3.0.0) and saw that the unit-symbol of the Item will not be used in case no label was defined and instead of this `null` will be returned.

Therefore I added a small check that will ensure that the unit-symbol of the item will be returned as long as no label is given.

Unfortunately I don't know how to test the changes, I tried replacing the corresponding JAR in the runtime directory, but seems that it doesn't work. If someone could give me some hints how I can check this in my running openHAB instance please let me know.

Anyway I added some unit-tests based on the information I was able to gather during debugging.

Fixes https://github.com/openhab/openhab-webui/issues/765

### DIFF
--- a/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java
+++ b/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java
@@ -1263,6 +1263,11 @@ public class ItemUIRegistryImpl implements ItemUIRegistry {
 
             // we require the item to define a dimension, otherwise no unit will be reported to the UIs.
             if (item instanceof NumberItem && ((NumberItem) item).getDimension() != null) {
+                if (w.getLabel() == null) {
+                    // if no Label was assigned to the Widget we fallback to the items unit
+                    return ((NumberItem) item).getUnitSymbol();
+                }
+
                 String unit = getUnitFromLabel(w.getLabel());
                 if (!UnitUtils.UNIT_PLACEHOLDER.equals(unit)) {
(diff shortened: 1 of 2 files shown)


andrewfg:
> don't know how to test the changes, I tried replacing the corresponding JAR in the runtime directory

@chris922 it is some time since I last did it but as I recall you need to drop the jar file in the addons folder and then use the console `bundle:install` command..

https://www.openhab.org/docs/administration/console.html#using-the-console

chris922:
@andrewfg : I know how to do it for bundles like additional bindings.. but this here are changes at the openHAB core, so I am unsure if it is still the same way or if I will destroy my setup when I try to uninstall/remove something there to drop in this new JAR :thinking:

andrewfg:
> I know how to do it for bundles like additional bindings

Its is a bit different for bindings and core jars..
- for bindings: you just drop the jar in the addons folder
- core jars: : you drop it in the addons folder, and then run `bundle:install`

andrewfg:
^
PS and to reverse it you do `bundle:uninstall`

andrewfg:
^
PPS and experiment with `bundle:list` first

chris922:
Ahh nice, perfect. I stopped the bundle `openHAB Core :: Bundles :: UI`, dropped in my new JAR and started this bundle from my JAR.. aaaand: Issue seems to be solved, I can change the temperature using the BasicUI via Web :) 

Thanks @andrewfg for supporting me how to test it

PR #2183, opened 2021-02-07
Make absolute color temperature channel type advanced

kaikreuzer:
Currently, Things that support this channel type always show two color temperature channels, which looks weird and distracting to the user:

Screenshot 2021-02-07 at 21 35 51

As the absolute channel is an optional feature for advanced users and use cases, it should imho also only appear as an advanced channel.

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/DefaultSystemChannelTypeProvider.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/DefaultSystemChannelTypeProvider.java
@@ -188,8 +188,8 @@ public class DefaultSystemChannelTypeProvider implements ChannelTypeProvider {
     public static final ChannelType SYSTEM_COLOR_TEMPERATURE_ABS = ChannelTypeBuilder
             .state(new ChannelTypeUID(BINDING_ID, "color-temperature-abs"), "Color Temperature", "Number")
             .withDescription("Controls the color temperature of the light in Kelvin").withCategory("ColorLight")
-            .withStateDescriptionFragment(StateDescriptionFragmentBuilder.create().withMinimum(new BigDecimal(1000))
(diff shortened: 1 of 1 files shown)


J-N-K:
I disagree. The absolute color temperature is much more intuitive than the relative one and this should be the default.

kaikreuzer:
The percent one is there by default, any ct thing supports it and it has always been there.

J-N-K:
Deconz does not support it. 

And as I said elsewhere: a Dimmer is wrong as can be for the colortemperature. ON/OFF don‘t make sense. Even if different lights may show smaller differences in absolute color temperature, 0-100% will in most cases not match. Color temperature in K is what is written on bulbs if you buy them, there is no percent value anywhere. 

„It has always been like that“ doesn‘t count as a valid argument in other cases, so what makes that a valid argument here?

kaikreuzer:
We had that discussion, I don't want to repeat it. Any binding is expected to have the percent channel for color temperature, since this is always possible. If this is not the case for deconz, I consider it a bug there.

J-N-K:
I consider it a bug in core.

cweitkamp:
What about renaming the channel type? We can have a "Color Temperature" beside "Color Temperature (Kelvin)" and "Farbtemperatur" / "Farbtemperatur (Kelvin)" respectively.

kaikreuzer:
It is still the same property that is controlled and thus duplicated.

cweitkamp (approved):
The 'advanced' flag does neither solves nor should end our discussion about color temperature. OHC now provides a solution for both ways. If bindings will implement them, a user can decide on its own which is the best way for him/her to control lights.

J-N-K:
Looks like it's a one-man-show. Then be it.

bodiroga:
The work that @J-N-K has done for the proyect can not end this way, this is really sad 😢

PR #2188, opened 2021-02-09
Add profile to generate semantics classes with groovy-maven-plugin

wborn:
This way you can generate the classes with Maven when adding `-DgenerateTagClasses` to your build command.
I.e. you no longer need to have Groovy locally installed for this.
Also properly sorts the generated imports so they match the Spotless configuration.


kaikreuzer (approved):
Very convenient, thanks!

PR #2190, opened 2021-02-10
[scheduler] Mitigation fix for earlier triggers #1976

### LINKED ISSUE #1976: OH3.RC2 - All my daily rules (cron trigger) run twice
See https://community.openhab.org/t/openhab-3-0-release-candidates-discussion/110200/184?u=moody_blue

Hilbrand:
Some users reported the scheduler to be run to earlier. It actual is known the Java scheduler can drift.
To handle this problem 2 changes have been made:
1. If the scheduler triggers > 2 sec before expected end time it will reschedule.
2. The scheduler keeps track of the timestamp the job should run. For recurring schedulers, like cron, when the next job is run it offsets of the calculated time, and not the actual time. This guaranties that the next scheduled time is actual the next scheduled time and not the same end time in case the scheduler would trigger to early.

Closes #1976


kaikreuzer (approved):
Many thanks @Hilbrand!

PR #2194, opened 2021-02-11
[metering] tag rule metric with human readable rule name in addition …

pravussum:
…to the rule id.

Signed-off-by: Robert Bach <openhab@mortalsilence.net>


cweitkamp (approved):
LGTM. Thanks.

Issue #2145, opened 2021-01-22
io transport mqtt sets wrong QoS levels to hiveMQ


stedon81:
Package: org.openhab.core.io.transport.mqtt.internal.client

`Mqtt3AsyncClientWrapper.publish()` and `Mqtt3AsyncClientWrapper.subscribe` use a function called `getMqttQosFromInt` to map the `int qos` to the enum value required by hiveMQ.
This function maps the integer values wrongly:
```
    protected MqttQos getMqttQosFromInt(int qos) {
        switch (qos) {
            case 0:
                return MqttQos.AT_LEAST_ONCE;
            case 1:
                return MqttQos.AT_MOST_ONCE;
            case 2:
                return MqttQos.EXACTLY_ONCE;
            default:
                throw new IllegalArgumentException("QoS needs to be 0, 1 or 2.");
        }
    }
```

According to https://www.hivemq.com/blog/mqtt-essentials-part-6-mqtt-quality-of-service-levels/, the QoS levels are:
* At most once (0)
* At least once (1)
* Exactly once (2)

Hence, QoS levels 0 and 1 must be exchanged in that function.

stedon81:
I prvided pull request #2208 to fix this issue

kaikreuzer:
Fixed by https://github.com/openhab/openhab-core/pull/2208 - thanks!

PR #2208, opened 2021-02-18
Fix MQTT QoS levels setting

stedon81:
Solves issue #2145 "io transport mqtt sets wrong QoS levels to hiveMQ" 
https://github.com/openhab/openhab-core/issues/2145

Fix is simply exchanging the enum values returned for the case 0 and case 1 branches, as they have been mixed up as compared to the mqtt specification:

> According to https://www.hivemq.com/blog/mqtt-essentials-part-6-mqtt-quality-of-service-levels/, the QoS levels are:
> 
> - At most once (0)
> - At least once (1)
> - Exactly once (2)


stedon81:
Will implicitly also fix issue in mqtt.homie binding:
https://github.com/openhab/openhab-addons/issues/9919

wborn:
I think the changes are good! Can you also sign off your commit or provide the `Signed-off-by:` line in a comment so we can add it? See [Sign your work](https://www.openhab.org/docs/developer/contributing.html#sign-your-work).

kaikreuzer:
I think we can treat that as a small patch exception and skip the DCO.

kaikreuzer (approved):

Issue #1983, opened 2020-12-22
Script files are loaded before scripting engine is available


cweitkamp:
I am using Jython Scripting and installed the bundle for it. During OH3 restarts I am seeing the following info in my logs. One message for each physical file. Approximately 5 min later all the files are reloaded and the rules can be used without an issue. This looks like we can implement a small improvement here.

```
2020-12-22 10:29:21.331 [INFO ] [ab.core.service.AbstractWatchService] - ScriptEngine for py not available
2020-12-22 10:29:21.338 [INFO ] [ab.core.service.AbstractWatchService] - ScriptEngine for py not available
2020-12-22 10:29:21.352 [INFO ] [ab.core.service.AbstractWatchService] - ScriptEngine for py not available
2020-12-22 10:29:21.358 [INFO ] [ab.core.service.AbstractWatchService] - ScriptEngine for py not available
2020-12-22 10:29:21.363 [INFO ] [ab.core.service.AbstractWatchService] - ScriptEngine for py not available
2020-12-22 10:29:21.369 [INFO ] [ab.core.service.AbstractWatchService] - ScriptEngine for py not available
2020-12-22 10:29:21.373 [INFO ] [ab.core.service.AbstractWatchService] - ScriptEngine for py not available
2020-12-22 10:29:21.380 [INFO ] [ab.core.service.AbstractWatchService] - ScriptEngine for py not available
2020-12-22 10:29:21.385 [INFO ] [ab.core.service.AbstractWatchService] - ScriptEngine for py not available
2020-12-22 10:29:21.391 [INFO ] [ab.core.service.AbstractWatchService] - ScriptEngine for py not available
2020-12-22 10:29:21.398 [INFO ] [ab.core.service.AbstractWatchService] - ScriptEngine for py not available
2020-12-22 10:29:21.401 [INFO ] [ab.core.service.AbstractWatchService] - ScriptEngine for py not available
2020-12-22 10:29:21.405 [INFO ] [ab.core.service.AbstractWatchService] - ScriptEngine for py not available
2020-12-22 10:29:21.409 [INFO ] [ab.core.service.AbstractWatchService] - ScriptEngine for py not available
2020-12-22 10:29:21.413 [INFO ] [ab.core.service.AbstractWatchService] - ScriptEngine for py not available
2020-12-22 10:29:21.417 [INFO ] [ab.core.service.AbstractWatchService] - ScriptEngine for py not available
[...]

PR #2222, opened 2021-03-03
[automation] Added automation script StartLevels

### LINKED ISSUE #2002: Script files not executed in order at startup
In OH2.5 script files were executed in order (alphanumeric based on script name) at startup. In OH3 there appears to be no specific order in their initial execution.

Looking at the code, I believe that the problem is in https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/loader/ScriptFileWatcher.java

In 2.5, the watcher starts early, before script engines are available, and with a specific "earliestStart" time, and enqueues all the script files. When the engine becomes available, the files are dequeued, sorted (in `checkFiles`), then passed on for execution. Note that this is a pretty opaque mechanism to sort the files - the method doesn't mention it, and it only happens dues to getting enqueued before the script engine is available/earliestStart is reached (if the script engine is available, they get submitted immediately in the order they are read/detected). It looks like in OH3 this behaviour was missed when adding the `ReadyState` stuff, because the first thing called is `importFiles` rather than `checkFiles`, which, if script engines are available, will _not_ submit the scripts in alphanumeric order.

jpg0:
Allows scripts to specify per-script start levels (which differ from the start level for the `ScriptFileWatcher` itself). Also extracted various functionality from ScriptFileWatcher into distinct components & added unit tests to cover all script loading functionality.

This fixes #2002 and possibly #1983 too. The vast majority of the new LOC are unit tests.

Signed-off-by: jpg0 <jpg@trillica.com>

### DIFF
--- a/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/loader/ScriptFileReference.java
+++ b/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/loader/ScriptFileReference.java
@@ -1,0 +1,120 @@
+/**
+ * Copyright (c) 2010-2021 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.automation.module.script.rulesupport.internal.loader;
+
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Optional;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.jetbrains.annotations.NotNull;
+import org.openhab.core.service.StartLevelService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Script File wrapper offering various methods to inspect the script
+ *
+ * @author Jonathan Gilbert - initial contribution
+ */
+@NonNullByDefault
+public class ScriptFileReference implements Comparable<ScriptFileReference> {
+
+    private static final Set<String> EXCLUDED_FILE_EXTENSIONS = new HashSet<>(
+            Arrays.asList("txt", "old", "example", "backup", "md", "swp", "tmp", "bak"));
+
+    private static final Pattern[] startLevelPatterns = new Pattern[] { Pattern.compile(".*/sl(\\d{2})/[^/]+"), // script
+                                                                                                                // in
+                                                                                                                // immediate
+                                                                                                                // slXX
+                                                                                                                // directory
+            Pattern.compile(".*/[^/]+\\.sl(\\d{2})\\.[^/.]+") // script named <name>.slXX.<ext>
+    };
+
+    private static final Logger logger = LoggerFactory.getLogger(ScriptFileReference.class);
+
+    private final URL scriptFileURL;
+
+    public ScriptFileReference(URL scriptFileURL) {
+        this.scriptFileURL = scriptFileURL;
+    }
+
+    public URL getScriptFileURL() {
+        return scriptFileURL;
+    }
+
+    public int getStartLevel() {
+        for (Pattern p : startLevelPatterns) {
+            Matcher m = p.matcher(scriptFileURL.getPath());
+            if (m.find() && m.groupCount() > 0) {
+                try {
+                    return Integer.parseInt(m.group(1));
+                } catch (NumberFormatException nfe) {
+                    logger.warn("Extracted start level {} from {}, but it's not an integer. Ignoring.", m.group(1),
+                            scriptFileURL.getPath());
+                }
+            }
+        }
+
+        return StartLevelService.STARTLEVEL_RULEENGINE;
+    }
+
+    public Optional<String> getScriptType() {
+        String fileName = scriptFileURL.getPath();
+        int index = fileName.lastIndexOf(".");
+        if (index == -1) {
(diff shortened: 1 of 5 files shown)


boc-tothefuture:
> Allows scripts to specify per-script start levels.

Can you provide an example of how the script accomplishes this?

jpg0:
I provided two mechanisms - either in the name of the script or the folder that contains it. So to run at start level 60:

`.../script.sl60.py` or
`.../sl60/script.js`

Happy to add documentation changes to the PR too, but I'm not sure where it goes.

jpg0:
Found the docs at https://github.com/openhab/openhab-docs/blob/main/configuration/jsr223.md

Happy to create a PR for this too, just want to ensure that the maintainers are happy with this change before I do so.

(I also see that the current docs describe the loading order, which is no longer correct as of OH3.x)

bundles/org.openhab.core/src/main/java/org/openhab/core/service/StartLevelService.java
```diff
@@ -249,10 +245,35 @@ private void setStartLevel(int level) {
...
+    public void registerTracker(StartLevelTracker tracker) {
+        slTrackers.add(tracker);
+        tracker.onStartLevelChanged(openHABStartLevel);
+    }
+
+    public void unregisterTracker(StartLevelTracker tracker) {
+        slTrackers.remove(tracker);
+    }
+
+    public interface StartLevelTracker {
```
kaikreuzer, line 276: I deliberately didn't introduce a dedicated tracker here as it somewhat duplicates the already existing `ReadyTracker` - this service can be used to track start level changes, see [this example](https://github.com/openhab/openhab-core/blob/127724c0e31ad8abf0855d87fbf96204267a39be/bundles/org.openhab.core.model.rule/src/org/openhab/core/model/rule/jvmmodel/RulesRefresher.java#L130-L131).

Would you be ok to refactor your PR accordingly?

kaikreuzer (changes requested):

bundles/org.openhab.core/src/main/java/org/openhab/core/service/StartLevelService.java
jpg0: @kaikreuzer sure! I've just pushed an update which switches to that patttern. Hope that is better.

(BTW I originally added the dedicated tracker to the StartLevelService because I didn't want to replicate the functionality that it provided - e.g. storing all the ready markers and determining the singular current start level. I've just used a much simpler implementation however.)

kaikreuzer (approved):
Thanks @jpg0, lgtm!

And yes, it would be awesome if you could create a PR for the docs repo to cover this new feature there - thanks!

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/loader/ScriptFileReference.java
```diff
@@ -0,0 +1,120 @@
...
+import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Optional;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.jetbrains.annotations.NotNull;
```
wborn, line 27: We don't use jetbrains null annotations. 

See https://www.openhab.org/docs/developer/guidelines.html#null-annotations

Can you do a follow up PR with adding proper null annotations?

bundles/org.openhab.core.automation.module.script.rulesupport/src/test/java/org/openhab/core/automation/module/script/rulesupport/internal/loader/DelegatingScheduledExecutorService.java
```diff
@@ -0,0 +1,133 @@
...
+
+    @NotNull
+    @Override
+    public ScheduledFuture<?> schedule(@NotNull Runnable command, long delay, @NotNull TimeUnit unit) {
+        return delegate.schedule(command, delay, unit);
+    }
+
+    @NotNull
+    @Override
+    public <V> ScheduledFuture<V> schedule(@NotNull Callable<V> callable, long delay, @NotNull TimeUnit unit) {
```
wborn, line 41: All the jetbrains `@NotNull` annotations in this file should be replaced with Eclipse JDT annotaitons.

kaikreuzer:
Good catches, @wborn, sorry for having merged without noticing. 😲

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/loader/ScriptFileReference.java
jpg0: Sorry about that! IDEA can be very keen on adding these sometimes; I hadn't noticed. I'll create a follow up PR to remove them.

wborn: Thanks!

wborn:
Those jetbrains annotations seem to be a dependency of hivemq-mqtt-client. Maybe we can exclude it to prevent it from happening. They've ended up in some add-ons too (https://github.com/openhab/openhab-addons/pull/10334).

jpg0:
PR to remove the introduced annotations: #2241

jpg0:
@wborn Can we add a build step to reject imports from the `org.jetbrains.annotations` package? I just had a look to see if it would be easy in spotless (possibly, not sure as the annotations don't always have equivalents), but it could be done in the custom sat plugin stuff to just break the build.

wborn:
SAT already scans for forbidden packages. But it will only result in warnings and not errors, so we could add the package to [ruleset.properties](https://github.com/openhab/openhab-addons/blob/ee59803df88b009c607b10d691f5c2804d1dbb28/tools/static-code-analysis/checkstyle/ruleset.properties#L3)

Excluding the dependency from hivemq-mqtt-client would also introduce another hurdle for using them. :wink:

jpg0:
PR for doc changes: https://github.com/openhab/openhab-docs/pull/1511

PR #2241, opened 2021-03-14
Remove Jetbrains @NonNull annotations

jpg0:
Removed Jetbrains annotations which were accidentally introduced in #2222. 


wborn (approved):
Thanks for the quick fix! :-)

PR #2244, opened 2021-03-17
Upgrade Gson to 2.8.6

wborn:
Upgrades Gson from 2.8.2 to 2.8.6.

See change log: https://github.com/google/gson/blob/master/CHANGELOG.md

### DIFF
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -91,9 +91,9 @@
 
     <!-- Gson -->
     <dependency>
-      <groupId>org.eclipse.orbit.bundles</groupId>
-      <artifactId>com.google.gson</artifactId>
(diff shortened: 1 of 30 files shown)


marcelrv:
Will this:
_Added static methods JsonParser.parseString and JsonParser.parseReader and deprecated instance method JsonParser.parse_

Make all cases where JsonParser.parse is used unusable? or only give a warning?

wborn:
Yes those methods still exist but are flagged as deprecated since 2019 :wink: , see https://github.com/google/gson/commit/c5a3f21fbad3828deba1deb88168e1f76829f807

kaikreuzer (approved):
Cool, no need for the Orbit bundling anymore. 🎉

Issue #2134, opened 2021-01-18
Integration tests often fail with "Stream closed" or "No input provided for project"


wborn:
Often our integration tests executed by the bnd-testing-maven-plugin:testing goal fail with one of the following reasons:

* `java.io.IOException: Stream closed`
* `java.io.EOFException: No input provided for project`

I spent some time to debug this recently, see:

* https://github.com/bndtools/bnd/issues/4253#issuecomment-761113709
* https://github.com/bndtools/bnd/issues/4509

It seems to be an issue in bnd where it may sometimes close `System.in` when other builder threads change `System.in` in multi-threaded Maven builds.

As a workaround I've reconfigured the core/add-ons jobs so it does a full multi threaded build without integration tests, followed by just a build that runs the integration tests and nothing else. Let's hope it reduces running into this bug until there is a proper fix.

wborn:
This bnd issue is still occuring sometimes.

wborn:
Hopefully this will be fixed with https://github.com/bndtools/bnd/pull/5177.

PR #2245, opened 2021-03-19
Removed outdated basic auth classes

kaikreuzer:
Afaics, these classes are not used and I doubt that we'll need them anytime soon, since we have now authentication in place by different means.


wborn (approved):
Thanks for the cleanup! The test failures are unrelated. Seems like Bnd 5.3.0 didn't fully fix https://github.com/openhab/openhab-core/issues/2134. :disappointed:

PR #2255, opened 2021-03-25
Add null annotations to org.openhab.core.voice

wborn:
Allows for using null annotations with the voice TTS add-ons.


kaikreuzer (approved):
Thanks!

PR #2256, opened 2021-03-25
Fix SAT findings

wborn:
Fixes 65 SAT findings.


kaikreuzer (approved):
Thanks for the massive cleanup!

PR #2270, opened 2021-04-04
[skeleton] Use real world examples; fix imports

fwolter:
Added some examples as many new bindings don't make use of context tags or supported bridges.

Also add a refresh parameter to show the `unit` and `min` attributes.

### DIFF
--- a/tools/archetype/binding/src/main/resources/archetype-resources/src/main/resources/OH-INF/thing/thing-types.xml
+++ b/tools/archetype/binding/src/main/resources/archetype-resources/src/main/resources/OH-INF/thing/thing-types.xml
@@ -6,6 +6,12 @@
 
 	<!-- Sample Thing Type -->
 	<thing-type id="sample">
+		<!-- Comment this in for Things which need a Bridge to function.
+		<supported-bridge-type-refs>
+			<bridge-type-ref id="${bindingIdCamelCase}Bridge" />
+		</supported-bridge-type-refs>
+		-->
+
 		<label>${bindingIdCamelCase} Binding Thing</label>
 		<description>Sample thing for ${bindingIdCamelCase} Binding</description>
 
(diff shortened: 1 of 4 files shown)


tools/archetype/binding/src/main/resources/archetype-resources/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -14,19 +20,23 @@
 		</channels>
 
 		<config-description>
-			<parameter name="config1" type="text" required="true">
-				<label>Sample Parameter</label>
-				<description>This is a sample text configuration parameter</description>
+			<parameter name="hostname" type="text" required="true">
```
cweitkamp, line 23: This is a nice improvement. We should change the example code accordingly. e.g.

https://github.com/openhab/openhab-core/blob/949313744d24b888a90a4e904fca67ae32978b94/tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/internal/__bindingIdCamelCase__Configuration.java#L28

tools/archetype/binding/src/main/resources/archetype-resources/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -6,6 +6,12 @@
 
 	<!-- Sample Thing Type -->
 	<thing-type id="sample">
+		<!-- Comment this in for Things which need a Bridge to function.
+		<supported-bridge-type-refs>
+			<bridge-type-ref id="${bindingIdCamelCase}Bridge" />
```
cweitkamp, line 11: I think it makes sense to add it but only if we provide the source for the resulting bridge too. Including complete handling in `ThingHandlerFactory`and so on. Maybe add a hint in the JavaDoc to delete the class if not needed.

Once we got this into out skeleton we can add more example code from the docs like discovery.

cweitkamp (commented):
Thanks for this draft. Imo it makes sense to change the skeleton.

Unfortunately the build failed because of spotless errors.

tools/archetype/binding/src/main/resources/archetype-resources/src/main/resources/OH-INF/thing/thing-types.xml
fwolter: We could copy the Thing handler class and the XML structure, but those would be nearly 1:1 copies. I'm unsure if that adds any benefit or just puzzles the user...

tools/archetype/binding/src/main/resources/archetype-resources/src/main/resources/OH-INF/thing/thing-types.xml
fwolter: Good point. I also adapted the i18n.

tools/archetype/binding/src/main/resources/archetype-resources/src/main/resources/OH-INF/thing/thing-types.xml
cweitkamp: No, there will be not much benefit when copying it. As I mentioned above an example discovery services as `ThingHandlerService` implementation provided by the bridge handler would be nice. But we can add it later.

cweitkamp (approved):

Issue #1842, opened 2020-11-22
Serial ports getting blocked after some re-connecting


fwolter:
When using the Modbus binding with OH3, the serial port with the Modbus RS485 transciever cannot be opened anymore after some time.

When the error appears, the following is printed on stdout: `initialise_event_info_struct: initialise failed!`

This is a known bug NeuronRobotics/nrjavaserial#111.

After compiling the native libs with `DISABLE_LOCKFILES` as suggested in above issue, the error is gone.

This has been discussed in https://community.openhab.org/t/modbus-binding-loses-its-serial-connection-every-few-days-without-a-reason-after-updating-to-2-5-0-release-build/89550/14

Maybe locking should be disabled in OH nrjavaserial again?

wborn:
For me this issue is not that critical wanting to start forking nrjavaserial again. :wink: Forking nrjavaserial implies all other native libraries need to be recompiled as well (linux/armhf/arm64) and you'd have to merge any upstream changes to the fork.

fwolter:
I understood you == me :) I'd do that if the pain gets too big. Until now there is only one user beside me facing the problem. Maybe @MrDOS manages to re-work the locking mechanism in the next time.

jumblejj:
Hi, Also seeing this issues with OH3, see [https://community.openhab.org/t/modbus-binding-not-working-on-oh3/111619](https://community.openhab.org/t/modbus-binding-not-working-on-oh3/111619)

ssalonen:
Is there workaround we could implement in the binding side?

ssalonen:
@fwolter could you enlighten whether the following workaround would help in the binding?

Current behaviour
- when I/O error is encountered, serial port is closed and re-opened before next retry

New behaviour
- when I/O error is encountered, serial port is kept open

ssalonen:
Comment from user in community https://community.openhab.org/t/modbus-binding-problem-with-serial-connection/117942/8?u=ssalonen shows how frustrating the nrjavaserial issue is in practice:

> I read all those threads.
> I gave up with this binding. It’s not usable.
>
> I have written my own rules to query the modbus unit using the modpoll command line utility.
> Works perfectly and absolutely reliable 100%.

fwolter:
> @fwolter could you enlighten whether the following workaround would help in the binding?
> 
> Current behaviour
> 
>     * when I/O error is encountered, serial port is closed and re-opened before next retry
> 
> 
> New behaviour
> 
[...]

PR #2272, opened 2021-04-05
[modbus] Workaround for nrjavaserial issues: do not disconnect serial

ssalonen:
Workaround for https://github.com/openhab/openhab-core/issues/1842.
Typically serial connections are disconnected and reconnected on IO
errors. This has turned out to be problematic with the recent
nrjavaserial updates brought by OH3 (locking behaviour of nrjavaserial
seems to be prone to deadlocks).

We workaround the issues by trying to keep the serial connection open
as long as possible.

This might not a foolproof fix as commented by @fwolter https://github.com/openhab/openhab-core/issues/1842#issuecomment-792978074 :
> [...] I often observed the error right after the start of the JVM.
> I agree that this problem is very annoying. A "quick" solution would be to disable the locks in general.

But on the other hand the community reports seem to be very often related to "I/O" errors and the disconnect/re-connect logic in modbus transport.

Pre-compiled jar in zip: 
[org.openhab.core.io.transport.modbus-3.1.0-SNAPSHOT.zip](https://github.com/openhab/openhab-core/files/6259049/org.openhab.core.io.transport.modbus-3.1.0-SNAPSHOT.zip)
(corresponding to commit https://github.com/openhab/openhab-core/pull/2272/commits/1dc4094b770242c626d784c3482552cad729b592) 

### DIFF
--- a/bundles/org.openhab.core.io.transport.modbus/src/main/java/org/openhab/core/io/transport/modbus/internal/ModbusManagerImpl.java
+++ b/bundles/org.openhab.core.io.transport.modbus/src/main/java/org/openhab/core/io/transport/modbus/internal/ModbusManagerImpl.java
@@ -627,9 +627,15 @@ public class ModbusManagerImpl implements ModbusManager {
                                 "Last try {} failed when executing request ({}). Aborting. Error was I/O error, so reseting the connection. Error details: {} {} [operation ID {}]",
                                 tryIndex, request, e.getClass().getName(), e.getMessage(), operationId);
                     }
-                    // Invalidate connection, and empty (so that new connection is acquired before new retry)
-                    timer.connection.timeConsumer(c -> invalidate(endpoint, c), connection);
-                    connection = Optional.empty();
+                    if (endpoint instanceof ModbusSerialSlaveEndpoint) {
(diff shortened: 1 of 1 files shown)


cweitkamp (approved):
LGTM. Thanks.

ppieczul:
Looks like I need a 3.1 version to use this transport jar. How can I switch to 3.1 on openhabian? Somehow I fail to find a good way, openhabian-config allows for selecting only 3.0

ssalonen:
@ppieczul unfortunately I am not familiar with openhabian... Perhaps best to ask in community forum.

ppieczul:
I got 3.1.M3 installed finally. How do I make the Modbus binding use the new transport jar? I stopped the default transport bundle, installed the jar and the Modbus bundle remains in Waiting state.
Screenshot 2021-04-15 at 21 54 40

ssalonen:
@ppieczul unfortunately I must say that M3 is too old (around 2 weeks) and does not include this fix (1 week ago merged).  Instead of utilizing milestone release, you need to refer to snapshot versions.

You can find these via openhab.org website but not sure what is the correct installation method with openhabian

PR #2278, opened 2021-04-06
Upgrade Xtext/Xtend to 2.25.0, LSP4J to 0.10.0

wborn:
For release notes, see:

https://www.eclipse.org/Xtext/releasenotes.html#/releasenotes/2021/03/02/version-2-25-0
https://github.com/eclipse/lsp4j/blob/master/CHANGELOG.md#v0100-nov-2020

Also aligns the spifly and xbean runtime dependencies with Karaf 4.3.1 which was not possible because the older Xtext required ASM 8.0.1.


features/karaf/openhab-tp/src/main/feature/feature.xml
```diff
@@ -175,9 +175,10 @@
 	</feature>
 
 	<feature name="openhab.tp-lsp4j" description="Eclipse LSP4J" version="${project.version}">
-		<capability>openhab.tp;feature=lsp4j;version=0.9.0</capability>
-		<bundle>mvn:org.eclipse.lsp4j/org.eclipse.lsp4j/0.9.0</bundle>
-		<bundle>mvn:org.eclipse.lsp4j/org.eclipse.lsp4j.jsonrpc/0.9.0</bundle>
+		<capability>openhab.tp;feature=lsp4j;version=0.10.0</capability>
+		<bundle>mvn:org.eclipse.lsp4j/org.eclipse.lsp4j/0.10.0</bundle>
+		<bundle>mvn:org.eclipse.lsp4j/org.eclipse.lsp4j.jsonrpc/0.10.0</bundle>
+		<bundle dependency="true">mvn:com.google.guava/guava/27.1-jre</bundle>
```
cweitkamp, line 181: Why Guava 27.1 here? We want rise 30.1. Don't we?

wborn: That would be nice but it is not possible because the version ranges aren't aligned and Guava doesn't use semantic versioning.

wborn: I see Guava does use semver, see [releases policy](https://github.com/google/guava/wiki/ReleasePolicy#versioning). So there are often breaking changes... so if you have many dependencies you will quickly end up with multiple versions. :-( LSP4J 0.11.0 does have a version range that supports Guava 30.1.. but Xtext first needs to be updated for that. So hopefully with Xtext 2.26.0 we can remove it again.

cweitkamp: Okay. Then we should merge this now.

cweitkamp (approved):

PR #2284, opened 2021-04-11
Upgrade jamod (to get rid of case with fragmented packets)

denis-ftc-denisov:
This PR bumps version of jamod modbus library to include fix for case with fragmented modbus packets received via TCP connection (discussed here https://github.com/openhab/jamod/pull/11)

### DIFF
--- a/bundles/org.openhab.core.io.transport.modbus/pom.xml
+++ b/bundles/org.openhab.core.io.transport.modbus/pom.xml
@@ -35,7 +35,7 @@
     <dependency>
       <groupId>net.wimpi</groupId>
       <artifactId>jamod</artifactId>
-      <version>[1.3.2.OH]</version>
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.io.transport.modbus/pom.xml
```diff
@@ -35,7 +35,7 @@
     <dependency>
       <groupId>net.wimpi</groupId>
       <artifactId>jamod</artifactId>
-      <version>[1.3.0.OH,1.3.1.OH]</version>
+      <version>[1.3.2.OH]</version>
```
wborn, line 38: It seems unnecessary to still have brackets for a single version.

```suggestion
      <version>1.3.2.OH</version>
```

wborn (approved):
Thanks!

PR #2289, opened 2021-04-13
[metering] Fix dependency issue in io.monitor bundle (#2288)

### LINKED ISSUE #2288: [metering] NoClassDefFoundError - mandatory transitive Micrometer dependencies missing
This has originally been reported in openhab-addons:
https://github.com/openhab/openhab-addons/issues/10515

The actual issue is located within the core, since the Metrics IO addon makes use of the MeterRegistryProvider provided by the core (org.openhab.core.io.monitor bundle). The Micrometer classes depend on a transitive dependency, which is only present in the Metrics IO addon, but not in the core bundle. 

The solution is to include mandatory transitive dependencies of Micrometer in the core bunde.

pravussum:
Fixes #2288. 

Signed-off-by: Robert Bach <openhab@mortalsilence.net>

### DIFF
--- a/bundles/org.openhab.core.io.monitor/bnd.bnd
+++ b/bundles/org.openhab.core.io.monitor/bnd.bnd
@@ -6,4 +6,6 @@ Import-Package: \
  org.osgi.service.*,\
  org.slf4j.*,\
  com.google.gson.*;version="[2.8,3)"
-Export-Package: io.micrometer.core.*
+Export-Package: io.micrometer.core.*;-split-package:=merge-first,\
+ org.HdrHistogram.*;-split-package:=merge-first,\
+ org.LatencyUtils.*;-split-package:=merge-first
--- a/bundles/org.openhab.core.io.monitor/pom.xml
+++ b/bundles/org.openhab.core.io.monitor/pom.xml
@@ -41,7 +41,7 @@
       <groupId>org.latencyutils</groupId>
       <artifactId>LatencyUtils</artifactId>
       <version>2.0.3</version>
-      <scope>runtime</scope>
+      <scope>compile</scope>
       <exclusions>
         <exclusion>
           <groupId>org.hdrhistogram</groupId>
--- a/bundles/org.openhab.core.io.monitor/pom.xml
+++ b/bundles/org.openhab.core.io.monitor/pom.xml
@@ -31,6 +31,24 @@
       <version>1.6.3</version>
       <scope>compile</scope>
     </dependency>
+    <dependency>
+      <groupId>org.hdrhistogram</groupId>
+      <artifactId>HdrHistogram</artifactId>
+      <version>2.1.12</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.latencyutils</groupId>
+      <artifactId>LatencyUtils</artifactId>
+      <version>2.0.3</version>
+      <scope>runtime</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>org.hdrhistogram</groupId>
+          <artifactId>HdrHistogram</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
   </dependencies>
 
   <build>
--- a/bundles/org.openhab.core.io.monitor/src/main/java/org/openhab/core/io/monitor/internal/metrics/RuleMetric.java
+++ b/bundles/org.openhab.core.io.monitor/src/main/java/org/openhab/core/io/monitor/internal/metrics/RuleMetric.java
@@ -16,6 +16,7 @@ import java.util.Collection;
 import java.util.Dictionary;
 import java.util.HashSet;
 import java.util.Hashtable;
+import java.util.Objects;
 import java.util.Set;
 
 import org.eclipse.jdt.annotation.Nullable;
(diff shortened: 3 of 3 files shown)


bundles/org.openhab.core.io.monitor/pom.xml
```diff
@@ -31,6 +31,24 @@
...
+      <groupId>org.hdrhistogram</groupId>
+      <artifactId>HdrHistogram</artifactId>
+      <version>2.1.12</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.latencyutils</groupId>
+      <artifactId>LatencyUtils</artifactId>
+      <version>2.0.3</version>
+      <scope>runtime</scope>
```
kaikreuzer, line 44: Don't you need to add this to the Karaf feature to make the bundle is installed at runtime?

bundles/org.openhab.core.io.monitor/bnd.bnd
```diff
@@ -6,4 +6,6 @@ Import-Package: \
  org.osgi.service.*,\
  org.slf4j.*,\
  com.google.gson.*;version="[2.8,3)"
-Export-Package: io.micrometer.core.*
+Export-Package: io.micrometer.core.*;-split-package:=merge-first,\
```
kaikreuzer, line 9: What is the split-package good for?

bundles/org.openhab.core.io.monitor/bnd.bnd
```diff
@@ -6,4 +6,6 @@ Import-Package: \
...
+ org.HdrHistogram.*;-split-package:=merge-first,\
```
kaikreuzer, line 10: Are those libraries included in the bundle jar?

bundles/org.openhab.core.io.monitor/src/main/java/org/openhab/core/io/monitor/internal/metrics/RuleMetric.java
```diff
@@ -80,9 +81,9 @@ public void unbind() {
         if (meterRegistry == null) {
             return;
         }
-        for (Meter meter : meterRegistry.getMeters()) {
+        for (Meter meter : Objects.requireNonNull(meterRegistry).getMeters()) {
```
kaikreuzer, line 84: That's rather ugly. Can't you simply do a proper `else` case for the if-statement above and assign `meterRegistry` to a local variable to make the null checks happy?

kaikreuzer (changes requested):

bundles/org.openhab.core.io.monitor/bnd.bnd
pravussum: I added them because there was a warning in the bnd-maven plugin:
`Split package, multiple jars provide the same package` - thb I don't know where these duplicates originate from, I'm quite inexperienced with bnd/OSGI. 
The split-package directive basically just sets the default strategy explicitly and thus removes the warning, not really the root cause. Any advice appreciated.

bundles/org.openhab.core.io.monitor/bnd.bnd
pravussum: Yes.

bundles/org.openhab.core.io.monitor/src/main/java/org/openhab/core/io/monitor/internal/metrics/RuleMetric.java
pravussum: Unfortunately there is a clash with the io.micrometer.core.instrument.@NonNull annotation, which is not applicable to local variables. And this the least verbose variant of the whole class I got to compile without warnings.

kaikreuzer: No, that's not the problem. What I meant is https://github.com/pravussum/openhab-core/pull/1.

kaikreuzer: No, this is not the problem. What I meant is https://github.com/pravussum/openhab-core/pull/1.

bundles/org.openhab.core.io.monitor/bnd.bnd
kaikreuzer: > I don't know where these duplicates originate from

My assumption is that this is because the original jar is on the build path as well as the extracted classes - so bnd thinks that two different bundles provide this package. The setting should thus be useless at runtime. It only suppresses the warning at build time. Not sure if we should keep it for this.

@wborn As our bnd expert: Do you have any opinion here?

bundles/org.openhab.core.io.monitor/src/main/java/org/openhab/core/io/monitor/internal/metrics/RuleMetric.java
pravussum: Hm, I think I tried exactly that and got some compiler error/warning that forced me to change the Nullable annotations (that's why I mentioned them), but I don't get it anymore. Never mind, probably an environment issue - your changes are merged, thanks for the taking the effort!

kaikreuzer (approved):
Thanks! (build issue is unrelated to that PR)

PR #2302, opened 2021-04-20
Fix Crowdin translation filename mismatch

wborn:
Looks like this is the root cause of the generated translation filenames not matching the original filename.


cweitkamp (approved):
Good catch. Thanks.

PR #2306, opened 2021-04-21
[i18n] Translation of configurable services

cweitkamp:
- Added property files for configurable services
- Renamed config files
- Removed default "required=false" flags

Not yet complete ...

Labels / titles of system services are set via OSGi component `service.config.label`. Could not figure out if they are i18n too.

grafik

https://github.com/openhab/openhab-core/blob/62c85a3214ac74e5c9ac4266cd6b9c805ebf329b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/i18n/I18nProviderImpl.java#L77-L81


wborn:
> Labels / titles of system services are set via OSGi component service.config.label. Could not figure out if they are i18n too.

It's just a property only used by openHAB. They are added to `ConfigurableServiceDTO`s by [ConfigurableServiceResource.getServicesByFilter(String)](https://github.com/openhab/openhab-core/blob/4a2954d4091359f24d39d13affb514d80896573f/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/service/ConfigurableServiceResource.java#L266-L301).

Note that there are also add-ons (and UIs) that have these configurable services. So we would also need a mechanism for looking up those translations.

cweitkamp:
@wborn Thanks for the hint. I will try to implement a solution for it.

PR is ready for review.

bundles/org.openhab.core.config.discovery/src/main/resources/OH-INF/i18n/inbox.properties
```diff
@@ -0,0 +1,7 @@
+config.system.inbox.autoIgnore.label = Auto Ignore
+config.system.inbox.autoIgnore.description = <![CDATA[
+                <p>If enabled, an inbox result matching an existing thing is automatically ignored.</p>
+                <p>If set to false, the inbox may contain results although identical things already exist.</p>
+            ]]>
```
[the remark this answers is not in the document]
wborn: It doesn't work for me. I tested with the following `inbox_nl.properties`:

```
system.config.inbox.autoIgnore.label = NL_Auto Ignore
system.config.inbox.autoIgnore.description = <![CDATA[
                <p>NL_If enabled, an inbox result matching an existing thing is automatically ignored.</p>
                <p>NL_If set to false, the inbox may contain results although identical things already exist.</p>
            ]]>
system.config.inbox.autoApprove.label = NL_Auto Approve
system.config.inbox.autoApprove.description = NL_If enabled, inbox results are automatically approved, unless they were marked as ignored.
````

This is what it looks like in the UI:

Screenshot from 2021-04-27 16-59-53

It ignores the content when it is not on the same line, see the REST API response JSON:

```json
{
   "uri":"system:inbox",
   "parameters":[
      {
         "defaultValue":"true",
         "description":"\u003c![CDATA[",
         "label":"NL_Auto Ignore",
         "name":"autoIgnore",
         "required":false,
         "type":"BOOLEAN",
         "readOnly":false,
         "multiple":false,
         "advanced":false,
         "verify":false,
         "limitToOptions":true,
         "options":[
            
         ],
         "filterCriteria":[
            
         ]
      },
      {
         "defaultValue":"false",
         "description":"NL_If enabled, inbox results are automatically approved, unless they were marked as ignored.",
         "label":"NL_Auto Approve",
         "name":"autoApprove",
         "required":false,
         "type":"BOOLEAN",
         "readOnly":false,
         "multiple":false,
         "advanced":false,
         "verify":false,
         "limitToOptions":true,
         "options":[
            
         ],
         "filterCriteria":[
            
         ]
      }
   ],
   "parameterGroups":[
      
   ]
}
```

So it will probably also not work with Crowdin. Besides that it is also not easy to translate strings that contain all this markup. Is the HTML still required for the UI @ghys?

wborn (changes requested):
Thanks for working on this. :+1: I gave it a test by adding some NL files, but it only worked after I updated the keynames.

ghys:
> So it will probably also not work with Crowdin. Besides that it is also not easy to translate strings that contain all this markup. Is the HTML still required for the UI @ghys?

@wborn For some reason I cannot reply to the review message above🤔 
If you want to do a simple linebreak I'd suggest to drop the multiline `<p></p>` and just use `<br>` i.e.
`If enabled...<br>If set to false...`
Crowdin recognizes those tags properly and invites contributors not to touch them.

bundles/org.openhab.core.io.rest.auth/src/main/resources/OH-INF/i18n/restauth.properties
```diff
@@ -0,0 +1,6 @@
+config.system.restauth.allowBasicAuth.label = Allow Basic Authentication
+config.system.restauth.allowBasicAuth.description = Allow the use of Basic authentication to access protected API resources, in addition to access tokens and API tokens.
+config.system.restauth.cacheExpiration.label = Cache Expiration Time
+config.system.restauth.cacheExpiration.description = When basic authentication is activated, credentials are put in a cache in order to speed up request authorization. The entries in the cache expire after a while in order to not keep credentials in memory indefinitely. This value defines the expiration time in hours. Set it to 0 for disabling the cache.
+config.system.restauth.allowBasicAuth.label = Implicit User Role
+config.system.restauth.allowBasicAuth.description = By default, operations requiring the "user" role are available when unauthenticated. Disabling this option will enforce authorization for these operations. Warning: This causes clients that do not support authentication to break.
```
wborn, line 6:
```suggestion
config.system.restauth.implicitUserRole.label = Implicit User Role
config.system.restauth.implicitUserRole.description = By default, operations requiring the "user" role are available when unauthenticated. Disabling this option will enforce authorization for these operations. Warning: This causes clients that do not support authentication to break.
```

bundles/org.openhab.core/src/main/resources/OH-INF/i18n/i18n.properties
```diff
@@ -0,0 +1,40 @@
+config.system.i18n.language.label = Language
+config.system.i18n.language.description = <![CDATA[
+                <p>The default language that should be used. If not specified, the system default locale is used.</p>
+                <p>The ISO 639 alpha-2 or alpha-3 language code (if there is no alpha-2 one).</p>
```
ghys, line 4: Just a thought - as the valid options are provided and the UI displays them, I wonder if the "ISO code" indication and the examples are necessary in these parameters' descriptions.
For those configuring them in the services.cfg file, these hints could be provided as comments.

bundles/org.openhab.core.config.discovery/src/main/resources/OH-INF/i18n/inbox.properties
```diff
@@ -0,0 +1,7 @@
...
+config.system.inbox.autoApprove.label = Auto Approve
+config.system.inbox.autoApprove.description = If enabled, inbox results are automatically approved, unless they were marked as ignored.
```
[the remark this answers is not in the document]
cweitkamp: I will change it accordingly. Thanks for the hint.

bundles/org.openhab.core.config.discovery/src/main/resources/OH-INF/i18n/inbox.properties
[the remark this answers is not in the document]
cweitkamp: Thank you both for testing and the proposals. I will remove the Character Data and replace HTML paragraph elements by line breaks.

bundles/org.openhab.core/src/main/resources/OH-INF/i18n/i18n.properties
cweitkamp: Yes, I agree. The examples are necessary for textual configuration but not in the UI. We can remove it here and the other configuration parameters for this service.

cweitkamp:
I incorporated everything. Thanks again for your useful and - as usual - detailed and valuable comments.

bundles/org.openhab.core.config.discovery/src/main/resources/OH-INF/config/inbox.xml
```diff
@@ -8,16 +8,12 @@
 	<config-description uri="system:inbox">
 		<parameter name="autoIgnore" type="boolean">
 			<label>Auto Ignore</label>
-			<description><![CDATA[
-				<p>If enabled, an inbox result matching an existing thing is automatically ignored.</p>
-				<p>If set to false, the inbox may contain results although identical things already exist.</p>
-			]]>
-			</description>
+			<description><![CDATA[If enabled, an Inbox result matching an existing thing is automatically ignored.<br/>If set to false, the Inbox may contain results although identical things already exist.]]></description>
```
wborn, line 11: Is the `<![CDATA[` still used here so you don't have to escape HTML tags?

I had a look at the webui translations and saw that it uses `<br>` instead of `<br/>`:

https://github.com/openhab/openhab-webui/blob/7c8e7875832344edf335f4095c3a2d6f847cf276/bundles/org.openhab.ui/web/src/assets/i18n/empty-states/en.json#L6

It would make it a bit more consistent if we would use the same tags everywhere. I would prefer `<br>` since it is shorter, WDYT @ghys? It seems that `<br />` (with space) should be used to be XHTML compliant.

wborn:
> I incorporated everything. 

Thanks, I generated some NL translation files and now saw that all texts were properly "translated". :+1:

bundles/org.openhab.core.config.discovery/src/main/resources/OH-INF/config/inbox.xml
ghys: Both are fine in HTML5, I tend to agree with https://stackoverflow.com/a/4288474 although depending on the context/surrounding code I sometimes use `<br />`. In parameter descriptions I think `<br>` is better since it prevents escaping issues.

cweitkamp: Yes, I left some Character Data stuff because of the escaping. The `<description>` tag is a simple type element.and does not allow other HTML/XML Tags inside.

Talking about labels and descriptions something else came to my mind. When using the i18n from properties files the values inside the XML are kind of useless. One day someone will.be confused by the double place to maintain the labels.or description. Because of that I thought it might be better to switch every config description to.use custom keys like `@text/<key>` which separates the definition from i18n logic and prevents a lot of work in the future.

I will rework this PR accordingly.

cweitkamp: I will not change it. Once we have a proper XSLT transformation the problem will be solved too.

I changed the `<be>` tags.

wborn (approved):
Thanks! Now we have some more translation work to do to get back too 100%. :wink:

PR #2307, opened 2021-04-21
[net] Filter duplicate broadcast addresses for service configuration if multiple network interfaces are available

cweitkamp:
- Filter duplicate broadcast addresses for service configuration if multiple network interfaces are available


wborn (approved):
Thank you!

PR #2313, opened 2021-04-23
Ignore illegal thing status updates from REMOVING

### LINKED ISSUE #2311: Ignore thing status updates after thing was placed in REMOVING
During development of the [Miele Cloud Binding](https://community.openhab.org/t/miele-cloud-binding/93266), we decided to ignore thing status updates after the thing was placed in the `REMOVING` status. We discussed this in the [PR](https://github.com/openhab/openhab-addons/pull/9146) with @kaikreuzer who requested us to integrate this change into openHAB core.

In summary:
`ThingHandler`s are only notified by `handleRemove` and do not recognize that their things are `REMOVING` until then. Thus, concurrent tasks that update the thing status might produce exceptions because they update the thing status to `ONLINE`, `OFFLINE`, etc. Handling this by ignoring these status updates seems reasonable to us.

I will open a PR for this once I'm done with the changes.

BjoernLange:
Ignore illegal thing status transitions from REMOVING instead of throwing IllegalArgumentException. Feedback is appreciated, especially concerning unexpected side-effects and the added integration test.

Closes #2311

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
@@ -120,6 +120,7 @@ import org.slf4j.LoggerFactory;
  * @author Henning Sudbrock - Consider thing type properties when migrating to new thing type
  * @author Christoph Weitkamp - Added preconfigured ChannelGroupBuilder
  * @author Yordan Zhelev - Added thing disabling mechanism
+ * @author Björn Lange - Ignore illegal thing status transitions instead of throwing IllegalArgumentException
  */
 @NonNullByDefault
 @Component(immediate = true, service = { ThingTypeMigrationService.class, ThingManager.class })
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
```diff
@@ -187,10 +188,11 @@ public void statusUpdated(Thing thing, ThingStatusInfo statusInfo) {
 
             if (ThingStatus.REMOVING.equals(oldStatusInfo.getStatus())
                     && !ThingStatus.REMOVED.equals(statusInfo.getStatus())) {
-                // only allow REMOVING -> REMOVED transition, all others are illegal
-                throw new IllegalArgumentException(MessageFormat.format(
-                        "Illegal status transition from REMOVING to {0}, only REMOVED would have been allowed.",
-                        statusInfo.getStatus()));
+                // only allow REMOVING -> REMOVED transition, all others are ignored because they are illegal
+                logger.warn(
```
kaikreuzer, line 192: If we say that bindings do not stand a chance to prevent thing status updates when REMOVING status is entered, we should probably not even log a warning as this would look like there's a bug in the binding. I'd therefore suggest to change it to debug.

BjoernLange:
I applied your suggestion, @kaikreuzer.

kaikreuzer (approved):
Thanks!

PR #2333, opened 2021-04-30
[i18n] Added i18n feature for labels of ConfigurableServices

cweitkamp:
- Added i18n feature for labels of ConfigurableServices

Notation is either ` service.<uri>.label = ...` (`:` has to be replaced by `.` in the `<uri>` part) or `@text/<key>`.


wborn (approved):
Thanks the code looks good and it works well after some testing. :+1:

PR #2334, opened 2021-05-01
[i18n] Added missing properties for HLI, LSP and MQTT

cweitkamp:
- Added missing properties for HLI, LSP and MQTT

Minor dependency on #2333 because of the final notation for service labels.


Rossko57:
MQTT System Broker is obsolete and should be removed from OH3?  (There being no embedded broker now)

PR #1817, opened 2020-11-13
Changed comparison from equals() to reference

cweitkamp:
- Changed comparison from equals() to reference

Fixes https://github.com/openhab/openhab-addons/issues/8845

Signed-off-by: Christoph Weitkamp <github@christophweitkamp.de>

PR #2335, opened 2021-05-01
Fix more SAT findings and add a few suppressions

wborn:
Signed-off-by: Wouter Born <github@maindrain.net>

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ChannelCommandDescriptionProvider.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ChannelCommandDescriptionProvider.java
@@ -90,7 +90,7 @@ public class ChannelCommandDescriptionProvider implements CommandDescriptionProv
                 CommandDescription dynamicCommandDescription = dynamicCommandDescriptionProvider
                         .getCommandDescription(channel, originalCommandDescription, locale);
                 if (dynamicCommandDescription != null) {
-                    if (dynamicCommandDescription == originalCommandDescription) {
+                    if (dynamicCommandDescription.equals(originalCommandDescription)) {
                         logger.error(
                                 "Dynamic command description matches original command description. DynamicCommandDescriptionProvider implementations must never return the original command description. {} has to be fixed.",
                                 dynamicCommandDescription.getClass());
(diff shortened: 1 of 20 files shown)


bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ChannelStateDescriptionProvider.java
```diff
@@ -134,7 +134,7 @@ public Integer getRank() {
             StateDescription dynamicStateDescription = provider.getStateDescription(channel, originalStateDescription,
                     locale);
             if (dynamicStateDescription != null) {
-                if (dynamicStateDescription == originalStateDescription) {
```
cweitkamp, line 137: Can we revert this? There was a reason to use the comparison against the reference here: https://github.com/openhab/openhab-core/pull/1817

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ChannelCommandDescriptionProvider.java
```diff
@@ -90,7 +90,7 @@ public ChannelCommandDescriptionProvider(final @Reference ItemChannelLinkRegistr
                 CommandDescription dynamicCommandDescription = dynamicCommandDescriptionProvider
                         .getCommandDescription(channel, originalCommandDescription, locale);
                 if (dynamicCommandDescription != null) {
-                    if (dynamicCommandDescription == originalCommandDescription) {
```
cweitkamp, line 93: Can we revert this? There was a reason to use the comparison against the reference here: https://github.com/openhab/openhab-core/pull/1817

cweitkamp (commented):
Thanks you. I left some remarks inside. If that does not work we have to check the implementation of the  `equals()` method in both classes to suppress the warning in a proper way.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ChannelStateDescriptionProvider.java
wborn: Good catch! I'll restore it and add a comment so we won't run into this regression again. :blush:

cweitkamp (approved):
LGTM. Thanks.

PR #2337, opened 2021-05-02
[automation] match on complete channel UID

t-8ch:
Matching only on "topic.contains(channelUID)" will lead to false
positivies.

This was reported for the linuxinput binding, where pressing the "left"
key would also trigger rules for the "l" key.

The channels for these are
`linuxinput:input-device:test:keypresses#KEY_L` and
`linuxinput:input-device:test:keypresses#KEY_LEFT`

Signed-off-by: Thomas Weißschuh <thomas@t-8ch.de>

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ChannelEventTriggerHandler.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ChannelEventTriggerHandler.java
@@ -25,6 +25,7 @@ import org.openhab.core.automation.handler.TriggerHandlerCallback;
 import org.openhab.core.events.Event;
 import org.openhab.core.events.EventFilter;
 import org.openhab.core.events.EventSubscriber;
+import org.openhab.core.thing.ChannelUID;
 import org.openhab.core.thing.events.ChannelTriggeredEvent;
 import org.osgi.framework.BundleContext;
 import org.osgi.framework.ServiceRegistration;
@@ -47,7 +48,7 @@ public class ChannelEventTriggerHandler extends BaseTriggerModuleHandler impleme
     private final Logger logger = LoggerFactory.getLogger(ChannelEventTriggerHandler.class);
 
     private final String eventOnChannel;
-    private final String channelUID;
+    private final ChannelUID channelUID;
     private final Set<String> types = new HashSet<>();
     private final BundleContext bundleContext;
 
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ChannelEventTriggerHandler.java
```diff
@@ -88,7 +88,7 @@ public boolean apply(Event event) {
         boolean eventMatches = false;
         if (event instanceof ChannelTriggeredEvent) {
             ChannelTriggeredEvent cte = (ChannelTriggeredEvent) event;
-            if (cte.getTopic().contains(this.channelUID)) {
+            if (cte.getTopic().equals("openhab/channels/" + this.channelUID + "/triggered")) {
```
cweitkamp, line 91: Why do we check against the topic at all? The `ChannelTriggeredEvent` stores the `ChannelUID` inside which can be used for comparison:

https://github.com/openhab/openhab-core/blob/0972ad5ede205db5ebbcde5e2c5bbe205bb94621/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/events/ChannelTriggeredEvent.java#L65-L70

cweitkamp (commented):
Thanks for your contribution. I left a question.

cweitkamp:
btw Are you feeling comfortable with adding a unit test for the `ChannelEventTriggerHandler`?

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ChannelEventTriggerHandler.java
t-8ch: Good point, I'll change it.

cweitkamp (changes requested):
Looks great. Thanks.

bundles/org.openhab.core.automation/src/test/java/org/openhab/core/automation/internal/module/handler/ChannelEventTriggerHandlerTest.java
```diff
@@ -0,0 +1,52 @@
+package org.openhab.core.automation.internal.module.handler;
```
cweitkamp, line 1: Please add the license header.

cweitkamp (approved):
LGTM. Thanks.

PR #2347, opened 2021-05-06
[magic] Added Channel for DateTime Item

cweitkamp:
- Added Channel for DateTime Item


wborn (approved):
This extra magic LGTM. :mage_man::sparkles:

PR #2349, opened 2021-05-06
Display raw item state when formatting fails

### LINKED ISSUE #2037: OH3: Switch item with linked channel and hysteresis profile leads to error on the item
I'm using openHAB 3.0 with docker on a linux host system.

My use case is an alarm indicating that the humidity in my bathroom is over a certain upper value. When this happens the bathroom fan is turned on to bring the humidity down. The fan should stay on until the humidity drops below a certain lower value.

For this I configured a switch item carrying the alarm value and linked it to the humidity sensor with a channel configured for the hysteresis profile.

The expected outcome should be a switch item handling the alarm value with the hysteresis in place. The current outcome is that the item goes into an error state.

The item configuration is:

image

The channel link configuration is:

image

As you can see in the screen shots, the state is "Err". The probable cause might be the following log entry:

```
2020-12-30 00:41:22.012 [WARN ] [e.internal.SseItemStatesEventBuilder] - Exception while formatting value 'OFF' of item Badezimmer_Sicherheit_Feuchtigkeitsalarm with format '%.0f': f != java.lang.String
```

brianwarner:
With the addition of Profiles, there are situations where a Channel sends format information that
does not match the type. For example, "Timestamp on update" always sends a DateTime, but the channel
may specify a different format (e.g., if it is a battery level, "%.0f %%"). In such a case, we
attempt to format the DateTime according to the pattern, it (predictably and correctly) fails, and
this results in an error in the UI and a warning in the log.

While this is an expected error if a user uses an incorrect format code in an Item's metadata (user
error), it is somewhat unexpected if a user is choosing an option as presented in the UI. Because
this is purely a formatting issue, one solution is to direct users to add a formatting code if they
see an error. However, the disadvantage of this approach is that it doesn't work "out of the box."

This patch attempts to address the issue by displaying the raw Item state when formatting fails
instead of displaying "Err". This should provide a more balanced approach with a predictable
outcome. In addition, when a user is intentionally changing the pattern via metadata and gets it
wrong, they should still see info in the log that helps them get to the root of the problem, but
I've changed it from warn to info so it will be a little less noisy for those who choose to ignore
it.

This solution was discussed in https://github.com/openhab/openhab-core/issues/2037 and
https://community.openhab.org/t/timestamps-not-rendering-consistently-in-ui-3-0-2/121891/2

Thank you to @Rossko57 and @cweitkamp for their help in figuring it out, and in suggesting the
solution.

Fixes #2037

Signed-off-by: Brian Warner <brian@bdwarner.com>

### DIFF
--- a/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/internal/SseItemStatesEventBuilder.java
+++ b/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/internal/SseItemStatesEventBuilder.java
@@ -152,7 +152,8 @@ public class SseItemStatesEventBuilder {
                             try {
                                 displayState = state.format(pattern);
                             } catch (IllegalArgumentException e) {
-                                logger.info("Unable to format value '{}' of item {} with format '{}': {}, displaying raw state",
+                                logger.info(
+                                        "Unable to format value '{}' of item {} with format '{}': {}, displaying raw state",
                                         state, item.getName(), pattern, e.getMessage());
                                 displayState = state.toString();
                             }
--- a/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/internal/SseItemStatesEventBuilder.java
+++ b/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/internal/SseItemStatesEventBuilder.java
@@ -152,9 +152,9 @@ public class SseItemStatesEventBuilder {
                             try {
                                 displayState = state.format(pattern);
                             } catch (IllegalArgumentException e) {
-                                logger.warn("Exception while formatting value '{}' of item {} with format '{}': {}",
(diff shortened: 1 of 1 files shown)


cweitkamp (changes requested):
Thank you for your contribution.

Our static-code-analysis tool is not happy with your changes. You can fix it by running `mvn spotless:apply` in the related folder.

If you want to keep the feedback to the user than a WARNING is the correct log level. If not, you should decrease it to DEBUG.

brianwarner:
Thanks @cweitkamp, turns out it was a linter error. Easy fix once I got Maven up and running. I pushed two commits, one that satisfies the linter, and another that changes the log level to DEBUG.

I also saw @Rossko57's comment on the issue that there may be a more comprehensive way to solve this, although to be honest I'm at the boundaries of my knowledge on what else would need to be changed. Happy to look into other ways to solve this, as well.

cweitkamp (approved):
LGTM. Thanks.

Lets keep the issue as this maybe is not the final solution but a small step forward.

PR #2358, opened 2021-05-11
Fix unstable MqttBrokerConnectionTests

wborn:
The `timeoutWhenNotReachable` and `timeoutWhenNotReachableFuture` tests often fail on systems with a lot of load.


wborn:
Yes the plan was indeed to retrigger the other builds after the likelihood of success had increased. 🙂

PR #2362, opened 2021-05-13
Fix decimal separator issues when creating QuantityType from String

### LINKED ISSUE #2360: QuantityType has no decimal separator for certain locales
@cweitkamp @wborn
After upgrading to build 2375 the power readings of my FRITZ!DECT 200 are much to high.
Instead of eg 72 W it shows 72062,00 W. It's a factor of 1000.
Maybe a side effect of PR openhab/openhab-addons#10583 ?

wborn:
Fixes #2360

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityType.java
@@ -15,6 +15,7 @@ package org.openhab.core.library.types;
 import static org.eclipse.jdt.annotation.DefaultLocation.*;
 
 import java.math.BigDecimal;
+import java.text.DecimalFormatSymbols;
 import java.time.Instant;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
@@ -58,6 +59,7 @@ public class QuantityType<T extends Quantity<T>> extends Number
         implements PrimitiveType, State, Command, Comparable<QuantityType<T>> {
 
     private static final long serialVersionUID = 8828949721938234629L;
+    private static final char DOT_DECIMAL_SEPARATOR = '.';
     private static final BigDecimal HUNDRED = BigDecimal.valueOf(100);
 
     public static final QuantityType<Dimensionless> ZERO = new QuantityType<>(0, AbstractUnit.ONE);
@@ -102,6 +104,14 @@ public class QuantityType<T extends Quantity<T>> extends Number
             BigDecimal bd = new BigDecimal(value);
             quantity = (Quantity<T>) Quantities.getQuantity(bd, AbstractUnit.ONE, Scale.RELATIVE);
         } else {
+            char defaultDecimalSeparator = DecimalFormatSymbols.getInstance().getDecimalSeparator();
+            // The quantity is parsed using a NumberFormat based on the default locale.
+            // To prevent issues, any dot decimal separators are replaced by the default locale decimal separator.
+            if (DOT_DECIMAL_SEPARATOR != defaultDecimalSeparator
+                    && formatted.contains(String.valueOf(DOT_DECIMAL_SEPARATOR))) {
+                formatted = formatted.replace(DOT_DECIMAL_SEPARATOR, defaultDecimalSeparator);
+            }
+
             Quantity<T> absoluteQuantity = (Quantity<T>) Quantities.getQuantity(formatted);
             quantity = Quantities.getQuantity(absoluteQuantity.getValue(), absoluteQuantity.getUnit(), Scale.RELATIVE);
         }
--- a/bundles/org.openhab.core/src/test/java/org/openhab/core/internal/items/ItemStateConverterImplTest.java
+++ b/bundles/org.openhab.core/src/test/java/org/openhab/core/internal/items/ItemStateConverterImplTest.java
@@ -17,12 +17,16 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.Mockito.*;
 
+import java.util.Locale;
+import java.util.stream.Stream;
+
 import javax.measure.quantity.Length;
 import javax.measure.quantity.Temperature;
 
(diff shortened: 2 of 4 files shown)


J-N-K:
Are you sure this is sufficient? I didn't check if the old library could patse `1,200.5 W`? If so, your replacement would (with a german locale) result in `1,200,5 W` which will fail, because the correct format would be `1.200,5 W`.

wborn:
I haven't checked those parts either yet, that's one of the reasons why it is still a draft. :wink:

cweitkamp:
Thanks for looking into it. Would be nice to have some unit tests for this - if possible.

wborn:
I wrote a test to print how different strings are parsed depending on the default locale and ran it with OH 3.0 and OH 3.1:

```java
    /**
     * Locales having a different decimal separator to test string parsing and generation.
     */
    static Stream<Locale> locales() {
        return Stream.of(
                // ٫ (Arabic, Egypt)
                Locale.forLanguageTag("ar-EG"),
                // , (German, Germany)
                Locale.forLanguageTag("de-DE"),
                // . (English, United States)
                Locale.forLanguageTag("en-US"));
    }

    public void printParsedQuantity(String value) {
        String prefix = "new QuantityType<>(\"" + value + "\").toString() = ";

        try {
            System.out.println(prefix + new QuantityType<>(value).toFullString());
        } catch (Exception e) {
            System.out.println(prefix + "ERROR!");
        }
    }

    @ParameterizedTest
    @MethodSource("locales")
    public void testtest(Locale locale) {
        Locale.setDefault(locale);

        System.out.println(
                "Locale: " + locale.toLanguageTag() + " " + DecimalFormatSymbols.getInstance().getDecimalSeparator()
                        + DecimalFormatSymbols.getInstance().getGroupingSeparator());

        printParsedQuantity("123٫56 °C");
        printParsedQuantity("123,56 °C");
        printParsedQuantity("123.56 °C");

        printParsedQuantity("123٬123٬123٫56 °C");
        printParsedQuantity("123.123.123,56 °C");
        printParsedQuantity("123,123,123.56 °C");

        System.out.println();
    }
```

## OH 3.0 Output

```
Locale: ar-EG ٫٬
new QuantityType<>("123٫56 °C").toString() = ERROR!
new QuantityType<>("123,56 °C").toString() = ERROR!
new QuantityType<>("123.56 °C").toString() = 123.56 °C
new QuantityType<>("123٬123٬123٫56 °C").toString() = ERROR!
new QuantityType<>("123.123.123,56 °C").toString() = ERROR!
new QuantityType<>("123,123,123.56 °C").toString() = ERROR!

Locale: de-DE ,.
new QuantityType<>("123٫56 °C").toString() = ERROR!
new QuantityType<>("123,56 °C").toString() = ERROR!
new QuantityType<>("123.56 °C").toString() = 123.56 °C
new QuantityType<>("123٬123٬123٫56 °C").toString() = ERROR!
new QuantityType<>("123.123.123,56 °C").toString() = ERROR!
new QuantityType<>("123,123,123.56 °C").toString() = ERROR!

Locale: en-US .,
new QuantityType<>("123٫56 °C").toString() = ERROR!
new QuantityType<>("123,56 °C").toString() = ERROR!
new QuantityType<>("123.56 °C").toString() = 123.56 °C
new QuantityType<>("123٬123٬123٫56 °C").toString() = ERROR!
new QuantityType<>("123.123.123,56 °C").toString() = ERROR!
new QuantityType<>("123,123,123.56 °C").toString() = ERROR!
```

## OH 3.1 Output

```
Locale: ar-EG ٫٬
new QuantityType<>("123٫56 °C").toString() = 123.56 °C
new QuantityType<>("123,56 °C").toString() = 123 °C
new QuantityType<>("123.56 °C").toString() = 123 °C
new QuantityType<>("123٬123٬123٫56 °C").toString() = 1.2312312356E8 °C
new QuantityType<>("123.123.123,56 °C").toString() = 123 °C
new QuantityType<>("123,123,123.56 °C").toString() = 123 °C

Locale: de-DE ,.
new QuantityType<>("123٫56 °C").toString() = 123 °C
new QuantityType<>("123,56 °C").toString() = 123.56 °C
new QuantityType<>("123.56 °C").toString() = 12356 °C
new QuantityType<>("123٬123٬123٫56 °C").toString() = 123 °C
new QuantityType<>("123.123.123,56 °C").toString() = 1.2312312356E8 °C
new QuantityType<>("123,123,123.56 °C").toString() = 123.123 °C

Locale: en-US .,
new QuantityType<>("123٫56 °C").toString() = 123 °C
new QuantityType<>("123,56 °C").toString() = 12356 °C
new QuantityType<>("123.56 °C").toString() = 123.56 °C
new QuantityType<>("123٬123٬123٫56 °C").toString() = 123 °C
new QuantityType<>("123.123.123,56 °C").toString() = 123.123 °C
new QuantityType<>("123,123,123.56 °C").toString() = 1.2312312356E8 °C
```

So it seems it never was able to parse thousands separators or locale specific separators. Only `.` was a valid decimal separator. With the changes in this PR and the new UoM library is better with parsing, but it doesn't always get it right.

The OH code has no input validation and the exceptions thrown in OH 3.0 were thrown by the old UoM library.

We probably could improve parsing locale specific strings by also creating a constructor with a `Locale` parameter one day. Then it could replace the locale specific separators with those of the default JVM `Locale` before creating a quantity using the library.

cweitkamp (approved):
The build is successful again with your changes.

I think the important thing is the correct parsing of decimal separators is working again. Of course thousands separators are a use case but from my pov it is used rarely in the context of openHAB.

Lets merge this for now and cross fingers we eliminated the culprit.

wborn:
> We probably could improve parsing locale specific strings by also creating a constructor with a Locale parameter one day. Then it could replace the locale specific separators with those of the default JVM Locale before creating a quantity using the library.

I have created https://github.com/unitsofmeasurement/indriya/issues/351 because I think other projects using Indrya would also benefit from being able to specify a locale when parsing strings. When implemented, we could for instance replace the workaround in this PR with using the en-US locale as default and simply pass the locale parameter of a new constructor to the Indrya method. :slightly_smiling_face:

PR #2353, opened 2021-05-08
[profiles] Remove update of timestamp on Change/Update for Item update

cweitkamp:
- Remove update of timestamp on Change/Update for Item update

Fixes #1537

Signed-off-by: Christoph Weitkamp <github@christophweitkamp.de>

PR #2364, opened 2021-05-15
Added "timestamp-trigger" Profile

cweitkamp:
- Added Timestamp on Trigger Profile

This Profile updates the State of an DateTime Item to now each time the Channel is triggered.

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/profiles/SystemProfileFactory.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/profiles/SystemProfileFactory.java
@@ -67,13 +67,13 @@ public class SystemProfileFactory implements ProfileFactory, ProfileAdvisor, Pro
             RAWBUTTON_TOGGLE_ROLLERSHUTTER_TYPE, RAWBUTTON_TOGGLE_SWITCH_TYPE, RAWROCKER_DIMMER_TYPE,
             RAWROCKER_NEXT_PREVIOUS_TYPE, RAWROCKER_ON_OFF_TYPE, RAWROCKER_PLAY_PAUSE_TYPE,
             RAWROCKER_REWIND_FASTFORWARD_TYPE, RAWROCKER_STOP_MOVE_TYPE, RAWROCKER_UP_DOWN_TYPE, TIMESTAMP_CHANGE_TYPE,
-            TIMESTAMP_OFFSET_TYPE, TIMESTAMP_UPDATE_TYPE);
+            TIMESTAMP_OFFSET_TYPE, TIMESTAMP_TRIGGER_TYPE, TIMESTAMP_UPDATE_TYPE);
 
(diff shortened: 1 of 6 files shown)


bundles/org.openhab.core.thing/src/main/resources/OH-INF/i18n/SystemProfiles.properties
```diff
@@ -23,4 +23,5 @@ profile.config.system.timestamp-offset.offset.label = Offset
 profile.config.system.timestamp-offset.offset.description = Offset to be applied on the state towards the item. The negative offset will be applied in the reverse direction.
 profile.config.system.timestamp-offset.timezone.label = Time Zone
 profile.config.system.timestamp-offset.timezone.description = A time zone to be applied on the state.
+profile-type.system.timestamp-trigger.label = Timestamp on Trigger
```
kaikreuzer, line 26:
```suggestion
profile-type.system.timestamp-trigger.label = Timestamp on trigger
```

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/profiles/SystemProfileFactory.java
```diff
@@ -179,7 +181,7 @@ public SystemProfileFactory(final @Reference ChannelTypeRegistry channelTypeRegi
                 case STATE:
                     return DEFAULT;
                 case TRIGGER:
-                    return null;
+                    return TIMESTAMP_TRIGGER;
```
kaikreuzer, line 184: Pfew, I am not so sure if this profile should really become the generally suggested one...
I was just [looking into profiles](https://github.com/openhab/openhab-webui/issues/886#issuecomment-844381424) today and was thinking that a "map-to-string" profile (need to find a better name) would probably make sense as a default profile for trigger channels. Wdyt?

kaikreuzer (changes requested):
Many thanks! Just two small comments/questions.

bundles/org.openhab.core.thing/src/main/resources/OH-INF/i18n/SystemProfiles.properties
cweitkamp: We changed spelling for other timestamp Profiles to camel case in https://github.com/openhab/openhab-core/pull/2353 (e.g. "Timestamp on Change"). I would like to keep it.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/profiles/SystemProfileFactory.java
cweitkamp: I reverted this change.

cweitkamp: If I understand your idea correctly you like to add a `TriggerProfile` which converts the triggered event into a `StringType`? 

What about simply "toString"?

kaikreuzer (approved):
Thanks!

PR #2365, opened 2021-05-15
Support parsing localized strings with DecimalType, PercentType and QuantityType

wborn:
Adds constructors with a Locale parameter for parsing localized string to DecimalType, PercentType and QuantityType.
This allows for parsing locale specific group and decimal seperators in strings.

Also fixes:
* IllegalArgumentException not being thrown if numbers in strings are only partially parsed as QuantityType
* MeasurementParseException not caught and rethrown as IllegalArgumentException

Several new unit tests cover the new functionality and fixed regressions.

---

Further improves https://github.com/openhab/openhab-core/pull/2362

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DecimalType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DecimalType.java
@@ -13,7 +13,11 @@
 package org.openhab.core.library.types;
 
 import java.math.BigDecimal;
+import java.text.DecimalFormat;
+import java.text.NumberFormat;
+import java.text.ParsePosition;
 import java.util.IllegalFormatConversionException;
+import java.util.Locale;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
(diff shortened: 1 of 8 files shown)


wborn:
Seems like some add-ons expect different exceptions to be thrown for illegal arguments since some tests now fail:

https://ci.openhab.org/job/PR-openHAB-Addons/4221/#showFailuresLink

IMHO it is better to always throw the same exception (IllegalArgumentException, new behavior) instead just throwing whatever runtime exception is thrown by the underlying implementations (NumberFormatException, MeasurementParseException, etc.).
This behavior was never documented or covered with unit tests in the Core.

WDYT @openhab/core-maintainers?

wborn:
Throwing a  `NumberFormatException` could still make sense for the `DecimalType` and `PercentType` constructors. It extends the `IllegalArgumentException` and has as doc:

```
 * Thrown to indicate that the application has attempted to convert
 * a string to one of the numeric types, but that the string does not
 * have the appropriate format.
```

For the `QuantityType` it is a bit more difficult since it can have a unit, but the unit is optional.

wborn:
Besides that, the `PercentType` also throws an `IllegalArgumentException` when the value is not in the `[0, 100]` range. So to be on the safe side, when using these constructors, you probably want to catch any `IllegalArgumentException`.

PR #2373, opened 2021-05-20
[automation] Remove readOnly flag on script type parameter

ghys:
Not sure if there's a good rationale behind it being read-only (maybe because it's not supposed to change after having been set) but removing it will fix https://github.com/openhab/openhab-webui/issues/1061.


kaikreuzer (approved):
I assume that dates back to times that scripts were provided only from the file system...

PR #2374, opened 2021-05-20
Rework thrown exceptions for DecimalType, PercentType and QuantityType

wborn:
Throw NumberFormatException when numbers are invalid and cannot be parsed to BigDecimals.
Add JavaDocs to document when which exception is thrown.
Add more unit tests to cover the thrown exceptions.

Related to #2365


kaikreuzer (approved):
Very nice - finally some JavaDoc here 😁

Issue #2148, opened 2021-01-23
Rules have misleading error-messages, if e.g. operator is missing


binderth:
As described here: https://community.openhab.org/t/oh3-rule-condition-module-is-not-well-configured/115150 I ran into a problem, while I forgot to set the "operator" in the rules-UI. So the rule wasn't fired and the logs showed:
```
2021-01-23 12:48:00.580 [ERROR] [le.handler.ItemStateConditionHandler] - Module is not well configured: itemName=Comfoair_Automatic  operator=null  state = ON
```
Perhaps the wording can be changed to something more speaking like "Condition of rule is not well configured: ..."? I couldn't make out, what I did wrong at first, unless I repeatedly opened the condition in the GUI and there I saw the missing operator.

J-N-K:
I do not understand the difference between the message above and the message you suggest. `operator=null` seems to be quite clear to me.

PR #2376, opened 2021-05-22
[automation] Fixed default value definition for module configuration parameters

cweitkamp:
- Fixed default value definition for module configuration parameters

The field is called "defaultValue" in `ConfigDescriptionParameterDTO` and was not parsed before.

Related to #2148

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/resources/OH-INF/automation/moduletypes/EphemerisConditions.json
+++ b/bundles/org.openhab.core.automation/src/main/resources/OH-INF/automation/moduletypes/EphemerisConditions.json
@@ -10,7 +10,7 @@
 					"type": "INTEGER",
 					"label": "Offset",
 					"description": "Today +/- offset days (+1 = tomorrow, -1 = yesterday).",
-					"default": 0,
+					"defaultValue": 0,
 					"stepsize": 1,
 					"required": false
 				}
@@ -26,7 +26,7 @@
 					"type": "INTEGER",
 					"label": "Offset",
 					"description": "Today +/- offset days (+1 = tomorrow, -1 = yesterday).",
-					"default": 0,
+					"defaultValue": 0,
 					"stepsize": 1,
 					"required": false
 				}
@@ -42,7 +42,7 @@
 					"type": "INTEGER",
 					"label": "Offset",
 					"description": "Today +/- offset days (+1 = tomorrow, -1 = yesterday).",
-					"default": 0,
+					"defaultValue": 0,
 					"stepsize": 1,
 					"required": false
 				}
@@ -58,7 +58,7 @@
 					"type": "INTEGER",
 					"label": "Offset",
 					"description": "Today +/- offset days (+1 = tomorrow, -1 = yesterday).",
-					"default": 0,
+					"defaultValue": 0,
 					"stepsize": 1,
 					"required": false
 				}
@@ -81,7 +81,7 @@
 					"type": "INTEGER",
 					"label": "Offset",
 					"description": "Today +/- offset days (+1 = tomorrow, -1 = yesterday).",
-					"default": 0,
+					"defaultValue": 0,
 					"stepsize": 1,
 					"required": false
 				}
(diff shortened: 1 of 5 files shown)


bundles/org.openhab.core.automation/src/main/resources/OH-INF/automation/moduletypes/EphemerisConditions.json
```diff
@@ -10,7 +10,7 @@
 					"type": "INTEGER",
 					"label": "Offset",
 					"description": "Today +/- offset days (+1 = tomorrow, -1 = yesterday).",
-					"default": 0,
+					"defaultValue": 0,
```
kaikreuzer, line 13: Imho, the JSON representation of a config description param should be identical to the XML representation where this attribute is called `default`. The only reason why it is called `defaultValue` in the DTO is that `default` is a reserved keyword in Java and thus cannot be used there.
The better fix would therefore imho be to adjust the JSON parsing.

bundles/org.openhab.core.automation/src/main/resources/OH-INF/automation/moduletypes/GenericConditions.json
```diff
@@ -66,7 +66,7 @@
 					"type": "TEXT",
 					"description": "the compare operator, allowed are <,>,=,!=,>=,<= matches",
 					"required": true,
```
kaikreuzer, line 68: I wonder why we have required set to true here, if we have a default value defined. According to our usual logic, it should be false - same for other occurrences here in this PR (I know, it was like that before, but I am just seeing it now).

bundles/org.openhab.core.automation/src/main/resources/OH-INF/automation/moduletypes/ItemConditions.json
```diff
@@ -18,8 +18,6 @@
 					"type": "TEXT",
 					"label": "Operator",
 					"description": "the compare operator (one of =,<,>,!=,>=,<=)",
-					"required": false,
-					"limitToOptions": true,
```
kaikreuzer, line 22: Why did you remove this attribute?

kaikreuzer (changes requested):

bundles/org.openhab.core.automation/src/main/resources/OH-INF/automation/moduletypes/EphemerisConditions.json
cweitkamp: We can add the `@SerializedName` annotation in the ConfigDescriptionParameterDTO` for this field. e.g.

```java
@SerializedName(value="default", alternate="defaultValue")
```

The first one ("value") will be used as a default for serialization and deserialization, while the second ("defaultValue") will be used to indicate an alternate name of the JSON field that we want to map. But in this way we will introduce a regression as it will be used for exposing configuration description via REST interface too. Which name should be used for exposing the DTO?

kaikreuzer: @ghys What's your opinion on the breaking change? Is this something that can be easily adjusted in the Main UI?
It seems to me as half of our config descriptions might not work since we do not use the same attribute name in all places. As it is an optional one, I'd actually hope that the change won't really cause issues in the UI, it might just not recognize the "other half" config descriptions, which currently use `defaultValue` in their JSON.

ghys: I count only 4 (relevant) occurrences of `defaultValue` in the UI's code, 2 here:
https://github.com/openhab/openhab-webui/blob/c71db73a6163026f383d352ace8362bc41ccd426/bundles/org.openhab.ui/web/src/components/config/config-sheet.vue#L88
and 2 related to https://github.com/openhab/openhab-webui/pull/1049.
So overall it's feasible but might break a few widgets that make use of the feature in the above PR (which is relatively recent so no big impact anyways).

kaikreuzer: Thanks for checking - should be a rather safe change then.

cweitkamp:
I will close this PR and submit a new one.

PR #2378, opened 2021-05-23
[automation] add tags to item modules

kaikreuzer:
Tags are required to enable auto-mapping of inputs and outputs of modules in a rule, see https://github.com/openhab/openhab-addons/pull/10205#discussion_r636433332.

Also made command/state config params optional and being regarded, if they are set. This way it is avoided that auto-wiring prevents users from defining dedicated values in the actions.


cweitkamp (approved):
LGTM. Thanks.

Thus auto-mapping feature sounds really interesting. I do not know very much about it but like to learn. Do we have some documentation about it?

kaikreuzer:
@cweitkamp The best documentation about auto-mapping can be found in [its JavaDoc](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleEngineImpl.java#L1246-L1255).
Note that it is a feature on top of "connections" between outputs and inputs. We don't use this feature in the Main UI yet - it would also be possible for the user to explicitly wire outputs and inputs together. The auto-mapping only deals with "what is left".

PR #2382, opened 2021-05-24
New translations SystemProfiles.properties (Poland)

grzegorz914:


cweitkamp:
Hi @grzegorz914.

Thanks for adding and improving Polish translations. Please use [Crowdin](https://crowdin.com/project/openhab-core/pl) platform for it. Otherwise the repository is not in synch and we might loose track on the correct ones.

PR #2383, opened 2021-05-25
[config] Changed 'ConfigDescriptionParameterDTO' field serialization 'defaultValue' -> 'default'

cweitkamp:
- Changed `ConfigDescriptionParameterDTO` field serialization: "defaultValue" -> "default"
- Fixed a minor bug in `ConfigDescriptionDTOMapper` for field "unit"
- Added unit tests for `ConfigDescriptionDTOMapper`, `ConfigDescriptionResource` and `AutomationIntegrationJsonTest`

Supersedes #2376
Related to #2148

### DIFF
--- a/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/dto/ConfigDescriptionParameterDTO.java
+++ b/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/dto/ConfigDescriptionParameterDTO.java
@@ -17,9 +17,10 @@ import java.util.List;
 
 import org.openhab.core.config.core.ConfigDescriptionParameter.Type;
 
+import com.google.gson.annotations.SerializedName;
+
 /**
- * This is a data transfer object that is used to serialize parameter of a
- * configuration description.
+ * This is a data transfer object that is used to serialize parameter of a configuration description.
  *
  * @author Dennis Nobel - Initial contribution
  * @author Alex Tugarev - Extended for options and filter criteria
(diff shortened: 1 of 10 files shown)


bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/dto/ConfigDescriptionParameterDTO.java
```diff
@@ -29,6 +30,7 @@
 public class ConfigDescriptionParameterDTO {
 
     public String context;
+    @SerializedName(value = "default")
```
cweitkamp, line 33: As mentioned in https://github.com/openhab/openhab-core/pull/2376#discussion_r637870924 we can use the following syntax to allow both names for parsing.

```suggestion
    @SerializedName(value="default", alternate="defaultValue")
```

kaikreuzer: Sounds like the safer option in any case. Shall we use this one? What is your preference?

cweitkamp: I'd prefer adding the fallback.

kaikreuzer (approved):
Lgtm, thank you!
I'll wait with merging it until after the 3.1.0.M5 build - just to make sure that we do not introduce any regression that we might have overlooked.

PR #2388, opened 2021-05-29
[automation] Expose TriggerHandlerCallback scheduler

fwolter:
Originated from https://github.com/openhab/openhab-addons/pull/10205#discussion_r637600080

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/handler/ModuleHandler.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/handler/ModuleHandler.java
@@ -26,16 +26,6 @@ import org.openhab.core.automation.ModuleHandlerCallback;
 public interface ModuleHandler {
 
     /**
-     * Initializes the module handler, e.g. schedules jobs.
-     * <p>
-     * This method is only called, if the {@link ModuleHandlerCallback} is set.
-     * <p>
-     * The framework expects this method to be non-blocking and return quickly. For longer running initializations,
-     * the implementation has to take care of scheduling a separate job.
-     */
-    public void initialize();
-
-    /**
      * The method is called by RuleManager to free resources when {@link ModuleHandler} is released.
      */
     public void dispose();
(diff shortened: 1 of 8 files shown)


bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/handler/ModuleHandler.java
```diff
@@ -25,6 +25,16 @@
...
 
+    /**
+     * Initializes the module handler, e.g. schedules jobs.
+     * <p>
+     * This method is only called, if the {@link ModuleHandlerCallback} is set.
+     * <p>
+     * The framework expects this method to be non-blocking and return quickly. For longer running initializations,
+     * the implementation has to take care of scheduling a separate job.
+     */
+    public void initialize();
```
kaikreuzer, line 36: Why do you introduce this method? Can't the handler initialize stuff when the callback is set?
I don't really see the need for both methods. We could also change the existing `setCallback` to be `initialize` with the callback as a parameter, but this would be a breaking change.

kaikreuzer:
Build fails with 
```
[ERROR] /home/jenkins/jenkins-agent1/workspace/PR-openHAB-Core/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/delegates/SimpleTriggerHandlerCallbackDelegate.java:[28,14] The type org.openhab.core.automation.module.script.rulesupport.internal.delegates.SimpleTriggerHandlerCallbackDelegate must implement the inherited abstract method org.openhab.core.automation.handler.TriggerHandlerCallback.getScheduler()
```

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/handler/ModuleHandler.java
fwolter: You're right that's redundant. I removed the initialize method again.

kaikreuzer (approved):
Thanks!

Issue #2223, opened 2021-03-03
[jupnp] waits 10 minutes after failed request leading to connection problems with some bindings


thinkingstone:
Openhab 3.0
Jupnp 2.5.2
Panasonic binding 3.0.snapshot from https://github.com/thinkingstone/openhab-addons
also reproducible with the official upnpcontrol binding

I discovered this issue while testing a migration of the panasonictv binding to OH3. 
See also: https://community.openhab.org/t/panasonictv-oh3/108124

Situation:
1. Openhab is online and jupnp and the panasonic binding is started
2. TV thing and items are configured. 
3. The TV is off. 
4. Now if I switch the TV on the binding can not connect to the TV for 10 minutes. 
5. After 10 minutes the binding can connect to the TV and is working correctly. 

Expected behaviour: the binding can connect to the TV shortly after switching it on. 

This is due: 
1. The TV sends a upnp event that it is ready
2. Jupnp retrieves this event and tries to get the service description 
3. but the TV webserver does not deliver the service description. Okay this an issue from panasonic but that is not solvable by us. 
4. Jupnp treats this as a failed request and does not retry request for 10 minutes. 

This can be seen in org.jupnp-2.5.2.jar!/org/jupnp/transport/spi/AbstractStreamClient.class

Introduced with https://github.com/jupnp/jupnp/pull/104
Optimize device information retrieval error handling #104

This is also reproducible with the upnpcontrol binding. 

The following log files need these log levels: 

org.jupnp                                          │ ERROR
org.jupnp.protocol                                 │ INFO
org.jupnp.registry                                 │ INFO
org.jupnp.transport                                │ TRACE

I would like to discuss possible solutions and maybe I can provide a pull request for that. 

possible solutions that I can see: 
1. change jupnp AbstractStreamClientConfiguration retryAfterSeconds to a much lower time-out. Lets say 30 Seconds. But I could not figure out why 10 minutes was chosen. Maybe there is a good reason for this. This should then be a issue in jupnp repo. 
2. change openhab so that it configures jupnp with a lower retryAfterSeconds. How is that done?
3. some other solution?

[logfile.txt](https://github.com/openhab/openhab-core/files/6075101/logfile.txt)

morph166955:
Please see https://github.com/jupnp/jupnp/pull/129 and see if it helps you.

thinkingstone:
How can I test that? 
[...]

PR #2396, opened 2021-06-01
upgrade to JUPnP 2.6.0

kaikreuzer:
Fixes https://github.com/openhab/openhab-core/issues/1886
Fixes https://github.com/openhab/openhab-core/issues/2223

### DIFF
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -143,7 +143,7 @@
     <dependency>
       <groupId>org.jupnp</groupId>
       <artifactId>org.jupnp</artifactId>
-      <version>2.5.2</version>
+      <version>2.6.0</version>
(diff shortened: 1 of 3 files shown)


morph166955:
Leonardo-DiCaprio-Toast-Fireworks-Gif

wborn (approved):
Thank you!

morph166955:
@kaikreuzer Given all of the potential touch points this can have, I wrote up a fun but hopefully helpful post on the forum on how to use the different configuration options (link above).  Do you think it may be helpful to specifically call this out (or something similar) when 3.1GA is done so people know to go look for it if they are having problems?  Would it be helpful to go through and list out all the different upnp based bindings that this could impact?

PR #2397, opened 2021-06-04
Fix group member links from rest/items/{itemname}

### LINKED ISSUE #2252: Openhab 3.0, Error in REST API?
Please find documentation here:
https://community.openhab.org/t/error-in-rest-api/119506/7

pacive:

Not really happy with hardcoding the path for the UriBuilder, but couldn't find a cleaner way.

The problem is that the UriBuilder is created without templates when creating it from the UriInfo, which caused it to always return the requested path, regardless of what arguments was passed to the `.build()` method.

Fixes #2252


kaikreuzer:
Thanks for your bug report and the analysis @pacive!
I looked into it and came up with a slightly more elegant fix - I created https://github.com/openhab/openhab-core/pull/2399 for it.

pacive:
Great! I'll close this one then.

PR #2398, opened 2021-06-05
[automation] Ephemeris Offset fix

Cossey:
This PR fixes the ephemeris offset which does not presently work.

Discussed in [this community post](https://community.openhab.org/t/ephemeris-condition-with-offset-not-working-in-rules-in-oh-3-1-0m5/123088/3?u=cossey)


cweitkamp (approved):
LGTM. Thanks for fixing it.

Issue #2350, opened 2021-05-06
[REST API] Allow user access to UUID resource


jsetton:
Related to the discussion in #1570, I am currently working on refactoring the Alexa skill code. Part of it is the need to generate an [Alexa custom identifier](https://developer.amazon.com/en-US/docs/alexa/device-apis/alexa-discovery-objects.html#additionalattributes-object) globally unique across user accounts.

Having the ability to use the server UUID as part of this identifier would be ideal. Unfortunately, this data point requires admin access currently. Is there any downside allowing user access to the UUID resource?

digitaldan:
hmm, we have the UUID as part of the registration for openHAB cloud.  Could we expose something there just for the skill and then for those not using the cloud service, make the UUID a optional parameter to self hosted skills targeting an OH instance directly ?

jsetton:
We could. I was trying to limit the number of ways on how to determine that data point since that resource is available for OH2 instances.

If it's a problem to allow user access to the UUID resource, we could just have the skill generate custom identifier for OH2 instances and only focus having the planned Alexa binding handle that part for OH3 instances.

digitaldan:
Yeah, reading #1570 sounds like its going to be a tough argument to make, so i would go with an alternative.  I was also thinking if we needed to, it may be possible to have myopenhab as part of the authentication flow for alexa, forward the user to log into their openhab after logging into the cloud, then request an api token, but this would be a lot more work if its even possible, and require a double login, so maybe not the best solution.  I'm not sure if somehow we could have the  cloud binding grab a token, but i worry we start degrading the security at this point.    

So long story short, your alternative sounds appealing ;-)

jsetton:
> Yeah, reading #1570 sounds like its going to be a tough argument to make

My understanding is that discussion was more about UUID potentially being added to the root resource then needing to be secured. And ultimately, the consensus was to keep it separate since it has its own resource endpoint.

> I was also thinking if we needed to, it may be possible to have myopenhab as part of the authentication flow for alexa, forward the user to log into their openhab after logging into the cloud, then request an api token, but this would be a lot more work if its even possible, and require a double login, so maybe not the best solution.

I agree. Too many authentications for little reward. Once the planned Alexa binding is implemented, there will be no need for it. This is more to keep consistency between skill only and skill + binding setup in the future.

> So long story short, your alternative sounds appealing ;-)

[...]

PR #2402, opened 2021-06-07
Make UUID available to users

kaikreuzer:
Implements https://github.com/openhab/openhab-core/issues/2350

### DIFF
--- a/bundles/org.openhab.core.id/src/main/java/org/openhab/core/id/internal/UUIDResource.java
+++ b/bundles/org.openhab.core.id/src/main/java/org/openhab/core/id/internal/UUIDResource.java
@@ -34,7 +34,6 @@ import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.media.Content;
 import io.swagger.v3.oas.annotations.media.Schema;
 import io.swagger.v3.oas.annotations.responses.ApiResponse;
-import io.swagger.v3.oas.annotations.security.SecurityRequirement;
 import io.swagger.v3.oas.annotations.tags.Tag;
 
 /**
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.id/src/main/java/org/openhab/core/id/internal/UUIDResource.java
```diff
@@ -49,7 +49,7 @@
 @JaxrsName(UUIDResource.PATH_UUID)
 @JaxrsApplicationSelect("(" + JaxrsWhiteboardConstants.JAX_RS_NAME + "=" + RESTConstants.JAX_RS_NAME + ")")
 @Path(UUIDResource.PATH_UUID)
-@RolesAllowed({ Role.ADMIN })
+@RolesAllowed({ Role.ADMIN, Role.USER })
 @SecurityRequirement(name = "oauth2", scopes = { "admin" })
```
cweitkamp, line 53: Do we have to remove the `SecurityRequirement` to make this endpoint accessible for users? I could not find any other user defined interface method with it.

kaikreuzer: Good point. I actually didn't notice that we have those annotations and am not sure how/if they are considered at all. But it seems to make sense to remove it indeed.

cweitkamp: Thanks for removing it. Can you please apply the formatter?

pavog:
Hello everyone!
The `@SecurityRequirement` annotation provides the Open Api specification generator with the information that authentication is required to access this endpoint.
In the end, this leads to:



 ```yaml
 /uuid:
    get:
      tags:
        - uuid
      summary: A unified unique id.
      operationId: getUUID
      responses:
        '200':
          description: OK
          content:
            text/plain:
              schema:
                type: string
```

vs


```yaml
  /uuid:
    get:
      tags:
        - uuid
      summary: A unified unique id.
      operationId: getUUID
      responses:
        '200':
          description: OK
          content:
            text/plain:
              schema:
                type: string
      security:
        - oauth2:
            - admin
```

For more info see [OpenAPI 3 docs -> Authentication and Authorization](https://swagger.io/docs/specification/authentication/).

kaikreuzer:
Thanks for the explanation, @pavog!

bundles/org.openhab.core.id/src/main/java/org/openhab/core/id/internal/UUIDResource.java
kaikreuzer: Sorry, didn't notice. Done now!

cweitkamp (approved):
LGTM. Thanks.

PR #2412, opened 2021-06-24
[item] Create a deep copy of the first found fragment before merging other fragments into it

### LINKED ISSUE #2409: Update of dynamic state options in UIs is broken
@cweitkamp : we have a major issue. The update of state options in UIs is now broken in snapshot 2431.
This is certainly a consequence of PR #1505 

Note that the new event is working well, I was able to catch it in the remote openHAB binding.

The problem is that after you update your state options provider for a channel, this update is ignored in UIs (Main UI and Basic UI). And this is not a cache problem of course.

Could it be because the webui is not yet in phase with the core changes ?

cweitkamp:
- Create a deep copy of the first found fragment before merging other fragments into it

Fixes #2409

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/service/StateDescriptionServiceImpl.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/service/StateDescriptionServiceImpl.java
@@ -75,7 +75,8 @@ public class StateDescriptionServiceImpl implements StateDescriptionService {
 
             // we pick up the first valid StateDescriptionFragment here:
             if (result == null) {
-                result = (StateDescriptionFragmentImpl) fragment;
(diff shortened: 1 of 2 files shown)


lolodomo (approved):
Thank you a lot for the fix. 

I have not checked in detail your changes in tests.

kaikreuzer (approved):
Many thanks for looking into this!

kaikreuzer:
I've directly launched a new core build: https://ci.openhab.org/job/openHAB-Core/796/

lolodomo:
Fantastic, I just tested your change in my Eclipse environment and it clearly solves the issue. List of state options is now correct in Basic UI, in Main UI and in item REST API.
Thank you a lot for fixing that quickly.

I will now be able to work on the UI changes to consider the new event.

PR #2421, opened 2021-06-29
RuleResource: Add schema specification for "Location" header

pavog:
Openapi requires either a `schema` or `content` property for `/rules.post.responses.201.headers.Location`. 
See openapi documentation [Description of Responses](https://swagger.io/docs/specification/describing-responses/#headers) for details. 

The code 
```java
final Rule newRule = ruleRegistry.add(RuleDTOMapper.map(rule));
return Response.status(Status.CREATED)
    .header("location", "rules/" + URLEncoder.encode(newRule.getUID(), StandardCharsets.UTF_8)).build();
```
tells us that the "Location" header is a string with the url to the newly created rule.

We should add the schema specification for the header as a string by extending the annotation `@Header` within the `@ApiResponse` with `schema = @Schema(implementation = String.class)`.


kaikreuzer (approved):
lgtm, thanks!

PR #2432, opened 2021-07-20
Use exec:exec instead of exec:java goal to avoid classloaders leak

gnodet:
This should help a lot as a workaround to https://github.com/mvndaemon/mvnd/issues/440 which has no real fix.

### DIFF
--- a/bundles/org.openhab.core.model.item/pom.xml
+++ b/bundles/org.openhab.core.model.item/pom.xml
@@ -47,19 +47,16 @@
         <artifactId>exec-maven-plugin</artifactId>
         <version>1.4.0</version>
         <configuration>
-          <includeProjectDependencies>true</includeProjectDependencies>
-          <includePluginDependencies>true</includePluginDependencies>
+          <executable>java</executable>
           <classpathScope>compile</classpathScope>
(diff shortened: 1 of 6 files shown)


gnodet:
@rogierhofboer can you have a look and see if this helps with mvndaemon/mvnd#440 ?

rogierhofboer:
@gnodet This fixes the problem for me. But shouldn't it also work with `java` instead of `exec`?

gnodet:
Yes, it should work with `java`, but the problems is not in `mvnd`, rather in the invoked java code, in this case, the eclipse EMF generator.  So you can raise an issue on their side, but I can't do anything in mvnd if they keep references to the classloaders somehow.

kaikreuzer:
Thanks for this fix @gnodet! May I ask you to sign-off your commit as per our [contribution guidelines](https://www.openhab.org/docs/developer/contributing.html#sign-your-work)? Thanks!

gnodet:
> Thanks for this fix @gnodet! May I ask you to sign-off your commit as per our [contribution guidelines](https://www.openhab.org/docs/developer/contributing.html#sign-your-work)? Thanks!

Done !

kaikreuzer (approved):
Perfect, thanks a lot!

PR #2435, opened 2021-07-22
[automation] Allow UoM in 'ItemStateCondition'

### LINKED ISSUE #2426: [automation] UoM in ItemStateCondition
Currently it is not possible to use UoM in ItemStateConditions. 

The `ItemStateConditionHandler` correctly creates a `QuantityType` as `comparingState`:

https://github.com/openhab/openhab-core/blob/79edf2b9e643ad802c31b2e5f54edb93534277fd/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ItemStateConditionHandler.java#L93

however, the `itemState` is always created as `DecimalType`: 

https://github.com/openhab/openhab-core/blob/79edf2b9e643ad802c31b2e5f54edb93534277fd/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ItemStateConditionHandler.java#L95

Later in the code the condition is only evaluated if the `comparingState` is of `DecimalType` (e.g.):

https://github.com/openhab/openhab-core/blob/79edf2b9e643ad802c31b2e5f54edb93534277fd/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ItemStateConditionHandler.java#L104-L107

IMO the `itemState` should be created with the same type as `comparingState` and the check for numeric conditions should include both, `DecimalType` and `QuantityType`

cweitkamp:
- Allow UoM in `ItemStateCondition`

The user will see a warning if Item has a QuantityState and the comparison value is defined as decimal (similar to `offset` Profile).

```
[main] WARN org.openhab.core.automation.internal.module.handler.ItemStateConditionHandler - Received a QuantityType state '23 °C' with unit, but the condition is defined as a plain number without unit (5), please consider adding a unit to the condition.
```

Fixes #2426


kaikreuzer (approved):
lgtm, thanks!

PR #2437, opened 2021-07-23
Consistent ordering of entries in jsondb

### LINKED ISSUE #2436: Consistent ordering of entries in jsondb to ensure minimal delta
As discussed in https://community.openhab.org/t/storing-configuration-userdata-in-git/124727/, we can sometimes experience "unexpected" re-ordering of JSON objects entries in jsondb.

This is on its own very unharmful on its own, as no program or application should rely on the order of entries of JSON objects. openHAB will naturally parse the JSON to corresponding Java `Map`  or similar, no matter what is the literal order in the file.

Similar problem most likely exists with JSON arrays and Java `Set`'s, although the JSON object structure is much more utilized by openHAB based on a quick glance.

However, the unexpected re-ordering of entries causes _unnecessary_ changes (changes in file content with no impact to runtime), and makes it harder to audit the "real" changes with certain tools such as `git diff`.

PR to follow.

ssalonen:
Resolves #2436.


bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/OrderingMapSerializer.java
```diff
@@ -0,0 +1,42 @@
...
+import com.google.gson.JsonSerializationContext;
+import com.google.gson.JsonSerializer;
+
+/**
+ * Serializes map data by ordering the keys
+ *
+ * @author Sami Salonen - Initial contribution
+ */
+@NonNullByDefault
+public class OrderingMapSerializer implements JsonSerializer<Map<String, Object>> {
```
ssalonen, line 32: GsonBuilder.registerTypeHierarchyAdapter(Map.class, new OrderingMapSerializer()) can result in OrderingMapSerializer to be used with non-Comparable keys and thus with runtime exception -- even though this class by itself looks OK (String keys are comparable)

bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/OrderingSetSerializer.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @author Sami Salonen - Initial contribution
+ */
+@NonNullByDefault
+public class OrderingSetSerializer implements JsonSerializer<Set<Object>> {
+
+    @Override
+    public JsonElement serialize(Set<Object> src, Type typeOfSrc, JsonSerializationContext context) {
+        JsonArray ordered = new JsonArray();
+        src.stream().map(context::serialize).sorted().forEachOrdered(ordered::add);
```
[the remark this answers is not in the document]
ssalonen: Resolved in https://github.com/openhab/openhab-core/pull/2437/commits/69f4c812c434e138dca67f0e2e0d7dbd21490886

bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/OrderingMapSerializer.java
ssalonen: Resolved in Resolved in https://github.com/openhab/openhab-core/pull/2437/commits/69f4c812c434e138dca67f0e2e0d7dbd21490886

kaikreuzer (approved):
Thanks @ssalonen, this looks good to me and the change makes sense in order to allow a proper diffing of changes in those files.

PR #2442, opened 2021-07-27
fix build

J-N-K:
Without this change current core can't be build on Windows.


kaikreuzer (approved):
Thanks!

PR #2448, opened 2021-08-03
[automation] Fix missing configuration in rules

### LINKED ISSUE #2447: [automation] Configuration discarded in ScriptedAutomationManager
In a `SimpleRule` we can use a `.setConfiguration` method to store some configuration data with a rule, however, when that rule is added via the `ScriptedAutomationManager` the configuration is not passed to the added `Rule` and can't be accessed in rule-based events (also the `RuleDTO` has a `configuration` field).

Is there a special reason for that or is it a bug?

Use case: I want to do some clean-up (cancelling `ScheduledFuture`s and so on) when the rule is removed. For that I need access to the original script (or at least parts of that). My idea was to store the needed information in the configuration, but unfortunately this is not possible.

J-N-K:
Fixes #2447 


kaikreuzer (approved):
Thanks!

PR #2451, opened 2021-08-04
[WIP][automation] expose three classes used as bindings in JSR-223 rules as interfaces

### LINKED ISSUE #2438: Some classes are internal but exposed "accidentally"
I'm currently working on a JSR223 compatible Java scripting addon.

The core exposes adds a binding `scriptExtension` or  `events` (and probably more that I didn't encounter yet) to the ScriptEngine. Unfortunately the classes for these bindings (`ScriptExtensionManagerWrapper` and `ScriptBusEvent`) are in internal bundles: `org.openhab.core.automation.module.script.internal` and `org.openhab.core.automation.module.script.internal.defaultscope` and are therefore not available in the classloader and the script loading fails (internal bundles are not exported).

Would it make sense to move these to public packages? My guess would be that this is not a breaking change since they can't be accessed at the moment anyway.

J-N-K:
Fixes #2438 


farion:
Would love to see this in main.

This is not only a problem for Java-Scripting as mentioned in #2438. This is also an Issue if you use Groovy with TypeChecked.

PR #2459, opened 2021-08-17
[automation] implement a listener for ScriptEngineFactory changes

### LINKED ISSUE #2458: [automation] Scripts should be unloaded when the ScriptEngineFactory becomes unavailable
Currently scripts are loaded on startup or when they are modified or newly created on runtime. If no corresponding `ScriptEngineFactory` is available they are queued and retried later when a `ScriptEngineFactory` is added. This works quite good. The other way is not working at all. Scripts stay in memory when the corresponding `ScriptEngineFactory` becomes unavailable. IMO they should be unloaded in that case.

The implementation could  be an extension to the `DependencyTracker` (which currently only tracks dependencies on libraries).

J-N-K:
Fixes #2458 

This is a different approach than the one proposed in the issue which implements a cleaner solution.


kaikreuzer:
PR build failed due to spotless checks.

J-N-K:
Oops. Already fixed that but forgot to push...

kaikreuzer (approved):
lgtm, thanks.

PR #2467, opened 2021-08-28
fix WatchQueueReader for multiple listeners

### LINKED ISSUE #2466: AbstractWatchService can't handle two listeners for same directory
The `AbstractWatchService` depends on a static instance of the `WatchQueueReader`. The `WatchQueueReader` stores the listeners in a `Map<WatchKey, AbstractWatchService>`. If a second `WatchService` registers the same directory, the first is no longer notified of changes because the value of the `Map` is silently overwritten.

This is even the case without an external `WatchService`: The `FolderObserver` registers `openhab/conf/`, later the `ScriptFileWatcher` registers `openhab/conf/automation/jsr223`. Changes to files in `openhab/conf/automation/jsr223` are not forwarded to the `FolderObserver`. I didn't check, but this is most likely not a problem due to dependencies between the `org.openhab.core.model.core` and `org.openhab.automation.module.script.rulesupport` bundles (so the correct registration order is always used).

A solution would be to change the map to `Map<WatchKey, Set<AbstractWatchService>>`. WDYT?

J-N-K:
Fixes #2466 

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/service/WatchQueueReader.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/service/WatchQueueReader.java
@@ -181,13 +181,10 @@ public class WatchQueueReader implements Runnable {
                     keysToRemove.add(key);
                 }
             }
-
-            Iterator<Notification> it = notifications.iterator();
-            while (it.hasNext()) {
-                Notification notification = it.next();
+            for (Notification notification : notifications) {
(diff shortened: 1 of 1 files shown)


J-N-K:
@openhab/core-maintainers any chance to get this reviewed and merged? Thanks,

kaikreuzer:
Why did you close it? Is it not relevant anymore?

J-N-K:
The issue is still present. But I implemented a workaround for my use-case.

Issue #2476, opened 2021-09-04
Feature: allow enabling authentication for all activities.


flo-mic:
Currently there are a lot activities on the API and WebUI allowed without authentication. There should be a possibility to force a user login before sending API commands or be able to access the WebUI. This feature should be disabled by default for the starting point but configurable from the WebUI so that a paranoid user can enable it if all his applications support the authentication or do not require the API top work.

This feature is especially important as soon you want to integrate a security system within openhab. It does not make sense to control a security system from openhab if I can change the item states without any authentication.

ghys:
This is what you're looking for:

![image](https://user-images.githubusercontent.com/2004147/132105060-b2169910-517b-4ad1-95ef-b72ba72f09e9.png)

You'll find it in Settings > API Security in the UI. It's an advanced option for now because it has the potential to break clients which don't expect authentication.

flo-mic:
ah perfect, I was not aware of this. Next time I will first check if there are any advanced settings. Thank you for the fast response.

PR #2477, opened 2021-09-04
Rework system:restauth config parameters

ghys:
Following the discussion in #2476, I believe the order and "advanced" status of config parameters in API Security (`system:restauth`) should be changed.
This promotes "Implicit User Role" to a non-advanced option (even if it can break some clients, it is clearly stated in the description), and demotes "Allow Basic Authentication" to an advanced option (API Tokens can be used to authenticate to the API as a better alternative to Basic).


kaikreuzer (approved):
Makes sense, thanks!

Issue #2392, opened 2021-05-30
Unchecked exceptions are swallowed by Executors


fwolter:
Unchecked exceptions raised in Executor tasks are not logged. The tasks are silently terminated and periodic tasks are cancelled completely. That's a "feature" of `FutureTask`, which needs to get `get()` called to retrieve the thrown exception. But many tasks don't have a result, so `get()` is never called and nobody notices if something went fundamentally wrong.

I spend quite a lot of time when tracking down the race condition #2389, whose error was hidden by above issue. The sympthom and the cause were in totally different components, which I'd never suspect to be connected.

Years ago I had an extensive error search in a project, too, which was also caused by above issue. I have found this, which I found again now: http://web.archive.org/web/20190307143204/http://code.nomad-labs.com:80/2011/12/09/mother-fk-the-scheduledexecutorservice
I think the language used is absolutely appropriate. They make a proposal how to fix it.

I'm wondering if you are aware of that behavior or isn't that that much critcal as I think and I'm too biased from above error search?  I find swallowing unchecked *and* uncought exceptions is a no go.

J-N-K:
That's a known issue and some bindings in addons take care of that (e.g. [`onewire`](https://github.com/openhab/openhab-addons/blob/6ae7f981344355632709d37b5b95f6aaec2891e5/bundles/org.openhab.binding.onewire/src/main/java/org/openhab/binding/onewire/internal/handler/OwserverBridgeHandler.java#L135-L202)).  

A general solution would be great of course, because it would reduce code on the bindings.

lolodomo:
Very well known, that is why in bindings we generally catch `Exception` in periodic threads.

kaikreuzer:
I remember that we tried all kind of things to improve that situation a while ago, but I don't remember the exact results.
Not sure whether we also tried extending the executer class. If that suggested solution is working without any drawbacks, I guess it would make sense to implement it in the core and provide it through our scheduled thread pool service.

Hilbrand:
For bindngs we wrapped the scheduler to catch uncatxhed exceptions:: https://github.com/eclipse-archived/smarthome/pull/4772

fwolter:
That's a very neat solution. Simpler than the proposed in above link. Is there any reason not using it everywere?

J-N-K:
Except from one test all usages of `ScheduledThreadPoolExecutor` now use the `WrappedScheduledExecutorService`, so this issue is solved.

Issue #2389, opened 2021-05-29
[monitor] Micrometer raises unchecked exception during OH start


fwolter:
This exception causes the task in `StartLevelService` to terminate when processing the callbacks of startlevel 40. The exception is swallowed. I added a temporary catch clause to log it. The only visible sympthom I noticed is that the rule engine is not started sometimes.
```
23:29:07.993 [DEBUG] [.internal.DefaultMetricsRegistration] - Registering meters...
23:29:08.010 [DEBUG] [o.monitor.internal.metrics.JVMMetric] - JVMMetric is being bound...
23:29:28.660 [WARN ] [enhab.core.service.StartLevelService] - 
java.lang.NoClassDefFoundError: javax/management/ListenerNotFoundException
	at org.openhab.core.io.monitor.internal.metrics.JVMMetric.bindTo(JVMMetric.java:57) ~[?:?]
	at org.openhab.core.io.monitor.internal.DefaultMetricsRegistration.lambda$0(DefaultMetricsRegistration.java:96) ~[?:?]
	at java.lang.Iterable.forEach(Iterable.java:75) ~[?:?]
	at org.openhab.core.io.monitor.internal.DefaultMetricsRegistration.registerMeters(DefaultMetricsRegistration.java:96) ~[?:?]
	at org.openhab.core.io.monitor.internal.DefaultMetricsRegistration.onReadyMarkerAdded(DefaultMetricsRegistration.java:105) ~[?:?]
	at org.openhab.core.internal.service.ReadyServiceImpl.lambda$0(ReadyServiceImpl.java:52) ~[?:?]
	at java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183) ~[?:?]
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195) ~[?:?]
	at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:177) ~[?:?]
	at java.util.HashMap$EntrySpliterator.forEachRemaining(HashMap.java:1746) ~[?:?]
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484) ~[?:?]
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474) ~[?:?]
	at java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150) ~[?:?]
[...]

PR #2481, opened 2021-09-09
[startlevel] Use wrapped scheduler to make swallowed exceptions visible

fwolter:
This makes unchecked and uncought exception visible, thrown in start level triggers. Tested with the exception from #2389.

Discussed in #2392 


wborn (approved):
Thanks!

PR #2482, opened 2021-09-10
Upgrade XStream to 1.4.18

wborn:
For release notes, see:

https://x-stream.github.io/changes.html#1.4.18

---

There were already PRs some time ago to properly initialize the XStream security framework in all code that uses XStream.


kaikreuzer (approved):
Thanks!

PR #2488, opened 2021-09-15
[Marketplace] Add option to include unpublished entries

ghys:
Add a config parameter to bypass the filtering of Dicourse topics
to include those which don't have the "published" tag.

Set the API key parameter as advanced (irrelevant to most users).


kaikreuzer:
What exactly is the use case to show unpublished stuff? I thought the whole idea is to only show entries with the published tag as all the rest is not (yet) meant to be shown in the marketplace?

ghys:
Yes, the default is to not show them. This option is mostly for the authors themselves and moderators so they can preview draft entries in the marketplace before regular users see them, and see it they work - hence the strongish language in the description ("harm your installation", "at your own risk" etc.) but it's true, there's still a risk people who have no business enabling this option will still try it.

kaikreuzer:
Ok, let's see if this causes issues as too many people activate it and then complain...

kaikreuzer (approved):

PR #2490, opened 2021-09-18
Extend marketplace to accept kar and some improvements

J-N-K:
Following up a discussion with @ghys this adds another `MarketplaceAddonHandler` for installing .kar files.

### DIFF
--- a/bundles/org.openhab.core.addon.marketplace.karaf/src/main/java/org/openhab/core/addon/marketplace/karaf/internal/community/CommunityKarafAddonHandler.java
+++ b/bundles/org.openhab.core.addon.marketplace.karaf/src/main/java/org/openhab/core/addon/marketplace/karaf/internal/community/CommunityKarafAddonHandler.java
@@ -18,19 +18,15 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.MalformedURLException;
-import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
-import java.util.Arrays;
 import java.util.List;
-import java.util.Objects;
 import java.util.stream.Collectors;
 
-import org.apache.karaf.features.FeaturesService;
-import org.apache.karaf.features.Repository;
+import org.apache.karaf.kar.KarService;
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.openhab.core.OpenHAB;
 import org.openhab.core.addon.Addon;
@@ -61,11 +57,11 @@ public class CommunityKarafAddonHandler implements MarketplaceAddonHandler {
 
     private final Logger logger = LoggerFactory.getLogger(CommunityKarafAddonHandler.class);
 
-    private final FeaturesService featuresService;
+    private final KarService karService;
 
     @Activate
-    public CommunityKarafAddonHandler(@Reference FeaturesService featuresService) {
-        this.featuresService = featuresService;
+    public CommunityKarafAddonHandler(@Reference KarService karService) {
+        this.karService = karService;
         ensureCachedBundlesAreInstalled();
     }
 
@@ -79,11 +75,10 @@ public class CommunityKarafAddonHandler implements MarketplaceAddonHandler {
     public boolean isInstalled(String addonId) {
         try {
             Path addonDirectory = getAddonCacheDirectory(addonId);
-            List<URI> repositories = Arrays.stream(featuresService.listRepositories()).map(Repository::getURI)
-                    .collect(Collectors.toList());
+            List<String> repositories = karService.list();
             if (Files.isDirectory(addonDirectory)) {
-                return Files.list(addonDirectory).filter(path -> path.endsWith(".kar")).findFirst().map(Path::toUri)
-                        .map(repositories::contains).orElse(false);
+                return Files.list(addonDirectory).filter(path -> path.endsWith(".kar")).findFirst()
+                        .map(Path::getFileName).map(Path::toString).map(repositories::contains).orElse(false);
             }
         } catch (Exception e) {
             logger.warn("Failed to determine installation status for {}: ", addonId, e);
@@ -108,14 +103,13 @@ public class CommunityKarafAddonHandler implements MarketplaceAddonHandler {
     public void uninstall(Addon addon) throws MarketplaceHandlerException {
         try {
             Path addonPath = getAddonCacheDirectory(addon.getId());
-            List<URI> repositories = Arrays
-                    .stream(Objects.requireNonNullElse(featuresService.listRepositories(), new Repository[] {}))
-                    .map(Repository::getURI).collect(Collectors.toList());
+            List<String> repositories = karService.list();
             if (Files.isDirectory(addonPath)) {
-                List<Path> repositoryFiles = Files.list(addonPath).filter(path -> repositories.contains(path.toUri()))
+                List<Path> repositoryFiles = Files.list(addonPath)
+                        .filter(path -> repositories.contains(path.getFileName().toString()))
                         .collect(Collectors.toList());
                 for (Path file : repositoryFiles) {
-                    featuresService.removeRepository(file.toUri(), true);
+                    karService.uninstall(file.getFileName().toString());
                     Files.delete(file);
                 }
             }
@@ -154,7 +148,7 @@ public class CommunityKarafAddonHandler implements MarketplaceAddonHandler {
                             "The local cache folder doesn't contain a single file: " + addonPath, null);
                 }
                 try {
-                    featuresService.addRepository(bundleFiles.get(0).toUri(), true);
+                    karService.install(bundleFiles.get(0).toUri(), true);
                 } catch (Exception e) {
                     throw new MarketplaceHandlerException(
                             "Cannot install bundle from marketplace cache: " + e.getMessage(), e);
--- a/bundles/org.openhab.core.addon.marketplace.karaf/src/main/java/org/openhab/core/addon/marketplace/karaf/internal/community/CommunityKarafAddonHandler.java
+++ b/bundles/org.openhab.core.addon.marketplace.karaf/src/main/java/org/openhab/core/addon/marketplace/karaf/internal/community/CommunityKarafAddonHandler.java
@@ -177,7 +177,7 @@ public class CommunityKarafAddonHandler implements MarketplaceAddonHandler {
                         });
             }
         } catch (IOException e) {
-            logger.warn("Failed to re-install bundled: {}", e.getMessage());
+            logger.warn("Failed to re-install bundles: {}", e.getMessage());
         }
     }
 
--- a/bundles/org.openhab.core.addon.marketplace.karaf/src/main/java/org/openhab/core/addon/marketplace/karaf/internal/community/CommunityKarafAddonHandler.java
+++ b/bundles/org.openhab.core.addon.marketplace.karaf/src/main/java/org/openhab/core/addon/marketplace/karaf/internal/community/CommunityKarafAddonHandler.java
@@ -25,6 +25,7 @@ import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
 import java.util.List;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.apache.karaf.kar.KarService;
 import org.eclipse.jdt.annotation.NonNullByDefault;
@@ -54,6 +55,7 @@ public class CommunityKarafAddonHandler implements MarketplaceAddonHandler {
     private static final List<String> SUPPORTED_EXT_TYPES = List.of("automation", "binding", "io", "persistence",
             "transformation", "ui", "voice");
     private static final String KAR_DOWNLOAD_URL_PROPERTY = "kar_download_url";
+    private static final String KAR_EXTENSION = ".kar";
 
     private final Logger logger = LoggerFactory.getLogger(CommunityKarafAddonHandler.class);
 
@@ -70,16 +72,24 @@ public class CommunityKarafAddonHandler implements MarketplaceAddonHandler {
         return SUPPORTED_EXT_TYPES.contains(type) && KAR_CONTENT_TYPE.equals(contentType);
     }
 
+    private Stream<Path> karFilesStream(Path addonDirectory) throws IOException {
+        return Files.isDirectory(addonDirectory) ? Files.list(addonDirectory).map(Path::getFileName)
+                .filter(path -> path.toString().endsWith(KAR_EXTENSION)) : Stream.empty();
+    }
+
+    private String pathToKarRepoName(Path path) {
+        String fileName = path.getFileName().toString();
+        return fileName.substring(0, fileName.length() - KAR_EXTENSION.length());
+    }
+
     @Override
     @SuppressWarnings("null")
     public boolean isInstalled(String addonId) {
         try {
             Path addonDirectory = getAddonCacheDirectory(addonId);
             List<String> repositories = karService.list();
-            if (Files.isDirectory(addonDirectory)) {
-                return Files.list(addonDirectory).filter(path -> path.endsWith(".kar")).findFirst()
-                        .map(Path::getFileName).map(Path::toString).map(repositories::contains).orElse(false);
-            }
+            return karFilesStream(addonDirectory).findFirst().map(this::pathToKarRepoName).map(repositories::contains)
+                    .orElse(false);
         } catch (Exception e) {
             logger.warn("Failed to determine installation status for {}: ", addonId, e);
         }
@@ -99,19 +109,16 @@ public class CommunityKarafAddonHandler implements MarketplaceAddonHandler {
     }
 
     @Override
-    @SuppressWarnings("null")
     public void uninstall(Addon addon) throws MarketplaceHandlerException {
         try {
             Path addonPath = getAddonCacheDirectory(addon.getId());
             List<String> repositories = karService.list();
-            if (Files.isDirectory(addonPath)) {
-                List<Path> repositoryFiles = Files.list(addonPath)
-                        .filter(path -> repositories.contains(path.getFileName().toString()))
-                        .collect(Collectors.toList());
-                for (Path file : repositoryFiles) {
-                    karService.uninstall(file.getFileName().toString());
-                    Files.delete(file);
+            for (Path path : karFilesStream(addonPath).collect(Collectors.toList())) {
+                String karRepoName = pathToKarRepoName(path);
+                if (repositories.contains(karRepoName)) {
+                    karService.uninstall(karRepoName);
                 }
+                Files.delete(addonPath.resolve(path));
             }
             Files.delete(addonPath);
         } catch (Exception e) {
@@ -148,7 +155,7 @@ public class CommunityKarafAddonHandler implements MarketplaceAddonHandler {
                             "The local cache folder doesn't contain a single file: " + addonPath, null);
                 }
                 try {
-                    karService.install(bundleFiles.get(0).toUri(), true);
+                    karService.install(bundleFiles.get(0).toUri(), false);
                 } catch (Exception e) {
                     throw new MarketplaceHandlerException(
                             "Cannot install bundle from marketplace cache: " + e.getMessage(), e);
--- a/bundles/org.openhab.core.addon.marketplace.karaf/src/main/java/org/openhab/core/addon/marketplace/karaf/internal/community/CommunityKarafAddonHandler.java
+++ b/bundles/org.openhab.core.addon.marketplace.karaf/src/main/java/org/openhab/core/addon/marketplace/karaf/internal/community/CommunityKarafAddonHandler.java
@@ -53,8 +53,7 @@ import org.slf4j.LoggerFactory;
 @NonNullByDefault
 public class CommunityKarafAddonHandler implements MarketplaceAddonHandler {
     private static final String KAR_CONTENT_TYPE = "application/vnd.openhab.feature;type=karfile";
-    private static final String BUNDLE_CACHE_PATH = OpenHAB.getUserDataFolder() + File.separator + "marketplace"
-            + File.separator + "kar";
+    private static final Path BUNDLE_CACHE_PATH = Path.of(OpenHAB.getUserDataFolder(), "marketplace", "kar");
     private static final List<String> SUPPORTED_EXT_TYPES = List.of("binding");
     private static final String KAR_DOWNLOAD_URL_PROPERTY = "kar_download_url";
 
@@ -70,7 +69,6 @@ public class CommunityKarafAddonHandler implements MarketplaceAddonHandler {
 
     @Override
     public boolean supports(String type, String contentType) {
-        // we support only certain extension types, and only as pure OSGi bundles
         return SUPPORTED_EXT_TYPES.contains(type) && KAR_CONTENT_TYPE.equals(contentType);
     }
 
@@ -78,12 +76,12 @@ public class CommunityKarafAddonHandler implements MarketplaceAddonHandler {
     @SuppressWarnings("null")
     public boolean isInstalled(String addonId) {
         try {
-            File addonDirectory = getAddonCacheDirectory(addonId);
+            Path addonDirectory = getAddonCacheDirectory(addonId);
             List<URI> repositories = Arrays.stream(featuresService.listRepositories()).map(Repository::getURI)
                     .collect(Collectors.toList());
-            if (addonDirectory.exists() && addonDirectory.isDirectory()) {
-                return Files.list(addonDirectory.toPath()).filter(path -> path.endsWith(".kar")).findFirst()
-                        .map(Path::toUri).map(repositories::contains).orElse(false);
+            if (Files.isDirectory(addonDirectory)) {
+                return Files.list(addonDirectory).filter(path -> path.endsWith(".kar")).findFirst().map(Path::toUri)
+                        .map(repositories::contains).orElse(false);
             }
         } catch (Exception e) {
             logger.warn("Failed to determine installation status for {}: ", addonId, e);
@@ -94,13 +92,12 @@ public class CommunityKarafAddonHandler implements MarketplaceAddonHandler {
 
     @Override
     public void install(Addon addon) throws MarketplaceHandlerException {
-        URL sourceUrl;
         try {
-            sourceUrl = new URL((String) addon.getProperties().get(KAR_DOWNLOAD_URL_PROPERTY));
+            URL sourceUrl = new URL((String) addon.getProperties().get(KAR_DOWNLOAD_URL_PROPERTY));
             addBundleToCache(addon.getId(), sourceUrl);
             installFromCache(addon.getId());
         } catch (MalformedURLException e) {
-            throw new MarketplaceHandlerException("Malformed source URL: " + e.getMessage());
+            throw new MarketplaceHandlerException("Malformed source URL: " + e.getMessage(), e);
         }
     }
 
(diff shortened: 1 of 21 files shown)


wborn:
Yes it would be nice to support KAR files as well. One of the reasons the ESH Marketplace probably didn't support it, could be that this is Karaf only. That's why many Karaf specific functionality is moved into separate bundles to keep it optional when using OHC. That way you can still easily use Felix instead of Karaf as OSGi runtime. Though I don't know how common it is to use OHC with another runtime these days. :slightly_smiling_face:

J-N-K:
Well, it would be easy to move the class to another bundle if you prefer that. Which one would fit best?

J-N-K:
@wborn Do you know what happens if two feature repositories provide features with the same name (e.g. a .kar has an updated version of a binding that is available in distribution)?

kaikreuzer:
I think it is ok to leave the class (and thus the Karaf dependency) in here. If someone builds their own distro without Karaf, he is likely to also not include the marketplace, which is hosted on our Discourse.

> Do you know what happens if two feature repositories provide features with the same name 

My guess is that both are available and upon an install (without specifying a specific version) the one with the highest version is picked.

J-N-K:
Ok, we can always make an administrative rule that the marketplace-version should include a qualifier to avoid duplicate version-numbers. 

This work is purely "theoretical" as I can't post to the marketplace category on the forum and therefore not test it. When will that be opened?

wborn:
> I think it is ok to leave the class (and thus the Karaf dependency) in here. If someone builds their own distro without Karaf, he is likely to also not include the marketplace, which is hosted on our Discourse.

That distro is what you have when debugging openHAB in Eclipse. :wink: With this dependency you cannot run or debug the Marketplace with Eclipse because there are no Karaf bundles in that runtime. You'd probably still be able to run and debug it when the Karaf code is moved into a separate bundle like `org.openhab.core.addon.marketplace.karaf`.

kaikreuzer:
@wborn Ok, I didn't think about that - this is then indeed be a strong reason to move the class to a dedicated bundle.

wborn:
Good good :+1: I think you are now working on to also add the new bundle to the BOM and Karaf feature file? :slightly_smiling_face:

J-N-K:
Sure 😀

bundles/org.openhab.core.addon.marketplace.karaf/src/main/java/org/openhab/core/addon/marketplace/karaf/internal/community/CommunityKarafAddonHandler.java
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        return false;
+    }
+
+    @Override
+    public void install(Addon addon) throws MarketplaceHandlerException {
+        URL sourceUrl;
+        try {
+            sourceUrl = new URL((String) addon.getProperties().get(KAR_DOWNLOAD_URL_PROPERTY));
```
wborn, line 99:
```suggestion
        try {
            URL sourceUrl = new URL((String) addon.getProperties().get(KAR_DOWNLOAD_URL_PROPERTY));
```

bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityMarketplaceAddonService.java
```diff
@@ -95,6 +95,7 @@
 
     private HashMap<String, String> contentTypes = new HashMap<String, String>(3);
```
wborn, line 96:
```suggestion
    private Map<String, String> contentTypes = new HashMap<>(3);
```

bundles/org.openhab.core.addon.marketplace.karaf/src/main/java/org/openhab/core/addon/marketplace/karaf/internal/community/CommunityKarafAddonHandler.java
```diff
@@ -0,0 +1,189 @@
...
+                for (File file : Objects.requireNonNullElse(addonPath.listFiles(), new File[] {})) {
+                    if (repositories.contains(file.toURI())) {
+                        featuresService.removeRepository(file.toURI(), true);
+                    }
+                    file.delete();
+                }
+            }
+            addonPath.delete();
+        } catch (Exception e) {
+            throw new MarketplaceHandlerException("Failed uninstalling bundle: " + e.getMessage());
```
wborn, line 125: Is it possible to add the stacktrace to all the `MarketplaceHandlerException`s? It would simplify debugging.

bundles/org.openhab.core.addon.marketplace.karaf/src/main/java/org/openhab/core/addon/marketplace/karaf/internal/community/CommunityKarafAddonHandler.java
```diff
@@ -0,0 +1,189 @@
...
+ * @author Kai Kreuzer - Initial contribution and API
+ * @author Yannick Schaus - refactoring
+ * @author Jan N. Klug - refactor to support kar files
+ *
+ */
+@Component(immediate = true)
+@NonNullByDefault
+public class CommunityKarafAddonHandler implements MarketplaceAddonHandler {
+    private static final String KAR_CONTENT_TYPE = "application/vnd.openhab.feature;type=karfile";
+    private static final String BUNDLE_CACHE_PATH = OpenHAB.getUserDataFolder() + File.separator + "marketplace"
```
wborn, line 56: Maybe it is possible to replace these strings with Java NIO `Path`? It's nicer to construct them with `Path.of(..)` and you would not have to convert between `File` and `Path` when using `Files`.

bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityMarketplaceAddonService.java
```diff
@@ -161,7 +162,7 @@ public void refreshSource() {
     @Override
     public List<Addon> getAddons(Locale locale) {
         try {
-            List<DiscourseCategoryResponse> pages = new ArrayList<DiscourseCategoryResponse>();
+            List<DiscourseCategoryResponseDTO> pages = new ArrayList<DiscourseCategoryResponseDTO>();
```
wborn, line 165:
```suggestion
            List<DiscourseCategoryResponseDTO> pages = new ArrayList<>();
```

bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityMarketplaceAddonService.java
```diff
@@ -280,6 +281,7 @@ private Addon convertTopicItemToAddon(DiscourseTopicItem topic, List<DiscourseUs
         String id = ADDON_ID_PREFIX + topic.id.toString();
         AddonType addonType = types.get(topic.category_id);
         String type = (addonType != null) ? addonType.getId() : "";
+        // TODO: content-type differentiation for kar and jar
```
wborn, line 284: Is there still something to be done here?

bundles/org.openhab.core.addon.marketplace.karaf/src/main/java/org/openhab/core/addon/marketplace/karaf/internal/community/CommunityKarafAddonHandler.java
```diff
@@ -0,0 +1,189 @@
...
+
+    @Activate
+    public CommunityKarafAddonHandler(@Reference FeaturesService featuresService) {
+        this.featuresService = featuresService;
+        ensureCachedBundlesAreInstalled();
+    }
+
+    @Override
+    public boolean supports(String type, String contentType) {
+        // we support only certain extension types, and only as pure OSGi bundles
```
wborn, line 73: That's what the `CommunityBundleAddonHandler` does not this KAR handler.

bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityMarketplaceAddonService.java
```diff
@@ -191,7 +192,7 @@ public void refreshSource() {
                     .map(t -> convertTopicItemToAddon(t, users)).collect(Collectors.toList());
         } catch (Exception e) {
             logger.error("Unable to retrieve marketplace add-ons", e);
-            return new ArrayList<Addon>();
+            return new ArrayList<>();
```
wborn, line 195:
```suggestion
            return List.of();
```

Less code and it returns a constant which reduces memory consumption and garbage collection.

wborn (changes requested):
Thanks for working on this! I have few comments/improvements below:

bundles/org.openhab.core.addon.marketplace.karaf/src/main/java/org/openhab/core/addon/marketplace/karaf/internal/community/CommunityKarafAddonHandler.java
J-N-K: Fixed. Also changed in the handler where it was copied from.

bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityMarketplaceAddonService.java
J-N-K: I also changed that to use `Map.of`.

bundles/org.openhab.core.addon.marketplace.karaf/src/main/java/org/openhab/core/addon/marketplace/karaf/internal/community/CommunityKarafAddonHandler.java
J-N-K: Agreed, I'll extend the `MarketplaceHandlerException` with a `Throwable cause`.

bundles/org.openhab.core.addon.marketplace.karaf/src/main/java/org/openhab/core/addon/marketplace/karaf/internal/community/CommunityKarafAddonHandler.java
J-N-K: Done.

bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityMarketplaceAddonService.java
J-N-K: I didn't change these because there are a lot of similar issues in the code. But I'll fix these two.

bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityMarketplaceAddonService.java
J-N-K: Yes, that's the one open question. Currently we distinguish between kar and jar by looking at the extension of the filename from the download link. Unfortunately that is not available here, , because it is not present in the `DiscourseTopicItem`. This method is used in `getAddons` which lists all addons. In `getAddon`, which only gets the addon for a single addon id, we have a `DiscourseTopicResponseDTO` which includes that field.

bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityMarketplaceAddonService.java
J-N-K: Done.

J-N-K:
Would you like to the the other bundle-addon-handler also refactored to the new `Path/Files`-API?

wborn:
Yes that would be nice too. :+1:

bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityMarketplaceAddonService.java
wborn: Yes that's even better!

bundles/org.openhab.core.addon.marketplace.karaf/src/main/java/org/openhab/core/addon/marketplace/karaf/internal/community/CommunityKarafAddonHandler.java
wborn: It certainly looks a lot nicer now. :slightly_smiling_face: :+1:

bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityMarketplaceAddonService.java
wborn: So this is still TBD, see also https://github.com/openhab/openhab-core/pull/2405#issuecomment-922330139

ghys: I have renamed the Bindings category to Bundles and added the following tag groups:

image

⬆️ it would be nice to have this added as a "maturity" property if one of these are found. This could come in handy later.

image

Also the category is configured like this:

image

wborn: Thanks @ghys, that should work and give us everything we want for the forseeable future. :slightly_smiling_face::crossed_fingers:

J-N-K: I'll implement that

bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/MarketplaceBundleInstaller.java
```diff
@@ -142,26 +146,25 @@ protected void uninstallBundle(BundleContext bundleContext, String addonId) thro
...
+                Files.list(BUNDLE_CACHE_PATH).filter(Files::isDirectory).map(p -> "marketplace:" + p.getFileName())
+                        .filter(addonId -> !isBundleInstalled(bundleContext, addonId)).forEach(addonId -> {
+                            logger.info("Reinstalling missing marketplace bundle: {}", addonId);
+                            try {
+                                installFromCache(bundleContext, addonId);
+                            } catch (MarketplaceHandlerException e) {
+                                logger.warn("Failed reinstalling add-on from cache", e);
+                            }
+                        });
+                // TODO: why is that: bundleFile.delete();
```
J-N-K, line 160: @ghys Can you comment here? Why is the file deleted from the cache after it has been installed?

ghys: It's the directory with the topic ID, and since it's not empty, nothing would happen. I think it's an unfortunate leftover, so the best is to remove it.

J-N-K:
I have implemented what I think should work. The last commit only contains those changes. I'm not sure what to do with the maturity tag, it's included in the tags-property but there is no field in `Addon` that could be used.

bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityMarketplaceAddonService.java
```diff
@@ -86,17 +90,20 @@
     private static final String YAML_CODE_MARKUP_START = "<pre><code class=\"lang-yaml\">";
     private static final String CODE_MARKUP_END = "</code></pre>";
 
-    private HashMap<Integer, AddonType> types = new HashMap<Integer, AddonType>(3);
-    private static final Integer BINDINGS_CATEGORY = 73;
+    private static final Integer ADDONS_CATEGORY = 73;
```
wborn, line 93: Rule templates and UI Widgets are also add-ons, these are bundle add-ons.

```suggestion
    private static final Integer BUNDLES_CATEGORY = 73;
```

J-N-K: done

wborn:
> there is no field in `Addon` that could be used

We can add another field, like all the other fields that were added in #2405, so the info can be shown in the UI and used for filtering.

J-N-K:
I added the field for maturity and license (also that is currently always null). Looking at `Addon`, I wonder if it really makes sense to do add more fields each time (which will always be breaking) or if we should change to a builder.

J-N-K:
re-triggering build

wborn:
I just tested this PR and got my Marketplace entry succesfully recognized as a KAR:

kar

However it did not install properly. The only logging that showed was:

`[Fatal Error] :1:1: Content is not allowed in prolog.`

So I added a try/catch to get more detailed logging and this what showed up:

```
[ERROR] [.community.CommunityKarafAddonHandler] - Error installing KAR add-on
org.openhab.core.addon.marketplace.MarketplaceHandlerException: Cannot install bundle from marketplace cache: Unable to load file:///home/wouter/software/openhab/instance/userdata/marketplace/kar/126674/org.openhab.binding.systeminfo-3.2.0-SNAPSHOT.kar : file:///home/wouter/software/openhab/instance/userdata/marketplace/kar/126674/org.openhab.binding.systeminfo-3.2.0-SNAPSHOT.kar
	at org.openhab.core.addon.marketplace.karaf.internal.community.CommunityKarafAddonHandler.installFromCache(CommunityKarafAddonHandler.java:166) ~[?:?]
	at org.openhab.core.addon.marketplace.karaf.internal.community.CommunityKarafAddonHandler.install(CommunityKarafAddonHandler.java:103) ~[?:?]
	at org.openhab.core.addon.marketplace.internal.community.CommunityMarketplaceAddonService.install(CommunityMarketplaceAddonService.java:233) ~[?:?]
	at org.openhab.core.io.rest.core.internal.addons.AddonResource.lambda$1(AddonResource.java:223) ~[?:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515) [?:?]
Caused by: java.lang.RuntimeException: Unable to load file:///home/wouter/software/openhab/instance/userdata/marketplace/kar/126674/org.openhab.binding.systeminfo-3.2.0-SNAPSHOT.kar : file:///home/wouter/software/openhab/instance/userdata/marketplace/kar/126674/org.openhab.binding.systeminfo-3.2.0-SNAPSHOT.kar
	at org.apache.karaf.features.internal.service.RepositoryImpl.load(RepositoryImpl.java:121) ~[?:?]
	at org.apache.karaf.features.internal.service.RepositoryImpl.<init>(RepositoryImpl.java:51) ~[?:?]
	at org.apache.karaf.features.internal.service.RepositoryCacheImpl.create(RepositoryCacheImpl.java:51) ~[?:?]
	at org.apache.karaf.features.internal.service.FeaturesServiceImpl.addRepository(FeaturesServiceImpl.java:386) ~[?:?]
	at org.openhab.core.addon.marketplace.karaf.internal.community.CommunityKarafAddonHandler.installFromCache(CommunityKarafAddonHandler.java:164) ~[?:?]
	... 8 more
Caused by: java.lang.RuntimeException: Unable to load file:///home/wouter/software/openhab/instance/userdata/marketplace/kar/126674/org.openhab.binding.systeminfo-3.2.0-SNAPSHOT.kar
	at org.apache.karaf.features.internal.model.JaxbUtil.unmarshalValidate(JaxbUtil.java:139) ~[?:?]
	at org.apache.karaf.features.internal.model.JaxbUtil.unmarshal(JaxbUtil.java:100) ~[?:?]
	at org.apache.karaf.features.internal.service.RepositoryImpl.load(RepositoryImpl.java:118) ~[?:?]
	at org.apache.karaf.features.internal.service.RepositoryImpl.<init>(RepositoryImpl.java:51) ~[?:?]
	at org.apache.karaf.features.internal.service.RepositoryCacheImpl.create(RepositoryCacheImpl.java:51) ~[?:?]
	at org.apache.karaf.features.internal.service.FeaturesServiceImpl.addRepository(FeaturesServiceImpl.java:386) ~[?:?]
	at org.openhab.core.addon.marketplace.karaf.internal.community.CommunityKarafAddonHandler.installFromCache(CommunityKarafAddonHandler.java:164) ~[?:?]
	... 8 more
Caused by: org.xml.sax.SAXParseException: Content is not allowed in prolog.
	at com.sun.org.apache.xerces.internal.parsers.DOMParser.parse(DOMParser.java:261) ~[?:?]
	at com.sun.org.apache.xerces.internal.jaxp.DocumentBuilderImpl.parse(DocumentBuilderImpl.java:339) ~[?:?]
	at javax.xml.parsers.DocumentBuilder.parse(DocumentBuilder.java:122) ~[?:?]
	at org.apache.karaf.util.XmlUtils.parse(XmlUtils.java:65) ~[?:?]
	at org.apache.karaf.features.internal.model.JaxbUtil.unmarshalValidate(JaxbUtil.java:112) ~[?:?]
	at org.apache.karaf.features.internal.model.JaxbUtil.unmarshal(JaxbUtil.java:100) ~[?:?]
	at org.apache.karaf.features.internal.service.RepositoryImpl.load(RepositoryImpl.java:118) ~[?:?]
	at org.apache.karaf.features.internal.service.RepositoryImpl.<init>(RepositoryImpl.java:51) ~[?:?]
	at org.apache.karaf.features.internal.service.RepositoryCacheImpl.create(RepositoryCacheImpl.java:51) ~[?:?]
	at org.apache.karaf.features.internal.service.FeaturesServiceImpl.addRepository(FeaturesServiceImpl.java:386) ~[?:?]
	at org.openhab.core.addon.marketplace.karaf.internal.community.CommunityKarafAddonHandler.installFromCache(CommunityKarafAddonHandler.java:164) ~[?:?]
	... 8 more
```

Most likely KAR files need to be installed using the [KarService](https://github.com/apache/karaf/blob/main/kar/src/main/java/org/apache/karaf/kar/KarService.java) and `FeaturesServiceImpl.addRepository` expects URLs to unpacked repositories?

I tested using this [org.openhab.binding.systeminfo-3.2.0-SNAPSHOT.kar](https://github.com/wborn/openhab-addons/releases/download/kar-test/org.openhab.binding.systeminfo-3.2.0-SNAPSHOT.kar). I also checked the file in /userdata/marketplace and the content was as expected. This KAR file works if copy it into my `/addons` dir.

---

Some other things I noticed:

There seems to be a limit of 3 tags on posts:

limit

Also, after installing a KAR file, most likely the UI will then list the add-ons of the KAR as any other "openHAB Distribution" add-on? That can be confusing.

J-N-K:
@wborn Correct, I didn't think about that (I used the `FeaturesService` to install a feature.xml from a maven repository). `KarService` looks indeed like the correct one. I changed that, but as I said before: I can't test anything because I don't have access to the marketplace.

bundles/org.openhab.core.addon.marketplace.karaf/src/main/java/org/openhab/core/addon/marketplace/karaf/internal/community/CommunityKarafAddonHandler.java
```diff
@@ -148,7 +155,7 @@ private void installFromCache(String addonId) throws MarketplaceHandlerException
                             "The local cache folder doesn't contain a single file: " + addonPath, null);
                 }
                 try {
-                    karService.install(bundleFiles.get(0).toUri(), true);
+                    karService.install(bundleFiles.get(0).toUri(), false);
```
J-N-K, line 158: From my understanding `true` installs the contained features. That’s what I would expect after installing an addon from the marketplace. Am I wrong here?

wborn: It is actually the otherway around which is confusing :-) 

```java
    /**
     * Install KAR from a given URI
     *
     * @param karUri URI of the kar to install
     * @param noAutoStartBundles true to not start automatically the bundles, false else
     * @throws Exception in case of installation failure
     */
    void install(URI karUri, boolean noAutoStartBundles) throws Exception;
```

wborn:
Using the KarService it works a lot better. :+1:

Though there were a few more issues that I've fixed in my commit. I think it is better to automatically start the bundles or they won't work. That way you can also use features names that do not start with openhab-binding etc. so they will not show up like "openHAB Distribution" add-ons, which prevents:

> Also, after installing a KAR file, most likely the UI will then list the add-ons of the KAR as any other "openHAB Distribution" add-on? That can be confusing.

I also had to manually remove lastnpe and slf4j from my KAR file and features definition. Otherwise it would not properly install, so we probably need to exclude those from KAR generation.

Now it seems to work well and I think it is mergable. :smile: 

> WDYT about the builder for Addon? Perhaps it would be better to do that in another PR, also it would simplify the code here and in the existing AddonService

Yes it is not convenient and ugle to haveso many constructor arguments. I'm also fine  to do that in a follow up PR.

features/karaf/openhab-core/src/main/feature/feature.xml
```diff
@@ -66,7 +66,9 @@
 
 	<feature name="openhab-core-addon-marketplace" version="${project.version}">
 		<feature>openhab-core-base</feature>
+		<bundle dependency="true">mvn:org.apache.karaf.kar/org.apache.karaf.kar.core/${karaf.tooling.version}</bundle>
```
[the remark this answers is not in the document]
J-N-K: I didn‘t find the correct feature 😀. Maybe you can point me there.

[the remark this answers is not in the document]
wborn: I see it was added to make the feature verification happy. So I've added the "kar" feature just like how some other OH features depend on Karaf features.

wborn (approved):
I think the code is ready to be merged now. :+1: Can you also review this @openhab/add-ons-maintainers? It also contains a few of my commits.

fwolter (approved):
The PR contains `.classpath` and `.project` files. Otherwise LGTM.

wborn:
> The PR contains .classpath and .project files.

Yes the openhab-core repo still contains these because IIRC there are still some issues to import projects (without compilation errors) in Eclipse using only POMs.

J-N-K:
It would be great if we can enable the category in Discourse then.

ghys:
There seems to be a limit of 3 tags on posts

image

the brown color & reset button means it was changed from the default (5). I don't know why that is, so I reverted to 5.

> Also, after installing a KAR file, most likely the UI will then list the add-ons of the KAR as any other "openHAB Distribution" add-on? That can be confusing.

What do you mean, I didn't understand?

ghys:
>  I can't test anything because I don't have access to the marketplace.

I created a special group of "marketplace curators" and put you in it, so you should have access. Currently only maintainers and forum moderators can create posts in the marketplace category until we unveil it to the public.

kaikreuzer (approved):
lgtm, many thanks! Great to have KAR files supported now.

PR #2495, opened 2021-09-25
Fix null analysis errors in marketplace bundle

wborn:
These changes resolve all the null analysis errors found in this bundle.


kaikreuzer (approved):
Thanks!

PR #2499, opened 2021-09-27
replace multi-parameter constructor with builder on org.openhab.core.Addon

J-N-K:
as discussed here: #2490 


wborn (approved):
Thanks for building the builder!

PR #2503, opened 2021-10-03
add item name to ScriptBusEvent log messages

J-N-K:
Currently the item name is missing and it is nearly impossible to find the cause of the warning as there is no hint which rule sent the wrong update/command. Now at least the item name is shown which helps to identify the offending rule.


wborn (approved):
Thanks for adding the hints!

Issue #2465, opened 2021-08-26
Prevent addons from shutting down shared thread pools


janvyb:
openHAB provides addons with API for obtaining shared thread pools ([ThreadPoolManager](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core/src/main/java/org/openhab/core/common/ThreadPoolManager.java)), which is also used to obtain `scheduler ` property in BaseThingHandler. It's a good thing so that everyone isn't spawning their own threads and most addons use it.

The problem is that nothing stops any addon from shutting down this shared pool, which has catastrophic consequences. When that happens, all addons/services which use that shared pool stop working, throwing `java.util.concurrent.RejectedExecutionException` when they try to schedule new jobs. There's no way to recover from that other than restarting openHAB, and it's also hard to diagnose since lots of things keeps working fine.

I've recently encountered this problem with [miio binding](https://github.com/openhab/openhab-addons/issues/11152), but forum seems to be full of similar problems (like [this](https://community.openhab.org/t/solved-task-java-util-concurrent-scheduledthreadpoolexecutor-scheduledfuturetask-not-completed-errors/124629) or [this](https://community.openhab.org/t/telegram-binding-bot-suddenly-not-running/117876), and I'm certaint that I encountered this problem a few times myself before) and it's pretty much impossible from logs to find out what happened, because the exception is thrown from first addon that tries to use that shared pool after it was shut down.

I'd propose overriding `shutdown()` and `shutdownNow()` methods on [WrappedScheduledExecutorService](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/common/WrappedScheduledExecutorService.java) (and maybe also on [QueueingThreadPoolExecutor](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core/src/main/java/org/openhab/core/common/QueueingThreadPoolExecutor.java)) to either do nothing, or at least log some warning with stack trace when someone calls them. It shouldn't cause any issues since `ThreadPoolManager` isn't shutting them down and noone else should either since they are supposed to be shared.

PR #2531, opened 2021-10-19
Prevent shutdown of shared thread pool

### LINKED ISSUE #760: shutdown of Executors provided by ThreadPoolManager
While writing https://github.com/openhab/openhab-core/pull/744#discussion_r279149985 I wonder if such "mistakes" should be prevented by the framework itself.

An executor is taken from the `ThreadPoolManager` and e.g. used at different places.
The user perhaps doesn't know that it is a shared executor and it must not be shut down.

Here my RFC:
* The class `WrappedScheduledExecutorService` is used by the `ThreadPoolManager` internally and it is the only one that uses that class, so it can be modified without affect other ones (it is also part of an internal package).
* Let's override the methods that shutdown an `ExecutorService` (e.g. `shutdown` and `shutdownNow`). The implementation should log a warning (or error) with a message "this method must not be called, please fix your code" and a stack trace. The specific method of the super class is not called.
* We provide a replacement for the specific methods with a prefix e.g. "really" or something similar (`reallyShutdown`, ...), that calls the respective super method.
* The "really shutdown" methods are only known by the `ThreadPoolManager`.
* The thread pool manager could provide methods like `shutdownScheduler(final String poolName)` that takes the pool from the pools map and calls the "real shutdown" method.
* This way it is explicit and very informative that a (shared) scheduler provided by the `ThreadPoolManager` should be shut down.

WDYT (e.g. all @openhab/core-maintainers and and of course all the others)?

janvyb:
Right now shared thread pools can be shut down, which leads to many problems (as described in #760 and #2465).

This PR fixes it by overriding shutdown and shutdownNow methods on wrapper classes around thread pools returned from factory methods on ThreadPoolManager, preventing shutdown and logging errors to help find out what addon calls them. This should not cause any issues since shared pools are never shut down by framework.

Closes #760

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/common/ThreadPoolManager.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/common/ThreadPoolManager.java
@@ -12,17 +12,25 @@
  */
 package org.openhab.core.common;
 
+import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 import java.util.WeakHashMap;
+import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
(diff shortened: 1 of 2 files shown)


kaikreuzer (approved):
Alright, thank you!
Let's merge and see if it all works as expected :-)

Issue #2537, opened 2021-10-25
OpenHAB stops watching any files for changes


ccutrer:
This has gotten to the point where I have to restart openhab constantly to see changes to my items or rules files. I'm on openhab 3.2.0 S2517. The symptom is that no file changes are ever noticed - I have to restart openhab to pick up any changes. I enabled debug logging for org.openhab.core.service (and some the ScriptFileService, as I'm using the JRuby add-on and that's where I first noticed it, but the problem also applies to regular .items files), and I see this in my log when I boot up:

```
2021-10-24 18:26:04.268 [INFO ] [.core.model.lsp.internal.ModelServer] - Started Language Server Protocol (LSP) service on port 5007
2021-10-24 18:26:04.322 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model 'controlpanels.rules'
2021-10-24 18:26:04.734 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model 'pool.rules'
2021-10-24 18:26:05.247 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model 'locks.rules'
2021-10-24 18:26:05.630 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model 'pigpio.rules'
2021-10-24 18:26:05.771 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model 'water.rules'
2021-10-24 18:26:05.887 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model 'myq.rules'
2021-10-24 18:26:06.020 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model 'tv.rules'
2021-10-24 18:26:06.263 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model 'alarm.rules'
2021-10-24 18:26:07.070 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model 'usg.rules'
2021-10-24 18:26:07.144 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model 'minisplit.rules'
2021-10-24 18:26:07.443 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model 'astro.rules'
2021-10-24 18:26:07.692 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model 'computers.rules'
[...]

PR #2538, opened 2021-10-25
Include full exception when WatchQueueReader thread dies

ccutrer:
See #2537

Signed-off-by: Cody Cutrer <cody@cutrer.us>


cweitkamp:
Hi @ccutre,

One question: when logging this error message will a user be able to fix the cause? How?

J-N-K:
No. The only possible way to fix is restarting openHAB.

cweitkamp:
The reason why I am asking - In case this exception leads to an unusable system e.g. because if a blocked rule engine we should give an advise how to solve the situation (e.g. by changing the logged message). This would be much better user experience.

J-N-K:
IMO this should fix itself. The exception should not stop the WatchQueueReader. Maybe the try-catch-Block should be moved inside the for-loop, not around it.

ccutrer:
+1, I'll move the catch inside the loop. I'm just completely in the dark as to what's causing the problem, so am a little afraid it's something that will constantly cause it to crash and restart. For me, the crash is happening every time OpenHAB starts.

Is there any documentation anywhere on how to put a locally compiled core bundle into an otherwise-running snapshot? That's the only place I can reproduce my error. I tried to replace the jar file that apt installs, but it didn't do anything, and if I put it in the addons folder karaf shows it sees the bundle, but it's inactive (presumably because some built-in one has already been loaded).

ccutrer:
It looks like the build failed, but I'm not familiar enough with how it should look to know if it's related. There is a separate exception handling block to exit the thread gracefully on interrupt, but I could see it being a possibility that it just hangs when it tries to shut down now.

jimtng:
https://github.com/openhab/openhab-core/pull/2563 fixed the problem that was discovered by this PR.

bundles/org.openhab.core/src/main/java/org/openhab/core/service/WatchQueueReader.java
```diff
@@ -272,11 +272,9 @@ public void run() {
                 }
 
                 key.reset();
+            } catch (Exception exc) {
+                logger.error("Exception caught in WatchQueueReader: ", exc);
```
wborn, line 276: It is probably better to move the `catch (InterruptedException exc)` for the `key = watchService.take();`  just above this catch. So `InterruptedException`s are still properly handled when they occur when the application is stopping.

If it can recover from these exceptions I don't think it is a good idea to bother end users with this by logging scary stacktraces. When exception handling works as expected there is nothing for them to do and everything will still work.

Furthermore it is always better to use `while (!Thread.currentThread().isInterrupted()) {` instead of `for (;;) {` or `while (true) {` when threads should stop when they are interrupted.

jimtng: > If it can recover from these exceptions I don't think it is a good idea to bother end users with this by logging scary stacktraces. When exception handling works as expected there is nothing for them to do and everything will still work.

I agree but perhaps do a `logger.debug`?

wborn: Yes it's always good to have some debug logging in case the exception handling needs to be tweaked. Debug logging can also be made very scary. :wink:

wborn: Can you address this comment so we can merge this PR @ccutrer?

ccutrer: I think I've addressed all the comments now - use a conditional loop instead of infinite, move the InterruptedException handling to the outermost loop, change that logging to debug, and try to make the message as friendly as possible even when it is logged.

bundles/org.openhab.core/src/main/java/org/openhab/core/service/WatchQueueReader.java
```diff
@@ -273,11 +268,12 @@ public void run() {
                 }
 
                 key.reset();
+            } catch (InterruptedException exc) {
+                logger.debug("WatchQueueReader interrupted; thread exiting: {}", exc.getLocalizedMessage());
+                return;
+            } catch (Exception exc) {
+                logger.error("Exception caught in WatchQueueReader: ", exc);
```
wborn, line 275:
```suggestion
                logger.debug("Exception caught in WatchQueueReader", exc);
```

wborn (approved):
Thanks for the fixes!

Issue #16, opened 2016-02-04
compat1x creation of DateTimeType does not produce equivalent objects


watou:
1.x bindings that put `DateTimeType` objects on the event bus will receive updates that fail the `equals` test because the [compat1x layer's use of `toString` on the ESH version to create the `org.openhab.core.library.types.DateTimeType` equivalent](https://github.com/kaikreuzer/openhab-core/blob/master/bundles/org.openhab.core.compat1x/src/main/java/org/openhab/core/compat1x/internal/TypeMapper.java#L68) does not create objects that pass the `equals` test.  This imprecise conversion is currently causing memory leaks and possibly other bugs under openHAB 2 for 1.x bindings.

Also, out of curiosity, why is this repo yours personally?

watou:
This:
```
    public static void test() {
        DateTimeType gen1 = new DateTimeType();
        DateTimeType gen2 = new DateTimeType(gen1.toString());
        DateTimeType gen3 = new DateTimeType(gen2.toString());

        System.out.println( "gen1 " + (gen1.equals(gen2) ? "equals" : "does not equal") + " gen2" );
        System.out.println( "gen2 " + (gen2.equals(gen3) ? "equals" : "does not equal") + " gen3" );
        System.out.println( "gen3 " + (gen3.equals(gen1) ? "equals" : "does not equal") + " gen1" );
    }
```
produces this:
```
gen1 does not equal gen2
gen2 equals gen3
gen3 does not equal gen1
```

maggu2810:
openhab-core is a set of bundle that are owned by @kaikreuzer 
It's not about the distribution or addons.
So, why not using its own repo if he maintains that bundles?

kaikreuzer:
@watou: I kept it separate as this is closer to the ESH code base - it is simply some code around the ESH runtime that for the one or other reason is not within the ESH repo. As a preparation for opening up (=write access) the openhab Github organisation to a broader group of people, I wanted to keep this governance of this part separate and closer to ESH, so that it can be closely synched.

But back to your issue: Thanks for the analysis, I will fix it. Are there already any issues open regarding the memory leaks that you are mentioning?

watou:
> So, why not using its own repo if he maintains that bundles?

[...]

PR #2543, opened 2021-10-27
Add compilation errors & SAT GitHub annotations

ghys:


wborn:
Many static analysis tools nowadays also support the SARIF format.

It seems you can also use these files to integrate with the GitHub code scanning, see:

https://docs.github.com/en/code-security/code-scanning
https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/about-integration-with-code-scanning

Might also be worth to look into what this exactly does and how well it works when PRs are created. :wink:

ghys:
So in the end I combined these two PRs into one:
- one that reports & displays Java compilation errors as annotations ([example](https://github.com/ghys/openhab-core/pull/1/files#diff-899c7989dc13bc74a38689d94df1fdf53a812cf558269e78df4e7258a11b1c47))
- one that reports SAT errors as annotations ([example](https://github.com/ghys/openhab-core/pull/2/files#diff-899c7989dc13bc74a38689d94df1fdf53a812cf558269e78df4e7258a11b1c47))

(Try Ctrl-F5 if the annotations don't show)

I eventually fixed some unrelated problems in this PR,  that's because the action originally had a bug and was treating every problem as errors unless there were warnings - so info levels were treated as failures... If they are not wanted I can revert them.

ghys:
> https://docs.github.com/en/code-security/code-scanning
https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/about-integration-with-code-scanning

Those are interesting as well but it might be better to postpone them to another workflow dedicated to code checks?

wborn (approved):
Thanks a lot. :+1: 
I gave it a brief test in my own repo and it seems to work. :slightly_smiling_face: 
So let's test how it works for real!

ghys:
If you think I'd be a good addition to other repos (notably openhab-addons) I can open similar PRs to those too.

wborn:
Yes it would be nice to test it in the add-ons repo too!

I plan on adding GHA to more repos when all bugs are sorted out and we're happy with the features in the most active repos. :slightly_smiling_face: That prevents a lot of PRs to tune GHA everywhere. It might also be worth looking into [reusing workflows](https://docs.github.com/en/actions/learn-github-actions/reusing-workflows) if we add GHA to many repos.

vivox80pro:
> 

#16

PR #2544, opened 2021-10-28
Add i18n-maven-plugin to make internationalization easier

wborn:
This plugin simplifies generating the default translation .properties files from the add-on XML information files.

It reuses the same XStream parsing classes that are used by openhab-core for parsing the binding/config/thing XML files.
It will also keep any existing default translations already present in property files for translations using `@text/`.
Furthermore it will nicely group and sort the translations.

After building this Maven plugin you can use it on add-ons using:

`mvn org.openhab.core.tools:i18n-maven-plugin:3.2.0-SNAPSHOT:generate-default-translations`

This [commit](https://github.com/wborn/openhab-addons/commit/769d7dbce4903f30ca78ff5b3da73fba22b61777) shows what it generates when run on all add-ons.

### DIFF
--- a/tools/i18n-plugin/src/main/java/org/openhab/core/tools/i18n/plugin/XmlToTranslationsConverter.java
+++ b/tools/i18n-plugin/src/main/java/org/openhab/core/tools/i18n/plugin/XmlToTranslationsConverter.java
@@ -43,7 +43,7 @@ import org.openhab.core.tools.i18n.plugin.Translations.TranslationsSection;
 import org.openhab.core.types.StateDescription;
 
 /**
- * Converts XML based {@link AddonInfo} to {@link Translations}.
+ * Converts XML based {@link BundleInfo} to {@link Translations}.
  *
  * @author Wouter Born - Initial contribution
  */
@@ -56,32 +56,32 @@ public class XmlToTranslationsConverter {
         this.log = log;
     }
 
-    public Translations convert(AddonInfo addonInfo) {
-        BindingInfoXmlResult bindingInfoXml = addonInfo.getBindingInfoXml();
+    public Translations convert(BundleInfo bundleInfo) {
+        BindingInfoXmlResult bindingInfoXml = bundleInfo.getBindingInfoXml();
         if (bindingInfoXml == null) {
             log.warn("Cannot generate translations because there is no XML binding info");
             return Translations.translations();
         }
 
         return Translations.translations( //
-                bindingSection(addonInfo), //
-                bindingConfigSection(addonInfo), //
-                thingTypesSection(addonInfo), //
-                thingTypesConfigSection(addonInfo), //
-                channelGroupTypesSection(addonInfo), //
-                channelTypesSection(addonInfo), //
-                channelTypesConfigSection(addonInfo));
+                bindingSection(bundleInfo), //
+                bindingConfigSection(bundleInfo), //
+                thingTypesSection(bundleInfo), //
+                thingTypesConfigSection(bundleInfo), //
+                channelGroupTypesSection(bundleInfo), //
+                channelTypesSection(bundleInfo), //
+                channelTypesConfigSection(bundleInfo));
     }
 
-    private TranslationsSection bindingSection(AddonInfo addonInfo) {
+    private TranslationsSection bindingSection(BundleInfo bundleInfo) {
         String header = "binding";
-        BindingInfoXmlResult bindingInfoXml = addonInfo.getBindingInfoXml();
+        BindingInfoXmlResult bindingInfoXml = bundleInfo.getBindingInfoXml();
         if (bindingInfoXml == null) {
             return section(header);
         }
 
         BindingInfo bindingInfo = bindingInfoXml.getBindingInfo();
-        String bindingId = addonInfo.getBindingId();
+        String bindingId = bundleInfo.getBindingId();
 
         String keyPrefix = String.format("binding.%s", bindingId);
 
@@ -91,9 +91,9 @@ public class XmlToTranslationsConverter {
         ));
     }
 
-    private TranslationsSection bindingConfigSection(AddonInfo addonInfo) {
+    private TranslationsSection bindingConfigSection(BundleInfo bundleInfo) {
         String header = "binding config";
-        BindingInfoXmlResult bindingInfoXml = addonInfo.getBindingInfoXml();
+        BindingInfoXmlResult bindingInfoXml = bundleInfo.getBindingInfoXml();
         if (bindingInfoXml == null) {
             return section(header);
         }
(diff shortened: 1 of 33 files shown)


lolodomo:
Trying with the vigicrues binding, I got these channels types:
```
channel-type.vigicrues.alert-icon.label = Pictogram
channel-type.vigicrues.alert-icon.description = Official pictogram associated to alert level.
channel-type.vigicrues.alert-level.label = Alert
channel-type.vigicrues.alert-level.state.option.0 = Green
channel-type.vigicrues.alert-level.state.option.1 = Yellow
channel-type.vigicrues.alert-level.state.option.2 = Orange
channel-type.vigicrues.alert-level.state.option.3 = Red
channel-type.vigicrues.comment.label = Comment
channel-type.vigicrues.flow.label = Current Flow
channel-type.vigicrues.gauge.label = Relative Measure
channel-type.vigicrues.height.label = Height
channel-type.vigicrues.height.description = Water level in the river
channel-type.vigicrues.observation-time.label = Observation Time
channel-type.vigicrues.observation-time.description = Observation date and time
```
But in the original french properties file created by @clinique, I have other entries like
```
channel-type.vigicrues.relative-flow.label = DÃ©bit relatif
channel-type.vigicrues.relative-flow.description = DÃ©bit relatif par rapport aux crues historiques.
channel-type.vigicrues.relative-height.label = Hauteur relative
channel-type.vigicrues.relative-height.description = Hauteur relative par rapport aux crues historiques.
channel-type.vigicrues.short-comment.label = Situation
channel-type.vigicrues.short-comment.description = Bref descriptif de la situation.
```
Here is how are defined these channel in the thing file:
```
		<channels>
			<channel id="height" typeId="height"/>
			<channel id="relative-height" typeId="gauge">
				<label>Relative Height</label>
				<description>Current height toward historical floods.</description>
			</channel>
			<channel id="flow" typeId="flow"/>
			<channel id="relative-flow" typeId="gauge">
				<label>Relative Flow</label>
				<description>Current flow toward historic floods.</description>
			</channel>
			<channel id="alert" typeId="alert-level"/>
			<channel id="alert-icon" typeId="alert-icon"/>
			<channel id="short-comment" typeId="comment">
				<label>Short description</label>
			</channel>
			<channel id="comment" typeId="comment"/>
			<channel id="observation-time" typeId="observation-time"/>
		</channels>
```

I am asking myself if there is not a bug in your tool which should create an entry for each channel, not only for eachchannel type ?

lolodomo:
Finally, I am not sure there is a bug. This looks more like a missing feature or maybe I don't know what path to use for example to translate the label of the `short-comment` channel. `channel-type.vigicrues.short-comment.label` is not working. `channel.vigicrues.short-comment.label` is not working. @cweitkamp : do you know what path should be used in this case (label or description defined on the channel itself) ?
If this is not implemented, the solution would be to use `@text`.

lolodomo:
I finally found the right key to use for channels. For example: `thing-type.vigicrues.station.channel.short-comment.label`

Ideally it should be considered by your i18n maven plugin.

wborn:
I think it should be possible. I remember I had a look at the astro binding where labels/descriptions are overriden for certain group types. But I probably forgot to add something similar for channel types. :blush:

wborn:
I think this is the logic that takes care of it for channel group types with overriden labels/descriptions:

https://github.com/openhab/openhab-core/pull/2544/files#diff-b7101042a9b1b8825f910f519872ac389701fb78eaf34c86116933f2b9d30f83R206-R220

lolodomo:
I checked the recent default properties file generated and I think we have to fix those for the following bindings: openuv, weathercompany, kodi and probably awmfritz. So you can test an update of your tool on one of these bindings.

wborn:
It should now also generate default translations for overridden channels in thing types @lolodomo.

Furthermore it now also generates translations for bundles that only have config descriptions, like automation/io/persistence/transform/voice or core bundles. :slightly_smiling_face:

lolodomo:
> It should now also generate default translations for overridden channels in thing types

Just tested it with the openuv bidnign and it works.
I am going to fix few bindings.

lolodomo:
One option would be to run your tool on all bindings without any properties file. This will open these bindings to translations in Crowdin. This will not be a full internationalisation of these bindings but probably an internationalisation at around 90% or even 95%. What could be missing (because it requires specific code) is thing status message, discovery result label or hardcoded English string value for channel state.

For bindings having already properties files, we have to take care and do it manually with precaution.

WDYT @wborn ?

wborn:
> This will not be a full internationalisation of these bindings but probably an internationalisation at around 90% or even 95%. 

Yes I also think the generated properties files will be a good start and it allows users to start translating while devs add the last remaining %. Add-ons that no longer have a maintainer would also greatly benefit from this.

My idea was to first get a feeling on how well it works and iron out the bugs before mass generating them all. I think it already works pretty well.

But even with all bugs ironed out it will not remove any duplicate strings. However being able to translate something multiple times is better than not being able to translate strings at all. :wink:

ghys:
> This will not be a full internationalisation of these bindings but probably an internationalisation at around 90% or even 95%.

If it is any incentive, I believe it could also trigger the i18n of the admin part of the UI as well; currently while all the user areas are translatable (and translated), most the admin pages still aren't because I figured a significant part of those are parameter sheets that would still appear in English, so IMO it is not worth the effort at this time (I mean the effort of replacing all these hardcoded strings, a tedious and thankless job... translating strings is another matter but past translation campaign experiences with Crowdin gives much confidence the community will step up).
If the admin needs to understand English to figure out the parameters, then the other parts of the admin UI should be no problem either so there's little point in translating it. But if we can achieve a good i18n "coverage" in most key languages across the most popular bindings then it becomes more appealing to go all the way.

cweitkamp:
I added a missing feature to translate `CommandOptions` in https://github.com/wborn/openhab-core/pull/4.

jimtng:
> `mvn org.openhab.core.tools:i18n-maven-plugin:3.2.0-SNAPSHOT:generate-default-translations`

I used this handy tool recently. Is there a way to make this command a lot simpler, e.g. `mvn tools:generate-translations` or something similar?

wborn:
I think when it is merged and added as plugin dependency you should also be able to use it with:

`mvn i18n:generate-default-translations`

See also:

* https://maven.apache.org/guides/plugin/guide-java-plugin-development.html#shortening-the-command-line
* https://maven.apache.org/guides/introduction/introduction-to-plugin-prefix-mapping.html

tools/i18n-plugin/src/main/java/org/openhab/core/tools/i18n/plugin/XmlToTranslationsConverter.java
```diff
@@ -0,0 +1,368 @@
...
+            if (stateDescription != null) {
+                stateDescription.getOptions().stream()
+                        .map(option -> entry(
+                                String.format("%s.%s.state.option.%s", keyPrefix, channelId, option.getValue()),
+                                option.getLabel()))
+                        .forEach(entriesBuilder::add);
+
+                if (stateDescription.getPattern() != null) {
+                    String pattern = stateDescription.getPattern();
+                    if (pattern != null && pattern.contains("%1$t")) {
```
cweitkamp, line 286: Did you add this check (`pattern.contains("%1$t")`) by intention to filter every other pattern except date time related ones? I was looking into this because I am missing pattern for channel types from feed binding for translation. It uses `"%tc"` as pattern. I personally would be fine with this and change it for feed binding to something else. Or should we extend it here?

cweitkamp (commented):
Your contribution is in a good shape @wborn . Do you consider to change this PRs type from draft to ready to be merged soon?

tools/i18n-plugin/src/main/java/org/openhab/core/tools/i18n/plugin/XmlToTranslationsConverter.java
cweitkamp: See https://github.com/openhab/openhab-addons/pull/11618

wborn: I added the filter because when it is not applied it will also start adding the patterns for the following states to .properties files:

* `<state pattern="%s">`
* `<state pattern="%.2f" readOnly="true"/>`
* `<state pattern="%.1f %unit%"/>`

I doubt these would often result in other translations.

Although later I realised that any string that has a space in it, is also a potential translation candidate for any language that is right-to-left instead of left to right. :slightly_smiling_face:

wborn:
> Do you consider to change this PRs type from draft to ready to be merged soon?

Adding some unit tests for this plugin is still on my todo list. Also I want to add a plugin config option to skip generating default translations when the .properties file already exists. That would simplify creating a PR that adds default translations to all add-ons that do not have these already.

cweitkamp (approved):
LGTM. Thank you very much 

This will help developers a lot and safes much time regarding creation of translation properties. I bet you will submit a bunch of documentation later 😉 .

Let's go!

cweitkamp:
Am I allowed to merge this as I submitted code - not much, but I did - for it?

//CC @openhab/core-maintainers

kaikreuzer:
> Am I allowed to merge this as I submitted code - not much, but I did - for it?

Fine with me, go ahead!

PR #2552, opened 2021-10-31
Exclude logback-classic from org.apache.aries.jax.rs.whiteboard

wborn:
There should only be one SLF4J binding used for logging and we already use slf4j-simple.

This fixes the following warning when running unit tests:

```
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/home/wouter/.m2/repository/org/slf4j/slf4j-simple/1.7.32/slf4j-simple-1.7.32.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/home/wouter/.m2/repository/ch/qos/logback/logback-classic/1.2.3/logback-classic-1.2.3.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [org.slf4j.impl.SimpleLoggerFactory]
```

### DIFF
--- a/bom/test/pom.xml
+++ b/bom/test/pom.xml
@@ -73,6 +73,10 @@
       <version>2.0.1</version>
       <exclusions>
         <exclusion>
+          <groupId>ch.qos.logback</groupId>
+          <artifactId>logback-classic</artifactId>
+        </exclusion>
+        <exclusion>
           <groupId>junit</groupId>
           <artifactId>junit</artifactId>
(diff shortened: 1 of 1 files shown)


cweitkamp (approved):
Works locally for me. Thanks.

PR #2553, opened 2021-11-01
Add feature to sort group members with metadata configuration

cweitkamp:
- Added feature to sort group members with metadata configuration

See https://github.com/openhab/openhab-core/pull/2123#issuecomment-765904668

This is an initial draft on sorting group members with metadata configuration. Items in groups can be sorted by name, label or state. Default ordering is ascending.

demo.items
```java
Group:Number testGroup "Test Group" {
    sortBy="STATE" [ ordering="DESCENDING" ]
}
```

Allowed values for `sortBy`: `LABEL`, `NAME` and `STATE`.
Allowed values for `ordering`: `DESCENDING`, otherwise ascending ordering will be used. Configuration of `ordering` is optional.

Sort members by state picks up a `Comparator` from items, which can be implemented individually for each item type. `Item` class provides a default `Comparator` to sort states in alphabetically order by applying `toFullString()` on the state first.

UIs and other services have to be changed afterwards to benefit from this feature.

@openhab/core-maintainers wdyt?
// CC @ghys @rkoshak

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ItemRegistryImpl.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ItemRegistryImpl.java
@@ -198,6 +198,7 @@ public class ItemRegistryImpl extends AbstractRegistry<Item, String, ItemProvide
         if (item instanceof GenericItem) {
             GenericItem genericItem = (GenericItem) item;
             genericItem.setEventPublisher(getEventPublisher());
+            genericItem.setMetadataRegistry(metadataRegistry);
             genericItem.setStateDescriptionService(stateDescriptionService);
             genericItem.setCommandDescriptionService(commandDescriptionService);
             genericItem.setUnitProvider(unitProvider);
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/items/GenericItem.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/items/GenericItem.java
@@ -62,6 +62,8 @@ public abstract class GenericItem implements ActiveItem {
 
     protected @Nullable EventPublisher eventPublisher;
 
+    protected @Nullable MetadataRegistry metadataRegistry;
+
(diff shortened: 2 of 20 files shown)


kaikreuzer (approved):
Code looks good to me in general and the feature might be useful.
Before merging, I'd like to see a comment from @ghys as you had pinged him specifically as well on whether this fits for the UI to support it or whether some tweaks are desired.

cweitkamp:
Thanks for looking into this. From my POV we at least need to tweak the REST API to make use of this feature or add a special method for retrieving sorted groups / member of groups.

ghys:
The UI has its own sorting function which will consider the index put in the `widgetOrder` metadata namespace first, then will sort labels (or fallback to the name if the item has no label) alphabetically. It was because in many cases the order of items in API responses seemed random and indeterminate.
When you use features like the "group popup" action, the home page cards, the "add from model" function in pages, or the `oh-repeater` component with a `itemsInGroup` source, that function will be used instead of the order of the API. So there would be a need to figure out that in some cases the order of the API is the correct one.

cweitkamp:
The feature to order based on `widgetOrder` can be integrated in this implementation too.

Would it make sense to add an `orderIndex` field to the `EnrichedItemDTO` which carries the previously determined ordering? This can be used by consumers to ensure correct ordering.

rkoshak:
Holy cow, there's been a lot of stuff going on on GitHub I've simply missed. Playing catchup.

All I have to add is this is something that has been asked for many times by sitemap users. I could see it potentially being useful in rules too, though there are already ways to sort members of a Group there. 

I can't comment on MainUI concerns.

J-N-K:
@cweitkamp What is the status here? Judging from your last comment I assume there's still work to do before we can merge?

kaikreuzer:
@cweitkamp Are you still around and could comment on the status here?

wborn:
Is this feature still wanted? The PR now conflicts with "main", and the earlier discussion did not resolve whether sorting should happen in core, REST, or the UI. Should it be updated or closed?

PR #2557, opened 2021-11-03
GitHub Actions workflow improvements

wborn:
* Checkout merged branches for pull requests
* Echo the mvn command used for builds
* Use more functions in maven-build so it is similar to what will be used for add-on builds

### DIFF
--- a/.github/scripts/maven-build
+++ b/.github/scripts/maven-build
@@ -1,7 +1,11 @@
 #!/bin/bash
 
 BUILD_LOG=build.log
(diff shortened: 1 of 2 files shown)


wborn:
> Checkout merged branches for pull requests

It did seem like that if you press the "Re-run all jobs" button, it reruns the original workflow and it does not pick up any workflow file updates. :slightly_frowning_face:

cweitkamp (approved):
LGTM. Thanks.

PR #2561, opened 2021-11-05
Use POM packaging with index BOMs

wborn:
Looks like only the POM dependencies are used by the indexer, so the JAR is unused and useless as this warning correctly indicates when you build them:

```
[WARNING] JAR will be empty - no content was marked for inclusion!
```


kaikreuzer (approved):
Thanks!

PR #2564, opened 2021-11-08
[marketplace] avoid collisions with multiple AddonProviders

J-N-K:
The current implementation could result in collision of file-path and failed in determining the installation status if other AddonProviders re-use the marketplace addon handlers.


J-N-K:
This will still not work. Update in progress.

J-N-K:
This is not the most elegant solution (which would be to have a structure like `userdata/marketplace/<addonservice>/<addonid>`. However, this would require a breaking change for all current users, because the addons installed are currently under `userdata/marketplace/<addonid>` and would need to be removed. I now opted for `userdata/marketplace//UIDUtils.encode(<addonservice>:<addonid>)` for non-`marketplace`-addonservices and kept the "old" way for the marketplace.

hmerk (approved):
LGTM

cweitkamp (approved):
LGTM. Thanks.

PR #2565, opened 2021-11-08
Make LifecycleScriptExtensionProvider and its Disposable interface public

### LINKED ISSUE #2287: [JSR223] The "Disposable" interface in "LifecycleScriptExtensionProvider" is not accessible from scripts as it is not declared public
The "ScriptExtensionProvider" provides a 'lifecycleTracker' object allowing scripts to register for disposal events.

But, when trying to use it in a JSR223 script, it fails as its inner interface "Disposable" is not declared public, so it cannot be instanciated:

```
var DisposeTracker = new org.openhab.core.automation.module.script.internal.defaultscope.LifecycleScriptExtensionProvider$Disposable() {
        dispose: function() {
                // Do whatever
        }
}

lifecycleTracker.addDisposeHook(DisposeTracker);

```

This produces the following error:

`Error during evaluation of script 'file:/etc/openhab/automation/jsr223/GroupAutoUpdate.js': TypeError: new cannot be used with non-public java type org.openhab.core.automation.module.script.internal.defaultscope.LifecycleScriptExtensionProvider$Disposable
`

Just making it public should fix this, unless I am missing something.

https://github.com/openhab/openhab-core/blob/dea597480f504fd0d1de3e979e4db6dd03cfd73d/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/defaultscope/LifecycleScriptExtensionProvider.java#L101

jimtng:
LifecycleScriptExtensionProvider's Disposable interface is package-private which caused scriptry engines, e.g. JRuby unable to create a hook needed by the LifecycleTracker.addDisposeHook(). The hook is implemented from the `Disposable` functional interface which was not public.

Just making Disposable public without moving LifecycleScriptExtensionProvide out of the internal package wasn't enough.

Resolve #2287 

See https://github.com/boc-tothefuture/openhab-jruby/issues/366

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/LifecycleScriptExtensionProvider.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/LifecycleScriptExtensionProvider.java
@@ -10,7 +10,7 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  */
-package org.openhab.core.automation.module.script.internal.defaultscope;
+package org.openhab.core.automation.module.script;
 
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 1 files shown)


jpg0 (approved):
LGTM

cweitkamp (approved):
LGTM. Thanks.

PR #2567, opened 2021-11-16
[DSL] Added variable 'triggeringChannel' to DSL rules

### LINKED ISSUE #2143: For channel-based triggers, can no longer extract the channel which triggered the event using the receivedEvent implicit variable
In OH2.x the receivedEvent implicit variable allowed the ability to extract the name of the channel which triggered the event

ie: I have a rule which triggers at dusk and includes the line ```val trigger = receivedEvent.getChannel().toString()```.
val gets returned as: ```astro:sun:home:civilDusk#event```.

In OH3 the way the implicit variable works has changed, and it seems we have lost the ability to get to the name of the channel.  For what it's worth the error is ```‘getChannel’ is not a member of ‘java.lang.String’```. I'm not sure if this would be considered a regression. 

Maybe a solution might be a new implicit variable called triggeringChannel?

cweitkamp:
- Added variable 'triggeringChannel' to DSL rules

Closes #2143 


kaikreuzer (approved):
lgtm, thanks!

PR #2569, opened 2021-11-16
[test] Added Magic Button triggering some events every 5 seconds

cweitkamp:
- Added Magic Button triggering some events every 5 seconds
- Minor improvements

### DIFF
--- a/bundles/org.openhab.core.test.magic/src/main/resources/OH-INF/thing/thing-types.xml
+++ b/bundles/org.openhab.core.test.magic/src/main/resources/OH-INF/thing/thing-types.xml
@@ -11,6 +11,17 @@
 		</config-description>
 	</thing-type>
 
+	<thing-type id="button">
+		<label>Magic Button</label>
+		<description>A button triggering different events.</description>
+
+		<channels>
+			<channel typeId="system.rawbutton" id="rawbutton"/>
+			<channel typeId="system.button" id="button"/>
+			<channel typeId="system.rawrocker" id="rawrocker"/>
+		</channels>
+	</thing-type>
+
 	<thing-type id="onoff-light" extensible="alert">
 		<label>Magic Light - On/Off</label>
 		<description>A on/off light</description>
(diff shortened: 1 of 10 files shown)


bundles/org.openhab.core.test.magic/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -11,6 +11,17 @@
...
 	</thing-type>
 
+	<thing-type id="button">
+		<label>Magic Button</label>
+		<description>A button triggering different events.</description>
+
+		<channels>
+			<channel typeId="system.rawbutton" id="rawbutton"/>
+			<channel typeId="system.button" id="button"/>
+			<channel typeId="system.rawrocker" id="rawrocker"/>
```
wborn, line 21:
```suggestion
			<channel typeId="system.button" id="button"/>
			<channel typeId="system.rawrocker" id="rawrocker"/>
```

wborn (changes requested):
I think the audience will like these new magic tricks. But there is a small issue in the added magic. :wink:

bundles/org.openhab.core.test.magic/src/main/resources/OH-INF/thing/thing-types.xml
cweitkamp: Apparently we have all this different buttons.i the default system channel types:

https://github.com/openhab/openhab-core/blob/f9d201c35bacfe48270526419b8ac91bf488db30/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/DefaultSystemChannelTypeProvider.java#L67-L69

wborn: It was still a bit early when I reviewed this so I thought it was using the same channel type twice. But I now see they are different. 🕵️‍♂️

wborn (approved):
Thanks!

PR #2570, opened 2021-11-17
Add Timer.isCancelled() method

jimtng:
I found the need for Timer.isCancelled() inside a long running timer that self-reschedules itself.

Example:

```java
var timer = null

rule "Long running timer"
when
    System started
then
    timer = createTimer(now.plusSeconds(1), [|
        downloadFile() // do something that takes a long time here, e.g. a http get from a large source            
        timer.reschedule(now.plusSeconds(1))
    ])
end

rule "Cancel the timer"
when
    Item Switch1 received command ON
then
    timer.cancel()
end
```

When Switch1 received a command ON while `downloadFile()` is executing, the timer got cancelled, but the currently running task is still continuing, it will reschedule the timer and therefore the timer continues to run despite having been "cancelled".

In this case, the timer code can check with `isCancelled()` before calling `reschedule` as such:
```java
rule "Long running timer"
when
    System started
then
    timer = createTimer(now.plusSeconds(1), [|
        downloadFile() // do something that takes a long time here, e.g. a http get from a large source
        if (!timer.isCancelled())  {
            timer.reschedule(now.plusSeconds(1))
        }
    ])
end
```

### DIFF
--- a/bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/internal/actions/TimerImpl.java
+++ b/bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/internal/actions/TimerImpl.java
@@ -36,8 +36,6 @@ public class TimerImpl implements Timer {
     private final SchedulerRunnable runnable;
     private ScheduledCompletableFuture<Object> future;
 
-    private boolean cancelled;
-
     public TimerImpl(Scheduler scheduler, ZonedDateTime startTime, SchedulerRunnable runnable) {
         this.scheduler = scheduler;
         this.startTime = startTime;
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/internal/actions/TimerImpl.java
```diff
@@ -48,26 +46,29 @@ public TimerImpl(Scheduler scheduler, ZonedDateTime startTime, SchedulerRunnable
 
     @Override
     public boolean cancel() {
-        cancelled = true;
         return future.cancel(true);
     }
 
     @Override
     public boolean reschedule(ZonedDateTime newTime) {
```
cweitkamp, line 53: Not related to your contribution but I am wondering if we should synchronize this method to have a thread safe implementation? Wdyt?

cweitkamp (approved):
LGTM. Thanks.

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/internal/actions/TimerImpl.java
jimtng: It appears to be thread safe already https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/CompletableFuture.html#cancel(boolean)

cweitkamp: What I am talking about is a small race condition on the `future` property of this implementation when calling the `reschedule(...)` method because we first cancel the existing future and afterwards assign a new `CompletableFuture` to it.

PR #2582, opened 2021-11-28
Use Maven 3.8.4 in GHA CI builds

wborn:
This new version fixes a few regressions, see:

https://issues.apache.org/jira/secure/ReleaseNote.jspa?projectId=12316922&version=12350685


cweitkamp (approved):
LGTM. Thanks.

PR #2584, opened 2021-12-01
Add i18n-maven-plugin dependency

wborn:
When the plugin dependency is managed you can also use the plugin without adding GAV parameters to commands.

E.g. it allows for using it with:

```
mvn i18n:generate-default-translations
```

Related to #2544

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -540,6 +540,12 @@ Import-Package: \\
         </plugin>
 
         <plugin>
+          <groupId>org.openhab.core.tools</groupId>
+          <artifactId>i18n-maven-plugin</artifactId>
+          <version>${project.version}</version>
+        </plugin>
+
+        <plugin>
           <groupId>org.openhab.tools.sat</groupId>
           <artifactId>sat-plugin</artifactId>
           <version>${sat.version}</version>
@@ -562,6 +568,7 @@ Import-Package: \\
             </execution>
           </executions>
         </plugin>
+
         <plugin>
           <groupId>com.github.ekryd.sortpom</groupId>
           <artifactId>sortpom-maven-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


cweitkamp (approved):
LGTM. Thanks.

pom.xml
```diff
@@ -539,6 +539,12 @@ Import-Package: \\
           </configuration>
         </plugin>
 
+        <plugin>
+          <groupId>org.openhab.core.tools</groupId>
+          <artifactId>i18n-maven-plugin</artifactId>
+          <version>${project.version}</version>
```
kaikreuzer, line 545: @wborn This seems to break our release build as it is a SNAPSHOT dependency, which is not allowed - see https://ci.openhab.org/view/Sandbox/job/sandbox-openhab3-release/439/console.
We probably would have to maintain it in a separate repo just like SAT or only add this dependency after releasing version 3.2.

J-N-K: The separate repo is probably the cleanest solution. You could also try to move the whole plugin configuration to a profile that is enabled by default but disabled during release. I'm not 100% sure if the no-snapshot check also includes disabled profiles, though.

wborn: I've reverted the PR for now. I'm not a big fan of many repos because of all the overhead and when code changes it can only be used after creating a release. Especially for new code like this tweaks may be necessary. :wink:

kaikreuzer: I understand your pov, but afaik you cannot build plugins and use them in the very same reactor. So they will require an independent release cycle - however we achieve this.

wborn: It could also be an idea to put a different label on the "box" currently called "static-code-analysis" (build-tools?). Then the name would allow for putting more similar stuff into that box. :slightly_smiling_face:

But is still possible that when the data in OH-INF changes, the plugin would start depending on OHC snapshot dependencies to be able to parse it. So you'd not be able to use an updated version of the plugin until OHC is released.

PR #2585, opened 2021-12-03
Revert "Add i18n-maven-plugin dependency"

wborn:
Reverts openhab/openhab-core#2584 because it makes the release build fail.


kaikreuzer:
Thanks!

PR #2587, opened 2021-12-03
[automation] Added basic time-related classes for usage in script and rules

cweitkamp:
- Added basic time-related classes for usage in script and rules

This allows to use DateTime functions a lot easier for users without defining additional imports. We have public methods in different types which makes use of these classes e.g.

https://github.com/openhab/openhab-core/blob/a9504bfced5054bb8c7a864876f66df27c69c661/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java#L170

https://github.com/openhab/openhab-core/blob/a9504bfced5054bb8c7a864876f66df27c69c661/bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/actions/Timer.java#L76

`org.joda.time` API has a method `DateTime.now().withTimeAtStartOfDay()` which was used by many users. The replacement is `ZonedDateTime.now().truncatedTo(ChronoUnit.DAYS)`.

`Duration` class allows easy calculation of differences between two timestamps (`Duration.between(Temporal startInclusive, Temporal endExclusive)` - `ZonedDateTime` implemets `Temporal` interface).

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/defaultscope/DefaultScriptScopeProvider.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/defaultscope/DefaultScriptScopeProvider.java
@@ -17,6 +17,10 @@ import java.net.URLEncoder;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.time.Duration;
+import java.time.ZoneId;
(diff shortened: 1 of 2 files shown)


kaikreuzer:
The replacement that I am actually using in my rules is 
```
now.toLocalDate.atStartOfDay.atZone(ZoneId.systemDefault)
```
but it all isn't as nice as the `withTimeAtStartOfDay` from old Jodatime. 😞

kaikreuzer (approved):
lgtm, thanks!

PR #2592, opened 2021-12-05
[marketplace] add a 3rd Party marketplace

J-N-K:
This adds a new `AddonService` implementation that allows to add marketplaces maintained by 3rd party individuals (similar to alternative app stores on the smartphone). 

The configuration allows one or multiple marketplaces which provide their addons via JSON files (instead of a topic in the community forum). Installation is done by re-using the already present .jar, .kar, .json, .yaml bundle installers (same as community marketplace).

The JSON referenced by the URL encapsulates an array of one or more addon definitions like

```
[
  {
    "id": "tld-johndoe-binding-foo",
    "type": "binding",
    "title": "Foo Binding",
    "description": "This binding integrates Foo smarthome devices with openHAB",
    "link": "https://johndoe.tld/foodoc.html",
    "version": "1.0.0",
    "author": "John Doe",
    "maturity": "stable",
    "content_type": "application/vnd.openhab.feature;type=karfile",
    "url": "https://johndoe.tld/tld.johndoe.binding.foo-1.0.0.kar"
  },
  {
    ...
  }
]
```

This could also be used to distribute version-specific (by using different JSON files) bindings with back ported feature or bugfixes.

### DIFF
--- a/bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/json/AddonEntryDTO.java
+++ b/bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/json/AddonEntryDTO.java
@@ -1,0 +1,35 @@
+/**
+ * Copyright (c) 2010-2021 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.addon.marketplace.internal.json;
+
+import com.google.gson.annotations.SerializedName;
+
+/**
+ * The {@link AddonEntryDTO} is a DTO for encapsulating a single addon information.
+ *
+ * @author Jan N. Klug - Initial contribution
+ */
+public class AddonEntryDTO {
+    public String id = "";
+    public String type = "";
+    public String description = "";
+    public String title = "";
+    public String link = "";
+    public String version = "";
+    public String author = "";
+    public String configDescriptionURI = "";
+    public String maturity = "unstable";
+    @SerializedName("content_type")
+    public String contentType = "";
+    public String url = "";
+}
--- a/bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/json/JsonAddonService.java
+++ b/bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/json/JsonAddonService.java
@@ -36,6 +36,7 @@ import org.openhab.core.addon.AddonService;
 import org.openhab.core.addon.AddonType;
 import org.openhab.core.addon.marketplace.MarketplaceAddonHandler;
 import org.openhab.core.addon.marketplace.MarketplaceHandlerException;
+import org.openhab.core.addon.marketplace.internal.json.model.AddonEntryDTO;
 import org.openhab.core.config.core.ConfigurableService;
 import org.openhab.core.events.Event;
 import org.openhab.core.events.EventPublisher;
@@ -59,10 +60,14 @@ import com.google.gson.reflect.TypeToken;
  */
 @Component(immediate = true, configurationPid = "org.openhab.core.marketplace.json", //
         property = Constants.SERVICE_PID + "=org.openhab.core.marketplace.json")
-@ConfigurableService(category = "system", label = "JSON 3rd Party Marketplace", description_uri = JsonMarketplaceAddonService.CONFIG_URI)
-public class JsonMarketplaceAddonService implements AddonService {
-    static final String CONFIG_URI = "system:jsonmarketplace";
-    private static final String ADDON_ID_PREFIX = "jsonmarketplace:";
+@ConfigurableService(category = "system", label = "JSON 3rd Party Addon Service", description_uri = JsonAddonService.CONFIG_URI)
+public class JsonAddonService implements AddonService {
+    static final String CONFIG_URI = "system:jsonaddonservice";
+    private static final String CONFIG_URLS = "urls";
+    private static final String CONFIG_SHOW_UNSTABLE = "showUnstable";
+
+    private static final String SERVICE_NAME = "json";
+    private static final String ADDON_ID_PREFIX = SERVICE_NAME + ":";
 
     private static final Map<String, AddonType> TAG_ADDON_TYPE_MAP = Map.of( //
             "automation", new AddonType("automation", "Automation"), //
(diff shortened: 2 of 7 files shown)


clinique:
I wish this will be accepted.

bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/json/JsonMarketplaceAddonService.java
```diff
@@ -0,0 +1,249 @@
...
+
+    @Activate
+    public void activate(Map<String, Object> config) {
+        modified(config);
+    }
+
+    @Modified
+    public void modified(Map<String, Object> config) {
+        String urls = Objects.requireNonNullElse((String) config.get("urls"), "");
+        marketplaceUrls = Arrays.asList(urls.split(","));
```
cweitkamp, line 94: We should use a different separator. A comma can be a valid separator in queries if an add-on provider implements a protocol like OData or similar. What about a pipe `|`?

cweitkamp (commented):
I am fine with this new service. From my pov we should call is just `JsonAddonService` without the phrase "Marketplace".

bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/json/AddonEntryDTO.java
```diff
@@ -0,0 +1,35 @@
...
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.addon.marketplace.internal.json;
```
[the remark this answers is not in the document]
J-N-K: I have moved it to `org.openhab.core.addon.marketplace.internal.json.model` which has the same naming convention used for the DTOs of the community marketplace. That way we stay consistent within the bundle.

bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/json/JsonMarketplaceAddonService.java
```diff
@@ -0,0 +1,249 @@
...
+    private static final String ADDON_ID_PREFIX = "jsonmarketplace:";
+
+    private static final Map<String, AddonType> TAG_ADDON_TYPE_MAP = Map.of( //
+            "automation", new AddonType("automation", "Automation"), //
+            "binding", new AddonType("binding", "Bindings"), //
+            "misc", new AddonType("misc", "Misc"), //
+            "persistence", new AddonType("persistence", "Persistence"), //
+            "transformation", new AddonType("transformation", "Transformations"), //
+            "ui", new AddonType("ui", "User Interfaces"), //
+            "voice", new AddonType("voice", "Voice"));
```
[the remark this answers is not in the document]
J-N-K: Easy, but the question is what the correct place would be. In general an `AddonService` is not required to supply all `AddonType`s, however each `AddonService` could pick what is available. We do not have an `AddonConstants` class or something similar which would be a good fit. Another solution would be to make the map `public` in the `CommunityMarketplaceAddonService` but that would mix different AddonServices which should IMO be avoided.

I hate duplicated code, but IMO here it seems the best solution.

J-N-K:
@cweitkamp Thanks for the review. I have addressed all but one comment (the refactoring of the duplicated map).

bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/json/JsonMarketplaceAddonService.java
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    @Override
+    public List<AddonType> getTypes(Locale locale) {
+        return new ArrayList<>(TAG_ADDON_TYPE_MAP.values());
+    }
+
+    @Override
+    public void install(String id) {
+        Addon addon = getAddon(id, null);
```
cweitkamp, line 167: The `getAddon(String, Locale)` method may return `null`. We should add an additional check here to prevent NPEs.

bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/json/JsonMarketplaceAddonService.java
```diff
@@ -0,0 +1,249 @@
...
+                }
+                return;
+            }
+        }
+        postFailureEvent(id, "Add-on not known.");
+    }
+
+    @Override
+    public void uninstall(String id) {
+        Addon addon = getAddon(id, null);
```
cweitkamp, line 188: The `getAddon(String, Locale)` method may return `null`. We should add an additional check here to prevent NPEs.

bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/json/JsonMarketplaceAddonService.java
J-N-K: I also added that in the `CommunityMarketplaceAddonService`

bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/json/JsonAddonService.java
```diff
@@ -0,0 +1,253 @@
...
+
+/**
+ * This class is a {@link AddonService} retrieving JSON marketplace information.
+ *
+ * @author Yannick Schaus - Initial contribution
+ * @author Jan N. Klug - Refactored for JSON marketplaces
+ */
+@Component(immediate = true, configurationPid = "org.openhab.core.marketplace.json", //
+        property = Constants.SERVICE_PID + "=org.openhab.core.marketplace.json")
+@ConfigurableService(category = "system", label = "JSON 3rd Party Addon Service", description_uri = JsonAddonService.CONFIG_URI)
```
ghys, line 63: From an UX perspective - it's the 3rd configurable service related to add-ons that will appear under "Settings > System Services" with barely any parameter which is beginning to become a lot and quite messy. If anything we should really think about reducing this proliferation and merging related configurations (I've talked about this before).
- Add-on Management + Community Marketplace + JSON 3rd Party Addon Service _(also note the difference in spelling of Add-on/Addon!)_ => Add-on Management
- Regional Settings + Ephemeris => Regional Settings
- Audio + Voice => Audio & Voice
- Maybe also Sitemap + Charts => ?

Maybe have configurable services for the sole purposes of providing the configuration and have the actual services inject them?

J-N-K: I thought about that, too. The problem is that we - at least I know of no way to do it - group in the UI. If we merge everything under e.g. `system:addon` (which in general would be preferred IMO), it's not easily visible to the user which parameter affects which service.

J-N-K: Would it help to introduce another level in der config description URI? E.g. `system:addon:management`, `system:addon:marketplace` and `system:addon:json`? The UI could then group by the second level (if present). 

The good thing to this approach is that it is fully backward compatible since the service PID stays the same and therefore the configuration is also used after the config description URI changed.

cweitkamp (approved):
LGTM. Thanks.

I agree on the grouping of settings in the UI. This PR does not need to wait for a solution for that as we obviously have to change other `ConfigurableService`s too to reach that goal.

PR #2595, opened 2021-12-08
Use the mime types provided by the ScriptEngineFactory

digitaldan:
Use the mime types provided by the ScriptEngineFactory and not the underlying script engines they provide. This allows mutliple versions of script engines to co-exist, like Nashorn and GraalVM .


cweitkamp (commented):
Looks good so far. Thanks.

Too bad we do not have test coverage for this.

digitaldan:
Thanks @cweitkamp with the other PR's merged, this is the last one needed to have both runtime co-exist!

cweitkamp (approved):
LGTM. Thanks.

PR #2596, opened 2021-12-08
[openhab.core] DecimalType-ctor with Number argument

kippAndMost:
The expectation is, that when a `DecimalType` was constructed with a
`float` value, the precision of its `doubleValue()`, `floatValue()`,
`toBigDecimal()` and `toString()` is preserved. But there are `float`
values like `4.2f` or `37.1f` that cannot directly converted to `double`
without precision loss.

This commit replaces all the numerical constructors of `DecimalType`
with a single constructor with a `Number` argument, so that all `float`
values can be used without precision loss.

NOTE: There is some special handling needed for `QuantityType` and
`HSBType` because these types have a special none conventional `toString`
implementation.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DecimalType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DecimalType.java
@@ -41,19 +41,27 @@ public class DecimalType extends Number implements PrimitiveType, State, Command
     protected BigDecimal value;
 
     public DecimalType() {
-        this.value = BigDecimal.ZERO;
+        this(BigDecimal.ZERO);
+    }
+
+    public DecimalType(Number value) {
+        this(bigDecimal(value));
+    }
+
+    private static BigDecimal bigDecimal(Number value) {
+        if (value instanceof QuantityType) {
+            return ((QuantityType<?>) value).toBigDecimal();
+        }
+
+        if (value instanceof HSBType) {
+            return ((HSBType) value).toBigDecimal();
+        }
+
+        return new BigDecimal(value.toString());
     }
 
     public DecimalType(BigDecimal value) {
         this.value = value;
-    }
-
-    public DecimalType(long value) {
-        this.value = BigDecimal.valueOf(value);
-    }
-
-    public DecimalType(double value) {
-        this.value = BigDecimal.valueOf(value);
     }
 
     /**
--- a/bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/DecimalTypeTest.java
+++ b/bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/DecimalTypeTest.java
@@ -13,7 +13,10 @@
 package org.openhab.core.library.types;
 
 import static org.junit.jupiter.api.Assertions.*;
+import static org.junit.jupiter.params.provider.Arguments.arguments;
 
+import java.math.BigDecimal;
+import java.math.BigInteger;
 import java.text.DecimalFormatSymbols;
 import java.util.IllegalFormatConversionException;
 import java.util.Locale;
@@ -22,8 +25,10 @@ import java.util.stream.Stream;
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
 import org.junit.jupiter.params.provider.ValueSource;
+import org.openhab.core.library.unit.Units;
 
 /**
  * @author Thomas Eichstaedt-Engelen - Initial contribution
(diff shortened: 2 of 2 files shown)


cweitkamp:
Thanks for your contribution.

Wouldn't it make more sense to refactor the call to have a generic constructor passing a `Number` as parameter instead of one constructor for each primitive type?

kippAndMost:
Yes, I am with you. That was actually my first thought. That would also be consistent with the implementation of `QuantityType`. Then one would discard all the other constructors which use a number in whatever form. However, I then decided to do the minimum impact possible.

But I would of course welcome the consistent solution with the `Number`-ctor. What would you recommend?

cweitkamp:
I would like to go for the `Number` c-tor.

We will have feature freezer on next Monday ( 13th or December ) thus I would like to reschedule this change to a later time to make sure we do not introduce unwanted side effects. I hope that is okay for you.

kippAndMost:
Yes of course, its OK for me.

I replaced my previous  commit with an appropriate one.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DecimalType.java
```diff
@@ -41,19 +41,27 @@
...
-    public DecimalType(BigDecimal value) {
-        this.value = value;
+    public DecimalType(Number value) {
+        this(bigDecimal(value));
     }
 
-    public DecimalType(long value) {
-        this.value = BigDecimal.valueOf(value);
+    private static BigDecimal bigDecimal(Number value) {
+        if (value.getClass() == QuantityType.class) {
```
cweitkamp, line 52: Why not use `instanceof` comparison?

```suggestion
        if (value instanceof QuantityType) {
```

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DecimalType.java
```diff
@@ -41,19 +41,27 @@
...
+            return ((QuantityType<?>) value).toBigDecimal();
+        }
+
+        if (value.getClass() == HSBType.class) {
```
cweitkamp, line 56: See comment above.

cweitkamp (commented):
Thanks for the changes.

cweitkamp:
@kippAndMost Are you still around to finalize this PR?

kippAndMost:
Hi [cweitkamp](https://github.com/cweitkamp), sorry, I have some work overload at the moment. I will have a look at as soon as possible.

cweitkamp (approved):
LGTM. Thanks.

wborn:
I think this PR has broken the add-ons build, can you have a look at it @kippAndMost?

See: https://ci.openhab.org/job/openHAB-Addons/638/

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:testCompile (default-testCompile) on project org.openhab.binding.upnpcontrol: Compilation failure: Compilation failure:
[ERROR] /home/jenkins/jenkins-agent1/workspace/openHAB-Addons/bundles/org.openhab.binding.upnpcontrol/src/test/java/org/openhab/binding/upnpcontrol/internal/handler/UpnpRendererHandlerTest.java:[925,64] Null type mismatch (type annotations): required 'java.lang.@NonNull Number' but this expression has type 'java.lang.@Nullable Integer'
[ERROR] /home/jenkins/jenkins-agent1/workspace/openHAB-Addons/bundles/org.openhab.binding.upnpcontrol/src/test/java/org/openhab/binding/upnpcontrol/internal/handler/UpnpRendererHandlerTest.java:[926,28] Null type mismatch (type annotations): required 'java.lang.@NonNull Number' but this expression has type 'java.lang.@Nullable Integer'
```

jimtng:
This broke Fronius binding:
```
java.lang.NoSuchMethodError: 'void org.openhab.core.library.types.DecimalType.<init>(double)'
```
A fix for Fronius binding has been submitted in a PR.

A quick search shows this addon file using the same cast to (double) `openhab-addons/bundles/org.openhab.binding.ihc/src/main/java/org/openhab/binding/ihc/internal/handler/IhcHandler.java`

J-N-K:
Looks like this is heavily breaking API
compatibility.

J-N-K:
Did you check if just re-compiling also fixes the problem? I believe this is only incompatible in binary code, not at source level.

cweitkamp:
@J-N-K is right.

We first should fix the build and release a new snapshot before starting to look into every binding. I hope the only think which I missed in this PR is the `DecimalType(BigDecimal)` constructor which in general is superfluous a as `BiggDecimal` is extends `Number`.

J-N-K:
@cweitkamp But maybe keeping the old constructors is also a good idea to maintain binary compatibility. They can all call a single new one. Maybe mark them as `@Deprecated`.

cweitkamp:
I submitted a PR to fix the build https://github.com/openhab/openhab-addons/pull/12374

You got a point. That is a strong argument.

@openhab/core-maintainers Wdyt? Should we reintroduce the removed constructors.

J-N-K:
I think this should be decided before M2.

Hilbrand:
This will actually break compatibility with 3.2 for any binding that is compiled against that 3.3 where `DecimalType` is used in combination with an object instead primitive (for example `Integer`) because that type will compile against `Number` constructor that doesn't exist in 3.2. Meaning anyone updating the binding compiling it with 3.3 and puts it in the market place will have this issue. :disappointed:

PR #2602, opened 2021-12-10
Use Accept-Language header for locale in ConfigurableServiceResource

wborn:
Most RESTResources already use the LocaleService to get the locale based on the Accept-Language header and use the system locale only as fallback.
Because the ConfigurableServiceResource does not do this, it results in mixed up languages whenever your browser locale is not the same as the configured system locale.

See:

* https://community.openhab.org/t/language-support-question-regional-settings-language-vs-browser-language/128010
* https://github.com/openhab/openhab-webui/issues/1083


ghys:
Great to have a solution for this! 👍

kaikreuzer (approved):
lgtm, thanks!

PR #2606, opened 2021-12-11
Change the syntax for references in automation modules configuration

### LINKED ISSUE #2510: Rule Templates reference resolver issues
Reporting 2 problems I identified while trying to write rule templates regarding the [ReferenceResolver](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/util/ReferenceResolver.java)

1. This code that identifies that a module configuration parameter has references in it isn't the best:

https://github.com/openhab/openhab-core/blob/1d17166c79c79dfff5cbb4e96803a7d7d67ccf8f/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/util/ReferenceResolver.java#L246-L254

For instance when the parameter has any closing brace before the first `${` identifying a reference, for example for a script like this:
```js
function {
  ...something...
}

var a = "${parameter}";
```
the `}` that appears on line 3 before the first `${` will lead to no placeholder replacements being made; this is rather confusing.

2. this syntax, `${parameter}`, is exactly the same as used in ES6+ [Template Strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) which will certainly pose problems when modern ECMAScript versions with GraalVM become the norm.

ghys:
Fixes #2510.

This changes the placeholder syntax used by the reference resolver notably
used in rule templates from `${reference}` to `{{reference}}`.

The former syntax is also used by other languages, notably JavaScript's
template strings, and therefore writing rule templates including JavaScript
code with template strings wouldn't behave as expected.

This also fixes a bug identified in #2510 where you couldn't have a
single closing brace before the start of the first placeholder - which
could more easily happen than having two closing braces.

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/util/ReferenceResolver.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/util/ReferenceResolver.java
@@ -43,19 +43,19 @@ import org.slf4j.Logger;
  * <ul>
  * <li>Single reference configuration value where whole configuration property value is replaced(if found) with the
  * referenced value <br/>
- * 'configurationProperty': '${singleReference}'</li>
+ * 'configurationProperty': '{{singleReference}}'</li>
  * <li>Complex reference configuration value where only reference parts are replaced in the whole configuration property
  * value. <br/>
- * 'configurationProperty': '{key1: ${complexReference1}, key2: ${complexReference2}'</li>
+ * 'configurationProperty': '{key1: {{complexReference1}}, key2: {{complexReference2}}}'</li>
  * </ul>
  *
  * Given Module 'A' is child of CompositeModule then its inputs can have '${singleReferences}' to CompositeModule.
  * <ul>
  * <li>Single reference to CompositeModule inputs where whole input value is replaced with the referenced value <br/>
- * 'childInput' : '${compositeModuleInput}'</li>
+ * 'childInput' : '{{compositeModuleInput}}'</li>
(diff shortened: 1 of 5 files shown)


ghys:
I'm aware this breaks existing rule templates... but AFAIK, only me, @kaikreuzer and @rkoshak ever wrote any so I'm sure that we're all willing to update them ;) So better do it now instead of when it's too late to make such a breaking change.

ghys:
The automation integration tests did succeed:
```
[INFO] openHAB Core :: Integration Tests :: Automation Tests SUCCESS [ 42.188 s]
[INFO] openHAB Core :: Integration Tests :: Automation Core Module Tests SUCCESS [ 44.086 s]
[INFO] openHAB Core :: Integration Tests :: Automation Module Timer Tests SUCCESS [ 34.236 s]
[INFO] openHAB Core :: Integration Tests :: Automation Script Module Tests SUCCESS [ 35.765 s]
[INFO] openHAB Core :: Integration Tests :: Automation Integration Tests SUCCESS [ 49.069 s]
```

The Jenkins build failure is unrelated:

```
TEST org.openhab.core.internal.common.SafeCallerImplTest#testAsyncExceedingThreadPoolDifferentIdentifier() <<< ERROR: Duration should have been below 1,800 but was 2,012 ==> expected: <true> but was: <false>
org.opentest4j.AssertionFailedError: Duration should have been below 1,800 but was 2,012 ==> expected: <true> but was: <false>
```

wborn (approved):
Thanks. Better change it now than later!

PR #2612, opened 2021-12-14
SemanticLication: Rename Bathroom to BathRoom

Rosi2143:
To make the naming of bathroom consistent with all other rooms

Signed-off-by: Schrotti <Schrott.Micha@web.de>


digitaldan:
So Bathroom and Bedroom are single words in the english language, hence why they are not called "BedRoom" or "BathRoom" here. There is not a "Livingroom" however in English, it's spelled with two words, hence why it's written as "LivingRoom" in the code.

kaikreuzer:
Let's listen to the native speaker. 😄

PR #2617, opened 2021-12-15
UIRegistry now extensible, like other Registries

### LINKED ISSUE #2513: Partial use of common registry API forced by UIComponentRegistryFactory
Most of openHAB APIs, whether these are things, items or links rely on dedicated `Registry` instance feed by one or more `Provider` services. This design helps to have various sources of core elements managed in various ways.
This is however not reflected within UI registries for pages and widgets since they are wrapped and maintained only via storage.

I believe this is not a major trouble, however current construction does not permit ie. read only UI elements to be present in system. They can't be dynamically computed based on supplied configurations, present items and such.
Changing that would open several fields - ie. where pages could be provided by further system extensions.

jpg0:
This is to allow the UI registry & providers to follow the standard extensible pattern to allow osgi to supply other widgets and pages.

Fixes #2513 

### DIFF
--- a/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentRegistryFactoryImpl.java
+++ b/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentRegistryFactoryImpl.java
@@ -12,34 +12,87 @@
  */
 package org.openhab.core.ui.internal.components;
 
-import java.util.HashMap;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
-import org.openhab.core.storage.StorageService;
+import org.eclipse.jdt.annotation.NonNull;
+import org.openhab.core.common.registry.ManagedProvider;
+import org.openhab.core.ui.components.RootUIComponent;
 import org.openhab.core.ui.components.UIComponentRegistryFactory;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Reference;
+import org.osgi.service.component.annotations.ReferenceCardinality;
+import org.osgi.service.component.annotations.ReferencePolicy;
 
 /**
- * Implementation for a {@link UIComponentRegistryFactory} using a {@link StorageService} and a
- * {@link UIComponentProvider}.
+ * Implementation for a {@link UIComponentRegistryFactory} using a set of {@link UIComponentProvider}.
  *
  * @author Yannick Schaus - Initial contribution
+ * @author Łukasz Dywicki - Removed explicit dependency on storage providers.
+ * @author Jonathan Gilbert - Made providers' collections immutable.
  */
 @Component(service = UIComponentRegistryFactory.class, immediate = true)
 public class UIComponentRegistryFactoryImpl implements UIComponentRegistryFactory {
-    Map<String, UIComponentRegistryImpl> registries = new HashMap<>();
-
-    @Reference
-    StorageService storageService;
+    Map<String, UIComponentRegistryImpl> registries = new ConcurrentHashMap<>();
+    Map<String, Set<UIProvider>> providers = new ConcurrentHashMap<>();
 
     @Override
     public UIComponentRegistryImpl getRegistry(String namespace) {
         UIComponentRegistryImpl registry = registries.get(namespace);
         if (registry == null) {
-            registry = new UIComponentRegistryImpl(namespace, storageService);
+            Set<UIProvider> namespaceProviders = this.providers.get(namespace);
+            ManagedProvider<@NonNull RootUIComponent, @NonNull String> managedProvider = null;
+            if (namespaceProviders != null) {
+                for (UIProvider provider : namespaceProviders) {
+                    if (provider instanceof ManagedProvider) {
+                        managedProvider = (ManagedProvider<@NonNull RootUIComponent, @NonNull String>) provider;
+                        break;
+                    }
+                }
+            }
+            registry = new UIComponentRegistryImpl(namespace, managedProvider, namespaceProviders);
             registries.put(namespace, registry);
         }
         return registry;
+    }
+
+    @Reference(cardinality = ReferenceCardinality.MULTIPLE, policy = ReferencePolicy.DYNAMIC)
+    void addProvider(UIProvider provider) {
+        if (registries.containsKey(provider.getNamespace())) {
+            registries.get(provider.getNamespace()).addProvider(provider);
+        }
+        registerProvider(provider);
+    }
+
+    void removeProvider(UIProvider provider) {
+        if (registries.containsKey(provider.getNamespace())) {
+            registries.get(provider.getNamespace()).removeProvider(provider);
+        }
+        deregisterProvider(provider);
+    }
+
+    private void registerProvider(UIProvider provider) {
+        Set<UIProvider> existing = providers.get(provider.getNamespace());
+
+        if (existing == null) {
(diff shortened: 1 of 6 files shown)


bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentRegistryFactoryImpl.java
```diff
@@ -12,34 +12,87 @@
...
+
+    void removeProvider(UIProvider provider) {
+        if (registries.containsKey(provider.getNamespace())) {
+            registries.get(provider.getNamespace()).removeProvider(provider);
+        }
+        deregisterProvider(provider);
+    }
+
+    private void registerProvider(UIProvider provider) {
+        Set<UIProvider> existing = providers.get(provider.getNamespace());
```
jpg0, line 78: @splatch I modified this section to make the collections immutable, as I was having issues with them not being (not sure why).

splatch: Making final set inserted into map immutable is wrong, cause you can't remove provider later on when it disappear. 

Given that component activation can be made through different threads (ie. declarative services, OSGi activator etc), you probably better of with synchronized collection rather than plain `HashSet`. Keep in mind that set registered for namespace will be manipulated.

splatch:
@jpg0 Since you made modifications to initial code I posted feel free to add yourself to authors :)

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentRegistryFactoryImpl.java
jpg0: Well, whilst the set was immutable, I replaced it with a completely new set when it was modified, so it did still work.

Saying this, I don't really like the verbose code to do this (Java mutable collections I guess). Instead I have now changed it to use a `CopyOnWriteArraySet`, and also synchronized write access (otherwise concurrent updates could lose elements), as it should happen very infrequently anyway.

wborn (dismissed):
Thanks for making it extensible! I added a few minor comments below:

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIProvider.java
```diff
@@ -0,0 +1,27 @@
+/**
+ * Copyright (c) 2010-2021 Contributors to the openHAB project
```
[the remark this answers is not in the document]
splatch: Fair spot, but this is something which isn't really holding up this PR from going in could be addressed by automatic spotless fix once PR is merged (or while it is merged through robotized process). This is a generic point on multiple PRs which suffer from same issue as there is bunch of them for core and addons.
IANAL, but I need note that date in source file header is authoring date which is different for each file. Copyright can be established at time of making the sources, hence in theory it can't be "2010" for file which was made in 2021. It might be also doubtful if changing a comment which does not change contributed works. To make an example - re-printing of an book released in 1999 without changes in 2022 does not extend the copyright as the printing date is just an annotation on the covers, just like source file header.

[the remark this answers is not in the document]
wborn: Sorry I normally don't merge PRs that will break builds.😉

splatch:
> +/**
> + * Copyright (c) 2010-2021 Contributors to the openHAB project
> Sorry I normally don't merge PRs that will break builds.😉
> 
Maybe then its worth to consider changing header template to have `Copyright (c) 2010-\d+`. It will save tens of pull requests stuck for same plain reason. Especially these made in last two months of each year. Unless it is better to comment each of affected files individually?

wborn:
I think it makes perfect sense to have the same headers everywhere.

> Unless it is better to comment each of affected files individually?

That's done automatically with GitHub annotations for new PRs nowadays.

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIProvider.java
```diff
@@ -0,0 +1,27 @@
...
+package org.openhab.core.ui.internal.components;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.openhab.core.common.registry.Provider;
+import org.openhab.core.ui.components.RootUIComponent;
+
+/**
+ * Provides components (pages, widgets, etc.) at runtime.
+ *
+ * @author Łukasz Dywicki - initial controbution.
```
wborn, line 22: Typo:

```suggestion
 * @author Łukasz Dywicki - initial contribution
```

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentRegistryImpl.java
```diff
@@ -23,26 +27,39 @@
  * It is instantiated by the {@link UIComponentRegistryFactoryImpl}.
  *
  * @author Yannick Schaus - Initial contribution
+ * @author Łukasz Dywicki - Support for dynamic registartion of providers.
```
wborn, line 30: Another typo: 

```suggestion
 * @author Łukasz Dywicki - Support for dynamic registration of providers.
```

splatch:
> That's done automatically with GitHub annotations for new PRs nowadays.

Why not making automatic spotless:apply as a part of PR processing? If comments are automated, why fixes aren't?
It would save everybodys time and let people focus on issues which are not handled by tooling.

wborn:
> If comments are automated, why fixes aren't?

It's a nice idea but adding commits to branches causes sync issues and conflicts which then need to be resolved. That's probably why most repos don't do this. I've seen more contributors struggle with Git than with fixing the trivial issues. :wink: Maybe a bot that can fix it on demand would work better like how @dependabot works.

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentRegistryFactoryImpl.java
```diff
@@ -12,34 +12,87 @@
...
     public UIComponentRegistryImpl getRegistry(String namespace) {
         UIComponentRegistryImpl registry = registries.get(namespace);
         if (registry == null) {
-            registry = new UIComponentRegistryImpl(namespace, storageService);
+            Set<UIProvider> namespaceProviders = this.providers.get(namespace);
+            ManagedProvider<@NonNull RootUIComponent, @NonNull String> managedProvider = null;
+            if (namespaceProviders != null) {
+                for (UIProvider provider : namespaceProviders) {
+                    if (provider instanceof ManagedProvider) {
+                        managedProvider = (ManagedProvider<@NonNull RootUIComponent, @NonNull String>) provider;
```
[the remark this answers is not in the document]
J-N-K: Yes they are, because otherwise the null constraints are not fulfilled. The reason is that the whole class is not null-annotated. Since the other issues are only typos / year correction and we are allowed to edit the PR, are you ok with merging after the suggestions have been committed?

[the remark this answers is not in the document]
wborn: OK I've added the annotation on the class and done some more trivial cleanup/fixes.
LGTM now, so if you are happy with it as well, feel free to merge it. :slightly_smiling_face:

J-N-K (approved):

J-N-K:
Thanks!

PR #2620, opened 2021-12-17
Use static BigDecimal values

cweitkamp:
- Use static BigDecimal values


wborn (approved):
Thanks!

PR #2622, opened 2021-12-17
[tls] Add a PEMTrustManager to deal with different PEM files (e.g. self-signed or global CA certificates)

cweitkamp:
- Added a PEMTrustManager to deal with different PEM files (e.g. self-signed or global CA certificates)

### DIFF
--- a/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/PEMTrustManager.java
+++ b/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/PEMTrustManager.java
@@ -1,0 +1,224 @@
+/**
+ * Copyright (c) 2010-2021 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.io.net.http;
+
+import java.io.ByteArrayInputStream;
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.MalformedURLException;
+import java.net.Socket;
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.security.KeyManagementException;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+import java.security.cert.Certificate;
+import java.security.cert.CertificateException;
+import java.security.cert.CertificateFactory;
+import java.security.cert.CertificateParsingException;
+import java.security.cert.X509Certificate;
+import java.util.Base64;
+
+import javax.net.ssl.HttpsURLConnection;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLEngine;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.X509ExtendedTrustManager;
+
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/PEMTrustManager.java
```diff
@@ -0,0 +1,224 @@
...
+    /**
+     * Creates a {@link PEMTrustManager} instance by passing the PEM certificate as {@link String}.
+     * The PEM format typically starts with <code>"-----BEGIN CERTIFICATE-----"</code> and ends with
+     * <code>"-----END CERTIFICATE-----"</code>. The base 64 encoded certificate information are placed in between.
+     *
+     * @param pemCert the PEM certificate
+     * @throws CertificateException
+     */
+    public PEMTrustManager(String pemCert) throws CertificateException {
+        if (!pemCert.isBlank() && pemCert.startsWith(BEGIN_CERT)) {
```
J-N-K, line 69: One of the use-cases (e.g. with the http binding) could be that there is a configuration field in the thing config that allows pinning the certificate permanently (the `getInstanceFromServer` does only pin until restarting the system). 

Using `line.separator` would require the configuration  to be different depending on the operating system that openHAB runs on and could lead to problems if the user configures on a different OS. Maybe the cert string should be normalized to the `line.separator` line endings before the checks are done.

bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/PEMTrustManager.java
```diff
@@ -0,0 +1,224 @@
...
+    public static PEMTrustManager getInstanceFromFile(String path) throws FileNotFoundException, CertificateException {
+        String pemCert = readPEMCertificateStringFromFile(path);
+        if (pemCert != null) {
+            return new PEMTrustManager(pemCert);
+        }
+        throw new CertificateInstantiationException(String.format("Unable to read certificate from file: %s", path));
+    }
+
+    /**
+     * Creates a {@link PEMTrustManager} instance by downloading the PEM certificate from the given file.
```
J-N-K, line 98:
```suggestion
     * Creates a {@link PEMTrustManager} instance by downloading the PEM certificate from the given server.
```

bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/PEMTrustManager.java
```diff
@@ -0,0 +1,224 @@
...
+        }
+        return null;
+    }
+
+    private static @Nullable String readPEMCertificateStringFromFile(String path) throws FileNotFoundException {
+        File certFile = new File(path);
+        if (certFile.exists()) {
+            try {
+                String cert = new String(Files.readAllBytes(certFile.toPath()), StandardCharsets.UTF_8);
+                if (cert.startsWith(BEGIN_CERT)) {
```
J-N-K, line 195: You check the same at the beginning of the constructor, isn't that enough?

J-N-K:
This is IMO one of the most useful extensions in the last years, it allows the remove the `TrustAllTrustManager` and similar implementations in quite a bunch of addons.

cweitkamp (commented):
Good to see that someone finds it very useful.

bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/PEMTrustManager.java
cweitkamp: Can you check if the new version handles the line endings even better?

kaikreuzer (approved):
lgtm, thank you!
Let's wait for @J-N-K final comments before merging.

J-N-K:
I'll check today and report back this evening.

J-N-K (approved):
LGTM

PR #2605, opened 2021-12-11
Added nullness annotations to Addon related classes

cweitkamp:
- Added nullness annotations to Addon related classes (except for the `Addon` itself)
- Added translations for marketplace bundles
- add resources to classpath

Signed-off-by: Christoph Weitkamp <github@christophweitkamp.de>

PR #2626, opened 2021-12-18
[addonservices] fix JsonAddonService

J-N-K:
This fixes a regression introduced by #2605 which results in addon installation exiting early if the first `MarketplaceAddonHandler` does not support the addon.


bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/json/JsonAddonService.java
```diff
@@ -181,7 +181,6 @@ public void install(String id) {
                     }
                     return;
                 }
-                return;
```
cweitkamp, line 184: What about line 207 in `uninstall`?

cweitkamp (approved):
LGTM. Thanks.

This can be handled as a critical bugfix. The code will work as it now is similar to the `CommunityMarketplaceAddonService`.

PR #2629, opened 2021-12-19
Improve logic to choose a script engine preferred MIME type

ghys:
As discussed in https://github.com/openhab/openhab-addons/issues/11818.
This should select `application/x-python` for Jython (instead of `application/python` currently)
and allow other script engines to advertise their MIME type:
- JRuby: `application/x-ruby`
- Groovy: `application/x-groovy`


kaikreuzer:
I tried to test it in my IDE, but for some reason cannot resolve the scripting addons...
So let's merge untested and let us check the distro build briefly afterwards!

PR #2607, opened 2021-12-11
[addonservices] allow uninstalling of removed addons and fix other issues

J-N-K:
Fixes #2576
Fixes #2581 
Fixes #2673 
Fixes #2687 

It stores a String representation of the addon in the storage service on installation and removes it when the addon is uninstalled. When compiling the list of available addons the information in the storage takes precedence over the information received from the remote so the displayed information (including version or links) stays correct for the installed addon. 

@wborn WDYT? I can extend that for the other Addon-Service very quickly, so we can ship that with 3.2.0, which would be very beneficial for users.

Signed-off-by: Jan N. Klug <jan.n.klug@rub.de>

Issue #2581, opened 2021-11-26
[addonservices] wrong implementation of AddonService


J-N-K:
The JavaDoc of the `AddonService` states that  `getAddons(Locale locale)` should be cheap to call and return quickly. The implementation does remote HTTP requests instead of the suggested caching which could take some time. The `refreshSource()` method would be the correct place for updating the cache but is not called before addons are displayed in the UI.

J-N-K:
@cweitkamp The `refreshSource` method is never called. When a request is made to the REST API to get all addons, only `getAddons` is called. So another solution than the one you mentioned in #2633 would be to use a timestamp for the last refresh and then call `refreshSource` after some time.

Please do not do any refactoring to the `AddonService` implementations as that would probably result in conflicts with #2607.

cweitkamp:
I implemented a draft and created a PR against your branch used in #2607.

J-N-K:
I thought about something like that before, but it does not solve the problem, it just mitigates it. Every time the cache is refreshed the `getAddons` method is not returning fast. IMO this needs to be solved at the place where `getAddons` is called:

https://github.com/openhab/openhab-core/blob/cabb3f73150d13f44d647f46af8362e396684b92/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/addons/AddonResource.java#L126-L145

To comply with the documentation, `getAddons` should never refresh but only return the cache. The caller should make sure that `refreshSource` is called when needed. Otherwise the `refreshSource` method as part of the API is not necessary:

https://github.com/openhab/openhab-core/blob/cabb3f73150d13f44d647f46af8362e396684b92/bundles/org.openhab.core/src/main/java/org/openhab/core/addon/AddonService.java#L47-L53

Caching is not needed for `getAddon` (and is also not documented in JavaDoc there), because it doesn't happen very often (only once per add-on, when that add-on is selected in the UI, so it is automatically limited by user-interaction).

cweitkamp:
I agree. There we have to think about the implementation too. Should the `AddonResource` call `refreshSource` first before retrieving all add-ons via `getAddons`?
I don't think that a scheduled polling for `AddonService`s is an option.

The code I submitted to your branch reduces potential load on the target source. This might be considered too as there are already complaints about in the community forum.

J-N-K:
[...]

PR #2633, opened 2021-12-20
[addonservices] allow offline mode

### LINKED ISSUE #2628: [marketplace] Expecting no access to the marketplace when remote is set to false
For users who decided to rely only on an offline distribution and so who setup openHAB like this:
```

# Access Remote Add-on Repository
# Defines whether the remote openHAB add-on repository should be used for browsing and installing add-ons. (default is true)
#
remote = false
```

I assume MainUI should not show marketplace in this case but only bindings present in the kar file.

The current behaviour is to have bindings from the marketplace displayed in UI.

J-N-K:
Fixes #2628 

### DIFF
--- a/bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityMarketplaceAddonService.java
+++ b/bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityMarketplaceAddonService.java
@@ -14,6 +14,7 @@ package org.openhab.core.addon.marketplace.internal.community;
 
 import static org.openhab.core.addon.Addon.CODE_MATURITY_LEVELS;
 
+import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.Reader;
 import java.net.URI;
@@ -50,6 +51,8 @@ import org.openhab.core.config.core.ConfigurableService;
 import org.openhab.core.events.Event;
 import org.openhab.core.events.EventPublisher;
 import org.osgi.framework.Constants;
+import org.osgi.service.cm.Configuration;
+import org.osgi.service.cm.ConfigurationAdmin;
 import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Modified;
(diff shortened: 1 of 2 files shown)


J-N-K:
Since there is a lot of common code between both addon services, it might be a good idea to refactor them to inherit from an `AbstractAddonService`. But this is out of scope here and should probably be done after a solution for #2607 is found.

J-N-K:
Test fails unrelated.

cweitkamp (commented):
Thank you for looking into this. From my pov we now reached some point where we need unit tests to cover the functionality of those addon services. Wdyt?
If you like I can give you a hand on that. Just ping me.

And I agree. The code can be reduced by introducing an abstract base class.

J-N-K:
Regarding the unit tests: I agree. Unfortunately it's not that easy to mock the remote end (and it's also different for both add-on services. 

This will improve with the introduction of the `AbstractAddonService` as a `TestAddonService` could at least test the common parts. 

IMO the bundle installers also need test coverage. I can have a look in a separate PR.

cweitkamp (approved):
LGTM. Thanks.

🎆 Happy New Year.

bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityMarketplaceAddonService.java
```diff
@@ -460,4 +475,13 @@ private void postFailureEvent(String extensionId, @Nullable String msg) {
         Event event = AddonEventFactory.createAddonFailureEvent(extensionId, msg);
         eventPublisher.post(event);
     }
+
+    private boolean remoteEnabled() {
+        try {
+            Configuration configuration = configurationAdmin.getConfiguration("org.openhab.addons", null);
+            return (boolean) Objects.requireNonNullElse(configuration.getProperties().get("remote"), true);
```
cweitkamp, line 482: It looks.like the [getProperties()](https://docs.osgi.org/javadoc/osgi.cmpn/7.0.0/org/osgi/service/cm/Configuration.html#getProperties--) can return `null` and we are potentially run into NPEs here.

J-N-K: Fixed in https://github.com/openhab/openhab-core/pull/2607/commits/fc25ddbf41f4671b165ec795d537e5a6adb1e35a

cweitkamp: Thank you.

I will try to find a solution for #2581 later.

PR #2635, opened 2021-12-21
[discovery.mdns] Devices may apply a grace period for removal from the Inbox

### LINKED ISSUE #1835: [MDNSDiscoveryService] potential bug deleting items from Inbox??
I have been developing a binding that implements a MDNSDiscoveryParticipant to discover the bridge. One user reports that every time the participant add such a bridge to the Inbox, the core then deletes it again. I think this could be a bug in the MDNSDiscoveryService class..

- the method `removeOlderResults()` method seems to be called, but it is not clear why.
- and also `timestampOfLastScan` seems not to be being initialized.

```
public class MDNSDiscoveryService extends AbstractDiscoveryService implements ServiceListener {
..
    protected synchronized void stopScan() {
        removeOlderResults(getTimestampOfLastScan());
        super.stopScan();
    }
```

andrewfg:
**BACKGROUND**

Some types of OpenHAB bindings have devices that can sometimes be a bit late in sending their mDNS renewal announcements, which means that such devices are repeatedly removed from, and (re)added to, the InBox. This leads to confusion in the UI, and repeated logger messages. This is an exact analog to the UPNP Discovery issue that was solved by https://github.com/openhab/openhab-core/pull/2144

**SOLUTION**

The solution is an exact analog to the solution in https://github.com/openhab/openhab-core/pull/2144
Resolves #1835

**DOCUMENTATION**

The binding developer guide will also be updated via this PR https://github.com/openhab/openhab-docs/pull/1708

Signed-off-by: Andrew Fiddian-Green software@whitebear.ch

### DIFF
--- a/bundles/org.openhab.core.config.discovery.mdns/src/main/java/org/openhab/core/config/discovery/mdns/internal/MDNSDiscoveryService.java
+++ b/bundles/org.openhab.core.config.discovery.mdns/src/main/java/org/openhab/core/config/discovery/mdns/internal/MDNSDiscoveryService.java
@@ -66,7 +66,7 @@ public class MDNSDiscoveryService extends AbstractDiscoveryService implements Se
 
     private final MDNSClient mdnsClient;
 
-    /*
+    /**
      * Map of scheduled tasks to remove devices from the Inbox.
      */
     private Map<String, ScheduledFuture<?>> deviceRemovalTasks = new ConcurrentHashMap<>();
@@ -238,12 +238,9 @@ public class MDNSDiscoveryService extends AbstractDiscoveryService implements Se
                     try {
                         DiscoveryResult result = participant.createResult(serviceEvent.getInfo());
                         if (result != null) {
+                            cancelRemovalTask(serviceEvent.getInfo());
                             final DiscoveryResult resultNew = getLocalizedDiscoveryResult(result,
                                     FrameworkUtil.getBundle(participant.getClass()));
-                            final ServiceInfo service = serviceEvent.getInfo();
-                            if (participant.getRemovalGracePeriodSeconds(service) > 0) {
-                                cancelRemovalTask(service);
-                            }
                             thingDiscovered(resultNew);
                         }
                     } catch (Exception e) {
@@ -254,11 +251,11 @@ public class MDNSDiscoveryService extends AbstractDiscoveryService implements Se
         }
     }
 
-    /*
+    /**
      * If the device has been scheduled to be removed, cancel its respective removal task.
      */
-    private void cancelRemovalTask(ServiceInfo service) {
-        ScheduledFuture<?> deviceRemovalTask = deviceRemovalTasks.remove(service.getQualifiedName());
+    private void cancelRemovalTask(ServiceInfo serviceInfo) {
+        ScheduledFuture<?> deviceRemovalTask = deviceRemovalTasks.remove(serviceInfo.getQualifiedName());
         if (deviceRemovalTask != null) {
             deviceRemovalTask.cancel(false);
         }
--- a/bundles/org.openhab.core.config.discovery.mdns/src/main/java/org/openhab/core/config/discovery/mdns/internal/MDNSDiscoveryService.java
+++ b/bundles/org.openhab.core.config.discovery.mdns/src/main/java/org/openhab/core/config/discovery/mdns/internal/MDNSDiscoveryService.java
@@ -67,7 +67,7 @@ public class MDNSDiscoveryService extends AbstractDiscoveryService implements Se
     private final MDNSClient mdnsClient;
 
     /*
-     * Map of scheduled tasks to remove devices from the Inbox
+     * Map of scheduled tasks to remove devices from the Inbox.
      */
     private Map<String, ScheduledFuture<?>> deviceRemovalTasks = new ConcurrentHashMap<>();
 
@@ -255,7 +255,7 @@ public class MDNSDiscoveryService extends AbstractDiscoveryService implements Se
     }
 
     /*
-     * If the device has been scheduled to be removed, cancel its respective removal task
+     * If the device has been scheduled to be removed, cancel its respective removal task.
      */
     private void cancelRemovalTask(ServiceInfo service) {
         ScheduledFuture<?> deviceRemovalTask = deviceRemovalTasks.remove(service.getQualifiedName());
--- a/bundles/org.openhab.core.config.discovery.mdns/src/main/java/org/openhab/core/config/discovery/mdns/internal/MDNSDiscoveryService.java
+++ b/bundles/org.openhab.core.config.discovery.mdns/src/main/java/org/openhab/core/config/discovery/mdns/internal/MDNSDiscoveryService.java
@@ -207,16 +207,13 @@ public class MDNSDiscoveryService extends AbstractDiscoveryService implements Se
                 try {
                     ThingUID thingUID = participant.getThingUID(serviceEvent.getInfo());
                     if (thingUID != null) {
-                        ServiceInfo service = serviceEvent.getInfo();
-                        long gracePeriod = participant.getRemovalGracePeriodSeconds(service);
+                        ServiceInfo serviceInfo = serviceEvent.getInfo();
+                        long gracePeriod = participant.getRemovalGracePeriodSeconds(serviceInfo);
                         if (gracePeriod <= 0) {
                             thingRemoved(thingUID);
                         } else {
-                            cancelRemovalTask(service);
-                            deviceRemovalTasks.put(service.getQualifiedName(), scheduler.schedule(() -> {
-                                thingRemoved(thingUID);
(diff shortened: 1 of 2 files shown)


andrewfg:
Resolves https://github.com/openhab/openhab-core/issues/1835 in general.
And in combination https://github.com/openhab/openhab-addons/pull/11834 it resolves the specific Miele issue.

jlaur:
@andrewfg - link #1835 to this PR?

andrewfg:
^
Yes. See my post above yours.

jlaur:
> ^ Yes. See my post above yours.

Ah, well, I meant reference it here to be able to directly cross-navigate and have issue auto-closed when PR is merged:

image

andrewfg:
AFAIK the keyword “Resolves” (or “Fixes”, “Solves” etc.) followed by the issue number should indeed cause Git to auto close the issue.

jlaur:
> AFAIK the keyword “Resolves” (or “Fixes”, “Solves” etc.) followed by the issue number should indeed cause Git to auto close the issue.

Yes, if you write it like "Fixes #1835" in the PR description, it will be linked correctly.

bundles/org.openhab.core.config.discovery.mdns/src/main/java/org/openhab/core/config/discovery/mdns/internal/MDNSDiscoveryService.java
```diff
@@ -223,6 +240,10 @@ private void considerService(ServiceEvent serviceEvent) {
                         if (result != null) {
                             final DiscoveryResult resultNew = getLocalizedDiscoveryResult(result,
                                     FrameworkUtil.getBundle(participant.getClass()));
+                            final ServiceInfo service = serviceEvent.getInfo();
+                            if (participant.getRemovalGracePeriodSeconds(service) > 0) {
+                                cancelRemovalTask(service);
```
jlaur, line 245: Maybe this should be done unconditionally? If grace period in binding has changed to 0 since the task was scheduled, this might prevent the removal in case of any exception thrown.

bundles/org.openhab.core.config.discovery.mdns/src/main/java/org/openhab/core/config/discovery/mdns/internal/MDNSDiscoveryService.java
```diff
@@ -223,6 +240,10 @@ private void considerService(ServiceEvent serviceEvent) {
                         if (result != null) {
                             final DiscoveryResult resultNew = getLocalizedDiscoveryResult(result,
                                     FrameworkUtil.getBundle(participant.getClass()));
+                            final ServiceInfo service = serviceEvent.getInfo();
```
jlaur, line 243: **Naming:** We have `ServiceEvent` and `ServiceInfo`. To avoid confusion, I would suggest to stick to this naming for both variables, i.e. `serviceEvent` (already named that) and `serviceInfo`, respectively.

bundles/org.openhab.core.config.discovery.mdns/src/main/java/org/openhab/core/config/discovery/mdns/internal/MDNSDiscoveryService.java
```diff
@@ -232,4 +253,14 @@ private void considerService(ServiceEvent serviceEvent) {
             }
         }
     }
+
+    /*
```
jlaur, line 257: javadoc missing character:
```suggestion
    /**
```

bundles/org.openhab.core.config.discovery.mdns/src/main/java/org/openhab/core/config/discovery/mdns/internal/MDNSDiscoveryService.java
```diff
@@ -64,6 +66,11 @@
 
     private final MDNSClient mdnsClient;
 
+    /*
```
jlaur, line 69: javadoc missing character:
```suggestion
    /**
```

bundles/org.openhab.core.config.discovery.mdns/src/main/java/org/openhab/core/config/discovery/mdns/internal/MDNSDiscoveryService.java
andrewfg: I don't think so. Most bindings (except the Miele one) will not have overridden the getRemoval method, so the default method is called and returns 0. So on those most bindings, we definitely must not differ from prior behavior. I don't want to break any other bindings.

jlaur: No harm done trying to remove it from the ConcurrentHashMap if it doesn't exist:
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#remove-java.lang.Object-

Of course it doesn't change anything as long as no bindings supports this method. The PR itself doesn't change anything in this case. But when bindings will start supporting this, we can run into this situation where grace period will be changed from value >0 to 0 while having a task running. So I think this problem can be easily avoided by removing the condition?

andrewfg: Ok.

bundles/org.openhab.core.config.discovery.mdns/src/main/java/org/openhab/core/config/discovery/mdns/internal/MDNSDiscoveryService.java
andrewfg: Ok.

bundles/org.openhab.core.config.discovery.mdns/src/main/java/org/openhab/core/config/discovery/mdns/internal/MDNSDiscoveryService.java
andrewfg: Ok.

bundles/org.openhab.core.config.discovery.mdns/src/main/java/org/openhab/core/config/discovery/mdns/internal/MDNSDiscoveryService.java
andrewfg: Ok.

bundles/org.openhab.core.config.discovery.mdns/src/main/java/org/openhab/core/config/discovery/mdns/internal/MDNSDiscoveryService.java
```diff
@@ -200,7 +207,17 @@ public void serviceRemoved(@NonNullByDefault({}) ServiceEvent serviceEvent) {
...
+                        ServiceInfo service = serviceEvent.getInfo();
+                        long gracePeriod = participant.getRemovalGracePeriodSeconds(service);
+                        if (gracePeriod <= 0) {
+                            thingRemoved(thingUID);
+                        } else {
+                            cancelRemovalTask(service);
+                            deviceRemovalTasks.put(service.getQualifiedName(), scheduler.schedule(() -> {
+                                thingRemoved(thingUID);
+                                cancelRemovalTask(service);
+                            }, gracePeriod, TimeUnit.SECONDS));
```
jlaur, line 219: Sorry for late comment, didn't catch in first round: consider extracting this to a private method (e.g. `scheduleRemovalTask`), so you would have one method for scheduling task and another method for cancelling task. Even though it would only have one usage, it should still simplify this method by making it slimmer, thus improve readability.

andrewfg: Ok.

jlaur:
I have tested this PR in combination with openhab/openhab-addons#11834 and it works as expected - no more constant removal and readding of the Miele XGW 3000 gateway. From code perspective it also looks fine to me. Thanks again for this fix @andrewfg. Now will leave it to core maintainers to review.

kaikreuzer (approved):
lgtm, thanks!

cweitkamp:
I am afraid we missed one place in the manual scan where we have to cancel the removal task.

https://github.com/openhab/openhab-core/blob/670c64b5c441d3a0408cd55927397c3a616251af/bundles/org.openhab.core.config.discovery.mdns/src/main/java/org/openhab/core/config/discovery/mdns/internal/MDNSDiscoveryService.java#L152-L172

I will provide a follow-up PR with a fix and some minor code refactoring.

andrewfg:
Ok. Thanks.

cweitkamp:
See https://github.com/openhab/openhab-core/pull/2647

PR #2650, opened 2021-12-28
[audio] Fix hardcoded value for WAV file

dalgwen:
This pull request aims to replace hardcoded value by values extracted from the wav file itself, for the System speaker audio sink to play correctly some files.

Things I'm not sure about : 
- I added a new AudioFormat constructor, instead of adding a parameter to the already existing one, because I'm not confortable with introducing a breaking change.
- The values I compute (bitrate, framerate and framesize). I'm not proficient in audio, and honestly I don't really know what I'm doing. But it works when I test it. If someone using other audio sinks (compatible with WAV) could test it would be great.

### DIFF
--- a/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioFormat.java
+++ b/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioFormat.java
@@ -147,6 +147,11 @@ public class AudioFormat {
     private final @Nullable Long frequency;
 
     /**
+     * Channels number
+     */
+    private final @Nullable Integer channels;
+
+    /**
      * Constructs an instance with the specified properties.
      *
      * Note that any properties that are null indicate that
@@ -173,6 +178,39 @@ public class AudioFormat {
         this.bitDepth = bitDepth;
         this.bitRate = bitRate;
         this.frequency = frequency;
+        this.channels = 1;
+    }
+
+    /**
+     * Constructs an instance with the specified properties.
+     *
+     * Note that any properties that are null indicate that
+     * the corresponding AudioFormat allows any value for
+     * the property.
+     *
+     * Concretely this implies that if, for example, one
+     * passed null for the value of frequency, this would
+     * mean the created AudioFormat allowed for any valid
+     * frequency.
+     *
+     * @param container The container for the audio
+     * @param codec The audio codec
+     * @param bigEndian If the audo data is big endian
+     * @param bitDepth The bit depth of the audo data
+     * @param bitRate The bit rate of the audio
+     * @param frequency The frequency at which the audio was sampled
+     * @param channels The number of channels
+     */
+    public AudioFormat(@Nullable String container, @Nullable String codec, @Nullable Boolean bigEndian,
+            @Nullable Integer bitDepth, @Nullable Integer bitRate, @Nullable Long frequency,
+            @Nullable Integer channels) {
+        this.container = container;
+        this.codec = codec;
+        this.bigEndian = bigEndian;
+        this.bitDepth = bitDepth;
+        this.bitRate = bitRate;
+        this.frequency = frequency;
+        this.channels = channels;
     }
 
     /**
@@ -229,6 +267,15 @@ public class AudioFormat {
      */
     public @Nullable Long getFrequency() {
         return frequency;
+    }
+
+    /**
+     * Gets channel number
+     *
+     * @return The frequency
+     */
+    public @Nullable Integer getChannels() {
+        return channels;
     }
 
     /**
@@ -345,10 +392,12 @@ public class AudioFormat {
                 continue;
             }
 
+            Integer channel = format.getChannels() == null ? Integer.valueOf(1) : format.getChannels();
+
             // If required set BigEndian, BitDepth, BitRate, and Frequency to default values
             if (null == format.isBigEndian()) {
                 format = new AudioFormat(format.getContainer(), format.getCodec(), Boolean.TRUE, format.getBitDepth(),
-                        format.getBitRate(), format.getFrequency());
+                        format.getBitRate(), format.getFrequency(), channel);
             }
             if (null == format.getBitDepth() || null == format.getBitRate() || null == format.getFrequency()) {
                 // Define default values
@@ -379,7 +428,7 @@ public class AudioFormat {
                 }
 
                 format = new AudioFormat(format.getContainer(), format.getCodec(), format.isBigEndian(), bitDepth,
-                        bitRate, frequency);
+                        bitRate, frequency, channel);
             }
 
             // Return preferred AudioFormat
@@ -414,7 +463,9 @@ public class AudioFormat {
                     : getFrequency().equals(format.getFrequency()))) {
                 return false;
             }
-            return true;
+            if (!(null == getChannels() ? null == format.getChannels() : getChannels().equals(format.getChannels()))) {
+                return false;
+            }
         }
         return super.equals(obj);
     }
@@ -429,6 +480,7 @@ public class AudioFormat {
         result = prime * result + ((codec == null) ? 0 : codec.hashCode());
         result = prime * result + ((container == null) ? 0 : container.hashCode());
         result = prime * result + ((frequency == null) ? 0 : frequency.hashCode());
+        result = prime * result + ((channels == null) ? 0 : channels.hashCode());
         return result;
     }
 
@@ -439,6 +491,7 @@ public class AudioFormat {
                 + (bigEndian != null ? "bigEndian=" + bigEndian + ", " : "")
                 + (bitDepth != null ? "bitDepth=" + bitDepth + ", " : "")
                 + (bitRate != null ? "bitRate=" + bitRate + ", " : "")
-                + (frequency != null ? "frequency=" + frequency : "") + "]";
+                + (frequency != null ? "frequency=" + frequency : "") + (channels != null ? "channels=" + channels : "")
+                + "]";
     }
 }
--- a/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioFormat.java
+++ b/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioFormat.java
@@ -149,7 +149,7 @@ public class AudioFormat {
     /**
      * Channels number
      */
-    private @Nullable Integer channels;
+    private final @Nullable Integer channels;
 
     /**
      * Constructs an instance with the specified properties.
@@ -172,13 +172,7 @@ public class AudioFormat {
      */
     public AudioFormat(@Nullable String container, @Nullable String codec, @Nullable Boolean bigEndian,
             @Nullable Integer bitDepth, @Nullable Integer bitRate, @Nullable Long frequency) {
-        this.container = container;
-        this.codec = codec;
-        this.bigEndian = bigEndian;
-        this.bitDepth = bitDepth;
-        this.bitRate = bitRate;
-        this.frequency = frequency;
-        this.channels = 1;
+        this(container, codec, bigEndian, bitDepth, bitRate, frequency, 1);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/FileAudioStream.java
```diff
@@ -70,6 +75,36 @@ private static AudioFormat getAudioFormat(File file) throws AudioException {
...
+                codecPCMSignedOrUnsigned = null;
+            }
+            Integer bitRate = Math.round(format.getFrameRate() * format.getFrameSize()) * format.getChannels();
+            Long frequency = Float.valueOf(format.getSampleRate()).longValue();
+            return new AudioFormat(AudioFormat.CONTAINER_WAVE, codecPCMSignedOrUnsigned, format.isBigEndian(),
+                    format.getSampleSizeInBits(), bitRate, frequency, format.getChannels());
+
+        } catch (UnsupportedAudioFileException | IOException e) {
+            // assume default format
+            return new AudioFormat(AudioFormat.CONTAINER_WAVE, AudioFormat.CODEC_PCM_SIGNED, false, 16, 705600, 44100L,
```
cweitkamp, line 103: Can we refactor this to be a const?

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/FileAudioStream.java
```diff
@@ -70,6 +75,36 @@ private static AudioFormat getAudioFormat(File file) throws AudioException {
...
+                codecPCMSignedOrUnsigned = AudioFormat.CODEC_PCM_SIGNED;
+            } else if (javaSoundencoding.equals(Encoding.PCM_UNSIGNED.toString())) {
+                codecPCMSignedOrUnsigned = AudioFormat.CODEC_PCM_UNSIGNED;
+            } else if (javaSoundencoding.equals(Encoding.ULAW.toString())) {
+                codecPCMSignedOrUnsigned = AudioFormat.CODEC_PCM_ULAW;
+            } else if (javaSoundencoding.equals(Encoding.ALAW.toString())) {
+                codecPCMSignedOrUnsigned = AudioFormat.CODEC_PCM_ALAW;
+            } else {
+                codecPCMSignedOrUnsigned = null;
+            }
```
cweitkamp, line 95: Can you use a `switch-case-statement` for this block?

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioFormat.java
```diff
@@ -231,6 +269,15 @@ public AudioFormat(@Nullable String container, @Nullable String codec, @Nullable
         return frequency;
     }
 
+    /**
+     * Gets channel number
+     *
+     * @return The frequency
```
cweitkamp, line 275:
```suggestion
     * @return The number of channels
```

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/javasound/AudioPlayer.java
```diff
@@ -136,11 +136,12 @@ public void run() {
         }
         final int sampleSizeInBits = bitDepth.intValue();
 
-        final int channels = 1;
+        final int channels = audioFormat.getChannels() == null ? Integer.valueOf(1) : audioFormat.getChannels();
+        ;
```
cweitkamp, line 140: Please remove this empty line.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioFormat.java
```diff
@@ -173,6 +178,39 @@ public AudioFormat(@Nullable String container, @Nullable String codec, @Nullable
         this.bitDepth = bitDepth;
         this.bitRate = bitRate;
         this.frequency = frequency;
+        this.channels = 1;
```
cweitkamp, line 181: When defining a new constructor you should reuse it here by calling `super(...)` and passing all given parameters to it.

cweitkamp (commented):
Thanks for looking into this. I left some comments.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/FileAudioStream.java
```diff
@@ -70,6 +75,36 @@ private static AudioFormat getAudioFormat(File file) throws AudioException {
         }
     }
 
+    private static AudioFormat parseWavFormat(File file) {
```
cweitkamp, line 78: It looks like this method is useful for `ByteArrayAudioStream` too (see https://github.com/openhab/openhab-addons/pull/11877). Maybe we should move it to a different class to be available as util method.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/FileAudioStream.java
dalgwen: I don't think so ?
I already tried, but the value "javaSoundencoding.equals(Encoding.SOMETHING.toString()" are not constant and the compilation failed because you can only switch to constant value.
Unless I miss something.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/FileAudioStream.java
dalgwen: Yes, good idea. Done (in the util class)

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioFormat.java
dalgwen: Done

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/javasound/AudioPlayer.java
dalgwen: Oops. Done

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioFormat.java
dalgwen: you mean this(...) and not super(...) ?
If so, then ok, done.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/FileAudioStream.java
dalgwen: Yes, good idea. Done

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioWaveUtils.java
```diff
@@ -0,0 +1,69 @@
...
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.audio;
```
cweitkamp, line 13: Can you move this calls to the `org.openhab.core.audio.utils` package.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioWaveUtils.java
```diff
@@ -0,0 +1,69 @@
...
+import javax.sound.sampled.AudioSystem;
+import javax.sound.sampled.UnsupportedAudioFileException;
+
+/**
+ * Some utility methods for parsing and cleaning wav files
+ *
+ * @author Gwendal Roulleau - Initial contribution
+ *
+ */
+public class AudioWaveUtils {
```
cweitkamp, line 28: Please add [Null Annotations](https://www.openhab.org/docs/developer/guidelines.html#null-annotations).

```suggestion
@NonNullByDefault
public class AudioWaveUtils {
```

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioWaveUtils.java
```diff
@@ -0,0 +1,69 @@
...
+                codecPCMSignedOrUnsigned = AudioFormat.CODEC_PCM_SIGNED;
+            } else if (javaSoundencoding.equals(Encoding.PCM_UNSIGNED.toString())) {
+                codecPCMSignedOrUnsigned = AudioFormat.CODEC_PCM_UNSIGNED;
+            } else if (javaSoundencoding.equals(Encoding.ULAW.toString())) {
+                codecPCMSignedOrUnsigned = AudioFormat.CODEC_PCM_ULAW;
+            } else if (javaSoundencoding.equals(Encoding.ALAW.toString())) {
+                codecPCMSignedOrUnsigned = AudioFormat.CODEC_PCM_ALAW;
+            } else {
+                codecPCMSignedOrUnsigned = null;
+            }
```
cweitkamp, line 56: > > Can you use a switch-case-statement for this block?
>
> I don't think so ?

Alright. Then we should compare the `Encoding`s directly without converting them to a String and put constants first in comparison:

```suggestion
            Encoding javaSoundencoding = format.getEncoding();
            String codecPCMSignedOrUnsigned = null;
            if (Encoding.PCM_SIGNED.equals(javaSoundencoding)) {
                codecPCMSignedOrUnsigned = AudioFormat.CODEC_PCM_SIGNED;
            } else if (Encoding.PCM_UNSIGNED.equals(javaSoundencoding)) {
                codecPCMSignedOrUnsigned = AudioFormat.CODEC_PCM_UNSIGNED;
            } else if (Encoding.ULAW.equals(javaSoundencoding)) {
                codecPCMSignedOrUnsigned = AudioFormat.CODEC_PCM_ULAW;
            } else if (Encoding.ALAW.equals(javaSoundencoding)) {
                codecPCMSignedOrUnsigned = AudioFormat.CODEC_PCM_ALAW;
            }
```

cweitkamp (changes requested):

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioFormat.java
dalgwen: Oh man, sometimes I can be really, really stupid ! Thanks, done.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioWaveUtils.java
```diff
@@ -0,0 +1,69 @@
...
+            Integer bitRate = Math.round(format.getFrameRate() * format.getFrameSize()) * format.getChannels();
+            Long frequency = Float.valueOf(format.getSampleRate()).longValue();
+            return new AudioFormat(AudioFormat.CONTAINER_WAVE, codecPCMSignedOrUnsigned, format.isBigEndian(),
+                    format.getSampleSizeInBits(), bitRate, frequency, format.getChannels());
+
```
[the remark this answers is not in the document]
dalgwen: Ok, done

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioWaveUtils.java
dalgwen: Oh I missed this package. I feel stupid again. Thanks !

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioWaveUtils.java
dalgwen: Done

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioWaveUtils.java
dalgwen: Yes, much cleaner. Thanks.

cweitkamp (approved):
LGTM. Thanks.

As I am not an audio expert I would like to have an opinion of another @openhab/core-maintainers too.

cweitkamp:
No veto. I will merge this now.

PR #2653, opened 2021-12-29
[automation] Log warning for equals condition for DecimalTypes and QuantityTypes

### LINKED ISSUE #2624: The "equals" condition does not work for the items storing dimensionless number values
I created the channel like this and linked it to the item "AqaraSmartHomeCube001_Side".
```
  - id: side
    channelTypeUID: mqtt:number
    label: Side
    description: null
    configuration:
      min: 0
      stateTopic: zigbee2mqtt/AqaraSmartHomeCube001
      transformationPattern: JSONPATH:$.side
      max: 6
```
I added the "But only if" condition to the rule like this:
```
conditions:
  - inputs: {}
    id: "4"
    configuration:
      itemName: AqaraSmartHomeCube001_Side
      state: "0"
      operator: "="
    type: core.ItemStateCondition
```

This condition always returns "false" even state of "AqaraSmartHomeCube001_Side" is equal to 0.

The problem is in the "org.openhab.core.automation.internal.module.handler.ItemStateConditionHandler". Look at the code (version 3.1.x, but i bet the version 3.2 has the same issue):
```
            Item item = itemRegistry.getItem(itemName);
            State compareState = TypeParser.parseState(item.getAcceptedDataTypes(), state);
            State itemState = item.getState();
            DecimalType decimalState = itemState.as(DecimalType.class);
            logger.debug("ItemStateCondition '{}'checking if {} (State={}) {} {}", module.getId(), itemName, itemState,
                    operator, compareState);
            switch (operator) {
                case "=":
                    logger.debug("ConditionSatisfied --> {}", itemState.equals(compareState));
                    return itemState.equals(compareState);
```
The "item.getAcceptedDataTypes()" returns "[class org.openhab.core.library.types.DecimalType, class org.openhab.core.library.types.QuantityType, class org.openhab.core.types.UnDefType]". It means the "TypeParser.parseState(item.getAcceptedDataTypes(), state)" returns object of "DecimalType" creating as "DecimalType.valueOf("0")". But the "item.getState()" returns the object of "QuantityType". The "compareState" and "itemState" have different types. and the equation "temState.equals(compareState)" never returns "true".

The safest way to solve it to rewrite the equation like this:
```
return (null != decimalState && compareState instanceof DecimalType && decimalState .equals((DecimalType)compareState)) ||  itemState.equals(compareState);
```
But I think the right path to fix it is to convert the "itemState" to the class of "compareState" before comparing.

I have found a workaround. The condition can be replaced by this couple which works well:
```
conditions:
  - inputs: {}
    id: "5"
    configuration:
      itemName: AqaraSmartHomeCube001_Side
      state: "0"
      operator: ">="
    type: core.ItemStateCondition
  - inputs: {}
    id: "4"
    configuration:
      itemName: AqaraSmartHomeCube001_Side
      state: "0"
      operator: <=
    type: core.ItemStateCondition
```

cweitkamp:
- Log warning for equals condition for `DecimalType`s and `QuantityType`s

Closes #2624 

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ItemStateConditionHandler.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ItemStateConditionHandler.java
@@ -77,7 +77,6 @@ public class ItemStateConditionHandler extends BaseConditionModuleHandler {
         itemRegistry = null;
     }
 
-    @SuppressWarnings({ "rawtypes", "unchecked", "null" })
     @Override
     public boolean isSatisfied(Map<String, Object> inputs) {
         String itemName = (String) module.getConfiguration().get(ITEM_NAME);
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.automation/src/test/java/org/openhab/core/automation/internal/module/handler/ItemStateConditionHandlerTest.java
```diff
@@ -65,9 +65,9 @@ public ParameterSet(String comparisonState, State itemState, boolean expectedRes
                 { new ParameterSet("5", new DecimalType(23), false) }, //
                 { new ParameterSet("5", new DecimalType(5), true) }, //
                 { new ParameterSet("5 °C", new DecimalType(23), false) }, //
-                { new ParameterSet("5 °C", new DecimalType(5), false) }, //
+                { new ParameterSet("5 °C", new DecimalType(5), true) }, //
```
kaikreuzer, line 68: I would claim that this change is not correct.
The unit test was just right before: 5 is NOT the same as 5 °C as it is lacking any unit and could thus well also be 5 °F.

kaikreuzer (changes requested):
Isn't the real issue about #2624 that `TypeParser.parseState(item.getAcceptedDataTypes(), state)` returns a `DecimalType` instead of a `QuantityType`? Shouldn't `item.getAcceptedDataTypes()` return `QuantityType` before `DecimalType` in case the number item has a dimension?

bundles/org.openhab.core.automation/src/test/java/org/openhab/core/automation/internal/module/handler/ItemStateConditionHandlerTest.java
cweitkamp: Yes, I agree.

cweitkamp:
From my pov the roit cause of #2624 is that the user did not configuration a comparison value including a unit. And additionally the framework did not inform him about his mistake.

IIRC we cannot change the order of accepted data types for `Number` items because every plain number without a unit will parsed by the `QuantityType` with unit `ONE`.

I will revert my changes partially to just log a warning if the item state is a `QuantityType` but comparison value does not have a unit.

bundles/org.openhab.core.automation/src/test/java/org/openhab/core/automation/internal/module/handler/ItemStateConditionHandlerTest.java
cweitkamp: Keeping this in mind we should adopt the remaining comparisons too and add a conversion to a base unit before. I will have a look.

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ItemStateConditionHandler.java
```diff
@@ -93,122 +92,99 @@ public boolean isSatisfied(Map<String, Object> inputs) {
...
+    }
+
+    @SuppressWarnings("null")
+    private boolean equalsToItemState(String itemName, String state) throws ItemNotFoundException {
+        Item item = itemRegistry.getItem(itemName);
+        State compareState = TypeParser.parseState(item.getAcceptedDataTypes(), state);
+        State itemState = item.getState();
+        if (itemState instanceof QuantityType && compareState instanceof DecimalType) {
+            logger.warn(
+                    "Received a QuantityType state '{}' with unit for item {}, but the condition is defined as a plain number without unit ({}), please consider adding a unit to the condition.",
```
kaikreuzer, line 185: "please consider adding" sounds as if it is optional and will work nonetheless without.
Shouldn't we maybe use some stronger formulation to make it clear that the current configuration is invalid and the comparison is simply not done (always returns false)?

cweitkamp: I reformulated the warning to be more precise.

Additionally I added a check to deal with dimensionless `QuantityType`s without unit (internally `AbstractUnit.ONE`) because for them it is not possible to define a unit in the compare condition as the unit symbol is an empty string.

kaikreuzer (approved):
Thank you!

PR #2658, opened 2021-12-31
"System reached start level XX" trigger for DSL rules

Flole998:
This implements a way to trigger dsl-based rules on various start levels, like it is already possible with ui-based rules.

### DIFF
--- a/bundles/org.openhab.core.model.rule.runtime/src/org/openhab/core/model/rule/runtime/internal/DSLRuleProvider.java
+++ b/bundles/org.openhab.core.model.rule.runtime/src/org/openhab/core/model/rule/runtime/internal/DSLRuleProvider.java
@@ -51,6 +51,7 @@ import org.openhab.core.model.rule.rules.GroupMemberUpdateEventTrigger;
 import org.openhab.core.model.rule.rules.RuleModel;
 import org.openhab.core.model.rule.rules.SystemOnShutdownTrigger;
 import org.openhab.core.model.rule.rules.SystemOnStartupTrigger;
+import org.openhab.core.model.rule.rules.SystemStartlevelTrigger;
 import org.openhab.core.model.rule.rules.ThingStateChangedEventTrigger;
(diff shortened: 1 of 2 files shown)


kaikreuzer:
What exactly is the intention here? The lowest start level that allows execution of rules is 40/50 when the "system started ones are triggered". Besides this, there are no higher start levels implemented so far, so I do not really see any use case here (yet)?

Flole998:
There are 4 reasons I did this (and you will probably only consider one or two of them a valid reason ;) ):
1. To reduce the feature gap between UI and text based rules
2. I had issues with rules executing while states haven't been restored yet and moving it to level 100 seems to resolve those cases
3. There are cases where you have rules that are supposed to "initialize" some items and other rules to run after that. This is now possible by having initialization rules run in level 40 and rules depending on those in level 50 (or 100).
4. Some rules may depend on the UI to be available because they trigger external scripts that fetch something from the UI. Then those can be moved to level 100. As far as I can see that should already be possible now, even though other levels

Anyways, I understand your concern about this being potentially confusing. To fix that I suggest adding a warning if a startup level < 40 is supplied that "Rules can not be executed until start level 40 is reached".

We would also have to change the documentation which is currently stating
"Startlevels (logically only if startlevel >= 40) are only available in UI-Rules, not in DSL-Rules with textual configuration."
to something like
"Startlevels less than 40 are not available as the rule engine needs to be ready first before it can run any rules." (not sure if "start level" or "startlevel")

We can also put this aside until other levels are implemented, but with proper documentation and even a warning in the logs it should do no harm to merge it now, even though the full use case will only be visible later.

kaikreuzer (approved):
Thanks for your explanation @Flole998!
Fine for me to add this so that DSL-rules have the same features as UI-rules.
Maybe then worth to pick up the implementation for higher start levels soon. ;-)

Could you be so kind to also provide a PR for the documentation?

Issue #2618, opened 2021-12-15
[Persistence] API to save data not working


TBail:
While working on some persistence rules i tried to use the API to save some data to a persistence service. Unfortunately this idi not work.

To find out what is wrong with my code i tried the same with the api explorer in OH3 latest nightly. I used the

`PUT /persistence/items/{itemname}`

with appropriate parameters, but unfortunaltely i got an error with both services installed. i used (influxdb and mapdb). The error is

```
{
  "error": {
    "message": "Persistence service not modifiable: mapdb",
    "http-code": 400
  }
}
```

To ensure the the parameters like service id and time etc are working in general i tried ti retrieve persistence data. This worked very well.

Based on this i assume that writing persistence data with thw rest api does not work.

cweitkamp:
Thanks for reporting this missing feature. To store values in persistence via REST interface the services have to implement the `ModifiablePersistenceService` interface - which extends the `QueryablePersistenceService` interface. After a quick look into [Add-Ons repository](https://github.com/openhab/openhab-addons/search?q=ModifiablePersistenceService&type=) none of the available persistence services implements the mentioned feature.

I will try to have a look later into mapdb and jdbc how it can be archived.

- [ ] dynamodb
- [ ] influxdb
- [ ] jdbc
- [ ] jpa
- [ ] mapdb
- [ ] mongodb
- [ ] rrd4j

spacemanspiff2007:
Thank you @cweitkamp for stepping in! That's really nice of you.

cweitkamp:
Does it really make sense to implement this interface for MapDB as it only can store the last value? Please continue discussion in https://github.com/openhab/openhab-addons/pull/11923.

lujop:
@cweitkamp It's possible to implement only the store part without removing one and throwing a not-supported exception?

InfuxDB 2.0 it's not very supportive in deletes, and implementing full FilterCriteria won't be possible and require some work.
Because delete predicates are totally different than query predicates and the API doesn't support to construct them, and there are operators like inequality that are not supported (see  https://docs.influxdata.com/influxdb/cloud/reference/syntax/delete-predicate)
However the store part it's trivial and ready to submit a PR.

cweitkamp:
Yes, I think so. Just throw an `UnsupportedOperationException` in the `remove` method. That's it.

lujop:
I've added [#12013](https://github.com/openhab/openhab-addons/issues/12013) to track InfluxDB implementation

[...]

PR #2660, opened 2022-01-02
Use ZonedDateTime instead of Date in ModifiablePersistenceService interface

wborn:
If PRs are created to implement this interface, it would be better if we can prevent the propagation of old APIs like Date.

Related to:

* https://github.com/openhab/openhab-addons/pull/11922
* https://github.com/openhab/openhab-addons/pull/11923
* https://github.com/openhab/openhab-core/issues/2618#issuecomment-1003544762


cweitkamp (approved):
LGTM. Thanks.

This is a good idea 👍🏼 .

PR #2661, opened 2022-01-03
[automation] Align parsing of startlevel config

cweitkamp:
- Align parsing of startlevel config

We assume that configuration values are `BigDecimal` in `SystemTriggerHandler` thus it might make sense to align the parsing in the `RuleEngine`.

https://github.com/openhab/openhab-core/blob/0bdaeef789a52d637134800f3dbaf4827b543c60/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/SystemTriggerHandler.java#L62

Otherwise it might happen that we see a warning in the log:

```
2022-01-03 12:35:24.420 [WARN ] [e.automation.internal.RuleEngineImpl] - Configuration '80.0' is not a valid start level!
```

Extract from Json Storage:

```json
{
  "e8c39a6060": {
    "class": "org.openhab.core.automation.dto.RuleDTO",
    "value": {
      "triggers": [
        {
          "id": "1",
          "configuration": {
            "startlevel": 80.0
          },
          "type": "core.SystemStartlevelTrigger"
        }
      ],
      "conditions": [],
...
```

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleEngineImpl.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleEngineImpl.java
@@ -12,6 +12,7 @@
  */
 package org.openhab.core.automation.internal;
 
+import java.math.BigDecimal;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


kaikreuzer (approved):
Thanks!

PR #2662, opened 2022-01-03
[profiles] Do not log a warning if QuantityState has unit AbstractUnit.ONE

cweitkamp:
- Do not log a warning if `QuantityState` has unit `AbstractUnit.ONE`

I added this check to deal with dimensionless `QuantityTypes` without unit (internally `AbstractUnit.ONE`) because for them it is not possible to define a unit in the compare condition as the unit symbol is an empty string.


kaikreuzer (approved):
lgtm, thank you!

PR #2664, opened 2022-01-04
ScriptFileWatcher subdirectory loading upon startup no longer hardcoded

jpg0:
… to true

This matters as subclasses can override this behaviour and it was previously being ignored upon startup (only).


cweitkamp (approved):
LGTM. Thanks.

PR #2668, opened 2022-01-05
[rest] Do not reuse FilterCriteria instance for more than one query

### LINKED ISSUE #2665: requesting persistence data with boundary parameter does not seem to include end date datapoint
While trying to resolve https://github.com/openhab/openhab-webui/issues/912 was toying a bit with the `boundary` parameter and it seems as when set to `true` the returned data array does not include end date (as it does for start date).

i.e. requesting something like ```http://localhost:8080/rest/persistence/items/gBasement_BoilerRoom_HP_RAD_InOperation_HP?boundary=true&serviceId=influxdb&starttime=2021-11-30T23:00:00.000Z&endtime=2021-12-31T22:59:59.999Z``` returns 

```
"name": "gBasement_BoilerRoom_HP_RAD_InOperation_HP",
    "datapoints": "1006",
    "data": [
        {
            "time": 1638313200000,
            "state": "23598.0"
        },
...
        {
            "time": 1640988002024,
            "state": "23882.0"
        }
    ]
}
```

so `1638313200000 == 2021-11-30T23:00:00.000Z (start date)` while `1640988002024 != 2021-12-31T22:59:59.999Z (end date)` (I checked there are points after end date in DB, fyi).

Also, when requesting with boundary true and then false, one would expect number of datapoints to be different by two elements (start & end, ofc if there are points in DB before and after), but the diff is only a single element (start as it seems).

Quickly checked codebase and my guess would be that since `filter` is reused [here](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/persistence/PersistenceResource.java#L346) both begin and [end](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/persistence/PersistenceResource.java#L320) dates are in effect (and set to the same value) ...

cweitkamp:
- Do not reuse `FilterCriteria` instance for more than one query

Closes #2665

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/persistence/PersistenceResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/persistence/PersistenceResource.java
@@ -432,13 +432,6 @@ public class PersistenceResource implements RESTResource {
                     "Persistence service not modifiable: " + serviceId);
         }
 
-        try {
-            Item item = itemRegistry.getItem(itemName);
-        } catch (ItemNotFoundException e) {
-            logger.warn("Item not found '{}'.", itemName);
-            return JSONResponse.createErrorResponse(Status.BAD_REQUEST, "Item not found: " + itemName);
-        }
-
         if (timeBegin == null || timeEnd == null) {
             return JSONResponse.createErrorResponse(Status.BAD_REQUEST, "The start and end time must be set");
         }
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/persistence/PersistenceResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/persistence/PersistenceResource.java
@@ -281,7 +281,6 @@ public class PersistenceResource implements RESTResource {
                     timeZoneProvider.getTimeZone());
         }
 
-        FilterCriteria filter;
         Iterable<HistoricItem> result;
         State state = null;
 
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/persistence/PersistenceResource.java
```diff
@@ -432,7 +432,12 @@ private Response deletePersistenceItemData(@Nullable String serviceId, String it
                     "Persistence service not modifiable: " + serviceId);
         }
 
-        ModifiablePersistenceService mService = (ModifiablePersistenceService) service;
+        try {
+            Item item = itemRegistry.getItem(itemName);
+        } catch (ItemNotFoundException e) {
+            logger.warn("Item not found '{}'.", itemName);
+            return JSONResponse.createErrorResponse(Status.BAD_REQUEST, "Item not found: " + itemName);
+        }
```
wborn, line 440: Is this check really necessary or do persistence services start throwing exceptions when items do not exist in the registry?

With this check it might for instance no longer be possible for users to clean up persisted item state after deleting the items.

wborn (commented):
These changes look good to me, but I have a question below.
Can you also fix the merge conflicts?

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/persistence/PersistenceResource.java
wborn: I think I now see this is only done for storing new data now. :slightly_smiling_face:

wborn: It's in `deletePersistenceItemData` so my question still stands. :thinking:

cweitkamp: If we want to allow users to truncate data for not existing items this check indeed is wrong. The question is what happens to the orphaned table then?

I can remove it. If I do I have to change my implementation for JDBC in https://github.com/openhab/openhab-addons/pull/11922 too because it checks for the item to be present in the registry too and returns false for delete operation if not - it does not throw an exception in that case.

cweitkamp:
I fixed the conflicts and removed the check if the item exists.

cweitkamp:
@wborn I changed my implementation for JDBC too: https://github.com/openhab/openhab-addons/pull/12052

wborn (approved):
Thanks!

PR #2671, opened 2022-01-07
Update license headers to 2022

cweitkamp:
- Update license headers to 2022

Signed-off-by: Christoph Weitkamp <github@christophweitkamp.de>

PR #2675, opened 2022-01-08
Update groovy license headers to 2022

wborn:
Updates the year in a few groovy file headers that were missed in #2671.

### DIFF
--- a/bundles/org.openhab.core.semantics/model/generateTagClasses.groovy
+++ b/bundles/org.openhab.core.semantics/model/generateTagClasses.groovy
@@ -12,10 +12,14 @@
  */
 @Grab('com.xlson.groovycsv:groovycsv:1.1')
 import static com.xlson.groovycsv.CsvParser.parseCsv
+import java.nio.file.Files
 import java.nio.file.Paths
 import java.time.Year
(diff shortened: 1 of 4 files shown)


cweitkamp (commented):
Thanks.
Do you have an idea why it wasn't picked up by the man command?

https://github.com/openhab/openhab-core/blob/89aae83b64510bd95f934047b069d332055fff80/pom.xml#L449

And add archetype Groovy script to the includes in the POM file?

wborn:
> Do you have an idea why it wasn't picked up by the man command?

It did update the header at the top of the file but not the header used in the groovy code. 
So I've tweaked the code so it will always use the current year.

cweitkamp:
That's an improvement. At best it would use the content from [licenses/epl-2.0/header.txt](https://github.com/openhab/openhab-core/blob/main/licenses/epl-2.0/header.txt) to add the header to generated classes :wink:. Wdyt?

wborn:
It was not too difficult to accomplish that too. :wink:

cweitkamp (approved):
LGTM. Thanks.

Good to know that next year we have a little less that we don't have to worry about.

PR #2682, opened 2022-01-13
enable config validation for thing creation and update

### LINKED ISSUE #2674: Thing parameter min and max values defined in binding XML files not applied to parameters defined in config files
This came up in the discussion of https://github.com/openhab/openhab-addons/pull/10310 and @fwolter suggested reporting it here.

With parameters defined in a thing configuration such as the following:

```
	<bridge-type id="cloudbridge">
		<label>SensorPush Cloud Bridge</label>
		<description>SensorPush gateway cloud access</description>
…

		<config-description>
…
			<parameter name="poll" type="integer" min="2" max="60" unit="min">
				<label>Poll Interval</label>
				<description>The interval at which the handler will poll the cloud service for updates</description>
				<unitLabel>minutes</unitLabel>
				<default>5</default>
			</parameter>
…
		</config-description>
	</bridge-type>
```

The min/max values in the XML only appear to be enforced by the GUI. It happily let me violate the limits in a text config file by specifying poll=1 when min="2" is set for the poll parameter in the XML.

I recall this being a known limitation from way back, but I don’t see it documented anywhere. Should this limitation be added to the developer docs, or is it considered a bug?

J-N-K:
Fixes #2674 

Depends on #2683

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
@@ -21,7 +21,6 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
@@ -42,11 +41,10 @@ import org.openhab.core.common.ThreadPoolManager;
 import org.openhab.core.common.registry.Identifiable;
 import org.openhab.core.common.registry.ManagedProvider;
 import org.openhab.core.common.registry.Provider;
-import org.openhab.core.config.core.ConfigDescription;
-import org.openhab.core.config.core.ConfigDescriptionParameter;
 import org.openhab.core.config.core.ConfigDescriptionRegistry;
 import org.openhab.core.config.core.Configuration;
 import org.openhab.core.config.core.validation.ConfigDescriptionValidator;
+import org.openhab.core.config.core.validation.ConfigValidationException;
 import org.openhab.core.events.EventPublisher;
 import org.openhab.core.service.ReadyMarker;
 import org.openhab.core.service.ReadyMarkerFilter;
@@ -558,6 +556,7 @@ public class ThingManagerImpl
             // called from the thing handler itself, therefore
             // it exists, is initializing/initialized and
             // must not be informed (in order to prevent infinite loops)
+            // we assume the handler knows what he's doing and don't check config validity
             replaceThing(getThing(thingUID), thing);
         } else {
             Lock lock1 = getLockForThing(thing.getUID());
@@ -571,7 +570,23 @@ public class ThingManagerImpl
                             oldThing.setHandler(null);
                         }
                         thing.setHandler(thingHandler);
-                        safeCaller.create(thingHandler, ThingHandler.class).build().thingUpdated(thing);
+
+                        if (isInitializable(thing, getThingType(thing))) {
+                            safeCaller.create(thingHandler, ThingHandler.class).build().thingUpdated(thing);
+                        } else {
+                            final ThingHandlerFactory thingHandlerFactory = findThingHandlerFactory(
+                                    thing.getThingTypeUID());
+                            if (thingHandlerFactory != null) {
+                                if (isBridge(thing)) {
+                                    unregisterAndDisposeChildHandlers((Bridge) thing, thingHandlerFactory);
+                                }
+                                disposeHandler(thing, thingHandler);
+                                setThingStatus(thing,
+                                        buildStatusInfo(ThingStatus.UNINITIALIZED,
+                                                ThingStatusDetail.HANDLER_CONFIGURATION_PENDING,
+                                                "Missing or invalid configuration."));
+                            }
+                        }
                     } else {
                         logger.debug(
                                 "Cannot notify handler about updated thing '{}', because handler is not initialized (thing must be in status UNKNOWN, ONLINE or OFFLINE).",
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
```diff
@@ -571,7 +570,23 @@ public void thingUpdated(final Thing thing, ThingTrackerEvent thingTrackerEvent)
...
+                                    thing.getThingTypeUID());
+                            if (thingHandlerFactory != null) {
+                                if (isBridge(thing)) {
+                                    unregisterAndDisposeChildHandlers((Bridge) thing, thingHandlerFactory);
+                                }
+                                disposeHandler(thing, thingHandler);
+                                setThingStatus(thing,
+                                        buildStatusInfo(ThingStatus.UNINITIALIZED,
+                                                ThingStatusDetail.HANDLER_CONFIGURATION_PENDING,
+                                                "Missing or invalid configuration."));
```
cweitkamp, line 587: As the `ThingStatusInfo` is translatable I am wondering if we should introduce a custom translation key for this description (e.g.`@text/missing-or-invalid-configuration`).

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
```diff
@@ -744,23 +759,25 @@ private void initializeHandler(Thing thing) {
...
                 doInitializeHandler(handler);
             } else {
-                logger.debug("Thing '{}' not initializable, check required configuration parameters.", thing.getUID());
-                setThingStatus(thing,
-                        buildStatusInfo(ThingStatus.UNINITIALIZED, ThingStatusDetail.HANDLER_CONFIGURATION_PENDING));
+                logger.debug(
+                        "Thing '{}' not initializable, check if required configuration parameters are present and set values are valid.",
+                        thing.getUID());
+                setThingStatus(thing, buildStatusInfo(ThingStatus.UNINITIALIZED,
+                        ThingStatusDetail.HANDLER_CONFIGURATION_PENDING, "Missing or invalid configuration."));
```
cweitkamp, line 766: Use the same custom key here.

cweitkamp (commented):
Thanks for this improvement and for streamlining the configuration validation.

I left two minor comments.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
J-N-K: Where Do I have to place the correct value then? `ThingManagerImpl.properties` in the i81n-folder? Or `ThingStatusInfo.properties`?

cweitkamp: I think it should be place in a new properties file with ame `SystemThingStatusInfos.properties`.

cweitkamp (approved):
LGTM from my side.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
```diff
@@ -786,35 +803,20 @@ private boolean isInitializable(Thing thing, @Nullable ThingType thingType) {
...
         }
 
-        List<String> requiredParameters = getRequiredParameters(description);
-        Set<String> propertyKeys = configuration.getProperties().keySet();
-        if (logger.isDebugEnabled()) {
-            logger.debug("Configuration of '{}' needs {}, has {}.", targetUID, requiredParameters, propertyKeys);
+        try {
+            configDescriptionValidator.validate(configuration.getProperties(), configDescriptionURI);
+        } catch (ConfigValidationException e) {
+            return false;
```
cweitkamp, line 816: Should we add a trace log here to allow users to see validation messages in the log file? This might help to find the culprit for #2710.

J-N-K:
Yes, I‘ll prepare that immediately. It seems that at least some of the issues arise from signed/unsigned byte/int values. I‘ll check if that is a problem in the code or in the affected binding.

petero-dk:
I am online and have this issue, so let me know how I can help

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
J-N-K: See https://github.com/openhab/openhab-core/pull/2711

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
```diff
@@ -558,6 +556,7 @@ public void thingUpdated(final Thing thing, ThingTrackerEvent thingTrackerEvent)
             // called from the thing handler itself, therefore
             // it exists, is initializing/initialized and
             // must not be informed (in order to prevent infinite loops)
+            // we assume the handler knows what he's doing and don't check config validity
```
J-N-K, line 559: It seems that this assumption is wrong as e.g. the ZWave handler updates the configuration with values outside of the valid range. @cweitkamp shall we add a validation here, too? And what shall happen if the thing handler tries to update with an invalid configuration? Log a warning and deny replacing the thing?

cweitkamp: Yes, I agree. Both methods in `BaseThingHandler` which allows to update the configuration of a thing (`updateThing(Thing)` and `updateConfiguration(Configuration)`) lead to an unchecked configuration. There is no validation done.

Both `updateThing()` and `updateConfiguration()` call `ThingHandlerCallback.thingUpdated()` -> `ThingRegistry.updated()` -> `ThingTracker.thingUpdated()` of `ThingManagerImpl`.

In theorie we have to to check for a changed configuration of the thing and do the same like the `ThingResource`: call `ThingRegistry.updateConfiguration()` instead of just replacing the thing here. imo the handler has to be reinitialized if configuration will change.

Or did I miss something.

cdjackson: > It seems that this assumption is wrong

I think the problem with this assumption is that the binding doesn't have access to the config description, so it isn't (easily) possible for it to do this as far as I can see. The framework only passes the values to the binding - maybe it should also make the descriptions more readily available as well now that we're enforcing these checks.

PR #2690, opened 2022-01-16
[config] Add nullness annotations to config validation

cweitkamp:
- Added nullness annotations to config validation
- Use constructor injection

### DIFF
--- a/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/ConfigDescriptionValidatorImpl.java
+++ b/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/ConfigDescriptionValidatorImpl.java
@@ -14,13 +14,13 @@ package org.openhab.core.config.core.internal.validation;
 
 import java.net.URI;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.config.core.ConfigDescription;
 import org.openhab.core.config.core.ConfigDescriptionParameter;
 import org.openhab.core.config.core.ConfigDescriptionRegistry;
@@ -32,10 +32,7 @@ import org.openhab.core.i18n.TranslationProvider;
 import org.osgi.framework.BundleContext;
 import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
-import org.osgi.service.component.annotations.Deactivate;
 import org.osgi.service.component.annotations.Reference;
-import org.osgi.service.component.annotations.ReferenceCardinality;
-import org.osgi.service.component.annotations.ReferencePolicy;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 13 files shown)


bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/ConfigDescriptionValidatorImpl.java
```diff
@@ -49,20 +47,28 @@
...
+    private final Logger logger = LoggerFactory.getLogger(ConfigDescriptionValidatorImpl.class);
 
-    private static final List<ConfigDescriptionParameterValidator> VALIDATORS = Collections
-            .unmodifiableList(Arrays.asList(ConfigDescriptionParameterValidatorFactory.createRequiredValidator(),
-                    ConfigDescriptionParameterValidatorFactory.createTypeValidator(),
-                    ConfigDescriptionParameterValidatorFactory.createMinMaxValidator(),
-                    ConfigDescriptionParameterValidatorFactory.createPatternValidator()));
+    private final ConfigDescriptionRegistry configDescriptionRegistry;
+    private final TranslationProvider translationProvider;
+    private @NonNullByDefault({}) BundleContext bundleContext;
```
J-N-K, line 64: Why don't you inject that with the constructor, too?

bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/ConfigDescriptionValidatorImpl.java
```diff
@@ -49,20 +47,28 @@
...
 
-    private BundleContext bundleContext;
+    @Activate
+    public ConfigDescriptionValidatorImpl(final @Reference ConfigDescriptionRegistry configDescriptionRegistry,
+            final @Reference TranslationProvider translationProvider) {
```
J-N-K, line 68: This makes the `TranslationProvider` and the `ConfigDescriptionRegistry` mandatory, they were optional before. That change makes sense to me, but did you check if this does not create circular dependencies?

bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/ConfigDescriptionValidatorImpl.java
cweitkamp: Ups, ... Will add it.

bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/ConfigDescriptionValidatorImpl.java
cweitkamp: No, but afair the `TranslationProvider` does not reference any other services. The `ConfigDescriptionRegistry` only references optional provider.

bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/ConfigDescriptionValidatorImpl.java
```diff
@@ -86,7 +83,7 @@ protected void deactivate() {
      * @throws NullPointerException if given config description URI or configuration parameters are null
      */
     @Override
-    @SuppressWarnings("unchecked")
+    @SuppressWarnings({ "unchecked", "null" })
     public void validate(Map<String, Object> configurationParameters, URI configDescriptionURI) {
         Objects.requireNonNull(configurationParameters, "Configuration parameters must not be null");
```
J-N-K, line 88: If the class is annotated `@NonNullByDefault` both parameters are `@NonNull`. Isn't this check unnecessary?

J-N-K (commented):
In my opinion we can also remove `ConfigDescriptionParameterValidatorFactory`. I don't think there is any benefit in calling the factory compared to instantiate directly. If the idea was to save resources, it would be better to create a single static instance of each validator.

bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/ConfigDescriptionValidatorImpl.java
cweitkamp: Yes, and there are plenty more of them which can be omitted in the related validation classes. But let us focus on annotations and c-tor injection here and move other improvements into a follow-up PR.

J-N-K: Fine with me. I don't understand why the build failed.

cweitkamp:
@openhab/core-maintainers May I ask you to have a look into this PR. Thanks.

bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/ConfigDescriptionValidatorImpl.java
```diff
@@ -49,29 +46,29 @@
...
+    private final Logger logger = LoggerFactory.getLogger(ConfigDescriptionValidatorImpl.class);
 
-    private BundleContext bundleContext;
+    private final ConfigDescriptionRegistry configDescriptionRegistry;
+    private final TranslationProvider translationProvider;
+    private final BundleContext bundleContext;
 
     @Activate
-    protected void activate(BundleContext bundleContext) {
+    public ConfigDescriptionValidatorImpl(final @Reference BundleContext bundleContext,
```
wborn, line 66: You can probably fix the build by removing `@Reference` here:

```suggestion
    public ConfigDescriptionValidatorImpl(final BundleContext bundleContext,
```

The `BundleContext` is not something you can get a reference on like another component. But I think the annotations processor is smart enough to inject the `BundleContext` when it sees such a parameter. We also use it like this with the `StartLevelService`.

cweitkamp: Thanks for the hint. I applied your suggestion.

PR #2691, opened 2022-01-16
[config] Add config validator for parameter options

cweitkamp:
- Added config validator for parameter options

Depends on #2690 

### DIFF
--- a/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/OptionsValidator.java
+++ b/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/OptionsValidator.java
@@ -32,9 +32,10 @@ final class OptionsValidator implements ConfigDescriptionParameterValidator {
         }
 
         // Option values are a string, so we can do a simple compare
-        if (!param.getOptions().stream().map(o -> o.getValue()).anyMatch(v -> v.equals(value.toString()))) {
+        if (param.getOptions().stream().map(o -> o.getValue()).noneMatch(v -> v.equals(value.toString()))) {
             MessageKey messageKey = MessageKey.OPTIONS_VIOLATED;
-            return new ConfigValidationMessage(param.getName(), messageKey.defaultMessage, messageKey.key);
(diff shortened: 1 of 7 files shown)


bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/OptionsValidator.java
```diff
@@ -0,0 +1,41 @@
...
+final class OptionsValidator implements ConfigDescriptionParameterValidator {
+
+    @Override
+    public @Nullable ConfigValidationMessage validate(ConfigDescriptionParameter param, @Nullable Object value) {
+        if (value == null || !param.getLimitToOptions() || param.getOptions().isEmpty()) {
+            return null;
+        }
+
+        // Option values are a string, so we can do a simple compare
+        if (!param.getOptions().stream().map(o -> o.getValue()).anyMatch(v -> v.equals(value.toString()))) {
```
J-N-K, line 35: Use `noneMatch`

bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/OptionsValidator.java
```diff
@@ -0,0 +1,41 @@
...
+            MessageKey messageKey = MessageKey.OPTIONS_VIOLATED;
+            return new ConfigValidationMessage(param.getName(), messageKey.defaultMessage, messageKey.key);
```
J-N-K, line 37: It would be nice to add the list of allowed options here. This might be helpful to correct the setting.

J-N-K (changes requested):
A general comment: IMO `ConfigDescriptionParameterValidatorFactory` should be removed. There is no need for it, it just returns a new instance which could also be achieved by inserting `new ...` in the map. But maybe we should put that in another PR (like removing the unnecessary checks we talked about in #2690.

J-N-K (approved):

wborn (approved):
Thanks for adding the validator!

PR #2692, opened 2022-01-16
[skeleton] Add Thing config table to readme; minor improvements

fwolter:
This addresses some review findings for new addons by making some small improvements to the binding skeleton.

- put each sentence in a seperate line
- add note that only PNG images are supported. See https://github.com/openhab/openhab-addons/pull/11116#discussion_r775838188
- add examples to "Supported Things"
- add RW column to example Channel table
- add advanced and default config value
- add note that textual config examples are mandatory
- make .sitemap example optional (see small internal discussion)
- put emphasis on initialize() shall return quickly
- add note about logging to INFO
- fix compiler warning

@openhab/add-ons-maintainers 

### DIFF
--- a/tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/internal/__bindingIdCamelCase__Configuration.java
+++ b/tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/internal/__bindingIdCamelCase__Configuration.java
@@ -15,17 +15,20 @@
  */
 package ${package}.internal;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
 /**
  * The {@link ${bindingIdCamelCase}Configuration} class contains fields mapping thing configuration parameters.
  *
  * @author ${author} - Initial contribution
  */
+@NonNullByDefault
 public class ${bindingIdCamelCase}Configuration {
 
     /**
      * Sample configuration parameters. Replace with your own.
      */
-    public String hostname;
-    public String password;
+    public String hostname = "";
+    public String password = "";
     public int refreshInterval;
 }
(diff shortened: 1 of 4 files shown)


martinvw (approved):

tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/internal/__bindingIdCamelCase__Configuration.java
```diff
@@ -15,17 +15,20 @@
...
 public class ${bindingIdCamelCase}Configuration {
 
     /**
      * Sample configuration parameters. Replace with your own.
      */
-    public String hostname;
-    public String password;
+    public String hostname = "";
+    public String password = "";
     public int refreshInterval;
```
lolodomo, line 33: Please set it to 600 by default to be compliant with the XML.

lolodomo (changes requested):

tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/internal/__bindingIdCamelCase__Configuration.java
fwolter: I left the default value intentionally to avoid any redundancy between here and the XML. But I'm okay with it to add it.

lolodomo: In practice, I always take care to set the initial value to match the default value in the XML, and I even recommend developers to do it when I review PR. So I think it would be better to show in the skeleton this good practice.

fwolter: I changed it, although I'm a big fan of avoiding redundancies.

lolodomo (approved):
LGTM

PR #2695, opened 2022-01-18
Do not throw CertificateParsingException all the time; Fixed pinned certificate creation;

cweitkamp:
- Do not throw `CertificateParsingException` all the time
- Fixed pinned certificate creation
- Added `hashCode()` and `equals()` to ensure correct removals from `Queue` in `ExtensibleTrustManager` implementation

See https://community.openhab.org/t/java-connection-to-https-with-self-signed-cert/130593

### DIFF
--- a/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/PEMTrustManager.java
+++ b/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/PEMTrustManager.java
@@ -207,7 +207,8 @@ public final class PEMTrustManager extends X509ExtendedTrustManager {
 
             byte[] bytes = ((X509Certificate) certs[0]).getEncoded();
             if (bytes.length != 0) {
-                return BEGIN_CERT + "\r\n" + Base64.getEncoder().encodeToString(bytes) + "\r\n" + END_CERT;
+                return BEGIN_CERT + System.lineSeparator() + Base64.getEncoder().encodeToString(bytes)
+                        + System.lineSeparator() + END_CERT;
             }
         } catch (NoSuchAlgorithmException | KeyManagementException | IOException e) {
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/PEMTrustManager.java
```diff
@@ -187,7 +207,7 @@ public void checkServerTrusted(X509Certificate @Nullable [] chain, @Nullable Str
 
             byte[] bytes = ((X509Certificate) certs[0]).getEncoded();
             if (bytes.length != 0) {
-                return BEGIN_CERT + Base64.getEncoder().encodeToString(bytes) + END_CERT;
+                return BEGIN_CERT + "\r\n" + Base64.getEncoder().encodeToString(bytes) + "\r\n" + END_CERT;
```
wborn, line 210: Are you sure `\r\n` will work on any OS? If not `System.lineSeparator()` might help?

cweitkamp: No, I am not sure.

I tested it on Linux which was working. There was a user in the community forum testing on Windows and he told me that it is working for him too.

Is there a difference between `System.getProperty("line.separator")` and `System.lineSeparator()`. I used the first command in my initial commit of https://github.com/openhab/openhab-core/pull/2622.

wborn: According to the the [System.lineSeparator() JavaDocs](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/System.html#lineSeparator()) it should usually return the same value:

> Returns the system-dependent line separator string. It always returns the same value - the initial value of the system property line.separator.

wborn (approved):
Thanks!

Issue #2687, opened 2022-01-15
Exception in JsonAddonService / marketplace bundle at startup


lolodomo:
last snapshot 2694, at startup of OH:
```
2022-01-15 16:13:03.120 [ERROR] [place.internal.json.JsonAddonService] - bundle org.openhab.core.addon.marketplace:3.3.0.202201150311 (152)[org.openhab.core.addon.marketplace.internal.json.JsonAddonService(49)] : Error during instantiation of the implementation object
java.lang.reflect.InvocationTargetException: null
        at jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method) ~[?:?]
        at jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62) ~[?:?]
        at jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45) ~[?:?]
        at java.lang.reflect.Constructor.newInstance(Constructor.java:490) ~[?:?]
        at org.apache.felix.scr.impl.inject.internal.ComponentConstructorImpl.newInstance(ComponentConstructorImpl.java:316) ~[bundleFile:?]
        at org.apache.felix.scr.impl.manager.SingleComponentManager.createImplementationObject(SingleComponentManager.java:286) [bundleFile:?]
        at org.apache.felix.scr.impl.manager.SingleComponentManager.createComponent(SingleComponentManager.java:115) [bundleFile:?]
        at org.apache.felix.scr.impl.manager.SingleComponentManager.getService(SingleComponentManager.java:1000) [bundleFile:?]
        at org.apache.felix.scr.impl.manager.SingleComponentManager.getServiceInternal(SingleComponentManager.java:973) [bundleFile:?]
        at org.apache.felix.scr.impl.manager.AbstractComponentManager.activateInternal(AbstractComponentManager.java:776) [bundleFile:?]
        at org.apache.felix.scr.impl.manager.AbstractComponentManager.enableInternal(AbstractComponentManager.java:674) [bundleFile:?]
        at org.apache.felix.scr.impl.manager.AbstractComponentManager.enable(AbstractComponentManager.java:437) [bundleFile:?]
        at org.apache.felix.scr.impl.manager.ConfigurableComponentHolder.enableComponents(ConfigurableComponentHolder.java:667) [bundleFile:?]
[...]

PR #2699, opened 2022-01-20
[config] Allow String values for other Types

cweitkamp:
- Allow String values for other Types

To not introduce regressions after #2682 on the type validation we should allow `String` values for all parameter types.

### DIFF
--- a/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/TypeIntrospections.java
+++ b/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/TypeIntrospections.java
@@ -123,7 +123,18 @@ final class TypeIntrospections {
          * @return true, if the given value can be assigned to the type of this introspection, otherwise false
          */
         boolean isAssignable(Object value) {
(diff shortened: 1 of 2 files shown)


J-N-K:
LGTM. It seems that the `ConfigMapper` already takes care of proper conversion of `String` values to the defined types.

cweitkamp:
Yes, exactly. And the `ConfigUtil` does take care of converting default values.

Looking into the `ConfigMapper` class I am wondering why we did not expose it and the static method `objectConvert(Object, Class<?>)` with it to reuse it everywhere where we have to parse configuration parameters e.g. to prevent issues like https://github.com/openhab/openhab-core/issues/2687. The mapping util can be used for OSGi service configurations too.

J-N-K:
In fact I was immediately thinking about exactly that issue. Sound like a good idea. Currently the `ConfigMapper` has not null-annotations. `objectConvert` is null-safe, it returns `null` on a `null` input value. The question is if that is a desired behavior, because it could lead to a field annotated `@NonNull` to have a null value injected.

J-N-K (approved):

kaikreuzer (approved):
Thanks!

PR #2703, opened 2022-01-22
[config] make configuration value parser available for export

### LINKED ISSUE #2700: [config] Expose ConfigMapper to streamline parsing of configuration parameters
[...]

Looking into the `ConfigMapper` class I am wondering why we did not expose it and the static method `objectConvert(Object, Class<?>)` with it to reuse it everywhere where we have to parse configuration parameters e.g. to prevent issues like https://github.com/openhab/openhab-core/issues/2687. The mapping util can be used for OSGi service configurations too.

_Originally posted by @cweitkamp in https://github.com/openhab/openhab-core/issues/2699#issuecomment-1017776072_

J-N-K:
Closes #2700 

Some points are still open:

- If a conversion is not possible (e.g. provided value is `"foo"` and desired type is `int`), shall we return `null` or throw an `IllegalArgumentException` (or better a newly defined checked `ConfigValueParserException`)?
- Currently `float`, `int`, `long` and `double` (and their wrapper classes) are supported. Shall we add code for `short`, `byte` and `char`?
- There is no support for arrays. Is that needed?
- All `Collections` are mapped to `List`. Shall we at least add `Set`?
-

### DIFF
--- a/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/ConfigParser.java
+++ b/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/ConfigParser.java
@@ -45,6 +45,8 @@ public final class ConfigParser {
             "double", Double.class, //
             "long", Long.class, //
             "int", Integer.class, //
+            "short", Short.class, //
+            "byte", Byte.class, //
             "boolean", Boolean.class);
 
     private ConfigParser() {
@@ -70,7 +72,8 @@ public final class ConfigParser {
      * @return The configuration holder object. All fields that matched a configuration option are set. If a required
      *         field is not set, null is returned.
      */
-    public static <T> @Nullable T configurationAs(Map<String, Object> properties, Class<T> configurationClass) {
+    public static <T> @Nullable T configurationAs(Map<String, @Nullable Object> properties,
+            Class<T> configurationClass) {
         T configuration;
         try {
             configuration = configurationClass.getConstructor().newInstance();
@@ -142,14 +145,14 @@ public final class ConfigParser {
     }
 
     /**
-     * Convert a value to a given type
+     * Convert a value to a given type or return default value
      *
      * @param value input value or String representation of that value
      * @param type desired target class
      * @param defaultValue default value to be used if conversion fails or input value is null
      * @return the converted value
      */
-    public static <T> T valueAs(@Nullable Object value, Class<T> type, T defaultValue) {
+    public static <T> T valueAsOrElse(@Nullable Object value, Class<T> type, T defaultValue) {
         return Objects.requireNonNullElse(valueAs(value, type), defaultValue);
     }
 
@@ -177,17 +180,27 @@ public final class ConfigParser {
                 result = bdValue.longValue();
             } else if (Integer.class.equals(typeClass)) {
                 result = bdValue.intValue();
+            } else if (Short.class.equals(typeClass)) {
+                result = bdValue.shortValue();
+            } else if (Byte.class.equals(typeClass)) {
+                result = bdValue.byteValue();
             }
         } else if (value instanceof Number) {
             Number number = (Number) value;
-            if (Long.class.equals(typeClass)) {
+            if (Float.class.equals(typeClass)) {
+                result = number.floatValue();
+            } else if (Double.class.equals(typeClass)) {
+                result = number.doubleValue();
+            } else if (Long.class.equals(typeClass)) {
                 result = number.longValue();
             } else if (Integer.class.equals(typeClass)) {
                 result = number.intValue();
-            } else if (Double.class.equals(typeClass)) {
-                result = number.doubleValue();
-            } else if (Float.class.equals(typeClass)) {
-                result = number.floatValue();
+            } else if (Short.class.equals(typeClass)) {
+                result = number.shortValue();
+            } else if (Byte.class.equals(typeClass)) {
+                result = number.byteValue();
+            } else if (BigDecimal.class.equals(typeClass)) {
+                result = new BigDecimal(number.toString());
             }
         } else if (value instanceof String && !String.class.equals(typeClass)) {
             // Handle the conversion case of String to Float,Double,Long,Integer,BigDecimal,Boolean
@@ -198,10 +211,14 @@ public final class ConfigParser {
                 result = Double.valueOf(strValue);
             } else if (Long.class.equals(typeClass)) {
                 result = Long.valueOf(strValue);
-            } else if (BigDecimal.class.equals(typeClass)) {
-                result = new BigDecimal(strValue);
             } else if (Integer.class.equals(typeClass)) {
                 result = Integer.valueOf(strValue);
+            } else if (Short.class.equals(typeClass)) {
+                result = Short.valueOf(strValue);
+            } else if (Byte.class.equals(typeClass)) {
+                result = Byte.valueOf(strValue);
+            } else if (BigDecimal.class.equals(typeClass)) {
+                result = new BigDecimal(strValue);
             } else if (Boolean.class.equals(typeClass)) {
                 result = Boolean.valueOf(strValue);
             } else if (type.isEnum()) {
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/ConfigParser.java
```diff
@@ -120,49 +141,81 @@
...
-    private static Object objectConvert(Object value, Class<?> type) {
+    /**
+     * Convert a value to a given type
+     *
+     * @param value input value or String representation of that value
+     * @param type desired target class
+     * @param defaultValue default value to be used if conversion fails or input value is null
+     * @return the converted value
+     */
+    public static <T> T valueAs(@Nullable Object value, Class<T> type, T defaultValue) {
```
cweitkamp, line 152: We can name it `valueAsElse()` or `valueAsOrElse()` to let the code have a better readability.

```suggestion
    public static <T> T valueAsElse(@Nullable Object value, Class<T> type, T defaultValue) {
```

cweitkamp (commented):
I think we should not throw exceptions rather log a warning. I assume most of the consumers have a proper `null` handling - if they do not use the `valueAsElse()` method - thus exception handling will lead to boilerplate code to be added again. The idea behind this PR is to reduce code.

Adding support for `short`, `byte` or `Set` does not hurt. Imo arrays are not needed.

J-N-K:
Ready for review from my side. I would suggest to adjust code to use this in a follow-up PR instead of changing it here.

bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/ConfigParser.java
```diff
@@ -0,0 +1,253 @@
...
+    /**
+     * Convert a value to a given type
+     *
+     * @param value input value or String representation of that value
+     * @param type desired target class
+     * @return the converted value or null if conversion fails or input value is null
+     */
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    public static <T> @Nullable T valueAs(@Nullable Object value, Class<T> type) {
+        // make sure primitives are converted to their respective wrapper class
```
cweitkamp, line 178: Should we add an early return if the` value` already is an instance of the requested `type`?

bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/ConfigParser.java
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    /**
+     * Use this method to automatically map a configuration collection to a Configuration holder object. A common
+     * use-case would be within a service. Usage example:
+     *
+     * <pre>
+     * {@code
+     *   public void modified(Map<String, Object> properties) {
+     *     YourConfig config = ConfigMapper.as(properties, YourConfig.class);
```
cweitkamp, line 65:
```suggestion
     *     YourConfig config = ConfigParser.configurationAs(properties, YourConfig.class);
```

bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/ConfigParser.java
```diff
@@ -0,0 +1,258 @@
...
+            Class<?> type = field.getType();
+
+            Object value = properties.get(fieldName);
+            // Consider RequiredField annotations
+            if (value == null) {
+                LOGGER.trace("Skipping field '{}', because config has no entry for {}", fieldName, fieldName);
+                continue;
+            }
+
+            // Allows to have List<int>, List<Double>, List<String> etc (and the corresponding Set<?>
```
cweitkamp, line 104:
```suggestion
            // Allows to have List<int>, List<Double>, List<String> etc (and the corresponding Set<?>)
```

bundles/org.openhab.core.config.core/src/test/java/org/openhab/core/config/core/ConfigParserTest.java
```diff
@@ -0,0 +1,129 @@
...
+            new TestParameter<>(1, Byte.class, (byte) 1), //
+            // boolean/Boolean
+            new TestParameter<>("true", boolean.class, true), //
+            new TestParameter<>("true", Boolean.class, true), //
+            new TestParameter<>(false, boolean.class, false), //
+            new TestParameter<>(false, Boolean.class, false), //
+            // BigDecimal
+            new TestParameter<>("7.5", BigDecimal.class, BigDecimal.valueOf(7.5)), //
+            new TestParameter<>(BigDecimal.valueOf(-7.5), BigDecimal.class, BigDecimal.valueOf(-7.5)), //
+            new TestParameter<>(1, BigDecimal.class, BigDecimal.valueOf(1)), //
```
cweitkamp, line 74:
```suggestion
            new TestParameter<>(1, BigDecimal.class, BigDecimal.ONE), //
```

bundles/org.openhab.core.config.core/src/test/java/org/openhab/core/config/core/ConfigurationTest.java
```diff
@@ -62,6 +66,7 @@ public void assertGetConfigAsWorks() {
         configuration.put("stringField", "test");
         configuration.put("enumField", "ON");
         configuration.put("listField", List.of("one", "two", "three"));
+        configuration.put("setField", List.of("one", "two", "three"));
```
cweitkamp, line 69: Shouldn't thi be as `Set`?

```suggestion
        configuration.put("setField", Set.of("one", "two", "three"));
```

bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/ConfigParser.java
```diff
@@ -0,0 +1,258 @@
...
+        }
+
+        // make sure primitives are converted to their respective wrapper class
+        Class<?> typeClass = WRAPPER_CLASSES_MAP.getOrDefault(type.getSimpleName(), type);
+
+        Object result = value;
+        // Handle the conversion case of BigDecimal to Float,Double,Long,Integer
+        if (value instanceof BigDecimal && !BigDecimal.class.equals(typeClass)) {
+            BigDecimal bdValue = (BigDecimal) value;
+            if (Float.class.equals(typeClass)) {
```
cweitkamp, line 190: This `if`-`else`-statement looks very similar to the block below and `BigDecimal` implements `Number` too thus I am wondering if we can reduce code by extracting a private method for it. Do you see any chance to apply that?

bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/ConfigParser.java
```diff
@@ -0,0 +1,258 @@
...
+        return fields;
+    }
+
+    /**
+     * Convert a value to a given type or return default value
+     *
+     * @param value input value or String representation of that value
+     * @param type desired target class
+     * @param defaultValue default value to be used if conversion fails or input value is null
+     * @return the converted value
```
cweitkamp, line 163:
```suggestion
     * @return the converted value or the given defaultValue if conversion fails or input value is null
```

cweitkamp (commented):
Thanks for your work. The feature looks really nice. Cannot wait to use it in different places.

I read through the code on last time and left some final remarks.

bundles/org.openhab.core.config.core/src/test/java/org/openhab/core/config/core/ConfigurationTest.java
J-N-K: No, I deliberately chose to use List as input to check if a List is properly converted to a Set if that is required. Maybe another Set->Set and Set->List test should be added.

bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/ConfigParser.java
J-N-K: The first `BigDecimal` case can be removed in total, it's covered by the previous check if the type is already ok and the following `Number`.

J-N-K:
Tests fails unrelated

cweitkamp (approved):
LGTM. Thanks.

PR #2681, opened 2022-01-11
[automation] Script engines now unloaded if replaced, and closed if AutoCloseable

jpg0:
This change:

- Prevents scripts (and their associated engines) from replacing themselves without proper unloading
- Closes any engines with implement `AutoCloseable`

(Required to fix https://github.com/openhab/openhab-addons/issues/11952)

Signed-off-by: Jonathan Gilbert <jpg@trillica.com>

PR #2705, opened 2022-01-24
Removes the scriptengine from local map after closing and removing a scriptengine

digitaldan:

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
@@ -226,6 +226,7 @@ public class ScriptEngineManagerImpl implements ScriptEngineManager {
             }
 
             removeScriptExtensions(engineIdentifier);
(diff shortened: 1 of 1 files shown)


digitaldan:
After we clean up a script engine and close it, we are not removing it from the local map/cache and the script manager tries to reuse it, which it can not and throws an error.  See my comment on https://github.com/openhab/openhab-core/pull/2681#issuecomment-1019688489

@jpg0 does this seem reasonable?

jpg0:
@digitaldan that's certainly possible - I can see that it's not explicitly removed. The problem that I was fixing is that new ones were getting inserted over the top of old ones (so I guess implicit removal), and I didn't try to change that behaviour, only close the engines properly when requested. Maybe this change has uncovered an existing bug that closed engines were being reused but we didn't realise because they were not actually closed properly? I will certainly take a look - possibly just removal from the map at the end of `removeEngine` is all that is needed.

digitaldan:
Thanks @jpg0 ! indeed https://github.com/openhab/openhab-core/pull/2706/files looks like a better solution.

jpg0:
Oh, I just realised that I has commented on a PR which did exactly the same as mine! Sorry! At least as you suggested it looks better - I purposefully removed from the map first so that any other threads couldn't so easily get at an engine which is in the process of being closed.

PR #2709, opened 2022-01-26
[REST] Fixed wrong openAPI schema for /inbox

### LINKED ISSUE #2707: [REST] GET /inbox wrong Swagger Schema
Hello,
it seems that the API Explorer shows the wrong Schema for the `GET /inbox` endpoint, namely `DiscoveryResultDTO` instead of `[DiscoveryResultDTO]`.
image
The `Response body` shows that an array should be returned (which is also what actually happens), only the `Schema` says that it is only one object instead of an array of `DiscoveryResultDTOs`.

Compare this for example to the `GET /items` endpoint where the `Schema` shows the result to be an array.
image

cweitkamp:
- Fixed wrong openAPI schema for /inbox

Fixes #2707


wborn (approved):
Thanks!

PR #2712, opened 2022-01-29
[config] Enable config validation for updates by handler

J-N-K:
See https://github.com/openhab/openhab-core/pull/2682#discussion_r795035470

It was assumed that thing handlers only update valid configuration. After merging #2682 it was discovered that this is not true. This PR enables the same validation for updates through the handler. If an invalid configuration or otherwise illegal (not-initializable) thing is detected, the changes are discarded and a warning logged.

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/BaseThingHandler.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/BaseThingHandler.java
@@ -242,7 +242,7 @@ public abstract class BaseThingHandler implements ThingHandler {
      * Updates the state of the thing. Will use the thing UID to infer the
      * unique channel UID from the given ID.
      *
-     * @param channel ID id of the channel, which was updated
+     * @param channelID id of the channel, which was updated
      * @param state new state
      */
     protected void updateState(String channelID, State state) {
@@ -392,13 +392,25 @@ public abstract class BaseThingHandler implements ThingHandler {
             throw new IllegalArgumentException(
                     "Changes must not be done on the current thing - create a copy, e.g. via editThing()");
         }
+        ThingHandlerCallback callback = this.callback;
+        if (callback == null) {
+            logger.warn("Handler {} tried updating thing {} although the handler was already disposed.",
+                    this.getClass().getSimpleName(), thing.getUID());
+            return;
+        }
+        try {
+            callback.validateConfigurationParameters(thing, thing.getConfiguration().getProperties());
+        } catch (ConfigValidationException e) {
+            logger.warn(
+                    "Attempt to update thing '{}' with a thing containing invalid configuration '{}', blocked. This is most likely a bug.",
+                    thing.getUID(), thing.getConfiguration());
+            return;
+        }
         synchronized (this) {
             if (this.callback != null) {
                 this.thing = thing;
                 this.callback.thingUpdated(thing);
             } else {
-                logger.warn("Handler {} tried updating thing {} although the handler was already disposed.",
-                        this.getClass().getSimpleName(), thing.getUID());
             }
         }
     }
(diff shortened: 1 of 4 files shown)


itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/binding/BindingBaseClassesOSGiTest.java
```diff
@@ -606,6 +611,36 @@ public void assertConfigurationParametersAreValidated() {
...
+        assertNotNull(handler);
+
+        handler.updateConfiguration(new Configuration(Map.of("parameter", "otherValue")));
+        Object parameter = handler.getThing().getConfiguration().get("parameter");
+        assertNotNull(parameter);
+        assertEquals("otherValue", parameter);
+
+        handler.updateConfiguration(new Configuration(Map.of()));
+        // configuration should not change
+        assertNotNull(parameter);
```
cweitkamp, line 640: You should get the parameter again before applying the assertion.

```suggestion
        parameter = handler.getThing().getConfiguration().get("parameter");
        assertNotNull(parameter);
```

itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/binding/BindingBaseClassesOSGiTest.java
```diff
@@ -606,6 +611,36 @@ public void assertConfigurationParametersAreValidated() {
...
+
+        ThingTypeUID thingTypeUID = new ThingTypeUID("bindingId:type");
+        ThingUID thingUID = new ThingUID("bindingId:type:thingId");
+        Thing thing = ThingBuilder.create(thingTypeUID, thingUID)
+                .withConfiguration(new Configuration(Map.of("parameter", "someValue"))).build();
+
+        managedThingProvider.add(thing);
+
+        SimpleThingHandler handler = (SimpleThingHandler) thing.getHandler();
+        assertNotNull(handler);
```
cweitkamp, line 631: Should we add an initial assertion for the parameter equals to `"some value"`?

cweitkamp (commented):
U have the slight feeling that this will it work because the `BaseThingHandler:: update configuration ()` methods replaces the configuration properties before calling the `updateThing()` methods on the callback.

https://github.com/openhab/openhab-core/blob/16bc9321f18ff2b17aaaf8c0800d8abb8b2a8867/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/BaseThingHandler.java#L425-L434

Meaning the wrong parameters will be applied even if you discard the changes later.

J-N-K:
I reworked that. Now the `BaseThingHandler` validates the configuration before they are applied or updated (both in `updateThing` and `updateConfiguration`). There is a slight chance that an invalid configuration is applied anyway if the implementation of the `ThingHandler` overrides these methods and does not do a validation. 

Do you think `updateProperties` should also validate its parameters?

itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/binding/BindingBaseClassesOSGiTest.java
J-N-K: done

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/BaseThingHandler.java
```diff
@@ -392,13 +392,25 @@ protected void updateThing(Thing thing) {
...
+        try {
+            callback.validateConfigurationParameters(thing, thing.getConfiguration().getProperties());
+        } catch (ConfigValidationException e) {
+            logger.warn(
+                    "Attempt to update thing '{}' with a thing containing invalid configuration '{}', blocked. This is most likely a bug.",
+                    thing.getUID(), thing.getConfiguration());
+            return;
+        }
         synchronized (this) {
             if (this.callback != null) {
```
cweitkamp, line 410: This check can be omitted by reusing the local variable `callback` defined earlier.

cweitkamp:
> Do you think `updateProperties` should also validate its parameters?

No, I don't think so. Do you see a specific reason for doing it?

J-N-K:
After checking the code: no. I was wondering if it is possible to change configuration with `updateProperties`, but it seems that the only place where properties and configuration are mixed is when building a discovery result. After the thing is aded from the INBOX, they are separated.

cweitkamp:
If you are afraid that we loose this safety check by overridden methods we can try to move it deeper into the framework. E.g. into the `ThingHandlerCallback::updateThing()` method.

J-N-K:
But then we are back where we were before: the wrong config is applied before it is validated. I think that overriding `updateThing` or `updateConfiguration` is no very common and if it is sone should be secured by review.

cweitkamp:
Let's add a note about that to the JavaDoc. This way people know what the method is doing and what to take care of if they override it.

J-N-K:
Done

cweitkamp (approved):
LGTM. Thanks.

Issue #2710, opened 2022-01-28
Config validation preventing things initialising


cdjackson:
This issue https://github.com/openhab/openhab-core/pull/2682 seems to prevent things with configuration starting. It seems to be a common problem across multiple bindings - users are reporting that saving the configuration, even after making no changes - will bring the thing online.

I've checked that default values in some devices are within bounds, so there should be no reason for these devices to initialise and I guess there is something wrong with the initialisation check.

Given the extent of this issue it probably should be considered as critical.

See the forum thread for ZWave discussion -:
https://community.openhab.org/t/my-z-wave-is-borked/132448/38

J-N-K:
Is there an easy way to reproduce the issue? I don‘t see that in my setup.

cdjackson:
Yes, it seems easily reproducible based on what I read on the forum - many users are reporting this.

I don’t currently have an OH system with this code, so I personally can’t reproduce it myself, but according to discussions on the forum, this is hitting most Zwave devices, and at least also Homematic so I would be surprised if other devices with parameters being checked aren’t also impacted by this.

cweitkamp:
Very soon after merging #2682 I realized that using String values for configuration parameters of other types is not possible anymore. e.g. `"1"` s not accepted for an `int` parameter or `"true"` for `boolean` whereas `1` and `true` are valid. This clearly is a regression and I submitted https://github.com/openhab/openhab-core/pull/2699 to fix it. This fix might help to solve some of the issues (or all of them).

J-N-K:
I'm not 100% sure this is the root cause here in both cases. HomeMatic has a very special way of defining things, it adds a `ThingTypeProvider` and a `ConfigDescriptionProvider` and only defines the bridge in an XML. I could imagine that this can cause problems, but unfortunately I'm not able to debug that because I don't have a HomeMatic Gateway to test.

Anyway #2699 should be merged ASAP and we can see if wee need to do further investigations.

petero-dk:
I have this issue also. On some devices it was enough to save the thing (i.e. the z-wave serial controller) but for others i.e. z-wave switch which keeps reverting settings to values that are out of range.
![image](https://user-images.githubusercontent.com/2478689/151653273-8613ac8e-236f-4ed7-89d1-d39dbb91d760.png)

Which logs can I provide?

J-N-K:
@petero-dk Which device is that?

petero-dk:
[...]

PR #2713, opened 2022-01-29
[config] improve type validation message

J-N-K:
The message did not contain the detected wrong type which makes it harder to define what the actual error is.

Found while investigating #2710.


petero-dk:
Ok, so two follow up questions, how did my system get in a wrong state like this? And how should I go about fixing it?

cweitkamp:
Can you please change the properties file for translations too? Thanks.

J-N-K:
How do I change the translations for languages I don't understand? I also fixed a typo for numeric validations.

cweitkamp:
You do not have to take care about other languages than default translation file in a PR. It will be uploaded to Crowdin and further process will happen over there.

cweitkamp (approved):
LGTM. Thanks.

PR #2724, opened 2022-02-01
allow multiple listeners for same directory in WatchService

### LINKED ISSUE #2466: AbstractWatchService can't handle two listeners for same directory

J-N-K:
Fixes #2466

Also adds null-annotations.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/service/WatchQueueReader.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/service/WatchQueueReader.java
@@ -338,9 +338,9 @@ public class WatchQueueReader implements Runnable {
      * on Stack Overflow.
      *
      * @param key the {@link WatchKey}
-     * @param event the {@link }WatchEvent} itself
+     * @param event the {@link WatchEvent} itself
      * @param resolvedPath the resolved {@link Path} for this event
-     * @param services a {@link Set<AbstractWatchService>} that subscribe to this event
+     * @param services the {@link AbstractWatchService}s that subscribe to this event
      */
     private void processModificationEvent(WatchKey key, WatchEvent<?> event, Path resolvedPath,
             Set<AbstractWatchService> services) {
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/service/WatchQueueReader.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/service/WatchQueueReader.java
@@ -12,7 +12,10 @@
  */
 package org.openhab.core.service;
 
-import static java.nio.file.StandardWatchEventKinds.*;
+import static java.nio.file.StandardWatchEventKinds.ENTRY_CREATE;
+import static java.nio.file.StandardWatchEventKinds.ENTRY_DELETE;
+import static java.nio.file.StandardWatchEventKinds.ENTRY_MODIFY;
+import static java.nio.file.StandardWatchEventKinds.OVERFLOW;
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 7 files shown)


J-N-K:
@cweitkamp I would rather consider it a bug. It's not documented that adding a second watch service for the same (or a parent directory) disables the original watch service.

bundles/org.openhab.core/src/main/java/org/openhab/core/service/WatchQueueReader.java
```diff
@@ -269,75 +315,79 @@ public void run() {
...
+     * on Stack Overflow.
      *
-     * @param key
-     * @param event
-     * @param resolvedPath
-     * @param service
+     * @param key the `WatchKey`
+     * @param event the `WatchEvent` itself
+     * @param resolvedPath the resolved `Path` for this event
+     * @param services a `Set` of `AbstractWatchServices` that subscribe to this event
```
wborn, line 343: JavaDocs are not Markdown, instead of 

```
`WatchKey`
```
 use something like `{@link WatchKey}` or `{@code WatchKey}`

bundles/org.openhab.core/src/main/java/org/openhab/core/service/WatchQueueReader.java
```diff
@@ -269,75 +315,79 @@ public void run() {
 
                 key.reset();
             } catch (InterruptedException exc) {
-                logger.debug("WatchQueueReader interrupted; thread exiting: {}", exc.getLocalizedMessage());
+                logger.info("Caught InterruptedException. Shutting down.");
```
wborn, line 318: This is expected to happen when the runtime is shutdown:

```suggestion
                logger.debug("Caught InterruptedException. Shutting down.");
```

J-N-K:
Thanks for the review. I addressed your comments. A related (but independent) issue: the tests for this class reside in the core integration tests bundle, but are not executed because they are in `main` instead of `test`. Any idea how to fix it? Moving them to `org.openhab.core` is not possible because they depend on `org.openhab.core.test` (the class extends `JavaTest`).

bundles/org.openhab.core/src/main/java/org/openhab/core/service/WatchQueueReader.java
```diff
@@ -269,75 +315,79 @@ public void run() {
...
+     * on Stack Overflow.
      *
-     * @param key
-     * @param event
-     * @param resolvedPath
-     * @param service
+     * @param key the {@link WatchKey}
+     * @param event the {@link }WatchEvent} itself
+     * @param resolvedPath the resolved {@link Path} for this event
+     * @param services a {@link Set<AbstractWatchService>} that subscribe to this event
```
wborn, line 343:
```suggestion
     * @param event the {@link WatchEvent} itself
     * @param resolvedPath the resolved {@link Path} for this event
     * @param services the {@link AbstractWatchService}s that subscribe to this event
```

wborn (approved):
Thanks!

wborn:
> Any idea how to fix it?

I see you already fixed it by moving the class to `src/test/java`.

IIRC when I ran into this issue when writing tests in `org.openhab.core` I just duplicated the small bits and pieces I needed from the `JavaTest` class in my test (which usually isn't a lot).

To break the cycle, the `SyntheticBundleInstaller` and `VolatileStorageService` should be moved to some other bundle as those classes depend on `org.openhab.core`. But that would make it less easy to reuse them when writing tests in all other bundles.

Another way could be to just duplicate `JavaTest` in `src/test/java` of  `org.openhab.core` because it is not a lot of code anyways. :wink:

cweitkamp:
Currently the openHAB Add-ons build is failing because of a conflicting nullness annotation in the `ExecWhitelistWatchService`. Can you please check that? Thanks.

```
[ERROR] /home/jenkins/jenkins-agent1/workspace/openHAB-Addons/bundles/org.openhab.binding.exec/src/main/java/org/openhab/binding/exec/internal/ExecWhitelistWatchService.java:[62,15] The default '@org.eclipse.jdt.annotation.NonNull' conflicts with the inherited '@org.eclipse.jdt.annotation.Nullable' annotation in the overridden method from org.openhab.core.service.AbstractWatchService
```

https://ci.openhab.org/job/openHAB-Addons/623/

J-N-K:
I have added the necessary changes to [openhab-addons#12221](https://github.com/openhab/openhab-addons/pull/12221), otherwise none of the PR would build.

The build failed on wrong null-annotations.

Issue #2702, opened 2022-01-22
[audio] JavaSoundAudioSource returning closed AudioStream


GiviMAD:
This came up while testing the pr #2693.
The method getInputStream of this class initialize the TargetDataLine for the microphone just on its first execution, storing the value and returning the same for next executions. This method returns a JavaSoundInputStream that wraps the microphone, and it's closing the microphone in its method close, causing that next calls to getInputStream return and stream that doesn't work as it's underling microphone instance is already closed.

Related code:

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/javasound/JavaSoundAudioSource.java
```
    @Override
    public synchronized AudioStream getInputStream(AudioFormat expectedFormat) throws AudioException {
        if (!expectedFormat.isCompatible(audioFormat)) {
            throw new AudioException("Cannot produce streams in format " + expectedFormat);
        }
        TargetDataLine mic = this.microphone;
        if (mic == null) {
            mic = initMicrophone(format);
        }
        return new JavaSoundInputStream(mic, audioFormat);
    }
```

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/javasound/JavaSoundInputStream.java
```
    @Override
    public void close() throws IOException {
        input.close();
    }
```
I think that a new microphone should be initialize on each call to getInputStream as it will be disposed when the returned stream get closed. Also in the DialogProcessor we are getting two stream and consuming it from two different threads and I don't think that sharing the same underling instance is thread save which I think was causing the Speech-to-Text service I was testing to miss some part of the audio.
I've been testing to not sharing the microphone instance and everything seems to work fine (tested on a Mac, my development environment).
Could this be changed? Or there is a reason to avoid opening multiple instances of the microphone?
Thank you in advance.

GiviMAD:
I've testing the 3.3.0M1 and same issue is there. Some words on the transcriptions are wrong. 
@kaikreuzer, when you have a moment, can you give us your opinion about stop sharing the mic TargetDataLine?
@lolodomo, Did you have a chance to test this? 
I would like to solve this before asking for feedback about the new voice add-ons on the forum, sorry for the hurry

kaikreuzer:
Hi @GiviMAD, I am no expert at all on how to correctly deal with the microphone input stream. From what you describe, I'd consider it indeed a bug in the current code and if it all works fine when correctly closing the streams and instantiating new ones where ever needed, I'd be all for it. 👍

[...]

PR #2732, opened 2022-02-05
[javasound/dialogprocessor] Not share mic ref on javasound + close audio streams and use RecognitionStartEvent on dialogprocessor

GiviMAD:
This is a proposed solution for https://github.com/openhab/openhab-core/issues/2702 about the system audio (javasound) returning closed audio streams after the first audio stream is closed. 
Also includes the use of RecognitionStartEvent on the dialog processor to wait for the STT add-on to be ready for toggling the "listening" item, so everything is really ready for the user to start talking.
Hope everything is in place. I have tested this changes with the porcupineKS and googleSTT add-ons.

### DIFF
--- a/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/javasound/JavaSoundAudioSource.java
+++ b/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/javasound/JavaSoundAudioSource.java
@@ -16,7 +16,6 @@ import java.util.Locale;
 import java.util.Set;
 
 import javax.sound.sampled.AudioSystem;
-import javax.sound.sampled.DataLine;
 import javax.sound.sampled.TargetDataLine;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
@@ -32,6 +31,7 @@ import org.osgi.service.component.annotations.Component;
  *
  * @author Kelly Davis - Initial contribution and API
  * @author Kai Kreuzer - Refactored and stabilized
+ * @author Miguel Álvarez - Don't share TargetDataLine
  *
  */
 @NonNullByDefault
@@ -50,11 +50,6 @@ public class JavaSoundAudioSource implements AudioSource {
     private final AudioFormat audioFormat = convertAudioFormat(format);
 
     /**
-     * TargetDataLine for the mic
-     */
-    private @Nullable TargetDataLine microphone;
-
-    /**
      * Constructs a JavaSoundAudioSource
      */
     public JavaSoundAudioSource() {
@@ -63,13 +58,7 @@ public class JavaSoundAudioSource implements AudioSource {
     private TargetDataLine initMicrophone(javax.sound.sampled.AudioFormat format) throws AudioException {
         try {
             TargetDataLine microphone = AudioSystem.getTargetDataLine(format);
-
-            DataLine.Info info = new DataLine.Info(TargetDataLine.class, format);
-            microphone = (TargetDataLine) AudioSystem.getLine(info);
-
             microphone.open(format);
-
-            this.microphone = microphone;
             return microphone;
         } catch (Exception e) {
             throw new AudioException("Error creating the audio input stream.", e);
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/javasound/JavaSoundAudioSource.java
```diff
@@ -63,13 +58,7 @@ public JavaSoundAudioSource() {
     private TargetDataLine initMicrophone(javax.sound.sampled.AudioFormat format) throws AudioException {
         try {
             TargetDataLine microphone = AudioSystem.getTargetDataLine(format);
-
-            DataLine.Info info = new DataLine.Info(TargetDataLine.class, format);
```
kaikreuzer, line 67: Looking at this code now, I wonder whether there wasn't a good reason to have it.
I personally develop on macOS as well, so the simpler code might just work out.
But I remember that Kelly Davis developed on Linux and the Linux audio system behaved quite differently to macOS.
Before merging this, I'd hence think it should be tested on Linux (and possibly Windows) as well. Do you have any chance to do so?

GiviMAD: Sure, I can perform those tests tomorrow.
I'll ping you with the results.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
```diff
@@ -160,8 +163,11 @@ private void abortKS() {
     private void closeStreamKS() {
         AudioStream stream = streamKS;
         if (stream != null) {
-            // Due to an issue in JavaSoundAudioSource ( https://github.com/openhab/openhab-core/issues/2702 )
-            // we do not try closing the stream
+            try {
+                stream.close();
+            } catch (IOException e) {
+                logger.warn("IOException closing ks audio stream: {}", e.getMessage(), e);
```
lolodomo, line 169: Full stack trace with a WARN level if the stream failed to close us maybe a little too much?

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
```diff
@@ -178,8 +184,11 @@ private void abortSTT() {
     private void closeStreamSTT() {
         AudioStream stream = streamSTT;
         if (stream != null) {
-            // Due to an issue in JavaSoundAudioSource ( https://github.com/openhab/openhab-core/issues/2702 )
-            // we do not try closing the stream
+            try {
+                stream.close();
+            } catch (IOException e) {
+                logger.warn("IOException closing stt audio stream: {}", e.getMessage(), e);
```
lolodomo, line 190: Same here

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
```diff
@@ -201,7 +210,6 @@ public void ksEventReceived(KSEvent ksEvent) {
         if (!processing) {
             isSTTServerAborting = false;
             if (ksEvent instanceof KSpottedEvent) {
-                toggleProcessing(true);
```
lolodomo, line 204: Please also remove in this case the two calls to toggleProcessing with false in the 2 catch in the code just after. They are now more needed.

lolodomo (changes requested):

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/javasound/JavaSoundAudioSource.java
GiviMAD: You were right. There are some limitations for open multiple TargetDataLine for same format, but under Windows I found comments reporting same problem for other projects. I also discover there are some problems on windows with the porcupineKS add-on, I'll look at them, so the tests where really meaningful. On linux (Ubuntu 20.04.3 LTS) everything works right. I didn't found any comments reporting the problem on linux.
I'll update the pr to keep sharing the microphone under windows, but closing it correctly when no more streams depend on it.

kaikreuzer: Alright, sounds good - thanks for the elaborate testing!

GiviMAD: I have updated the implementation to behave different on windows.

lolodomo:
The code changes related to dialog processor are fine for me. 
I have not checked the changes related to audio.

kaikreuzer (approved):
Thank you!

PR #2735, opened 2022-02-05
fix removal of thing properties (single and bulk)

### LINKED ISSUE #2623: [Bindings] Deleting thing properties from handler, weird behavior
Hi, I'm adding a functionality to the androiddebugbridge binding where I add/remove some thing properties to store some device specific commands. https://github.com/openhab/openhab-addons/pull/11692

 I can easily create/update properties calling the updateProperty method of the base class (BaseThingHandler).
 ```java
            updateProperty(recordPropertyName, eventCommand);
 ```

To  remove a property I'm doing this which is quite tricky, but is the only way I found to get the property correctly removed:
```java
                this.getThing().setProperty(recordPropertyName, null);
                updateProperties(editProperties());
```
This is because the updateProperties function expects a map with non nullable values (and the updateProperty function a non nullable value).

This which seems more correct doesn't work.
```java
               // NOT WORKING
                var currentProps = editProperties();
                currentProps.remove(recordPropertyName);
                updateProperties(currentProps);
```

I think the ideal solution will be to change the updateProperties implementation so it actually remove the removed properties apart from the null ones, so the change won't break the current behavior. Also adding a function to remove a single property would be nice (like the updateProperty one).

Hope the explanation is clear, let me know if I'm missing something.

J-N-K:
Fixes #2623 

Since it only extends the allowed parameters to be @Nullable, I don't think it will cause any downstream-problems.

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/BaseThingHandler.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/BaseThingHandler.java
@@ -12,7 +12,6 @@
  */
 package org.openhab.core.thing.binding;
 
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -492,7 +491,7 @@ public abstract class BaseThingHandler implements ThingHandler {
             this.thing.setProperties(Map.of());
             propertiesUpdated = true;
         } else if (properties != null) {
-            for (Entry<String, String> property : properties.entrySet()) {
+            for (Entry<String, @Nullable String> property : properties.entrySet()) {
                 String propertyName = property.getKey();
                 String propertyValue = property.getValue();
                 String existingPropertyValue = thing.getProperties().get(propertyName);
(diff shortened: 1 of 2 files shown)


J-N-K:
Test fails unrelated

cweitkamp:
Be careful with this fix. I gave it a try couple of months ago but realized that it does have an impact on add-ons build. If you try to build this fixed OHC against the openHAB Add-Ons repository it fails early

```
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/christoph/openhab/git/openhab-addons/bundles/org.openhab.binding.alarmdecoder/src/main/java/org/openhab/binding/alarmdecoder/internal/handler/ADBridgeHandler.java:[321,26] Null type mismatch (type annotations): required 'java.util.@Nullable Map<java.lang.@NonNull String,java.lang.@Nullable String>' but this expression has type 'java.util.Map<java.lang.@NonNull String,java.lang.@NonNull String>'
[INFO] 1 error
[INFO] -------------------------------------------------------------
```

This afair happens when the the `editProperties()` method is used to initialize a local variable to be passed to the `updatePorperties(Map<String, String>)` method - which is a common way to got to edit properties.

```
        Map<String, String> properties = editProperties();
        properties.put(propertyName, propertyValue);
        updateProperties(properties);
```

J-N-K:
I‘ll check that. IMO it would be enough to have `updateProperties(@Nullable Map<String, Object>)` and `updateProperty(String name, @Nullable Object value)`. It allows to delete all properties by passing `null` instead of a `Map` and also deleting a single property. What would be missing is the ability to delete several properties at once and keeping other properties at the same time. But I believe that is not a very common case (keeping in mind that the current code does not allow deletions at all).

J-N-K:
This is still not 100% backward compatible because handlers overriding one of the methods need adjustments, but this is a very low number. I have created https://github.com/openhab/openhab-addons/pull/12221 for the necessary changes.

cweitkamp (approved):
LGTM. Thanks.

PR #2739, opened 2022-02-06
[expiry] extend ExpiryManager to allow ignoring state updates

### LINKED ISSUE #1994: Expire does not work in some situations
I have this item file
```
Switch aa_test1 {channel="daikin:ac_unit:sala:power", expire="2m,command=OFF"}
Switch aa_test2 {expire="2m,command=OFF"}
```
If I switch on both items I get from the logs
```
2020-12-23 11:03:19.993 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'aa_test1' received command ON
2020-12-23 11:03:19.999 [INFO ] [penhab.event.ItemStatePredictedEvent] - Item 'aa_test1' predicted to become ON
2020-12-23 11:03:20.005 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'aa_test1' changed from OFF to ON
2020-12-23 11:03:26.545 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'aa_test2' received command ON
2020-12-23 11:03:26.548 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'aa_test2' changed from OFF to ON
2020-12-23 11:05:27.168 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'aa_test2' received command OFF
2020-12-23 11:05:27.170 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'aa_test2' changed from ON to OFF
```
Item `aa_test2` (not linked to a channel) expired after 2 minutes but `aa_test1` (linked to a channel) did not.

### LINKED ISSUE #2542: [expire] Add an optional flag to have Expire reschedule on change as opposed to update
NOTE: this might be a duplicate of #1994 and I can close it if that's the case.

The problem is that lots of bindings will update Items periodically, sometimes as often as once a minute, even when the Item's state has not changed. Unfortunately, if one sets an Expire timeout greater than how often the binding updates the Item that means that the Expire will never happen.

If we had a flag to tell Expire that we only care about changes, not updates, we can support use cases like commanding the bathroom fan to OFF after five minutes of it being ON even when the binding keeps updating the Item every minute.

Looking at [ExpireManager](https://github.com/openhab/openhab-core/blob/79edf2b9e643ad802c31b2e5f54edb93534277fd/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ExpireManager.java) I think it would just be a matter:

- Adding the optional flag to ExpireConfig class
-  In processEvent, if the change flag is set ignore events of type ItemStateUpdate

J-N-K:
Fixes #1994 
Fixes #2542 

This needs an additional PR in UI to make it configurable via UI.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ExpireManager.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ExpireManager.java
@@ -288,6 +288,7 @@ public class ExpireManager implements EventSubscriber, RegistryChangeListener<It
     }
 
     static class ExpireConfig {
+        static final String CONFIG_IGNORE_STATE_UPDATES = "ignoreStateUpdates";
 
         private static final StringType STRING_TYPE_NULL_HYPHEN = new StringType("'NULL'");
         private static final StringType STRING_TYPE_NULL = new StringType("NULL");
@@ -329,7 +330,7 @@ public class ExpireManager implements EventSubscriber, RegistryChangeListener<It
                     ? configString.substring(commaPos + 1).trim()
                     : null;
 
-            Object ignoreStateUpdatesConfigObject = configuration.get("ignoreStateUpdates");
+            Object ignoreStateUpdatesConfigObject = configuration.get(CONFIG_IGNORE_STATE_UPDATES);
             if (ignoreStateUpdatesConfigObject instanceof String) {
                 ignoreStateUpdates = Boolean.parseBoolean((String) ignoreStateUpdatesConfigObject);
             } else if (ignoreStateUpdatesConfigObject instanceof Boolean) {
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ExpireManager.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ExpireManager.java
@@ -41,6 +41,7 @@ import org.openhab.core.items.MetadataRegistry;
 import org.openhab.core.items.events.GroupItemStateChangedEvent;
 import org.openhab.core.items.events.ItemCommandEvent;
 import org.openhab.core.items.events.ItemEventFactory;
+import org.openhab.core.items.events.ItemStateChangedEvent;
 import org.openhab.core.items.events.ItemStateEvent;
 import org.openhab.core.library.types.StringType;
 import org.openhab.core.types.Command;
@@ -77,10 +78,10 @@ public class ExpireManager implements EventSubscriber, RegistryChangeListener<It
 
     private final Logger logger = LoggerFactory.getLogger(ExpireManager.class);
 
-    private Map<String, @Nullable Optional<ExpireConfig>> itemExpireConfig = new ConcurrentHashMap<>();
-    private Map<String, Instant> itemExpireMap = new ConcurrentHashMap<>();
+    private final Map<String, @Nullable Optional<ExpireConfig>> itemExpireConfig = new ConcurrentHashMap<>();
+    private final Map<String, Instant> itemExpireMap = new ConcurrentHashMap<>();
 
-    private ScheduledExecutorService threadPool = ThreadPoolManager
+    private final ScheduledExecutorService threadPool = ThreadPoolManager
             .getScheduledPool(ThreadPoolManager.THREAD_POOL_NAME_COMMON);
 
     private final EventPublisher eventPublisher;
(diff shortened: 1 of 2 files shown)


J-N-K:
I don't understand why the test fails. I can't get it to fail on my MacBook Air M1 and not on an old Windows (i5-4570) machine. Neither with serial build nor with parallel build.

mherwege:
Don't you have to change the [event types subscriptions](https://github.com/openhab/openhab-core/blob/c9fe8c0e7467090e7a1c9d923abc9129aea0ffb4/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ExpireManager.java#L76-L77) as well? If I understand it well, the ExpiryManager does not subscribe to the ItemStateChangedEvents at the moment. Or am I missing something?

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ExpireManager.java
```diff
@@ -240,13 +244,13 @@ public void receive(Event event) {
         }
         if (event instanceof ItemStateEvent) {
             ItemStateEvent isEvent = (ItemStateEvent) event;
-            processEvent(isEvent.getItemName(), isEvent.getItemState());
+            processEvent(isEvent.getItemName(), isEvent.getItemState(), true);
         } else if (event instanceof ItemCommandEvent) {
             ItemCommandEvent icEvent = (ItemCommandEvent) event;
-            processEvent(icEvent.getItemName(), icEvent.getItemCommand());
-        } else if (event instanceof GroupItemStateChangedEvent) {
```
mherwege, line 247: Is there a specific reason to remove this? I can see a use case for the expire functionality on a group (e.g. to turn off all lights in a group if they stay on for too long).

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ExpireManager.java
```diff
@@ -166,7 +170,7 @@ private void processEvent(String itemName, Type type) {
     private @Nullable ExpireConfig getExpireConfig(String itemName) {
         Optional<ExpireConfig> itemConfig = itemExpireConfig.get(itemName);
         if (itemConfig != null) {
-            return itemConfig.isPresent() ? itemConfig.get() : null;
+            return itemConfig.orElse(null);
```
mherwege, line 173: return itemConfig.get() may be enough. You should remove the Nullable annotation on the itemExpireConfig map value. As it is a ConcurrentHashMap, null values are not allowed.

mherwege:
I am just wondering if explictely making the syntax 'ignoreStateUpdates=true' or 'ignoreStateUpdates=false' (with a default of false) would not be cleaner. It also further reduces the (very limited) risk of conflict with a '..., ignoreStateUpdates' channel update value. I was on a path implementing that before I noticed you had already done this.

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ExpireManager.java
mherwege: Mistake on my side, you do need the code you put in, but you can still remove the @Nullable annotation from the declaration.

cweitkamp:
The syntax `expire="<duration>[state|command]=<stateOrCommand>" was kept for backwards compatibility. As we are using expiry as metatdata information new feature should use metadata related syntax and pass additional configurations like this:

```
ga="Light" [ inverted=true ]
```

Meaning we now should use a syntax like this:
```
expire="<duration>[,(state|,command)=<stateOrCommand>]" [ignoreStateUpdates=true]
```

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ExpireManager.java
```diff
@@ -325,6 +329,15 @@ public ExpireConfig(Item item, String configString) throws IllegalArgumentExcept
                     ? configString.substring(commaPos + 1).trim()
                     : null;
 
+            Object ignoreStateUpdatesConfigObject = configuration.get("ignoreStateUpdates");
```
cweitkamp, line 332: Did you remember the `ConfigParser`? Will be a one liner if you use it.

And please refactor the String `"ignoreStateUpdates"` to be a static final constant.

bundles/org.openhab.core/src/test/java/org/openhab/core/internal/items/ExpireManagerTest.java
```diff
@@ -128,6 +130,55 @@ void testCancelExpiry() throws InterruptedException, ItemNotFoundException {
...
+        verify(eventPublisherMock, never()).post(any());
+        Thread.sleep(2000L);
+        verify(eventPublisherMock, times(1)).post(any());
+    }
+
+    @Test
+    void testIgnoreStateUpdateExtendsExpiryOnCommand() throws InterruptedException, ItemNotFoundException {
+        Item testItem = new NumberItem(ITEMNAME);
+        when(itemRegistryMock.getItem(ITEMNAME)).thenReturn(testItem);
+        when(metadataRegistryMock.get(METADATA_KEY)).thenReturn(config("2s", Map.of("ignoreStateUpdates", true)));
```
cweitkamp, line 154: The const can be reused in tests.

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ExpireManager.java
J-N-K: Yes, but it's not available here. It's part of `org.openhab.core.config.core` which is not a dependency of `org.openhab.core`.

cweitkamp: Oh, I see. I missed that

cweitkamp (approved):
LGTM. Thanks.

PR #2740, opened 2022-02-07
[discovery] keep format of thing properties

### LINKED ISSUE #1098: Discovery result properties issue
On a thing discovery, all properties that are created as long objects, after a thing is approved and created, are presented in thing properties as floating point numbers (in string representation). For example a property equal to long 50 in DiscoveryResult, in thing will be 50.0

The same property after subsequent thing initialization is equal to 50 not 50.0. 

This caused issue for example in LIFX binding when product ID property was parsed as long from string and on first run after discovery product was incorrectly determined. 

I suppose it should be the same representation any time. 

I look at the code but could not find a clear reason yet. Thing properties seem to be created from discovery result using String.valueOf which should return good value without fraction part. Any hints where it is converted to such form?

J-N-K:
Fixes #1098 

This fixes the reported problem with thing properties having a wrong representation. When adding a thing to the `PersistentInbox` the config description is retrieved and all non-configuration properties are converted to `String`. Properties with the same name as a configuration parameter stay unchanged.

There is one thing to consider:

The implementation relies on a config description being available. This is currently not enforced as it breaks a lot of tests for the persistent inbox. IMO it would be ok to not accept discovery results if the thing-type is missing in the registry or the config-description is not available.

### DIFF
--- a/bundles/org.openhab.core.config.discovery/src/test/java/org/openhab/core/config/discovery/internal/PersistentInboxTest.java
+++ b/bundles/org.openhab.core.config.discovery/src/test/java/org/openhab/core/config/discovery/internal/PersistentInboxTest.java
@@ -18,8 +18,9 @@ import static org.junit.jupiter.api.Assertions.*;
 import static org.mockito.ArgumentMatchers.*;
 import static org.mockito.Mockito.*;
 
-import java.math.BigDecimal;
 import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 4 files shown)


lolodomo:
No review of this PR?

bundles/org.openhab.core.config.discovery/src/test/java/org/openhab/core/config/discovery/internal/PersistentInboxTest.java
```diff
@@ -102,6 +103,8 @@ public void setup() {
     }
 
     @Test
+    @Disabled
+    // TODO: check if test should be removed
```
wborn, line 107: Did you already check this?

bundles/org.openhab.core.test.magic/src/main/java/org/openhab/core/magic/binding/internal/MagicDiscoveryService.java
```diff
@@ -42,10 +42,12 @@ public MagicDiscoveryService() throws IllegalArgumentException {
     @Override
     protected void startScan() {
         String serialNumber = createRandomSerialNumber();
+        // TODO: revert change here
```
wborn, line 45: Can it be reverted?

wborn (commented):
The changes look good to me, but there are still some TODOs.

bundles/org.openhab.core.config.discovery/src/test/java/org/openhab/core/config/discovery/internal/PersistentInboxTest.java
J-N-K: Yes. It's not needed any more, no guessing for properties and the logic for non-properties is the same.

bundles/org.openhab.core.test.magic/src/main/java/org/openhab/core/magic/binding/internal/MagicDiscoveryService.java
J-N-K: Yes, done.

wborn (approved):
Thanks for the fix!

PR #2742, opened 2022-02-08
Add more null annotations

wborn:
Adds null annotations to most of the tests as well as a few other classes.
Also fixes a few other SAT findings.
Fixes ~300 SAT findings in total.

### DIFF
--- a/bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/Stream2JSONInputStream.java
+++ b/bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/Stream2JSONInputStream.java
@@ -19,6 +19,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.Iterator;
 import java.util.stream.Stream;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.openhab.core.library.types.DateTimeType;
 
 import com.google.gson.Gson;
@@ -32,6 +33,7 @@ import com.google.gson.GsonBuilder;
  *
  * @author Henning Treu - Initial contribution
  */
+@NonNullByDefault
 public class Stream2JSONInputStream extends InputStream implements JSONInputStream {
 
     private final Iterator<String> iterator;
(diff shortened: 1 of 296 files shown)


bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/Stream2JSONInputStream.java
```diff
@@ -49,10 +51,6 @@
      * @throws IllegalArgumentException in case the source is null.
```
J-N-K, line 51: If it is annotated @NonNull it can't (shouldn't) throw `IllegalArgumentException` if the parameter is null.

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/extension/ScriptEngineConsoleCommandExtension.java
```diff
@@ -32,22 +35,23 @@
...
 
     public ScriptEngineConsoleCommandExtension() {
         super("script", "Execute scripts");
     }
 
     @Override
     public void execute(String[] args, Console console) {
+        ScriptEngine scriptEngine = this.scriptEngine;
         if (scriptEngine != null) {
             String scriptString = Arrays.stream(args).collect(Collectors.joining(" "));
```
J-N-K, line 52:
```suggestion
            String scriptString = String.join(" ", args);
```

J-N-K (commented):
Great! Also nice that you renamed all mocks to `*mock`

bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/Stream2JSONInputStream.java
wborn: I've fixed all the comments! :+1:

J-N-K (approved):

cweitkamp (approved):
LGTM. Thanks for this work.

Issue #2727, opened 2022-02-02
[automation] ScriptFileWatcher breaks Jython reloading on change


rkoshak:
I don't know enough to be able to say whether this belongs here or on openhba-addons. I'm putting it here for now because the timing of reports of problems almost exactly coincided with the merging of PR #2708. I'm seeing lots of reports from Jython users that changes made to their .py files are no longer being picked up. They have to restart openHAB for changes to be detected and loaded. 

In a typical Jython install using the Helper Libraries, a user will place their files into $OH_CONF/automation/jsr223/python/personal and library files would go in $OH_CONF/automation/lib/python/personal. The changes made in #2708 appears to stop watching those subfolders.

jimtng:
I'm on the 3.3.0-SNAPSHOT - Build #2737 (3rd Feb 2022). My jruby rule files in $OH_CONF/automation/jsr223/ruby/personal/ are being reloaded whenever I changed them. So this issue may be specific to jython.

wezzix:
After some 20 restarts, I can now report that **the issue is intermittent**. Sometimes it may reload a change minutes after service restart, but not 10 minutes later. Most of the time it doesn't reload at all. Just now, it reloads different files more than 10 minutes later and more than once (like expected).

J-N-K:
I doubt it is related to #2708 then.

wezzix:
Ok, I have a clue for you. Three saves in a row:

1. First save reloads ok. 
2. Second save dequeues but does not reload. 
3. Third save does not dequeue. 

My guess is that something hangs or crashes during file processing.

```
		<Logger level="TRACE" name="org.openhab.core.service.WatchQueueReader"/>
		<Logger level="TRACE" name="org.openhab.core.automation.module.script.rulesupport.loader.ScriptFileWatcher"/>
```
First save ok:
```
2022-02-03 19:14:57.628 [TRACE] [org.openhab.core.service.WatchQueueReader         ] - Modification event for /etc/openhab/automation/jsr223/python/personal/bathroom.py 
2022-02-03 19:14:57.630 [TRACE] [org.openhab.core.service.WatchQueueReader         ] - Scheduled processing of /etc/openhab/automation/jsr223/python/personal/bathroom.py
2022-02-03 19:14:57.632 [TRACE] [org.openhab.core.service.WatchQueueReader         ] - Modification event for /etc/openhab/automation/jsr223/python/personal/bathroom.py 
[...]

PR #2745, opened 2022-02-10
Fix WatchQueueReader cancelling jobs too early

J-N-K:
Related to #2727 

While investigating the issue above I found that the new version of the `WatchQueueReader` erroneously cancelled the job when removing it from the stored jobs. 

This is done on two occasions: 
- If a jobs is stored and we try to replace it. In this case cancellation is correct.
- If the job executes. Cancelling the job here is wrong. It is not noticed by the tests because it interrupts the job, which is not checked for our short-running jobs. In the issue above the job is checking if the thread is interrupted and therefore fails.

@wborn I also moved the tests to the `org.openhab.core` package and added the `JavaTest` as you suggested.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/service/WatchQueueReader.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/service/WatchQueueReader.java
@@ -268,16 +268,16 @@ public class WatchQueueReader implements Runnable {
                         }
                         if (services != null) {
                             File f = resolvedPath.toFile();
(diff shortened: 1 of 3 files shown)


wborn (commented):
It looks OK to me! :+1: But I'd also then use `.equals` instead of `==` on line 255:

https://github.com/openhab/openhab-core/blob/e584507a6532cb675b3ff11718e31cfb9ada1c67/bundles/org.openhab.core/src/main/java/org/openhab/core/service/WatchQueueReader.java#L255

J-N-K:
Done.

wborn (approved):

PR #2747, opened 2022-02-10
Avoid throwing raw exception types

wborn:
This fixes all 36 AvoidThrowingRawExceptionTypes SAT findings.

### DIFF
--- a/itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/firmware/Constants.java
+++ b/itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/firmware/Constants.java
@@ -12,6 +12,7 @@
  */
 package org.openhab.core.thing.firmware;
 
+import java.io.UncheckedIOException;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URL;
@@ -134,7 +135,7 @@ public class Constants {
         try {
             return new URL(url);
         } catch (MalformedURLException e) {
-            throw new IllegalArgumentException(e);
+            throw new UncheckedIOException(e);
         }
     }
 }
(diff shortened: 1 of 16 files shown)


itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/firmware/Constants.java
```diff
@@ -131,19 +130,11 @@
...
-            throw new RuntimeException(e);
-        }
-    }
-
-    private static URI newURI(String urlString) {
-        try {
-            return new URI(urlString);
-        } catch (URISyntaxException e) {
-            throw new RuntimeException(e);
+            throw new IllegalArgumentException(e);
```
cweitkamp, line 137: I think `UncheckedIOException` is the better choice here.

cweitkamp (approved):
LGTM. One minor remark. Wdyt?

itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/firmware/Constants.java
J-N-K: If it throws IllegalArgumentException you could also use URI.create(…) instead of catching URISyntaxException.

wborn: OK I've switched it to an `UncheckedIOException`.

> If it throws IllegalArgumentException you could also use URI.create(…) instead of catching URISyntaxException and remove the whole method.

Unfortunately it's an URL and not an URI, so it doesn't have this method. You could create it as an URI and call `toURL()` but then again it will throw a `MalformedURLException`. :slightly_frowning_face:

PR #2748, opened 2022-02-11
Support multi bundle bindings in i18n-maven-plugin

wborn:
The plugin did not generate translations for bundles of bindings that do not have a binding.xml file.
With this change it should also be possible to generate the default translations of the modbus and mqtt bindings.

Related to openhab/openhab-addons#12220

### DIFF
--- a/tools/i18n-plugin/src/main/java/org/openhab/core/tools/i18n/plugin/BundleInfoReader.java
+++ b/tools/i18n-plugin/src/main/java/org/openhab/core/tools/i18n/plugin/BundleInfoReader.java
@@ -117,6 +117,9 @@ public class BundleInfoReader {
                     } else if (type instanceof ChannelTypeXmlResult) {
                         ChannelTypeXmlResult result = (ChannelTypeXmlResult) type;
                         bundleInfo.getChannelTypesXml().add(result);
+                        if (bundleInfo.getBindingId().isBlank()) {
+                            bundleInfo.setBindingId(result.toChannelType().getUID().getBindingId());
+                        }
                     }
                 }
             } catch (ConversionException | MalformedURLException e) {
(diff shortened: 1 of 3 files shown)


tools/i18n-plugin/src/main/java/org/openhab/core/tools/i18n/plugin/BundleInfoReader.java
```diff
@@ -102,11 +103,20 @@ private void readThingInfo(Path ohinfPath, BundleInfo bundleInfo) throws IOExcep
...
-                        bundleInfo.getChannelGroupTypesXml().add((ChannelGroupTypeXmlResult) type);
+                        ChannelGroupTypeXmlResult result = (ChannelGroupTypeXmlResult) type;
+                        bundleInfo.getChannelGroupTypesXml().add(result);
+                        if (bundleInfo.getBindingId().isBlank()) {
+                            bundleInfo.setBindingId(result.getUID().getBindingId());
+                        }
                     } else if (type instanceof ChannelTypeXmlResult) {
-                        bundleInfo.getChannelTypesXml().add((ChannelTypeXmlResult) type);
+                        ChannelTypeXmlResult result = (ChannelTypeXmlResult) type;
+                        bundleInfo.getChannelTypesXml().add(result);
```
cweitkamp, line 119: What about setting the `bindingId` here in the `bundleInfo` too?

wborn: These kind of XML files don't have a `bindingId` attribute which is used in a `UID`. The parsed objects only have `URI`s which can be based on anything.

wborn: I see the `ChannelType` does extend from `AbstractDescriptionType` that has a `UID`, so I'll check if I can add a bindingId based on that. :slightly_smiling_face:

cweitkamp (approved):
LGTM. Thanks.

PR #2749, opened 2022-02-11
[rest] show start level in /systeminfo

### LINKED ISSUE #2640: [REST API] Provide start level in root endpoint
Currently the startup is hard to handle by apps that connect to the rest api.
Especially the loading of things and restoring data from persistence can lead to some issue if item/thing synchronization happened earlier.
If would be a tremendous simplification if the current start level of openhab would be served e.g. in the root endpoint.

Then one could wait until the desired value and only then sync/load the desired data.

J-N-K:
Closes #2640 


wborn (approved):
Thanks!

PR #2752, opened 2022-02-12
Add even more null annotations

wborn:
This PR adds even more missing null annotations which did not fit in #2742 as it grew too big.


cweitkamp (approved):
LGTM. Thanks.

PR #2754, opened 2022-02-12
Log a message if a model is parsed and no thing handler factory can be found

### LINKED ISSUE #2218: .things install silently fails if binding is missing
If you have a typo in a .things file (like `Bridge nqtt:broker:...`) or forgot to add the binding to addons.cfg, there is no warning, there is only the usual log line
`2021-02-28 18:17:20.945 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model 'mymqtt.things'`
but nothing more. The thing simply will not show in gui or things list.

There should be a log WARN line like `no binding found for nqtt, will ignore mymqtt.things`

J-N-K:
Fixes #2218 

A message at INFO level is logged when a thing could not be added because there is no corresponding binding (which could indicate a typo in the binding name or thing-type). This is only relevant for textual configuration as the UI prevents typos.

The message is suppressed on initial loading to prevent flooding the log (the model most likely is parsed before all ThingHandlerFactories have been added).


wborn (approved):
LGTM! @kaikreuzer can you also have a look at this because of your start level experience?

kaikreuzer (approved):
lgtm, many thanks!

PR #2756, opened 2022-02-12
Add triggeringThing name to DSL rules

jamesmelville:
I wanted to be able to detect which thing had triggered a DSL rule in order to write rules which used a single set of functionality when detecting that one of many things had gone offline. Whilst I was at it I figured it might be useful to include the previous and new statuses and remove the reference to previousState which didn't seem to make sense. I've tried to follow the pattern in the code and this PR #2567


jamesmelville:
Related request here: https://community.openhab.org/t/oh3-getting-triggeringitemname-and-previousstate-of-thing-changed/128794

kaikreuzer (approved):
Lgtm, thanks for the addition.
Could you please also create a PR to add these new implicit vars to the docs at https://www.openhab.org/docs/configuration/rules-dsl.html#implicit-variables-inside-the-execution-block?

PR #2757, opened 2022-02-12
Add bearingTo method for PointType

jamesmelville:
I wanted to be able to calculate the bearing from one PointType to another, and thought it would make sense to include in the PointType as we already had distanceFrom. I've included test coverage and tried to tidy the tests a little


wborn (approved):
Thanks! Seems like a useful addition. :slightly_smiling_face:

PR #2758, opened 2022-02-12
Add new Imperial Pressure unit psi

jamesmelville:
https://en.wikipedia.org/wiki/Pound_per_square_inch


wborn (approved):
Thanks for adding the unit!

wborn:
Can you also create a PR to add it to the list if units in the docs @jamesmelville?

See: 
* https://github.com/openhab/openhab-docs/blob/main/concepts/units-of-measurement.md#list-of-units
* https://www.openhab.org/docs/concepts/units-of-measurement.html#list-of-units

PR #2760, opened 2022-02-13
Fix context being re-used for next execution

### LINKED ISSUE #2751: event is defined during rule execution with startup trigger
I have a rule with two triggers:

```
configuration: {}
triggers:
  - id: "2"
    configuration:
      startlevel: 100
    type: core.SystemStartlevelTrigger
  - id: "1"
    configuration:
      itemName: myItem
    type: core.ItemStateChangeTrigger
conditions: []
```
If myItem is restored on startup, the rule is triggered twice on startup (which is what I would expect, once for the restored item plus once for the startup). The problem is, that in both cases "event" is defined during rule execution and pointing at one of the restored items.

If myItem is not restored on startup or if I remove the ItemStateChangeTrigger, the rule is triggered once and "event" is not defined during rule execution.

I'm using OpenHab 3.2 (openhabian)

J-N-K:
Fixes #2751 

Trigger information is inserted in the execution context. This information is changed to the new information each time the context is updated with the same keys. If the context of the next execution does not contain values for each key, the old key is re-used, leading e.g. to wrong event information in the context.

The solution is to remove the content of the triggering context after each execution.

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/AbstractScriptModuleHandler.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/AbstractScriptModuleHandler.java
@@ -101,11 +101,11 @@ public abstract class AbstractScriptModuleHandler<T extends Module> extends Base
     }
 
     /**
-     * Adds the passed context variables of the rule engine to the context scope of the ScriptEngine, this should be
-     * updated each time the module is executed
+     * Adds the passed context variables of the rule engine to the context scope of the ScriptEngine
+     * this should be done each time the module is executed to prevent leaking context to later executions
      *
      * @param engine the script engine that is used
-     * @param context the variables and types to put into the execution context
+     * @param context the variables and types to remove from the execution context
      */
     protected void setExecutionContext(ScriptEngine engine, Map<String, ?> context) {
         ScriptContext executionContext = engine.getContext();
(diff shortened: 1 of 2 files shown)


J-N-K:
This is not just DSL, it affects all JSR-223 rule engines, too.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/AbstractScriptModuleHandler.java
```diff
@@ -130,4 +130,26 @@ protected void setExecutionContext(ScriptEngine engine, Map<String, ?> context)
...
+     * Removes passed context variables of the rule engine from the context scope of the ScriptEngine, this should be
+     * updated each time the module is executed
+     *
+     * @param engine the script engine that is used
+     * @param context the variables and types to put into the execution context
+     */
+
+    protected void resetExecutionContext(ScriptEngine engine, Map<String, ?> context) {
+        ScriptContext executionContext = engine.getContext();
+
```
wborn, line 144:
```suggestion
```

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/AbstractScriptModuleHandler.java
```diff
@@ -130,4 +130,26 @@ protected void setExecutionContext(ScriptEngine engine, Map<String, ?> context)
...
     }
+
+    /**
+     * Removes passed context variables of the rule engine from the context scope of the ScriptEngine, this should be
+     * updated each time the module is executed
+     *
+     * @param engine the script engine that is used
+     * @param context the variables and types to put into the execution context
+     */
+
```
wborn, line 141:
```suggestion
```

wborn (approved):
LGTM and thanks!

PR #2761, opened 2022-02-14
[DialogProcessor] processor aborting before response

GiviMAD:

Hi, this small change fix the execution of the dialog processor.

It was aborting its own thread, because the event was produced by the stt service. So it causes an interruption execution during the tss execution after the hli.

I already tested this and change should not have other side effects as the whole method is synchronized.

Is this change ok @kaikreuzer, @lolodomo?


lolodomo (approved):
LGTM

cweitkamp (approved):
LGTM. Thanks.

PR #2762, opened 2022-02-14
Fix dispose is called on handlers for disabled things

### LINKED ISSUE #1893: Dispose method of disabled bridge thing handler is called at system shutdown
If a bridge thing is DISABLED, it appears the bridge handler's dispose method is called at shutdown even though the bridge handler's initialize method was never called. This also happens if you stop the binding from the Karaf console.

Steps to reproduce:
- create a bridge thing (I don't think it matters which binding you use as long as it has a bridge)
- disable the bridge thing using the UI
- shutdown openHAB
- start openHAB (bridge thing handler's initialize method is not called, as expected)
- shutdown openHAB (bridge thing handler's dispose method will be called) 

I also noticed that if the binding has a discovery service, it is activated at startup and deactivated at shutdown, even though the bridge thing is DISABLED.

J-N-K:
Fixes #1893 

The reason for the reported behavior is that a handler is registered on startup for disabled things. The thing is not initialized because the enabled-state is check while initializing the handler. 

Since we unregister the handler if the thing is disabled at runtime (and register a new one once the thing is enabled again), I believe we should not register a handler at this point.

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
@@ -401,9 +401,10 @@ public class ThingManagerImpl
         this.thingStatusInfoI18nLocalizationService = thingStatusInfoI18nLocalizationService;
         this.thingTypeRegistry = thingTypeRegistry;
 
+        storage = storageService.getStorage(THING_STATUS_STORAGE_NAME, this.getClass().getClassLoader());
+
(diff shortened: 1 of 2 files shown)


J-N-K:
I don‘t understand why this keeps failing in different tests. Neither Jenkins nor my own local build show this behavior.

wborn:
When it fails the itests with "stream closed" this is due to #2134. The Jenkins build uses a different configuration as workaround. It runs a seperate build for the itests. The issue is that bnd tries to use streams that can be closed by other Maven plugins in a parallel build.

wborn (approved):
Thanks for making the fix. :+1:

PR #2763, opened 2022-02-14
Add more null annotations to automation bundles

wborn:
This adds null annotations to many classes in the automation bundles and a few dimension classes.


bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/shared/simple/SimpleTriggerHandlerCallback.java
```diff
@@ -14,12 +14,14 @@
...
+import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.openhab.core.automation.handler.TriggerHandlerCallback;
 
 /**
  *
  * @author Simon Merschjohann - Initial contribution
  */
+@NonNullByDefault
 public interface SimpleTriggerHandlerCallback extends TriggerHandlerCallback {
     public void triggered(Map<String, ?> context);
```
J-N-K, line 26: I beliebe this is API breaking, right? Before a null context was allowed (even if it was not used in our own classes).

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/RuleStatusInfo.java
```diff
@@ -12,18 +12,22 @@
...
  * @author Yordan Mihaylov - Initial contribution
  * @author Kai Kreuzer - Refactored to match ThingStatusInfo implementation
  */
+@NonNullByDefault
 public class RuleStatusInfo {
 
-    private RuleStatus status;
-    private RuleStatusDetail statusDetail;
-    private String description;
+    private @NonNullByDefault({}) RuleStatus status;
```
J-N-K, line 28: Are we sure that `null` is NEVER present (even when deserializing with GSON)?

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/shared/simple/SimpleTriggerHandlerCallback.java
wborn: Yes it is API breaking. IIRC when `null` is used the `CompositeTriggerHandler` starts throwing NPEs. We can also change that class, but when null is allowed it usually adds more null checks everywhere.

J-N-K: I agree with that. You should just add the label :-)

wborn: Label added!

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/RuleStatusInfo.java
J-N-K: Fine, if it's only used for serialization. The comment in the no-arg constructor is a bit mis-leading: 

```
/**
     * Default constructor for deserialization e.g. by Gson.
     */
    protected RuleStatusInfo() {
    }
```

wborn: I just learned they are deserialzed using Gson when events are processed in `RuleEventFactory.createRuleStatusInfoEvent`. :slightly_smiling_face: I guess the same class will also be used to serialize the events.

J-N-K: Well, then my original comment is correct. We have no control over the payload that reaches that method (it is called by `createEvent` which is a public method of the component). A JSON with a `status` of `null` will not fail during deserialization but potentially lead to an NPE later.

wborn: I've added some validation just in case.

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/handler/TriggerHandlerCallback.java
```diff
@@ -49,7 +48,7 @@
      *            <li><code>value</code> - represents output value of the {@link Trigger}'s {@link Output}
      *            </ul>
      */
-    public void triggered(Trigger trigger, @Nullable Map<String, ?> context);
+    public void triggered(Trigger trigger, Map<String, ?> context);
```
kaikreuzer, line 51: @wborn I am in general a fan of the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle), hence I prefer to have such method inputs being treated conservatively - i.e. also allowing null values as it is here seems to be an optional parameter anyway.
Why do you think it should be made a non-null param?

wborn: I approached it from a clean code standpoint where allowing null values results in more complex code (potential bugs) and more unneccessary null checks, see https://github.com/openhab/openhab-core/pull/2763#discussion_r806764518.

This page summarizes what's in the Clean Code book on this subject: https://blog.markpearl.co.za/Working-With-Nulls

If providing no context is common, we could also add a `public void triggered(Trigger trigger)` method that calls `triggered(trigger, Map.of())`

kaikreuzer: > If providing no context is common, we could also add a public void triggered(Trigger trigger) method that calls triggered(trigger, Map.of())

I am not sure how common it is, but it is possible. So yes, instead of forcing the consumer to add a `Map.of()` as a second parameter, offering the method with a single parameter is probably the cleanest option.

kaikreuzer: Thanks!

kaikreuzer (approved):
lgtm, thanks!

Issue #2453, opened 2021-08-11
Add/remove user roles via karaf console


tarag:
Today (as of 3.1) the main web UI supports users belonging to several roles but the only console command to manage user role can create a user and assign it to a single role. 

There should be at least 2 additional commands to:
- add a role to an existing user
- remove a role from an existing user

In addition, a command to list all roles could be useful.

PR #2765, opened 2022-02-15
Manage user roles in the Karaf console

gennartn:
I implemented the functions to manage the role in the karaf console (I did it in the concoleCommandExecution and userRegistry class in the bundle org.openhab.core). We can now list/add/delete/change a role in the karaf console. Also, an administrator credential is requested for each command that handles the user, administrator role and API token. 

As a result, I have manage the issue https://github.com/openhab/openhab-core/issues/2453.


bundles/org.openhab.core.io.console/src/main/java/org/openhab/core/io/console/internal/extension/UserConsoleCommandExtension.java
```diff
@@ -12,9 +12,9 @@
  */
 package org.openhab.core.io.console.internal.extension;
 
-import java.util.List;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
```
wborn, line 15: Please import individual classes instead of complete packages using `*`.

> Rationale: Importing all classes from a package or static members from a class leads to tight coupling between packages or classes and might lead to problems when a new version of a library introduces name clashes.

bundles/org.openhab.core.io.console/src/main/java/org/openhab/core/io/console/internal/extension/UserConsoleCommandExtension.java
```diff
@@ -27,11 +27,14 @@
...
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Reference;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Console command extension to manage users, sessions and API tokens
  *
  * @author Yannick Schaus - Initial contribution
+ * @autor Nicolas Gennart - roles management
```
wborn, line 37:
```suggestion
 * @author Nicolas Gennart - roles management
```

bundles/org.openhab.core.ui/pom.xml
```diff
@@ -46,6 +46,11 @@
       <artifactId>org.openhab.core.bom.compile-model</artifactId>
       <type>pom</type>
     </dependency>
+    <dependency>
+      <groupId>org.openhab.core.bundles</groupId>
+      <artifactId>org.openhab.core.model.sitemap</artifactId>
+      <version>3.3.0-SNAPSHOT</version>
+    </dependency>
```
wborn, line 53: What is this import used for and is is really needed?

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/auth/UserRegistryImpl.java
```diff
@@ -15,12 +15,7 @@
 import java.security.NoSuchAlgorithmException;
 import java.security.SecureRandom;
 import java.security.spec.InvalidKeySpecException;
-import java.util.Arrays;
-import java.util.Base64;
-import java.util.HashSet;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
```
wborn, line 18: Please don't use `*` imports here either.

wborn (commented):
Thanks for creating the PR!
I haven't done a full review but I did notice a few things immediately. :wink:

gennartn:
Is my last commit good? Are there any changes to make?

MarkusThur:
Hello @gennartn I found your interesting paper "Role-based access control for openHAB users". I have no idea how to contact you other way, but I'd like to discuss this with you. especially as it is definitely necessary, that there is more control on roles.

You describe different models how to solve it, I think asigning a user to multiple roles could already help.

gennartn:
> Hello @gennartn I found your interesting paper "Role-based access control for openHAB users". I have no idea how to contact you other way, but I'd like to discuss this with you. especially as it is definitely necessary, that there is more control on roles.
> 
> You describe different models how to solve it, I think asigning a user to multiple roles could already help.
> 
> As a temporary "workauround" for my issue, I declared a user "A" which has the role user and then limit the visibility of things to role:administrator user:A if I only want this user to see it.. But it is stupid in the long run.
> 
> Cause if I now want a user "B" with the exact same rights, I would have to add B everywhere where I allowed A. If there would be a possibility to asign two roles to a user like (user, privilegedUser) this would help. Or being able to see privilegedUser as subrole of User.

gennartn:
Hello,
Thankfully for having read what I wrote. I have finished to clean my code but I did not realised yet all the test cases.. I have quite finished, I keep you posted if I succeed to finish my code for a PR, you can find my code here https://github.com/openHAB-clean-code/openhab-core/tree/main

King regards,
Nicolas

straiforos:
@gennartn any reason this PR was closed, I am very interested in continuing your work one way or another.

Issue #2764, opened 2022-02-14
Attempt to apply invalid configuration 


lolodomo:
With snapshot 2753, here is a systematic warning I got when I start my OH setup:
```

23:34:39.644 [WARN ] [b.core.thing.binding.BaseThingHandler] - Attempt to apply invalid configuration 'Configuration[{key=zwave_version; type=String; value=1.55}, {key=group_5; type=EmptyList; value=[]}, {key=group_4; type=EmptyList; value=[]}, {key=group_6; type=EmptyList; value=[]}, {key=group_1; type=ArrayList; value=[controller]}, {key=group_3; type=EmptyList; value=[]}, {key=group_2; type=EmptyList; value=[]}, {key=switchall_mode; type=BigDecimal; value=0}, {key=wakeup_node; type=BigDecimal; value=0}, {key=zwave_devicetype; type=String; value=1}, {key=config_21_1; type=BigDecimal; value=50}, {key=wakeup_interval; type=BigDecimal; value=0}, {key=zwave_deviceid; type=String; value=3}, {key=config_1_1; type=BigDecimal; value=2}, {key=zwave_nodeid; type=String; value=2}, {key=config_26_2; type=BigDecimal; value=5}, {key=config_27_1; type=BigDecimal; value=7}, {key=config_3_1; type=BigDecimal; value=3}, {key=config_22_2; type=BigDecimal; value=2000}, {key=config_4_1; type=BigDecimal; value=0}, {key=config_23_2; type=BigDecimal; value=1000}, {key=config_24_2; type=BigDecimal; value=900}, {key=node_id; type=BigDecimal; value=2}, {key=zwave_manufacturer; type=String; value=357}, {key=config_25_2; type=BigDecimal; value=20}]' on thing 'zwave:nodon_msp31x1_00_000:controller:node2' blocked. This is most likely a bug.
```
Probably something related to recent changes with configuration handling ?

lolodomo:
My previous version was snapshot 2694 and I had no warning of this kind.

I am back to 2694.

lolodomo:
It is failing in code introduced by #2712 
Does it exhibit a problem that was previously hidden ? A bug in the Z-Wave binding ?
By the way, that is a serious problem if the Z-Wave binding is now broken.

J-N-K:
The binding tries to update the thing configuration with values that are incompatible with the config description. Check if that is the case. If not, please post the Config description and I‘ll have a look.

lolodomo:
[...]

PR #2766, opened 2022-02-15
Improve log message for invalid configuration updates by handler

J-N-K:
Related to #2764 

The reason why the new configuration was rejected is now included in the log message.


wborn (approved):
That might help indeed to figure out what is wrong!

PR #2484, opened 2021-09-14
[monitor] Import javax.management package to correctly register GC metrics

kaikreuzer:
Fixes https://github.com/openhab/openhab-core/issues/2389

Signed-off-by: Kai Kreuzer <kai@openhab.org>

PR #2480, opened 2021-09-09
[metrics] Catch any Throwable during initialization

fwolter:
Workaround for #2389

Catches `Throwable`, because a `NoClassDefFoundError` is thrown.

Signed-off-by: Fabian Wolter <github@fabian-wolter.de>

PR #2767, opened 2022-02-15
Stop catching Throwables in DefaultMetricsRegistration

wborn:
Looks like this workaround introduced in #2480 is unnecessary nowadays because #2484 addressed the root cause.
This fixes the AvoidCatchingThrowable SAT findings.


cweitkamp (approved):
LGTM. Thanks.

PR #2770, opened 2022-02-15
Avoid `URISyntaxException` in tests

cweitkamp:
- Avoid `URISyntaxException` in tests by using`URI.create()` method


wborn (approved):
Thanks for the cleanup!

PR #2771, opened 2022-02-15
Prevent non-profile links for trigger channels

### LINKED ISSUE #2407: [REST] links must not be possible for trigger channels without a profile configuration
The REST API should prevent items being linked to trigger channels without a correct profile configuration.

See https://github.com/openhab/openhab-webui/issues/886#issuecomment-855368068.

J-N-K:
Fixes #2407 

I'm not fully happy with this PR. The reason is that this still allows illegal links:

1. For trigger channels if they are configured in text files.
2. For state channels in any case (e.g. it's possible to link a Switch channel to a Contact item).

The first is very difficult to fix because the item-channel-links are parsed and added from text file before things are added, so we can't use a check like it is implemented here. Delaying channel links would be a big refactoring.

The second has the same problem for textual configuration but also allows invalid links via REST API. It's more difficult to check here, because the default and follow profile allow every item type and every channel type, so if one of those is selected, the check will always return "valid", even if the item type and channel-type are not compatible. I'm unsure if it is possible to find a rule that properly covers all cases.


kaikreuzer (approved):
Lgtm - many new checks in here.
While you are right that it still does not cover all situations, I think it is a nice step forward and will help preventing certain illegal setups.

Issue #299, opened 2018-02-19
User-friendly solution for plug-in debugging


bgilmer77:
This Issue is prompted by discussion in openhab/openhab-docs#645

Frequently new users to OH do not experience the "happy path" for a number of reasons.  One area that causes issues is understanding add-ons.  The point of this Issue is to request a simple way for end-users to see potentially helpful log output for add-ons without having to have a PhD in Karaf, Javascript, OH2, or RAML, etc...).

Use case -
A new user to OH is having problems with their system.  They do not know they are missing a plugin, or perhaps they have two two networks in their house and mDNS is not working between them.  In any case, OH does not work, and they are "dead in the water" with nowhere to go.

Imagine a debug feature in an OH2 UI (Paper, perhaps), where the user can enter the name of the device and they get a simple result - such as below:

```
> smarthome:debug sonos
Debugging for sonos add-on enabled.

> smarthome:debug sonos
Debugging for sonos add-on disabled.
```

Or perhaps they see the following:
```
>smarthome:debug sonos
Debugging for sonos add-on enabled.
Analyzing logs...
Sonos plugin installed and running.
No Sonos devices found on network.
Analysis complete.
```
This Issue could be interpreted as a big and complex ask.  So a "crawl, walk, run" approach might be appropriate.  In any case, providing users with information they can understand when they are on the "unhappy path" can increase OH2 adoption and lead to a higher WAF (less time spent heads-down trying to get all the promised benefits of home automation delivered).

WDYT?

Josar:
This is a very nice suggestion.

I would allready be glad to have the log as a panel in paperUI.
And A sidebar containing a list of all installed packets to see/set log levels.

bgilmer77:
Is it as simple as just providing a log panel for starters?  Maybe yes, maybe no.  I run a lot of z-wave stuff, and I remember when I was on the "unhappy path", the log was full of all sorts of error messages.  I am not scared away by log output, but I suspect new users who are only moderately computer savvy might not be able to understand the output.  But as I said earlier, crawl, walk, run.

Josar:
@bgilmer77 lets imagine there is a side bar with buttons like set all off. And then you just check the mark at the binding you want to follow.  And the filter is applied for you and you jsut follow this dedicated Binding.

kaikreuzer:
> This Issue is prompted by discussion in openhab/openhab-docs#645

For the idea mentioned there, we actually already have https://github.com/openhab/openhab-core/issues/298.

[...]

PR #2772, opened 2022-02-15
[rest] Add ability to change loggers and expose package names of addons

J-N-K:
Related to #299

This adds the necessary parts in core for solving the mentioned issue. With these extensions it is possible 

- to add/modify/delete log settings via REST API
- get the Java package names that belong to an add-on

A future UI component can then retrieve the package names for an add-on and modify the log-levels of these packages. This could e.g. be integrated in the add-on page (next to the "remove" button). 

This is much more user-friendly than directing user to use the karaf console. On openHABian the log is displayed in frontail, so no console interaction is needed at all to get DEBUG logs.

### DIFF
--- a/bundles/org.openhab.core.karaf/pom.xml
+++ b/bundles/org.openhab.core.karaf/pom.xml
@@ -38,6 +38,12 @@
       <scope>provided</scope>
     </dependency>
     <dependency>
+      <groupId>org.apache.karaf.log</groupId>
+      <artifactId>org.apache.karaf.log.core</artifactId>
+      <version>${karaf.compile.version}</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
       <groupId>org.apache.karaf.wrapper</groupId>
       <artifactId>org.apache.karaf.wrapper.core</artifactId>
       <version>${karaf.compile.version}</version>
@@ -48,6 +54,12 @@
       <artifactId>org.apache.karaf.jaas.modules</artifactId>
       <version>${karaf.compile.version}</version>
       <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.openhab.core.bundles</groupId>
+      <artifactId>org.openhab.core.io.rest</artifactId>
+      <version>3.3.0-SNAPSHOT</version>
+      <scope>compile</scope>
     </dependency>
   </dependencies>
 
--- a/bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/KarafAddonService.java
+++ b/bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/KarafAddonService.java
@@ -19,6 +19,7 @@ import java.util.Comparator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Locale;
+import java.util.stream.Collectors;
 
 import org.apache.karaf.features.Feature;
 import org.apache.karaf.features.FeaturesService;
@@ -146,9 +147,18 @@ public class KarafAddonService implements AddonService {
                 break;
         }
 
+        // for openHAB add-on bundles the package is the same as the bundle name
+        List<String> packages = feature.getBundles().stream().filter(bundle -> !bundle.isDependency()).map(bundle -> {
+            String location = bundle.getLocation();
+            location = location.substring(0, location.lastIndexOf("/")); // strip version
+            location = location.substring(location.lastIndexOf("/") + 1); // strip groupId and protocol
+            return location;
+        }).collect(Collectors.toList());
+
         return Addon.create(type + "-" + name).withType(type).withLabel(feature.getDescription())
                 .withVersion(feature.getVersion()).withContentType(ADDONS_CONTENTTYPE).withLink(link)
-                .withAuthor(ADDONS_AUTHOR, true).withInstalled(featuresService.isInstalled(feature)).build();
+                .withLoggerPackages(packages).withAuthor(ADDONS_AUTHOR, true)
+                .withInstalled(featuresService.isInstalled(feature)).build();
     }
 
     @Override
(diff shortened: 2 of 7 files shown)


J-N-K:
Should the logging part be moved to a new bundle `org.openhab.core.io.rest.karaf` instead of adding a dependency to `org.openhab.core.io.rest` to `org.openhab.core.karaf`? If so, should it be added to a new feature or an existing one? I decided against adding it to `org.openhab.core.io.rest.log` because that bundle can be used without karaf and I need services from karaf.

bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/KarafAddonService.java
```diff
@@ -146,9 +147,18 @@ private Addon getAddon(Feature feature) {
                 break;
         }
 
+        // for openHAB add-on bundles the package is the same as the bundle name
+        List<String> packages = feature.getBundles().stream().filter(bundle -> !bundle.isDependency()).map(bundle -> {
```
kaikreuzer, line 151: Do we really need a list of packages for an add-on? We usually have a strict naming convention that all classes of an add-on should reside under the same (parent) package. The only exception is usually 3rd-party code that is packaged within or as `src/3rdparty/java`, but as this would be in the same bundle, this would not be covered by the logic here.

Thinking about how this feature can be supported in the UI, I'd think it would be best, if there's a single toggle through which the user sets the log level for the add-on. Wdyt?

kaikreuzer (changes requested):
Thanks, lgtm and it seems to be a nice feature, if the UI will eventually support it.

Indeed tricky to decide, where to put it. The Karaf bundle is probably indeed a good choice, so that no new dependencies to Karaf are introduced in other bundles.

I have just left some minor comments below.

bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/KarafAddonService.java
J-N-K: Yes, the user should only see a single switch. I'm not good at UI design, but my idea would be that in the add-ons overview (Settings/Bindings or similar) a drop-down element is displayed next to the "remove" button (and not shown when the button is labelled "installed" as it makes no sense to change loggers for an uninstalled add-on). The drop-down should contain the log-levels (with a default "DEFAULT" and TRACE/DEBUG/INFO/WARN/ERROR as other options). So the packages itself are never shown to the user.

It's true that add-ons that are provided by the `KarafAddonService` cannot make use of a list at the moment. Other implementations of `AddonService` (like the `JsonAddonService`) may have different logic to determine the packages (and also support add-ons which do not follow the strict naming policy that openHAB uses). IMO the `List` is more future-proof and does not hurt if it is only a single bundle.

kaikreuzer: My concern is that the UI would be forced to have more complex logic, as it can't simply do a single REST call when logging level is changed for an add-on, but it would have to read the package list and do multiple REST calls sequentially. That's imho not nice to move such logic to the UI, it should be handles in the core, if nothing is expected to be seen by the user anyway. It might be more elegant to have a `setLogLevel` method on the `AddonService` itself - which could then be independently implemented by the different implementations and which could at the same time be used by e.g. console commands or rule actions, which might be future additions. Wdyt?

J-N-K: That would be elegant from the API point of view (and in fact my first idea), but there are several reasons why I decided against than:

- We would need a dependency to a karaf bundle in each `AddonService` because it is needed to set the log level. Currently we have decoupled karaf from that.
- We would need to duplicate the logging code (three times).
- It would be API breaking because we would need to require each service to implement the new method.

It would be possible to workaround the first point by introducing a new OSGi service `LogLevelService` as an optional dependency. That‘s not very nice either.

I would suggest another option: we could make the REST API accept either a list of logger names and a single level (the UI can then just re-use the result from the first call) or a list of logger names/log levels (more flexible but requires some processing on client side). I would go with the first one.

J-N-K: I did check the code again, and the UI can already change all loggers in one request: The list of logger names/log-levels is already implemented. So all the UI would have to do is process

```
["some.nice.package", "another.nice.package"]
```

to

```
[
  { "loggerName": "some.nice.package", "level" : "DEBUG" },
  { "loggerName": "another.nice.package", "level" : "DEBUG" }
]
```

which I believe is a one-liner in most languages (e.g. JS: `let loggerSettings = loggerNames.map(v => ({"loggerName" : v, "level" : "DEBUG"}));`).

kaikreuzer: Ok, thanks for your explanations, let's go for it then and hope that @ghys will make use of it in the UI. 😄

J-N-K: I would try doing it myself, unfortunately I can't build the UI repo due to https://github.com/openhab/openhab-webui/issues/1165.

wborn:
I think this PR has broken the add-ons build, see:

https://github.com/openhab/openhab-addons/runs/5343766722?check_suite_focus=true

`
Message: Unable to resolve root: missing requirement [root] osgi.identity; osgi.identity=openhab-binding-souliss; type=karaf.feature; version=3.3.0.SNAPSHOT; filter:="(&(osgi.identity=openhab-binding-souliss)(type=karaf.feature)(version>=3.3.0.SNAPSHOT))" [caused by: Unable to resolve openhab-binding-souliss/3.3.0.SNAPSHOT: missing requirement [openhab-binding-souliss/3.3.0.SNAPSHOT] osgi.identity; osgi.identity=openhab-runtime-base; type=karaf.feature [caused by: Unable to resolve openhab-runtime-base/3.3.0.SNAPSHOT: missing requirement [openhab-runtime-base/3.3.0.SNAPSHOT] osgi.identity; osgi.identity=org.openhab.core.karaf; type=osgi.bundle; version="[3.3.0.202202260305,3.3.0.202202260305]"; resolution:=mandatory [caused by: Unable to resolve org.openhab.core.karaf/3.3.0.202202260305: missing requirement [org.openhab.core.karaf/3.3.0.202202260305] osgi.wiring.package; filter:="(&(osgi.wiring.package=org.apache.karaf.log.core)(version>=4.3.0)(!(version>=5.0.0)))"]]]
`

bundles/org.openhab.core.karaf/pom.xml
```diff
@@ -49,6 +55,12 @@
       <version>${karaf.compile.version}</version>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.openhab.core.bundles</groupId>
+      <artifactId>org.openhab.core.io.rest</artifactId>
+      <version>3.3.0-SNAPSHOT</version>
```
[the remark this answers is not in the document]
J-N-K: My bad. I'll provide a fix immediately.

PR #2773, opened 2022-02-16
Make ConfigDescription for thing and channels available to ThingHandler

### LINKED ISSUE #2719: Make config description available to things
After enabling config validation in #2682 it was found out that some thing handlers introduced invalid configurations which prevented thing initialization (reported as #2710). While the erroneous updates were prevented by #2712, this still leaves the handler with the task of validating configuration values. 

The `ThingHandlerCallback.validateConfigurationParameters` method can be used to find out if a given configuration is valid but does not allow checking of individual parameters and does not provide a simple way to programmatically check the types/limits of these. 

The `ThingHandlerCallback` has access to the `ConfigDescriptionRegistry` and could provide the `ConfigDescription`s of thing and channel to the handler. The question is if it would be better to allow access to the registry (`ConfigDescriptionRegistry getConfigDescriptionRegistry()`), and let the handler do all the work or if the retrieval based on the URI (`ConfigDescription getConfigDescription(String configDescriptionURI)`) or even the object (thing/channel) would be preferable.

Any thoughts?

J-N-K:
Closes #2719 

This enables a thing handler to check the configuration of the thing and it's channels prior to changing them. 

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/BaseThingHandler.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/BaseThingHandler.java
@@ -205,7 +205,7 @@ public abstract class BaseThingHandler implements ThingHandler {
     /**
      * Get the {@link ConfigDescription} of the thing
      *
-     * @return the config description (or null if not found or handler dispsed)
+     * @return the config description (or null if not found or handler disposed)
      */
     protected @Nullable ConfigDescription getConfigDescription() {
         ThingHandlerCallback callback = this.callback;
@@ -221,7 +221,7 @@ public abstract class BaseThingHandler implements ThingHandler {
     /**
      * Get the {@link ConfigDescription} for a {@link ChannelTypeUID}
      *
-     * @return the config description (or null if not found or handler dispsed)
+     * @return the config description (or null if not found or handler disposed)
      */
     protected @Nullable ConfigDescription getConfigDescription(ChannelTypeUID channelTypeUID) {
         ThingHandlerCallback callback = this.callback;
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/BaseThingHandler.java
```diff
@@ -191,14 +193,48 @@ public void channelUnlinked(ChannelUID channelUID) {
...
         } else {
             logger.warn("Handler {} tried validating its configuration although the handler was already disposed.",
                     this.getClass().getSimpleName());
         }
     }
 
+    /**
+     * Get the {@link ConfigDescription} of the thing
+     *
+     * @return the config description (or null if not found or handler dispsed)
```
cweitkamp, line 208:
```suggestion
     * @return the config description (or null if not found or handler disposed)
```

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/BaseThingHandler.java
```diff
@@ -191,14 +193,48 @@ public void channelUnlinked(ChannelUID channelUID) {
...
+            logger.warn("Handler {} tried retrieving its config description although the handler was already disposed.",
+                    this.getClass().getSimpleName());
+        }
+        return null;
+    }
+
+    /**
+     * Get the {@link ConfigDescription} for a {@link ChannelTypeUID}
+     *
+     * @return the config description (or null if not found or handler dispsed)
```
cweitkamp, line 224:
```suggestion
     * @return the config description (or null if not found or handler disposed)
```

cweitkamp (approved):
LGTM. Two minor spelling mistakes.

cweitkamp (approved):
LGTM. Two minor spelling mistakes.

PR #2775, opened 2022-02-17
Add more null annotations to XML processing classes

wborn:
This adds null annotations to many XML processing classes and a few others.


kaikreuzer (approved):
Thanks, looks good from what I can see.
One final question, though, as you have added "API breaking" to the PR: I assume this concerns only APIs that are not being used by any (official) add-ons and thus it does not really have an impact. Is this correct?

wborn:
Yes that's right. The add-ons are not impacted by this. I added the label because I removed the `ConverterAssertion` utility class. I found that only the `assertEndOfType` method in that class was used by the `ConfigDescriptionConverter` . So it seemed like a nice simplification to just move those 3 lines to where they were actually used.

kaikreuzer:
Thanks!

bundles/org.openhab.core.thing.xml/src/main/java/org/openhab/core/thing/xml/internal/ChannelTypeConverter.java
```diff
@@ -70,29 +74,30 @@ private boolean readBoolean(Map<String, String> attributes, String attributeName
     }
 
     private String readItemType(NodeIterator nodeIterator) throws ConversionException {
-        return (String) nodeIterator.nextValue("item-type", false);
+        return requireNonEmpty((String) nodeIterator.nextValue("item-type", false),
+                "ChannelType 'itemType' must not be null or empty.");
```
J-N-K, line 78: @wborn I think this change causes #2831. For trigger-channels an empty `itemType` is ok.

J-N-K, line 78: @wborn I think this change causes #2831. For trigger-channels an empty `itemType` is ok.

PR #2777, opened 2022-02-17
[config] Fixed config validation message for parameter options

cweitkamp:
- Fixed config validation message for parameter options

The message has two placeholder ans we should pass the current value and the possible options to it. 


wborn (approved):
Thanks!

PR #2778, opened 2022-02-17
Flush old JsonStorage to file before recreating a new one

### LINKED ISSUE #2774: JsonStorageService first reads database then flushes the old cache
While debugging homekit binding storage issue i found this method which does not look correct to me
JsonStorageService.getStorage

https://github.com/openhab/openhab-core/blob/44da7a4e0ee0b7b1f33c49f3f13f6cf9c2747058/bundles/org.openhab.core.storage.json/src/main/java/org/openhab/core/storage/json/internal/JsonStorageService.java#L121-L137

line 130: it creates first newStorage, i.e. reads json database into JsonStorage cache
line 134: it flushes the old instance of storage, i.e. writes old JsonStorage cache into json database (with some delay). 

this sounds to me as wrong order.  
probability that old instance has dirty cache at getStorage call is low, but still. 

it was not reason for homekit binding issue. In homekit binding I was calling getStorage twice from different classes and as result i had 2 different instances of JsonStorage which were overwriting each other.

wborn:
Fixes #2774

### DIFF
--- a/bundles/org.openhab.core.storage.json/src/main/java/org/openhab/core/storage/json/internal/JsonStorageService.java
+++ b/bundles/org.openhab.core.storage.json/src/main/java/org/openhab/core/storage/json/internal/JsonStorageService.java
@@ -116,7 +116,7 @@ public class JsonStorageService implements StorageService {
         logger.debug("Json Storage Service: Deactivated.");
     }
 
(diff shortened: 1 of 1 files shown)


J-N-K (approved):
I was about to push the same changes 😀

cweitkamp:
If so we can merge it quickly. :wink:

PR #2780, opened 2022-02-18
[DialogProcessor] change non null check for non empty check

GiviMAD:

After recent changes to null checks, seems like this need to be changed as the way of return no response from an hli has changed to return an empty string.
@lolodomo, @wborn WDYT?

### DIFF
--- a/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
+++ b/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
@@ -249,10 +249,7 @@ public class DialogProcessor implements KSListener, STTListener {
                 String question = sre.getTranscript();
                 try {
                     toggleProcessing(false);
-                    String answer = hli.interpret(locale, question);
-                    if (answer != null) {
-                        say(answer);
-                    }
+                    say(hli.interpret(locale, question));
                 } catch (InterpretationException e) {
                     String msg = e.getMessage();
                     if (msg != null) {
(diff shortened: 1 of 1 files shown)


wborn:
When it is changed like this we need to be absolutely sure that any `HumanLanguageInterpreter` implementation never returns `null` here. This can be still an issue with the [OpenNLPInterpreter](https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui.habot/src/main/java/org/openhab/ui/habot/nlp/internal/OpenNLPInterpreter.java) in the webui repo because it hasn't been fully annotated.

What might be better is to change `DialogProcessor.say(..)` by making the parameter `@Nullable` and then do a null and empty check there. If it is null is or empty we could `log.debug` this and return. That will also stop saying any empty exception messages and with the debug logging will make it clear that there was nothing to be said.

GiviMAD:
I have a PR HLI https://github.com/openhab/openhab-addons/pull/12260 powered by OpenNLP, should I change its name? because seems to be the same. Mine is primary focus on allow the user to define their own expressions so I think is still valuable. If you have any rename suggestion let me know please.

> What might be better is to change `DialogProcessor.say(..)` by making the parameter `@Nullable` and then do a null and empty check there. If it is null is or empty we could `log.debug` this and return. That will also stop saying any empty exception messages and with the debug logging will make it clear that there was nothing to be said.

Ok I will update the pr, but feel free to superseded if you are doing the same.

GiviMAD:
@wborn I have updated the PR following your comments.

GiviMAD:
The build error seems to be no related to this pr

wborn:
> I have a PR HLI https://github.com/openhab/openhab-addons/pull/12260 powered by OpenNLP, should I change its name? because seems to be the same. Mine is primary focus on allow the user to define their own expressions so I think is still valuable. If you have any rename suggestion let me know please.

Yes it might indeed become confusing when both are named OpenNLP. We could also hide that the actual implementation is done using OpenNLP. Then one day it is also more easy to swap OpenNLP for something else without end users noticing. :wink: For instance we could name them after the functionality they provide instead of what implements them. WDYT @ghys?

> I have updated the PR following your comments.

Thanks! Looks good to me now. :slightly_smiling_face: Meanwhile I also added null annotations to the OpenNLP HLI used by HABot in https://github.com/openhab/openhab-webui/pull/1318.

wborn (approved):

PR #2782, opened 2022-02-19
[rest] Allow editing transformation files

J-N-K:
This solves the necessary core-part of #2068.

The implementation makes sure that only files with registered extensions are processed and that breaking out of the configured transformation directory is not possible. The new `ManagedTransformationService` could easily be integrated into `TransformationService`, however since it provides a file extension and `TransformationService` is agnostic of the way the configuration is provided I decided against it.

@ghys Do you think this will be sufficient for the UI to support editing these files?

### DIFF
--- a/bundles/org.openhab.core.io.rest.transform/pom.xml
+++ b/bundles/org.openhab.core.io.rest.transform/pom.xml
@@ -1,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.openhab.core.bundles</groupId>
+    <artifactId>org.openhab.core.reactor.bundles</artifactId>
+    <version>3.3.0-SNAPSHOT</version>
+  </parent>
+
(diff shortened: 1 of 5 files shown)


ghys:
> @ghys Do you think this will be sufficient for the UI to support editing these files?

I suppose, but I have to raise the question: this is the first instance of a REST endpoint being able to modify the config in `$OPENHAB_CONF` (or wherever the config is).
As I understand it the common pattern is to have a Provider for files (in `conf`, maybe meant to be readonly to openHAB) and a ManagedProvider for objects backed by the JSONDB (in `userdata`) and a Registry to combine the two (and possibly others).
Having a REST endpoint to modify transformation files in the conf folder is kind of breaching this principle, isn't it?

J-N-K:
Yes, I thought about that quite a while. There are two main reasons why I decided to go this way:

- It requires next to no changes for the existing transformation add-ons, so it can easily integrated (in fact only a simple method that returns the supported file extensions needs to be implemented).
- It requires no changes on user-side, as it is fully backward compatible.

In my opinion there is no benefit in storing transformations in a database. The reason is that each transformation has its own requirements, so there is not much abstraction we can add. I originally thought about a `TransformationConfigurationProvider`, but it's quite hard to find a way to integrate that into the existing transformation scheme. 

We would need to add each transformation two times (as managed and unmanaged transformation), with different names. This is very inconvenient for the average user. Image the user has to decide when to use `MAP` (the existing service) and when to use `MMAP` (the manages service). The only way around that would be to place restrictions on the allowed file names for transformation files (e.g. exclude file names in the format `config:<transformation-name>:<uid>`. I did check that there is nearly no character that can be used as prefix except the `|` which unfortunately is already used by the javascript transformation to distinguish inline script from script files.

J-N-K:
After think one more night about it, I'll go with the Registry and twi providers. Closing here, I'll open a new PR.

PR #2784, opened 2022-02-19
Allow type migrations in JSONStorage

J-N-K:
This is a pre-requisite to solving #1301. It allows maintaining backward compatibility when types are renamed or changed.

Inspired by https://github.com/maggu2810/openhab-core/commit/b5c224e

### DIFF
--- a/bundles/org.openhab.core.storage.json/src/main/java/org/openhab/core/storage/json/internal/JsonStorage.java
+++ b/bundles/org.openhab.core.storage.json/src/main/java/org/openhab/core/storage/json/internal/JsonStorage.java
@@ -26,6 +26,7 @@ import java.util.Set;
 import java.util.Timer;
 import java.util.TimerTask;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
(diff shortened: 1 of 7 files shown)


bundles/org.openhab.core.storage.json/src/main/java/org/openhab/core/storage/json/internal/JsonStorage.java
```diff
@@ -210,20 +215,34 @@ public boolean containsKey(final String key) {
         }
 
         try {
+            String entityClassName = entry.getEntityClassName();
+            JsonElement entityValue = (JsonElement) entry.getValue();
+
+            TypeMigrator migrator = typeMigrators.get(entityClassName);
+            if (migrator != null) {
+                entityClassName = migrator.getNewType();
+                entityValue = migrator.migrate(entityValue);
```
wborn, line 224: Did you also consider migrating the whole storage when it is loaded? With this implementation it looks like it will migrate deserialized values that are never persisted forever?

It might also be easier for users to manipulate the .json file with other tools when it is fully migrated.

Doing the migration immediately may also make it easier to spot migration issues due to `TypeMigrationException`s.

J-N-K: Yes, I thought about that. But that would require iterating over the whole storage each time it is opened, which is probably not good for performance. 

Maybe it would be better to `put` the migrated value to the storage (if a migration was necessary).

J-N-K: I implemented storing the migrated value after the type migration. This keeps the benefits from gradual migration and also migrate only once. For most storages this will be a complete migration (because all storages used in managed provider will be read in total at startup).

wborn (approved):
Thank you very much! I think this should also bring us one step closer to thing type migrations. :slightly_smiling_face:

PR #2785, opened 2022-02-19
Fix deadlock when removing ScriptEngine

### LINKED ISSUE #2727: [automation] ScriptFileWatcher breaks Jython reloading on change

J-N-K:
Fixes #2727 

Awaiting feed-back from reporting user.

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
@@ -54,9 +54,9 @@ import org.slf4j.LoggerFactory;
 @NonNullByDefault
 @Component(service = ScriptEngineManager.class)
 public class ScriptEngineManagerImpl implements ScriptEngineManager {
+    private static final String THREAD_POOL_NAME = "OH-scriptenginemanager";
 
-    private final ScheduledExecutorService scheduler = ThreadPoolManager
-            .getScheduledPool(ThreadPoolManager.THREAD_POOL_NAME_COMMON);
+    private final ScheduledExecutorService scheduler = ThreadPoolManager.getScheduledPool(THREAD_POOL_NAME);
 
     private final Logger logger = LoggerFactory.getLogger(ScriptEngineManagerImpl.class);
     private final Map<String, ScriptEngineContainer> loadedScriptEngineInstances = new HashMap<>();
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
```diff
@@ -216,12 +222,14 @@ public void removeEngine(String engineIdentifier) {
             }
 
             if (scriptEngine instanceof AutoCloseable) {
-                AutoCloseable closeable = (AutoCloseable) scriptEngine;
-                try {
-                    closeable.close();
-                } catch (Exception e) {
-                    logger.error("Error while closing script engine", e);
-                }
+                scheduler.schedule(() -> {
```
J-N-K, line 225: I didn't use `.execute` here because the JavaDoc says, `.execute` may use the calling thread to execute the task. That is exactly what we want to avoid here.

wborn: Sounds like a good comment in the code. :wink:

J-N-K:
Looks like it fixed the issue.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
```diff
@@ -51,6 +54,9 @@
 @NonNullByDefault
 @Component(service = ScriptEngineManager.class)
 public class ScriptEngineManagerImpl implements ScriptEngineManager {
+    private static final String THREAD_POOL_NAME = "OH-scriptenginemanager";
+
+    private final ScheduledExecutorService scheduler = ThreadPoolManager.getScheduledPool(THREAD_POOL_NAME);
```
wborn, line 59: If it only occassionaly runs these tasks is might be better to reuse `ThreadPoolManager.THREAD_POOL_NAME_COMMON` for this instead of creating a new thread pool. 

```java
    /**
     * The common thread pool is reserved for occasional, light weight tasks that run quickly, and
     * use little resources to execute. Tasks that do not fit into this category should setup
     * their own dedicated pool or permanent thread.
     */
    public static final String THREAD_POOL_NAME_COMMON = "common";
```

Does it only schedule these tasks when rules are deleted/modified? Do you know how resource heavy the scheduled tasks will be?

J-N-K: I decided for the new thread pool to rule out that it is the same thread as the calling one. But I believe the `ScriptFileWatcher` has its own thread pool, correct?

Yes, only on delete/modify. I'm not sure how heavy the tasks are, probably not too much. And it currently only affects Jython, so I'll change it. We can change it in the future if we experience issues.

PR #2787, opened 2022-02-19
Fix rules stay uninitialized when using Java 17

wborn:
On Java 17 there is no Nashorn scripting engine so it takes a bit longer before `ScriptEngine`s are available.
Rules would stay uninitialized forever because the `ScriptModuleTypeProvider` did not notify its listeners whenever the `script.ScriptAction`, `script.ScriptCondition` module types became available.


J-N-K:
LGTM.

kaikreuzer (approved):
Thanks!

PR #2788, opened 2022-02-20
Fix test failures when building with Java 17

wborn:
* Adds some --add-opens to maven-surefire-plugin for:
  * modbus transport tests (java.net)
  * tests using Gson/XStream (java.util)
* Only run ScriptScopeOSGiTest when Nashorn is available as it has been removed since JDK 15


kaikreuzer (approved):
Thanks!

Issue #2794, opened 2022-02-21
[dialog processor] Potential bug with choice of audio format


lolodomo:
@GiviMAD : I think there is a bug in the dialog processor about the choice of the audio format. Rather Rather than choosing a matching format between the audio source and the KS service (respectively the audio source and the STT service), the code tries to find the best match between the source and the sink ! I do not understand the logic. It makes no sense IMHO.
WDYT ?

GiviMAD:
Yes I saw that when I checked the code for first time. It makes no sense for me at that moment.
I will check it again tomorrow, but I don't see any reason why the source and the sink formats should match, I think the source should be checked against the KS and STT and the sink with the TTS.

GiviMAD:
@lolodomo I have opened a PR about this. Let me if the changes makes sense to you.

lolodomo:
Should be closed by #2795

lolodomo:
Closed by #2795

PR #2795, opened 2022-02-22
[Voice] get compatible audio format refactor

GiviMAD:
This is a proposed solution for https://github.com/openhab/openhab-core/issues/2794.

Includes:

* use specific format for ks, stt and tts on the dialog processor.
* <s>refactor of static method getBestMatch in AudioFormat.</s>
* remove duplicated method getBestMatch in VoiceManagerImpl and use the one from AudioFormat.

### DIFF
--- a/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
+++ b/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
@@ -12,12 +12,9 @@
  */
 package org.openhab.core.voice.internal;
 
-import static org.openhab.core.voice.internal.VoiceManagerImpl.getBestMatch;
-
 import java.io.IOException;
 import java.util.HashSet;
 import java.util.Locale;
-import java.util.Set;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
@@ -87,7 +84,9 @@ public class DialogProcessor implements KSListener, STTListener {
     private final TranslationProvider i18nProvider;
     private final Bundle bundle;
 
-    private final @Nullable AudioFormat format;
+    private final @Nullable AudioFormat ksFormat;
+    private final @Nullable AudioFormat sttFormat;
+    private final @Nullable AudioFormat ttsFormat;
 
     /**
      * If the processor is currently processing a keyword event and thus should not spot further ones.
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
```diff
@@ -212,8 +212,7 @@ public void ksEventReceived(KSEvent ksEvent) {
             if (ksEvent instanceof KSpottedEvent) {
                 abortSTT();
                 closeStreamSTT();
-                isSTTServerAborting = false;
```
lolodomo, line 215: This line is necessary because abortSTT is setting the variation to true.

lolodomo (changes requested):

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioFormat.java
```diff
@@ -304,26 +305,28 @@ public boolean isCompatible(@Nullable AudioFormat audioFormat) {
...
-                if (output.isCompatible(preferredFormat)) {
-                    return preferredFormat;
-                } else {
-                    for (AudioFormat input : inputs) {
-                        if (output.isCompatible(input)) {
-                            return input;
-                        }
-                    }
-                }
+    public static @Nullable AudioFormat getBestMatch(Set<AudioFormat> requiredFormats,
```
lolodomo, line 316: My feeling is that you potentially changed the result of getBestMatch and I don't know if this is a good change or not.

GiviMAD: I think that's the desired behavior as the previous one did not ensure the "best" (the one selected by getPreferredFormat from the list of formats) was returned. 
I can move this to a separate PR if you think is better.

GiviMAD: Sorry, you were right.
I was not checking all available formats as the previous implementation did.
Now is fixed.
This implementation should work like the previous one but always prioritizing by getPreferredFormat.

GiviMAD: You are right, I will remove the changes to this method

lolodomo (approved):
LGTM

lolodomo (approved):
LGTM

kaikreuzer (approved):
Thanks!

PR #2796, opened 2022-02-23
Rename Thing status variables to be more explicit

jamesmelville:
Follow up to #2756

Suggestion from [rossko57](https://community.openhab.org/t/oh3-getting-triggeringitemname-and-previousstate-of-thing-changed/128794/9) that we make the variable names more explicit to avoid confusion of `previousStatus` with `previousState`.

I'll log a docs PR assuming this is accepted.


kaikreuzer (approved):
Sounds sensible, it could indeed avoid some confusion.

lolodomo:
It looks like a breaking change and we should alert users in the release notes.

J-N-K:
This feature is only available since two weeks or so (see #2756).

lolodomo:
Ok

PR #2799, opened 2022-02-25
Fix proxy can't handle requests with forwarded host list

### LINKED ISSUE #2684: ProxyFilter cannot handle comma-separated list in X-Forwarded-Host header
When using OH through **multiple apache2 reverse proxies** (like in my unusual use case), the proxies may maintain a comma-separated list of the Host headers which have been used by the client/previous proxies, which is written into / appended to the `X-Forwarded-Host` header. See https://github.com/xp-framework/xp-framework/issues/162 where the problem has been described well for another HTTP service.

Example request:
```
GET / HTTP/1.1
[...]
X-Forwarded-Host: foo.external.my.home, foo.internal.my.home
[...]
```
Here, `foo.external.my.home` is the original Host header the client has used at the external proxy, `foo.internal.my.home` the one the external proxy has used at the internal proxy. The list may go on.

If we feed such a multi-proxied request to an OH instance, this does not go well (hostnames redacted):

```
2022-01-14 16:31:54.556 [ERROR] [.io.rest.internal.filter.ProxyFilter] - Invalid X-Forwarded-Proto + X-Forwarded-Host header combination: https://foo.external.my.home, foo.internal.my.home
java.net.URISyntaxException: Illegal character in authority at index 8: https://foo.external.my.home, foo.internal.my.home
	at java.net.URI$Parser.fail(URI.java:2913) ~[?:?]
	at java.net.URI$Parser.parseAuthority(URI.java:3247) ~[?:?]
	at java.net.URI$Parser.parseHierarchical(URI.java:3158) ~[?:?]
	at java.net.URI$Parser.parse(URI.java:3114) ~[?:?]
	at java.net.URI.<init>(URI.java:600) ~[?:?]
	at org.openhab.core.io.rest.internal.filter.ProxyFilter.filter(ProxyFilter.java:90) ~[?:?]
	at org.apache.cxf.jaxrs.utils.JAXRSUtils.runContainerRequestFilters(JAXRSUtils.java:1810) ~[bundleFile:3.4.5]
	at org.apache.cxf.jaxrs.interceptor.JAXRSInInterceptor.processRequest(JAXRSInInterceptor.java:110) ~[bundleFile:3.4.5]
	at org.apache.cxf.jaxrs.interceptor.JAXRSInInterceptor.handleMessage(JAXRSInInterceptor.java:79) ~[bundleFile:3.4.5]
	at org.apache.cxf.phase.PhaseInterceptorChain.doIntercept(PhaseInterceptorChain.java:308) ~[bundleFile:3.4.5]
	at org.apache.cxf.transport.ChainInitiationObserver.onMessage(ChainInitiationObserver.java:121) ~[bundleFile:3.4.5]
	at org.apache.cxf.transport.http.AbstractHTTPDestination.invoke(AbstractHTTPDestination.java:265) ~[bundleFile:3.4.5]
	at org.apache.cxf.transport.servlet.ServletController.invokeDestination(ServletController.java:234) ~[bundleFile:3.4.5]
	at org.apache.cxf.transport.servlet.ServletController.invoke(ServletController.java:208) ~[bundleFile:3.4.5]
	at org.apache.cxf.transport.servlet.ServletController.invoke(ServletController.java:160) ~[bundleFile:3.4.5]
	at org.apache.cxf.transport.servlet.CXFNonSpringServlet.invoke(CXFNonSpringServlet.java:225) ~[bundleFile:3.4.5]
	at org.apache.cxf.transport.servlet.AbstractHTTPServlet.handleRequest(AbstractHTTPServlet.java:298) ~[bundleFile:3.4.5]
```
The culprit is [ProxyFilter.java ](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/internal/filter/ProxyFilter.java) where the plain Host header value is directly used to construct a URL without caring about the list nature appearing if we use more than one proxy.

https://github.com/openhab/openhab-core/blob/d6933021901ca28a4122f7057be3fbd4ef3d873d/bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/internal/filter/ProxyFilter.java#L56-L86

I assume ProxyFilter is used to render base URIs for generated HTML documents etc. sent to the client, right?
In this case the fix would be to take the first element of the `X-Forwarded-Host` comma-separated list and use it for the URI, either with a regex pattern or with a `String.split(..)[0]`. (I currently don't know if more checks need to be done on the header).

Tested with **openHAB 3.2.0**, but when looking into the code the problem also seems to exist on the current main branch.

(As a workaround I have configured my "inner" apache2 to replace the `X-Forwarded-Host` header appropriately.)

Sorry for not adding a PR, but I currently don't have a Java IDE + toolchain set up, so I cannot lint, build & test any changes...

J-N-K:
Fixes #2684 

This seems to be safe, because a comma is not an allowed character in a host-name.

### DIFF
--- a/bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/internal/filter/ProxyFilter.java
+++ b/bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/internal/filter/ProxyFilter.java
@@ -59,7 +59,7 @@ public class ProxyFilter implements ContainerRequestFilter {
         String host = getValue(ctx.getHeaders(), HOST_PROXY_HEADER);
         String scheme = getValue(ctx.getHeaders(), PROTO_PROXY_HEADER);
 
-        // if our request does not have neither headers end right here
+        // if our request does not have scheme or headers end right here
         if (scheme == null && host == null) {
             return;
         }
@@ -84,9 +84,15 @@ public class ProxyFilter implements ContainerRequestFilter {
             }
         }
 
+        // host may contain a list of hosts, cf. https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#x-headers
+        // we only take the first hostname
+        if (host.indexOf(",") > 0) {
+            host = host.substring(0, host.indexOf(","));
+        }
+
         // create a new URI from the current scheme + host in order to validate
         // it
-        String uriString = scheme + "://" + host;
+        String uriString = scheme + "://" + host.trim();
 
         URI newBaseUri = null;
         try {
--- a/bundles/org.openhab.core.io.rest/src/test/java/org/openhab/core/io/rest/internal/filter/ProxyFilterTest.java
+++ b/bundles/org.openhab.core.io.rest/src/test/java/org/openhab/core/io/rest/internal/filter/ProxyFilterTest.java
@@ -87,6 +87,21 @@ public class ProxyFilterTest {
     }
 
     @Test
+    public void hostListTest() throws Exception {
+        String baseURI = "http://localhost:8080/rest";
+        String requestURI = "http://localhost:8080/rest/test";
+        setupContextURIs(baseURI, requestURI);
+
+        setupContextHeaders("https", "eclipse.org, foo.bar");
+
+        filter.filter(contextMock);
+
+        URI newBaseURI = new URI("https://eclipse.org/rest");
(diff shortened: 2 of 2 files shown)


wborn (approved):
Thanks!

PR #2800, opened 2022-02-25
Log the remote address of failed login attempts

### LINKED ISSUE #2483: [auth] Failed login attempts should log out the IP the attempt came from
In the event that a user has exposed their openHAB instance to a dangerous environment (e.g. Internet) that server may be discovered and brute force login attempts may commence. There is already a dial back to increase the amount of time between failed login attempts which can be effective. However, a user might want to deploy additional mitigations such as blocking the IP at the firewall, fail2ban, etc. They also may want to correlate the login attempts across multiple services.

However, it's challenging if not impossible to do additional mitigations and correlation without having the IP address associated with the failed login attempts. We can know that we are under attack but not necessarily from where.

It's a little ambiguous to me if this is possible, but there might be a getRemoteAddr() method on the HttpServletRequest. I'm aware that may not always work and instead show what ever the closest proxy is between OH and the client but there are supposedly settings one can make in the reverse proxies to correct for that. I'm not expert here so maybe it's not possible?

J-N-K:
Closes #2483 

### DIFF
--- a/bundles/org.openhab.core.io.http.auth/src/main/java/org/openhab/core/io/http/auth/internal/AbstractAuthPageServlet.java
+++ b/bundles/org.openhab.core.io.http.auth/src/main/java/org/openhab/core/io/http/auth/internal/AbstractAuthPageServlet.java
@@ -142,12 +142,12 @@ public abstract class AbstractAuthPageServlet extends HttpServlet {
         return user;
     }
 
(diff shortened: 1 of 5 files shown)


wborn (approved):
It's nice to know where these attemps originate from. :+1:

Issue #2801, opened 2022-02-25
TTS audio broken after recent refactoring


t2000:
When I tried to use the latest OH snapshot yesterday I noticed that my picotts isn't working with my squeezebox audio anymore.

In the logs I found:

```
2022-02-25 08:11:26.526 [WARN ] [core.voice.internal.VoiceManagerImpl] - Error saying 'My Message here': No compatible audio format found for TTS 'picotts' and sink 'squeezebox:squeezeboxplayer:piCorePlayer:SQ_Stefan'
```

I looked into the `VoiceManagerImpl` and found a change that was just done one day before my snapshot:

https://github.com/openhab/openhab-core/pull/2795

Apparently the refactoring in there is broken, because:

Before:

```
public static @Nullable AudioFormat getBestMatch(Set<AudioFormat> inputs, Set<AudioFormat> outputs) {
        AudioFormat preferredFormat = getPreferredFormat(inputs);
        for (AudioFormat output : outputs) {
            if (output.isCompatible(preferredFormat)) {
                return preferredFormat;
            } else {
                for (AudioFormat input : inputs) {
                    if (output.isCompatible(input)) {
                        return input;
                    }
                }
            }
        }
        return null;
    }
```

Now (since it uses the one in `AudioFormat.java`):

```
public static @Nullable AudioFormat getBestMatch(Set<AudioFormat> inputs, Set<AudioFormat> outputs) {
        AudioFormat preferredFormat = getPreferredFormat(inputs);
        if (preferredFormat != null) {
            for (AudioFormat output : outputs) {
                if (output.isCompatible(preferredFormat)) {
                    return preferredFormat;
                } else {
                    for (AudioFormat input : inputs) {
                        if (output.isCompatible(input)) {
                            return input;
                        }
                    }
                }
            }
        }
        return null;
    }
```
    
Which means that now there is a check for `if (preferredFormat != null)` around everything.
    
But its even a bit more complicated. The `VoiceManagerImpl` has (still!) its own `public static @Nullable AudioFormat getPreferredFormat(Set<AudioFormat> audioFormats)` which is DIFFERENT from the one in `AudioFormat`.

Apparently the one in `VoiceManagerImpl` is "older" since the last changes are > 2 years ago and the one in `AudioFormat` had some changes in the last months.

The `defaultFrequency` here:

[...]

PR #2802, opened 2022-02-25
[Voice] restore getBestMatch method

GiviMAD:

Partially revert https://github.com/openhab/openhab-core/pull/2795
The getBestMatch implementation from AudioFormat behaves different. 
This restores the previous one.
Described here https://github.com/openhab/openhab-core/issues/2801


GiviMAD:
@lolodomo, I reverted the change

t2000:
Thanks for reverting this as a quick fix. But I agree that this should not be the final solution. Code duplication is always bad and as we just have seen it always hits you back :(

I will probably try the new snapshot with the reverted code soon and report back if there are still issues because auf #2650.

In #2801 I already posted what I found out about the issue so far, if I find some more time I could continue from there, but if anyone has more time, feel free to pick up from where I left of ;)

PR #2803, opened 2022-02-26
Re-introduce removed DecimalType ctors to maintain compatibility

J-N-K:
This re-introduces the constructors removed in #2596. This should restore binary compatibility with 3.2.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DecimalType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DecimalType.java
@@ -49,8 +49,8 @@ public class DecimalType extends Number implements PrimitiveType, State, Command
         this(bigDecimal(value));
     }
 
-    // TODO: Remove this constructors. They are still in place to maintain binary compatibility and will be removed once
-    // another change breaking binary compatibility is merged
+    // TODO: Remove these constructors. They are still in place to maintain binary compatibility and will be removed
+    // once another change breaking binary compatibility is merged
     public DecimalType(BigDecimal value) {
         this.value = value;
     }
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DecimalType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DecimalType.java
@@ -49,6 +49,21 @@ public class DecimalType extends Number implements PrimitiveType, State, Command
         this(bigDecimal(value));
     }
 
+    @Deprecated
+    public DecimalType(BigDecimal value) {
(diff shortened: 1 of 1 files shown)


wborn:
> This should restore binary compatibility with 3.2.

I think it will only postpone the inevitable breaking change? Won't all existing code then always be compiled against the deprecated constructors unless you explicitly cast it to a `Number`?

J-N-K:
Yes, but we would get time to enhance the external add-on services and allow compatibility checks for different versions of the binding and different versions of openHAB.

wborn:
I think it would then be better to add a comment to the deprecated methods and tell people why it is deprecated and what they have to do. :slightly_smiling_face: 

Then after a while the old methods can be removed as well as the explicit casts everybody added.

cweitkamp:
Compatibility information should be added at all. Together with an add-on version for providing the automatic update feature. Currently you are not able to see if a 3rd Party add-on is built against latest release or against snapshot and users do not know if it works on their system or not. Using snapshots usually is risky because of breaking changes like this regardless if we are talking about users or developers.

We have more than four months until next release thus should be enough time for everyone to adopt changes.

wborn:
Adding these deprecated methods back seems to add a lot of warnings to add-ons. For instance in the systeminfo binding it would create 35 deprecation warnings.

With some grepping I found ~1400 `new DecimalType(` occurences in the add-ons repo which would probably all be flagged as deprecated.

J-N-K:
Maybe we should not add the annotation then. But the shear amount of warning shows how common it is and what will not work afterwards.

J-N-K:
Before I need to update the PR many times, shall I add a comment in the code that these will be removed? And if so, as `TODO:` comment?

wborn:
I'm fine with a todo comment explaining why they are still there and that they will eventually be removed. We can also create an issue for that and put it on the project board. :slightly_smiling_face:

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DecimalType.java
```diff
@@ -49,6 +49,20 @@ public DecimalType(Number value) {
         this(bigDecimal(value));
     }
 
+    // TODO: Remove this constructors. They are still in place to maintain binary compatibility and will be removed once
```
wborn, line 52: I still have a review comment. :wink: 

```suggestion
    // TODO: Remove these constructors. They are still in place to maintain binary compatibility and will be removed once
```

cweitkamp (approved):
LGTM. 5hanks for both of you to work out a intermediate solution.

wborn (approved):

PR #2804, opened 2022-02-26
Fix missing feature for Karaf log bundle

J-N-K:


wborn (approved):
Thanks for the quick fix!

PR #2810, opened 2022-02-28
[rest] Allow invoking ThingActions via API

### LINKED ISSUE #926: [REST] Endpoint for binding actions missing
Currently it is only possible to get persistence providers through the Rest API.
Threre should be a generic way to interact with binding actions analog to persistance, too.

J-N-K:
Closes #926 

May I ask for guidance for the correct endpoint? Originally I thought that `/things/{thingUid}/action` would be a good idea but it turned out that having two classes providing endpoints under `/things` results in strange effects (like 404s even if the bundle is loaded). It is not possible to integrate this resource in the resource that provides `/things` without adding a dependency to the automation bundle.

### DIFF
--- a/bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/ThingActionsResource.java
+++ b/bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/ThingActionsResource.java
@@ -189,11 +189,11 @@ public class ThingActionsResource implements RESTResource {
     @Operation(operationId = "executeThingAction", summary = "Executes a thing action.", responses = {
             @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = Map.class))),
             @ApiResponse(responseCode = "404", description = "Action not found"),
-            @ApiResponse(responseCode = "500", description = "Creation of action handler failed") })
+            @ApiResponse(responseCode = "500", description = "Creation of action handler or execution failed") })
     public Response executeThingAction(@PathParam("thingUID") @Parameter(description = "thingUID") String thingUID,
             @PathParam("actionUid") @Parameter(description = "action type UID (including scope, separated by '.')") String actionTypeUid,
             @HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) @Parameter(description = "language") @Nullable String language,
-            @Parameter(description = "action inputs as map") Map<String, Object> actionInputs) {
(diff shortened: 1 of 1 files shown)


J-N-K:
Test fails unrelated.

kaikreuzer:
> May I ask for guidance for the correct endpoint?

I see your issue with the `/things` endpoint and don't have a solution for it.
If I see it correctly, you've now chosen a `/actions/<thingUID>/<actionUID>` path. I'd say that is a sensible choice, given the situation.

kaikreuzer:
> May I ask for guidance for the correct endpoint?

I see your issue with the `/things` endpoint and don't have a solution for it.
If I see it correctly, you've now chosen a `/actions/<thingUID>/<actionUID>` path. I'd say that is a sensible choice, given the situation.

jlaur:
@kaikreuzer - browsing through some old PR's I found this one. 🙂 With your latest comment, does it mean this PR is actually in a condition where it could be reviewed and potentially merged?

kaikreuzer:
Thanks for the ping @jlaur, seems this got lost from my radar.

kaikreuzer (approved):
Thanks for the PR and sorry for having it hanging for so long...

spacemanspiff2007:
I am currently looking at the actions of a DMX bridge.
What does the `tag` and `reference` do?
```json
[
  {
    "actionUid": "dmx.sendFade",
    "label": "immediately fade channels",
    "description": "Immediately performs fade on selected DMX channels.",
    "inputs": [
      {
        "name": "channels",
        "type": "java.lang.String",
        "label": "",
        "description": "",
        "required": false,
        "tags": [],
        "reference": "",
        "defaultValue": ""
      },
      {
        "name": "fade",
        "type": "java.lang.String",
        "label": "",
        "description": "",
        "required": false,
        "tags": [],
        "reference": "",
        "defaultValue": ""
      },
      {
        "name": "resumeAfter",
        "type": "java.lang.Boolean",
        "label": "",
        "description": "",
        "required": false,
        "tags": [],
        "reference": "",
        "defaultValue": ""
      }
    ],
    "outputs": []
  }
]
```

spacemanspiff2007:
@J-N-K : Since you create the PR:
What does the `tag` and `reference` do?

J-N-K:
@spacemanspiff2007 I have no idea. TBH I don't even know how you an assign tags to actions. The "reference" seems to be related to the (up to now) not implemented chaining of modules for passing output to other modules or getting input from other modules.

spacemanspiff2007:
Thank you for your quick response. I'll gladly ignore these fields for the HABApp users then.

PR #2811, opened 2022-02-28
[addonservices] Add version filtering

### LINKED ISSUE #2806: Add compatibility information to external add-on services
As discussed in #2803 the external add-on services are currently not able to determine if a given add-on is compatible with the core version. This is a mid-term problem since currently there is no change breaking binary compatibility since the introduction of the community marketplace/JSON 3rd party service. A solution to this issue is required prior to introducing the changes proposed in #2805.

@openhab/core-maintainers You can assign this issue to me.

J-N-K:
Fixes #2806

This enhances the add-on services with a compatible version filter. This allows displaying only compatible versions after breaking changes in openHAB core.

Discussion partly in the linked issue. Add-on services can (optionally) provide a Maven-like version range (`[3.1.0;3.3.0.M1]` or `[3.0.0;3.1.0.M4)`).

Tasks:

- [x] add parsing of versions
- [x] adjust JSON add-on service
- [x] adjust Community Marketplace add-on service 
- [x] add tests


wborn:
>  it could look like UniFi binding (beta) [3.3.0.M2;3.4).

I think that is currently the only way to add this data. It would be nice if there was a way to add metadata to Discourse topics that can also be retrieved using the Discourse REST API. But probably @ghys already looked into this before going with this approach. :wink: 

> What should happen with (potentially) incompatible add-ons. 

I wouldn't mind having an extra "Show Incompatible Add-ons" filter option in the UI for this similar to the "Show Non-Stable Bundles" filter option. I'd also replace "Bundles" with "Add-ons" in that other config option, because I think this filter is also applied to widgets/templates/block libs and KAR files.

bundles/org.openhab.core.karaf/src/main/resources/OH-INF/config/addons.xml
```diff
@@ -11,6 +11,11 @@
 			<description>Defines whether openHAB should access the remote repository for add-on installation.</description>
 			<default>true</default>
 		</parameter>
+		<parameter name="includeIncompatible" type="boolean">
```
J-N-K, line 14: Shall we move this file to the `org.openhab.core` bundle? It affects all add-on services (`remote` is used by the other add-on services, too) and is IMO placed wrong in `org.openhab.core.karaf`.

wborn: I see it is already used in the `AbstractRemoteAddonService` so that would make sense.

It might still be nice if there was also a config option to easily disable the marketplace completely. :wink:

J-N-K: I added that option to the community marketplace. The JSON 3rd party add-on service can be easily disabled by deleting the URL.

wborn: > I added that option to the community marketplace.

Yes I now see the option! :detective:

> should be shown in one place in the UI

For openHAB itself that probably makes sense. But if the other distros do not deploy some add-on service bundles, you'll probably end up with unused options in the UI. :thinking:

J-N-K: I would not use the same PID, the UI should just group these PIDs in one page.

wborn: If it can do that it would indeed be nice. IIRC the MQTT broker config that was removed in 
https://github.com/openhab/openhab-core/pull/2842 also had some kind of mechanism to group multiple configs. Though I don't know if it used different PIDs.

J-N-K: I have no idea, UI is a mystery to me. Maybe @ghys can comment on that.

Anyway, I think this is out of scope here and none of the changes makes it harder to implement that later.

bundles/org.openhab.core.addon.marketplace/src/main/resources/OH-INF/config/marketplace.xml
```diff
@@ -29,9 +35,9 @@
 				thus harm your system.</description>
 		</parameter>
 		<parameter name="showUnstable" type="boolean">
-			<label>Show Non-Stable Bundles</label>
+			<label>Show Non-Stable Add-Ons</label>
```
wborn, line 38:
```suggestion
			<label>Show Non-Stable Add-ons</label>
```

bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/BundleVersion.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    public boolean inRange(@Nullable String range) {
+        if (range == null || range.isBlank()) {
+            // if no range is given, we assume the range covers everything
+            return true;
+        }
+        Matcher matcher = RANGE_PATTERN.matcher(range);
+        if (!matcher.matches()) {
+            logger.warn("Argument {} does not define a valid version range. Assuming inRange", range);
```
wborn, line 76: If you'd throw a `IllegalArgumentException` here, you could catch it and add a bit more context to the logging so it is easier to troubleshoot which add-on has an invalid range.

wborn (commented):
Thanks for the PR.  :+1: I only have a few minor comments left!

bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityMarketplaceAddonService.java
```diff
@@ -276,7 +285,23 @@ private Addon convertTopicItemToAddon(DiscourseTopicItem topic, List<DiscourseUs
         String type = (addonType != null) ? addonType.getId() : "";
         String contentType = getContentType(topic.categoryId, tags);
 
-        String title = topic.title;
+        String title;
+        boolean compatible = true;
+
+        int compatibilityStart = topic.title.lastIndexOf("["); // version range always starts with [
```
wborn, line 291: To prevent false positives, it might also help to only check the last word of the title.

bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityMarketplaceAddonService.java
```diff
@@ -276,7 +285,23 @@ private Addon convertTopicItemToAddon(DiscourseTopicItem topic, List<DiscourseUs
...
+        if (compatibilityStart == -1) {
+            title = topic.title;
+        } else {
+            String potentialRange = topic.title.substring(compatibilityStart + 1);
+            Matcher matcher = BundleVersion.RANGE_PATTERN.matcher(potentialRange);
+            if (matcher.matches()) {
+                compatible = coreVersion.inRange(potentialRange);
+                title = topic.title.substring(0, compatibilityStart);
```
wborn, line 299: Trimming whitespace from the string might also be a good idea.

wborn (commented):
Thanks for the PR.  :+1: I only have a few minor comments left!

J-N-K:
I think I addressed all your comments.

wborn (approved):
Great! Let's give it a test. :rocket:

wborn:
For some reason if I disable "Enable Community Marketplace" it still keeps showing marketplace add-ons.

J-N-K:
I‘ll check that tomorrow. If they are installed, that is expected, as it would be impossible to uninstall them otherwise.

PR #2813, opened 2022-03-02
[Voice] fix checklocale with RuleHumanLanguageInterpreter

dalgwen:
Check for null and emptyness, as, for example, RuleHumanLanguageInterpreter returns an empty set to tell it has no locale preference.
Found by testing the new voice system (ks, stt)

### DIFF
--- a/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
+++ b/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
@@ -549,9 +549,8 @@ public class VoiceManagerImpl implements VoiceManager, ConfigOptionProvider {
         dialogProcessors.clear();
     }
 
-    private boolean checkLocales(@Nullable Set<Locale> supportedLocales, Locale locale) {
-        if (supportedLocales == null) {
-            // rule interpreter returns null so allowing it
+    private boolean checkLocales(Set<Locale> supportedLocales, Locale locale) {
+        if (supportedLocales.isEmpty()) {
             return true;
         }
         return supportedLocales.stream().anyMatch(sLocale -> {


bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -550,7 +550,7 @@ private void stopAllDialogs() {
     }
 
     private boolean checkLocales(@Nullable Set<Locale> supportedLocales, Locale locale) {
-        if (supportedLocales == null) {
+        if (supportedLocales == null || supportedLocales.isEmpty()) {
             // rule interpreter returns null so allowing it
```
lolodomo, line 554: In case the comment is wrong (you said the rule interpreyer returns an empty list and not null), it had to be fixed or removed. 
I did not verified myself.

lolodomo (changes requested):

lolodomo:
@GiviMAD : I remember a discussion about a STT service you did not know what locales to return. It could finally make sense to return an empty list. And the javadoc of the method should be updated to explain it is allowed to return an empty list of locales in case it is not possible to determine what locales are really supported by the current setup service.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
lolodomo: Verified. We have no KS, STT or HLI services return null as supported locales.

lolodomo: So please remove `@Nullable` from the parameter and remove the comment about rule interpreter.

dalgwen: Thanks & done.

lolodomo (approved):
LGTM

cweitkamp (approved):
LGTM. Thanks for the fix.

PR #2814, opened 2022-03-02
Add accepted types to ProfileContext

J-N-K:
I'm working on a very generic script profile which can be used with every item type. E.g. a script could half all commands from the handler to the item. Since a `Dimmer` channel accepts different item types, I need to know which types are accepted by the receiving item. This could e.g. also be used to link a `String` channel (containing a JSON array as result of an HTTP request) to an `Number` item (for element count in the array).

Even tough I doubt that there are other implementations of `ProfileContext` out there, I decided to keep it backward compatible and make the implementation of the newly introduced methods optional. If you don't mind the breaking change, I can easily remove them.

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ProfileContextImpl.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ProfileContextImpl.java
@@ -38,9 +38,7 @@ public class ProfileContextImpl implements ProfileContext {
     private final List<Class<? extends Command>> acceptedCommandTypes;
 
     public ProfileContextImpl(Configuration configuration) {
-        this.configuration = configuration;
-        this.acceptedDataTypes = List.of();
-        this.acceptedCommandTypes = List.of();
+        this(configuration, List.of(), List.of());
     }
 
     public ProfileContextImpl(Configuration configuration, List<Class<? extends State>> acceptedDataTypes,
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ProfileContextImpl.java
```diff
@@ -12,26 +12,42 @@
...
     private final Configuration configuration;
 
+    private final List<Class<? extends State>> acceptedDataTypes;
+    private final List<Class<? extends Command>> acceptedCommandTypes;
+
     public ProfileContextImpl(Configuration configuration) {
         this.configuration = configuration;
+        this.acceptedDataTypes = List.of();
+        this.acceptedCommandTypes = List.of();
+    }
```
cweitkamp, line 44: Should we reuse the second ctor here?

```suggestion
    public ProfileContextImpl(Configuration configuration) {
        this(configuration, List.of(), List.of());
    }
```

cweitkamp (approved):
I second your doubt on the different implementations of the `ProfileContext` interface but agree on the default methods.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ProfileContextImpl.java
J-N-K: Done.

cweitkamp (approved):

PR #2818, opened 2022-03-05
Correct GHA build step condition

wborn:
Fixes that the wrong build step was made conditional in #2817.
It should conditionally add annotations.


kaikreuzer (approved):
Thanks!

PR #2820, opened 2022-03-06
Remove temporary OSGi framework inclusion

wborn:
The itests are run using an R7 framework so the inclusion no longer seems required.
This also allows for easily customizing Import-Package headers in itest bundles.

Sometimes it useful to be able to customize this header when including libraries without OSGi bundle headers like in https://github.com/openhab/openhab-addons/pull/12431/.


kaikreuzer (approved):
lgtm

PR #2824, opened 2022-03-08
Fix PersistentInbox tries to update unmanaged things

### LINKED ISSUE #1599: [Discovery] Do not update the thing configuration of unmanaged things
When a thing is discovered (by the discovery of a binding), in case the thing is already existing, its configuration is first updated and then the managed thing is updated (which probably triggers the dispose/initialize on the thing handler).
In case the thing is unmanaged (thing declared in a config file), the thing update is failing with the following error:
```
13:44:39.298 [DEBUG] [ig.discovery.internal.PersistentInbox] - Discovery result with thing 'astro:sun:local' not added as inbox entry. It is already present as thing in the ThingRegistry.
13:44:39.309 [DEBUG] [ig.discovery.internal.PersistentInbox] - The configuration for thing 'astro:sun:local' is updated...
13:44:39.316 [WARN ] [mmon.registry.AbstractManagedProvider] - Could not update element with key astro:sun:local in ManagedThingProvider, because it does not exists.
```
I believe the warning is normal because the thing is unmanaged. The problem is that the thing configuration is updated by the call to the method `synchronizeConfiguration`. It leads to an updated thing configuration in memory (you can see it in Paper UI) but a configuration which is in fact not applied.

The problem is at this place: https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.config.discovery/src/main/java/org/openhab/core/config/discovery/internal/PersistentInbox.java#L242

I thing `synchronizeConfiguration` should first check if the thing is managed. If not, it should show a DEBUG log and return false.

J-N-K:
Fixes #1599 

### DIFF
--- a/bundles/org.openhab.core.config.discovery/src/main/java/org/openhab/core/config/discovery/internal/PersistentInbox.java
+++ b/bundles/org.openhab.core.config.discovery/src/main/java/org/openhab/core/config/discovery/internal/PersistentInbox.java
@@ -250,7 +250,8 @@ public final class PersistentInbox implements Inbox, DiscoveryListener, ThingReg
                             inboxResult.getClass().getName());
                 }
             }
-        } else {
+        } else if (managedThingProvider.get(thingUID) != null) {
+            // only try to update properties if thing is managed
             logger.debug(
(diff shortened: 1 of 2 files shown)


wborn (approved):
Thanks!

PR #2825, opened 2022-03-08
[rest] Allow binary MediaType for RawType item states

### LINKED ISSUE #1581: REST API: /items/{itemName}/state returns RawTypes as encoded plain text
Currently when using the `/items/{itemName}/state` to retrieve the state of a RawType (for instance ImageType) we get the base64-encoded version with a `Content-type: text/plain` response header like so:

```
$ http http://localhost:8080/rest/items/Spotify_AlbumImage/state
HTTP/1.1 200 OK
Content-Type: text/plain
Date: Sun, 09 Aug 2020 13:48:35 GMT
Server: Jetty(9.4.20.v20190813)
Transfer-Encoding: chunked

data:image/jpeg;base64,/9j/4AAQSkZJRgAB...
```

This is quite inefficient for images, especially when they change often; I wonder if anything prevents outputting the binary content with the MIME type set in the `Content-Type` HTTP header (if known, otherwise maybe `application/octet-stream`) i.e. something like:

```
$ http http://localhost:8080/rest/items/Spotify_AlbumImage/state
HTTP/1.1 200 OK
Content-Type: image/jpeg <<< *RawType MIME type here*
Date: Sun, 09 Aug 2020 13:48:35 GMT
Server: Jetty(9.4.20.v20190813)

<image bytes>
```

This would I think allow a web UI to display an Image item with a simple:
```html
<img src="/rest/items/Spotify_AlbumImage/state">
```
This is API-breaking, but there could be a way to make it it controllable so that there's no breakage, with a query parameter or the `Accept` header?

J-N-K:
Closes #1581 

This PR adds the ability to use different media types for RawType states. If the `Accept` header does not specify `text/plain` (which is the default used before), a new method checks if any of the provided media types in the header matches the media type of the state and sends the response with that media-type set. If no match is found, but the `Accept` header specifies `application/octet-stream` this media type is used. In case of media-type mismatch status 415 (unsupported media type) is returned. If the item state is not `RawType` the request is considered invalid and a 400 (bad request) is returned.

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
@@ -13,14 +13,17 @@
 package org.openhab.core.io.rest.core.internal.item;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Predicate;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import javax.annotation.security.RolesAllowed;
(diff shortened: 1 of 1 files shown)


wborn (approved):
LGTM! Thanks.

Issue #2517, opened 2021-10-10
Thing that is disabled and then deleted, stays indefinite on status 'removing'


arjan-io:
When a Thing is disabled with the pause button, then removed via either the button in the thing-config or via `select->remove selected`, the Thing is not removed at all but will show status `REMOVING` indefinitely. When the pause button is clicked again, the Thing goes to `UNINITALIZED` and when clicked once more to `ONLINE`. At no point is the removal process completed.

J-N-K:
I think this is essentially the same as https://github.com/eclipse-archived/smarthome/issues/942. 

What should be fixed is that enabling a thing in `REMOVING` state causes the device to initialize.

mhilbush:
> What should be fixed is that enabling a thing in REMOVING state causes the device to initialize.

This reminded me of an issue I opened a while ago. There's this behavior of a disabled thing when shutting down the system.

https://github.com/openhab/openhab-core/issues/1893

PR #2828, opened 2022-03-09
Fix things in REMOVING state initialize instead of getting removed

J-N-K:
This solves https://github.com/eclipse-archived/smarthome/issues/942

If a thing changes its status to REMOVING when it is not initialized the thing cannot be removed because `handleRemoval` cannot be called on an uninitialized handler. This is likely to happen when a binding is not available. When the binding becomes available again, the `ThingManager` erroneously changed the thing status to INITIALIZING and the thing handler get's initialized and ONLINE later. This PR calls `initialize` on the handler (so `handleRemoval` can be called later) but preserves the REMOVING status. When the `ThingHandler` tries to change the status to ONLINE, `handleRemoval` is called and the thing finally removed.

IMO #2517 should be closed if this is merged, as it is expected that either `handleRemoval` is called or the thing is force removed. 

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
@@ -105,8 +105,9 @@ import org.slf4j.LoggerFactory;
  * {@link ThingManagerImpl} tracks all things in the {@link ThingRegistry} and mediates the communication between the
  * {@link Thing} and the {@link ThingHandler} from the binding. Therefore it tracks {@link ThingHandlerFactory}s and
  * calls {@link ThingHandlerFactory#registerHandler(Thing)} for each thing, that was added to the {@link ThingRegistry}.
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
```diff
@@ -795,7 +802,14 @@ private void initializeHandler(Thing thing) {
             }
 
             if (isInitializable(thing, thingType)) {
-                setThingStatus(thing, buildStatusInfo(ThingStatus.INITIALIZING, ThingStatusDetail.NONE));
+                if (ThingStatus.REMOVING.equals(thing.getStatus())) {
+                    // preserve REMOVING state so the callback can later device to remove the thing after it has been
```
wborn, line 806:
```suggestion
                    // preserve REMOVING state so the callback can later decide to remove the thing after it has been
```

wborn (approved):
:+1:

PR #2832, opened 2022-03-11
Allow using * as wildcard in thing status triggers

### LINKED ISSUE #2167: Improve thing status handling features
In automation thing/device status/reachability is probably most important thing. Like device going offline should make all previous data received form it obsolete. Or other example - its useless to expect fire alarm from device if device is offline and we are not detecting it.
Unfortunately OH lacks possibility to easily work with thing statuses from end user perspective.

As a best option I see implementing possibility to bind thing status (https://www.openhab.org/docs/concepts/things.html) directly to OH Item (probably via some fake channel for each and every thing). That would solve probably all or at least most of thing status related problems/questions.

Thing based triggers are good step but definitely lacks flexibility and ease to use (see also #1956). Now those triggers requires hardcoding of thing ids (and hardcoding is bad). It would be nice to have some wildcard triggering possibility like "binding:\*", "binding:thingtype:\*". Also introducing some analog of "triggeringItem" here would be huge improvement.

J-N-K:
Closes #2167

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ThingStatusTriggerHandler.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ThingStatusTriggerHandler.java
@@ -17,6 +17,7 @@ import java.util.HashMap;
 import java.util.Hashtable;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Pattern;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
@@ -65,6 +66,7 @@ public class ThingStatusTriggerHandler extends BaseTriggerModuleHandler implemen
     private final BundleContext bundleContext;
 
     private final ServiceRegistration<?> eventSubscriberRegistration;
+    private final Pattern eventTopicFilter;
 
     public ThingStatusTriggerHandler(Trigger module, BundleContext bundleContext) {
         super(module);
@@ -77,6 +79,9 @@ public class ThingStatusTriggerHandler extends BaseTriggerModuleHandler implemen
             this.types = Set.of(ThingStatusInfoChangedEvent.TYPE);
         }
         this.bundleContext = bundleContext;
+
+        this.eventTopicFilter = Pattern.compile("^openhab/things/" + thingUID.replace("*", ".*?") + "/.*$");
+
         Dictionary<String, Object> properties = new Hashtable<>();
         properties.put("event.topics", "openhab/things/" + thingUID + "/*");
         eventSubscriberRegistration = this.bundleContext.registerService(EventSubscriber.class.getName(), this,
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ThingStatusTriggerHandler.java
```diff
@@ -77,6 +79,9 @@ public ThingStatusTriggerHandler(Trigger module, BundleContext bundleContext) {
             this.types = Set.of(ThingStatusInfoChangedEvent.TYPE);
         }
         this.bundleContext = bundleContext;
+
+        this.eventTopicFilter = Pattern.compile("^openhab/things/" + thingUID.replace("*", ".*?") + "/.*$");
+
         Dictionary<String, Object> properties = new Hashtable<>();
         properties.put("event.topics", "openhab/things/" + thingUID + "/*");
```
J-N-K, line 86: What is this used for? I didn't find code that uses this property. It also contradicts the description [here](https://www.openhab.org/docs/developer/utils/events.html#receive-events) which states that regular expressions should be used as filters.

bundles/org.openhab.core.automation/src/main/resources/OH-INF/automation/moduletypes/ThingTriggers.json
```diff
@@ -10,7 +10,7 @@
 					"type": "TEXT",
 					"context": "thing",
 					"label": "Thing",
-					"description": "The UID of the thing.",
+					"description": "The UID of the thing. You can use '*' as wildcard to match multiple things.",
```
cweitkamp, line 13: Not really related to your PR. Just wondering if we already have a way to translate these labels an descriptions.

J-N-K: Isn't that what `ModuleTypeI18nService` and `ModuleTypeI18nServiceImpl` is for?

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ThingStatusTriggerHandler.java
cweitkamp: I did the initial implementation based on the [GenericEventTriggerHandler](https://github.com/openhab/openhab-core/blob/9eef6a3b6d2a75514b5495e8400c2712b66bdb68/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/GenericEventTriggerHandler.java#L48) which uses nearly the same code for event subscription. Currently I cannot tell you whether there is a spacial purpose or not.

bundles/org.openhab.core.automation/src/main/resources/OH-INF/automation/moduletypes/ThingTriggers.json
cweitkamp: Yes, that is right.

All we have to do next is to define the related property files for translation in [JSON based](https://github.com/openhab/openhab-core/tree/main/bundles/org.openhab.core.automation/src/main/resources/OH-INF/automation/moduletypes) module types.

cweitkamp (approved):
LGTM. Thanks.

PR #2837, opened 2022-03-12
Run org.openhab.core tests in forks to reduce build time

wborn:
This saves about 1 minute when building this bundle.
Most bundles depend on org.openhab.core so cores would stay idle until this bundle is build in a parallel build.

### DIFF
--- a/bundles/org.openhab.core/pom.xml
+++ b/bundles/org.openhab.core/pom.xml
@@ -14,4 +14,20 @@
 
   <name>openHAB Core :: Bundles :: Core</name>
 
+  <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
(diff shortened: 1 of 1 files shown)


J-N-K:
Please wait before merging. I'm currently investigating why this seems to make my build fail again.

J-N-K:
It seems the failures (`Automation Integration Tests` never completes) are not related. I did six builds with and without this change each and the failure occurred in total in three builds, but independent of this change. So I need to look for the cause in another place. So I'm fine with merging. On succeeding builds it also saves me 1 minute.

wborn:
Great! 1 minute is a lot on a ~10 minute build. :slightly_smiling_face:

kaikreuzer (approved):
Wonderful, thanks!

PR #2838, opened 2022-03-12
Fix CertificateGenerator package

wborn:
The class is in a package that does not match the bundle name.


cweitkamp (approved):
LGTM. Thanks.

PR #2842, opened 2022-03-13
Remove MQTT system broker leftovers

### LINKED ISSUE #2841: MQTT System broker still in settings
I have upgraded to 3.3.0 M2 and System MQTT Broker is still in the settings

Screenshot 2022-03-04 at 09 14 46

wborn:
The MqttService was used to keep track of the system MQTT brokers.
Since there are none it can be removed as well as the observer, tests, configuration and translations.

Fixes #2841


kaikreuzer:
Hm, I'm just wondering: We are removing all this now, because we do not have the "embedded MQTT broker" add-on anymore available, right?
This embedded broker was just one possible "System MQTT broker", though. The MQTT service in general is about "system-wide configured Mqtt broker connections".
In general I'd claim it makes sense in many cases to configure a single MQTT broker for the whole system and make use of this in different add-ons - that was the intention of this at least.
I am not sure, though, if there are any add-ons besides the MQTT binding making use of MQTT brokers (I remember that there was this requirement, though, I just do not remember if any add-on made it into the official distro). If this is not the case, we might indeed wipe this feature.

wborn:
Yes the system wide brokers make sense. Though the only place that I found where it was used was in the MQTT binding and it got removed with https://github.com/openhab/openhab-addons/pull/12157. So either that PR should be reverted (and https://github.com/openhab/openhab-addons/pull/12472)  or the whole feature should be removed because it is now confusing to users as you can configure the system wide broker but it can no longer be used. :wink:

kaikreuzer (approved):

PR #2844, opened 2022-03-14
Fix config validation for decimal parameters with limited options

### LINKED ISSUE #2843: ConfigValidation fails on options if type is decimal
```
                <parameter name="volumeScale" type="decimal">
			<label>Volume Scale</label>
			<description>Applies a scale to the volume.</description>
			<default>1</default>
			<advanced>true</advanced>
			<options>
				<option value="1">0-100</option>
				<option value="2">0-100 in 0.5 steps</option>
				<option value="0.8">0-80</option>
				<option value="0.5">0-50</option>
			</options>
		</parameter>
```

This fails for UI configured things because the following is stored in the JSON database:

```
      "configuration": {
        "properties": {
          "ipAddress": "127.0.0.1",
          "port": 60128,
          "refreshInterval": 0,
          "volumeLimit": 100,
          "volumeScale": 2.0
        }
      },
```

For the validation `2.0` is converted to the string `"2.0"` which can't be found in the list of allowed options (`"1", "2", "0.8", "0.5"`).

J-N-K:
Fixes #2843

### DIFF
--- a/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/OptionsValidator.java
+++ b/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/OptionsValidator.java
@@ -36,7 +36,8 @@ final class OptionsValidator implements ConfigDescriptionParameterValidator {
 
         boolean invalid;
 
-        if (param.getType() == ConfigDescriptionParameter.Type.DECIMAL || param.getType() == ConfigDescriptionParameter.Type.INTEGER) {
+        if (param.getType() == ConfigDescriptionParameter.Type.DECIMAL
+                || param.getType() == ConfigDescriptionParameter.Type.INTEGER) {
             BigDecimal bdValue = new BigDecimal(value.toString());
             invalid = param.getOptions().stream().map(o -> new BigDecimal(o.getValue()))
                     .noneMatch(v -> v.compareTo(bdValue) == 0);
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/OptionsValidator.java
```diff
@@ -31,12 +34,22 @@
             return null;
         }
 
-        // Option values are a string, so we can do a simple compare
-        if (param.getOptions().stream().map(o -> o.getValue()).noneMatch(v -> v.equals(value.toString()))) {
+        boolean invalid;
+
+        if (param.getType() == ConfigDescriptionParameter.Type.DECIMAL) {
```
cweitkamp, line 39: I am afraid checking for `DECIMAL` is not enough. What if a parameter has type `INTEGER`?

J-N-K: Originally I included `integer` but after thinking about it, I thought it‘s not necessary, because they are stored as `1` and not as `1.0` but I‘ll check.

lolodomo:
Will it work with either 2 or 2.0 as option value?

J-N-K:
The fix? Yes.

cweitkamp (approved):
LGTM. Thanks.

PR #2753, opened 2022-02-12
Fix exception during startup when kar is not yet loaded

J-N-K:
Fixes #2571 

I can't test it, though (I was not able to reproduce the issue). Is there an easy way to reproduce the problem? @lolodomo I can provide instructions for a test, if you can reproduce it.

Signed-off-by: Jan N. Klug <github@klug.nrw>

PR #2845, opened 2022-03-15
Fix KAR installation check

J-N-K:
Follow-Up To #2753

It was reported on the forum that under some circumstances the service PID of the fileinstall service is not `org.apache.felix.fileinstall~deploy` but something like `org.apache.felix.fileinstall.a702e1ba-1635-4154-906d-a9646a043a56`. As a result the check if all KARs are installed failed because the addons directory could not be determined.

This PR requests all configurations for the factoryPID `org.apache.felix.fileinstall` and then takes the first to determine the addons directory. I believe this is fine because the service should not exist multiple times.


J-N-K:
@openhab/core-maintainers I believe this is critical as it seems to affect quite a lot of people.

wborn (approved):
Thanks for the fix!

PR #2852, opened 2022-03-18
Extend ProfileContext to allow data type changing profiles

J-N-K:
This is a follow-up to #2814.

It makes use of the methods added in #2814 and also adds a new method to access the command types accepted by the handler. Additionally it adds support for thing/item change detection. In case an update/creation/removal event is received, all corresponding profiles are removed from the cache. This ensures that the accepted data types and command types are always up-to-date with the current link configuration. If the item or channel cannot be found in the registry/thing, no profile is created. While this seems like a breaking change at first glance it is not, because the link would not work anyway if either item or thing(-channel) are not available.

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/profiles/ProfileContext.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/profiles/ProfileContext.java
@@ -66,4 +66,15 @@ public interface ProfileContext {
     default List<Class<? extends Command>> getAcceptedCommandTypes() {
         return List.of();
     }
+
+    /**
+     * Get the list of accepted command types for commands send to the handler
+     *
+     * This is an optional method and will return an empty list if not implemented.
+     *
+     * @return A list of all accepted command types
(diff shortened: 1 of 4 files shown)


J-N-K:
@openhab/core-maintainers Would it be possible to merge this?

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/profiles/ProfileContext.java
```diff
@@ -66,4 +66,15 @@ default List<Class<? extends State>> getAcceptedDataTypes() {
     default List<Class<? extends Command>> getAcceptedCommandTypes() {
         return List.of();
     }
+
+    /**
+     * Get the list of accepted command types for commands send to the handler
```
kaikreuzer, line 71:
```suggestion
     * Get the list of accepted command types for commands sent to the handler
```

kaikreuzer (approved):
Lgtm, thank you. Looking forward to your scripting profile. 👍

PR #2859, opened 2022-03-20
Improve HostFragmentSupportTest

wborn:
* Remove duplications
* Better null handling
* Use constants
* Use default waitForAssert timeout (10s)

This may help to get the test more stable (#1089).


J-N-K (commented):
LGTM.

One general question: Most integration tests use the hamcrest matchers while most unit tests use plain JUnit assertions. Is there any specific reason why we don't use the same style everywhere?

wborn:
I'm not aware of any particular reason and always try to use the hamcrest matchers because such tests are more readable and have better error messages. If a test does not yet use hamcrest matchers, I typically don't add them... unless the whole test is refactored to use them.

J-N-K (approved):

cweitkamp (approved):
LGTM. Thanks.

PR #2864, opened 2022-03-21
Fix enabling/disabling CommunityMarketplace

J-N-K:
For some reason if I disable "Enable Community Marketplace" it still keeps showing marketplace add-ons.

_Originally posted by @wborn in https://github.com/openhab/openhab-core/issues/2811#issuecomment-1073362324_

Simple typo when I refactored the config key to a static constant. In the `marketplace.xml` the configuration option is named `enable` not `enabled`.


wborn (approved):
Thanks for the fix!

PR #2868, opened 2022-03-23
Add missing default translations for add-ons/marketplace configuration options

wborn:
This makes it possible to use Crowdin for translating the new configuration options introduced in #2811.


J-N-K (approved):

cweitkamp (approved):
LGTM. Thanks.

PR #2876, opened 2022-03-27
Fix null annotations on PersistenceItemConfiguration

### LINKED ISSUE #2875: PersistenceItemConfiguration has wrong null annotations
Calling `toString` will result in NPEs, depending on if and which fields are `null`. IMO the `@Nullable` on `strategies` and `filters` should be removed and if `null` is provided in the constructor, an empty `List` should be added. This would also simplify code at other places. WDYT?

J-N-K:
Fixes #2875 

I didn't check if we can omit null-checks, because of the bigger refactoring in #2871 which would create merge conflicts.


wborn (approved):
Thanks!

PR #2883, opened 2022-03-30
Add a generic script transformation

J-N-K:
Depends on #2821
Related to #2872 

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptTransformationService.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptTransformationService.java
@@ -14,6 +14,8 @@ package org.openhab.core.automation.module.script;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import javax.script.Compilable;
 import javax.script.CompiledScript;
@@ -30,7 +32,10 @@ import org.openhab.core.transform.TransformationException;
 import org.openhab.core.transform.TransformationService;
 import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Deactivate;
 import org.osgi.service.component.annotations.Reference;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * The {@link ScriptTransformationService} implements a {@link TransformationService} using any available script
@@ -43,6 +48,11 @@ import org.osgi.service.component.annotations.Reference;
 public class ScriptTransformationService
         implements TransformationService, RegistryChangeListener<TransformationConfiguration> {
     public static final String OPENHAB_TRANSFORMATION_SCRIPT = "openhab-transformation-script-";
+
+    private static final Pattern SCRIPT_CONFIG_PATTERN = Pattern
+            .compile("(?<scriptType>.*?):(?<scriptUid>.*?)(\\?(?<params>.*?))?");
+
+    private final Logger logger = LoggerFactory.getLogger(ScriptTransformationService.class);
 
     private final Map<String, ScriptEngineContainer> scriptEngineContainers = new HashMap<>();
     private final Map<String, CompiledScript> compiledScripts = new HashMap<>();
@@ -57,16 +67,22 @@ public class ScriptTransformationService
             @Reference ScriptEngineManager scriptEngineManager) {
         this.transformationConfigurationRegistry = transformationConfigurationRegistry;
         this.scriptEngineManager = scriptEngineManager;
+        transformationConfigurationRegistry.addRegistryChangeListener(this);
+    }
+
+    @Deactivate
+    public void deactivate() {
+        transformationConfigurationRegistry.removeRegistryChangeListener(this);
     }
 
     @Override
     public @Nullable String transform(String function, String source) throws TransformationException {
-        int splitPoint = function.indexOf(":");
-        if (splitPoint < 1) {
+        Matcher configMatcher = SCRIPT_CONFIG_PATTERN.matcher(function);
+        if (!configMatcher.matches()) {
             throw new TransformationException("Script Type must be prepended to transformation UID.");
         }
-        String scriptType = function.substring(0, splitPoint);
-        String scriptUid = function.substring(splitPoint + 1);
+        String scriptType = configMatcher.group("scriptType");
+        String scriptUid = configMatcher.group("scriptUid");
 
         String script = scriptCache.get(scriptUid);
         if (script == null) {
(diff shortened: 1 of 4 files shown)


J-N-K:
I don‘t understand why this happens. Every second local build fails if I don‘t use -`DwithResolver=true` and the same bundles are changed back and forth with every new resolving build.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptTransformationService.java
```diff
@@ -0,0 +1,147 @@
...
+
+    private final TransformationConfigurationRegistry transformationConfigurationRegistry;
+    private final ScriptEngineManager scriptEngineManager;
+
+    @Activate
+    public ScriptTransformationService(
+            @Reference TransformationConfigurationRegistry transformationConfigurationRegistry,
+            @Reference ScriptEngineManager scriptEngineManager) {
+        this.transformationConfigurationRegistry = transformationConfigurationRegistry;
+        this.scriptEngineManager = scriptEngineManager;
```
wborn, line 59: I think `transformationConfigurationRegistry.addRegistryChangeListener(this)` is missing here. This causes the issue that scripts are not recompiled when they change.

Similarly `transformationConfigurationRegistry.removeRegistryChangeListener(this)` should be called on deactivation

wborn (changes requested):
Thanks for the PR! I finally got some time to test it and got it working with my [JSScripting (Nashorn)](https://community.openhab.org/t/jsscripting-nashorn/134770) add-on on Java 17. :+1: I did find a small issue while testing:

wborn:
One additional functionality that the existing JavaScript transformation has is that you can supply additional parameters used in scripts.

See also the [examples documentation](https://www.openhab.org/addons/transformations/javascript/#examples):

> As input name is reserved for transformed data, it can't be used in query parameters. Also ? and & characters are reserved, but if they need to passed as additional data, they can be escaped according to URI syntax.
> 
> transform/scale.js:
> 
> (function(data, cf, d) {
>     return parseFloat(data) * parseFloat(cf) / parseFloat(d);
> })(input, correctionFactor, divider)
> transform/scale.js?correctionFactor=1.1&divider=10

Maybe it is also possible to implement this so this new transformation has feature parity and there is no good reason to keep using the existing JavaScript transformation. :wink:

J-N-K:
Good idea. The nice thing is that this also works with other languages.

wborn:
Yes it works very well! So far I've tested Groovy, JS (Graal) and JS (Nashorn) transformations in parallel on Java 17. :slightly_smiling_face:

J-N-K:
There is one other feature missing, but I'm not sure if we should implement that: inline scripts.

wborn:
Maybe that is something for a follow up PR.
We could for instance use the inline script when the function has positive index for "|" which is smaller as ":"

E.g. `JS(| input / 10)`  would become `SCRIPT(js| input / 10)`

In that case the `function` value could be used as key for the `scriptCache`. But it would probably also be a good idea to then track if scripts are still in use... otherwise the cache will likely become yet another memory leak. :open_mouth:

wborn (approved):
Many thanks! :+1:

J-N-K:
Your last comment brought my attention to potential leaking memory when script engines are not properly disposed. In `ScriptEngineManagerImpl` we check if the `ScriptEngine` implements `AutoClosable` and call `close` in that case. I was just implementing that, when you pressed "merge". Do you think it's worth making a new PR for that?

wborn:
Yes it would certainly help to prevent any potential memory leak. :+1:

florian-h05:
Is this actually documented anywhere?

J-N-K:
Unfortunately there is no good place where to put that documentation. I can provide the content if someone finds a place to put it.

rkoshak:
I would expect it to find information about this in the Transformations page under Configuration, unless I'm misunderstanding what documentation is being asked about here. That's at least the first place I'd personally look for documentation about transformation capabilities provided by core.

J-N-K:
IMO it doesn't fit very well [here](https://www.openhab.org/docs/configuration/transformations.html) because this is only a general documentation that does not go into detail about specific transformations. And it can't be added to [this page](https://www.openhab.org/addons/) because that is generated automatically from the openhab-addons repository.

rkoshak:
Just because the Transformations Configuration is generic now doesn't mean we can't go as specific as necessary. I don't think this warrants inventing a whole new page to support it. I agree that we can't put it under add-ons not only because it's generated automatically but also because this isn't an add-on, it's part of core. 

I look at this similarly to Ephemeris. The docs for it didn't really "belong" anywhere either. But short of creating a new page just for it, placing the docs for it on the Actions page was the best we could manage. I'm still not happy with that but it was the best we could do. 

I think this is a similar situation. It doesn't really fit the page now but there really isn't anywhere else to put it. And it's not wholly unexpected to look there. The docs shouldn't need to go into to too many specific details on syntax. We can push that on the references for the specific languages I think. But the generic stuff like "it supports all the languages installed", the files need to go into the transformation folder, the tranformation should be wrapped in a function that takes one argument and returns the result as Strings (is this true?), stuff like that. An example or two would be nice but may not even be necessary.

florian-h05:
> I think this is a similar situation. It doesn't really fit the page now but there really isn't anywhere else to put it. And it's not wholly unexpected to look there. The docs shouldn't need to go into to too many specific details on syntax.

I totally agree on Rich’s opinion here, IMO it is not the perfect place. This feature is just to good to keep it undocumented because it’s docs do not really fit into the existing docs. So, let’s put a general, language-independent section about the generic script transformation with one/two examples there, I can then take care of adding a example to the JS Scripting add-on docs.

florian-h05:
@J-N-K Does the SCRIPT transformation work for you when you follow your docs? For me it is always failing with `Failed transforming the state 'hello' on item 'test_string' with pattern 'SCRIPT(dsl:stringlength):%s': Could not get script for UID 'stringlength'.`

J-N-K:
No. But `SCRIPT(dsl:stringlength.script):%s` does.

florian-h05:
The SCRIPT transformation is using Nashorn, right? 
How can I use JS Scripting with it?

J-N-K:
Use `application/javascript;version=ECMAScript-2021` as script type. Maybe an alias would be nice, but that needs to be implemented in the add-on ([here](https://github.com/openhab/openhab-addons/blob/7cbae110a024423f370f8f5e8524845b6f9d604a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/GraalJSScriptEngineFactory.java#L58-L72)). Something like `grails` would be a good fit IMO.

wborn:
> The SCRIPT transformation is using Nashorn, right?
> How can I use JS Scripting with it?

If you use Java 17 there will be no Nashorn. :wink:

florian-h05:
> Use application/javascript;version=ECMAScript-2021 as script type.

I could have come up with that myself, but the alias for RhinoJS confused me a bit. 
You are right that an alias would be nice, but I‘d have to check if introducing one would have any unwanted side effects (such as JS Scripting being listed two times in the script creation wizard in MainUI).

When we switch to Java 17 and Nashorn leaves us, JS Scripting can take over the `js` script type. 
Given the simplicity of transformation scripts only relying on native JS methods, they should work without any modification required.

J-N-K:
But it works fine with your Nashorn-addon (which should IMO be merged to openhab-addons after we switch to Java 17).

J-N-K:
IMO the alias will not affect that decision, but would be much easier for all those using 3.4 (which as last version of 3.x will probably be around for quite a while, if we look at the number of questions we still receive for 2.5).

florian-h05:
> IMO the alias will not affect that decision, but would be much easier for all those using 3.4

You are right, I‘ll have a look at introducing an alias, let‘s take your proposal `graaljs`.

wborn:
The extra mime type would certainly be nice for people who transition from Nashorn to GraalJS... or from GraalJS to something else whenever that goes EOL. I'm all for adding [JSScripting (Nashorn)](https://community.openhab.org/t/jsscripting-nashorn/134770) to the addons repo so it is easier to optionally install it after an upgrade. :slightly_smiling_face:

PR #2884, opened 2022-03-30
Update semantic metadata of group members when groups are added/removed

### LINKED ISSUE #2106: Mixing gui and files for semantic model does not work
Defining the semantic model partly in the gui and partly in item-files breaks after restarting openhab.

Steps to reproduce:

- Add a location group "test" in the gui.
- Add a point in an item-file:
`Number test_item "Test Item" <temperature> (test) ["Measurement", "Temperature"]`

At first the Test Item is shown as child in the semantic model but after a restart of openhab it is moved to the root level.
See also [here](https://community.openhab.org/t/mixing-gui-and-files-for-semantic-model/112952/10)

### LINKED ISSUE #2117: Items from .items files are outside the desired location in the Model tree view after reboot
*@LCN-User commented on Dec 31, 2020, 3:54 PM UTC:*

## The problem

The Item in the .items file (The group "gBadezimmerUnten" inserts the item in the desired location):
[Screenshot 2020-12-31-item](https://user-images.githubusercontent.com/76812252/103415934-a7ec9c80-4b84-11eb-81c9-e62a2deaba9c.jpg)

The Model in the Main UI looks good:
[Screenshot 2020-12-31-Model-tree-view1](https://user-images.githubusercontent.com/76812252/103415982-ceaad300-4b84-11eb-9338-6ac68a68d327.jpg)

After reboot the RaspberryPi the item is outside the house in the Model tree view:
[Screenshot 2020-12-31-Model-tree-view2](https://user-images.githubusercontent.com/76812252/103416229-0ebe8580-4b86-11eb-8ded-a98b51ceabc1.jpg)

Now I delete "gBadezimmerUnten" in the .items file and save the file. After that I (re-)add the deleted group and all is fine, untill the next reboot.

## Expected behavior

see above

## Steps to reproduce

see above

## Your environment

Raspberry Pi 4B, 4GB, Rev. 1.2
Openhabian 3.0.0 release, fresh installed on a new 64 GB SD-card
.items, .sitemaps files copied from the old openhab 2.5 installation
LCN-Binding

## Additional information

*This issue was moved by [ghys](https://github.com/ghys) from [openhab/openhab-webui#736](https://github.com/openhab/openhab-webui/issues/736).*

wborn:
This fixes the issue that the semantic metadata is incorrect when defining group member items in files and their groups in the UI.

Fixes #2106
Fixes #2117

### DIFF
--- a/bundles/org.openhab.core.semantics/src/main/java/org/openhab/core/semantics/internal/SemanticsMetadataProvider.java
+++ b/bundles/org.openhab.core.semantics/src/main/java/org/openhab/core/semantics/internal/SemanticsMetadataProvider.java
@@ -82,10 +82,10 @@ public class SemanticsMetadataProvider extends AbstractProvider<Metadata>
     @Activate
     protected void activate() {
         initRelations();
-        itemRegistry.addRegistryChangeListener(this);
         for (Item item : itemRegistry.getAll()) {
             processItem(item);
         }
+        itemRegistry.addRegistryChangeListener(this);
     }
 
     @Deactivate
--- a/bundles/org.openhab.core.semantics/src/main/java/org/openhab/core/semantics/internal/SemanticsMetadataProvider.java
+++ b/bundles/org.openhab.core.semantics/src/main/java/org/openhab/core/semantics/internal/SemanticsMetadataProvider.java
@@ -82,10 +82,10 @@ public class SemanticsMetadataProvider extends AbstractProvider<Metadata>
     @Activate
     protected void activate() {
         initRelations();
+        itemRegistry.addRegistryChangeListener(this);
         for (Item item : itemRegistry.getAll()) {
             processItem(item);
         }
-        itemRegistry.addRegistryChangeListener(this);
     }
 
     @Deactivate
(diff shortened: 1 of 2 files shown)


wborn:
Maybe you have some time to review this small bugfix so it makes it into the next milestone @openhab/core-maintainers?

bundles/org.openhab.core.semantics/src/main/java/org/openhab/core/semantics/internal/SemanticsMetadataProvider.java
```diff
@@ -82,10 +82,10 @@ public SemanticsMetadataProvider(final @Reference ItemRegistry itemRegistry) {
     @Activate
     protected void activate() {
         initRelations();
+        itemRegistry.addRegistryChangeListener(this);
```
kaikreuzer, line 85: Is moving this line up here really needed? Listening to changes should happen only after the initial item set is processed and it should be made sure that those calls only come in once the service is activated.

wborn: I moved it because without this change there is a chance that it will not receive changes when the thread gets suspended while processing all the items.

kaikreuzer: Iirc, we are using this pattern at many places.
To be really clean, we should then have a marker that the component is not yet activated and queue the listener changes to have them processed at a later moment. Because otherwise, it could also happen that you process a change before the activate method processed its older state. That would make the overall code pretty complex, though.

wborn: I've moved it back. Let's fix this when it really becomes an issue. :wink:

kaikreuzer (approved):
Many thanks!

PR #2886, opened 2022-03-31
Fix ScriptEngine not removed if created by ScriptAction

### LINKED ISSUE #2885: Script engines created in UI rules are never removed from the ScriptEngineManager
UI created rules using a script action create a script engine with a randomly created UUID as identifier. This is part of the `AbstractScriptModuleHandler`. In its `dispose()` method the script engine is removed from the `ScriptEngineManager`:

https://github.com/openhab/openhab-core/blob/738149d44b79333cf6fc4e6cd21bb4407e1f84ee/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/AbstractScriptModuleHandler.java#L82-L85

The `ScriptActionHandler` overrides `dispose()`

https://github.com/openhab/openhab-core/blob/9bf181bf0a4d6f86bfbd185f93e474dcfd17df98/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/ScriptActionHandler.java#L51-L53

and because of that the `dispose` method of the `AbstractScriptModuleHandler` is never called.The engine created for this module is therefore not removed from the `ScriptEngineManager` when the module is removed and the `ScriptExtensionProvider`s are never notified of the script's disposal (because that happens when the script engine is removed).

This could lead to memleaks, because the created script engine is still stored in the `ScriptEngineManagerImpl` internally but can no longer be used (because the UUID is random and lost). Also `ScriptExtensionProvider`s that track access by script may run into trouble.

J-N-K:
Fixes #2885 

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/ScriptActionHandler.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/ScriptActionHandler.java
@@ -41,15 +41,11 @@ public class ScriptActionHandler extends AbstractScriptModuleHandler<Action> imp
     /**
      * constructs a new ScriptActionHandler
      *
-     * @param module
-     * @param ruleUid the UID of the rule this handler is used for
+     * @param module the module
+     * @param ruleUID the UID of the rule this handler is used for
      */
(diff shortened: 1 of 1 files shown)


wborn (approved):
Thanks for the fix!

PR #2887, opened 2022-04-01
Add an access-tracking cache to be used in rules

### LINKED ISSUE #2084: Javascript based rules lack the ability for global variables

### LINKED ISSUE #2881: Move shared cache from jsscripting to core
JSScripting has a [shared cache implementation](https://github.com/openhab/openhab-addons/blob/4f4dfcca20f39c2d8ed059b67e1dba09c9c83957/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scope/SharedCache.java) that is interesting and could be useful for other scripting languages.

Currently to take advantage of this feature, one must install jsscripting addon then call `scriptExtension.importPreset('cache')` from another scripting language. By moving this to core, installing jsscripting addon is not required just to get this functionality.

The alternative is to duplicate this within other scripting addons, e.g. the one I'm interested in is jrubyscripting addon.

Thoughts, @jpg0?

### LINKED ISSUE #2943: Reloading a DSL-rule file doesn't clear old scheduled timers
When I have a DSL-rule that does something like this:

```
	mytimer = createTimer(now.plusSeconds(60)) [|
		MyItem.sendCommand(ON)
		mytimer = null
	]
```

and the rule file is modified/reloaded after the timer has been scheduled but before it was executed there is an error thrown:

```
Scheduled job failed and stopped
java.lang.reflect.UndeclaredThrowableException: null
        at com.sun.proxy.$Proxy128.apply(Unknown Source) ~[?:?]
        at org.openhab.core.model.script.actions.ScriptExecution.lambda$0(ScriptExecution.java:82) ~[?:?]
        at org.openhab.core.internal.scheduler.SchedulerImpl.lambda$12(SchedulerImpl.java:184) ~[?:?]
        at org.openhab.core.internal.scheduler.SchedulerImpl.lambda$1(SchedulerImpl.java:87) ~[?:?]
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515) [?:?]
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304) [?:?]
Caused by: org.openhab.core.model.script.engine.ScriptExecutionException: The name '<unkown>' cannot be resolved to an item or type; line 0, column 0
        at org.openhab.core.model.script.interpreter.ScriptInterpreter.invokeFeature(ScriptInterpreter.java:141) ~[?:?]
        at org.eclipse.xtext.xbase.interpreter.impl.XbaseInterpreter._doEvaluate(XbaseInterpreter.java:1008) ~[?:?]
        at org.eclipse.xtext.xbase.interpreter.impl.XbaseInterpreter._doEvaluate(XbaseInterpreter.java:971) ~[?:?]
        at org.eclipse.xtext.xbase.interpreter.impl.XbaseInterpreter.doEvaluate(XbaseInterpreter.java:247) ~[?:?]
        at org.openhab.core.model.script.interpreter.ScriptInterpreter.doEvaluate(ScriptInterpreter.java:226) ~[?:?]
        at org.eclipse.xtext.xbase.interpreter.impl.XbaseInterpreter.internalEvaluate(XbaseInterpreter.java:227) ~[?:?]
        at org.eclipse.xtext.xbase.interpreter.impl.XbaseInterpreter._doEvaluate(XbaseInterpreter.java:874) ~[?:?]
        at org.eclipse.xtext.xbase.interpreter.impl.XbaseInterpreter.doEvaluate(XbaseInterpreter.java:243) ~[?:?]
        at org.openhab.core.model.script.interpreter.ScriptInterpreter.doEvaluate(ScriptInterpreter.java:226) ~[?:?]
        at org.eclipse.xtext.xbase.interpreter.impl.XbaseInterpreter.internalEvaluate(XbaseInterpreter.java:227) ~[?:?]
        at org.eclipse.xtext.xbase.interpreter.impl.XbaseInterpreter._doEvaluate(XbaseInterpreter.java:475) ~[?:?]
        at org.eclipse.xtext.xbase.interpreter.impl.XbaseInterpreter.doEvaluate(XbaseInterpreter.java:251) ~[?:?]
        at org.openhab.core.model.script.interpreter.ScriptInterpreter.doEvaluate(ScriptInterpreter.java:226) ~[?:?]
        at org.eclipse.xtext.xbase.interpreter.impl.XbaseInterpreter.internalEvaluate(XbaseInterpreter.java:227) ~[?:?]
        at org.eclipse.xtext.xbase.interpreter.impl.XbaseInterpreter.evaluate(XbaseInterpreter.java:213) ~[?:?]
        at org.eclipse.xtext.xbase.interpreter.impl.ClosureInvocationHandler.doInvoke(ClosureInvocationHandler.java:47) ~[?:?]
        at org.eclipse.xtext.xbase.interpreter.impl.AbstractClosureInvocationHandler.invoke(AbstractClosureInvocationHandler.java:30) ~[?:?]
        ... 10 more
```

I would expect that upon unloading the outstanding timers are cleared so that there is no attempt to execute them when they are gone.

J-N-K:
Closes #2084
Closes #2881 
Closes #2943 
Depends on #2886

This implements option 1 that was discussed in #2084 (which seems to be what most comments prefer). It is remotely based on the code here: https://github.com/openhab/openhab-addons/blob/4f4dfcca20f39c2d8ed059b67e1dba09c9c83957/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scope/SharedCache.java.

Two separate key-value-caches are implemented:

1. `privateCache`: It is created per `scriptIdentifier` and can only be accessed from the same script. It is not cleared between two subsequent runs of a rule but removed when the script is unloaded. This can be used e.g. to store timers. 

2. `sharedCache`: It is shared between all scripts. Every access to a certain key in the cache is tracked. The key is removed from the map if all scripts that ever accessed (`put` or `get`) that key have been unloaded.

Since script unloading is tracked by the framework nothing special needs to be done in the script itself. That makes it usable also in UI rules, where `scriptUnloaded` is not available.

~~I am currently investigating if I can auto-cancel `ScheduledFutures` and `Timer` when a key is removed (i.e. the last script that used the key was unloaded).~~ `ScheduledFuture<?>` and `Timer` objects are cancelled when the script is unloaded in a private cache or if the last accessor was removed in the shared cache. They are not cancelled on `remove` because I might be that the script wants to use it or take special actions before cancellation.

@jpg0: do you think the preset name `cache` should be changed, so that we do not run into conflicts? I changed the case of the `sharedcache` to be more consistent with the camel-case we use in general. As an alternative, I could implement `sharedcache` as an alias to `sharedCache` and log a warning if that is used instead of the new name. The old name could then be removed in a later version.

### DIFF
--- a/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/shared/ValueCache.java
+++ b/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/shared/ValueCache.java
@@ -21,7 +21,6 @@ import org.eclipse.jdt.annotation.Nullable;
  * The {@link ValueCache} can be used by scripts to share information between subsequent runs of the same script or
  * between scripts (depending on implementation).
  *
- *
  * @author Jan N. Klug - Initial contribution
  */
 @NonNullByDefault
--- a/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/shared/ValueCache.java
+++ b/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/shared/ValueCache.java
@@ -1,0 +1,66 @@
+/**
+ * Copyright (c) 2010-2022 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.automation.module.script.rulesupport.shared;
+
+import java.util.function.Supplier;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+
+/**
+ * The {@link ValueCache} can be used by scripts to share information between subsequent runs of the same script or
+ * between scripts (depending on implementation).
+ *
+ *
+ * @author Jan N. Klug - Initial contribution
+ */
+@NonNullByDefault
+public interface ValueCache {
+
+    /**
+     * Add a new key-value-pair to the cache. If the key is already present, the old value is replaces by the new value.
+     *
+     * @param key a string used as key
+     * @param value an {@code Object} to store with the key
+     * @return the old value associated with this key or {@code null} if key didn't exist
+     */
+    @Nullable
+    Object put(String key, Object value);
+
+    /**
+     * Remove a key (and its associated value) from the cache
+     *
+     * @param key the key to remove
+     * @return the previously associated value to this key or {@code null} if key not present
+     */
+    @Nullable
+    Object remove(String key);
+
+    /**
+     * Get a value from the cache
+     *
+     * @param key the key of the requested value
+     * @return the value associated with the key or {@code null} if key not present
+     */
+    @Nullable
+    Object get(String key);
+
+    /**
+     * Get a value from the cache or create a new key-value-pair from the given supplier
+     *
+     * @param key the key of the requested value
+     * @param supplier a supplier that returns a non-null value to be used if the key was not present
+     * @return the value associated with the key
+     */
+    Object get(String key, Supplier<Object> supplier);
+}
--- a/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/CacheScriptExtension.java
+++ b/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/CacheScriptExtension.java
@@ -1,0 +1,206 @@
+/**
+ * Copyright (c) 2010-2022 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.automation.module.script.rulesupport.internal;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Supplier;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+import org.openhab.core.automation.module.script.ScriptExtensionProvider;
+import org.openhab.core.automation.module.script.rulesupport.shared.ValueCache;
+import org.osgi.service.component.annotations.Component;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The {@link CacheScriptExtension} extends scripts to use a cache shared between rules
+ *
+ * @author Jan N. Klug - Initial contribution
+ */
+@Component(immediate = true)
+@NonNullByDefault
+public class CacheScriptExtension implements ScriptExtensionProvider {
+    static final String PRESET_NAME = "cache";
+    static final String SHARED_CACHE_NAME = "sharedCache";
+    static final String PRIVATE_CACHE_NAME = "privateCache";
+
+    private final Logger logger = LoggerFactory.getLogger(CacheScriptExtension.class);
+
+    private final Lock cacheLock = new ReentrantLock();
+    private final Map<String, Object> sharedCache = new HashMap<>();
+    private final Map<String, Set<String>> sharedCacheKeyAccessors = new HashMap<>();
+
+    private final Map<String, ValueCacheImpl> privateCaches = new HashMap<>();
+
+    public CacheScriptExtension() {
+    }
+
+    @Override
+    public Collection<String> getDefaultPresets() {
+        return Set.of();
+    }
+
+    @Override
+    public Collection<String> getPresets() {
+        return Set.of(PRESET_NAME);
+    }
+
+    @Override
+    public Collection<String> getTypes() {
+        return Set.of(PRIVATE_CACHE_NAME, SHARED_CACHE_NAME);
+    }
+
+    @Override
+    public @Nullable Object get(String scriptIdentifier, String type) throws IllegalArgumentException {
+        if (SHARED_CACHE_NAME.equals(type)) {
+            return new TrackingValueCacheImpl(scriptIdentifier);
+        } else if (PRIVATE_CACHE_NAME.equals(type)) {
+            return privateCaches.computeIfAbsent(scriptIdentifier, ValueCacheImpl::new);
+        }
+
+        return null;
+    }
+
+    @Override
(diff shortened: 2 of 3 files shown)


jpg0:
@J-N-K tbh I have not been very careful with naming in the past because:

- I personally always explicitly import everything as I've had huge issues in the past with these implicit things clashing
- There has not been much care taken here so far in naming - for example `event` is surely one of the most used variable names possible!
- Any add-on can just insert it's own new preset with whatever name it chooses at whatever time it chooses and cause clashes

As you are probably aware I have been against auto-importing these symbols due to these issues - however as both (auto-import and common names) things have already happened, my view now is that to remediate the problem we now should try to be clear to the user when a clash actually occurs (most likely by hooking into the JS engine) if it becomes a problem. (This hasn't been done.)

As for the immediate concern, I appreciate the concern of clashing with existing scripts but also fear needing to come up with more and more obscure preset names as we introduce additional ones. Not sure where we should draw the line though.

J-N-K:
@openhab/core-maintainers As this is a much requested feature it would be great if someone could find the time to review. About half of the changed lines are tests.

jimtng:
Could this make it into 3.4?

florian-h05:
This is also on our "wishlist" at JS Scripting, as well as it is closing 3 issues at once.

@openhab/core-maintainers Would be great if it could make it into 3.4.0.

rkoshak:
I'll chime in. From the end user's perspective, this will be very important to universally support many use cases that are only possible through sharing variables between script actions and conditions or preserving variables between runs of a single script action/condition.

I've been anticipating this PR's merge for some time. It would be awesome if we could get it into 3.4 release.

jimtng:
Who do we need to ping? Perhaps this is overlooked?

florian-h05:
I already pinged the core maintainers, if we want to, we can also explicitely ping kaikreuzer, cweitkamp or wborn.

J-N-K:
It seems that there is an issue with using this new feature in JSScripting as well as DSL rules. @florian-h05, do you know how to import additional presets? According to the [JSR223 documentation](https://www.openhab.org/docs/configuration/jsr223.html) this should work via `scriptExtension.importPreset` but `scriptExtension` (or its alias `se`) seems to be unavailable in JSScripting. In Groovy the following works:

```
import org.openhab.core.automation.*
import org.openhab.core.automation.module.script.rulesupport.shared.simple.*
import org.openhab.core.config.core.Configuration

scriptExtension.importPreset("RuleSupport")
scriptExtension.importPreset("RuleSimple")
scriptExtension.importPreset("cache")

def sRule = new SimpleRule() {
    Object execute(Action module, Map<String, ?> inputs) {
	    sharedCache.put("x", "y")
    }
}
sRule.setTriggers([
    TriggerBuilder.create()
        .withId("aTimerTrigger")
        .withTypeUID("timer.GenericCronTrigger")
        .withConfiguration(new Configuration([cronExpression: "0 * * * * ?"]))
        .build()
    ])

automationManager.addRule(sRule)
```

jpg0:
@J-N-K you can reference it like this:
```
const { cache } = require('@runtime/Defaults');
```
(assuming that it's in the default set of presets)

florian-h05:
@J-N-K 
AFAIK, you are right that `scriptExtension` is not available, but reading throught https://github.com/openhab/openhab-addons/blob/622654ff1d020454c60980ccfc500905f7f2c068/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/ScriptExtensionModuleProvider.java, I think it should be possible to import the cache using `require('@runtime/cache')` (as it is not in the default preset, but in the `cache` preset).

J-N-K:
Thanks, I can confirm that

```
var { sharedCache } = require("@runtime/cache")

console.log(sharedCache.get("x"))
```

successfully logs `y` which is the value set in the Groovy script. So it works across languages, just DSL needs to be figured out. I'm currently not sure how to implement the `privateCache`, the `sharedCache` seems to be possible.

J-N-K:
I also found a way to make the cache available in DSL. It feels a bit like an ugly hack, so I would prefer to do that in another PR, as it‘ll most likely be a longer review process.

florian-h05:
So this is finished and ready for review now?

J-N-K:
Yes, it‘s also rebased and confirmed working with latest changes to automation.

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/shared/ValueCache.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.function.Supplier;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+
+/**
+ * The {@link ValueCache} can be used by scripts to share information between subsequent runs of the same script or
+ * between scripts (depending on implementation).
+ *
+ *
```
wborn, line 24:
```suggestion
 *
```

wborn (approved):
Very nice! :+1: Let's do some testing with a bigger audience. :rocket:
Can you also create a PR to add some docs for this?

wborn:
Is the plan now also to remove the existing [SharedCache](https://github.com/openhab/openhab-addons/blob/e979be49114a0e9f234f19ce57b9e11cc2b4493a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scope/SharedCache.java) from the jsscripting add-on to prevent "user cache confusion"?

florian-h05:
> Is the plan now also to remove the existing [SharedCache](https://github.com/openhab/openhab-addons/blob/e979be49114a0e9f234f19ce57b9e11cc2b4493a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scope/SharedCache.java) from the jsscripting add-on to prevent "user cache confusion"?

Yes, but the coexistence should not lead to real user confusion, as the SharedCache is not documented directly and only exposed through the helper library.
I just need to figure out how to handle both the shared and the private cache in the JavaScript library whilst not breaking existing installations.
I‘ll think about this and probably ping a few people to get their opinion.

florian-h05:
@J-N-K Does the cancellation of timers work for you in JS Scripting?

I am using the following script, and I can edit it and save it, openHAB reloads the script, but the timer runs (you need to have a look at the log timestamps, there should run no timer 15 seconds after the first reload of the script):

```javascript
const { privateCache } = require('@runtime/cache');
const { actions, time } = require('openhab');

console.log(privateCache.get('timer'));

const timer = actions.ScriptExecution.createTimer(time.toZDT().plusSeconds(15), () => {
  console.warn('Timer ran.');
})

privateCache.put('timer', timer);
```

Log: 
```
19:35:58.228 [INFO ] [port.loader.AbstractScriptFileWatcher] - Loading script '/etc/openhab/automation/js/test.js'
19:35:59.221 [INFO ] [penhab.automation.script.file.test.js] - null
19:36:06.185 [INFO ] [port.loader.AbstractScriptFileWatcher] - Loading script '/etc/openhab/automation/js/test.js'
19:36:07.172 [INFO ] [penhab.automation.script.file.test.js] - null
19:36:14.223 [WARN ] [penhab.automation.script.file.test.js] - Timer ran.
19:36:22.172 [WARN ] [penhab.automation.script.file.test.js] - Timer ran.
```

J-N-K:
Can't work, because the import is wrong.

PR #2893, opened 2022-04-03
Reduce Mockito stubbing warnings

### LINKED ISSUE #2892: Reduce Mockito warnings
A lot of unit tests currently use 

```
@MockitoSettings(strictness = Strictness.WARN)
```

which produces a ton of „Unused“ warning for stubbings which are created in a `@BeforeEach` method but not used in every `@Test`. In my opinion warnings should only be printed for conditions that need attention and this is not the case here. It just makes the build output harder to analyze.

I therefore suggest to change to

```
@MockitoSettings(strictness = Strictness.LENIENT)
```

WDYT?

J-N-K:
Closes #2892 


wborn (approved):
Thanks for the noise reduction!

PR #2894, opened 2022-04-03
Allow changing watched directory of a WatchService

J-N-K:
The current implementation of the `AbstractWatchService` requires the directory to be known on instantiation. It is not possible to retrieve it from anywhere because the parent's constructor must be called first. The `WatchQueueReader` is customized (subscribed by WatchService instance) in `activate` and unsubscribed in `deactivate`. This PR adds a new method that allows changing the directory. If called, the old subscription is removed, the directory changed and the service activated again.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/service/AbstractWatchService.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/service/AbstractWatchService.java
@@ -52,7 +52,7 @@ public abstract class AbstractWatchService {
      * @param pathToWatch the new path
      */
     protected void changeWatchDirectory(String pathToWatch) {
-        watchQueueReader.stopWatchService(this);
+        deactivate();
         this.pathToWatch = pathToWatch;
         activate();
     }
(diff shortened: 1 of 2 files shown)


J-N-K:
This will not work.

bundles/org.openhab.core/src/main/java/org/openhab/core/service/AbstractWatchService.java
```diff
@@ -37,14 +37,25 @@
...
 
     /**
-     * The queue reader
+     * Change the watch directory for this WatchService
+     *
+     * @param pathToWatch the new path
      */
-    protected WatchQueueReader watchQueueReader = WatchQueueReader.getInstance();
+    protected void changeWatchDirectory(String pathToWatch) {
+        watchQueueReader.stopWatchService(this);
```
wborn, line 55: Would it also possible to call `deactivate()` here instead? Subclasses would then have a way to cleanup stuff that gets added/created in `activate()`.

wborn (approved):
Thanks!

PR #2895, opened 2022-04-04
Fix UnitUtils.getDimensionName returns Angle instead of Dimensionless

### LINKED ISSUE #2880: UnitUtils.getDimensionName wrong result
Dimensionless units like Units.DECIBEL, Units.PARTS_PER_MILLION, Units.PERCENT are reported as an "Angle" instead of "Dimensionless" by UnitUtils.getDimensionName()

Some binding have implemented a counter measure but root correction would be better : 
````
                String dimension = UnitUtils.getDimensionName(unit);
                dimension = (Units.DECIBEL.equals(unit) || Units.PARTS_PER_MILLION.equals(unit)
                        || Units.PERCENT.equals(unit)) ? "Dimensionless" // strangely it is given as an Angle
                                : dimension;
````

wborn:
[An Angle is also Dimensionless](https://en.wikipedia.org/wiki/Angle#Units) and because it is the first compatible unit it is always returned instead of any other Dimensionless unit.
With these changes `getDimensionName` will prefer the dimension of equal units over those of compatible units.

Fixes #2880

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/types/util/UnitUtils.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/types/util/UnitUtils.java
@@ -110,6 +110,7 @@ public class UnitUtils {
      * @return The Dimension string or null if the unit can not be found in any of the SystemOfUnits.
      */
     public static @Nullable String getDimensionName(Unit<?> unit) {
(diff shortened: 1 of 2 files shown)


kaikreuzer (approved):
Thanks for the fix!

PR #2897, opened 2022-04-06
Fix memory leak in ScriptedRuleProvider

J-N-K:
The Rule was never removed from the internal tracking map and therefore allocated resources were never freed.

### DIFF
--- a/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/shared/ScriptedRuleProvider.java
+++ b/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/shared/ScriptedRuleProvider.java
@@ -60,7 +60,7 @@ public class ScriptedRuleProvider implements RuleProvider {
     }
 
     public void removeRule(String ruleUID) {
-        Rule rule = rules.get(ruleUID);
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/shared/ScriptedRuleProvider.java
```diff
@@ -60,13 +60,14 @@ public void addRule(Rule rule) {
     }
 
     public void removeRule(String ruleUID) {
-        Rule rule = rules.get(ruleUID);
+        Rule rule = rules.remove(ruleUID);
```
kaikreuzer, line 63: Isn't this now duplicating the removal from `rules`?
I'd think it would be cleaner to keep the "get" here and have it only removed in `removeRule(Rule rule)`, wdyt?

J-N-K: I thought about that, too. This way we avoid possible race conditions where the method with the id called twice and then calls the removal of the rule a second time. But that’s probably a very rare case, because it would require a bug at another place that calls remove. I‘ll change it as suggested.

kaikreuzer (approved):
Thank you!

PR #2901, opened 2022-04-07
Do not allow empty last segment in UID

### LINKED ISSUE #2900: ThingUID constructor allows creation of invalid Thing UID
I installed a binding (epsonprojector) whose discovery service added an invalid thing to my inbox. I can't remove this discovery result using the UI (an exception is thrown). I'll probably need to hand edit the jsondb to remove it.

This is the UID of the thing in my inbox (note there's nothing after the final colon):
```
epsonprojector:projector-tcp:
```

The problem is that under certain circumstances the `epsonprojector` binding is calling the `ThingUID` constructor with the `id` parameter containing an empty string.
```
new ThingUID (THING_TYPE_UID, "")
```

Somebody (possibly me, if necessary) will submit a fix to the binding to prevent this. But, shouldn't the `ThingUID` constructor also defend against an empty string being passed as an `id`?

J-N-K:
Fixes #2900 

The last segment is not allowed to be empty. Also adds tests for UID construction.

Originally I thought (and still think) that empty segments make no sense at any position. But at least for `ThingUID` there is a test that ensures that a missing thing-type is considered a valid UID.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/common/AbstractUID.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/common/AbstractUID.java
@@ -75,6 +75,9 @@ public abstract class AbstractUID {
             String segment = segments.get(i);
             validateSegment(segment, i, numberOfSegments);
         }
+        if (segments.get(numberOfSegments).isBlank()) {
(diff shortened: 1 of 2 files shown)


kaikreuzer (approved):
lgtm

PR #2905, opened 2022-04-09
Allow adding/changing Jetty server certificates via REST API

J-N-K:
Depends on https://github.com/openhab/openhab-distro/pull/1383

This adds a REST resource to add/change the Jetty server certificate. An empty DTO resets the certificate to a self-signed certificate.

Originally I thought about a `WatchService` for certificate files but this seems unnecessary. If certificates are changed on a regular basis (like with Let's encrypt), it is easy to add a post-hook script that uses keytool to insert the certificates retrieved in the keystore. For changing long-term certificates, the REST API is enough.


splatch:
My sense of what this change is tells me that it could deserve a 'feature' label as it brings a certificate management to upper layers than they were before.

wborn:
Why not use the existing settings mechanism?

wborn:
A separate REST endpoint might make sense if we can also use it to add other self signed certificates to the keystore via the UI: https://github.com/openhab/openhab-addons/issues/10446

ghys:
I appreciate the general principle, but I actually have professional experience in this (PKIs and stuff), and I would only make this word of caution because I just have to: you're allowing a security feature to be altered by an API, so the weakest link becomes the API, and the OH REST API is not the strongest... most of the instances are not even accessed with HTTPS.

I would strongly advise you to reconsider this or secure it properly as the certificate and private key provide more security than this endpoint (because it allows to change them).

Issue #2904, opened 2022-04-09
[HumanLanguageInterpreter] Allow HLI chaining


GiviMAD:
I have this open PR on the add-ons https://github.com/openhab/openhab-addons/pull/12260 for a customizable HLI and I would like to be able to use it in combination with other interpreters.

It is not quite clear for me how this can be done. I was thinking on adding a new method to the HLI interface which returns a boolean (maybe called 'understand' or something similar) so the core can check if a certain HLI can process a specific text while keeping this logic inside each interpreter, then we just need to implement a way of sorting the interpreters.

WDYT?

Also I would like to comment there is getGrammar method on the HLI interface, I think the intention for this one was to export the HLI capabilities in a common format so the core can analyze it, but I see this as a difficult feature to implement in all the interpreters, so I think it could be removed if we add the new one.

I can work on it if anybody is interested, but I would like to get some feedback from the maintainers and other contributors.

Regards!

lolodomo:
We have currently this method:
```
    /**
     * Interprets a human language text fragment of a given {@link Locale}
     *
     * @param locale language of the text (given by a {@link Locale})
     * @param text the text to interpret
     * @return a human language response
     */
    String interpret(Locale locale, String text) throws InterpretationException;
```
Why not simply considering throwing InterpretationException when the HLI does not understand the text ? We should check if this is the current solution implemented with existing HLIs.
We should of course also check that this exception is correctly handled everywhere this method is called. This is done correctly in the dialog processor for example.

Then your idea is to have an ordered list of HLI to consider until one of them is able to handle a text. This could be done by a replacing the HLI parameter in the dialog processor by an ordered list of HLI. All "callers" should then be updated to replace the HLI ID string parameter by a comma separated list of HLI IDs.
Some of the rule actions which allow the user to provide a HLI ID should also be updated the same way to give the user the opportunity to provide a list of HLI.
This change will be backward compatible.

And regarding your comment about getGrammar, I agree with you.

WDYT ?

lolodomo:
Regarding the first point, it looks like the system built-in HLI is well implemented, meaning it throws an exception if the text cannot be understood.:
https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/text/AbstractRuleBasedInterpreter.java#L110

[...]

PR #2906, opened 2022-04-10
[Voice] Allow hli list

GiviMAD:

This is a proposed solution for https://github.com/openhab/openhab-core/issues/2904 based on the feedback from @lolodomo.

Basically allows to pass a comma separate list of hli ids instead of a single one.

I have some doubts about renaming the hli vars, and the 'getHLI' methods, as there is already a getHLIs method that returns all the available interpreters, let me know if it's better to rename them.


GiviMAD:
ups, adding WIP

lolodomo:
You should keep getHLI unchanged and add a new method taking a comma separated list of IDs and returning a list of HLI. . 

Note the typo: comma separated list

GiviMAD:
> You should keep getHLI unchanged and add a new method taking a comma separated list of IDs and returning a list of HLI

Ok, there will be no uses for the 'getHLI(String id)' method other than inside the own VoiceManager, as all the places will call the new method 'getHLIByIds(String idList)'. Is this ok?

lolodomo:
I think it is better, it avoids breaking the interface and it avoids having a method name not matching what it does.

GiviMAD:
> I think it is better, it avoids breaking the interface and it avoids having a method name not matching what it does.

Yes it makes sense, specially the last part, thanks for the explanation.

GiviMAD:
I forgot to apply spotless last time.

GiviMAD:
Now is ready for review.

bundles/org.openhab.core.io.rest.voice/src/main/java/org/openhab/core/io/rest/voice/internal/VoiceResource.java
```diff
@@ -168,17 +187,32 @@ public Response interpret(
             @HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) @Parameter(description = "language") @Nullable String language,
             @Parameter(description = "text to interpret", required = true) String text) {
         final Locale locale = localeService.getLocale(language);
-        HumanLanguageInterpreter hli = voiceManager.getHLI();
+        Collection<HumanLanguageInterpreter> hli = voiceManager.getDefaultHLIs();
```
lolodomo, line 190: Please do not update this REST API. Its description is clear: "Sends a text to the default human language interpreter".
There is only one default HLI.

bundles/org.openhab.core.io.rest.voice/src/main/java/org/openhab/core/io/rest/voice/internal/VoiceResource.java
```diff
@@ -264,9 +298,9 @@ public Response startDialog(
                 return JSONResponse.createErrorResponse(Status.NOT_FOUND, "Text-to-Speech not found");
             }
         }
-        HumanLanguageInterpreter hli = null;
+        Collection<HumanLanguageInterpreter> hli = null;
         if (hliId != null) {
-            hli = voiceManager.getHLI(hliId);
```
lolodomo, line 269: Please update the parameter description, rename it into `hliIds` and rename your local variable into `hlis` or `interpreters`.

bundles/org.openhab.core.io.rest.voice/src/main/java/org/openhab/core/io/rest/voice/internal/VoiceResource.java
```diff
@@ -349,9 +383,9 @@ public Response listenAndAnswer(
...
+            hli = voiceManager.getHLIByIds(hliId);
```
lolodomo, line 388: Please update the parameter description, rename it into `hliIds` and rename your local variable into `hlis` or `interpreters`.

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/actions/Voice.java
```diff
@@ -157,7 +158,7 @@ public static String interpret(@ParamDoc(name = "text") Object text) {
      * In case of interpretation error, the error message is played using the default audio sink.
      *
      * @param text The text to interpret
-     * @param interpreter The Human Language Interpreter to be used
+     * @param interpreter Comma separated list of human language text interpreters to use or null to use the default services
```
lolodomo, line 161: Please rename the parameter into `interpreters`
This parameter should remain non null as you have another action without this parameter. So please remove the last part of the description.

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/actions/Voice.java
```diff
@@ -180,7 +181,7 @@ public static String interpret(@ParamDoc(name = "text") Object text,
      * If sink parameter is null, the error message is simply not played.
      *
      * @param text The text to interpret
-     * @param interpreter The Human Language Interpreter to be used
+     * @param interpreter Comma separated list of human language text interpreters to use or null to use the default services
```
lolodomo, line 184: Please rename the parameter into `interpreters`
This parameter should remain non null. So please remove the last part of the description.

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/actions/Voice.java
```diff
@@ -218,10 +219,10 @@ public static void startDialog(@ParamDoc(name = "source") @Nullable String sourc
      * @param ks the keyword spotting service to use or null to use the default service
      * @param stt the speech-to-text service to use or null to use the default service
      * @param tts the text-to-speech service to use or null to use the default service
-     * @param interpreter the human language text interpreter to use or null to use the default service
+     * @param interpreter comma separated list of human language text interpreters to use or null to use the default services
```
lolodomo, line 222: Please rename the parameter into `interpreters`
... or null to use the default **service**

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/actions/Voice.java
```diff
@@ -339,10 +340,10 @@ public static void listenAndAnswer(@ParamDoc(name = "source") @Nullable String s
      *
      * @param stt the speech-to-text service to use or null to use the default service
      * @param tts the text-to-speech service to use or null to use the default service
-     * @param interpreter the human language text interpreter to use or null to use the default service
+     * @param interpreter  comma separated list of human language text interpreters to use or null to use the default services
```
lolodomo, line 343: Please rename the parameter into `interpreters`
... or null to use the default **service**

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/VoiceManager.java
```diff
@@ -106,7 +106,7 @@ public interface VoiceManager {
      * Interprets the passed string using a particular HLI service and the default locale.
      *
      * @param text The text to interpret
-     * @param hliId The id of the HLI service to use or null
+     * @param hliId Comma separated list of HLI service ids to use or null
```
lolodomo, line 109: Please rename the parameter into `hliIdList` (name you are using in your implementation and looks good to me)

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/VoiceManager.java
```diff
@@ -144,17 +144,17 @@ public interface VoiceManager {
      * @param ks the keyword spotting service to use or null to use the default service
      * @param stt the speech-to-text service to use or null to use the default service
      * @param tts the text-to-speech service to use or null to use the default service
-     * @param hli the human language text interpreter to use or null to use the default service
+     * @param hli collection of human language text interpreters to use or null to use the default services
```
lolodomo, line 147: Please rename the parameter into `hlis`
... the default **service**

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/VoiceManager.java
```diff
@@ -187,17 +187,17 @@ void startDialog(@Nullable KSService ks, @Nullable STTService stt, @Nullable TTS
      *
      * @param stt the speech-to-text service to use or null to use the default service
      * @param tts the text-to-speech service to use or null to use the default service
-     * @param hli the human language text interpreter to use or null to use the default service
+     * @param hli collection of human language text interpreters to use or null to use the default services
```
lolodomo, line 190: Please rename the parameter into `hlis`
... the default **service**

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/VoiceManager.java
```diff
@@ -280,6 +280,29 @@ void listenAndAnswer(@Nullable STTService stt, @Nullable TTSService tts, @Nullab
...
+    /**
+     * Retrieves a HumanLanguageInterpreter collection.
+     * If a default names are configured and the services are available, those are returned. Otherwise, the first
+     * available service is returned.
+     * 
+     * @return a Collection<HumanLanguageInterpreter> or null, if no services are available or if some defaults are
+     *         configured, but no according services are found.
+     */
+    @Nullable
+    Collection<HumanLanguageInterpreter> getDefaultHLIs();
```
lolodomo, line 304: I really think we should keep a unique default interpreter.
So I would suggest to remove this method.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/VoiceManager.java
```diff
@@ -280,6 +280,29 @@ void listenAndAnswer(@Nullable STTService stt, @Nullable TTSService tts, @Nullab
...
+     * Retrieves a HumanLanguageInterpreter collection.
+     * If no services are available returns null.
+     * 
+     * @param idList Comma separated list of HLI service ids to use or null
+     * @return a Collection<HumanLanguageInterpreter> or null, if no services are available or if some defaults are
+     *         configured, but no
+     *         according services are found
+     */
+    @Nullable
+    Collection<HumanLanguageInterpreter> getHLIByIds(String idList);
```
lolodomo, line 293: I would suggest rather this name: `getHLIsByIds`

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
```diff
@@ -74,7 +75,7 @@ public class DialogProcessor implements KSListener, STTListener {
     private final @Nullable KSService ks;
     private final STTService stt;
     private final TTSService tts;
-    private final HumanLanguageInterpreter hli;
+    private final Collection<HumanLanguageInterpreter> hli;
```
lolodomo, line 78: Please rename the variable into `hlis`

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
```diff
@@ -104,7 +105,7 @@ public class DialogProcessor implements KSListener, STTListener {
     private @Nullable AudioStream streamKS;
     private @Nullable AudioStream streamSTT;
 
-    public DialogProcessor(KSService ks, STTService stt, TTSService tts, HumanLanguageInterpreter hli,
+    public DialogProcessor(KSService ks, STTService stt, TTSService tts, Collection<HumanLanguageInterpreter> hli,
```
lolodomo, line 108: Please rename the parameter into `hlis`

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
```diff
@@ -124,7 +125,7 @@ public DialogProcessor(KSService ks, STTService stt, TTSService tts, HumanLangua
         this.ttsFormat = VoiceManagerImpl.getBestMatch(tts.getSupportedFormats(), sink.getSupportedFormats());
     }
 
-    public DialogProcessor(STTService stt, TTSService tts, HumanLanguageInterpreter hli, AudioSource source,
+    public DialogProcessor(STTService stt, TTSService tts, Collection<HumanLanguageInterpreter> hli, AudioSource source,
```
lolodomo, line 128: Please rename the parameter into `hlis`

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
```diff
@@ -282,15 +283,24 @@ public synchronized void sttEventReceived(STTEvent sttEvent) {
...
+                        break;
+                    } catch (InterpretationException e) {
+                        logger.debug("Interpretation exception: {}", e.getMessage());
+                        var msg = e.getMessage();
+                        if (msg != null) {
+                            error = msg;
+                        }
+                    }
                 }
+                toggleProcessing(false);
```
lolodomo, line 302: This should probably remain before calling the interpreters.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceConsoleCommandExtension.java
```diff
@@ -135,7 +135,8 @@ public void execute(String[] args, Console console) {
                     try {
                         AudioSource source = args.length < 2 ? null : audioManager.getSource(args[1]);
                         AudioSink sink = args.length < 3 ? null : audioManager.getSink(args[2]);
-                        HumanLanguageInterpreter hli = args.length < 4 ? null : voiceManager.getHLI(args[3]);
+                        Collection<HumanLanguageInterpreter> hli = args.length < 4 ? null
```
lolodomo, line 138: Rather rename the variable into `hlis`

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceConsoleCommandExtension.java
```diff
@@ -158,7 +159,8 @@ public void execute(String[] args, Console console) {
...
```
lolodomo, line 162: Rather rename the variable into `hlis`

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceConsoleCommandExtension.java
```diff
@@ -243,11 +245,12 @@ private void say(String[] args, Console console) {
     private void listInterpreters(Console console) {
         Collection<HumanLanguageInterpreter> interpreters = voiceManager.getHLIs();
         if (!interpreters.isEmpty()) {
-            HumanLanguageInterpreter defaultHLI = voiceManager.getHLI();
+            var defaultHLIs = voiceManager.getDefaultHLIs();
```
lolodomo, line 248: As previously mentioned, I believe we should keep only one default interpreter.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -485,34 +502,34 @@ public void startDialog() throws IllegalStateException {
 
     @Override
     public void startDialog(@Nullable KSService ks, @Nullable STTService stt, @Nullable TTSService tts,
-            @Nullable HumanLanguageInterpreter hli, @Nullable AudioSource source, @Nullable AudioSink sink,
+            @Nullable Collection<HumanLanguageInterpreter> hli, @Nullable AudioSource source, @Nullable AudioSink sink,
```
lolodomo, line 505: Rather `hlis`

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -288,20 +288,37 @@ public String interpret(String text) throws InterpretationException {
...
     @Override
-    public String interpret(String text, @Nullable String hliId) throws InterpretationException {
-        HumanLanguageInterpreter interpreter;
-        if (hliId == null) {
-            interpreter = getHLI();
-            if (interpreter == null) {
+    public String interpret(String text, @Nullable String hliIdList) throws InterpretationException {
+        Collection<HumanLanguageInterpreter> interpreters;
+        if (hliIdList == null) {
+            interpreters = getDefaultHLIs();
```
lolodomo, line 294: `List.of(getHLI())` instead of `getDefaultHLIs()`

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -485,34 +502,34 @@ public void startDialog() throws IllegalStateException {
...
-            @Nullable HumanLanguageInterpreter hli, @Nullable AudioSource source, @Nullable AudioSink sink,
+            @Nullable Collection<HumanLanguageInterpreter> hli, @Nullable AudioSource source, @Nullable AudioSink sink,
             @Nullable Locale locale, @Nullable String keyword, @Nullable String listeningItem)
             throws IllegalStateException {
         // use defaults, if null
         KSService ksService = (ks == null) ? getKS() : ks;
         STTService sttService = (stt == null) ? getSTT() : stt;
         TTSService ttsService = (tts == null) ? getTTS() : tts;
-        HumanLanguageInterpreter interpreter = (hli == null) ? getHLI() : hli;
+        Collection<HumanLanguageInterpreter> interpreters = (hli == null) ? getDefaultHLIs() : hli;
```
lolodomo, line 512: `List.of(getHLI())` instead of `getDefaultHLIs()`

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -556,31 +573,31 @@ public void listenAndAnswer() throws IllegalStateException {
 
     @Override
     public void listenAndAnswer(@Nullable STTService stt, @Nullable TTSService tts,
-            @Nullable HumanLanguageInterpreter hli, @Nullable AudioSource source, @Nullable AudioSink sink,
+            @Nullable Collection<HumanLanguageInterpreter> hli, @Nullable AudioSource source, @Nullable AudioSink sink,
```
lolodomo, line 576: Rather `hlis`

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -556,31 +573,31 @@ public void listenAndAnswer() throws IllegalStateException {
...
             @Nullable Locale locale, @Nullable String listeningItem) throws IllegalStateException {
         // use defaults, if null
         STTService sttService = (stt == null) ? getSTT() : stt;
         TTSService ttsService = (tts == null) ? getTTS() : tts;
-        HumanLanguageInterpreter interpreter = (hli == null) ? getHLI() : hli;
+        Collection<HumanLanguageInterpreter> interpreters = (hli == null) ? getDefaultHLIs() : hli;
```
lolodomo, line 581: `List.of(getHLI())` instead of `getDefaultHLIs()`

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -736,11 +753,47 @@ public Collection<KSService> getKSs() {
         return hli;
     }
 
+    @Override
+    public @Nullable Collection<HumanLanguageInterpreter> getDefaultHLIs() {
```
lolodomo, line 757: To be removed

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -736,11 +753,47 @@ public Collection<KSService> getKSs() {
...
         return humanLanguageInterpreters.get(id);
     }
 
+    @Override
+    public @Nullable Collection<HumanLanguageInterpreter> getHLIByIds(String idList) {
+        List<HumanLanguageInterpreter> interpreters = new ArrayList<>();
+        for (var id : idList.split(",")) {
+            var hli = humanLanguageInterpreters.get(id);
+            if (hli == null) {
+                logger.warn("Default HumanLanguageInterpreter '{}' not available!", id);
```
lolodomo, line 786: "Default" to be removed from the message

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -736,11 +753,47 @@ public Collection<KSService> getKSs() {
...
+            } else {
+                interpreters.add(hli);
+            }
+        }
+        if (interpreters.isEmpty()) {
```
lolodomo, line 791: Shouldn't we return null as soon as one of the interpreter is not found (rather than ignoring it) ?

lolodomo (changes requested):

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/VoiceManager.java
GiviMAD: But if we remove this it will be no possible to enable multi selection on the ui, I was planning to open a followup issue on the ui to ask for this functionally.
I think it makes sense, is something common that you want to use the same set of interpreters in all the places, if we remove this we are forcing the user to pass the interpreter ids in all places.
Allowing a list of interpreters on the call but not allowing it to be configured as default seems a little awkward for me.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
GiviMAD: I also have doubts about it.
As I was thinking on support configuring a chain of interpreters as default I though it makes more sense not to fail, to prevent errors in case one of the configured interpreters is uninstalled.

GiviMAD:
@lolodomo I have solve part of the review.
Can you explain me the reason why you think is better to support just one default interpreter while allowing to use multiples? For me it breaks a little the default functionality as if you want to always use multiple interpreters (this is my case) you can not rely on the default value anymore.

GiviMAD:
@lolodomo I have finally followed your advice so I can complete the review.
I'll open an issue on the future to discuss there if it is a good idea to support multiple interpreters as default or not.

bundles/org.openhab.core.io.rest.voice/src/main/java/org/openhab/core/io/rest/voice/internal/VoiceResource.java
```diff
@@ -133,7 +137,7 @@ public Response getInterpreter(
     }
 
     @POST
-    @Path("/interpreters/{id: [a-zA-Z_0-9]+}")
+    @Path("/interpreters/{ids: [a-zA-Z_0-9,]+}")
```
lolodomo, line 140: For compatibility reason, I think we should keep `id`.

bundles/org.openhab.core.io.rest.voice/src/main/java/org/openhab/core/io/rest/voice/internal/VoiceResource.java
```diff
@@ -142,19 +146,34 @@ public Response getInterpreter(
     public Response interpret(
             @HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) @Parameter(description = "language") @Nullable String language,
             @Parameter(description = "text to interpret", required = true) String text,
-            @PathParam("id") @Parameter(description = "interpreter id") String id) {
+            @PathParam("ids") @Parameter(description = "comma separated list of interpreter ids") String ids) {
```
lolodomo, line 149: Related to my previous comment:
```
@PathParam("id") @Parameter(description = "comma separated list of interpreter ids") String ids) {
```

bundles/org.openhab.core.io.rest.voice/src/main/java/org/openhab/core/io/rest/voice/internal/VoiceResource.java
```diff
@@ -133,7 +137,7 @@ public Response getInterpreter(
     }
 
     @POST
-    @Path("/interpreters/{id: [a-zA-Z_0-9]+}")
+    @Path("/interpreters/{ids: [a-zA-Z_0-9,]+}")
     @Consumes(MediaType.TEXT_PLAIN)
     @Operation(operationId = "interpretText", summary = "Sends a text to a given human language interpreter.", responses = {
```
lolodomo, line 142: You could adjust the summary with "Sends a text to given human language interpreter(s)."

bundles/org.openhab.core.io.rest.voice/src/main/java/org/openhab/core/io/rest/voice/internal/VoiceResource.java
```diff
@@ -142,19 +146,34 @@ public Response getInterpreter(
...
+                    error = msg;
+                }
+            }
+        }
+        if (answer != null) {
+            return Response.ok(answer, MediaType.TEXT_PLAIN).build();
+        } else if (error != null) {
+            return JSONResponse.createErrorResponse(Status.BAD_REQUEST, error);
         }
+        return Response.ok("", MediaType.TEXT_PLAIN).build();
```
lolodomo, line 176: I suggest rather:
```
        if (error != null) {
            return JSONResponse.createErrorResponse(Status.BAD_REQUEST, error);
        } else {
            return Response.ok(answer == null ? "" : answer, MediaType.TEXT_PLAIN).build();
        }
```
As the previous code was building the OK message with null, I am not sure we have to check if answer is null or not. Maybe this code is OK:
```
        if (error != null) {
            return JSONResponse.createErrorResponse(Status.BAD_REQUEST, error);
        } else {
            return Response.ok(answer, MediaType.TEXT_PLAIN).build();
        }
```

bundles/org.openhab.core.io.rest.voice/src/main/java/org/openhab/core/io/rest/voice/internal/VoiceResource.java
```diff
@@ -325,7 +344,7 @@ public Response listenAndAnswer(
             @QueryParam("sourceId") @Parameter(description = "source ID") @Nullable String sourceId,
             @QueryParam("sttId") @Parameter(description = "Speech-to-Text ID") @Nullable String sttId,
             @QueryParam("ttsId") @Parameter(description = "Text-to-Speech ID") @Nullable String ttsId,
-            @QueryParam("hliId") @Parameter(description = "interpreter ID") @Nullable String hliId,
+            @QueryParam("hliIds") @Parameter(description = "interpreter ID") @Nullable String hliIds,
```
lolodomo, line 347:
```
          @QueryParam("hliIds") @Parameter(description = "comma separated list of interpreter IDs") @Nullable String hliIds,
```

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/actions/Voice.java
```diff
@@ -339,16 +340,16 @@ public static void listenAndAnswer(@ParamDoc(name = "source") @Nullable String s
      *
      * @param stt the speech-to-text service to use or null to use the default service
      * @param tts the text-to-speech service to use or null to use the default service
-     * @param interpreter the human language text interpreter to use or null to use the default service
+     * @param interpreters  comma separated list of human language text interpreters to use or null to use the default service
```
lolodomo, line 343: One space to remove

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/actions/Voice.java
```diff
@@ -157,14 +158,14 @@ public static String interpret(@ParamDoc(name = "text") Object text) {
      * In case of interpretation error, the error message is played using the default audio sink.
      *
      * @param text The text to interpret
-     * @param interpreter The Human Language Interpreter to be used
+     * @param interpreters Comma separated list of human language text interpreters to use
      */
     @ActionDoc(text = "interprets a given text by a given human language interpreter", returns = "human language response")
```
lolodomo, line 163: "interprets a given text by given human language interpreter(s)"

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/actions/Voice.java
```diff
@@ -180,15 +181,15 @@ public static String interpret(@ParamDoc(name = "text") Object text,
      * If sink parameter is null, the error message is simply not played.
      *
      * @param text The text to interpret
-     * @param interpreter The Human Language Interpreter to be used
+     * @param interpreters Comma separated list of human language text interpreters to use
      * @param sink The name of audio sink to be used to play the error message
      */
     @ActionDoc(text = "interprets a given text by a given human language interpreter", returns = "human language response")
```
lolodomo, line 187: "interprets a given text by given human language interpreter(s)"

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/VoiceManager.java
```diff
@@ -106,11 +106,11 @@ public interface VoiceManager {
      * Interprets the passed string using a particular HLI service and the default locale.
```
lolodomo, line 106: Please update the description to explain how is handled the comma separated list of interpreters.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/VoiceManager.java
```diff
@@ -144,17 +144,17 @@ public interface VoiceManager {
      * @param ks the keyword spotting service to use or null to use the default service
      * @param stt the speech-to-text service to use or null to use the default service
      * @param tts the text-to-speech service to use or null to use the default service
-     * @param hli the human language text interpreter to use or null to use the default service
+     * @param hlis collection of human language text interpreters to use or null to use the default service
```
lolodomo, line 147: A small sentence in the description to explain the principle of this collection would be appreciated.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/VoiceManager.java
```diff
@@ -187,17 +187,17 @@ void startDialog(@Nullable KSService ks, @Nullable STTService stt, @Nullable TTS
      *
      * @param stt the speech-to-text service to use or null to use the default service
      * @param tts the text-to-speech service to use or null to use the default service
-     * @param hli the human language text interpreter to use or null to use the default service
+     * @param hlis collection of human language text interpreters to use or null to use the default service
```
lolodomo, line 190: A small sentence in the description to explain the principle of this collection would be appreciated.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/VoiceManager.java
```diff
@@ -280,6 +280,18 @@ void listenAndAnswer(@Nullable STTService stt, @Nullable TTSService tts, @Nullab
      */
     Collection<KSService> getKSs();
 
+    /**
+     * Retrieves a HumanLanguageInterpreter collection.
+     * If no services are available returns null.
+     * 
+     * @param idList Comma separated list of HLI service ids to use or null
```
lolodomo, line 287: "or null" to be removed

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -485,34 +503,40 @@ public void startDialog() throws IllegalStateException {
...
         KSService ksService = (ks == null) ? getKS() : ks;
         STTService sttService = (stt == null) ? getSTT() : stt;
         TTSService ttsService = (tts == null) ? getTTS() : tts;
-        HumanLanguageInterpreter interpreter = (hli == null) ? getHLI() : hli;
+        Collection<HumanLanguageInterpreter> interpreters = hlis;
+        if (hlis == null) {
+            HumanLanguageInterpreter hli = getHLI();
+            if (hli != null) {
+                interpreters = List.of(hli);
+            }
```
lolodomo, line 518: `interpreters` should be set to null if there is no default:
```
interpreters = hli == null ? null : List.of(hli);
```

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -556,31 +580,37 @@ public void listenAndAnswer() throws IllegalStateException {
...
         // use defaults, if null
         STTService sttService = (stt == null) ? getSTT() : stt;
         TTSService ttsService = (tts == null) ? getTTS() : tts;
-        HumanLanguageInterpreter interpreter = (hli == null) ? getHLI() : hli;
+        Collection<HumanLanguageInterpreter> interpreters = hlis;
+        if (hlis == null) {
+            HumanLanguageInterpreter hli = getHLI();
+            if (hli != null) {
+                interpreters = List.of(hli);
+            }
```
lolodomo, line 593: `interpreters` should be set to null if there is no default:
```
interpreters = hli == null ? null : List.of(hli);
```

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
```diff
@@ -282,15 +283,24 @@ public synchronized void sttEventReceived(STTEvent sttEvent) {
...
+                        break;
+                    } catch (InterpretationException e) {
+                        logger.debug("Interpretation exception: {}", e.getMessage());
+                        var msg = e.getMessage();
+                        if (msg != null) {
+                            error = msg;
+                        }
+                    }
                 }
+                say(answer != null ? answer : error);
```
lolodomo, line 303: Shouldn't it be rather ?
```
say(error != null ? error : answer);
```

lolodomo (changes requested):
Still few comments.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
lolodomo: Ok but in this case, please mention it in the method description (in interface).

bundles/org.openhab.core.io.rest.voice/src/main/java/org/openhab/core/io/rest/voice/internal/VoiceResource.java
J-N-K: Since this is internal only, you can change to `ids` without any issue.

bundles/org.openhab.core.io.rest.voice/src/main/java/org/openhab/core/io/rest/voice/internal/VoiceResource.java
J-N-K: You should be able to implicitly deserialize it to a list (change to `List<String>`).

bundles/org.openhab.core.io.rest.voice/src/main/java/org/openhab/core/io/rest/voice/internal/VoiceResource.java
```diff
@@ -232,7 +251,7 @@ public Response startDialog(
             @QueryParam("ksId") @Parameter(description = "keywork spotter ID") @Nullable String ksId,
             @QueryParam("sttId") @Parameter(description = "Speech-to-Text ID") @Nullable String sttId,
             @QueryParam("ttsId") @Parameter(description = "Text-to-Speech ID") @Nullable String ttsId,
-            @QueryParam("hliId") @Parameter(description = "interpreter ID") @Nullable String hliId,
+            @QueryParam("hliIds") @Parameter(description = "comma separated list of interpreter IDs") @Nullable String hliIds,
```
J-N-K, line 254: This is not internal. Please keep the old `hliId`, mark it as deprecated and properly handle it in code. The new `hliIds` should return a `List<String>`, not a `String`.

bundles/org.openhab.core.io.rest.voice/src/main/java/org/openhab/core/io/rest/voice/internal/VoiceResource.java
J-N-K: See above.

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/actions/Voice.java
J-N-K: Please do not break API. Add new action which accepts a `List<String>`. You can call the new action from the "old" one by using `List.of(interpreter)`. Also below at other places.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/VoiceManager.java
```diff
@@ -106,11 +106,11 @@ public interface VoiceManager {
      * Interprets the passed string using a particular HLI service and the default locale.
      *
      * @param text The text to interpret
-     * @param hliId The id of the HLI service to use or null
+     * @param hliIdList Comma separated list of HLI service ids to use or null
      * @throws InterpretationException
      * @return a human language response
      */
-    String interpret(String text, @Nullable String hliId) throws InterpretationException;
+    String interpret(String text, @Nullable String hliIdList) throws InterpretationException;
```
J-N-K, line 113: See above, don't remove, add to API if possible. Use `List<String>` for multiple interpreters.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
```diff
@@ -282,15 +283,24 @@ public synchronized void sttEventReceived(STTEvent sttEvent) {
...
+                    try {
+                        answer = interpreter.interpret(locale, question);
+                        logger.debug("Interpretation result: {}", answer);
+                        error = null;
+                        break;
+                    } catch (InterpretationException e) {
+                        logger.debug("Interpretation exception: {}", e.getMessage());
+                        var msg = e.getMessage();
+                        if (msg != null) {
+                            error = msg;
```
J-N-K, line 299: If there is more than one error message, earlier messages are discarded. Is that intentional?

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
J-N-K: Probably, yes.

J-N-K (changes requested):
Sorry for being late to the party. I have one general question (which is probably due to me missing knowledge on this part, I never used HLI): If there are multiple HLI and they are all listed, the first result is taken. Could it happen that the user is interested in the result of another HLI? Since `Collection` in general does not guarantee order, the user can't control which HLI is queried first.

bundles/org.openhab.core.io.rest.voice/src/main/java/org/openhab/core/io/rest/voice/internal/VoiceResource.java
lolodomo: Yes, correct. My comment was useless. You can ignore it.

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/actions/Voice.java
lolodomo: This is not a breaking change because the parameter is a string.
But why not adding a new method for the case we want to provide a list of interpreters.
But I am not sure that inside a DSL rule for example, it is possible to build a List of objects.

bundles/org.openhab.core.io.rest.voice/src/main/java/org/openhab/core/io/rest/voice/internal/VoiceResource.java
lolodomo: The String type was a way to keep compatibility.
But I agree with you with th equery param which must not be renamed.

lolodomo:
> If there are multiple HLI and they are all listed, the first result is taken. Could it happen that the user is interested in the result of another HLI? Since `Collection` in general does not guarantee order, the user can't control which HLI is queried first.

The order was very important. The user just wants to get the result of the first interpreter able to interpret.
So `Collection` was a bad choice. I have this question in mind and finally totally forgot it !

lolodomo:
Considering @J-N-K comments, I think we should duplicate REST API and actions for "interpret", keeping the existing ones unchanged.

For "startDialog" and "listenAndAnswer", as this is recent features added, we can directly consider the list of interpreters as parameter without duplicating them.

For REST API, having a List<String> as parameter is not a problem.
But for actions, will it be possible to pass a list of strings as parameter in a rule ? What would be the syntax in a DSL rule for example ?

And finally, is `List` the right class for an ordered list ?

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
GiviMAD: already initialized here:
```
List<HumanLanguageInterpreter> interpreters = hlis;
```

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
GiviMAD: already initialized here:
```
List<HumanLanguageInterpreter> interpreters = hlis;
```

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
GiviMAD: Yes, this is done in order to find the first interpreter which returns a successful response. 

Maybe this can be improved in a separate pr by adding a subclass of the general Interpretation Exception to ensure those are unsuccessful responses instead of any other execution error. WDYT?

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
lolodomo: Yes, you're right because hlis is null in this case.
Comment to be ignored.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
lolodomo: Yes, you're right because hlis is null in this case.
Comment to be ignored.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/VoiceManager.java
GiviMAD: As @lolodomo said, we are not really breaking the api here. 
After taking a look all the consumers of this method requires allow passing many interpreters, so if I create another one which accepts List<string> all the consumer will use that instead with the inconvenience of handling the "split by comma" operation, so I think this is want we want. Reopen this if I'm wrong please.

GiviMAD:
@J-N-K @lolodomo, I think I've taken all the comments in account. Let me know if I have missed something.

bundles/org.openhab.core.io.rest.voice/src/main/java/org/openhab/core/io/rest/voice/internal/VoiceResource.java
```diff
@@ -325,7 +348,8 @@ public Response listenAndAnswer(
             @QueryParam("sourceId") @Parameter(description = "source ID") @Nullable String sourceId,
             @QueryParam("sttId") @Parameter(description = "Speech-to-Text ID") @Nullable String sttId,
             @QueryParam("ttsId") @Parameter(description = "Text-to-Speech ID") @Nullable String ttsId,
-            @QueryParam("hliId") @Parameter(description = "interpreter ID") @Nullable String hliId,
+            @Deprecated @QueryParam("hliId") @Parameter(description = "interpreter ID") @Nullable String hliId,
```
J-N-K, line 351:
```suggestion
            @QueryParam("hliId") @Parameter(description = "interpreter Id", deprecated=true) @Nullable String hliId,
```

J-N-K (changes requested):

bundles/org.openhab.core.io.rest.voice/src/main/java/org/openhab/core/io/rest/voice/internal/VoiceResource.java
```diff
@@ -133,27 +137,39 @@ public Response getInterpreter(
...
+        for (var interpreter : hlis) {
+            try {
+                answer = interpreter.interpret(locale, text);
+                logger.debug("Interpretation result: {}", answer);
+                error = null;
+                break;
+            } catch (InterpretationException e) {
+                logger.debug("Interpretation exception: {}", e.getMessage());
+                var msg = e.getMessage();
+                error = Objects.requireNonNullElse(msg, "Unexpected error");
```
J-N-K, line 166:
```suggestion
                error = Objects.requireNonNullElse(e.getMessage(), "Unexpected error");
```

bundles/org.openhab.core.io.rest.voice/src/main/java/org/openhab/core/io/rest/voice/internal/VoiceResource.java
```diff
@@ -133,27 +137,39 @@ public Response getInterpreter(
...
         }
-
-        try {
-            hli.interpret(locale, text);
-            return Response.ok(null, MediaType.TEXT_PLAIN).build();
-        } catch (InterpretationException e) {
-            return JSONResponse.createErrorResponse(Status.BAD_REQUEST, e.getMessage());
+        String answer = "";
+        String error = null;
+        for (var interpreter : hlis) {
```
J-N-K, line 157:
```suggestion
        for (HumanLanguageInterpreter interpreter : hlis) {
```

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/VoiceManager.java
```diff
@@ -280,6 +283,26 @@ void listenAndAnswer(@Nullable STTService stt, @Nullable TTSService tts, @Nullab
...
 
+    /**
+     * Retrieves a HumanLanguageInterpreter collection.
+     * If no services are available returns null.
+     * 
+     * @param ids Comma separated list of HLI service ids to use
+     * @return a List<HumanLanguageInterpreter> or null, if none of the services is available
+     */
+    @Nullable
+    List<HumanLanguageInterpreter> getHLIsByIds(String ids);
```
J-N-K, line 294: As I said earlier, why do you return `null` and not an empty list? Checking `list.isEmpty()` or `list == null` is the same, but avoids unnecessary null.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/VoiceManager.java
```diff
@@ -280,6 +283,26 @@ void listenAndAnswer(@Nullable STTService stt, @Nullable TTSService tts, @Nullab
...
+
+    /**
+     * Retrieves a HumanLanguageInterpreter collection.
+     * If no services are available returns null.
+     *
+     * @param ids List of HLI service ids to use or null
+     * @return a List<HumanLanguageInterpreter> or null, if none of the services is available
+     */
+    @Nullable
+    List<HumanLanguageInterpreter> getHLIsByIds(List<String> ids);
```
J-N-K, line 304: See above

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
```diff
@@ -282,15 +284,24 @@ public synchronized void sttEventReceived(STTEvent sttEvent) {
...
-                    String answer = hli.interpret(locale, question);
-                    logger.debug("Interpretation result: {}", answer);
-                    say(answer);
-                } catch (InterpretationException e) {
-                    logger.debug("Interpretation exception: {}", e.getMessage());
-                    say(e.getMessage());
+                toggleProcessing(false);
+                String answer = "";
+                String error = null;
+                for (var interpreter : hlis) {
```
J-N-K, line 290:
```suggestion
                for (HumanLanguageInterpreter interpreter : hlis) {
```

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
```diff
@@ -282,15 +284,24 @@ public synchronized void sttEventReceived(STTEvent sttEvent) {
...
+                        answer = interpreter.interpret(locale, question);
+                        logger.debug("Interpretation result: {}", answer);
+                        error = null;
+                        break;
+                    } catch (InterpretationException e) {
+                        logger.debug("Interpretation exception: {}", e.getMessage());
+                        var msg = e.getMessage();
+                        if (msg != null) {
+                            error = Objects.requireNonNullElse(msg, "Unexpected error");
+                        }
```
J-N-K, line 301:
```suggestion
                        error = Objects.requireNonNullElse(e.getMessage(), "Unexpected error");
```

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -288,20 +289,38 @@ public String interpret(String text) throws InterpretationException {
...
 
     @Override
-    public String interpret(String text, @Nullable String hliId) throws InterpretationException {
-        HumanLanguageInterpreter interpreter;
-        if (hliId == null) {
-            interpreter = getHLI();
+    public String interpret(String text, @Nullable String hliIdList) throws InterpretationException {
+        List<HumanLanguageInterpreter> interpreters;
+        if (hliIdList == null) {
+            var interpreter = getHLI();
```
J-N-K, line 295:
```suggestion
            HumanLanguageInterpreter interpreter = getHLI();
```

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -741,6 +772,28 @@ public Collection<KSService> getKSs() {
...
 
+    @Override
+    public @Nullable List<HumanLanguageInterpreter> getHLIsByIds(String ids) {
+        return getHLIsByIds(Arrays.asList(ids.split(",")));
+    }
+
+    @Override
+    public @Nullable List<HumanLanguageInterpreter> getHLIsByIds(List<String> ids) {
+        List<HumanLanguageInterpreter> interpreters = new ArrayList<>();
+        for (var id : ids) {
```
J-N-K, line 783:
```suggestion
        for (String id : ids) {
```

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -741,6 +772,28 @@ public Collection<KSService> getKSs() {
...
+            var hli = humanLanguageInterpreters.get(id);
```
J-N-K, line 784:
```suggestion
            HumanLanguageInterpreter hli = humanLanguageInterpreters.get(id);
```

J-N-K (changes requested):
Please also adapt the documentation: https://github.com/openhab/openhab-docs/blob/main/configuration/multimedia.md

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/VoiceManager.java
```diff
@@ -144,17 +145,18 @@ public interface VoiceManager {
...
      * @param keyword the keyword to use during keyword spotting or null to use the default keyword
      * @param listeningItem the item to switch ON while listening to a question
      * @throws IllegalStateException if required services are not all available or the provided locale is not supported
      *             by all these services or a dialog is already started for this audio source
      */
     void startDialog(@Nullable KSService ks, @Nullable STTService stt, @Nullable TTSService tts,
-            @Nullable HumanLanguageInterpreter hli, @Nullable AudioSource source, @Nullable AudioSink sink,
+            List<HumanLanguageInterpreter> hlis, @Nullable AudioSource source, @Nullable AudioSink sink,
             @Nullable Locale locale, @Nullable String keyword, @Nullable String listeningItem)
             throws IllegalStateException;
```
J-N-K, line 161: As I said above: add to API, do not remove from API. This removes the old method from the API and breaks all code USING these interface. As it is exported to scripts, this is bad. Adding new methods only breaks code IMPLEMENTING this interface. Since I believe it is unlikely that there are other implementations, this is less invasive and preferred, therefor keep the old method:

```
        void startDialog(@Nullable KSService ks, @Nullable STTService stt, @Nullable TTSService tts,
             List<HumanLanguageInterpreter> hlis, @Nullable AudioSource source, @Nullable AudioSink sink,
             @Nullable Locale locale, @Nullable String keyword, @Nullable String listeningItem)
             throws IllegalStateException;

        void startDialog(@Nullable KSService ks, @Nullable STTService stt, @Nullable TTSService tts,
             @Nullable HumanLanguageInterpreter hli, @Nullable AudioSource source, @Nullable AudioSink sink,
             @Nullable Locale locale, @Nullable String keyword, @Nullable String listeningItem)
             throws IllegalStateException;

```

The implementation of the old method is easy, something like

```
        void startDialog(@Nullable KSService ks, @Nullable STTService stt, @Nullable TTSService tts,
             @Nullable HumanLanguageInterpreter hli, @Nullable AudioSource source, @Nullable AudioSink sink,
             @Nullable Locale locale, @Nullable String keyword, @Nullable String listeningItem)
             throws IllegalStateException {
            if (hli == null) {
                startDialog(ks, stt, tts, List.of(), source, sink, locale, keyword, listeningItem); 
            } else {
                startDialog(ks, stt, tts, List.of(hli), source, sink, locale, keyword, listeningItem); 
            }
```

If you think that there is no use-case for the old method, you can mark it as `@Deprecated`. But in this case I believe it is better to keep it.

Applies also below for `listenAndAnswer`.

GiviMAD: Done, sorry for the delay and thank you for the detailed explanation.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/VoiceManager.java
```diff
@@ -144,20 +145,46 @@ public interface VoiceManager {
      * @param ks the keyword spotting service to use or null to use the default service
      * @param stt the speech-to-text service to use or null to use the default service
      * @param tts the text-to-speech service to use or null to use the default service
-     * @param hli the human language text interpreter to use or null to use the default service
+     * @param hli the human language text interpreters to use or null to use the default service
```
lolodomo, line 148: Please keep "interpreter" in the deprecated method.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/VoiceManager.java
```diff
@@ -187,18 +214,41 @@ void startDialog(@Nullable KSService ks, @Nullable STTService stt, @Nullable TTS
      *
      * @param stt the speech-to-text service to use or null to use the default service
      * @param tts the text-to-speech service to use or null to use the default service
-     * @param hli the human language text interpreter to use or null to use the default service
+     * @param hli the human language text interpreters to use or null to use the default service
```
lolodomo, line 217: Please keep "interpreter" in the deprecated method.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/VoiceManager.java
```diff
@@ -280,6 +330,24 @@ void listenAndAnswer(@Nullable STTService stt, @Nullable TTSService tts, @Nullab
      */
     Collection<KSService> getKSs();
 
+    /**
+     * Retrieves a HumanLanguageInterpreter collection.
+     * If no services are available returns null.
```
lolodomo, line 335: returns an empty list

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/VoiceManager.java
```diff
@@ -280,6 +330,24 @@ void listenAndAnswer(@Nullable STTService stt, @Nullable TTSService tts, @Nullab
...
+     * 
+     * @param ids Comma separated list of HLI service ids to use
+     * @return a List<HumanLanguageInterpreter> or empty, if none of the services is available
+     */
+    List<HumanLanguageInterpreter> getHLIsByIds(String ids);
+
+    /**
+     * Retrieves a HumanLanguageInterpreter collection.
+     * If no services are available returns null.
```
lolodomo, line 344: returns an empty list

bundles/org.openhab.core.io.rest.voice/src/main/java/org/openhab/core/io/rest/voice/internal/VoiceResource.java
```diff
@@ -232,7 +247,8 @@ public Response startDialog(
             @QueryParam("ksId") @Parameter(description = "keywork spotter ID") @Nullable String ksId,
             @QueryParam("sttId") @Parameter(description = "Speech-to-Text ID") @Nullable String sttId,
             @QueryParam("ttsId") @Parameter(description = "Text-to-Speech ID") @Nullable String ttsId,
-            @QueryParam("hliId") @Parameter(description = "interpreter ID") @Nullable String hliId,
+            @QueryParam("hliId") @Parameter(description = "interpreter ID", deprecated = true) @Nullable String hliId,
```
lolodomo, line 250: @J-N-K : is it really necessary to keep this parameter and mark it as deprecated while this API is not present in OH 3.2 ? It was recently added by me in 3.3.

bundles/org.openhab.core.io.rest.voice/src/main/java/org/openhab/core/io/rest/voice/internal/VoiceResource.java
```diff
@@ -325,7 +347,8 @@ public Response listenAndAnswer(
             @QueryParam("sourceId") @Parameter(description = "source ID") @Nullable String sourceId,
             @QueryParam("sttId") @Parameter(description = "Speech-to-Text ID") @Nullable String sttId,
             @QueryParam("ttsId") @Parameter(description = "Text-to-Speech ID") @Nullable String ttsId,
-            @QueryParam("hliId") @Parameter(description = "interpreter ID") @Nullable String hliId,
+            @QueryParam("hliId") @Parameter(description = "interpreter ID", deprecated = true) @Nullable String hliId,
```
lolodomo, line 350: @J-N-K : is it really necessary to keep this parameter and mark it as deprecated while this API is not present in OH 3.2 ? It was recently added by me in 3.3.

lolodomo (changes requested):
Final review

lolodomo (approved):
LGTM, thank you @GiviMAD 
The parameters for REST API that have been marked as deprecated could be discussed as these REST API were added very recently and are only available in 3.3 snapshots and 3.3 milestones.

bundles/org.openhab.core.io.rest.voice/src/main/java/org/openhab/core/io/rest/voice/internal/VoiceResource.java
J-N-K: If it was not in a release, it can be removed IMO. But this only applies to the REST parameters, correct?

lolodomo: It also applies to method listenAndAnswer marked as deprecated in interface VoiceManager while it is also a recently added method not present in any release. So I suggest to suppress the one currently marked as deprecated in the PR. 

But your remark about method startDialig was fully correct as this method is there since several years. So deprecating the method is the good option in that case.

lolodomo:
@GiviMAD ; is it clear for you what must be kept as deprecated and what must be removed?

GiviMAD:
> @GiviMAD ; is it clear for you what must be kept as deprecated and what must be removed?

@lolodomo I think so, I'll go for it.

lolodomo:
Perfect IMHO

lolodomo:
Once merged, I will propose a big update in the documentation to add/update all existing actions and console commands about voice management.

J-N-K (approved):
Thanks.

PR #2657, opened 2021-12-30
Implement "forceAsItem" functionality for Charts in sitemaps

Flole998:
This is part of implementing #2185

PR #2910, opened 2022-04-11
Fix hanging build

J-N-K:
CI builds hang forever (also observed locally).

Probably related to #2657


wborn (approved):
Thanks! :+1:

PR #2912, opened 2022-04-13
Adding missing system channels descriptions.

clinique:
and while there, removed some redundant null checks of the same package.

Signed-off-by: clinique <gael@lhopital.org>


kaikreuzer (approved):
lgtm, thanks!

PR #2913, opened 2022-04-13
Add .gradle to .gitignore

J-N-K:
While openHAB uses Maven, under some circumstances IntelliJ creates files in the .gradle/ directory.


wborn (approved):
IntelliJ has its own quirks (just like Eclipse). :neutral_face:

PR #719, opened 2019-04-14
[core] Revert ThingImpl and BridgeImpl changes as the storage uses those implementations

cweitkamp:
- Partially revert #693

Closes #717 - But did not solve the root cause

Signed-off-by: Christoph Weitkamp <github@christophweitkamp.de>

PR #2914, opened 2022-04-15
Use ThingStorageEntity instead of ThingImpl/BridgeImpl for storage

### LINKED ISSUE #731: Use DTOs of Thing interface to store and restore the data

J-N-K:
Fixes #731

#719 should be reverted in another PR.

### DIFF
--- a/bundles/org.openhab.core.storage.json/src/main/java/org/openhab/core/storage/json/internal/JsonStorageService.java
+++ b/bundles/org.openhab.core.storage.json/src/main/java/org/openhab/core/storage/json/internal/JsonStorageService.java
@@ -25,6 +25,8 @@ import org.openhab.core.OpenHAB;
 import org.openhab.core.config.core.ConfigurableService;
 import org.openhab.core.storage.Storage;
 import org.openhab.core.storage.StorageService;
+import org.openhab.core.storage.json.internal.migration.BridgeImplTypeMigrator;
+import org.openhab.core.storage.json.internal.migration.ThingImplTypeMigrator;
 import org.openhab.core.storage.json.internal.migration.TypeMigrator;
 import org.osgi.framework.Constants;
 import org.osgi.service.component.annotations.Activate;
@@ -51,7 +53,8 @@ public class JsonStorageService implements StorageService {
     /**
      * Contains a map of needed migrations, key is the storage name
      */
-    private static final Map<String, List<TypeMigrator>> MIGRATORS = Map.of();
+    private static final Map<String, List<TypeMigrator>> MIGRATORS = Map.of( //
+            "org.openhab.core.thing.Thing", List.of(new BridgeImplTypeMigrator(), new ThingImplTypeMigrator()));
 
     private final Logger logger = LoggerFactory.getLogger(JsonStorageService.class);
 
--- a/bundles/org.openhab.core.storage.json/src/main/java/org/openhab/core/storage/json/internal/migration/ThingImplTypeMigrator.java
+++ b/bundles/org.openhab.core.storage.json/src/main/java/org/openhab/core/storage/json/internal/migration/ThingImplTypeMigrator.java
@@ -1,0 +1,77 @@
+/**
+ * Copyright (c) 2010-2022 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.storage.json.internal.migration;
+
+import java.util.Spliterator;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+
+/**
+ * The {@link ThingImplTypeMigrator} implements a {@link TypeMigrator} for stored things
+ *
+ * @author Jan N. Klug - Initial contribution
+ */
+@NonNullByDefault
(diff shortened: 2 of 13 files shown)


J-N-K:
I tested it with my production database and it worked. You can easily try that yourself: copy your `org.openhab.core.thing.Thing.json` to `resources` in the `org.openhab.core.storage.json.tests` bundle and adjust the number of expected things/bridges in line 59 of `ThingMigrationOSGiTest.java`.

bundles/org.openhab.core.storage.json/src/main/java/org/openhab/core/storage/json/internal/JsonStorageService.java
```diff
@@ -51,7 +53,8 @@ public class JsonStorageService implements StorageService {
     /**
      * Contains a map of needed migrations, key is the storage name
      */
-    private static final Map<String, List<TypeMigrator>> MIGRATORS = Map.of();
+    private static final Map<String, List<TypeMigrator>> MIGRATORS = Map.of( //
+            "org.openhab.core.thing.Thing", List.of(new BridgeImplTypeMigrator(), new ThingImplTypeMigrator()));
```
wborn, line 57: It would be more modular when the migrators are part of the bundles that use the storage (like `org.openhab.core.thing` in this case). That way non-OHC bundles can also migrate their data. :slightly_smiling_face:

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/dto/ThingStorageDTO.java
```diff
@@ -0,0 +1,32 @@
...
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.thing.dto;
+
+/**
+ * The {@link ThingStorageDTO} is a transfer object for Thing storage
```
wborn, line 16: Isn't it a POJO used for mapping stored data? When using JPA such POJOs are called "Entities". It is also very common to keep entities internal so you can more easily change how data is stored without other code breaking.

wborn (commented):
Thanks for working on this! I've added a few comments after having a quick look.

bundles/org.openhab.core.storage.json/src/main/java/org/openhab/core/storage/json/internal/JsonStorageService.java
J-N-K: Yes, but then we would have to draw in a dependency to org.openhab.core.storage.json to every bundle that uses migrations. If we keep the migrators in the storage bundle, other bundles can stay agnostic of the storage and one could use e.g. MapDB instead of JSON storage.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/dto/ThingStorageDTO.java
J-N-K: Good point, I will rename and move to an internal package.

bundles/org.openhab.core.storage.json/src/test/java/org/openhab/core/storage/json/internal/ThingStorageDTOMigratorTest.java
```diff
@@ -0,0 +1,105 @@
...
+import com.google.gson.GsonBuilder;
+import com.google.gson.JsonElement;
+
+/**
+ * The {@link ThingStorageDTOMigratorTest} contains tests for the ThingImpl and BridgeImpl migrators
+ *
+ * @author Jan N. Klug - Initial contribution
+ */
+@NonNullByDefault
+public class ThingStorageDTOMigratorTest {
```
wborn, line 49: This class name now also needs an update. :wink:

bundles/org.openhab.core.storage.json/src/main/java/org/openhab/core/storage/json/internal/migration/ThingImplTypeMigrator.java
```diff
@@ -0,0 +1,77 @@
...
+public class ThingImplTypeMigrator implements TypeMigrator {
+
+    @Override
+    public String getOldType() {
+        return "org.openhab.core.thing.internal.ThingImpl";
+    }
+
+    @Override
+    public String getNewType() {
+        return "org.openhab.core.thing.dto.ThingStorageDTO";
```
wborn, line 39: This also needs an update.

bundles/org.openhab.core.storage.json/src/test/resources/thingMigration-result.json
```diff
@@ -0,0 +1,70 @@
+{
+  "deconz:deconz:00313E041ED0": {
+    "class": "org.openhab.core.thing.dto.ThingStorageDTO",
```
wborn, line 3: Same for these test results.

J-N-K:
@wborn How shall we proceed here?

wborn (approved):
It seems to work very well for me after some testing. :+1:  So let's merge this! :rocket:

PR #2923, opened 2022-04-21
Add new module DateTimeTrigger for triggering on a DateTime state

### LINKED ISSUE #2726: DateTime as rule trigger
I'd like to request a new rule trigger, specifically an item of DateTime type. Rule should fire when the time is equal to the one specified in the item.
My use case is that I often set a timer a fair bit in the future and either experience a restart or something else unexpected that causes the timer to reset. Setting an item that's persisted would both simplify the rule and imho make operation more reliable.

Mikael

J-N-K:
Closes #2726 

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/DateTimeTriggerHandler.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/DateTimeTriggerHandler.java
@@ -1,0 +1,182 @@
+/**
+ * Copyright (c) 2010-2022 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.automation.internal.module.handler;
+
+import java.time.format.DateTimeFormatter;
+import java.util.Dictionary;
+import java.util.Hashtable;
+import java.util.Set;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+import org.openhab.core.automation.ModuleHandlerCallback;
+import org.openhab.core.automation.Trigger;
+import org.openhab.core.automation.handler.BaseTriggerModuleHandler;
+import org.openhab.core.automation.handler.TimeBasedTriggerHandler;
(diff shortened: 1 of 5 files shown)


bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/DateTimeTriggerHandler.java
```diff
@@ -0,0 +1,182 @@
...
+    public static final String MODULE_TYPE_ID = "timer.DateTimeTrigger";
+    public static final String CONFIG_ITEM_NAME = "itemName";
+
+    private static final DateTimeFormatter CRON_FORMATTER = DateTimeFormatter.ofPattern("s m H d M * uuuu");
+
+    private final Logger logger = LoggerFactory.getLogger(DateTimeTriggerHandler.class);
+
+    private final CronScheduler scheduler;
+    private final String itemName;
+    private String cronExpression = "@reboot";
```
kaikreuzer, line 67: This magic string `@reboot` looks a bit weird to me. How about simply setting it to `null`, if no cron expression is set yet? Or if you really prefer to set a magic string in order to avoid null values, please at least make it a constant.

kaikreuzer (changes requested):
Thanks, lgtm. Just one very small comment below.

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/DateTimeTriggerHandler.java
J-N-K: I can make it a constant. I used this string because it is properly handled in the CronAdjuster, so probably we should make it a constant there. WDYT?

kaikreuzer: Ah, I didn't even know that this value was defined in the CronAdjuster. So yeah, it makes a lot of sense to declare it as a constant in there and just use it.

kaikreuzer (approved):
Thanks!

Flole998:
I am thinking about extending this so it can handle a "time only match", something like
`Time is time of myDateTimeItem`
in which case the CronExpression simply needs to be modified to be a `s m H * * * *` instead so it ignores the date. This will be very helpful when trying to implement rules that run daily at a user-configurable time. This option could be named `timeOnly` and for DSL Rules there could be a new `TimeTrigger` that's basically the `DateTimeTrigger` with that `timeOnly` option set.

Also the documentation/helptext for this feature can be misunderstood easily: It mentions that this triggers at a time specified in an item. More accurately it should say something like "at a time and date..." and when my idea is added "at a time and optionally a date...".

PR #2928, opened 2022-04-23
[Semantics] Return parent Location and Equipment

### LINKED ISSUE #2924: How to traverse up the model hierarchy using Semantics Actions
From the documentation:
> An Equipment is normally a Group Item that can contain sub-Equipment, and Points.

There seems to be an issue with [Semantics::getEquipment](https://github.com/openhab/openhab-core/blob/cabb3f73150d13f44d647f46af8362e396684b92/bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/actions/Semantics.java#L98-L113)

1. When an item is a point, getEquipment(Item) would return the (parent) equipment that it belongs to.
2. When an (group)item itself is an equipment, it would return itself
3. But when an equipment is a sub-equipment, getEquipment(Item) would still return itself, not the parent equipment that it belongs to. 

This seems like an inconsistency to me. @cweitkamp could you suggest a way to get the parent equipment of a sub-equipment, or should we modify `Semantics::getEquipment` to always return the parent equipment, and never the given item itself?

I understand that currently I could do this in my script:

```
if Semantics.isEquipment(item) {
  foreach group in item.getGroups() {
    parent_equipment = Semantics.getEquipment(group) // work around it
    if (parent_equipment != null) break 
  }
} else {
  parent_equipment = Semantics.getEquipment(item)
}
```

There's probably a counter argument to this.

My proposal is to change the Semantics::getEquipment() to:

```java
    @ActionDoc(text = "gets the Equipment Item an Item belongs to")
    public static @Nullable Item getEquipment(Item item) {
             return SemanticsActionService.getEquipmentItemFromGroupNames(item.getGroupNames());
    }
```

But this would affect `getEquipmentType()` which should still return the equipment type of the current item, if it's an Equipment. So getEquipmentType() would change to:
```java
    @ActionDoc(text = "gets the Equipment type an Item belongs to")
    public static @Nullable Class<? extends Equipment> getEquipmentType(Item item) {
        Item equipmentItem = isEquipment(item) ? item : getEquipment(item);
        return equipmentItem != null ? SemanticTags.getEquipment(equipmentItem) : null;
    }
```

With this proposed change, my script above to get the parent item would simply become:

```
parent_equipment = Semantics.getEquipment(item)
```

This seems to also apply to `getLocation()` - how can I get a location's parent location easily?

jimtng:
Resolve #2924 

Instead of returning the given item, getEquipment() and getLocation() return the item's parent Equipment / Location. This allows the caller to navigate the hierarchy, e.g. get the parent location of the given location.

### DIFF
--- a/bundles/org.openhab.core.model.script/src.moved/test/java/org/openhab/core/model/script/actions/SemanticsTest.java
+++ b/bundles/org.openhab.core.model.script/src.moved/test/java/org/openhab/core/model/script/actions/SemanticsTest.java
@@ -31,6 +31,7 @@ import org.openhab.core.items.ItemRegistry;
 import org.openhab.core.library.CoreItemFactory;
 import org.openhab.core.model.script.internal.engine.action.SemanticsActionService;
 import org.openhab.core.semantics.model.equipment.CleaningRobot;
+import org.openhab.core.semantics.model.equipment.Battery;
 import org.openhab.core.semantics.model.location.Bathroom;
 import org.openhab.core.semantics.model.location.Indoor;
 
@@ -50,6 +51,7 @@ public class SemanticsTest {
     private GroupItem equipmentItem;
     private GenericItem temperaturePointItem;
     private GenericItem humidityPointItem;
+    private GenericItem subEquipmentItem;
 
     @BeforeEach
     public void setup() throws ItemNotFoundException {
@@ -84,6 +86,13 @@ public class SemanticsTest {
         humidityPointItem.addTag("Measurement");
         humidityPointItem.addTag("Humidity");
 
+        subEquipmentItem = itemFactory.createItem(CoreItemFactory.NUMBER, "TestBattery");
+        subEquipmentItem.addTag("Battery");
+
+        // Equipment (TestBattery) is a part of Equipment (Cleaning Robot)
+        equipmentItem.addMember(subEquipmentItem);
+        subEquipmentItem.addGroupName(equipmentItem.getName());
+
         when(mockedItemRegistry.getItem("TestHouse")).thenReturn(indoorLocationItem);
         when(mockedItemRegistry.getItem("TestBathRoom")).thenReturn(bathroomLocationItem);
         when(mockedItemRegistry.getItem("Test08")).thenReturn(equipmentItem);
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.model.script/src.moved/test/java/org/openhab/core/model/script/actions/SemanticsTest.java
```diff
@@ -95,8 +104,8 @@ public void setup() throws ItemNotFoundException {
 
     @Test
     public void testGetLocation() {
-        assertThat(Semantics.getLocation(indoorLocationItem), is(indoorLocationItem));
-        assertThat(Semantics.getLocation(bathroomLocationItem), is(bathroomLocationItem));
+        assertNull(Semantics.getLocation(indoorLocationItem));
```
jlaur, line 107: For consistency with the constraint-based Assert model:
```suggestion
        assertThat(Semantics.getLocation(indoorLocationItem), is(nullValue()));
```

bundles/org.openhab.core.model.script/src.moved/test/java/org/openhab/core/model/script/actions/SemanticsTest.java
```diff
@@ -115,7 +124,8 @@ public void testGetLocationType() {
 
     @Test
     public void testGetEquipment() {
-        assertThat(Semantics.getEquipment(equipmentItem), is(equipmentItem));
+        assertNull(Semantics.getEquipment(equipmentItem));
```
jlaur, line 127:
```suggestion
        assertThat(Semantics.getEquipment(equipmentItem), is(nullValue()));
```

jlaur (changes requested):
Nice! Two minor comments for the tests.

jimtng:
@jlaur thanks for your review. I have changed the assertion as suggested. I'm not sure about the spacing in-between lines and why they are that way, but I added spacing to make them all spaced the same way.

wborn (approved):
Thanks! It looks good to me, but it would be nice if @cweitkamp can also have a quick look.

cweitkamp (approved):
LGTM. Thanks.

It sounds reasonable to change the logic.

PR #2932, opened 2022-04-24
Properly close script engines in script transformation

J-N-K:
Follow-up to #2883 


wborn (approved):
Thanks! I think something similar needs to be done in #2872?

J-N-K:
I think I'll refactor the profile to use the transformation, so that should be covered as well and we can make use of the latest additions like parameters without duplicating code.

PR #2933, opened 2022-04-25
Bump jackson to 2.12.6(.1)

J-N-K:
The `jackson-databind` bundle contains a vulnerability that is fixed in version 2.12.6.1. This version is only available for that bundle, all other bundles are still available as 2.12.6. This requires a manual revert for the `jackson-databind` bundle. I have added a note as comment. 

See #2882


wborn (approved):
Thanks!

PR #2939, opened 2022-04-30
[test] Allow intercepting logs and make assertions for log messages

J-N-K:
Signed-off-by: Jan N. Klug <github@klug.nrw>

PR #2940, opened 2022-04-30
Refactor thing update handling of ThingManagerImpl

### LINKED ISSUE #1463: ThingManagerImpl#thingUpdated never gets hold of the oldElement
While hunting a strange bug I stumpled accros an even stranger bug: 

https://github.com/openhab/openhab-core/blob/aca4513601194485e53a028b2a47b9f304ddd8e9/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java#L517

is being called as a `RegistryTracker` of the thing registry. It only receives the new element of the registry and tries to construct the previous, old element by calling getThing here:

https://github.com/openhab/openhab-core/blob/aca4513601194485e53a028b2a47b9f304ddd8e9/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java#L581

Besides it's bad implementation of searching by iterating through a list instead of using a map it is just not working! We could even completely remove this internal list and ask the registry directly because the ThingRegistry just returns a reference of the thing, which is of course the same reference which is stored here. So it ends up with `oldElement == element` in https://github.com/openhab/openhab-core/blob/aca4513601194485e53a028b2a47b9f304ddd8e9/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java#L528

For example consider this: 

1. getting an object with (id=1) from the registry
2. update it 
3. Call `thingRegsitry.update`
4. ManagedThingProvider will store the new element (id=1) and return the old element from storage as a `PersistedElement`and construct the old thing out of it (id=2) 
5. Calls `notifyListeners(oldElement, element)`
6. the listener calls `thingRegistry#updated(oldElement, element)`
7. this will call `notifyListeneryAbotuUpdatedElement(oldElement, element)` 
8. which finally calls `notifyTrackers(element)` -- now the oldElement is getting lost, the tracker does not have it
9. `ThingManagerImpl#thingUpdated` is the tracker and reads the oldElement from its shadow storage as explained above which does not work

This will never work as expected. I'd suggest to add the oldElement to the tracker API and to completely remove the internal list of things of the `ThingManager`.

J-N-K:
Fixes #1463 
Depends on #2939 (for tests)

Also code clean-up.

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
@@ -132,13 +132,13 @@ public class ThingManagerImpl
     static final String XML_THING_TYPE = "openhab.xmlThingTypes";
 
     private static final String THING_STATUS_STORAGE_NAME = "thing_status_storage";
-    private static final String FORCEREMOVE_THREADPOOL_NAME = "forceRemove";
-    private static final String THING_MANAGER_THREADPOOL_NAME = "thingManager";
+    private static final String FORCE_REMOVE_THREAD_POOL_NAME = "forceRemove";
+    private static final String THING_MANAGER_THREAD_POOL_NAME = "thingManager";
 
     private final Logger logger = LoggerFactory.getLogger(ThingManagerImpl.class);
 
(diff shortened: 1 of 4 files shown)


wborn (approved):
LGTM and thanks for the cleanup!

PR #2941, opened 2022-04-30
Fix item name checking in ItemUtil

### LINKED ISSUE #2243: Clarification on valid Item naming needed
Hi everyone,

I would like to get some generic clarification for the valid usage of Numbers in  Item names.
Reason for this is an issue that is currently discussed in the documentation repository.
(Reference: openhab/openhab-docs#1481)

According to the issue item names are currently just validated for allowed characters and numbers, but not against starting with numbers. 
(Reference: [ItemUtil.java#L56](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core/src/main/java/org/openhab/core/items/ItemUtil.java#L56))

Documentation states for a long time that item names must not begin with a number.
So there is a difference between the docs and the specific core implementation.

According to a newer comment we have a misbehavior when using such items in rules dsl.
(Reference: https://github.com/openhab/openhab-docs/issues/1481#issuecomment-794594645) 

So we should clarify:

- What's the correct limitation
- Fix either docs or core code components involved

To get a long term clarification.

J-N-K:
Fixes #2243 

Since numbers as first character of the item name cause issues and the documentation already states that they are not allowed, the validity check is adjusted to reflect the correct policy.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/items/ItemUtil.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/items/ItemUtil.java
@@ -54,7 +54,7 @@ public class ItemUtil {
      * @return true if the specified name is a valid item name, otherwise false
      */
     public static boolean isValidItemName(final @Nullable String itemName) {
-        return itemName != null && !itemName.isEmpty() && itemName.matches("[a-zA-Z0-9_]*");
+        return itemName != null && !itemName.isEmpty() && itemName.matches("[a-zA-Z][a-zA-Z0-9_]*");
     }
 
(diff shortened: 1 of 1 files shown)


lolodomo:
You also reject _ as first character. 
Would it be a potential breaking change for users?

lolodomo:
My question was rather: what will happen if users defined items starting with a number ?
The system will no more consider them ?

J-N-K:
Yes, a warning will be logged when a provider tries to add that to the item registry.

kaikreuzer (approved):
lgtm, thanks!

PR #2945, opened 2022-05-01
Fix transformations not removed if file deleted

J-N-K:
The `isRegularFile` check fails for `DELETE` events. As a result deleted files where not removed from the registry.


kaikreuzer (approved):
lgtm, thanks!

PR #2946, opened 2022-05-01
Allow sub-directories for icons

### LINKED ISSUE #2896: Subfolders for custom icons
Right now it's not possible to have icons arranged in subfolders.
Personally I think this would be a good idea, I have custom icons for many items, not to mention the 100 I am designing for weather states.
I understand that naming icons 'kinda' groups them together, but having icon sets in their own folders would tidy up everyone's filesystem.

J-N-K:
Closes #2896 

This immediately works in Main UI, but needs an adjustment in Basic UI.

### DIFF
--- a/bundles/org.openhab.core.ui.icon/src/main/java/org/openhab/core/ui/icon/internal/CustomIconProvider.java
+++ b/bundles/org.openhab.core.ui.icon/src/main/java/org/openhab/core/ui/icon/internal/CustomIconProvider.java
@@ -12,10 +12,10 @@
  */
 package org.openhab.core.ui.icon.internal;
 
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileNotFoundException;
+import java.io.IOException;
 import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.Collections;
 import java.util.Locale;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


lolodomo (approved):
LGTM

kaikreuzer:
If I understand the code right, this now also requires the http request to contain the folder names next to the icon name (i.e. the relative path). My understanding of the issue was that it is desired to be able to structure icons into subfolders in the file system, but that they should all still be served the very same way to the outside - i.e. hiding their storage path. Wdyt?

J-N-K:
Yes, you have to include the folder. After re-reading the issue I beliebe that both interpretations (supporting sub-folders and present them with or without folder-name) are possible. IMO the solution here is better, because we will run into unpredictable behavior if an icon with the same name in different folders (which is perfectly possible in a file system). Taking first/last discovered item would be non-deterministic since it depends on the order in which the folders are processed.

kaikreuzer:
I'm not feeling well about that change as it changes the semantic of how icons are handled throughout openHAB. The "category" used to be a plain string so far - now it would be a path. Note that icon categories are used at many places: In the general `IconProvider` interface, within DSL item definitions, in sitemaps, etc. I don't see the benefit that would justify such a huge change.
Wrt duplicate files in the folder structure, we could probably define a simple rule like saying the first one found when traversing the tree is chosen.

J-N-K:
I'm fine with that, too. I'll check how to implement it.

kaikreuzer:
Thank you!

kaikreuzer (approved):
Thanks, I like this much more and it also looks pretty straight forward.

PR #2947, opened 2022-05-02
Fix issues with UI Component Registries

J-N-K:
Follow-Up to #2617 

See https://community.openhab.org/t/openhab-3-3-milestone-discussion/132715/98
Solution provided by @splatch in the same thread.

This is required to allow UI bundles to extend the UIComponentProvider for storing configurations server-side.

There were some more issues, e.g. a managed provider not available at the time of registry creation was added as provider bot not set as managed provider. I also moved the namespace-specific providers to the respective bundles.


splatch:
I was not certain if extracted `UIComponentProvider` qualifies to become part of a public package. Given that it holds some logic its better to open access to it. Note that relying on an base class might have side effects under certain scenarios. For example when ui.component package is updated you might need to manually refresh habpanel provider. I bet later will affect far more people than generic failure in habpanel. ;-)

J-N-K:
Since e.g. `ManagedThingProvider` and `ManagedItemProvider` are also in a public package I guess it's ok. If the changes are API breaking, this will be quite obvious because the openhab-webui nightly build will fail.

J-N-K:
@openhab/core-maintainers This needs to be merged together with https://github.com/openhab/openhab-webui/pull/1384 or we'll run into even more trouble than we have now.

kaikreuzer:
Just for my understanding:
Usually, the pattern is that we have a single managed provider per registry, which lives in the core and which manages writable elements through a StorageService.
For UI components, we seem to divert from this pattern: We have namespaces and consider every namespace to be specific to an add-on (UI), so that the add-on itself is responsible for providing the managed provider. This also means that we will never ever share a namespace between multiple add-ons.
Did I get this right?

J-N-K:
Yes. What I propose to wait for (I need a day or two to get it right) is essentially what we do with the other registries, so we can manage all providers in core and stay agnostic of the required namespaces (that is why I moved the providers out of the core bundle, because we would need to adapt core and add a new provider for each namespace, which is wrong IMO).

The drawback is that the blockly marketplace and habpanel are broken until I'm done with it. 

What is also possible: add the habpanel and blockly namespaces to core for the time being and implement the provider factory in a second step. That would be the fastest way of getting back to a usable state.

PR #2948, opened 2022-05-03
Add support for automatic creation of ManagedProviders for UI components

J-N-K:
Supersedes #2947
Follow-Up to https://github.com/openhab/openhab-core/pull/2617

See https://community.openhab.org/t/openhab-3-3-milestone-discussion/132715/98

Every registry needs a managed provider for storing configurations on server-side.


J-N-K:
@kaikreuzer I guess this is more in line with what we do with the other registries and does not require a component for each namespace.

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentProvider.java
```diff
@@ -15,34 +15,41 @@
...
 /**
  * A namespace-specific {@link ManagedProvider} for UI components.
  *
  * @author Yannick Schaus - Initial contribution
+ * @author Jan N. Klug - Refactored to component factory
  */
 @NonNullByDefault
+@Component(factory = "org.openhab.core.ui.component.provider.factory")
 public class UIComponentProvider extends AbstractProvider<RootUIComponent>
         implements ManagedProvider<RootUIComponent, String>, UIProvider {
```
ghys, line 42: Probably late to the "party" but I think it would be more logical to call this `UIProvider` interface `UIComponentProvider` (as it provides UICompnents, not UIs) and this `UIComponentProvider` class something like `UIComponentProviderImpl`.

ghys (commented):
IMHO this should be probably be merged ASAP if it effectively fixes the regressions of the incomplete refactoring that unfortunately made it to M4 and broke several important features.

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentProvider.java
J-N-K: Good point. Since this was internal only, we can change it without issues.

ghys: > this UIComponentProvider class something like UIComponentProviderImpl.

Or maybe even `ManagedUIComponentProvider[Impl]` since the pattern seems to be `ManagedThingProvider implements ThingProvider`, `ManagedItemProvider implements ItemProvider` etc.

ghys:
Another remark: to be complete, other objects allowing both managed and non-managed providers have a `editable` flag in their REST API schema in order to give clues to UIs which ones can be saved using the API and which ones are effectively read-only, so they can adapt the display (🔒 icons & hidden functionality).
This has also been omitted in the recent refactoring so UIs can't make these adaptations for UI components that cannot be modified or deleted since they're not provided from the ManagedProvider.

FWIW that's part of the reason there wasn't any support for non-managed Providers in the initial implementation (I'm not questioning the added value of having them): there wasn't an initial need for them since they were supposed to be managed by WYSIWYG UI editors only.

kaikreuzer (approved):
Great solution, many thanks!

kaikreuzer:
@J-N-K & @ghys Do you consider this a critical fix, which would warrant doing an M5 release out-of-schedule the next days?

J-N-K:
Probably, since it affects quite a few people. I would suggest to merge #2945  (not critical, but easy to review and potentially irritating behavior) and possibly #2949 (reported on the forum) before that.

ghys:
I agree there's a relatively strong case for that. In M4 users can't see or save:
1) UI-designed sitemaps
2) HABPanel configuration
3) Block libraries (for Blockly rules).

kaikreuzer:
Ok, I'll suggest we test a bit tomorrow and I'll see to do a new milestone build on Saturday then.

PR #2949, opened 2022-05-05
Fix EphemerisManager crashing on invalid configuration

J-N-K:
It has been reported in the past (and in the forum) that the EphemerisManagerImpl can't handle illegal configurations. Due to dependencies in other bundles this results in the whole automation component to be unavailable.

This PR adds code to strip illegal characters (which will solve most reported issues with additional double quotes) and just skips a days configuration with unknown day names (instead of failing to activate/modify the service in total).


kaikreuzer (approved):
Thanks!

PR #2951, opened 2022-05-07
Fix MQTT status message when disconnected while connecting

### LINKED ISSUE #1202: [MQTT] Receiving CONNECTED state change when the MQTT broker is shutdown
I have a MQTT client (binding) connected to a MQTT broker (Snips running on a RPI).
The connection is established and I can see states changes through the method connectionStateChanged, CONNECTING and then CONNECTED.

If I shutdown my RPI, the binding receioves a state change with state CONNECTED and error "Server closed connection without DISCONNECT.".
I think the state should be DISCONNECTED in this case, not CONNECTED.

J-N-K:
Fixes #1202


wborn (approved):
LGTM! 👍

PR #2954, opened 2022-05-08
Expire ignore commands

### LINKED ISSUE #2952: Enhancement: Expire to optionally ignore commands
This is a feature request / enhancement to expire functionality, as introduced in https://github.com/openhab/openhab-core/pull/1803

I would prefer to have additional optional syntax to ignore e.g. commands when determining whether to expire item.

My use case is as follows
- item is linked to two channels, one for state updates (measurement), one for writing the data to another system (commands)
- I have rule that triggers regularly, commanding the item to write the latest measurement to the another system

With current implementation of the expire, the regular commands naturally prevent the item expiration. 

Currently I am working around this with bunch of rules and expire-like metadata.

In this feature request, I propose to have additional syntax in expire to ignore command updates, allowing the  item to expire in case of regular commands like in my use case.

If this is something that is considered suitable, I can contribute PR to introduce the feature.

ssalonen:
Resolves #2952

This needs an additional PR in UI to make it configurable via UI. Similar to `ignoreStateUpdates` parameter support in `openhab-webui` https://github.com/openhab/openhab-webui/pull/1325

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ExpireManager.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ExpireManager.java
@@ -330,14 +330,7 @@ public class ExpireManager implements EventSubscriber, RegistryChangeListener<It
                     ? configString.substring(commaPos + 1).trim()
                     : null;
 
-            Object ignoreStateUpdatesConfigObject = configuration.get(CONFIG_IGNORE_STATE_UPDATES);
-            if (ignoreStateUpdatesConfigObject instanceof String) {
-                ignoreStateUpdates = Boolean.parseBoolean((String) ignoreStateUpdatesConfigObject);
-            } else if (ignoreStateUpdatesConfigObject instanceof Boolean) {
-                ignoreStateUpdates = (Boolean) ignoreStateUpdatesConfigObject;
-            } else {
-                ignoreStateUpdates = false;
-            }
+            ignoreStateUpdates = getBooleanConfigValue(configuration, CONFIG_IGNORE_STATE_UPDATES);
 
             if ((stateOrCommand != null) && (stateOrCommand.length() > 0)) {
                 if (stateOrCommand.startsWith(COMMAND_PREFIX)) {
@@ -373,6 +366,27 @@ public class ExpireManager implements EventSubscriber, RegistryChangeListener<It
                 expireCommand = null;
                 expireState = UnDefType.UNDEF;
             }
+        }
+
+        /**
+         * Parse configuration value as primitive boolean. Supports parsing of String and Boolean values.
+         *
+         * @param configuration map of configuration keys and values
+         * @param configKey configuration key to lookup configuration map
+         * @return configuration value parsed as boolean. Defaults to false when configKey is not present in
+         *         configuration
+         */
+        private boolean getBooleanConfigValue(Map<String, Object> configuration, String configKey) {
+            boolean configValue;
+            Object configValueObject = configuration.get(configKey);
+            if (configValueObject instanceof String) {
+                configValue = Boolean.parseBoolean((String) configValueObject);
+            } else if (configValueObject instanceof Boolean) {
+                configValue = (Boolean) configValueObject;
+            } else {
+                configValue = false;
+            }
+            return configValue;
         }
 
         private Duration parseDuration(String durationString) throws IllegalArgumentException {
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ExpireManager.java
```diff
@@ -330,14 +334,8 @@ public ExpireConfig(Item item, String configString, Map<String, Object> configur
...
 
-            Object ignoreStateUpdatesConfigObject = configuration.get(CONFIG_IGNORE_STATE_UPDATES);
-            if (ignoreStateUpdatesConfigObject instanceof String) {
-                ignoreStateUpdates = Boolean.parseBoolean((String) ignoreStateUpdatesConfigObject);
-            } else if (ignoreStateUpdatesConfigObject instanceof Boolean) {
-                ignoreStateUpdates = (Boolean) ignoreStateUpdatesConfigObject;
-            } else {
-                ignoreStateUpdates = false;
-            }
+            ignoreStateUpdates = getBooleanConfigValue(configuration, CONFIG_IGNORE_STATE_UPDATES);
```
J-N-K, line 337:
```suggestion
            ignoreStateUpdates = ConfigParser.valueAsOrElse(configuration.get( CONFIG_IGNORE_STATE_UPDATES), Boolean.class, false);
```

We now have a core-method for that.

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ExpireManager.java
```diff
@@ -244,13 +246,13 @@ public void receive(Event event) {
...
+            processEvent(isEvent.getItemName(), isEvent.getItemState(), EventType.StateUpdate);
         } else if (event instanceof ItemCommandEvent) {
             ItemCommandEvent icEvent = (ItemCommandEvent) event;
-            processEvent(icEvent.getItemName(), icEvent.getItemCommand(), false);
+            processEvent(icEvent.getItemName(), icEvent.getItemCommand(), EventType.Command);
         } else if (event instanceof ItemStateChangedEvent) {
             ItemStateChangedEvent icEvent = (ItemStateChangedEvent) event;
-            processEvent(icEvent.getItemName(), icEvent.getItemState(), false);
+            processEvent(icEvent.getItemName(), icEvent.getItemState(), EventType.StateChanged);
         }
```
J-N-K, line 256: Wouldn't it make sense to check for `ItemEvent` only and move this logic to `processEvent`? You could do something like
 
```
String itemName;
Type value;

if (event instance ItemStateEvent && !expireConfig.ignoreStateUpdates) {
    itemName = event.getName();
    type = ((ItemStateEvent) event).getItemState;
} else if (...) {
    ...
} else {
    return;
} 
```

The new enum could then be omitted. Looks cleaner to me.

J-N-K (changes requested):

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ExpireManager.java
ssalonen: `ConfigParser` is defined in `org.openhab.core.config.core`. Is it possible to add it as dependency to `org.openhab.core` (circular dependency)?

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ExpireManager.java
ssalonen: Essentially you are suggesting to merge `processEvent` and `receive`? Currently `processEvent` does not have access to the `event` object.

J-N-K: Yes, currently the code just duplicates the check for the item type (one time in `receive` and one time in `processEvent`.

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ExpireManager.java
J-N-K: No, unfortunately not. Keep your method then.

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ExpireManager.java
ssalonen: Makes sense. See https://github.com/openhab/openhab-core/pull/2954/commits/fa9493626981d37a3eb7af6c6bac7782dfc1c2f4

J-N-K (approved):
Thanks, LGTM.

PR #2955, opened 2022-05-08
Fix NPE in ItemUIRegistry

J-N-K:
Reported on the forum. Also some code-cleanup.


wborn (approved):
Thanks for the fix and cleanup!

PR #2958, opened 2022-05-11
Normalize Metadata configuration from managed provider

### LINKED ISSUE #1904: metadata difference between mainUI and .items
following item created via .items text configuration
`
Number temp_sensor {homekit="TemperatureSensor" [minValue=13, maxValue=43] }
`
result in meta data minValue and maxValue with values of type BigDecimal.

the same item created via mainUI results in values of type Double

J-N-K:
Fixes #1904 

The reason is that GSON de-serializes numeric metadata configuration values from JSON storage to `Double`, while textual defined values are `BigDecimal`. This PR adds normalization of the entries provided by the managed provider.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ManagedMetadataProviderImpl.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ManagedMetadataProviderImpl.java
@@ -102,9 +102,15 @@ public class ManagedMetadataProviderImpl extends AbstractManagedProvider<Metadat
     }
 
     private Map.Entry<String, Object> normalizeConfigEntry(Map.Entry<String, Object> entry) {
-        if (entry.getValue() instanceof Number) {
-            return Map.entry(entry.getKey(), new BigDecimal(entry.getValue().toString()));
+        Object value = entry.getValue();
+        if (value instanceof Integer) {
+            BigDecimal newValue = new BigDecimal(value.toString());
+            newValue.setScale(0);
+            return Map.entry(entry.getKey(), newValue);
+        } else if (value instanceof Number) {
+            return Map.entry(entry.getKey(), new BigDecimal(value.toString()));
         }
+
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ManagedMetadataProviderImpl.java
```diff
@@ -76,4 +81,30 @@ public void removeItemMetadata(String name) {
...
+    }
+
+    private Metadata normalizeMetadata(Metadata metadata) {
+        return new Metadata(metadata.getUID(), metadata.getValue(), metadata.getConfiguration().entrySet().stream()
+                .map(this::normalizeConfigEntry).collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue)));
+    }
+
+    private Map.Entry<String, Object> normalizeConfigEntry(Map.Entry<String, Object> entry) {
+        if (entry.getValue() instanceof Number) {
+            return Map.entry(entry.getKey(), new BigDecimal(entry.getValue().toString()));
```
kaikreuzer, line 106: I'm not sure if it is relevant, but in case the number is an int, we [here](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/normalization/IntNormalizer.java#L51) also set the scale to 0 to make sure that it is not considered a floating point value. Wdyt, would this make sense here as well?

J-N-K: I'm not sure if it is necessary, but it doesn't hurt. Implemented.

J-N-K:
@kaikreuzer Do you think we can merge here, too?

kaikreuzer (approved):
Sure thing. Sorry that this was getting out of my sight.

PR #2959, opened 2022-05-13
Fix resource leaks in AudioServlet

### LINKED ISSUE #532: resource leak: audio manager

J-N-K:
Fixes #532

Add closing of streams on deactivation and add tests to ensure that all streams are closed after they are expired or retrieved.

There might be still one possible resource leak: a one-time stream that is served but never retrieved will stay in the map forever. But since that is exactly the wanted behavior for streams without timeout, there is nothing we can do about it. Those streams are only closed if the servlet is deactivated.


lolodomo:
Code looks good to me.

kaikreuzer (approved):
Thank you!

PR #2960, opened 2022-05-13
Add processing of Accept header to AudioServlet

### LINKED ISSUE #1350: Discussion: AudioServlet and wav files
I have just debugged why the PicoTTS service is not working with my Squeezebox installation.

I am running the Picoreplayer project with Logitech Media Server and squeezelite instances.

PicoTTS produces ".wav" files and openHAB serves them on its webserver. Browsers like Chromium are able to play it from there. 

The squeezelite instance however, got the URL in its playlist correctly, but refuses to play it.

First I thought that my LMS or squeezelite has problems with wave files in a certain format, like earlier versions of these programs did, but that is not the case for me. I could add the file downloaded from openhab on the harddisk of the picoreplayer and play it from there without issues.

Then I setup an apache webserver on my development machine and served the wav file from there. Result was that the file could be played by squeezelite without issues!

Wireshark helped me to find the difference between apache and the openHAB audio servlet. It is right here:

https://github.com/openhab/openhab-core/blob/e63db5f5c56bdb370707a395b78ee4d7b215a9d3/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/AudioServlet.java#L100

Apache uses the type `audio/x-wav` instead.

For testing purposes I have edited the openHAB servlet to serve the wav files as `audio/x-wav` too. This is working perfectly fine. Squeezelite plays them without any issues!

by searching the internet I found that this isn't a new issue:
https://github.com/jshttp/mime-db/issues/22

However, their recommendation `audio/vnd.wave` from http://tools.ietf.org/html/rfc2361 doesn't work for the squeezebox case.

Long story short: This leaves me to the question whether we could simply replace the `audio/wav` by `audio/x-wav` or if we have to enhance the code so we can somehow dynamically decide which mimetype to use.

Any suggestions?

AFAIK picoTTS is the **only** TTS service supported by openHAB that does NOT need a cloud connection, that is why I really want to use it!

J-N-K:
Closes #1350

This checks the Accept header of the incoming request and adjusts the content-type (if possible). This is the best we can do to support clients that need a non-standard MIME type.


lolodomo:
Code looks good to me.

What could have be done too is to add an additional `serve` method with a new parameter to provide the MINE type to consider when the audio servlet is returning data. Like that, the squeezebox binding in its audio sink could have set it here:
https://github.com/openhab/openhab-addons/blob/da59cdd255a66275dd7ae11dd294fedca4942d30/bundles/org.openhab.binding.squeezebox/src/main/java/org/openhab/binding/squeezebox/internal/SqueezeBoxAudioSink.java#L96
WDYT ?

J-N-K:
Why would one want to do that? The possible MIME types can be determined from `Accept` header (client side) and the `stream.getFormat` (server-side). Maybe a mis-match between those two (i.e. the determined MIME type is not present in the `Accept`-header should result in a 406 error.

lolodomo:
You're probably right.
I was thinking about the strange case where the client was not providing any Accept header. I don't know if the software of the Squeezebox is providing one.

kaikreuzer (approved):
lgtm, thanks!

PR #2961, opened 2022-05-13
Fix console extension output to wrong console

### LINKED ISSUE #1545: OSGiConsole outputs to where thread was started, not in current console window
The class `org.eclipse.smarthome.io.console.karaf.internal.OSGiConsole` prints to the output stream associated with the thread using `System.out.print*` methods. This can cause the results from console commands to potentially get displayed elsewhere. I have encountered this while trying to implement support to configure and monitor Insteon devices. An example of what is happening is:

- user create a ssh session to OH
- user send message to device from ssh session
- device responds and is handled by OH daemon thread
- output from console.println() appears where start.sh was ran, not ssh session

Ideally the output would show up in the console that sent the message. The workaround is to write stdout to a log file and monitor the log file while sending commands to a device. For example I use the following to stat up OH:

` /opt/openhab/start.sh server >/opt/openhab/userdata/logs/console.log 2>&1 &`

J-N-K:
Fixes #1545

The `PrintStream` used for printing should not be `System.out` but the `PrintStream` associated with the `Session` of the karaf shell.


kaikreuzer (approved):
Nice fix, thanks!

PR #2962, opened 2022-05-14
[rest] allow retrieving metadata for item

### LINKED ISSUE #1455: [REST API] Missing method for metadata namespaces
Currently it is only possible to set and delete metadata of an item when the namespace is known.
However the endpoint to read the available namespaces or an option to get all metadata regardless of namespaces is missing.

J-N-K:
Closes #1455

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
@@ -569,8 +569,8 @@ public class ItemResource implements RESTResource {
             return Response.status(Status.NOT_FOUND).build();
         }
 
-        Map<String, Object> metadata = metadataRegistry.stream().filter(metadataFilter).map(this::mapMetadata)
-                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+        Map<String, Object> metadata = metadataRegistry.stream().filter(metadataFilter)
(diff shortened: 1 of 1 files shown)


spacemanspiff2007:
Would it be possible to add a switch like that to the ``/items`` endpoint?
Currently I have to make every request with ``metadata=.+``

J-N-K:
This would change the way the `/items` endpoint works, since currently "no selector" means "no metadata". IMO maintaining backward compatibility is more important here than the burden of adding `metadata=.+`.

spacemanspiff2007:
I am not suggesting changing the behavior of the ``/items`` endpoint but rather making an addition so it's easier to request all metadata. Maybe a Parameter ``allMetadata`` for which the default is False which can be set to True.

J-N-K:
But what is the benefit of adding `allMetadata=true` instead of `metadata=.+`?

spacemanspiff2007:
If you look through the issues there has been some confusion by other users on how to get all metadata.
That would make things clearer.

J-N-K:
IMO we can expect that someone who uses the REST API reads the documentation. Maybe the wording can be improved, although I don't have a good idea how to do so.

spacemanspiff2007:
It was just an idea based on my observations. If I figured it out so can others.

---

Back to the PR:
The original request was to get a list of available name spaces over all items so they can be requested when reading the items because the metadata has be explicitly known.
In the mean time the ability to request metadata based on an regex has been added.
So imho we don't need this change since `metadata=.+` achieves the same.

spacemanspiff2007:
Sorry that I had not closed the issue and you invested time in fixing it.
I'll look into my old oh issues and try to double check the ones that I created for OH2

PR #2963, opened 2022-05-14
Add DateTimeTrigger to DSL rule model

J-N-K:
Follow-Up to #2726 

The new trigger can't be used in DSL rules. This adds the missing parts in the model:

```
rule "foo"
when
    Time equals MyDateTimeItem
then
    ...
end
```


kaikreuzer (approved):
lgtm, thanks.

sasbibic:
In version Build #2947 appears error:
[320,1]: no viable alternative at input 'equals'

J-N-K:
Can you show the full file? Thanks.

sasbibic:
I think I was a bit too quick, there was change in syntax?

From:
Time equals <DateTimeItem>

to:
Time is <DateTimeItem>

Is This correct ?

Regards
/Sas

J-N-K:
Exactly. The old syntax cause the same error you see, but in all places in rules where .equals is used. Since this was the latest addition, the syntax was changed here. The documentation should reflect that change.

sasbibic:
Great, thanks!

PR #2964, opened 2022-05-14
Fix spaces in options not properly handled in i18n-maven-plugin

### LINKED ISSUE #2636: [i18n:generate-default-translations] keys in options list with spaces are incorrectly generated
The `i18n:generate-default-translations` maven plugin to generate properties files from xml files incorrectly handles keys with spaces. This can occur in options lists, where spaces are allowed in keys. It should escape the spaces in the key with a backslash, see the [properties documentation](https://docs.oracle.com/javase/6/docs/api/java/util/Properties.html#load(java.io.Reader)).

Because this plugin was used to generate most of the properties files for the addons, some addons are now broken. i.e. They show the wrong key values. Addons known to be affected are: `networkupstools`, `roku`, `wlanthermo`, and yamahareceiver` . For the  networkupstools this is already been reported, as a pr https://github.com/openhab/openhab-addons/pull/11828, and on the forum: https://community.openhab.org/t/networkupstools-new-ups-status-in-version-3-2-0/130397

J-N-K:
Fixes #2636 

Spaces in option values are now properly escaped


kaikreuzer (approved):
lgtm, thanks!

PR #2965, opened 2022-05-15
Add event information in rules for time, manual and RunRuleAction trigger

### LINKED ISSUE #1391: GenericCronTriggerHandler does not give back context information
I am referring to the closed issue in JSR223/Jython:
[Class Decoration - CronTrigger - inputs map is empty](https://github.com/openhab-scripters/openhab-helper-libraries/issues/304#issuecomment-597796005)

I am not familiar with the code, but it would be enough if just the information which is put out in the log output would somehow be packed into the map:

`re.logger.debug("The trigger '{}' of rule '{}' is triggered.", trigger.getId(), ruleUID);`

In the module GenericCronTriggerHandler.java nothing is passed back.

If I have multiple Cron triggers inside one rule, it would be good to know which particular trigger was fired. Today this is not possible as no context information is passed.

J-N-K:
Closes #1391 

This creates event information for time-based triggers (DateTimeTrigger, GenericCronTrigger, TimeOfDayTrigger). The topic contains the trigger type, the trigger id (or label) is passed as source and the trigger configuration as payload of the event. 

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/events/TimerEvent.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/events/TimerEvent.java
@@ -27,7 +27,7 @@ public class TimerEvent extends AbstractEvent {
     public static final String TYPE = TimerEvent.class.getSimpleName();
 
     /**
-     * constructs a new rule added event
+     * Constructs a new timer event
      *
      * @param topic the topic of the event
      * @param payload the payload of the event (contains trigger configuration)
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/events/TimerEvent.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/events/TimerEvent.java
@@ -1,0 +1,49 @@
+/**
+ * Copyright (c) 2010-2023 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.automation.events;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+import org.openhab.core.events.AbstractEvent;
+
+/**
+ * An {@link TimerEvent} is only used to notify rules when timer triggers fire.
+ *
+ * @author Jan N. Klug - Initial contribution
+ */
+@NonNullByDefault
+public class TimerEvent extends AbstractEvent {
+
+    public static final String TYPE = TimerEvent.class.getSimpleName();
+
+    /**
+     * constructs a new rule added event
+     *
+     * @param topic the topic of the event
+     * @param payload the payload of the event (contains trigger configuration)
+     * @param source the source of the event
+     */
+    public TimerEvent(String topic, String payload, @Nullable String source) {
+        super(topic, payload, source);
+    }
+
+    @Override
+    public String getType() {
+        return TYPE;
+    }
+
+    @Override
+    public String toString() {
+        return "Timer " + getSource() + " triggered.";
+    }
+}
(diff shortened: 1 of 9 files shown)


ccutrer:
+1. This enhancement would allow the jrubyscripting helper gem (https://github.com/boc-tothefuture/openhab-jruby/) to remove its own CronTrigger type that's used simply to correlate which trigger triggered the rule.

ccutrer:
cc @boc-tothefuture

J-N-K:
What would be the "payload" of these events?

rkoshak:
Even if it's just a boolean flag that would be sufficient. Or maybe a new field could be added to `event` that is always there and has an enum indicating how the rule was triggered. I like that idea because it means we don't need to figure it out based on what's `undefined` and instead we could just check.

    if(this.event.triggeredBy = 'Time')
    if(this.event.triggeredBy = 'Manual')
    if(this.event.triggeredBy = 'CalledBy') // could carry an additional parameter with the UID of the rule that called it
    if(this.event.triggeredBy = 'ItemUpdate')
    if(this.event.triggeredBy = 'ItemCommand')

and so on.

ccutrer:
at the lowest level, _anything_ would be sufficient, so that the trigger's module id is in the context map in some way that's sent to the action's execute method. ideal would be an "event" object, descended from AbstractEvent. I don't even care if it doesn't have extra fields (a calling rule UID would definitely be nice for the call-rule action).

ccutrer:
> Even if it's just a boolean flag that would be sufficient. Or maybe a new field could be added to `event` that is always there and has an enum indicating how the rule was triggered. I like that idea because it means we don't need to figure it out based on what's `undefined` and instead we could just check.
> 
> ```
> if(this.event.triggeredBy = 'Time')
> if(this.event.triggeredBy = 'Manual')
> if(this.event.triggeredBy = 'CalledBy') // could carry an additional parameter with the UID of the rule that called it
> if(this.event.triggeredBy = 'ItemUpdate')
> if(this.event.triggeredBy = 'ItemCommand')
> ```
> 
> and so on.

Assuming we get a distinct descendant of `AbstractEvent` for each possibility, you could just check the event's class type instead of having an explicit enum.

J-N-K:
`getType` should be available on all existing events. So what we need is an event for triggering manually/by other rule.

J-N-K:
In JS a rule with

```
console.log("received: '" + event + "' with payload '" + event.getPayload() + "'")
```

result in the following logs

- `received: 'Execution triggered by manual' of type 'ExecutionEvent' with payload '{}'` when triggered from UI
- `received: 'Execution triggered by runRuleAction' of type 'ExecutionEvent' with payload '{"previous":{"topic":"openhab/execution/2d4f245946/triggered","payload":"{}","source":"manual"}}'`when triggered from another rule that was triggered manually
- `received: 'Timer 1 triggered.' of type 'TimerEvent' with payload '{"cronExpression":"0 0/5 * * * ? *"}'` when triggered from a cron trigger

LukasA83:
This feels really useful, can we get it merged?

rkoshak:
Ping @openhab/maintainers, this PR is awaited with great anticipation by me and others. It would be a great new feature to brag about for OH 4 if we can get it merged.

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/events/TimerEvent.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @author Jan N. Klug - Initial contribution
+ */
+@NonNullByDefault
+public class TimerEvent extends AbstractEvent {
+
+    public static final String TYPE = TimerEvent.class.getSimpleName();
+
+    /**
+     * constructs a new rule added event
```
kaikreuzer, line 30:
```suggestion
     * Constructs a new rule added event
```

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/DateTimeTriggerHandler.java
```diff
@@ -115,6 +119,9 @@ public void run() {
         ModuleHandlerCallback callback = this.callback;
         if (callback instanceof TriggerHandlerCallback triggerHandlerCallback) {
             triggerHandlerCallback.triggered(module);
```
kaikreuzer, line 121: Shouldn't this line be removed?

kaikreuzer (changes requested):
Thanks, just one small bug to fix.

kaikreuzer (approved):
Thanks!

PR #2968, opened 2022-05-16
Also escape special characters in state option values

Hilbrand:
And escaped space, = and : symbols (all legal according to properties documentation).

Follow up on #2964


tools/i18n-plugin/src/main/java/org/openhab/core/tools/i18n/plugin/XmlToTranslationsConverter.java
```diff
@@ -271,7 +271,8 @@ private TranslationsSection channelTypesSection(BundleInfo bundleInfo) {
             if (stateDescription != null) {
                 stateDescription.getOptions().stream()
                         .map(option -> entry(
-                                String.format("%s.%s.state.option.%s", keyPrefix, channelId, option.getValue()),
+                                String.format("%s.%s.state.option.%s", keyPrefix, channelId,
+                                        option.getValue().replaceAll("[ :=]", Matcher.quoteReplacement("\\ "))),
```
J-N-K, line 275: I'm not sure here. Replacing them all with an escaped space makes it impossible to have `:` and `=` as values (which could e.g. be used to select the separator for key-value pairs).

Hilbrand: Correct it should replace it properly. I've updated this.

PR #2970, opened 2022-05-18
Add REST support for deleting links and removing orphaned links

### LINKED ISSUE #1454: orphaned links after deleting item
When deleting an item through the rest api and the item has links those links do not get deleted automatically. They remain orphaned and still point to the deleted item.
When an item is deleted, all links that point to the item should be deleted, too.

[HABApp Issue](https://github.com/spacemanspiff2007/HABApp/issues/150)

J-N-K:
Closes #1454 

This add support for deleting all item-channel-links to an item or to a thing. This allows support for preventing orphaned links in UI (similar to the "delete item" when unlinking a channel). 

Also support for removing all orphaned links in the registry is added. I'm not sure where to expose this in the API. Something like `POST /links/compress` would be my first choice, but considering that we have the same issue with orphaned metadata and `POST /items/compress` is not possible (that is already used for sending a command to an item names `compress`), I'm not convinced that this is the best solution. Any ideas?

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
@@ -16,6 +16,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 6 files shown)


spacemanspiff2007:
To get a better understanding: what happens if you call this on a item from the `*items` file with

1.  a link that's defined in the file
2.  two links, one defined in the file and one from the GUI

J-N-K:
In all cases: managed links are removed, unmanaged links stay. That‘s a the drawback of allowing to configure links in .items files, they can’t be removed.

kaikreuzer:
> Also support for removing all orphaned links in the registry is added. 

Wouldn't "purge" be a better word instead of "compress" for this feature?

> Any ideas?

`POST /links/purge` would work and for items, we are specifically talking about metadata, so how about `POST /items/metadata/purge`? POST is not yet used on this path, so this should imho work.

J-N-K:
Done.

kaikreuzer (approved):
thanks, lgtm!

PR #2975, opened 2022-05-22
Assume installed addons are compatible

J-N-K:
It was reported that after an upgrade from aprevious version installed addons from marketplaces are not shown in the add-ons pages. This is a result of the new compatibility check which excludes incompatible add-ons by default. Already installed add-ons must be excluded from this check otherwise they can't be uninstalled.


wborn (approved):
Makes sense!

PR #2976, opened 2022-05-23
Fix null type mismatch errors

wborn:
This fixes the null type errors in Eclipse which were introduced by #2906.


J-N-K (approved):
Looks much cleaner. Thank you.

lolodomo:
@wborn : I believe you broke what was introduced for the interpret method:
```
        if (!interpreters.isEmpty()) {
            Locale locale = localeProvider.getLocale();
            for (var interpreter : interpreters) {
                try {
                    String answer = interpreter.interpret(locale, text);
                    logger.debug("Interpretation result: {}", answer);
                    return answer;
                } catch (InterpretationException e) {
                    logger.debug("Interpretation exception: {}", e.getMessage());
                    throw e;
                }
            }
        }
```
This will throw an exception as soon as it fails for one interpreter.
The idea was to try all interpreters until one is not failing.

lolodomo:
@GiviMAD for information.

GiviMAD:
@wborn @lolodomo yes in fact it breaks what we did, I will open another pr and ping you there. Watching the changes I think it could be simplified.

PR #2981, opened 2022-05-25
Wait for thing-type available in PersistentInbox

### LINKED ISSUE #1393: PersistentInbox should wait for ThingTypeRegistry
During the development of a new piece of code, involving a custom ThingTypeProvider, I noticed that Things are approved by the PersistentInbox (PersistentInbox.approve(ThingUID thingUID, @Nullable String label)), thereby mapping properties of DiscoveryResults to Thing properties and configuration using the ThingTypeRegistry. However, this is done before the ThingTypeProviders are added to the ThingTypeRegistry, and specifically in the case that autoApprove if Inbox items is set to true

PersistentInbox should use the ReadyService to delay itself until the ThingTypeRegistry is fully equipped.

J-N-K:
Fixes #1393

An alternative implementation would not break API (the return type of the `add` method would still be `boolean`), but this would move the retry-logic to the discovery service. IMO it is cleaner to have it in the `PersistentInbox` as it is specific to this implementation. It would also be difficult if the discovery result is rejected because the thing is already present.

The benefit of waiting on the thing type is that the normalization of the configuration values will always be possible.

### DIFF
--- a/bundles/org.openhab.core.config.discovery/src/main/java/org/openhab/core/config/discovery/inbox/Inbox.java
+++ b/bundles/org.openhab.core.config.discovery/src/main/java/org/openhab/core/config/discovery/inbox/Inbox.java
@@ -13,6 +13,7 @@
 package org.openhab.core.config.discovery.inbox;
 
 import java.util.List;
+import java.util.concurrent.CompletableFuture;
 import java.util.stream.Stream;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
@@ -57,7 +58,7 @@ public interface Inbox {
      * @param result the discovery result to be added to this inbox (could be null)
      * @return true if the specified discovery result could be added or updated, otherwise false
      */
-    boolean add(@Nullable DiscoveryResult result);
+    CompletableFuture<Boolean> add(@Nullable DiscoveryResult result);
 
     /**
      * Removes the {@link DiscoveryResult} associated with the specified {@code Thing} ID from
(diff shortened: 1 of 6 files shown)


kaikreuzer (approved):
lgtm, thanks!

jlaur:
@J-N-K - the openhab-addons build was broken by this, because the mielecloud binding expects `Inbox.add` to return boolean. I'm not sure what would be the correct fix for this, can you advise?

For reference:
https://github.com/openhab/openhab-addons/blob/848ec83d52620125da9aeaaba1933307b18b8f94/bundles/org.openhab.binding.mielecloud/src/main/java/org/openhab/binding/mielecloud/internal/config/servlet/CreateBridgeServlet.java#L130-L134

J-N-K:
Can you explain what you are trying to achieve with that code?

jlaur:
> Can you explain what you are trying to achieve with that code?

Unfortunately not, but I can have a second look and see if I can figure it out. Perhaps @BjoernLange is able to comment?

bundles/org.openhab.core.config.discovery/src/main/java/org/openhab/core/config/discovery/inbox/Inbox.java
```diff
@@ -57,7 +58,7 @@ public interface Inbox {
      * @param result the discovery result to be added to this inbox (could be null)
      * @return true if the specified discovery result could be added or updated, otherwise false
```
jlaur, line 59: @J-N-K - btw, Javadoc should be updated after this change.

BjoernLange:
> Unfortunately not, but I can have a second look and see if I can figure it out. Perhaps @BjoernLange is able to comment?

This code is supposed to create a Miele Cloud Account Thing after a successful OAuth2 pairing by putting a new discovery result into the inbox and approving it. Because the thing might already exist it is checked whether an inbox entry can be created. Depending on the result a new thing is approved from the inbox or it is tried to change the configuration of an existing thing.

PR #2982, opened 2022-05-26
Fix log levels and duplicate implementation warning in tests

J-N-K:
Related to #2939

Needed to solve issues in openhab-addons 


wborn (approved):
Thanks!

J-N-K:
It seems that a dependency is missing somewhere. I'll check that.

PR #2983, opened 2022-05-28
Fix dangling SSE subscriptions

### LINKED ISSUE #598: High memory usage & complete openHAB shutdown with many SSE subscriptions
Hi All,

as already mentioned in [an issue on the openHAB-distro repo](https://github.com/openhab/openhab-distro/issues/862) I've been experiencing a high memory and CPU usage after some time of running openHAB, up until the point, where the whole system becomes unresponsive after a couple of hours/days.

I developed a Homebridge plugin that uses a single subscription for every item used in Homebridge through the `http://<openhab-host>/rest/events?topics=smarthome/items/<habItem>/statechanged` interface. Due to the size of my installation I ended up having about 60 of those subscriptions opened. The updates should have been delivered consistently and all subscriptions were kept opened over the time (no stale subscriptions afaik). Nevertheless after changing the model to use a single subscription to `http://<openhab-host>/rest/events?topics=smarthome/items/` and filtering the relevant items on the receiving side I have not experienced any memory or CPU issues, therefore I think there is some unexpected behavior in this area.

I created a dump file using `dev:dump-create` from the console during the time, where the issue was happening, but I am not sure how to interpret the information. [Maybe you can have a look at that](https://www.dropbox.com/s/4z5nv63zjncqgg0/2019-02-18_123012.zip?dl=0)

I will continue looking at the issue and try to create a PoC. Until then maybe you can give your feedback/experience with that.

Thanks,

Frank

J-N-K:
Fixes #598

Unlike the sitemap SSE subscriptions generic event subscriptions did not implement a connection lost monitor.

@spacemanspiff2007 Can you confirm that this additional event does not cause issues in your application?

### DIFF
--- a/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/SseResource.java
+++ b/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/SseResource.java
@@ -20,6 +20,9 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
 
 import javax.annotation.security.RolesAllowed;
 import javax.inject.Singleton;
@@ -42,6 +45,7 @@ import javax.ws.rs.sse.SseEventSink;
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.auth.Role;
+import org.openhab.core.common.ThreadPoolManager;
 import org.openhab.core.events.Event;
 import org.openhab.core.io.rest.RESTConstants;
 import org.openhab.core.io.rest.RESTResource;
@@ -99,6 +103,10 @@ public class SseResource implements RESTResource, SsePublisher {
 
     private final Logger logger = LoggerFactory.getLogger(SseResource.class);
 
+    private final ScheduledExecutorService scheduler = ThreadPoolManager
+            .getScheduledPool(ThreadPoolManager.THREAD_POOL_NAME_COMMON);
+    private final ScheduledFuture<?> cleanSubscriptionsJob;
+
     private @Context @NonNullByDefault({}) Sse sse;
 
     private final SseBroadcaster<SseSinkItemInfo> itemStatesBroadcaster = new SseBroadcaster<>();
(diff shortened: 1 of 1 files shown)


lolodomo:
Is it for sitemao SSE? 
Because I have the strange feeling I already implemented this ALIVE message several years ago...

spacemanspiff2007:
@J-N-K 
Thanks for asking!

Any event that can't be mapped to the corresponding python type will generate an error in the logs (every time it is received).
Under which topic is the published? Or does this not contain the topic information at all and this will only be a "ALIVE" string?

lolodomo:
I think about MainUI and the remote openHAB binding as other users of the OH SSE. To be checked if they will not be impacted.

J-N-K:
@lolodomo Can you check for the remoteopenhab-binding? I have not seen issues in Main UI, but I‘ll check again.

lolodomo:
I have the same question as @spacemanspiff2007 : under which topic is the published?

lolodomo:
Hre are the topics "listened" by the remote openHAB binding: openhab/items/*/*,openhab/things/*/*,openhab/channels/*/triggered,openhab/channels/*/descriptionchanged

J-N-K:
There is no topic. It's just '{ "TYPE" : "ALIVE" }', similar to the alive-event from the sitemap (which contains some additional information about the sitemap not needed here).

spacemanspiff2007:
Do I have to respond to this item with a rest call in any way?

J-N-K:
No, the only reason is to detect a closed TCP connection (then the send will fail and the listener removed).

lolodomo:
@J-N-K : for the remote openHAB binding  I believe this would require a simple additional case to cover the type ALIVE here:
https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.remoteopenhab/src/main/java/org/openhab/binding/remoteopenhab/internal/rest/RemoteopenhabRestClient.java#L416
Could you confirm and ideally prepare a PR?

lolodomo:
If you change nothing, the only risk is to have regular DEBUG logs mentioning ALIVE type messages not supported. 
I can check that after this PR is merged and then propose the small enhancement to avoid the logs.

spacemanspiff2007:
> No, the only reason is to detect a closed TCP connection (then the send will fail and the listener removed).

Wouldn't it only be necessary to send this some time after the last message was sent?

J-N-K:
Technically yes, but that would require tracking which event made it to which subscription. Just keeping track of the last processed event globally is not sufficient, because of the filter it'll not reach all subscribers. IMO the burden of one extra event every 2 minutes does not justify more complex logic here.

spacemanspiff2007:
> IMO the burden of one extra event every 2 minutes does not justify more complex logic here.

No worries - I thought the filters are connection bound and the connection knows when it's sending an event.
I'll have to support it anyway so I am fine with your proposed solution. :+1:

wborn (approved):
Thanks for the fix!

lolodomo:
> If we change nothing, the only risk is to have regular DEBUG logs mentioning ALIVE type messages not supported. I can check that after this PR is merged and then propose the small enhancement to avoid these logs.

This is confirmed.

```
22:19:46.568 [DEBUG] [nternal.rest.RemoteopenhabRestClient] - Received unhandled event with name 'event' and data '{"type":"ALIVE"}'
```

I will have to propose a small update.

lolodomo:
Just for my curiosity: all OH SSE events have "message" as name while your ALIVE event had "event". Is it expected?

PR #2986, opened 2022-05-29
Fix SSE topic filters

### LINKED ISSUE #2473: SSE topics filter not matching correctly
The [event API](https://www.openhab.org/docs/developer/utils/events.html) allows us to subscribe to item events via HTML5 [server-sent events](https://html.spec.whatwg.org/multipage/server-sent-events.html) of openHAB [Rest API](https://www.openhab.org/docs/configuration/restdocs.html).

## Expected Behavior
Using the openHAB Rest API endpoint `/events` to subscribe to the specific topic `openhab/items/{itemname}/state` should lead to a subscription of this topic only.

## Current Behavior
A subscription to the specific topic `openhab/items/{itemname}/state` leads to a subscription of the two topics `openhab/items/{itemname}/state` and `openhab/items/{itemname}/statechanged`.

## Possible Solution
I think the topic filtering is done in the file [SseSinkTopicInfo.java](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/internal/SseSinkTopicInfo.java):36 using the regex filter anymatch. Maybe an adjustment of the regex by appending a $ would help but I am not sure if that collides with the TOPIC_VALIDATE_PATTERN in file [SseUtil.java](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/internal/util/SseUtil.java):70.

## Steps to Reproduce (for bugs)
`curl "http://{openHAB_IP}:8080/rest/events?topics=openhab/items/{itemname}/state"`

## Your Environment
Docker openhab/openhab:3.1.0 - linux/arm/v7

J-N-K:
Fixes #2473 

According to the [documentation](https://www.openhab.org/docs/developer/utils/events.html), topic filters are regular expressions. The SSE topic filters are converted to regular expressions, therefore `openhab/items/*/state` should result in `openhab/items/.*/state`. The code however adds a `.*` at the end `openhab/items/.*/state.*`. This makes it impossible to filter for `/state` but exclude `/statechanged`, which is IMO expected. 

### DIFF
--- a/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/internal/util/SseUtil.java
+++ b/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/internal/util/SseUtil.java
@@ -85,7 +85,7 @@ public class SseUtil {
         } else {
             StringTokenizer tokenizer = new StringTokenizer(topicFilter, ",");
             while (tokenizer.hasMoreElements()) {
-                String regex = tokenizer.nextToken().trim().replace("*", ".*") + ".*";
+                String regex = tokenizer.nextToken().trim().replace("*", ".*") + "$";
                 filters.add(regex);
             }
         }
--- a/bundles/org.openhab.core.io.rest.sse/src/test/java/org/openhab/core/io/rest/sse/internal/util/SseUtilTest.java
+++ b/bundles/org.openhab.core.io.rest.sse/src/test/java/org/openhab/core/io/rest/sse/internal/util/SseUtilTest.java
@@ -60,7 +60,7 @@ public class SseUtilTest {
     @Test
     public void testFilterMatchers() {
         List<String> regexes = SseUtil
-                .convertToRegex("openhab/*/test/test/test/test,    openhab/test/*/test/test/test, openhab,qivicon");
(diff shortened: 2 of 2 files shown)


ghys:
Please hold on merging this as the UI is probably making requests that assume the current syntax, so the impact has to be assessed.

J-N-K:
According to the list you showed in the linked issue everything should be fine, since proper matching is done for all things, items, rules and addons (`*` at the end or where a wildcard should be used). 

You would only be affected if you subscribed to `openhab/items` (without `*`) and expected a subscription for all topics starting with `openhab/items`. As long as a `*` is at the end of the filter, nothing changes.

lolodomo:
@J-N-K : does this change break my  event subscription? 
https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.remoteopenhab/src/main/java/org/openhab/binding/remoteopenhab/internal/rest/RemoteopenhabRestClient.java#L333

J-N-K:
No, it‘ll only break if your subscription does not end with * AND you rely on an implicitly added * at the end (your subscription ends with `state` instead of `state*` but you expect to also cover `statechanged`).

PR #2987, opened 2022-05-30
Remove unnecessary script type configuration from script transformation

J-N-K:
This is breaking change, but IMO since the script transformation was only recently introduced, it is acceptable. With this change it is possible to use `my_transformation.js` as configuration instead of `js:my_transformation.js` which is more clear.


kaikreuzer:
I'm wondering, if we should maybe keep the prefix and instead change the file extension to ".script", so that it respects the naming convention for files in the `transform` folder - see also my comment [here](https://github.com/openhab/openhab-core/pull/2872#issuecomment-1146587570).

J-N-K:
Good point.

PR #2993, opened 2022-06-02
Bump maven-plugin-plugin to 3.6.4

J-N-K:
Needed for Java 17 source level. This also removes build errors with wrong scope and an unused dependency.


wborn (approved):
Thank you!

PR #2996, opened 2022-06-06
Fix DateTime trigger syntax to not break the equals function in the execution script

kaikreuzer:
This fixes a critical regression caused by #2963, which made the `equals` method disfunctional within rules.

This resulted in log messages like:
```
2022-06-06 11:30:46.587 [WARN ] [el.core.internal.ModelRepositoryImpl] - Configuration model 'test.rules' has errors, therefore ignoring it: [91,83]: no viable alternative at input 'equals'
```

It seems that Xtext is able to parse everything, if we use the same vocabulary as for the TimerTrigger, i.e. "is". The only drawback might be that the DateTime trigger cannot be used with items of name `midnight` or `noon`, but I think this is acceptable.

### DIFF
--- a/bundles/org.openhab.core.model.rule/src/org/openhab/core/model/rule/Rules.xtext
+++ b/bundles/org.openhab.core.model.rule/src/org/openhab/core/model/rule/Rules.xtext
@@ -76,7 +76,7 @@ TimerTrigger:
 ;
 
 DateTimeTrigger:
-    'Time' 'equals' item=ItemName
+    'Time' 'is' item=ItemName
 ;
 
 SystemTrigger:


J-N-K:
Interesting. IIRC it didn't work for me with `is`, that's why I choose `equals`.

kaikreuzer:
I only tested it within my IDE, but there it seemed to work. And compilation definitely worked, so Xtext does not see any conflict in the definition.

J-N-K:
I'm currently compiling your change locally to check that it also works in a regular runtime.

J-N-K (approved):

Issue #844, opened 2019-05-31
Historic state updates from bindings


wborn:
<a alt="bounty" href="https://www.bountysource.com/issues/74911053"><img src="https://api.bountysource.com/badge/issue?issue_id=74911053"></a>

With the implementation of https://github.com/eclipse/smarthome/pull/1872 there is now some basic infrastructure for handling historic state. However there still seems to be no API for historic state updates from bindings.

Ofcourse to get it all working there also needs to be a `ModifiablePersistenceService` implementation for this.

See also:
* [OH community discussion](https://community.openhab.org/t/howto-integrate-a-device-returning-a-history-of-measurements/21127)
* https://github.com/eclipse/smarthome/issues/4765

wborn:
The bounty has also been migrated from ESH! After closing the ESH issue I claimed it myself. :dollar:

IamSomin:
I have run into this issue while experimenting with bindings. After reading the related issues, as far as I can tell, no progress has been made on this? Would it still make sense to implement ModifiablePersistenceService?

There is one main issue I can think of: Bindings usually do not work with items directly. I think there either needs to be an easy way for bindings to figure out what items new states should be persisted to, or ideally an option to provide historic states directly to channels (although I think that would require a lot more work).

I think the best way forward would be to start compiling a list of all persistence services that could implement ModifiablePersistenceService and work with their maintainers? That would solve half of the problem and as the REST API already has the capability to work with these services, it would actually add some improved functionality by itself.

wborn:
There hasn't been any progress so far and I think the `ModifiablePersistenceService` interface can still be used for this. We would probably still want to hide the Items from `ThingHandler`s and let the framework figure out which Items are linked and have what persistence configured. I.e. we could add some methods similar to the `updateState(...)` methods, e.g. `historicState(...)` which would have the same method signature but also have a `ZonedDateTime` parameter.

This issue is about providing the API in the core. The persistence add-ons can already implement a `ModifiablePersistenceService` interface today, although it is not very useful since you can only use them with REST calls currently. :wink: 

My use case for this would be to store historic state into persistence in the event that my openHAB instance was offline for some reason. Then when it is up and running again, the binding can request missing historic state from devices and persist it using this API. That way I would be less bothered about downtime. :slightly_smiling_face:

marvkis:
[...]

Issue #1298, opened 2019-12-26
Extending states with attributes


cdjackson:
For OH3, I think we should take a serious look at how states / commands are defined. We've had the legacy system for 10 years now, and I personally think it's creaking at the seems and limiting the way we can run the system.

Currently, we have a single dimension in the state - e.g. ON/OFF, a single value, etc. This is fine for basic information that goes to a user or UI, however, in many cases further information is required (or at least, is provided from the device sending the state) to fully describe an event. While this information may not be of primary concern, and may not be used in a UI, it is useful for triggering rules/scenes and I feel it would be useful to think if this could be attached to the state event.

A few use cases -:
* A ZigBee light can provide a transition time - ie the amount of time required to move from ON to OFF. We want to be able to define this attribute, along with the state.
* Z-Wave scene controller can can send a scene 'command' - this usually has a number of attributes - the scene number, but also the keypress type (KEY-DOWN, KEY-UP, etc), and sometimes a time is also provided. This is currently handled by defining a number - (scene number).(type) - e.g. 6.3 might be used for scene 6, keydown - we ignore the time so the device can not be fully utilised.
* An alarm can have a type, and a state. Instead of implementing 25 channels for 25 different alarms, each with an on/off condition, it might be more convenient to provide a single Alarm channel with on/off state and alarm type attribute. The UI can show there's an alarm (i.e. showing the channel and it's state - as we now have) but a rule would be able to look at the attribute and decide exactly what alarm type this is and act with the appropriate response.
* A smoke alarm can go into alarm - we want to show this on the UI, however the notification rule that rings the fire brigade probably doesn't want to do this if it's a test - we could add the "Test=true" attribute so that the system can work around this.
* A door lock can have the state ON or OFF (LOCKED or UNLOCKED) but this can have attributes such as the reason it changed state (eg the user who opened the door).

Of course there are always ways around this with the current system - we can concatenate data together(e.g. alarm type number x 100 + alarm value), or we can use two channels, but then the rule has to resynchronise these two streams. Or we can use json strings to hold multiple bits of data which has been the recommendation so far.

I'd like to propose an extension to State updates to hold attributes - ie secondary data that provides context to the primary state.

Currently, we have -:

```
updateState(channelUID, state)
```

I believe (following an extremely quick look at the code!) that this could be extended with another parameter containing an object / attribute map. eg

```
updateState(channelUID, state, object)
```
eg in pseudo code.
```
[...]

PR #3000, opened 2022-06-14
Enable binding to store historic states

altaroca:
please refer to #844
this adds a method `BaseThingHandler.updateHistoricState (ChannelUID channelUID, State state, ZonedDateTime dateTime)` and all logic to route the corresponding `ItemHistoricStateEvent` through the core to the persistence layer and any `ModifiablePersistenceService`.
I have tested it using my [fork](https://github.com/altaroca/openhab-addons/tree/E3DC) of the E3DC binding and influxdb modified to implement `ModifiablePersistenceService`.
I made some choices for simplicity sake:
- the item current state and its historic state only meet in the persistence. Thus the historic state (even if younger) does not affect the current state.
- the events are separate but share the same publication topic


bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/ThingHandlerCallback.java
```diff
@@ -57,6 +58,15 @@ public interface ThingHandlerCallback {
      */
     void stateUpdated(ChannelUID channelUID, State state);
 
+    /**
+     * Informs about an update to a historic state for a channel.
+     *
+     * @param channelUID channel UID (must not be null)
```
J-N-K, line 64: You can remove all "must not be null" in this class, as the class is annotated with `@NonNullByDefault`

bundles/org.openhab.core/src/main/java/org/openhab/core/items/events/ItemEventFactory.java
```diff
@@ -268,6 +281,26 @@ public static ItemEvent createStateEvent(String itemName, State state) {
...
+     * @param itemName the name of the item to send the state update for
+     * @param state the new state to send
+     * @param source the name of the source identifying the sender (can be null)
+     * @return the created item state event
+     * @throws IllegalArgumentException if itemName or state is null
+     */
+    public static ItemHistoricStateEvent createHistoricStateEvent(String itemName, State state, ZonedDateTime dateTime,
+            @Nullable String source) {
+        assertValidArguments(itemName, state, "state");
+        // TODO: use a different topic for historic state ? Would require changes to ItemUpdater.
```
J-N-K, line 296: You already defined a constant for that above. Why don't you use it here? IMO the topic has to be different, because the payload is different and the event also serves a different purpose

bundles/org.openhab.core/src/main/java/org/openhab/core/items/events/ItemEventFactory.java
```diff
@@ -268,6 +281,26 @@ public static ItemEvent createStateEvent(String itemName, State state) {
...
+    /**
+     * Creates an item historic state event.
+     *
+     * @param itemName the name of the item to send the state update for
+     * @param state the new state to send
+     * @param source the name of the source identifying the sender (can be null)
+     * @return the created item state event
+     * @throws IllegalArgumentException if itemName or state is null
+     */
+    public static ItemHistoricStateEvent createHistoricStateEvent(String itemName, State state, ZonedDateTime dateTime,
```
J-N-K, line 293: `dateTime` is missing from JavaDoc

bundles/org.openhab.core/src/main/java/org/openhab/core/items/events/ItemHistoricStateEvent.java
```diff
@@ -0,0 +1,80 @@
...
+     * Constructs a new item state event.
+     *
+     * @param topic the topic
+     * @param payload the payload
+     * @param itemName the item name
+     * @param itemState the item state
+     * @param source the source, can be null
+     */
+    protected ItemHistoricStateEvent(String topic, String payload, String itemName, State itemState,
+            ZonedDateTime dateTime, @Nullable String source) {
```
J-N-K, line 46: `dateTime` is missing from JavaDoc

bundles/org.openhab.core/src/main/java/org/openhab/core/items/GenericItem.java
```diff
@@ -269,6 +294,22 @@ protected void notifyListeners(final State oldState, final State newState) {
         }
     }
 
+    protected void notifyHistoryListeners(final State state, final ZonedDateTime dateTime) {
+        // if nothing has changed, we send update notifications
```
J-N-K, line 298: That's not what you do. Remove comment

bundles/org.openhab.core/src/main/java/org/openhab/core/items/HistoricStateChangeListener.java
```diff
@@ -0,0 +1,39 @@
...
+ * <p>
+ * This interface must be implemented by all classes that want to be notified about changes in the state of an item.
+ *
+ * <p>
+ * The {@link GenericItem} class provides the possibility to register such listeners.
+ *
+ * @author Kai Kreuzer - Initial contribution
+ */
+@NonNullByDefault
+public interface HistoricStateChangeListener extends StateChangeListener {
```
J-N-K, line 30: I would prefer the new method to be integrated in the old `StateChangeListener` interface with a default implenentation that throws an `UnsupportedOperationException`.

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/internal/PersistenceManagerImpl.java
```diff
@@ -158,6 +160,39 @@ private void handleStateEvent(Item item, boolean onlyChanges) {
         }
     }
 
+    /**
+     * Calls all persistence services which use change or update policy for the given item
+     *
+     * @param item the item to persist
+     * @param onlyChanges true, if it has the change strategy, false otherwise
+     */
+    private void handleHistoricStateEvent(Item item, State state, ZonedDateTime dateTime, boolean onlyChanges) {
```
J-N-K, line 169: What is the `onlyChanges` parameter used for? If I don't overlook something, you always call it with `false`.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/BaseThingHandler.java
```diff
@@ -287,6 +288,39 @@ protected void updateState(String channelID, State state) {
...
     }
 
+    /**
+     *
+     * Updates a historic state of the thing.
+     *
+     * @param channelUID unique id of the channel, which was updated
+     * @param state new state
+     */
+    protected void updateHistoricState(ChannelUID channelUID, State state, ZonedDateTime dateTime) {
```
J-N-K, line 298: `dateTime` is missing in JavaDoc

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/BaseThingHandler.java
```diff
@@ -287,6 +288,39 @@ protected void updateState(String channelID, State state) {
...
+        synchronized (this) {
+            if (this.callback != null) {
+                // TODO: check if this historic state is later than any existing state
```
J-N-K, line 301: What is needed here?

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/BaseThingHandler.java
```diff
@@ -287,6 +288,39 @@ protected void updateState(String channelID, State state) {
...
+
+    /**
+     *
+     * Updates a historic state of the thing. Will use the thing UID to infer the
+     * unique channel UID from the given ID.
+     *
+     * @param channelID id of the channel, which was updated
+     * @param state new state
+     */
+    protected void updateHistoricState(String channelID, State state, ZonedDateTime dateTime) {
```
J-N-K, line 319: `dateTime` missing in JavaDoc

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/HistoricState.java
```diff
@@ -0,0 +1,52 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.thing.internal;
+
+import java.time.ZonedDateTime;
+
+import org.eclipse.jdt.annotation.Nullable;
+import org.openhab.core.types.State;
+
+public class HistoricState implements State {
```
J-N-K, line 20: annotate with @NonNullByDefault, JavaDoc missing

altaroca:
wow, that was fast. I must apologize. I won’t be able to react that quickly. Please give me a few days.

altaroca:
@J-N-K i agree with your proposals and have made changes accordingly. 

> What happens if the state has a date this is in the future?
that date is treated as any other. In other words, the binding is responsible for providing a correct date. There may be instances where a future date is needed. Maybe some kind of forecast? I cannot think of anything useful now.

splatch:
While I like a lot that this topic come to concrete form I suppose that a bit better way would be proper tagging of states instead of multiplication of events. As outlined - bindings can provide estimations which are ahead of time, hence it would require another kind of event.

I think that making a type `TimeBoudState[DateTime, State]` would better reflect that case because:
- This could be still a valid state
- It would allow to get whole thing working with channel-item profiles
- It can be detected by persistence adapters

Thing is, if core is ready for such step?

altaroca:
@splatch I do not fully understand your comment. What do you mean by 'can still be a valid state'?
The PR provides the entity you propose. Only difference is it is called `HistoricState`.

splatch:
> The PR provides the entity you propose. Only difference is it is called `HistoricState`.

@altaroca roger, I missed type itself and got lost by changes in other areas. If you rely on the state then main question I could raise is why distinguish situation by introducing `ThingCallback.historicStateUpdated(ChannelUID channelUID, State state, ZonedDateTime dateTime)` which hide access to this state in bindings. If bindings have access to String/DateTime and other representations, then this part of change, strictly speaking is not required.

I see a point in making such state a first level citizen because some of bindings could provide a timestamp of received update. To give more concrete example - a CAN interface controller can attach timestamp to the message which delivers state update. Binding then knows exact timestamp of an update and it is not a historic state per say as it is instant value with receive time information.

altaroca:
@splatch I had thought about that and decided against it. There were two reasons:

1. sub-classes of `State` are often checked by using expressions like `instanceof DecimalType`. This breaks if there was another type hierarchy based on `HistoricState` like `DecimalHistoricType` etc. This might be solved but requires a lot of refactoring in the core and in add-ons.
2. I fear the complexity of deciding which state is the current one. Will a younger `HistoricState` overwrite the current state? Yes, for timed measurements. No, for forecasts in the same channel. I do not see a clear rule. That's why I kept `State` and `HistoricState` separate.

splatch:
1) Why you consider checking and casting for `DecimalState` as a fine operation, but checking for `TimeBoundState` as something wrong? It is a state just like others. In case if primitive type is required you can rely on composition such as below

```
interface WrapperState extends State /*, ComplexType? */ {
  State unwrap();
}
class TimeBoundState implements WrapperState {
  private DateTime dateTime;
  private State state;
  State unwrap() {
    State myState = state;
    do {
       // it can be also moved to State/Type itself so DecimalType will always return this, but wrapper states will unpack themselves
       myState = myState instanceof WrapperState ? ((WrapperState) state).unpack() : state;
    } while (myState instanceof WrapperState);
    return myState;
  }
}

// for callers doing persistence and other parties which need to process concrete state instance
State actualState = state instanceof WrapperState ? ((WrapperState) state).unpack() : state;
if (actualState instanceof DecimalType) {
   ...
}
```

2) If you keep going with a `HistoricState` then its a matter of question when we will need to start dealing with `FutureState`, `ScheduledState` and related framework changes. Obviously these situations may not appear immediately, but what you signal with historic state is a paired information state+timestamp. If you look at it in such way then it doesn't matter if its past, future or almost now.
Processing of time bound states might be completely omitted by items, or based on time accuracy criteria. Just like you do it right now via `applyHistoricState`. Its based on separate event but may be based on state kind iself.

altaroca:
> 1. Why you consider checking and casting for `DecimalState` as a fine operation, but checking for `TimeBoundState` as something wrong? 

Err, no. That's not my point. I meant to say that no existing code will understand the meaning of types like `DecimalTimeBoundState` because of the way everyone uses `instanceof` on known sub-classes of `State`. And we would have to refactor classes like `InfluxDBStateConvertUtils`.

> 2. If you keep going with a `HistoricState` then its a matter of question when we will need to start dealing with `FutureState`, `ScheduledState` and related framework changes. Obviously these situations may not appear immediately, but what you signal with historic state is a paired information state+timestamp. If you look at it in such way then it doesn't matter if its past, future or almost now.

Of course. I completely agree. And I do not care about the name. It's up for discussion what's the best name: `HistoricState`, `TimeBoundState` or something else. 

What I mean is that people, seeing `State` and `HistoricState` being passed along as equal in the same channel will start to wonder which one takes precedence if any. Or do you suggest to have entirely different channels for `State` and `HistoricState`? My current implementation uses the same channel and you could interpret `HistoricState` as a kind of sub-channel on the primary channel.

J-N-K:
IMO if timestamps in the future are allowed, something like `TimeboundState` would be better than `HistoricState` as that implies that the time is in the past.

I wonder if it would not be better to extend `State` with a `DateTime getDateTime()`.(I personally would prefer a `Long getTimeStamp()`, but that's debatable). A default implementation could return the current time (which is similar to what we have now. If the consumer does not support timed states, but the state does have time information, the time information would be discarded. 

This could also replace the need for the `HistoricState` in the persistence implementations.

J-N-K:
@openhab/core-maintainers I would really like to hear your opinion here.

wborn:
Thanks for working on this @altaroca! :+1: 

> I would really like to hear your opinion here.

I don't have a strong opinion but it seems that in previous related PRs the timestamp always has been kept separate from the `State` as well.

See for instance this code:

* [org.openhab.core.persistence.ModifiablePersistenceService.store(Item, ZonedDateTime, State)](https://github.com/openhab/openhab-core/blob/cabb3f73150d13f44d647f46af8362e396684b92/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/ModifiablePersistenceService.java#L47)
* [org.openhab.core.persistence.HistoricItem.getTimestamp()](https://github.com/openhab/openhab-core/blob/cabb3f73150d13f44d647f46af8362e396684b92/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/HistoricItem.java#L39)

J-N-K:
Yes, but I don't understand the reasoning behind that. From what @kaikreuzer said in #3033 a possible argument is that this timestamp might not be the actual timestamp of the value change. While this might be true, if the timestamp is @Nullable, a binding can chose not to set it, if it is not known. But having a timestamp together with a state will remove the need for extra-code to handle future values (e.g. forecasts) and historic values (retrieved from persistence service) and will also allow bindings that CAN determine the exact point in time of a value change to communicate that. And it is fully backward compatible because all add-ons agnostic of the timestamp will not be affected (they produce states without timestamps, just like before and they consume states ignoring the timestamp, just like before), but persistence services can properly store the timestamp together with the value. If no timestamp is present, they use the system time, like it is now.

altaroca:
i cannot foresee all downstream implications of this change. Let's approach it semantically: 
_state_ means (abbreviated from Merriam Webster)
1. mode or condition of being,
2. a condition or stage in the physical being of something

This definition does not refer to a timestamp. So if you wanted to always combine a State object with a time you should at least call it differently, such as _StateWithTime_ or the above _TimeboundState_.

Now the question is: can openhab do without a State object and can you replace it in all places with a TimeboundState object?

wborn:
Adding forecasts on the same channels probably will also have its own set of issues. You probably don't want to persist forcasted state and actual state using the same item. I can imagine there are also devices that store their forecasts made at a certain time in a buffer. So if you then read these forecasts at a later time you might want to persist "historic forecasts". :wink:

J-N-K:
It's questionable if someone wants to persist forecasts at all (maybe to use `restoreOnStartup`). Anyhow, if the timestamp is persisted together with the value, I don't see any issues arising from that.

My concern is that we now have `State`, add `HistoricState` for persisted states (or states that shall be persisted) here and then add `FutureState` for forecasts later. Then we have three non-interchangable types for representing a state. 

Keeping aside that this leads to a very bulky API (and therefore a high burden for developers), I doubt that users will benefit from that. We now separate `Command` and `State` and even that is not understood by many. Do we really expect a majority to understand why a `State` is not a `State` depending on the presence of a time information? And even more why there are two different types of `State`s with time-information? Or would the user prefer to have one single type to deal with? 

I doubt that this `HistoricState` can be kept internal. As soon as such functionality is added someone will request to make it available e.g. in rules.

seime:
I do see usecases for persisting forecasts - electricity prices and weather are two. If my network goes down and openHAB is restarted, I would loose the ability to make decisions on these data without persistence.

IMHO a `State` _should_ have a timestamp reflecting the actual time this value was sampled/produced - by the sensor/service or by the framework if the underlying sensor/service lacks this.

The `State` timestamp is typically different from the `Event` timestamp. which is the timestamp the `Event` _itself_ was produced. I think we need both - if only not to mix the two concepts and create confusion.

Could we explore the feasability to create new `Event` types and topics for historic and forecast timeseries? Ie one (set) for bindings that produce these time series and that the framework listens for (and persists the included `States` with timestamp), and another set of events for rule engines to listen for when new timeseries are available? 
I'm not very familiar with the inner workings of openHAB here, so bear with me if it breaks backwards compat, creates a lot of work or is simply a bad idea.

spacemanspiff2007:
I too think that this adds way too much complexity without providing enough benefit.

Persisting forecasts/historic states is something that can easily be done in rules/through the REST API (see #12663).
If a binding needs this information why not just let the user pass it in through a channel (e.g. in a rule or through an item).
An additional benefit would be that the binding logic does not depend on the persistence service and the user can define a custom fallback e.g. if the service is not available or some other error occurs.

Timestamps have been (partially) discussed in #3033 and imho the only place where it might make sense is for a new `ItemStateUpdatedEvent`. There it's clear that it's the timestamp of the item state update.

cdjackson:
> Persisting forecasts/historic states is something that can easily be done in rules/through the REST AP

I don't think a binding should be forced to use the REST API - IMHO the REST API is for external use and forcing "internal" users to use this feels like a kludge and (IMHO) makes a binding a second class citizen.

> If a binding needs this information why not just let the user pass it in through a channel (e.g. in a rule or through an item).

If you mean to provide a separate channel for the time information, then IMHO this is not a good idea. You loose the consistency - you now have to find some way to correlate the state and time into a single object and that's not easy. OH doesn't guarantee that events will be kept in order, so a state1/time1/state2/time2 stream might be received as state1/state2/time2/time1. This also stands true for other information where states are split into multiple channels (eg a dimmer value, and the rate of change) - these really should all be considered as "attributes" of a state rather than as separate channels. I know this was discussed and rejected previously though...

> An additional benefit would be that the binding logic does not depend on the persistence service

I've not looked at all the comments etc, but I don't understand why adding a timestamp to a state requires interaction between the binding and the persistence. I think @J-N-K stated earlier that this should be backward compatible - if a binding adds a time to a state, then the persistence can persist the time - if not, it falls back to the existing implementation.

There are quite a few devices that can provide historical data - eg both Zigbee and ZWave meters can store information in the event that the controller is offline, and provide this historical data later. Currently it's not possible (without some sort of kludge) for the binding to incorporate this data back into the system so I would certainly support something that can do this.

splatch:
> Adding forecasts on the same channels probably will also have its own set of issues

FYI forecasts and states could be easily filtered out through appropriate profiles. You can use a single channel as source of data and fence items with proper profile to keep forecast or actual state.

> It's questionable if someone wants to persist forecasts at all (maybe to use `restoreOnStartup`). Anyhow, if the timestamp is persisted together with the value, I don't see any issues arising from that.

Well... a consumption of energy from beginning of month till now can be stored at the end of current month. It is stored ahead of time at the very last point of a time bucket.

> Do we really expect a majority to understand why a `State` is not a `State` depending on the presence of a time information? And even more why there are two different types of `State`s with time-information? Or would the user prefer to have one single type to deal with?

Developers and end users so far lived without a time next to state, so they will survive without it a little bit more. They will survive introduction of other *kind* of state if they will benefit from it. Otherwise they will complain.
My personal rationale is that we can't keep state limited to existing ones in wider scenarios. We either think ahead and design it or keep fooling ourselves that we can cover all new born scenarios with tight list of states existing since a decade (ie. by use of strings with json payloads) or individual fields and processors added by case basis here and there. Yet it will pollute and complicate core even more.

spacemanspiff2007:
> > Persisting forecasts/historic states is something that can easily be done in rules/through the REST AP
> 
> I don't think a binding should be forced to use the REST API - IMHO the REST API is for external use and forcing "internal" users to use this feels like a kludge and (IMHO) makes a binding a second class citizen.

My argument was not about the binding having to use the rest api. It was
"If the user wants to persist future/historic states he shall do so in rules ***OR*** the rest api". 

If bindings write to persistence a whole new set of issues arise:
- what if I don't use persistence at all, can I use the binding?
- what if my persistence doesn't provide the features the binding uses
- what if I want to use a different persistence for historic/future states
- Assuming that the same item is used:
  if things fail and I look a couple of days later at persisted values, are they values that actually happened or forecast values?

I view bindings as a hardware abstraction layer and as such it shouldn't persist forecast or historic states at all.
We have this good separation between binding, persistence, rules, etc. and it's a good idea to keep these blocks separated.

> If you mean to provide a separate channel for the time information, then IMHO this is not a good idea.

I mean to provide channels that accept the aggregated values as input, e.g. as comma separated values or json.
It's not pretty but at least it's transparent at what's happening and whats going in/out of the binding.

From my understanding this whole "the binding shall use persistence" is only because it's not possible to pass in multiple values in a structured way. 
Maybe it really is time for a native json or array datatype in openHAB ... .

cdjackson:
> My argument was not about the binding having to use the rest api. It was
"If the user wants to persist future/historic states he shall do so in rules OR the rest api".

Ok, no problem there then - the REST API is fine for the user :)

> If bindings write to persistence a whole new set of issues arise:

Maybe i misunderstood, but I don't think anyone is proposing that bindings write to persistence? The binding updates a state - the persistence system effectively receives these events, and persists data. The binding doesn't directly communicate to persistence. There's no change from the status quo.

> I view bindings as a hardware abstraction layer and as such it shouldn't persist forecast or historic states at all.

Yes, fundamentally this is correct - it's a hardware abstraction. However what if the hardware provides historic data? Eg what if the hardware provides an alert - at 3PM an alarm went off and the system wants to report this when the power comes back up at 5PM. Currently it would report that the power went off, and then came back on, both at 5PM.

Again, the binding is not "persisting" any data - it's just providing data from the hardware, with a timestamp. The persistence layer does the persisting with no direct interaction from the binding.

> I mean to provide channels that accept the aggregated values as input, e.g. as comma separated values or json.

Sadly this is unfortunately the way it currently needs to work and as you say, it's very ugly and very non-standard. ie every binding does something different, and therefore there's no standardisation, and the user is left with the mess. I think the point here about adding times, or even other supplementary data as per my earlier example was to try and provide a standard way to manage this sort of information.

> Maybe it really is time for a native json or array datatype in openHAB

I don't think adding a new type is the right approach as IMHO it doesn't solve the problem and creates a compatability mess. I proposed some time back the idea of adding additional attributes - eg times, speeds, etc to existing data types. This could I believe be added and made backward compatible. However this went nowhere unfortunately.

https://github.com/openhab/openhab-core/issues/1298

In any case I think that's off topic as this issue is really only looking at adding the time and not other data attributes.

J-N-K:
I think the original issue is exactly what @cdjackson describes: have a way for bindings to publish historical data with the correct timestamp if they are available. If they are only displayed, they should probably be discarded, if they are persisted, this will be taken care of by the persistence service that is responsible for handling states of the respective item. 

This is not about entering historical data via UI (this should clearly be handled by the REST API, and I believe the necessary endpoints are already there). For bindings to publish historical data or forecasts the REST API is not the way to go. I believe it boils down to three options:

1. Add a timestamp to the `State`. This can be done by adding a `getTimeStamp` to the `State` interface or adding a new interface `TimeboundState` that can be implemented by the implementations of `State` (or even `HistoricalState` and `Forecast`, that would allow us to handle state updates differently depending on what they are by simple `instance`-checks).
2. Add a new datatype `TimeboundState` and duplicate all implementations of `State`, so that we have `TimeboundDecimalType`, `TimeboundHSBType` and so on.
3. Add new methods like `updateState(State state, long timestamp)` or similar. 

IMO 2 is the worst solution, 3 is most simple (but clutters API) and 1 is the most flexible but needs some thoughts on the handling. The big plus here is that it is fully backward compatible.

splatch:
> 2\. Add a new datatype `TimeboundState` and duplicate all implementations of `State`, so that we have `TimeboundDecimalType`, `TimeboundHSBType` and so on.

This is deeply wrong. Time is additional marker for a state and can be composed with it. With wrapper/decorators you can keep raw states as is. See example I made: https://github.com/openhab/openhab-core/pull/3000#issuecomment-1164560513. By this way timebound states will be ignored unless affected places acknowledge it and do necessary `unpack` call.

cdjackson:
> 1. Add a timestamp to the State. 

My vote goes here. Possibly an extended interface as you suggest might be a good idea as it can act as a marker (as you also mention).


> 1b would also be a possible way for https://github.com/openhab/openhab-core/issues/1298 (add ExtendedState interface for additional data).

Again, maybe off topic, but I'd be very keen if this was picked up again and happy to help with this. I think it's probably a V4 issue (I tried pushing it for the OH2 -> OH3 transition) more because it's a significant change/addition to the concept, even if I think it's not a breaking change.

kaikreuzer:
If I see it correctly, the PR at hand more or less corresponds to option (3), which is clearly the least intrusive.

I like the separation between `State` and `HistoricState` and think the way they are used are very different.
The reason I am reluctant with adding a timestamp to the `State` itself (besides what I already wrote in #3033) is that State updates on the event bus are considered to be a time-series. That is: A continuous data stream which is unmodifiable and with a strict order by time - i.e. more recent events in the stream are always also newer datapoints. This is what everything relies on. No consumer (like the various UIs) have to check whether they ever received a state with a more recent timestamp before showing the latest received state update - it would make their logic very complex if state updates could suddenly come in un-chronological order. Please note here that he very first persistence service in openHAB was the "logging persistence". All it did was writing the state updates to a log file in sequential order.

I understand @cdjackson's use cases and I agree that allowing bindings to also provide historical values can be helpful. But using `HistoricState`s would nicely avoid above issues and express that this is data for specific use cases that should hence be also handled exceptionally.

J-N-K:
> The reason I am reluctant with adding a timestamp to the `State` itself (besides what I already wrote in #3033) is that State updates on the event bus are considered to be a time-series. That is: A continuous data stream which is unmodifiable and with a strict order by time - i.e. more recent events in the stream are always also newer datapoints.

Where can I find this limitation? There is nothing in the developer documentation that requires any order for events or guarantees an order. Anyway, we could work around that by only sending events over SSE that DO NOT implement `TimeboundState`. This way we could still use all the infrastructure we currently have for handling state updates.

kaikreuzer:
> Where can I find this limitation?

I don't think this is anywhere explicitely written, but it is an implicit assumption by the overall design. As mentioned, the UIs e.g. rely on the fact that the event stream is chronologically sorted, so does every persistence service, etc.

> we could work around that by only sending events over SSE that DO NOT implement `TimeboundState`

I don't think that this would be a clean solution. The source of truth are the events on the event bus, not the ones streamed through SSE. What we could do is to introduce a new event type of `TimeboundStateEvents`. This way we could make sure that existing logic never comes across them (as it might misinterpret them) and only new code that is specifically interested in them could handle them. I'm not sure, if it is worth the effort, though, or if option (3) wouldn't do the trick in a more lightweight fashion.

masipila:
Hi, 

just found this issue here.

Regarding the use cases for persisting future-timestamped values.

I have optimized our house heating and hot water heating to the cheapest hours of the day, which has saved me 400 € in August - December. The full solution documentation is available here:
https://community.openhab.org/t/control-a-water-heater-and-ground-source-heat-pump-based-on-cheap-hours-of-spot-priced-electricity/136566/13

I also opened this thread expressing my interest to contribute a more generic solution:
https://community.openhab.org/t/entsoe-e-binding-for-nordpool-spot-prices/143833/4

But many things in this comes back to the need to persist future-timestamped values. Spot prices of electricity is not the only thing needed here. Weather forecast, wind power production forecast and solar power production forecast are also needed to calculate tomorrow's (or a longer time periods) energy optimizations.

I also need the ability to persist values from the past.  My use case is that I do not own a real time energy consumption meter, but the hourly consumption data is available from the API of my electircity company. I can fetch yesterday's data next afternoon.

Cheers,
Markus

masipila:
Elaborating on the use case I mentioned in my previous post, which highlights the need for persisting future timestamps.

This is a screenshot of today's energy optimization, it was calculated yesterday.

image

The blue step-line represents the spot prices of each hour. As you can see, the prices vary between 0.2 c / kWh (at 04:00 at night) and 11.55 c/kWh at 11.00. So the price difference is about 60x between the cheapest hour and the most expensive hour. The data was published, fetched and persisted yesterday.

The green line represents the weather forecast (with wind chill compensation). This data was also fetched and persisted yesterday and this data was used to calculate how many hours I will need to heat our house.

The stacked bar chart represents the biggest energy consumption loads. Yellow is hour heat pump compressor (read: when then house heating is on), red is the water boiler and the blue is when the heat pump is pre-heating hot tap water which will be then fed to the boiler.

I can currently achieve this "optimize future energy consumption" because I bypass openHab in persisting the data. I would love to participate in generalizing this solution to the community. Doing this kind of energy consumption optimization requires the capability of persisting future-timestamped data.

Regarding the use case for persisting historical data. This chart illustrates our energy consumption vs. spot prices. The blue bars are our consumption and the green line are spot prices. My hourly consumption data of yesterday is available from the electricity company's API this afternoon so I need to persist it with historical timestamps. I'm also doing this currently bypassing the openHab and writing the data directly to the database from my script.

image

Cheers,
Markus

masipila:
Hi all,

I've read through the comments and I'm trying to summarize the current status of this issue so that we could conclude the next steps and remaining work. 

**Conceptual discussion regarding persisting historical data**
There were first comments debating if it should be possible to persist **historical** and/or **future** data. I interpret that we reached a consensus that there is a need to persist **historical** data. One use cases in the discussion was fetching the energy consumption data from a 3rd party and persisting it.

@cdjackson and and @J-N-K summarized it quite well and three options were proposed in comment
[#issuecomment-1225872970](#issuecomment-1225872970)

@cdjackson was in favor of option 1 but I interpret @kaikreuzer was more in favor of option 3.

**How about the time series with future timestamps?**
Can we please finalize the discussion if we should address the need to persist data with future timestamps as well? I'd like to reach conclusion if that will be out of scope this issue (at the moment the title of this PR limits the scope to historical data only) or should we address also this future data need at the same time? 

Example use cases have been presented in the comments above, one of them being tomorrow's day-ahead electricity prices so that it can be used to optimize tomorrow's energy consumption.

I believe the conceptual question can be formulated as follows: 
_If we allow bindings to persist data with timestamps in the past, are there any reasons why we would not allow binding to persist data with future timestamps as well?_

kaikreuzer:
Picking this up again after the recent discussions in #3478.

Thinking about it again, I'm indeed still clearly for option 3 and would think that we can keep it very simple, if we completely leave events out of the game. The thing handlers could get the possibility to post time-bound states for a channel through a new method in the callback. The callback can then process these calls and pass them on to the persistence manager to update the persisted data accordingly. Do I miss anything crucial here?

>  if we should address the need to persist data with future timestamps as well? 

I'd say it makes a lot of sense to include this in the scope as well - we would simply not care whether the timestamp is in the past or the future.
There is just one extra thing to think about, if we allow future values to be persisted: Imho we should make sure that an item state event is sent automatically, whenever such a future state becomes "present". By this we would make sure that the item state reflects what the persistent data claims the state is. As a result, a binding would more or less schedule future item state update events by writing future data to the persistence layer.

J-N-K:
I don't think it's possible to leave events out here, because events are the way the persistence manager gets the data it passes to the persistence services. Of course we can work around that by adding new event types that are only subscribed by the persistence manager and not by other event listeners, but I'm not sure this is a good idea. It would be easier and more straight forward if the timestamp was always associated with the state (and have an implicit timestamp "now" if none is set, which is the same we have now). The SSE event stream could also be modified to only pass events that are "now" (maybe with a window of +/- 1000ms so we don't remove too many events), so historic or future events are not sent. This keeps the order of events for that and is backward compatible.

Posting the (future) states at the time they were added for requires the persistence services to keep track of the next future event for each item and restore this map when the system starts up. This is of course possible, but requires implementation in the persistence services. Another way would be to leave this to the persistence manager which could query the "next state" from the services, similar to `restoreOnStartup` - this is also something to consider, the last persisted might not be the last state, if last persisted has a timestamp later than now.

splatch:
> Kai: (...) There is just one extra thing to think about, if we allow future values to be persisted: Imho we should make sure that an item state event is sent automatically, whenever such a future state becomes "present"

Persistence services during restore retrieve last state using chronological order. They will handle that case in consistent way, if our main driver is "latest by date", and not "closes to now" (closest ahead or in the past?).
Forcing persistence services to track and distinguish "now" vs "future" will complicate these services logic and result in situations where we will need to DB introduce polling or complex caching layer for each and every service which is out there. Given that these services tend to have completely different characteristics (from rrd4j up to dynamodb / influx), common denominator might not exist. I think tracking now vs future is something more suited to `EnergyManager` refereed in other issue.
Issue we discuss can be served with simple approach: future state should only take over item if its ahead of current time for current state (this is I think initial proposal from author?). There are very rare cases where forecast and present state are stored in same item, hence we do not need to care much about "now", because by definition future or past states provided by binding impact different items than "present".

Important question to bring is where we place code which propagates future state to item state? From stability and maintenance point of view, number of ways in which item can change its state should be limited. The more code paths leading to state change we have, the more surprises end users will face.

kaikreuzer:
> I don't think it's possible to leave events out here, because events are the way the persistence manager gets the data it passes to the persistence services.

The implementation in this PR actually shortcuts the persistence manager and directly reaches out directly to the modifiable persistence services. I'm not yet decided whether this is good or bad, but it raises the general question of whether we should consider persistence configurations for such "binding provided" time-bound states. What if the user has two modifiable persistence services configured, but only wants to use one of them for future prices?

> Another way would be to leave this to the persistence manager which could query the "next state" from the services, similar to restoreOnStartup - this is also something to consider, the last persisted might not be the last state, if last persisted has a timestamp later than now.

Yes, I would tend to leave this to the persistence manager and not add logic to the individual services for that.
The notion of "last" will indeed be carefully specified. The method `lastUpdate` should imho return the most recent state update in the past and not what was last written to the database (e.g. a future state).

> Important question to bring is where we place code which propagates future state to item state? 

As @J-N-K mentions, I'd see this logic to be very similar to the `restoreOnStartup` feature, so it would be the persistence manager's job.

J-N-K:
Bypassing the persistence manager is wrong and should not happen. This would require selecting the persistence service in the binding and is totally anti-pattern to our configuration handling.

This is also important because the persistence manager should be notified by the item, not by anyone else. We would re-introduce UoM issues if we allow someone else to provide values to persistence services (because they could have a different unit). This currently handled in the persistence service implementations which leads to different levels of UoM support in different services. With #3121 we'll have a clear way to ensure that always the same unit is persisted and restored. 

It would be much better to merge #2871 and add a `FutureTimestampFilter` and a `HistoricTimestampFilter` (or a `TimestampNowFilter`). 

IMO - if you prefer the methods over timestamps for states - then `GenericItem` should be extended with a `.setState(State state, Instant time)` method and that does not set item set state but creates a `TimedItemStateUpdatedEvent` and notifies listeners of the state change. This way we can still have the the benefits of the `ItemStateUpdatedEvent` (state as it would be if it was immediately set) and do not interfere with the state handling we have now. This would also work around the problems with the SSE streams because those don't receive the new event. Websocket listeners could decide if they want to subscribe or not.

rkoshak:
> Imho we should make sure that an item state event is sent automatically, whenever such a future state becomes "present". By this we would make sure that the item state reflects what the persistent data claims the state is. As a result, a binding would more or less schedule future item state update events by writing future data to the persistence layer.

What about the case where the actual didn't match the prediction? Or conditions alter before the actual time of the prediction and the new prediction is different?

As far as I'm aware, Persistence doesn't support a method to replace values but I think it would need to in order to support this.

> the last persisted might not be the last state, if last persisted has a timestamp later than now.

Sadly that just exacerbates an existing problem. `lastUpdate` almost never returns what was actually the last update to the Item unless the persistence strategy is carefully structured to do so.

J-N-K:
> > the last persisted might not be the last state, if last persisted has a timestamp later than now.
> Sadly that just exacerbates an existing problem. `lastUpdate` almost never returns what was actually the last update to the Item unless the persistence strategy is carefully structured to do so.

That's because we don't have timestamps for states. But it seems we can't get a consensus to introduce them,

masipila:
> What about the case where the actual didn't match the prediction? 

I have difficulties to think of a use case where the prediction and actual would be the same Item. If I would fetch forecasted temperature and have a thermometer, I would have two Items for these. One for the forecast and another for the actual.

> Or conditions alter before the actual time of the prediction and the new prediction is different?

If (when) the weather forecast changes so that tomorrow 12.00 is first 5 degrees and then the forecast gets updated that tomorro 12.00 is 5.5 degrees, I would expected that the latter takas presedence. That's how I currently do it.

rkoshak:
> I have difficulties to think of a use case where the prediction and actual would be the same Item.

Weather is one.

> One for the forecast and another for the actual.

True, but over the course of a day, at least where I am, even just the forecast min, max, and precipitation for the following day can change several times over the course of the day, sometimes by significant amounts. They change even more further out into the future is predicted. I live in one of the most difficult places in the world to predict the weather unfortunately.

> I would expected that the latter takas presedence.

Agreed, that's what *should* happen. But I don't see a mechanism discussed here to make that happen. Do we now have two entries in Persistence, one for 5.0 and another for 5.5? If so, how does it know the 5.5 takes precedence? How does that look in a chart or to the Persistence Actions? On-the-other-hand, if the 5.0. entry gets *replaced* by the 5.5, Persistence needs to be updated so it know when to do that and to do that. 

I just want to make sure this edge case isn't forgotten.

masipila:
> see a mechanism discussed here to make that happen. Do we now have two entries in Persistence, one for 5.0 and another for 5.5? If so, how does it know the 5.5 takes precedence? How does that look in a chart or to the Persistence Actions? On-the-other-hand, if the 5.0. entry gets replaced by the 5.5, Persistence needs to be updated so it know when to do that and to do that.

I don't see a problem overwriting the previous value with a more recent forecast.

masipila:
> I just want to make sure this edge case isn't forgotten.

Forgot to mention in my previous comment that  I full agree that this needs to be handled. And it's not even an edges case IMHO, it's completely normal that a forecast changes as time goes by.

splatch:
> Picking this up again after the recent discussions in #3478.
> 
> Thinking about it again, I'm indeed still clearly for option 3 and would think that we can keep it very simple, if we completely leave events out of the game. 

This opens up a question if profiles which are defined for channels linked to future or past items will be fired, and if they will be fired then how they will deal with reported timestamp?

J-N-K:
> > I just want to make sure this edge case isn't forgotten.
> 
> Forgot to mention in my previous comment that I full agree that this needs to be handled. And it's not even an edges case IMHO, it's completely normal that a forecast changes as time goes by.

I don't think we can replace/update values in RRD4J, in JDBC or InfluxDB it should be possible.

kaikreuzer:
> Bypassing the persistence manager is wrong and should not happen.

Ok, so let's let the persistence manager deal with this. My question from above still stays, though: Will we require to extend the persistence configuration to define how time-based updates should be dealt with? After all, a strategy like "everyMinute" simply does not have any meaning here.

>> Forgot to mention in my previous comment that I full agree that this needs to be handled. And it's not even an edges case IMHO, it's completely normal that a forecast changes as time goes by.

> I don't think we can replace/update/remove values in RRD4J, in JDBC or InfluxDB it should be possible.

We are clearly only talking about `ModifiablePersistenceService`s here. These support changing values, be it ones in the past or in the future. And yes, as @masipila already stated, updating (future) values will be a regular use case.

> I think the actual update to persistence should be done by the item, not by someone else. 

I agree that we have a lot of relevant logic on the items that we need to (re-)use, such as the UoM handling, the profiles, etc.
My main concern is that I don't think that the `CommunicationManager` is the the right place to deal with such updates. To me, the `CommunicationManager` (and also the events) are the "real-time view" of what is happening. They directly influence what the UIs are showing, they are triggering the rules, they are reflecting the here and now.
Changing historic or future values in a database is something substantially different to me - yes, we should allow a binding to do that (possibly by means of a new method in the handler callback), but similar to `PersistenceResource.putItemState()` this does not have to go through the `CommunicationManager`, but rather be handled by the persistence layer, i.e. the `PersistenceManager` directly.

Maybe we will end up doing it through the existing "communication" infrastructure after all, but I would regard this rather as a workaround because it is simpler to implement, but than a clean solution. So if we there is a chance to deal with it on the persistence side only, I would prefer this.

splatch:
> Changing historic or future values in a database is something substantially different to me - yes, we should allow a binding to do that (possibly by means of a new method in the handler callback), but similar to PersistenceResource.putItemState() this does not have to go through the CommunicationManager, but rather be handled by the persistence layer, i.e. the PersistenceManager directly.

The thing handler callback have already bunch of methods, adding one more does not impact it much (its already over bloated and does everything from configuration validation through thing status ending with channel updates).
What I’d be careful of is assumption that past or future states are reserved for persistence alone. Close time forecasts are most relevant for pre-heating/cooling, watering and other "plan" based situations which will often be based on rules (when forecast change trigger?).

Basic fact that REST resource does not use communication manager does not mean that it can't be used elsewhere. It is an internal component which is not used directly by any REST endpoint (even SSE which is closest by nature).

Reason why I was mentioning profiles is because they could be used i.e to calculate mean/average over a moving window, effectively delivering re-computed state from forecast and time

kaikreuzer:
> Reason why I was mentioning profiles is because they could be used i.e to calculate mean/average over a moving window, effectively delivering re-computed state from forecast and time

I do not see much value here as forecast values will very often come in bulk. If you process 240 events for updated values for the next 24 hours sequentially, it will not only cause huge unnecessary CPU utilization, but also can result in inadequate intermediate results, because the full picture simply wasn't there yet.
What is true is that rules might be interested to know when an update of future values might have happened; but this could be provided as an event (of a new type) by the persistence manager after receiving a bulk update.

splatch:
> I do not see much value here as forecast values will very often come in bulk. If you process 240 events for updated values for the next 24 hours sequentially, it will not only cause huge unnecessary CPU utilization, but also can result in inadequate intermediate results, because the full picture simply wasn't there yet.

@kaikreuzer I didn't mean ingestion of future/past states. What I meant is necessity to continuously poll database in order to re-calculate items which are derivative of these states. While rules have a cron trigger, using that mechanism to average i.e. future temperature predictions will cost extra. Doing that in profile which can keep last couple of states will be lighter than above. 
At present `PersistenceManager` does not post any events, in fact it is a listener for state change/update. I do see a risk of endless loop, however without code it is just a unjustified fear.

henfiber:
I am also interested in this for various use cases. I drafted this flowchart:

oh-historical-and-future-states

Some notes:

- Since Persistence services are not called imperatively but listen for events, these timestamps should be included in  events.
- Since the State is not always updated (for historical or future timestamps), we cannot reuse `*StateUpdated` events. I have introduced two new events (with the suffix `Received` instead of `Updated` since the State is not updated).
- Besides events, adding `lastUpdate` to the State would allow the framework to compare with the previous `lastUpdate` and take action accordingly (update if newer than previous one, and not in the future).
- The separate `HistoricalStateReceived` and `FutureStateReceived` events allow new Persistence strategies to be added with the same name. For instance, some users may opt to use a different database for future states or disable historical or future values entirely for some items where they would be considered erroneous.

Therefore, I am thinking that adding timestamps both to the State and Events may seem more work but it may help with ingestion and persistence of historical/future states. With the added benefit that it will make it easier (and more performant) for users to have access to lastUpdate in rules and the UI (with some extra work there) without reaching to persistence extensions or creating DateTime items.

(The mermaid diagram is [here](https://mermaid-js.github.io/mermaid-live-editor/edit#pako:eNq1VV1vmzAU_SuWH6pNSlNCCGlQmodlnfawrFPpNG1lDw7cJqhgItukyxD_vcY4lAgnedl4iMj9OseHc-UCh1kE2MNPSfYSrgkT6OFDQJF87iGEeAsL4Jys4DHA-SYiAnwhfwL8uy6aZ1SQmPKHOAUuSLop5iRJUKjDSOzjZV3__W3Gj1is_R0XkFbNcv50OavTiFd5b3q1nE2X7GqGVH1_S5Ic0A3SjOr_qqDOV1hfCBd6yA2i2cu79w3Vb8B4zMU8ZwyoUC2PErICX2Y5jVRE995uZUmDXw32kAmkhxQJr81wr8wawuc3WQK8p91IgmbGmQry4gJ166fNkcxqLviqAfwfenYYNdp-lsJmLA5Jolq1daJK3yMpo8QdgEbfA44a9FMucgYdQEP4H4Bp9wANwQe2jUMoWiGkY_K7mJYHXV7OuqtyZINkMfp6p1qObsuZZVLNBr-fQPx569csD2xrsnK7-oQBzzr0DMkOptbkMHGXRMCKzqesbT417ldpAlBz2ic7YtuTvZqhwYB7WxiO25ah5a6Te3Wyx-T_Y_W4h1NgKYkjeQUUVX-AxRqkwbAnXyPCngMc0FLWkVxk_o6G2HsiCYeevg4-xmTFSNpEN4Rir8B_sOfYfcuyrZE7sAfXtjsY9_AOe0OnP3Jtx50417YjE-NR2cN_s0xOsPoTd2gNXJkbOq41ccdq3C-VFCyX0yGKpR6L-spSN1f5CrEAWgk) for anyone wanting to make changes)

J-N-K:
@splatch Can't fix stupid. You can already create rules that run endlessly and also create two rules that always call each other. IMO applying a profile that does such things just makes no sense (while a profile that converts values or data types makes sense and is not critical).

@henfiber Conceptually it would make more sense to couple timestamp and state than timestamp and event, especially since we then would not need any additional methods. States without timestamps are considered to be "now" (like it is at the moment) and everyone not interested in future or past events can simply discard every event that has a timestamp. Items would only set states that have no timestamp.

That aside: We don't need new events. Since #3141 we differentiate between 

- item shall update: `ItemStateEvent`
- item has updated: `ItemStateUpdatedEvent`
- item has changed: `ItemStateChangedEvent`

so using `ItemStateEvent` would be no issue. The item can then decide if it uses the event to update its state (like it does now, if the state can be converted) or just passes to the listeners (like persistence services).

@kaikreuzer Bulk updates are another issue, we currently have no means to do so (although I see a benefit for that). This should really be a new event type but there is another issue: If we don't couple timestamps and states but timestamps and events we can't transport states for different timestamps in one event.

jlaur:
> > I don't think it's possible to leave events out here, because events are the way the persistence manager gets the data it passes to the persistence services.
> 
> The implementation in this PR actually shortcuts the persistence manager and directly reaches out directly to the modifiable persistence services. I'm not yet decided whether this is good or bad, but it raises the general question of whether we should consider persistence configurations for such "binding provided" time-bound states. What if the user has two modifiable persistence services configured, but only wants to use one of them for future prices?

I'm not sure if I understand this correctly, but shouldn't this work pretty much like now, i.e. the persistence service configured for item `price` (which is a future price if timestamp is in the future) will receive those time-bound states?

My doubt is if we are considering getting anything back from persistence, i.e. like **restoreOnStartup**? I would assume **restoreOnStartup** to be able to push back time-bound states as well, similar to what bindings would do.

But besides that, shouldn't we be able to handle time-bound states completely independent of persistence services? If a binding would update an item **not persisted** with a state one hour into the future, this should work. But what will happen in an hour, will there be an event that the (current) item value changed? If not, all bindings would have to implement this by themselves. And if yes, then how will it work? Will it be feasible to keep queued future states in memory, or do we need persistence?

The persistence we might need here I see as something **different** from the persistence the user has configured for items. This is persistence needed by the event bus to keep track of future state updates, and only to preserve memory. Also it should work for all items whether persisted or not by persistence services. When openHAB is restarted, we can flush all this. It may be repopulated by bindings or by persistence services when using **restoreOnStartup**, just like the current states are repopulated or restored today.

Sorry if I misunderstood something, just want to understand the thinking here and be sure we are not coupling existing persistence services too much with the need for the event bus to be updated when future states become present.

J-N-K:
A possible way would be that the item keeps track of future states and schedules its own update for the time that the state is for. When `restoreOnStartup` is used, we could then also restore the future states from persistence (which is easy, because it is a `HistoricItem` that already contains a timestamp).

masipila:
> A possible way would be that the item keeps track of future states and schedules its own update for the time that the state is for. When restoreOnStartup is used, we could then also restore the future states from persistence (which is easy, because it is a HistoricItem that already contains a timestamp).

This would be an ideal solution in my opinion. This would allow to use the same spot price Item to for example display "price right now" UI element. At the moment, I have an hourly Rule that updates the "spot price now" Item, which is a separate Item than the "spot price" Item where I persist the future timestamped prices.

> However, for calculations like "when is the best time to start the dish washer" we would still probably need persistence.

Yes,  I think so. At least I am very curious to do analysis of the "control points" vs. "actuals".

jlaur:
> A possible way would be that the item keeps track of future states and schedules its own update for the time that the state is for.

Exactly. This is similar to what I currently (have to) do manually in the EDS binding:
https://github.com/openhab/openhab-addons/pull/14376/files#diff-ba1eee4670519df1d485d8cc46217096018fb204bd597ea8d6237ad27ef457e7R492-R497

> When `restoreOnStartup` is used, we could then also restore the future states from persistence (which is easy, because it is a `HistoricItem` that already contains a timestamp).

Yes, so they will then also keep track of future states and schedule updates after being restored.

> However, for calculations like "when is the best time to start the dish washer" we would still probably need persistence.

Definitely. Comments above was meant for the event bus. Here we need access to persistence from the Energy Management Service.

henfiber:
@J-N-K it seems that many people have interest from a slightly different angle in this initiative, so I think you're right to be asking for some consensus before spending time on this. I will be happy to contribute (testing, updating docs, code if possible etc.) regardless of the decisions. That said I have some comments.

Let me summarize what I have understood as your current proposal:

- A timestamp will be coupled to state. We are going to have a new `TimeBoundState` (or similarly-named) object having both a `state` and a `dateTime` parameter. 
- The timestamp may also be in the future, so we will not use the name "Historic".
- No new event types will be created. `ItemStateEvent` will play the role of notifying consumers that a new State was received even if the Item state was not updated.

Please correct me if I have misunderstood something.

My questions/comments on that:

**1. Will a binding be able to send a timestamp with every state and make the items (indirectly) updated with this timestamp-state pair?**

From your previous comment, I understand that this won't be the case:
> States without timestamps are considered to be "now" (like it is at the moment) and everyone not interested in future or past events can simply discard every event that has a timestamp. Items would only set states that have no timestamp.

In my view, Items should also set states **that have a timestamp**. With the condition that the timestamp is **newer** than the previously received timestamp (stored in `TimeBoundState`) and is not in the future.

If Items only set states without a timestamp, then:

- A binding/rule that wants to send `N` timestamped measurements to an OpenHAB item for the last hour, will have to send `N-1` measurements with a timestamp, and the last one without a timestamp in order to set it as the current state.
  - This leads to inconsistency, since the N-1 measurements will be persisted with the Device time and the last measurement will be persisted with OpenHAB or Database server time.
- Or the binding can send all `N` timestamped measurements, and also send the latest (current) state without a timestamp so the Item state is also updated.
  - this will introduce an erroneous duplicate in the database.

So, it will not be possible to both send timestamped measurements and set the current state at once without introducing inconsistency or duplicates.

**2. Will the timestamp(s) be available to SSE consumers?**

If the state will only be coupled with the State and not with events, will SSE consumers have a way to receive the timestamp with the stream of state updates?

For a Java binding it may not be an issue to get the timestamp from the registry, but it may be an issue for SSE consumers like Habapp or the UI if they need to make an extra HTTP request to get the timestamp on every change.

I have read your concerns regarding event timestamps breaking the strict order in the SSE stream, but I don't see why OpenHAB should change the order. It should continue emitting the events in the incoming order. The timestamps would be just reported as an extra attribute. It is up to the SSE consumer to re-order the events (by introducing latency/buffering) if an application has such a requirement.

**3. Should we save the timestamp only when set by a binding?**

I would argue that it would be useful to store the timestamp even if a binding did not include one with the state. In this case, it would be the ingestion time, i.e. the system clock time ("now") when the new state was received.

With the suggested (and current) approach the ingestion time is not stored at all, so we have to use the consumer's time. Persistence services are one of the consumers and some of them use their own `now()` function (not even the consumer's **Java** time). 

An even greater issue is when someone opens the OpenHAB dashboard after some time or makes a REST API call to `/items`, there will be no way to get the information when the item state was updated. One would argue, you can get the timestamps from persistence, but there are cases where this will be sub-optimal at best. For instance, if one wants to make a "monitoring" page listing all of their items with the last update, they would have to make persistence queries to each one (may be 100's). The same for rules with a similar role that run on a schedule.

In contrast, if the timestamp was always saved (in the existing `State` object or elsewhere) it could be always made available to any interested consumer.

---

Now, all that said (and apologies for the long comment), even if this is implemented as suggested it will be an improvement to the existing situation. 

Just wanted to share some concerns in case others have similar use cases.

kaikreuzer:
For me it feels that this discussion goes into a fairly wrong direction and I do not agree with the summary of @henfiber, sorry.
Wrt @jlaur's question on whether or not we need persistence: The PR title says "Enable binding to store historic states" and its description mentions "logic to route [...] through the core to the persistence layer and any ModifiablePersistenceService". This is imho a fairly "simple" requirement. Changing this to "an item must not only have a state, but needs to keep its own history (both past and future) in memory" is imho a very different kind of beast and I would try to avoid opening such a can of worms. So my answer to your question is simple: Yes, we need a `ModifiablePersistenceService` in order to provide the feature.

We already had different options discussed [above](https://github.com/openhab/openhab-core/pull/3000#issuecomment-1225872970) and for various reasons, I [still prefer option 3](https://github.com/openhab/openhab-core/pull/3000#issuecomment-1500291644).

For the requirement that this PR wants to address, I do not see the need to have timestamps added to existing state update events. This would introduce many issues that were already mentioned above (like ordering, measurement time vs. processing time, the fact that persistence services persist states, but not events, bulk update handling, ...) and I believe this is a fully independent discussion that should not be in scope here.

Please also note that imho the requirement in this PR is also primarily about bulk updates of history. Be it missing history that is recovered and added as bulk from a cloud service, be it the prices for electricity for every hour of the next day, be it the weather forecast for the next 7 days. In all these use cases, single events do not make sense on their own, but what would be needed is an event that a certain historic (past or future) time range has just been updated - that's why I suggested a new event type that is issued by the persistence manager.

J-N-K:
I also don't think we need timestamps on events, I think we need to find a way to couple timestamps and states, and probably also (as @kaikreuzer correctly points out) also a way to group several of these "combined states" for bulk updates. Nothing of that requires timestamps for events and if I understood correct what @kaikreuzer is saying, I also agree that we should probably not re-use the existing events for that and I'm fine with using "option 3" if the item and persistence manager are not bypassed by these methods.

But I think we cannot separate the discussion here from the "full picture" (maybe it would have been better to start a discussion independent from the code here). Historic data and forecasts are quite similar in that they don't represent the current state of an entity but the state of that entity at a different point in time. Inserting them in persistence is only the half way for using them (it's fine for historic events, but not sufficient for forecasts). I'm not saying that the "second half" needs to be addressed in this PR (or even in the very near future), but we should have a picture of what the design should look like, so we don't introduce something now that we have to revert or break because we find out that we forgot something very important that we need to use the forecasts for our energy management system.

We already have a coupling of states and timestamps, the `HistoricItem` (which is a bit weird in naming when we talk about forecasts, but we can probably live with that. What we would need is a `HistoricItemFactory` (which we currently don't have) that the `BaseThingHandler` can use. We can then use a `postUpdate(ChannelUID, List<HistoricItem>)` method to inform the callback about these values. The callback should then route this list in the same direction like the state from `postUpdate(ChannelUID, State)`. The `CommunicationManager` should pass it to the item.  We can add a `default` method to the `StateProfile` that applies the profile for all values of the list before passing it to the item. A new method `setTimedStates(List<HistoricItem>)` is added to the `GenericItem` that in a first step only notifies listeners and by set puts the values to persistence (because the persistence manager is a listener to the items).

This is what should be covered here. In a follow-up PR we can add filters to persistence services so that different persistence series attached to the same item can be used for current states and the timed states and (if we agreed that it is desirable) an update/restore logic for future values for the item.

spacemanspiff2007:
> But I think we cannot separate the discussion here from the "full picture" (maybe it would have been better to start a discussion independent from the code here). Historic data and forecasts are quite similar in that they don't represent the current state of an entity but the state of that entity at a different point in time. Inserting them in persistence is only the half way for using them (it's fine for historic events, but not sufficient for forecasts). I'm not saying that the "second half" needs to be addressed in this PR (or even in the very near future), but we should have a picture of what the design should look like, so we don't introduce something now that we have to revert or break because we find out that we forgot something very important that we need to use the forecasts for our energy management system.

I think that's a good idea. 
Maybe take a step back and describe how everything should work from a feature/user perspective and then see what needs to be done.

It's trivial to have a binding generate a `timestamp -> value` table. But then what?
I'd expect this to go through the event bus somehow (and consequently the RestAPI consumers).
Then an item will receive this table. What will the state of the item be?
Will the state of the item automatically change according to the table as time moves on?
If persistence shall override already existing values there will be code changes needed for almost all persistence services.
How many values from the future will be restored for the item state? The whole table?
Can I feed the databases future values myself and have openHAB automatically restore them?
Will the item automatically restore from the future states in the database or will there be the need for a rule for every item?

Or will this still behave like a normal item and there will be a function to query the current value from persistence which can only be used in rules.

kaikreuzer:
>  Inserting them in persistence is only the half way for using them (it's fine for historic events, but not sufficient for forecasts)

Where do you see a difference here? (Btw, ChatGPT suggested to use the expression "time-series data" as it covers historic data and forecasted data equally well.)
My idea was to have the new "time-series update" event be published by the persistence manager and this could provide a time range within data has been changed. This could be past or future or even a range from past to future, so there wouldn't be a need to differentiate at all. Wrt potential use cases, I would also think that this should be what is needed to inform anyone interested. If we think in future that we additionally need the concrete updated values to be sent with the event as well, this is something that can be added at a later point and I do not see the risk that we would have to revert anything.

>  What we would need is a HistoricItemFactory (which we currently don't have) that the BaseThingHandler can use. 

To avoid the "weird" name, we might also think about a `TimeSeriesFactory`, which could create structs that directly relate many time-based values with a single item. The call `postUpdate(ChannelUID, TimeSeries)` would then be unambiguously refer to the (or even multiple) item(s) that are linked to the channel and not leave room for passing updates for different items in a single call (noting that the handler itself should not even know about the items that are linked to the channel).
The method `GenericItem.setTimedStates(List<HistoricItem>)` would then likewise become `GenericItem.setTimeSeries(TimeSeries)`.

In general, I like your proposal to pass it this way through the `CommunicationManager`, the `GenericItem` and through a listener to the `PersistenceManager` - it feels to me as if we are getting closer to a solution that shouldn't be too hard to implement and that fits into the existing architecture.

J-N-K:
`TimeSeries` sounds like a good name and I agree that it would be better to use something like that than a `List<Something>` because we can then change the internals whenever we like. We don't need a factory then, we can just use the `TimeSeries` constructor. The class could look similar to this (with proper `.equals` and `.toString` methods):

```
public class TimeSeries {
    private final List<Entry> states = new ArrayList<>();

    public void add(Instant timestamp, State state) {
        states.add(new Entry(timestamp, state);
    }
    
    public Stream<Entry> getStates() {
        return List.copyOf(states).stream();
    }

    public record Entry(Instant timestamp, State state) {
    }
}
```

WDYT?

Regarding the difference: for historic data we have a good coverage of everything we need in the `PersistenceExtensions`, we can retrieve and analyze data and the state with the latest timestamp is always the state that needs to be restored on startup. 

If we allow future state in the persistence, how we determine the state that we restore, if we want to automatically update the item when we reach the respective time. None of this would affect the persistence itself, but I would like to discuss as much of the design as possible before we add code, since there might be something that you or me don't think of at the moment which needs fundamental changes - even though I don't really expect that.

So my summary would be:

1. We add a new class `TimeSeries` that can be used to transport a list of timestamp/state tuples.
2. We route this via new methods in the `ThingHandlerCallback` and the `CommunicationManager` to the `GenericItem` which then informs its listeners after applying whatever it would do with a simple `State`.
3. We add a new `ItemTimeSeriesUpdatedEvent` (with similar content to the `ItemStateUpdatedEvent`, with converted units, ...).

Open questions:

1. Should we add a new type of listener (`TimeSeriesListener`), or do we just extend the `StateChangeListener`? I think a new listener would be better, because the `StateChangeListener` really deals with single states that are updated and changed.
2. Do we need another type of event (`ItemTimeSeriesEvent`) to allow the REST API or websockets to send a `TimeSeries`? I would say yes.
3. a) Which value shall be restored if there are historic and future values? I think the last value that is before now is correct. b) Shall this be handled by the persistence manager or by the individual persistence services?
4. Shall the item auto-update its state according to the persisted forecast? IMO yes, because using the same item for forecast and actual value makes no sense. 

The actual implementation of 4 can be discussed in the future. If the answer to 3b is "the persistence service" then it needs more changes in the services.

kaikreuzer:
> WDYT?

`TimeSeries` class looks good. 👍 

> how we determine the state that we restore, if we want to automatically update the item when we reach the respective time. 

I would see this as a responsibility of the `PersistenceManager` - it could query whether there are any future states stored for an item and do a query for the very first of them in order to keep this in memory for sending a state update event when the time has come. If new future states are stored, the `PersistenceManager` is the first to know and can update its internal memory map.

> So my summary would be

Sounds good to me.

> Open questions:

1. I'd also vote for a new listener interface.
2. Yes
3. a) Clearly always the latest before now. That's the whole idea about this persistence strategy.
  b) Not sure if there needs much to be changed about the current implementation? We can have a closer look and decide.
4. Yes. That's actually what I described above.

J-N-K:
Looks like we have the same picture now, nice. @altaroca Do you also agree and would you be willing to adapt the PR accordingly?

henfiber:
@J-N-K, @kaikreuzer I understand that the discussion regarding the implementation of this feature is nearing its final decision, and I respect the expertise and knowledge that you and the other core maintainers bring to the table.

However, I would like to kindly request a reconsideration of the concerns I raised in my last comment, as I believe they could lead to a more robust and effective solution. I will rephrase and shorten them a bit here: 

1. Current State should also be affected by a `postUpdate(ChannelUID, TimeSeries)` call (under conditions listed in my previous comment) if we want to have a consistent Persistence service without mixed clock sources, ignored timestamps or duplicates.
2. External Event Listeners may be interested in getting this new information as well.
3. Current State is not special; it is actually a specific Historic state which should have a unique recorded timestamp replicated in persistence.

If I should rank their relative importance I would go with **1 > 3 > 2**. So, we may ignore the 2nd one for now, i.e. attaching timestamps to events. The 3rd one can also be addressed separately although I think it is closely related to what we are discussing here.

Regarding the 1st corcern though, I have listed a common case which I believe cannot be handled effectively if the `GenericItem` ignores the `TimeSeries` payload and just passes it to Persistence.  So, I would greatly appreciate it if you could review this and check whether the proposed implementation addresses it. 

I apologize if I come across as overly persistent. I am genuinely interested in this feature, so I am trying to provide feedback in the spirit of what @J-N-K mentioned:

>  don't introduce something now that we have to revert or break because we find out that we forgot something very important

>  discuss as much of the design as possible before we add code, since there might be something that you or me don't think of at the moment which needs fundamental changes

J-N-K:
@henfiber 

1. What kind of state change would you expect if the data is only for the past? And what if they are only for the future. We already agreed on setting the item to the correct state when the time that is associated with the state has come.
2. What event listeners are you think of? With the `ItemTimeSeriesUpdatedEvent` the information is on the event bus. Whoever is interested, can subscribe to that event and should get the information.
3. I don't understand what you mean? We already set the state if it is the current state and we also persist it - if configured.

henfiber:
@J-N-K 

> 2. What event listeners are you think of? With the ItemTimeSeriesUpdatedEvent the information is on the event bus. Whoever is interested, can subscribe to that event and should get the information.

If `ItemTimeSeriesUpdatedEvent` will be made avaialble to external consumers (e.g. Habapp through SSE) I am already covered, that would be great.


> 1. What kind of state change would you expect if the data is only for the past? And what if they are only for the future. We already agreed on setting the item to the correct state when the time that is associated with the state has come.

I think you have covered the future states situation nicely.

What I am more concerned about is the "gray" area between what is considered "past" and what "current" state. For me "current" (virtual) state is the latest state we have received (with or without timestamps attached).

I will explain below my perspective about the issues arising when Items only accept states without timestamps.

## How historical state is usually retrieved

Devices or services that keep a local history of measurements, do not usually separate between "past" and "current" data. They forward, export or return the last measurements in one of the following ways:

- Return the `X` last measurements for the `<last Y period>` in batch mode (including the current/latest state, i.e. the latest sample)
- Queue up to `X` last measurements when communication becomes unavailable and send them in a FIFO manner when communication is restored. This includes the latest ("current") state.

I don't know of any device that sends only "past" data (i.e. excluding the current, latest state), unless it has an API where you can explicitly add a filter argument to do so.

## Specific example: OwnTracks

Let's look at a specific example, [OwnTracks](https://owntracks.org/booklet/), and see how it operates and what happens depending on how we handle historical states.

The OwnTracks [json payload](https://owntracks.org/booklet/tech/json/) includes a `tst` field which contains the UNIX epoch timestamp of the GPS fix. As mentioned [in the documentation](https://owntracks.org/booklet/tech/http/):

> In the event that the endpoint is unreachable, the payload will be queued and posted at a later time.
> we queue messages many thousands on the app until connectivity can be established

[Additionally](https://owntracks.org/booklet/features/android/):

> Outgoing messages may only be sent in batched intervals. Location updates or event messages may be delayed.

So,

- **OwnTracks queues messages and sends them in a FIFO manner when communication becomes available**.
- the GPS-fix timestamps are important since you cannot assume that a received location update occurred "now".


Let's see the sequence of events:

We have an OpenHAB item `Bob_Location` which is indirectly mapped through a binding to an OwnTracks device tracker.

OwnTracks detects a location update (at 2023-04-18 **13:38:25**) and queues the message due to a lack of connectivity:

```
{
  "_type": "location",
  "tid": "BOB",
  "lat": 48.85833,
  "lon": 2.29513,
  "tst": 1681825105
}
```

OwnTracks detects a second location update (at 2023-04-18 **13:39:25**):

```
{
  "_type": "location",
  "tid": "BOB",
  "lat": 48.85950,
  "lon": 2.29640,
  "tst": 1681825165
}
```

OwnTracks now has connectivity, so it first sends the previously queued message and then the latest one (FIFO).


**What should be the in-memory and persisted states of `Bob_Location` after these two messages?**

I would expect that:

- The `Bob_Location` Item has the state `48.85950, 2.29640` since this is the most recent location update.
- There are two new records in persistence with the timestamps and coordinates of the two messages.

oh-historical-persistence-from-bindings-desired-situation

---

**What will be the resulting OH state with currently accepted implementation plan?**

From my understanding:

> “Items would only set states that have no timestamp”.  

i.e. TimeSeries updates are only concerned with Past or Future states but not the current state.

So we end up with this situation:

oh-historical-persistence-from-bindings-proposed-no-state-update

The `Bob_Location` Item is not updated and does not match the latest state we received from OwnTracks.

---

It does not make sense to leave the `Bob_Location` item unchanged though (the location has changed), so an extra `postUpdate` call **without a timestamp** has to be made, which will lead to the following situation with duplicate persisted states and mixed clocks:

oh-historical-persistence-from-bindings-proposed-plus-extra-call

---

If we want to avoid the duplicates, then we somehow need to separate the last ("current") value from the "past" (queued) values. 

So we end up with this situation:

oh-historical-persistence-from-bindings-proposed-n-minus-one

Unfortunately, this is not possible for FIFO (non-batch) updates since the binding cannot know which update is the latest.  And is not ideal from a consistency perspective since we end up with mixed clock sources.

---

Finally, we can completely ignore the timestamps and only send `postUpdate` calls which defeats the purpose of this feature since we are coming back to the present situation:

oh-historical-persistence-from-bindings-current-situation

-----

Therefore, what I advocate for, is the situation in the first diagram which is accomplished by updating the state of the `Bob_Location` item, even from States with timestamps, when the `msg.timestamp > State.timestamp` (the condition is also shown in the flow chart in one of my previous comments) .  

Which seems a reasonable expectation to me since an OpenHAB item's state represents the most recent received state.



---

## Appendix: Devices that may operate in a similar manner to OwnTracks

1. Environmental monitoring sensors (temperature, humidity, air quality, soil moisture etc) which often operate on battery power and need to conserve energy. They may wake up periodically, take readings, timestamp them, and store the data in a local buffer. When communication is available, they send the buffered data in batch mode or one-by-one (FIFO).
2. Wearables: Devices like fitness trackers, smartwatches (heart rate, step count etc.) buffer the data with timestamps and synchronize with a cloud service or mobile device when communication is available, either through Bluetooth, Wi-Fi, or other means.
3. Asset tracking devices for vehicles or other objects which may store the location data along with timestamps in a local buffer and transmit the buffered data when in range of a communication network, such as cellular or Wi-Fi.
4. Some Smart home devices such as smart thermostats (e.g. the Nest Learning thermostat which keeps local history), Air Conditioners which keep consumed energy history (see [Panasonic Comfort Cloud binding](panasonic-comfort-cloud-binding/133848)), and Alarm panels which have local memory.
5. DIY sensors or wearables, such as a ESP32-based wearable device with a heart rate sensor (e.g., MAX30102) which can be programmed to operate in this way.
6. Mobile apps such as OwnTracks which I described above. In general, any moving IoT devices which
7. Other specific devices and services mentioned in the forum such as [GROHE ONDUS](https://www.grohe.de/de_de/smarthome/grohe-sense-guard/) and the E3DC binding which triggered this Pull Request.


----

Regarding (3)

> Current State is not special; it is actually a specific Historic state which should have a unique recorded timestamp replicated in persistence.
> I don't understand what you mean?  We already set the state if it is the current state and we also persist it - if configured.

I will try to explain it better in a subsequent comment.

---

Apologies again for my long comment, hopefully you will find it helpful.

J-N-K:
I don't think this is an issue. The `PersistenceManager` can check if the timestamp of the persisted value closest to now (in the past) is newer or older than the an element in the time series and issue an `ItemStateEvent` we only have to find a way to veto persistence for this value, so it isn't duplicated in the persisted data.

masipila:
Hi all,

I thought I'd drop a note to others in this thread that I'm most likely not able to contribute too much on making the energy optimizations after #3597 lands even though I was earlier showing strong interest to contribute and generalize these things to be as easy for non-developer background users as possible.

We got fantastic news a couple of weeks ago that the Finnish Olympic Committee will support our curling team next winter season on our run to Milano Cortina 2026. This will mean that I will work less to be able to curl more, which means that when I'm home I want to spend that time with the family.

Cheers,
Markus

kaikreuzer:
Hey @masipila, That's an awesome opportunity for you! All the best for your preparations and whenever you feel you need a break from curling and dive into some code instead, please think of us!

J-N-K:
Superseded by #3597

PR #3001, opened 2022-06-15
Exclude org.osgi.service.prefs to fix build

wborn:
Someone added a dependency on a non-existing artifact in the new Equinox release and now many builds are broken because Xtext defines a version range for Equinox :disappointed:, see:

* https://github.com/eclipse-equinox/equinox.bundles/issues/54
* https://github.com/eclipse/xtext/issues/2077

### DIFF
--- a/bom/compile-model/pom.xml
+++ b/bom/compile-model/pom.xml
@@ -38,6 +38,11 @@
           <groupId>org.eclipse.platform</groupId>
           <artifactId>org.eclipse.osgi</artifactId>
         </exclusion>
+        <exclusion>
+          <!-- See: https://github.com/eclipse-equinox/equinox.bundles/issues/54 -->
+          <groupId>org.osgi.service</groupId>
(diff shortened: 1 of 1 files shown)


J-N-K:
Did you check if this is still needed after they deployed a new version an hour ago?

wborn:
Yes see this recent build:

https://ci.openhab.org/job/openHAB-Core/1014/

It will probably try to keep resolving it until it is excluded in Xtext.
It's another good example why using Maven dependency version ranges is not a good idea.

J-N-K:
It still resolves 3.10.0, while the new artifact has 3.10.1, maybe this issue only persists until the maven cache expires and 3.10.1 is included? But I'm fine with this solution, too. We can revert it later if we run into issues.

J-N-K (approved):

J-N-K:
It seems we can revert the change in the near future: https://issues.sonatype.org/browse/OSSRH-81766

PR #3007, opened 2022-06-19
Log a warning when item for trigger/condition is missing or removed

### LINKED ISSUE #2526: [Rule DSL] log a WARN if item does not exist
In reference to [this](https://community.openhab.org/t/log-non-existing-item-in-rule-condition/127652/3), would it be possible to add a WARN log if a referenced item in a rule dsl file condition does not exist? Would be really helpful when debugging, thanks :D!

J-N-K:
Closes #2526 

The solution here not only works for DSL but for all rules since the check is done in the module handler.

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/GroupCommandTriggerHandler.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/GroupCommandTriggerHandler.java
@@ -28,7 +28,9 @@ import org.openhab.core.events.EventFilter;
 import org.openhab.core.events.EventSubscriber;
 import org.openhab.core.items.Item;
 import org.openhab.core.items.ItemRegistry;
+import org.openhab.core.items.events.ItemAddedEvent;
 import org.openhab.core.items.events.ItemCommandEvent;
+import org.openhab.core.items.events.ItemRemovedEvent;
 import org.openhab.core.types.Command;
 import org.osgi.framework.BundleContext;
 import org.osgi.framework.ServiceRegistration;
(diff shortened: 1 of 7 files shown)


kaikreuzer:
The discussion linked in #2526 suggests that this should be addressed through the VSCode extension as it seems risky to add it to the runtime due to potential issues with the lifecycle of rules and items, especially when editing them at runtime.
Did you consider those concerns here?

J-N-K:
That‘s why I added it here and subscribed to the ItemAddedEvent/ItemRemovedEvent.That way runtime changes to items will be properly considered. And it also works for UI defined rules and all other languages, IMO it‘s superior to VSCode.

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/GroupCommandTriggerHandler.java
```diff
@@ -59,21 +61,29 @@ public class GroupCommandTriggerHandler extends BaseTriggerModuleHandler impleme
...
+        this.ruleUID = ruleUID;
         Dictionary<String, Object> properties = new Hashtable<>();
-        this.topic = "openhab/items/";
+        this.topic = "openhab/items/*";
         properties.put("event.topics", topic);
         eventSubscriberRegistration = this.bundleContext.registerService(EventSubscriber.class.getName(), this,
                 properties);
+
+        if (itemRegistry.get(groupName) == null) {
+            logger.warn("Group '{}' needed for rule '{}', trigger '{}' not present in registry. Trigger will not work.",
```
kaikreuzer, line 84: I think the message about the trigger is incorrect/misleading. The trigger is in the registry, the item is not.
I'd suggest to change it to something like, which is anyhow shorter and has all relevant information:
```suggestion
            logger.warn("Group '{}' needed for rule '{}', Trigger '{}' will not work.",
```
Same of course for all the other log messages.

kaikreuzer (changes requested):
> That way runtime changes to items will be properly considered.

I am not sure how this addresses potential lifecycle issues, but I assume you have tested this, so I'm fine here.

So just my wish to slightly improve the log messages remains, then we should be good to merge.

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/GroupCommandTriggerHandler.java
J-N-K: Done.

kaikreuzer (approved):
Thank you!

PR #3008, opened 2022-06-20
Fix null analysis error in ThingMigrationOSGiTest

wborn:
In Eclipse the `storageService` would be inferred as nullable which conflicts with the `ManagedThingProvider` non-null constructor argument.


J-N-K (approved):
Thanks. It‘s interesting to see what differences in null-analysis happen between different IDEs.

wborn:
Yes there is still some room for improving the compiler arguments. I think it does show a warning for this on command line builds (IntelliJ). The OH Eclipse config has it [configured as error](https://github.com/openhab/openhab-distro/blob/main/launch/openHAB2.setup#L120-L123). It can be turned into an error by removing [+nullAnnotConflict](https://github.com/openhab/openhab-core/blob/c2ec97cfaf396c50edb7f3c54a4f2edee6c0705c/pom.xml#L286) in the compiler args. However that results in some more errors on CLI builds that don't show in Eclipse. So it requires a bit more tweaking to get it right everywhere.

PR #3009, opened 2022-06-20
Remove redundant `@NonNullByDefault`

wborn:
Nullness default is redundant with a default specified for the enclosing type AutomationIntegrationJsonTest.


J-N-K (approved):
Thanks.

PR #3010, opened 2022-06-21
Fix config validation for integer values

J-N-K:
Reported on the forum. The problem is that on deserialization 255 is deserialized to a `BigDecimal` without scale. The validation the fails because the `BigDecimal` is converted to a `255.0` string, which is not equal to the `255` string from the config description option. The value is now properly normalized.

### DIFF
--- a/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/MinMaxValidator.java
+++ b/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/MinMaxValidator.java
@@ -39,11 +39,12 @@ final class MinMaxValidator implements ConfigDescriptionParameterValidator {
             return null;
         }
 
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/MinMaxValidator.java
```diff
@@ -37,9 +39,11 @@ final class MinMaxValidator implements ConfigDescriptionParameterValidator {
             return null;
         }
 
-        // Allow specified options to be outside of the min/max value
+        Object normalizedValue = ConfigUtil.normalizeType(value, parameter).toString();
```
maniac103, line 42: It technically doesn't matter, but make this `String normalizedValue = ...` for clarity?

wborn (approved):
Thanks for the fix!

PR #3012, opened 2022-06-22
Fix grep console not working

### LINKED ISSUE #3011: Piped grep not working anymore in karaf (3.3.0-SNAPSHOT - Build #2963)
I just installed the above snapshot and issuing a 

 openhab:things list | grep OFFLINE  

returns all things, not a filtered list, as if the piped command is not executed

J-N-K:
Fixes #3011 

I currently can't test if #1545 is back with this solution. The `PrintStream` is now determined in the same way as grep does and that at least makes grep work locally and for SSH connections. Maybe @robnielsen can test with the Insteon extension once it is merged.


wborn (approved):
Thanks!

PR #3016, opened 2022-06-26
Fix ConcurrentModificationException in WatchQueueReader

kaikreuzer:
This addresses https://community.openhab.org/t/openhab-3-3-milestone-discussion/132715/220

@openhab/core-maintainers: I don't know how likely this race condition is to happen - please decide whether you consider it critical and merge it for 3.3 or leave it for later.


J-N-K:
I believe it's not very likely, considering that we have not seen reports about that in the past (and the code didn't change for some time). But changing `HashMap` to `ConcurrentHashMap` is also a very low risk modification.

J-N-K (approved):

jlaur:
@J-N-K - should milestone be set to 3.3?

PR #3022, opened 2022-06-28
Improve UID error message

J-N-K:
https://community.openhab.org/t/openhab-3-3-release-discussion/136925/18

The segments are now integrated in the message so it's easier to determine which UID was wrong.


wborn (approved):
This will help!

PR #3024, opened 2022-07-01
Normalize thing configuration before initialization

J-N-K:
There were some reports on the forum that de-serialized values (and probably other) do not have the correct type. We already normalize thing configurations when they are set via REST API, but it seems that under some circumstances integer values are not properly de-serialized. This PR adds normalization of the configuration before the thing is initialized. 

This is not the best solution, but unfortunately it'll get very complex otherwise. The best would be to normalize the configuration immediately when it is read from the database, but that can't be done because the config description needed for that is not available at the time the thing registry is first filled. Holding back things in the `ManagedProvider` would require keeping track of the config descriptions and thus duplicating what is already in the `ThingManagerImpl`.

Doing it before the initialization might also mitigate problems with integer/string values in textual configuration.


kaikreuzer (approved):
Thanks for this fix.
It is difficult to judge whether there could be any side effects, but let's merge and we will find out.

PR #3026, opened 2022-07-03
Remove OH2 leftover in Karaf Add-On service

J-N-K:
OH1 add-ons are not present in the OH3 distribution as they are not supported anyway, so we do not to filter them.


wborn (approved):
Thanks for cleaning it up!

PR #3027, opened 2022-07-03
Add missing & refactor existing button toggle profiles

pfink:
Signed-off-by: Patrick Fink <mail@pfink.de>

This pull requests does the following:

**Functionality improvements**

* Missing toggle profiles for trigger channel SYSTEM.BUTTON are added
* Existing togle profiles for SYSTEM.RAWBUTTON do get the new parameter `event` where the trigger event could be changed. Example (this makes profile `rawbutton-toggle-switch` use the `RELEASED` event as a trigger instead of the default `PRESSED`):
```
Switch MyItem { channel="<bindingID>:<thing-typeID>:MyThing:myChannel" [profile="rawbutton-toggle-switch", event="RELEASED"]}
```
 * The parameter 'event' is available for all button profiles, it's content is validated, logging throws a warning if the user sets an event that is not valid for the channel. It's an optional parameter that defaults to `SHORT_PRESSED` for SYSTEM.BUTTON and to `PRESSED` for SYSTEM.RAWBUTTON. Therefore, the PR is fully backward-compatible.

**Toggle Profile Refactoring**

Instead of having one class per profile, I created the new class `ToggleProfile` which can handle all 6 profiles ("button-toggle-switch", "button-toggle-player", "button-toggle-rollershutter", "rawbutton-toggle-player" "rawbutton-toggle-rollershutter" and "rawbutton-toggle-switch"). Therefore, the existing rawbutton profile classed would be obsolete and are deleted in this PR. This avoids redundant code and (hopefully) improves maintainability.

**Known Issues / Limitations**

* The PR is in proposal state. I'd first like to get a general feedback if maintainers agree with the approach I took (using 1 class for all 6 profiles instead of 6 classes) before I put more effort into it. I didn't look into style issues and just tested a small subset of use cases yet. So it's not ready for merge.
* The constructor of `ToggleProfile` is quite long. I tried it with builder pattern as an alternative, but this needs quite a lot code and the handling of the Eclipse null annotations would be quite complicated. Other ideas to manage this are welcome.
* RawButtonToggleProfileTest is currently failing. The ProfileContext still has to be mocked properly.

@cweitkamp As you wrote some of the rawbutton profiles, I think it would be good if you could have a look on this.


bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/profiles/ToggleProfile.java
```diff
@@ -0,0 +1,94 @@
...
+        if (triggerEvent.equals(event)) {
+            StateType newState = initialState.equals(previousState) ? alternativeState : initialState;
+            callback.sendCommand(newState);
+            previousState = newState;
+        }
+    }
+
+    @Override
+    public void onStateUpdateFromItem(State state) {
+        previousState = state.as(OnOffType.class);
```
J-N-K, line 92: This looks wrong. Previously the RollerShutter returned `UpDownType` here.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/profiles/SystemProfileFactory.java
```diff
@@ -103,14 +107,32 @@ public SystemProfileFactory(final @Reference ChannelTypeRegistry channelTypeRegi
             return new SystemOffsetProfile(callback, context);
         } else if (RANGE.equals(profileTypeUID)) {
             return new SystemRangeStateProfile(callback, context);
+        } else if (BUTTON_TOGGLE_SWITCH.equals(profileTypeUID)) {
+            return new ToggleProfile<OnOffType>(callback, context, BUTTON_TOGGLE_SWITCH,
+                    DefaultSystemChannelTypeProvider.SYSTEM_BUTTON, OnOffType.ON, OnOffType.OFF,
+                    CommonTriggerEvents.SHORT_PRESSED);
```
J-N-K, line 113: In the case of a raw button there are only `PRESSED` and `RELEASED`,  but here we have `SHORT_PRESSED`, `LONG_PRESSED` and `RELEASED`. Would it be possible to make it configurable if short or long events are used to toggle the state?

J-N-K (changes requested):
In general I'm all for re-using code and this seems to be a good choice. I have left two comments.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/profiles/SystemProfileFactory.java
pfink: It is already configurable. The event passed to the constructor is just the default event that is being used if the user does not configure it explicitely.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/profiles/ToggleProfile.java
pfink: You're right, that was a bug. It's fixed now!

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/profiles/ToggleProfile.java
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * This profile reads the triggered events and uses the item's current state
+ * to toggle it. Being configurable via the constructor, {@link ProfileFactory}
+ * can use it for various system channel types and item types.
+ *
+ * @author Patrick Fink - Initial contribution
+ */
+@NonNullByDefault
+public class ToggleProfile<StateType extends State & Command> implements TriggerProfile {
```
J-N-K, line 36: We mostly use `T` as naming for generics. I would suggest to change it to that.

J-N-K (changes requested):
It would be great if you could also add tests for the other cases (i.e. `UpDownType` and `PlayPauseType`. Also a test that shows that non-trigger events are ignored would be great.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/profiles/ToggleProfile.java
```diff
@@ -0,0 +1,94 @@
...
+    public ToggleProfile(ProfileCallback callback, ProfileContext context, ProfileTypeUID uid, ChannelType channelType,
+            StateType initialState, StateType alternativeState, String defaultEvent) {
+        this.uid = uid;
+        this.channelType = channelType;
+        this.initialState = initialState;
+        this.alternativeState = alternativeState;
+        this.callback = callback;
+
+        String triggerEventParam = (String) context.getConfiguration().get(EVENT_PARAM);
+        if (isValidEvent(triggerEventParam)) {
```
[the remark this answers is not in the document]
pfink: This is already ensured by line 63.

bundles/org.openhab.core.thing/src/test/java/org/openhab/core/thing/internal/profiles/RawButtonToggleSwitchProfileTest.java
```diff
@@ -39,18 +42,19 @@
 public class RawButtonToggleSwitchProfileTest {
```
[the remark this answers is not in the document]
pfink: done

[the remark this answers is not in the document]
pfink: done

bundles/org.openhab.core.thing/src/test/java/org/openhab/core/thing/internal/profiles/ToggleProfileTest.java
```diff
@@ -0,0 +1,188 @@
...
+        Map<String, Object> params = triggerEvent == null ? Collections.emptyMap()
+                : Collections.singletonMap(ToggleProfile.EVENT_PARAM, triggerEvent);
+        when(contextMock.getConfiguration()).thenReturn(new Configuration(params));
+    }
+
+    private void prepareContextMock() {
+        initializeContextMock(null);
+    }
+
+    private ListAppender<ILoggingEvent> createLogAppender() {
```
J-N-K, line 160: There is a method for that in `JavaTest`: `setupInterceptedLogger(Class<?> clazz, LogLevel minLogLevel)` and also assertions for no log (`assertNoLogMessage`) or log messages (`assertLogMessage(Class<?> clazz, LogLevel logLevel, String message)`.

cweitkamp (approved):
LGTM. Thanks for the improvement and reducing the code in such an elegant way. 👍🏼

bundles/org.openhab.core.thing/src/test/java/org/openhab/core/thing/internal/profiles/ToggleProfileTest.java
pfink: done

J-N-K (approved):

PR #3030, opened 2022-07-06
Add a profile for converting Non-Undef/Null to ON

ssalonen:
This is a profile to convert non-UNDEF non-NULL values from {@link ThingHandler} as ON/OFF.

State updates from `ThingHandler` that are UNDEF or NULL are converted to OFF, while other values are converted to ON.
- Commands (never UNDEF/NULL) from `ThingHandler` towards items are always converted to ON.
- Commands from items are forwarded to the `ThingHandler` as-is.

Docs updated in https://github.com/openhab/openhab-docs/pull/1856

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/profiles/SystemNonUndefProfile.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/profiles/SystemNonUndefProfile.java
@@ -1,0 +1,81 @@
+/**
+ * Copyright (c) 2010-2022 Contributors to the openHAB project
+ *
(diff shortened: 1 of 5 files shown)


J-N-K:
Isn't that the same that can be achieved by the MAP profile?

ssalonen:
I guess you are right. For all practical purposes we could have same effect using MAP with

NULL=OFF
UNDEF=OFF
=ON

I really did not think of this, perhaps since I did not realize that MAP supports default value as well.

ssalonen:
I do not think this code "carries its own weight" when there is a MAP alternative present.

Users benefit more from learning the MAP transformation since that concept can be applied in different situations, not just in this specific case.

I will close this PR.

PR #2966, opened 2022-05-15
Allow generation of default translations for automation modules

J-N-K:
Closes #2839 

Signed-off-by: Jan N. Klug <github@klug.nrw>

PR #3035, opened 2022-07-09
Add default translations for automation module types

wborn:
These were generated using openhab/openhab-core#2966.

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/resources/OH-INF/i18n/automation.properties
+++ b/bundles/org.openhab.core.automation/src/main/resources/OH-INF/i18n/automation.properties
@@ -1,0 +1,405 @@
+# core.ChannelEventTrigger
+
+module-type.core.ChannelEventTrigger.label = a trigger channel fires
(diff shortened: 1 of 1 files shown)


J-N-K (approved):
Thanks!

Nadahar:
As a translator, I've started to look at these. There are quite a few instances of the word `item` for example. There are "special rules" for Item, Thing, Sitemap etc, where theses should not be translated and always be written capitalized. Is it safe to assume that when I come across `item` that it doesn't mean "OH Item" but the generic word "item", and that it should be translated?

J-N-K:
No, there are also cases where "item" in lower-case should not be translated. In fact, in the translations added here, "item" should never be translated and always refers to the openHAB item.

Nadahar:
I think that is a potential source of confusion. I'm pretty sure that I read somewhere that they should always be capitalized, but in addition to that only the capitalized versions are in the "Crowdin glossary" that triggers a dotted underline under the words which states that they should not be translated.

PR #3036, opened 2022-07-10
Enhance transformation configuration

J-N-K:
This is needed to give the users the best experience when editing transformations in UI.

### DIFF
--- a/bundles/org.openhab.core.io.rest.transform/src/main/java/org/openhab/core/io/rest/transform/TransformationConfigurationDTO.java
+++ b/bundles/org.openhab.core.io.rest.transform/src/main/java/org/openhab/core/io/rest/transform/TransformationConfigurationDTO.java
@@ -26,6 +26,7 @@ public class TransformationConfigurationDTO {
     public String uid;
     public String label;
     public String type;
+    public String context;
     public @Nullable String language;
     public String content;
     public boolean editable = false;
@@ -34,6 +35,7 @@ public class TransformationConfigurationDTO {
         this.uid = transformationConfiguration.getUID();
         this.label = transformationConfiguration.getLabel();
         this.type = transformationConfiguration.getType();
+        this.context = transformationConfiguration.getContext();
         this.content = transformationConfiguration.getContent();
         this.language = transformationConfiguration.getLanguage();
     }
--- a/bundles/org.openhab.core.io.rest.transform/src/main/java/org/openhab/core/io/rest/transform/TransformationConfigurationDTO.java
+++ b/bundles/org.openhab.core.io.rest.transform/src/main/java/org/openhab/core/io/rest/transform/TransformationConfigurationDTO.java
@@ -16,10 +16,10 @@ import java.util.Map;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
-import org.openhab.core.transform.TransformationConfiguration;
+import org.openhab.core.transform.Transformation;
 
 /**
- * The {@link TransformationConfigurationDTO} wraps a {@link TransformationConfiguration}
+ * The {@link TransformationConfigurationDTO} wraps a {@link Transformation}
  *
  * @author Jan N. Klug - Initial contribution
  */
@@ -32,11 +32,11 @@ public class TransformationConfigurationDTO {
     public Map<String, String> configuration;
     public boolean editable = false;
 
-    public TransformationConfigurationDTO(TransformationConfiguration transformationConfiguration) {
-        this.uid = transformationConfiguration.getUID();
-        this.label = transformationConfiguration.getLabel();
-        this.type = transformationConfiguration.getType();
-        this.language = transformationConfiguration.getLanguage();
-        this.configuration = transformationConfiguration.getConfiguration();
+    public TransformationConfigurationDTO(Transformation transformation) {
+        this.uid = transformation.getUID();
+        this.label = transformation.getLabel();
+        this.type = transformation.getType();
+        this.language = transformation.getLanguage();
+        this.configuration = transformation.getConfiguration();
     }
 }
--- a/bundles/org.openhab.core.io.rest.transform/src/main/java/org/openhab/core/io/rest/transform/TransformationConfigurationDTO.java
+++ b/bundles/org.openhab.core.io.rest.transform/src/main/java/org/openhab/core/io/rest/transform/TransformationConfigurationDTO.java
@@ -12,6 +12,8 @@
  */
 package org.openhab.core.io.rest.transform;
 
+import java.util.Map;
+
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.transform.TransformationConfiguration;
@@ -26,17 +28,15 @@ public class TransformationConfigurationDTO {
     public String uid;
     public String label;
     public String type;
-    public String context;
     public @Nullable String language;
-    public String content;
+    public Map<String, String> configuration;
     public boolean editable = false;
 
     public TransformationConfigurationDTO(TransformationConfiguration transformationConfiguration) {
         this.uid = transformationConfiguration.getUID();
         this.label = transformationConfiguration.getLabel();
         this.type = transformationConfiguration.getType();
-        this.context = transformationConfiguration.getContext();
-        this.content = transformationConfiguration.getContent();
         this.language = transformationConfiguration.getLanguage();
+        this.configuration = transformationConfiguration.getConfiguration();
     }
 }
--- a/bundles/org.openhab.core.transform/src/main/java/org/openhab/core/transform/Transformation.java
+++ b/bundles/org.openhab.core.transform/src/main/java/org/openhab/core/transform/Transformation.java
@@ -20,12 +20,12 @@ import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.common.registry.Identifiable;
 
 /**
- * The {@link TransformationConfiguration} encapsulates a transformation configuration
+ * The {@link Transformation} encapsulates a transformation configuration
  *
  * @author Jan N. Klug - Initial contribution
  */
 @NonNullByDefault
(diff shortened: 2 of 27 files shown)


bundles/org.openhab.core.transform/src/main/java/org/openhab/core/transform/TransformationConfiguration.java
```diff
@@ -28,21 +28,24 @@ public class TransformationConfiguration implements Identifiable<String> {
...
     private final String type;
+    private final @Nullable String context; // for backward deserialization compatibility
     private final @Nullable String language;
     private final String content;
 
     /**
      * @param uid the configuration UID. The format is config:&lt;type&gt;:&lt;name&gt;[:&lt;locale&gt;]. For backward
      *            compatibility also filenames are allowed.
      * @param type the type of the configuration (file extension for file-based providers)
+     * @param context context (e.g. script type or blockly blocks) of the configuration as JSON string
```
kaikreuzer, line 39: I'm not a big fan of expecting JSON format in Java String parameters. It especially is totally unclear what kind of values are expected to be passed here and how they are potentially used. Why isn't the transformation type not enough? It is supposed to describe the transfomation and with it the expected configuration of it.

bundles/org.openhab.core.transform/src/test/java/org/openhab/core/transform/internal/TransformationConfigurationRegistryImplTest.java
```diff
@@ -47,22 +47,22 @@ public class TransformationConfigurationRegistryImplTest {
     private static final String MANAGED_WITH_DE_LANGUAGE_UID = "config:" + SERVICE + ":managed:de";
 
     private static final TransformationConfiguration MANAGED_WITHOUT_LANGUAGE = new TransformationConfiguration(
-            MANAGED_WITHOUT_LANGUAGE_UID, "", SERVICE, null, MANAGED_WITHOUT_LANGUAGE_UID);
+            MANAGED_WITHOUT_LANGUAGE_UID, "", SERVICE, SERVICE, null, MANAGED_WITHOUT_LANGUAGE_UID);
```
kaikreuzer, line 50: This shows what I mention above: As an API user, one does not expect to pass JSON and instead passes some other stuff, like duplicated and thus useless information like here.

kaikreuzer (changes requested):

bundles/org.openhab.core.transform/src/test/java/org/openhab/core/transform/internal/TransformationConfigurationRegistryImplTest.java
J-N-K: That's just a leftover from my first attempt (a "subType" which contained either additional information or the duplicated information). But this turned out not to be sufficient for blocky.

bundles/org.openhab.core.transform/src/main/java/org/openhab/core/transform/TransformationConfiguration.java
J-N-K: There are two reasons: 

- For good UI integration we need to give a hint to the UI which editor to use. For "map" or "scale" we can also integrate that into the UI but e.g. the "script" transformation needs different editors for blocky, JS or Python and there is no way to auto-detect the correct scripting language. 
- For blocky "scripts" we need to store additional data (the blockly "code") to edit them later. 

If you have a good idea how to store arbitrary data other than an encoded JSON, I'm happy to implement that.

J-N-K:
Please have a look at the images in the linked PR in webui to see what I want to achieve. I don't know why they vanished the first time.

bundles/org.openhab.core.transform/src/main/java/org/openhab/core/transform/TransformationRegistry.java
```diff
@@ -21,16 +21,16 @@
...
 
 /**
- * The {@link TransformationConfigurationRegistry} is the interface for the transformation configuration registry
+ * The {@link TransformationRegistry} is the interface for the transformation registry
  *
  * @author Jan N. Klug - Initial contribution
  */
 @NonNullByDefault
-public interface TransformationConfigurationRegistry extends Registry<TransformationConfiguration, String> {
+public interface TransformationRegistry extends Registry<TransformationConfiguration, String> {
```
kaikreuzer, line 29: The `TransformationRegistry` should (by the definition of its name) be a registry for `Transformations`, not `TransformationConfigurations`. Does anything speak against refactoring `TransformationConfiguration` to `Transformation`? I know that we have already the action with that name, but as this is in a different (and internal) package, this should hopefully not block us.

kaikreuzer (changes requested):

bundles/org.openhab.core.transform/src/main/java/org/openhab/core/transform/TransformationRegistry.java
J-N-K: I agree. My idea was to keep it backward compatible, but with renaming the `TransformationConfigurationRegistry` to `TransformationRegistry` we need to adapt the transformation add-ons that use it in openhab-addons anyway, so I'll change it.

bundles/org.openhab.core.io.rest.transform/src/main/java/org/openhab/core/io/rest/transform/TransformationConfigurationDTO.java
```diff
@@ -27,14 +29,14 @@ public class TransformationConfigurationDTO {
...
+    public Map<String, String> configuration;
     public boolean editable = false;
 
-    public TransformationConfigurationDTO(TransformationConfiguration transformationConfiguration) {
-        this.uid = transformationConfiguration.getUID();
-        this.label = transformationConfiguration.getLabel();
-        this.type = transformationConfiguration.getType();
-        this.content = transformationConfiguration.getContent();
-        this.language = transformationConfiguration.getLanguage();
+    public TransformationConfigurationDTO(Transformation transformation) {
```
kaikreuzer, line 35: Seems you have missed to also refactor the name of the DTO. 😄

bundles/org.openhab.core.transform/src/main/java/org/openhab/core/transform/Transformation.java
```diff
@@ -12,39 +12,43 @@
...
  */
 @NonNullByDefault
-public class TransformationConfiguration implements Identifiable<String> {
+public class Transformation implements Identifiable<String> {
+    public static final String FUNCTION = "function";
+
     private final String uid;
     private final String label;
     private final String type;
     private final @Nullable String language;
```
kaikreuzer, line 34: I would claim that we do not need a `language` property anymore - all language specific stuff is within the `configuration` and those are localizable by default already. The `Transformation` instance itself should not be dependent on a single language.

kaikreuzer (changes requested):

bundles/org.openhab.core.io.rest.transform/src/main/java/org/openhab/core/io/rest/transform/TransformationConfigurationDTO.java
J-N-K: Done. The question is if I should also move the resource to `/transformations` instead of `/transformations/configurations`.

bundles/org.openhab.core.transform/src/main/java/org/openhab/core/transform/Transformation.java
J-N-K: Well, it has always been this way for text-based transformations. I'm not sure we'll break something if the leave it out.

kaikreuzer: Could you please elaborate on this? I'm not sure I understand where we have a language dependency in text-based transformations.

bundles/org.openhab.core.io.rest.transform/src/main/java/org/openhab/core/io/rest/transform/TransformationConfigurationDTO.java
kaikreuzer: I would say so, yes. Then it is in line with other resources.

bundles/org.openhab.core.transform/src/main/java/org/openhab/core/transform/Transformation.java
J-N-K: This has always been the case: https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.transform/src/main/java/org/openhab/core/transform/AbstractFileTransformationService.java. It's implemented by two characters separated with a `_` before the file extension.

kaikreuzer: Interesting, I've never heard about this feature and it does not turn up in any documentation either.
But looking at it, I think it is something different than saying a `Transformation` is language specific. Instead, it brings multi-language support to a single `Transformation` - in your example, the transformation is a MAP with the parameter `switch` and the logic then fetches the right content based on the currently set locale. I.e. all files (`switch.map`, `switch_de.map` and `switch_fr.map`) belong to the very same `Transformation`.

J-N-K: I‘m not sure how long this feature has been there, looking at the commit history at least since 2.5 and I doubt that many people use it. And BTW I don‘t think it‘s working as intended.

I would think that the use-case would be to use it in the UI to display the switch status in different languages (would be very useful for installations with international guests). But then it should select the language based on the browser locale, but this is not the case, it‘s the openHAB system‘s locale that is used.

I agree that it is more „one transformation, different languages“, but it seems to be very difficult to find a way to have different languages in one transformation when configured from UI. We would need to add different functions for each language, that makes the UI very complex.

I think the way it is now is a better compromise between „good modeling“, „backward compatibility“ and „usability from UI“.

kaikreuzer (approved):
Thank you, I think this looks pretty clean now!

PR #3046, opened 2022-07-20
Fix build warnings in addons

J-N-K:
Related to #2982

This fixes the warning reported in the attached PR. Technically this dependency is not necessary but it's missing in the PMD class path if it is not explicitly added here.


jlaur:
@J-N-K - thank you for these fixes.

kaikreuzer (approved):
thanks!

jlaur:
@J-N-K - I can confirm that the warnings disappeared today after the merge yesterday. Thanks!

Issue #3025, opened 2022-07-03
openHAB add-ons are not shown as installed anymore 


cweitkamp:
Beside openHAB add-ons I am using a third-party JSON add-on provider. After a restart of my system all openHAB add-ons are not listed as installed anymore - all things are working properly - whereas all the third-party add-ons are listed correctly.

![Screenshot_20220703-111804](https://user-images.githubusercontent.com/5131747/177033460-18ce9120-b4a3-4ea0-b667-57a94bc2c230.jpg)

![Screenshot_20220703-111821](https://user-images.githubusercontent.com/5131747/177033463-82bace63-7b97-43cf-a60c-38b27ac740ed.jpg)

Extract of REST API response of endpoint `addons/<ADDONID>` looks like this

```
{
  "id": "binding-avmfritz",
  "label": "AVM FRITZ!Box Binding",
  "version": "3.3.0",
  "compatible": true,
  "contentType": "application/vnd.openhab.feature;type=karaf",
  "link": "https://www.openhab.org/addons/bindings/avmfritz/",
  "author": "openHAB",
  "verifiedAuthor": true,
  "installed": false,
  "type": "binding",
  "configDescriptionURI": "",
  "keywords": "",
  "countries": "",
  "connection": "",
  "properties": {},
  "loggerPackages": [
    "org.openhab.binding.avmfritz"
  ]
}
```

After removal of an openHAB add-on e.g. via REST API or console and reinstallation the state of all add-ons is shown correctly again.

J-N-K:
Interesting. I don't see that. 

In the `KarafAddonService` the list of add-ons is created by iterating over all features and calling `getAddon`. That uses the `featureService` to determine if an add-on is installed. Can you check in the karaf console if the features are listed as installed?

cweitkamp:
It looks like the problem is only present after an upgrade - probably related to clearing the cache. After the afore mentioned workaround the listing is correct.

After an upgrade the feature is listed as `Uninstalled` whereas the bundle is `Active`:

```
openhab> bundle:list | grep -i avm
249 │ Active │  80 │ 3.4.0.202207040331     │ openHAB Add-ons :: Bundles :: AVM FRITZ! Binding
openhab> feature:list | grep -i avm                                                                                                                                                   
openhab-binding-avmfritz                          │ 3.4.0.SNAPSHOT   │          │ Uninstalled │ openhab-addons-3.4.0-SNAPSHOT                        │ AVM FRITZ!Box Binding
[...]

PR #3004, opened 2022-06-16
Refactor WatchService

J-N-K:
Depends on #2994 

This replaces the `AbstractWatchService` with a new `WatchService` interface and implementation that can be consumed as OSGi service. It makes use of a library that does a lot of what we currently do (hashing, removing duplicate events) and also implements a native watch service on MacOS which decreases response time to file changes.

- [x] Implement WatchService
- [x] Adjust implementations
- [x] Remove Java17 changes to core (should be merged in a separate PR).
- [x] Resolve itests (not done yet as it will create unnecessary merge conflicts)
- [x] Adjust usages in openhab-addons (https://github.com/openhab/openhab-addons/pull/14004)

@wborn Since we discussed this issue in the past, WDYT about the design?

PR #3049, opened 2022-07-24
Improve FeatureInstaller

J-N-K:
There are some flaws in the `FeatureInstaller` which this PR tries to solve. While debugging #3025 I found that the periodic sync job sets the config cache before the configuration is applied. This might lead to issues if another call to `modified` re-sets the configuration to something else but does not update the configuration cache. It is also not ensured that all configuration updates are processed in the order they arrive, so the asynchronous processing of the updates might result in an older configuration overwriting a newer one. 

This PR 

- introduces a queue that holds the configuration supplied to `modified` and processes them in the correct order 
- ensures that configuration cache is properly updated when an update is processed (independent from the configuration source)
- removed nearly all synchronized as they are no longer needed
- some other code clean-up

### DIFF
--- a/bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/FeatureInstaller.java
+++ b/bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/FeatureInstaller.java
@@ -203,7 +203,7 @@ public class FeatureInstaller implements ConfigurationListener {
             }
             installAddons(config); // we don't have to wait even if the job is running, because method is synchronized
         }
-        
+
         processingConfigQueue.set(false);
     }
 
--- a/bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/FeatureInstaller.java
+++ b/bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/FeatureInstaller.java
@@ -212,6 +212,12 @@ public class FeatureInstaller implements ConfigurationListener {
         }
 
         processingConfigQueue.set(false);
+
+        try {
+            featuresService.refreshFeatures(EnumSet.of(FeaturesService.Option.Upgrade));
+        } catch (Exception e) {
+            logger.error("Failed to refresh bundles after processing config update", e);
+        }
     }
 
     @Nullable
@@ -485,7 +491,8 @@ public class FeatureInstaller implements ConfigurationListener {
         try {
             Feature[] features = featuresService.listInstalledFeatures();
             if (!anyMatchingFeature(features, withName(name))) {
-                featuresService.installFeature(name);
+                featuresService.installFeature(name,
+                        EnumSet.of(FeaturesService.Option.Upgrade, FeaturesService.Option.NoAutoRefreshBundles));
                 features = featuresService.listInstalledFeatures();
                 if (anyMatchingFeature(features, withName(name))) {
                     logger.debug("Installed '{}'", name);
--- a/bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/FeatureInstaller.java
+++ b/bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/FeatureInstaller.java
@@ -187,6 +187,7 @@ public class FeatureInstaller implements ConfigurationListener {
         }
 
         Map<String, Object> config;
+        boolean changed = false;
 
         while ((config = configQueue.poll()) != null) {
             // cache the last processed config
@@ -203,18 +204,23 @@ public class FeatureInstaller implements ConfigurationListener {
             }
 
             if (installPackage(config)) {
+                changed = true;
                 // our package selection has changed, so let's wait for the values to be available in config admin
                 // which we will receive as another call to modified
                 continue;
             }
 
-            installAddons(config); // we don't have to wait even if the job is running, because method is synchronized
+            if (installAddons(config)) {
+                changed = true;
+            }
         }
 
         processingConfigQueue.set(false);
 
         try {
-            featuresService.refreshFeatures(EnumSet.of(FeaturesService.Option.Upgrade));
+            if (changed) {
+                featuresService.refreshFeatures(EnumSet.noneOf(FeaturesService.Option.class));
+            }
         } catch (Exception e) {
             logger.error("Failed to refresh bundles after processing config update", e);
         }
@@ -386,7 +392,7 @@ public class FeatureInstaller implements ConfigurationListener {
         return changed;
     }
 
-    private void installAddons(final Map<String, Object> config) {
+    private boolean installAddons(final Map<String, Object> config) {
         final Set<String> currentAddons = new HashSet<>(); // the currently installed ones
         final Set<String> targetAddons = new HashSet<>(); // the target we want to have installed afterwards
         final Set<String> installAddons = new HashSet<>(); // the ones to be installed (the diff)
@@ -435,6 +441,8 @@ public class FeatureInstaller implements ConfigurationListener {
 
         // do the de-installation
         uninstallAddons.forEach(this::uninstallFeature);
+
+        return !installAddons.isEmpty() || !uninstallAddons.isEmpty();
     }
 
     private Set<String> getAllFeatureNamesWithPrefix(String prefix) {
--- a/bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/FeatureInstaller.java
+++ b/bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/FeatureInstaller.java
@@ -118,8 +118,6 @@ public class FeatureInstaller implements ConfigurationListener {
 
     private @Nullable String onlineRepoUrl = null;
 
-    private @Nullable ScheduledFuture<?> installJob;
-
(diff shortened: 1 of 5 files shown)


bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/FeatureInstaller.java
```diff
@@ -451,17 +439,17 @@ private Set<String> getAllFeatureNamesWithPrefix(String prefix) {
...
-    private synchronized void installFeatures(Set<String> addons) {
+    private void installFeatures(Set<String> addons) {
         try {
             if (logger.isDebugEnabled()) {
-                logger.debug("Installing '{}'", addons.stream().collect(Collectors.joining(", ")));
+                logger.debug("Installing '{}'", String.join(", ", addons));
             }
-            featuresService.installFeatures(addons,
-                    EnumSet.of(FeaturesService.Option.Upgrade, FeaturesService.Option.NoFailOnFeatureNotFound));
+            featuresService.installFeatures(addons, EnumSet.of(FeaturesService.Option.NoAutoRefreshBundles,
```
J-N-K, line 451: @wborn Do you know if this might cause issues? The reason I added it is that some add-ons cause `org.openhab.core` to refresh if it is not set. As a result the `EventPublisher` is unloaded and this service is stopped. If a lot of add-ons are installed at the same time, this will result in an `InterruptedException` in `FeaturesService.installFeatures` which in turn shows up as ERROR in the log. At the next start everything is fixed.

It would be great if there is a documentation for the `FeaturesService.Option` (at least I didn't find it). Do you know of such a document?

wborn: I don't think there is much documentation for the `FeaturesService` besides JavaDocs and the code itself. So if that doesn't help, mailing questions on the Karaf mailing list might help.

Maybe you can figure out what causes the bundle refresh in the first place and prevent it from happening?

The [Karaf docs](https://karaf.apache.org/manual/latest/#_feature_and_resolver) describes the reasons for automatically refreshing bundles:

> For instance, a bundle has to be refreshed:
> 
> 1. when bundle A uses an import package, and bundle B is installed providing a new version of the package (matching bundle A import version range). Then bundle A has to be refreshed to use the new version of the package.
> 2. when bundle A uses an optional import package, and bundle B is installed providing this package. Then, bundle A has to be refreshed to actually use the package.
> 3. when bundle A uses a package provided by bundle B, then, bundle A will be refreshed as well. This is kind of "cascading" refresh.

J-N-K:
I think I found it (or at least one reason). This is the tree for `org.openhab.core` before any add-on is installed:

```
Bundle org.openhab.core [148] is currently ACTIVE

org.openhab.core [148]
+- tech.units.indriya [230]
|  +- javax.measure.unit-api [52]
|  +- org.apache.servicemix.bundles.javax-inject [86]
|  +- uom-lib-common [231]
|     +- javax.measure.unit-api [52]
+- org.apache.karaf.services.eventadmin [2]
|  +- org.apache.felix.configadmin [12]
|  |  +- org.apache.felix.coordinator [10]
|  +- org.apache.felix.metatype [4]
+- uom-lib-common [231]
+- si-units [227]
|  +- javax.measure.unit-api [52]
|  +- tech.units.indriya [230]
|  +- si.uom.si-quantity [228]
|  |  +- javax.measure.unit-api [52]
|  +- org.apache.servicemix.bundles.javax-inject [86]
+- javax.measure.unit-api [52]
+- com.google.gson [35]
+- org.apache.felix.scr [68]
|  +- org.apache.felix.configadmin [12]
|  +- org.apache.felix.metatype [4]
|  +- org.osgi.util.promise [9]
|  |  +- org.osgi.util.function [8]
|  +- org.apache.karaf.shell.core [23]
|     +- org.apache.karaf.services.eventadmin [2]
|     +- org.apache.felix.configadmin [12]
|     +- org.ops4j.pax.logging.pax-logging-api [5]
|     |  +- org.apache.karaf.services.eventadmin [2]
|     +- org.jline [25]
|        +- org.apache.sshd.sftp [93]
|        |  +- org.apache.sshd.osgi [91]
|        |  |  +- bcpkix [26]
|        |  |  |  +- bcprov [27]
|        |  |  |  +- bcutil [28]
|        |  |  |     +- bcprov [27]
|        |  |  +- org.ops4j.pax.logging.pax-logging-api [5]
|        |  |  +- bcprov [27]
|        |  +- org.ops4j.pax.logging.pax-logging-api [5]
|        +- org.apache.sshd.scp [92]
|        |  +- org.apache.sshd.osgi [91]
|        |  +- org.ops4j.pax.logging.pax-logging-api [5]
|        +- org.fusesource.jansi [6]
|        +- org.apache.sshd.osgi [91]
+- org.ops4j.pax.logging.pax-logging-api [5]
```

When installing `binding-amazondashbutton` the bundle reloads. This is because the binding also installs the feature `openhab-runtime-jna` and after the installation the tree looks like that:

```
Bundle org.openhab.core [148] is currently ACTIVE

org.openhab.core [148]
+- tech.units.indriya [230]
|  +- javax.measure.unit-api [52]
|  +- org.apache.servicemix.bundles.javax-inject [86]
|  +- uom-lib-common [231]
|     +- javax.measure.unit-api [52]
+- org.apache.karaf.services.eventadmin [2]
|  +- org.apache.felix.configadmin [12]
|  |  +- org.apache.felix.coordinator [10]
|  +- org.apache.felix.metatype [4]
+- uom-lib-common [231]
+- si-units [227]
|  +- javax.measure.unit-api [52]
|  +- tech.units.indriya [230]
|  +- si.uom.si-quantity [228]
|  |  +- javax.measure.unit-api [52]
|  +- org.apache.servicemix.bundles.javax-inject [86]
+- javax.measure.unit-api [52]
+- com.google.gson [35]
+- org.apache.felix.scr [68]
|  +- org.apache.felix.configadmin [12]
|  +- org.apache.felix.metatype [4]
|  +- org.osgi.util.promise [9]
|  |  +- org.osgi.util.function [8]
|  +- org.apache.karaf.shell.core [23]
|     +- org.apache.karaf.services.eventadmin [2]
|     +- org.apache.felix.configadmin [12]
|     +- org.ops4j.pax.logging.pax-logging-api [5]
|     |  +- org.apache.karaf.services.eventadmin [2]
|     +- org.jline [25]
|        +- org.apache.sshd.sftp [93]
|        |  +- org.apache.sshd.osgi [91]
|        |  |  +- bcpkix [26]
|        |  |  |  +- bcprov [27]
|        |  |  |  +- bcutil [28]
|        |  |  |     +- bcprov [27]
|        |  |  +- org.ops4j.pax.logging.pax-logging-api [5]
|        |  |  +- bcprov [27]
|        |  +- org.ops4j.pax.logging.pax-logging-api [5]
|        +- org.apache.sshd.scp [92]
|        |  +- org.apache.sshd.osgi [91]
|        |  +- org.ops4j.pax.logging.pax-logging-api [5]
|        +- org.apache.sshd.osgi [91]
|        +- com.sun.jna [256]
|        +- org.fusesource.jansi [6]
+- org.ops4j.pax.logging.pax-logging-api [5]
```

So an optional dependency is fulfilled and because of (2) the core bundle is refreshed. This causes a lot of other bundles to refresh (including all UI bundles). I believe there are similar issues with other bundles/features and this might also be the reason for a lot of strange behavior when installing add-ons, especially if several add-ons are installed at the same time (e.g. after an upgrade). This not only affects the `FeatureInstaller` but also other services (e.g. the `KarafAddonService`.

I'm not sure how to address that. We could add `openhab-runtime-jna` to `openhab-runtime-base`, this will solve the issue I found here. It does not prevent similar issues and while there may be a general benefit if we provide JNA by default, this might not be the case for other optional dependencies. Maybe also creating an override that blacklists the JNA bundle might work.

bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/FeatureInstaller.java
wborn: > As a result the `EventPublisher` is unloaded and this service is stopped

Would the feature installation succeed when refreshing occurs while the `EventPublisher` has reference cardinaltiy optional? If that works, it could continue and for instance post any events whenever the `EventPublisher` is set again.

wborn:
> We could add openhab-runtime-jna to openhab-runtime-base, this will solve the issue I found here.

Looks like this is a required dependency for https://github.com/openhab/openhab-core/pull/3004 anyhow.

bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/FeatureInstaller.java
J-N-K: I don't think that makes a difference. If the `EventPublisher` is removed, this is most likely because the `org.openhab.core` bundle is refreshed. This will automatically refresh the `org.openhab.core.karaf` bundle and the service is de-activated anyway. I now removed immediate refreshing during installation and refresh after all config updates have been processed (but only if the add-on configuration changed).

J-N-K:
I moved JNA to an `openhab.tp-jna` feature and included that in the `openhab-core-base` feature. This needs an adjustment in openhab-addons: https://github.com/openhab/openhab-addons/pull/13298

wborn (approved):
Thanks! Let's test if this fixes all the flaws. :slightly_smiling_face:

PR #3051, opened 2022-07-28
Add "time only"-Option for DateTimeTrigger in rules

Flole998:
Extends the DateTimeTrigger introduced in ca94fd5 and dd13da5 to allow triggering only when the time matches to allow for daily triggers.

I tried to use "Time is time of item" as trigger for DSL-Rules first as that seems to better fit into the existing model but unfortunately that breaks the existing rules, so "Time matches time of" was chosen instead. Other ideas are very welcome.

### DIFF
--- a/bundles/org.openhab.core.model.rule/src/org/openhab/core/model/rule/Rules.xtext
+++ b/bundles/org.openhab.core.model.rule/src/org/openhab/core/model/rule/Rules.xtext
@@ -76,7 +76,7 @@ TimerTrigger:
 ;
 
 DateTimeTrigger:
-    'Time' 'is' item=ItemName
+    'Time' 'is' item=ItemName (timeOnly?='timeOnly')?
 ;
 
 SystemTrigger:
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/DateTimeTriggerHandler.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/DateTimeTriggerHandler.java
@@ -57,14 +57,17 @@ public class DateTimeTriggerHandler extends BaseTriggerModuleHandler
 
     public static final String MODULE_TYPE_ID = "timer.DateTimeTrigger";
     public static final String CONFIG_ITEM_NAME = "itemName";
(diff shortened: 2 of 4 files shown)


J-N-K:
I wonder if we really need a new trigger here. If a `DateTimeType` is constructed with time only, a date of `1970-01-01` is assumed:

https://github.com/openhab/openhab-core/blob/193da095a8a8d2966b0c456c34c52f1d27e1148d/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java#L86-L89

Couldn't we just check if the date is `1970-01-01` and use the time-only match in that case? This seems to be easier for the user, because it'll take the `DateTimeType` that is created if only a time is passed to the constructor.

Flole998:
Sometimes you don't have control over the date that ends up in an item, for example if a binding is setting it to 1900-1-1 then it would no longer work, an additional rule is needed to create a modified DateTimeItem. Or if you store the DateTime when a light was turned on last time and then trigger on that time then you could compare the date to see if it's longer ago than x days and if it is the light will be switched on by the rule to simulate people turning on and off the lights. I think in certain scenarios this would be a benefit, but I agree that my implementation might not be perfect, it would be better if this could all be put into the DateTimeTrigger somehow, but I have no idea how that could be done.

bundles/org.openhab.core.model.rule/src/org/openhab/core/model/rule/Rules.xtext
```diff
@@ -79,6 +80,10 @@ DateTimeTrigger:
     'Time' 'is' item=ItemName
```
J-N-K, line 80:
```suggestion
    'Time' 'is' item=ItemName (timeOnly?='timeOnly')?
```

Wouldn't this work? It should create a boolean `timeOnly`.

Flole998: Changed it now to use this approach

Flole998:
I additionally implemented a detection if the item is a time-only item to automatically switch to timeOnly mode in that case as it doesn't make sense then to use the date there aswell.

J-N-K (approved):
LGTM, thanks.

PR #3053, opened 2022-07-29
Prevent crash on invalid entry in ManagedThingProvider

J-N-K:
Invalid entries in the database (e.g. duplicate channels) result in an uncatched exception which prevents adding things from storage to the `ThingRegistry`. This PR adds a try-catch-block around the thing mapping and logs a warning. As a result only the affected thing will be missing.

Reported on the forum: https://community.openhab.org/t/issues-with-3-3-on-win10/137734


wborn (approved):
Thanks!

PR #3069, opened 2022-08-22
Use configured locale for SSE displayState

### LINKED ISSUE #3067: [REST] /rest/events displayState not localized
When working on translating the SMHI binding (which has some stateOptions defined) I noticed that the displayState in the SSE events isn't translated. I tracked it down to [this line](https://github.com/openhab/openhab-core/blob/cabb3f73150d13f44d647f46af8362e396684b92/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/internal/SseItemStatesEventBuilder.java#L77) in the SseItemStatesEventBuilder, which uses `Locale.getDefault()` instead of the browser's Accept-Language as the other REST endpoints do.

pacive:
Fixes #3067 

### DIFF
--- a/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/internal/SseItemStatesEventBuilder.java
+++ b/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/internal/SseItemStatesEventBuilder.java
@@ -25,6 +25,7 @@ import javax.ws.rs.sse.OutboundSseEvent.Builder;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
+import org.openhab.core.io.rest.LocaleService;
 import org.openhab.core.io.rest.sse.internal.dto.StateDTO;
 import org.openhab.core.items.Item;
 import org.openhab.core.items.ItemNotFoundException;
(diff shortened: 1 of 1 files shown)


J-N-K (commented):
LGTM, can you add a test that ensures this will not break again? It should ensure that the locale returned by the `LocaleService` is different from the system-locale and then check that the display state is in the correct format.

pacive:
I've only ever done simple unit tests unfortunately, not sure how I would go about testing something like this, with so many different parts I'm afraid. If I understand it correctly, it would require a mocked Thing with stateOptions defined for a channel, (in two different languages), an Item connected to that channel, plus configuring the [I18nProviderImpl](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/i18n/I18nProviderImpl.java) (the default LocaleProvider), and finally calling SseItemStatesEventBuilder.buildEvent(), to check that the correct language is used? I honestly have no idea how to even begin, but if I could get some guidance I would be happy to learn more about it.

J-N-K:
I think you can mock everything. If you

- mock the bundle context 
- mock the item registry and let it return return a mocked item when an item is requested
- let the mocked item return a state and two different state descriptions, one for the system locale and one for another locale
- mock the locale service and let it return the other locale
- build the item state and check that it is formatted with the other locale, not the system locale

pacive:
Ok, please bear with me trying to get my head around this.

* If I mock the LocaleService, I implicitly make an assumption that this is working as intended.

* The SseItemStatesEventBuilder passes the locale directly to the [getDisplayState()](https://github.com/openhab/openhab-core/blob/cabb3f73150d13f44d647f46af8362e396684b92/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/internal/SseItemStatesEventBuilder.java#L95) method, which passes it on to the Items getStateDescription() method, which in turn passes it to its configured StateDescriptionService, and that is responsible for getting the StateDescription with the correct locale. 

But if I mock this in the Item I thereby make the assumption that the StateDescriptionService part also works, which means the only thing the test will prove is that this method actually gets called. Maybe I'm misunderstanding the purpose of testing, but I can't really see that this proves anything useful?

J-N-K (approved):
Alright, we'll see if it breaks again.

PR #3072, opened 2022-08-28
Fix missing initial thing status event

J-N-K:
A thing always has a status. This status is not properly propagated to the event bus when the thing is added. This e.g. leads to a situation where a thing's first status that is received by am event subscriber is "INITIALIZING" instead of "UNINITIALIZED". The status should always be sent so that the thing lifecycle can properly be tracked by event listeners.

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
@@ -305,7 +305,6 @@ public class ThingManagerImpl
             if (channelType != null) {
                 URI configDescriptionURI = channelType.getConfigDescriptionURI();
                 if (configDescriptionURI != null) {
-                    configDescriptionValidator.validate(configurationParameters, configDescriptionURI);
                 }
             }
         }
@@ -1140,7 +1139,7 @@ public class ThingManagerImpl
                 }
             } else {
                 setThingStatus(thing, new ThingStatusInfo(ThingStatus.UNINITIALIZED,
-                        ThingStatusDetail.HANDLER_REGISTERING_ERROR, "Handler factory not found"));
+                        ThingStatusDetail.HANDLER_MISSING_ERROR, "Handler factory not found"));
                 logger.debug("Not registering a handler at this point. No handler factory for thing '{}' found.",
                         thing.getUID());
             }
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
@@ -559,6 +559,7 @@ public class ThingManagerImpl
             logger.error("A thing with UID '{}' is already tracked by ThingManager. This is a bug.", thing.getUID());
         }
         this.things.put(thing.getUID(), thing);
+        eventPublisher.post(ThingEventFactory.createStatusInfoEvent(thing.getUID(), thing.getStatusInfo()));
         logger.debug("Thing '{}' is tracked by ThingManager.", thing.getUID());
         if (!isHandlerRegistered(thing)) {
             registerAndInitializeHandler(thing, getThingHandlerFactory(thing));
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1125,7 +1126,7 @@ private void registerAndInitializeHandler(final Thing thing,
             final @Nullable ThingHandlerFactory thingHandlerFactory) {
         if (isDisabledByStorage(thing.getUID())) {
             logger.debug("Not registering a handler at this point. Thing is disabled.");
-            thing.setStatusInfo(new ThingStatusInfo(ThingStatus.UNINITIALIZED, ThingStatusDetail.DISABLED, null));
+            setThingStatus(thing, new ThingStatusInfo(ThingStatus.UNINITIALIZED, ThingStatusDetail.DISABLED, null));
```
J-N-K, line 1129: If the status is set directly, no event is emitted.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
```diff
@@ -559,6 +559,7 @@ public void thingAdded(Thing thing, ThingTrackerEvent thingTrackerEvent) {
             logger.error("A thing with UID '{}' is already tracked by ThingManager. This is a bug.", thing.getUID());
         }
         this.things.put(thing.getUID(), thing);
+        eventPublisher.post(ThingEventFactory.createStatusInfoEvent(thing.getUID(), thing.getStatusInfo()));
```
J-N-K, line 562: The thing already has that status, we only need to notify listeners about it.

spacemanspiff2007:
> A thing always has a status.

Would it be possible to just use this status for the `ThingAddedEvent` instead of sending another `ThingStatusEvent`?

J-N-K:
IMO since we have the `ThingStatusEvent` it makes more sense to emit one of these than adding a status to an event that is not directly connected to a status. It's also more consistent since the `ItemAddedEvent` also does not carry information about the item's state.

spacemanspiff2007:
A freshly added item has a defined state (NULL). I'm not aware of such a definition for a Thing.
That's why I think it would be good to include the status in the `ThingAddedEvent` so the status on the client side has a defined state.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1138,6 +1139,8 @@ private void registerAndInitializeHandler(final Thing thing,
                             identifier);
                 }
             } else {
+                setThingStatus(thing, new ThingStatusInfo(ThingStatus.UNINITIALIZED,
+                        ThingStatusDetail.HANDLER_REGISTERING_ERROR, "Handler factory not found"));
```
kaikreuzer, line 1143: Wouldn't the right sub-status be `HANDLER_MISSING_ERROR` here?
Have you checked, how often we run into this code? My hope would be that at startup we only add the things once the handlers are already available - if not, we will see one such event for every Thing on startup, which would be a bit annoying.

J-N-K: Unfortunately this happens quite often (at least for managed things). This will not show up in the log, though.

I agree on the `HANDLER_MISSING_ERROR`.

kaikreuzer: > This will not show up in the log, though.

Not even in the event log?

J-N-K: No. I guess this is because it happens very early in the startup process where no events are logged.

kaikreuzer: Ok, then I'd hope we can live with that during startup.
So could you please change the code to `HANDLER_MISSING_ERROR`? Once this is done, I can merge the PR.

J-N-K: Done

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
```diff
@@ -305,7 +305,6 @@ public void validateConfigurationParameters(Channel channel, Map<String, Object>
             if (channelType != null) {
                 URI configDescriptionURI = channelType.getConfigDescriptionURI();
                 if (configDescriptionURI != null) {
-                    configDescriptionValidator.validate(configurationParameters, configDescriptionURI);
```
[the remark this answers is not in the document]
J-N-K: Yes.

kaikreuzer (approved):
Thank you!

PR #3075, opened 2022-09-01
Add visibility and color parameters to UI sitemap provider

mherwege:
This adds the missing parameters to the sitemap provider. This allows the UI to be further extended in a follow-up PR and make the sitemap UI configuration support the same features as textual configuration.

Signed-off-by: Mark Herwege <mark.herwege@telenet.be>

### DIFF
--- a/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
+++ b/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
@@ -19,6 +19,8 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.eclipse.emf.common.util.EList;
 import org.eclipse.jdt.annotation.NonNullByDefault;
@@ -28,13 +30,16 @@ import org.openhab.core.config.core.ConfigUtil;
 import org.openhab.core.model.core.EventType;
 import org.openhab.core.model.core.ModelRepositoryChangeListener;
 import org.openhab.core.model.sitemap.SitemapProvider;
+import org.openhab.core.model.sitemap.sitemap.ColorArray;
 import org.openhab.core.model.sitemap.sitemap.LinkableWidget;
 import org.openhab.core.model.sitemap.sitemap.Mapping;
 import org.openhab.core.model.sitemap.sitemap.Sitemap;
 import org.openhab.core.model.sitemap.sitemap.SitemapFactory;
 import org.openhab.core.model.sitemap.sitemap.SitemapPackage;
+import org.openhab.core.model.sitemap.sitemap.VisibilityRule;
 import org.openhab.core.model.sitemap.sitemap.Widget;
 import org.openhab.core.model.sitemap.sitemap.impl.ChartImpl;
+import org.openhab.core.model.sitemap.sitemap.impl.ColorArrayImpl;
 import org.openhab.core.model.sitemap.sitemap.impl.ColorpickerImpl;
 import org.openhab.core.model.sitemap.sitemap.impl.DefaultImpl;
 import org.openhab.core.model.sitemap.sitemap.impl.FrameImpl;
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
```diff
@@ -318,6 +330,77 @@ private void addWidgetMappings(EList<Mapping> mappings, UIComponent component) {
...
+                    if (matcher.find()) {
+                        condition = matcher.group();
+                    }
+                    String sign = "";
+                    String state = sourceVisibility.toString().split(COMPARATOR_STRING)[1].trim();
+                    if ((state.charAt(0) == '+') || (state.charAt(0) == '-')) {
+                        sign = state.substring(0, 0);
+                        state = state.substring(1).trim();
+                    }
+                    VisibilityRuleImpl visibilityRule = (VisibilityRuleImpl) SitemapFactory.eINSTANCE
```
J-N-K, line 349: I think this can be simplified by using a better regex pattern:

```
(?<item>[\w_]+)?\s*(?<condition>==|!=|<=|>=|<|>)\s*(?<sign>\+|-)?\s*(?<value>\S+)
```

You can then use `.find` and have direct access to all groups. I also added named groups (`item`, `condition`, `sign` and `state`), so you can use `.group("item")` instead of `.group(0)`. I believe this results in much better performance (and simpler code) than matching, splitting and trimming several times.

Did you check if this also works for the "no item name provided, use current item name" case?

A similar regex can be used for the color below.

J-N-K (changes requested):

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
```diff
@@ -318,6 +334,71 @@ private void addWidgetMappings(EList<Mapping> mappings, UIComponent component) {
...
+                        String condition = matcher.group("condition");
+                        String sign = matcher.group("sign");
+                        String state = matcher.group("state");
+                        VisibilityRuleImpl visibilityRule = (VisibilityRuleImpl) SitemapFactory.eINSTANCE
+                                .createVisibilityRule();
+                        visibilityRule.setItem(item);
+                        visibilityRule.setCondition(condition);
+                        visibilityRule.setSign(sign);
+                        visibilityRule.setState(state);
+                        visibility.add(visibilityRule);
```
J-N-K, line 353:
```suggestion
                        VisibilityRuleImpl visibilityRule = (VisibilityRuleImpl) SitemapFactory.eINSTANCE
                                .createVisibilityRule();
                        visibilityRule.setItem(matcher.group("item"));
                        visibilityRule.setCondition(matcher.group("condition"));
                        visibilityRule.setSign(matcher.group("sign"));
                        visibilityRule.setState(matcher.group("state"));
                        visibility.add(visibilityRule);
```

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
```diff
@@ -79,6 +85,13 @@ public class UIComponentSitemapProvider implements SitemapProvider, RegistryChan
     private static final String SITEMAP_PREFIX = "uicomponents_";
     private static final String SITEMAP_SUFFIX = ".sitemap";
 
+    private static final String COMPARATOR_STRING = "==|!=|<=|>=|<|>";
+    private static final Pattern COMPARATOR = Pattern.compile(COMPARATOR_STRING);
```
J-N-K, line 89: I think you can remove these two now, correct?

J-N-K (changes requested):
Otherwise LGTM, thanks.

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
mherwege: Thanks

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
mherwege: Indeed

J-N-K (approved):

J-N-K:
Thanks!

PR #3078, opened 2022-09-08
[tls] Rethrow 'IOException' instead of swallowing it

cweitkamp:
- Rethrow `IOException` instead of swallowing it

In case of unreachable servers there is an `IOException` leading to an error message in the log file. This case can be handled even better when rethrowing the thrown exception which allows the calling instance to react to server side errors.

https://community.openhab.org/t/vizio-tv-binding/138367/12

### DIFF
--- a/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/PEMTrustManager.java
+++ b/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/PEMTrustManager.java
@@ -40,6 +40,7 @@ import javax.net.ssl.X509ExtendedTrustManager;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
+import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/PEMTrustManager.java
```diff
@@ -62,15 +62,17 @@ public final class PEMTrustManager extends X509ExtendedTrustManager {
...
+     * @throws CertificateParsingException
      */
     public PEMTrustManager(String pemCert) throws CertificateException {
         if (!pemCert.isBlank() && pemCert.startsWith(BEGIN_CERT)) {
             try (InputStream certInputStream = new ByteArrayInputStream(pemCert.getBytes(StandardCharsets.UTF_8))) {
                 trustedCert = (X509Certificate) CertificateFactory.getInstance("X.509")
                         .generateCertificate(certInputStream);
             } catch (IOException e) {
-                throw new CertificateException(e);
+                LoggerFactory.getLogger(PEMTrustManager.class).debug("An IOException occurred: {}", e.getMessage());
```
J-N-K, line 74: Why is the logger not an instance field? Otherwise LGTM

J-N-K (approved):

bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/PEMTrustManager.java
cweitkamp: Done.

PR #3079, opened 2022-09-08
[thing] Added default channel types for electrical energy

cweitkamp:
- Added default channel types for electrical energy (electric power, electric current, electric potential and electrical energy)


bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/DefaultSystemChannelTypeProvider.java
```diff
@@ -347,13 +355,56 @@ public class DefaultSystemChannelTypeProvider implements ChannelTypeProvider {
...
+
+    /**
+     * Electric-potential: system wide {@link ChannelType} which shows the electric potential
+     */
+    public static final ChannelType SYSTEM_ELECTRIC_POTENTIAL = ChannelTypeBuilder
+            .state(SYSTEM_CHANNEL_TYPE_UID_ELECTRIC_POTENTIAL, "Electric Potential", "Number:ElectricPotential")
+            .withDescription("Current electric potential").withCategory("Energy")
+            .withStateDescriptionFragment(
+                    StateDescriptionFragmentBuilder.create().withReadOnly(true).withPattern("%.1f %unit%").build())
+            .withTags(List.of("Measurement", "Voltage")).build();
```
J-N-K, line 388: Are you sure that this is a good naming? Usually it'll probably be a potential difference (aka voltage), not na absolute potential.

cweitkamp: No, I am not. I tried to create a relation to the dimension to be used with the ItemType and a cluster for electrical Channels. In general we can name all of them just Current, Power, Voltage and Energy - without the prefix.

J-N-K: IMO that would be better from the user perspective.

cweitkamp: Me too.

But there is one problem. The channel type `power` already is user for turning devices on and off:

https://github.com/openhab/openhab-core/blob/2779c0fe0fd36872c6cdcc1cab006d3c774b2d90/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/DefaultSystemChannelTypeProvider.java#L70

J-N-K: Maybe keep the "Electric" prefix then. There will probably be no clash for "voltage" but "current" could also be related to water or gas.

cweitkamp: Okay. Then I will rename just electric potential to electric voltage and keep the rest.

J-N-K (approved):

PR #3091, opened 2022-09-22
Add new ParameterizedTransformationService interface

ccutrer:
This allows new uses for transformation services where more structured objects can be passed in. In my particular use case, the MQTT Home Assistant binding already depends on the Jinja transform service in order to comply with Home Assistant's behavior. But some use cases require passing more state to Jinja than a single `value`: https://www.home-assistant.io/integrations/light.mqtt/#command_on_template

https://github.com/openhab/openhab-addons/pull/13413 is my adding PR that will utilize this. In its state as of posting, I've just added the interface to the Jinja bundle, but it's being problematic (and introduces a hard dependency between the Home Assistant binding and the Jinja bundle, which could be problematic because the Home Assistant binding is part of the MQTT feature which many people install, but never use Home Assistant or Jinja). Also note that I considered adding a special syntax to the Jinja service to pass additional variables (similar to how the script transformation service does), but it's actually important to be able to pass typed data to Jinja, not just strings.

### DIFF
--- a/bundles/org.openhab.core.transform/src/main/java/org/openhab/core/transform/ParameterizedTransformationService.java
+++ b/bundles/org.openhab.core.transform/src/main/java/org/openhab/core/transform/ParameterizedTransformationService.java
@@ -1,0 +1,30 @@
+/**
+ * Copyright (c) 2010-2022 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.transform;
+
+import java.util.Map;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+
+/**
+ * A ParameterizedTransformationService extends TransformationService in order to allow
+ * providing specific parameters to the transform function, instead of a single `value`.
+ *
+ * @author Cody Cutrer - Initial contribution
+ */
+@NonNullByDefault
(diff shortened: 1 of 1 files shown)


J-N-K:
In general bundles in add-ons are not allowed to depend on other bundles in add-ons. I don't know why this was introduced here. 

IMO the external dependencies for the JINJA transformation should be moved to a core feature and the JINJA transformation and the MQTT bundle should add that feature as a dependency, so it gets installed if either of these is installed. This also removes the need for introducing such a new interface.

ccutrer:
> IMO the external dependencies for the JINJA transformation should be moved to a core feature and the JINJA transformation and the MQTT bundle should add that feature as a dependency, so it gets installed if either of these is installed. This also removes the need for introducing such a new interface.

👍  I'll look into going that route. I wanted to avoid having MQTT depend directly on the jinjava external dependency, but hadn't considered just making an intermediary feature that both bundles can depend on. So that would just be adding some metadata to `features/karaf/openhab-core/src/main/feature/feature.xml` in this repo, and then updating the `feature.xml` for MQTT and Jinja bundles to depend on the new feature, and finally updating the `pom.xml` for them to be a `provide` instead of a `compile`, correct?

ccutrer:
Oooh wait, `features/karaf/openhab-tp/src/main/feature.xml` in this repo. and I just figured out that `tp` stands for "third party".

ccutrer:
Oh no, it means "target platform". I still think it's the correct place to put it?

J-N-K:
There has been some discussion about the naming and correct placing in the past (and if addon-bundles should depend on tp-features), but IMO it's comparable to e.g. jackson or netty where we also have tp-features and addon-bundles depending on them (e.g. binding.groheondus or io.homekit). 

Yes, the dependencies in the POM need to be changed. You should probably keep them as scope `compile` and add them to `bnd.noembedding`, otherwise you'll probably run into trouble when running in Eclipse.

ccutrer:
Will take a different approach...

PR #3092, opened 2022-09-23
Fix StartLevelService "hanging" if config is changed.

ccutrer:
Unregistered trackers aren't clear, so then we never register new trackers, and thus never get notified of any changes.


J-N-K (approved):
LGTM, Thanks.

PR #3099, opened 2022-10-03
[rulesupport] sort JSR223 scripts by start level before filename

ccutrer:
The docs say that JSR223 scripts will be loaded in order of start level (as defined by filename or containing folder). this works (mostly) fine when you're stepping through start levels sequentially on startup, but if you're already at SL100, and have scripts located in sl folders, and then install an addon for those scripts, they were loading in alphabetic order instead. This ensures they'll still respect start level order relative to each other.

This should also fix the case of script start levels not matching start levels OpenHAB actually steps through (which are all multiples of 10) - i.e. if you define scripts in sl30, sl31, and sl32, when OpenHAB jumps from 30 to 40, 31 should be executed before 32.


ccutrer:
cc: @jpg0

J-N-K (approved):
LGTM, thanks.

PR #3100, opened 2022-10-05
[voice] Allow stop single dialogs and start dialog on an existing processor

GiviMAD:
Introduces two small changes on the Voice manager:

* Register current running single shot dialog instances so they can be stoped by the voice manager stop method.
* Allow the execution of a single dialog when calling the listenAndAnswer method with the same configurations that other dialog processor running persistently. 

Explanation on the second point: Before, any call to this method that collide by source id with a registered processor was rejected, but if you are using same configurations (ignoring the keyword),I think that makes sense to trigger a dialog execution with the registered dialog processor and the DialogProcessor class implementation already ensures that the keyword will be ignored while processing so there should be no problems on doing so.

Let me know if anything needs to be changed or improved.

### DIFF
--- a/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
+++ b/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
@@ -386,15 +386,16 @@ public class DialogProcessor implements KSListener, STTListener {
     }
 
     /**
-     * Check if other DialogProcessor instance have same configuration ignoring the configured keyword
+     * Check if other DialogProcessor instance have same configuration ignoring the keyword spotting configuration
      *
      * @param dialogProcessor Other DialogProcessor instance
      */
     public boolean isCompatible(DialogProcessor dialogProcessor) {
         return this.sink.equals(dialogProcessor.sink) && this.source.equals(dialogProcessor.source)
-                && Objects.equals(this.ks, dialogProcessor.ks) && this.stt.equals(dialogProcessor.stt)
-                && this.tts.equals(dialogProcessor.tts) && this.hlis.size() == dialogProcessor.hlis.size()
-                && this.hlis.containsAll(dialogProcessor.hlis) && this.locale.equals(dialogProcessor.locale)
+                && this.stt.equals(dialogProcessor.stt) && this.tts.equals(dialogProcessor.tts)
+                && Objects.equals(this.prefVoice, dialogProcessor.prefVoice)
+                && this.hlis.size() == dialogProcessor.hlis.size() && this.hlis.containsAll(dialogProcessor.hlis)
+                && this.locale.equals(dialogProcessor.locale)
                 && Objects.equals(this.listeningItem, dialogProcessor.listeningItem);
     }
 }
--- a/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
+++ b/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
@@ -150,6 +150,10 @@ public class DialogProcessor implements KSListener, STTListener {
         this.ttsFormat = VoiceManagerImpl.getBestMatch(tts.getSupportedFormats(), sink.getSupportedFormats());
     }
 
+    public void startSingleDialog() {
+        executeSimpleDialog();
+    }
+
     public void start() {
         KSService ksService = ks;
         if (ksService != null) {
@@ -209,6 +213,10 @@ public class DialogProcessor implements KSListener, STTListener {
         abortKS();
         closeStreamKS();
         toggleProcessing(false);
+    }
+
+    public boolean isProcessing() {
+        return processing;
     }
 
     private void abortKS() {
@@ -375,5 +383,18 @@ public class DialogProcessor implements KSListener, STTListener {
                 logger.warn("Error saying '{}': {}", text, e.getMessage());
             }
         }
+    }
+
+    /**
+     * Check if other DialogProcessor instance have same configuration ignoring the configured keyword
+     *
+     * @param dialogProcessor Other DialogProcessor instance
+     */
+    public boolean isCompatible(DialogProcessor dialogProcessor) {
+        return this.sink.equals(dialogProcessor.sink) && this.source.equals(dialogProcessor.source)
+                && Objects.equals(this.ks, dialogProcessor.ks) && this.stt.equals(dialogProcessor.stt)
+                && this.tts.equals(dialogProcessor.tts) && this.hlis.size() == dialogProcessor.hlis.size()
+                && this.hlis.containsAll(dialogProcessor.hlis) && this.locale.equals(dialogProcessor.locale)
+                && Objects.equals(this.listeningItem, dialogProcessor.listeningItem);
     }
 }
--- a/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
+++ b/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
@@ -127,7 +127,7 @@ public class VoiceManagerImpl implements VoiceManager, ConfigOptionProvider {
     private final Map<String, String> defaultVoices = new HashMap<>();
 
     private Map<String, DialogProcessor> dialogProcessors = new HashMap<>();
-    private ConcurrentHashMap<String, DialogProcessor> singleDialogProcessors = new ConcurrentHashMap<>();
+    private Map<String, DialogProcessor> singleDialogProcessors = new ConcurrentHashMap<>();
 
     @Activate
     public VoiceManagerImpl(final @Reference LocaleProvider localeProvider, final @Reference AudioManager audioManager,
@@ -546,7 +546,7 @@ public class VoiceManagerImpl implements VoiceManager, ConfigOptionProvider {
         AudioSource audioSource = (source == null) ? audioManager.getSource() : source;
         if (audioSource != null) {
             DialogProcessor processor = dialogProcessors.remove(audioSource.getId());
-            cleanSingleDialogProcessors();
+            singleDialogProcessors.values().removeIf(e -> !e.isProcessing());
             if (processor == null) {
                 processor = singleDialogProcessors.get(audioSource.getId());
             }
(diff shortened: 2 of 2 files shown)


GiviMAD:
WDYT @lolodomo?

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
```diff
@@ -376,4 +384,17 @@ protected void say(@Nullable String text) {
...
     }
+
+    /**
+     * Check if other DialogProcessor instance have same configuration ignoring the configured keyword
+     *
+     * @param dialogProcessor Other DialogProcessor instance
+     */
+    public boolean isCompatible(DialogProcessor dialogProcessor) {
+        return this.sink.equals(dialogProcessor.sink) && this.source.equals(dialogProcessor.source)
+                && Objects.equals(this.ks, dialogProcessor.ks) && this.stt.equals(dialogProcessor.stt)
```
lolodomo, line 395: What is the difference leading to use `Objects.equals` for ks ?

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -598,13 +603,24 @@ public void listenAndAnswer(@Nullable STTService stt, @Nullable TTSService tts,
...
                 logger.debug("Executing a simple dialog for source {} ({})", audioSource.getLabel(null),
                         audioSource.getId());
-                processor = new DialogProcessor(sttService, ttsService, prefVoice, interpreters, audioSource, audioSink,
-                        loc, item, this.eventPublisher, this.i18nProvider, b);
                 processor.start();
+                singleDialogProcessors.put(audioSource.getId(), processor);
+            } else if (!isSingleDialog && activeProcessor.isCompatible(processor)) {
+                logger.debug("Executing a simple dialog for active source {} ({})", audioSource.getLabel(null),
+                        audioSource.getId());
+                activeProcessor.startSingleDialog();
```
lolodomo, line 623: This will interrupt a potential running dialogue.
And I am not sure the permanent dialogue will be restored after ?

lolodomo (changes requested):

lolodomo:
> WDYT @lolodomo?

This looks like a good idea.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
GiviMAD: Because both are nullable, but it was an error as I pretended to ignore all the keyword spotting related config in this methods.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
GiviMAD: Yes, it's safe, the SpotEvent of the ks also calls this method under the hood, so it works well, it ignore other SpotEvent in the mean time but it continue with the spotting after it's finish.

J-N-K:
I must admit that my knowledge in this area is not very good.Can you explain what a "single dialog" is and what makes the difference to a "dialog"?

GiviMAD:
> I must admit that my knowledge in this area is not very good.Can you explain what a "single dialog" is and what makes the difference to a "dialog"?

Sure, the dialog processor expose two different capabilities at the moment:
* Running persistently waiting for the keyword to be spotted, to fire the dialog flow which can be simplified as: source => stt => hli => tts => sink. This is exposed through the "start dialog" functionality. This is what we were calling dialog.
* Run  just one dialog iteration (source => stt => hli => tts => sink). This is exposed through the "listen and answer" functionality. This is what we were calling single dialog.

Let me know if some part is unclear.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -598,13 +603,24 @@ public void listenAndAnswer(@Nullable STTService stt, @Nullable TTSService tts,
             throw new IllegalStateException(
                     "Cannot execute a simple dialog as provided locale is not supported by all services.");
         } else {
-            DialogProcessor processor = dialogProcessors.get(audioSource.getId());
-            if (processor == null) {
+            boolean isSingleDialog = false;
+            DialogProcessor activeProcessor = dialogProcessors.get(audioSource.getId());
+            cleanSingleDialogProcessors();
```
J-N-K, line 608: Above (in l. 548-551) you call the clean method only if `processor == null`, here you call it always. Is there any reason to do that differently?

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -126,6 +126,7 @@ public class VoiceManagerImpl implements VoiceManager, ConfigOptionProvider {
     private final Map<String, String> defaultVoices = new HashMap<>();
 
     private Map<String, DialogProcessor> dialogProcessors = new HashMap<>();
+    private Map<String, DialogProcessor> singleDialogProcessors = new HashMap<>();
```
J-N-K, line 129: You might want to use `ConcurrentHashMap` here since you modify the same map from different points in code which I believe can be called concurrently.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -863,4 +879,12 @@ private Comparator<Voice> createVoiceComparator(Locale locale) {
...
         return null;
     }
+
+    private void cleanSingleDialogProcessors() {
+        singleDialogProcessors.forEach((key, processor) -> {
+            if (!processor.isProcessing()) {
+                singleDialogProcessors.remove(key);
+            }
+        });
+    }
```
J-N-K, line 889: I don't think you need a method here, it's a one-liner:

```
    singleDialogProcessors.removeIf(e -> !e.getValue().isProcessing());
```

J-N-K (changes requested):
Sorry for coming back to this so late.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
GiviMAD: I'll let both offside the if. I think in both places it will ensure you not end with a lot of finished dialog references in memory, but better to remove those as soon as posible and should not impact the performance.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
GiviMAD: Ends up like this:
```
singleDialogProcessors.values().removeIf(e -> !e.isProcessing());
```
I didn't know about the entrySet and values methods retuning views that you can use to modify the initial collection. Really good to know, thank you!

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -126,6 +127,7 @@ public class VoiceManagerImpl implements VoiceManager, ConfigOptionProvider {
     private final Map<String, String> defaultVoices = new HashMap<>();
 
     private Map<String, DialogProcessor> dialogProcessors = new HashMap<>();
+    private ConcurrentHashMap<String, DialogProcessor> singleDialogProcessors = new ConcurrentHashMap<>();
```
J-N-K, line 130:
```suggestion
    private HashMap<String, DialogProcessor> singleDialogProcessors = new ConcurrentHashMap<>();
```

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -598,13 +604,24 @@ public void listenAndAnswer(@Nullable STTService stt, @Nullable TTSService tts,
             throw new IllegalStateException(
                     "Cannot execute a simple dialog as provided locale is not supported by all services.");
         } else {
-            DialogProcessor processor = dialogProcessors.get(audioSource.getId());
-            if (processor == null) {
+            boolean isSingleDialog = false;
+            DialogProcessor activeProcessor = dialogProcessors.get(audioSource.getId());
+            cleanSingleDialogProcessors();
```
J-N-K, line 609:
```suggestion
            singleDialogProcessors.values().removeIf(e -> !e.isProcessing());
```

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -598,13 +604,24 @@ public void listenAndAnswer(@Nullable STTService stt, @Nullable TTSService tts,
...
+            }
+            var processor = new DialogProcessor(sttService, ttsService, prefVoice, interpreters, audioSource, audioSink,
+                    loc, item, this.eventPublisher, this.i18nProvider, b);
+            if (activeProcessor == null) {
                 logger.debug("Executing a simple dialog for source {} ({})", audioSource.getLabel(null),
                         audioSource.getId());
-                processor = new DialogProcessor(sttService, ttsService, prefVoice, interpreters, audioSource, audioSink,
-                        loc, item, this.eventPublisher, this.i18nProvider, b);
                 processor.start();
+                singleDialogProcessors.put(audioSource.getId(), processor);
```
J-N-K, line 620:
```suggestion
                singleDialogProcessors.values().removeIf(e -> !e.isProcessing());
```

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -863,4 +880,8 @@ private Comparator<Voice> createVoiceComparator(Locale locale) {
         }
         return null;
     }
+
+    private void cleanSingleDialogProcessors() {
+        singleDialogProcessors.values().removeIf(e -> !e.isProcessing());
+    }
```
J-N-K, line 886:
```suggestion
```

J-N-K (changes requested):
Otherwise LGTM

J-N-K (approved):
Thanks

GiviMAD:
> Thanks

Thank you for the review.

PR #3106, opened 2022-10-07
[auth] grant implicit user role by network

GiviMAD:
This PR introduces a new advanced option on the Api Security settings to define a list comma separated CDIR, to grant the requests originated from those networks with an implicit user role.

This configuration is ignored when "implicit user role" option is enabled.

For me it covers to necessities:

Allowing the Alexa skill to work the "implicit user role" option is disabled.
Allowing me to have an "Implicit user role" when I'm inside my house but always require credentials when I'm accessing through my domine.
Hope everything is in place and let me know if any documentation update is needed.

### DIFF
--- a/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
+++ b/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
@@ -13,13 +13,17 @@
 package org.openhab.core.io.rest.auth.internal;
 
 import java.io.IOException;
+import java.net.InetAddress;
 import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.time.Duration;
+import java.util.Arrays;
 import java.util.Base64;
+import java.util.List;
 import java.util.Map;
 import java.util.Random;
+import java.util.stream.Collectors;
 
 import javax.annotation.Priority;
 import javax.servlet.http.HttpServletRequest;
@@ -66,6 +70,7 @@ import org.slf4j.LoggerFactory;
  * @author Yannick Schaus - Add support for API tokens
  * @author Sebastian Gerber - Add basic auth caching
  * @author Kai Kreuzer - Add null annotations, constructor initialization
+ * @author Miguel Álvarez - Add trusted networks for implicit user role
  */
 @PreMatching
 @Component(configurationPid = "org.openhab.restauth", property = Constants.SERVICE_PID + "=org.openhab.restauth")
@@ -80,14 +85,15 @@ public class AuthFilter implements ContainerRequestFilter {
 
     private static final String ALT_AUTH_HEADER = "X-OPENHAB-TOKEN";
     private static final String API_TOKEN_PREFIX = "oh.";
-
     protected static final String CONFIG_URI = "system:restauth";
     private static final String CONFIG_ALLOW_BASIC_AUTH = "allowBasicAuth";
     private static final String CONFIG_IMPLICIT_USER_ROLE = "implicitUserRole";
+    private static final String CONFIG_TRUSTED_NETWORKS = "trustedNetworks";
     private static final String CONFIG_CACHE_EXPIRATION = "cacheExpiration";
 
     private boolean allowBasicAuth = false;
     private boolean implicitUserRole = true;
+    private List<String> trustedNetworks = List.of();
     private Long cacheExpiration = 6L;
 
     private ExpiringUserSecurityContextCache authCache = new ExpiringUserSecurityContextCache(
@@ -139,6 +145,9 @@ public class AuthFilter implements ContainerRequestFilter {
             allowBasicAuth = value != null && "true".equals(value.toString());
             value = properties.get(CONFIG_IMPLICIT_USER_ROLE);
             implicitUserRole = value == null || !"false".equals(value.toString());
+            value = properties.get(CONFIG_TRUSTED_NETWORKS);
+            trustedNetworks = value == null ? List.of()
+                    : Arrays.stream(value.toString().split(",")).map(String::trim).collect(Collectors.toList());
             value = properties.get(CONFIG_CACHE_EXPIRATION);
             if (value != null) {
                 try {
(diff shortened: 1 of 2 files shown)


GiviMAD:
I forgot to regenerate the default translations.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -267,4 +276,55 @@ public void filter(@Nullable ContainerRequestContext requestContext) throws IOEx
...
+        }
+        return trustedNetworks.stream().anyMatch((networkCDIR) -> {
+            var clientIp = getClientIp(requestContext);
+            var isInTrustedNetwork = isIpInSubnet(clientIp, networkCDIR);
+            if (isInTrustedNetwork) {
+                logger.debug("Granted implicit user role to request with ip {} as part of network {}", clientIp,
+                        networkCDIR);
+            }
+            return isInTrustedNetwork;
+        });
```
J-N-K, line 295:
```suggestion
        String clientIp = getClientIp(requestContext);
        return trustedNetworks.stream().anyMatch(networkCIDR -> isInSubNet(clientIp, networkCIDR);
```

You could also create `SubnetUtils.SubnetInfo`s when parsing the config, so you can call `SubnetInfo::isInRange` here.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -267,4 +276,55 @@ public void filter(@Nullable ContainerRequestContext requestContext) throws IOEx
...
+    }
+
+    private String getClientIp(ContainerRequestContext requestContext) {
+        String ipForwarded = requestContext.getHeaderString("x-forwarded-for");
+        String[] ips = ipForwarded == null ? null : ipForwarded.split(",");
+        String client_ip = (ips == null || ips.length == 0) ? null : ips[0];
+        client_ip = (client_ip == null || client_ip.isEmpty()) ? servletRequest.getRemoteAddr() : client_ip;
+        return client_ip;
+    }
```
J-N-K, line 304:
```suggestion
    private String getClientIp(ContainerRequestContext requestContext) {
        String ipForwarded = Objects.requireNonNullElse(requestContext.getHeaderString("x-forwarded-for"), "");
        String clientIp = ipForwarded.split(",")[0];
        return clientIp.isBlank() ? servletRequest.getRemoteAddr() : clientIp;
    }
```

`String.split` return an array with only one element (the string itself) if no match is found, so if you ensure `ipForwarded` is at least an empty string, you can always take the first element, it is either the first part (if there are more than one),  the only part (if there is only one) or an empty string (if the header was not present). This also reduces the code needed later.

bundles/org.openhab.core.io.rest.auth/src/main/resources/OH-INF/i18n/restauth.properties
```diff
@@ -4,5 +4,7 @@ system.config.restauth.cacheExpiration.label = Cache Expiration Time
 system.config.restauth.cacheExpiration.description = When basic authentication is activated, credentials are put in a cache in order to speed up request authorization. The entries in the cache expire after a while in order to not keep credentials in memory indefinitely. This value defines the expiration time in hours. Set it to 0 for disabling the cache.
 system.config.restauth.implicitUserRole.label = Implicit User Role
 system.config.restauth.implicitUserRole.description = By default, operations requiring the "user" role are available when unauthenticated. Disabling this option will enforce authorization for these operations. Warning: This causes clients that do not support authentication to break.
+system.config.restauth.trustedNetworks.label = Trusted Networks
+system.config.restauth.trustedNetworks.description = Grant implicit user role to requests originated from these networks; Comma separated list of CDIR (ignored if "Implicit User Role" is enabled).
```
J-N-K, line 8:
```suggestion
system.config.restauth.trustedNetworks.description = Grant implicit user role to requests originating from these networks; Comma separated list of CDIR (ignored if "Implicit User Role" is enabled).
```

bundles/org.openhab.core.io.rest.auth/src/main/resources/OH-INF/config/restauth.xml
```diff
@@ -12,6 +12,12 @@
 				authentication to break.</description>
 			<default>true</default>
 		</parameter>
+		<parameter name="trustedNetworks" type="text">
+			<advanced>true</advanced>
+			<label>Trusted Networks</label>
+			<description>Grant implicit user role to requests originated from these networks; Comma separated list of CDIR
```
J-N-K, line 18:
```suggestion
			<description>Grant implicit user role to requests originating from these networks; Comma separated list of CDIR
```

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -267,4 +276,55 @@ public void filter(@Nullable ContainerRequestContext requestContext) throws IOEx
...
         }
     }
+
+    private boolean isImplicitUserRole(ContainerRequestContext requestContext) {
+        if (implicitUserRole) {
+            return true;
+        }
+        if (trustedNetworks.isEmpty()) {
+            return false;
+        }
```
J-N-K, line 286: This is not needed, if `trustedNetworks` is an empty list, `Stream.anyMatch` also returns false.

J-N-K (changes requested):

spacemanspiff2007:
Could you please make a PR that adds this option to the [runtime.cfg](https://github.com/openhab/openhab-distro/blob/199b81e01fabe8281f1ffee801252d813dc4318a/distributions/openhab/src/main/resources/conf/services/runtime.cfg#L81), too?

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
GiviMAD: Changes applied, thank you for the suggestion!
Package for SubnetUtils is not installed, that's why I look up for a function that relies on InetAddress.
Please reopen this if you think it's better to install the package and rely on SubnetUtils.

GiviMAD:
> Could you please make a PR that adds this option to the [runtime.cfg](https://github.com/openhab/openhab-distro/blob/199b81e01fabe8281f1ffee801252d813dc4318a/distributions/openhab/src/main/resources/conf/services/runtime.cfg#L85), too?

Sure, thank you!

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -258,13 +268,52 @@ public void filter(@Nullable ContainerRequestContext requestContext) throws IOEx
...
+        return trustedNetworks.stream().anyMatch(networkCIDR -> isIpInSubnet(clientIp, networkCIDR));
+    }
+
+    private String getClientIp(ContainerRequestContext requestContext) {
+        String ipForwarded = Objects.requireNonNullElse(requestContext.getHeaderString("x-forwarded-for"), "");
+        String clientIp = ipForwarded.split(",")[0];
+        return clientIp.isBlank() ? servletRequest.getRemoteAddr() : clientIp;
+    }
+
+    private boolean isIpInSubnet(final String ip, final String subnetCDIR) {
```
J-N-K, line 295: I think this code should be improved as it does a lot of things that can be pre-parsed. If you do something like

```
private static class CIDR {
    private byte[] networkBytes;
    private long prefix;

    public CIDR(String cidr) {
        String[] parts = cidr.split("/");
        this.prefix = Integer.parseInt(parts[1]);
        this.bytes =  InetAddress.getByName(parts[0]).getAddress();
    }

    public boolean inRange(byte[] address) {
        <... enter your code here ...>
    }
}
```

and then do `byte[] address = InetAddress.getByName(ip).getAddress()` before l. 286, you only have to do the `String/byte[]` conversion only once and the parsing of the network parts only when parsing the config. This saves a ton of compute time for each request.

J-N-K: Another option would be something along the lines of https://www.baeldung.com/java-check-ip-address-range#using-javas-inetaddress-class-for-ipv4, storing the range with start and end values and checking if it is between those. Of course we would need to use  `BigInteger` to make sure this also works for IPv6.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -258,13 +267,73 @@ public void filter(@Nullable ContainerRequestContext requestContext) throws IOEx
...
+    }
+
+    private List<CIDR> parseTrustedNetworks(String value) {
+        var cidrList = new ArrayList<CIDR>();
+        for (var cidrString : value.split(",")) {
+            try {
+                cidrList.add(new CIDR(cidrString.trim()));
+            } catch (UnknownHostException e) {
+                logger.warn("Error parsing trusted networks configuration, disabled.");
+                return List.of();
```
J-N-K, line 300: This discards the whole list, not only this entry. Wouldn't it be better to just log the warning and continue parsing?

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -139,6 +146,8 @@ protected void modified(@Nullable Map<String, Object> properties) {
             allowBasicAuth = value != null && "true".equals(value.toString());
             value = properties.get(CONFIG_IMPLICIT_USER_ROLE);
             implicitUserRole = value == null || !"false".equals(value.toString());
+            value = properties.get(CONFIG_TRUSTED_NETWORKS);
+            trustedNetworks = value == null ? List.of() : parseTrustedNetworks(value.toString());
```
J-N-K, line 150:
```suggestion
            trustedNetworks = parseTrustedNetworks(ConfigParser.valueAsOrElse(properties.get(CONFIG_TRUSTED_NETWORKS), String.class, ""));
```

J-N-K (changes requested):
Otherwise LGTM.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
GiviMAD: Yes, I was not able to decide what to do, I will change it.

J-N-K (approved):
Thanks! FTR: If necessary (which I believe is not), we can add an ExpiringCache for already validated remote IPs to further reduce the load.

PR #3107, opened 2022-10-09
[core.thing] Prefer concurrent collector with parallel streams

ssalonen:
This should avoid behind-the-scenes serial behaviour when collecting to set (the assumption is that Collectors.toSet does not expose CONCURRENT characteristic).

See for more information https://stackoverflow.com/a/41045442

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/link/AbstractLinkRegistry.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/link/AbstractLinkRegistry.java
@@ -17,6 +17,7 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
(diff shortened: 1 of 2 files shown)


ssalonen:
The idea here is to have hopefully performance behaviour -- although I have not benchmarked this.

J-N-K:
```
        List<ItemChannelLink> links2 = IntStream.range(0, 10000)
                .mapToObj(i -> new ItemChannelLink(String.valueOf(i), new ChannelUID("foo:bar:zinga:hu")))
                .collect(Collectors.toList());

        long start = System.nanoTime();
        // Set<String> set = links2.parallelStream().map(AbstractLink::getItemName)
        // .collect(Collectors.toConcurrentMap(Function.identity(), x -> Boolean.TRUE)).keySet();
        Set<String> set = links2.parallelStream().map(AbstractLink::getItemName).collect(Collectors.toSet());
        long stop = System.nanoTime();
        LoggerFactory.getLogger(getClass()).error("time: {}", (stop - start) / 1000000);
```

If I enable the upper code (i.e. the `toConcurrentMap`), it takes about 10ms to complete, with the lower code (i.e. `toSet`) takes about 4ms to complete, so the old code would be 2x faster. This changes if the collections gets larger, for an initial list of 1000000 links, the results are around 130 ms vs. 200 ms. But I believe that 10000 links is even more than the usual user will experience here. I believe the reason is that creating the additional objects is time-consuming and there will only be a benefit if the collection is very large.

ssalonen:
yeah now looking at the code as well, this is not really a bottle neck in practice :) 

The original reason why I paid attention to this is due to some issues with async MQTT client, but I think I need to keep looking

PR #3108, opened 2022-10-10
Support mired units

ccutrer:
Mireds are fairly common to describe the color temperature of lightbulbs (slightly less common than Kelvin), but are very useful for various calculations when adjusting the color temperature, as well as being necessary for various integerations that require mired units.

This commit makes them a well-known unit (previously they were still usable, using "MK^-1"), as well as making them easier to work with on QuantityType. The hiccup is that Mireds aren't technically a Temperature dimension, because they're a reciprocal. So add a `inverse` method that delegates to javax.measure's same method, and then use it as necessary when doing unit conversions and comparisons. Unfortunately, because the dimension changes, the return value of a conversion won't necessarily be the same type, an additional method is added for callers that are willing to handle the change in dimension. This is implemented for all callers that can use it in core.

By implementing this in core, it makes it possible to link a channel that's quantity aware (such as MQTT's number channels with arbitrary units) that expresses it in say mireds, with a Number:Temperature item with a state description in K, seamlessly allowing the UI to display the user-preferred value. Simultaneously it makes it automatic for other add-ons--like Homekit (change incoming to openhab-addons building on this commit), or say the Hue binding--to easily support these conversions, and not need addon-specific metadata to inform it how to do said conversion.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityType.java
@@ -262,7 +262,7 @@ public class QuantityType<T extends Quantity<T>> extends Number
      * Convert this QuantityType to a new {@link QuantityType} using the given target unit.
      *
      * @param targetUnit the unit to which this {@link QuantityType} will be converted to.
-     * @return the new {@link QuantityType} in the given {@link Unit} or {@code null} in case of an erro.
+     * @return the new {@link QuantityType} in the given {@link Unit} or {@code null} in case of an error.
      */
     @SuppressWarnings("unchecked")
     public @Nullable QuantityType<T> toUnit(Unit<?> targetUnit) {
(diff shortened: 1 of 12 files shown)


bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityType.java
```diff
@@ -255,7 +262,7 @@ public Dimension getDimension() {
      * Convert this QuantityType to a new {@link QuantityType} using the given target unit.
      *
      * @param targetUnit the unit to which this {@link QuantityType} will be converted to.
-     * @return the new {@link QuantityType} in the given {@link Unit} or {@code null} in case of a
+     * @return the new {@link QuantityType} in the given {@link Unit} or {@code null} in case of an erro.
```
kaikreuzer, line 265:
```suggestion
     * @return the new {@link QuantityType} in the given {@link Unit} or {@code null} in case of an error.
```

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityType.java
```diff
@@ -283,6 +290,22 @@ public Dimension getDimension() {
         return null;
     }
 
+    /**
+     * Convert this QuantityType to a new {@link QuantityType} using the given target unit.
+     * 
+     * Implicit conversions using inverse units are allowed (i.e. mired <=> Kelvin). This may
+     * changed the dimension.
```
kaikreuzer, line 297:
```suggestion
     * change the dimension.
```

kaikreuzer (changes requested):
Thanks for your contribution and your extensive explanation. Indeed, mired seem to be a "weird" unit due to its reciprocal style, which asks for some special treatment.
From what I can see, the code looks pretty clean to me and should not have any side effects, so I guess it is good to be merged.
I've merely like to ask you to fix two very small typos before merging - thanks!

kaikreuzer (approved):
Thank you!

PR #3110, opened 2022-10-11
Fix AuthFilter crash if trusted network not configured

J-N-K:
Follow-up to #3106 

See also: https://community.openhab.org/t/authfilter-has-arrayindexoutofboundexception/139866/7

The code did not properly handle an empty network. This also adds test for implicit user role handling and trusted networks.

@openhab/core-maintainers I beliebe this is critical as it breaks all installations until either this is merged or the user configures a trusted network (which should not be done in general).

### DIFF
--- a/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
+++ b/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
@@ -294,7 +294,7 @@ public class AuthFilter implements ContainerRequestFilter {
         var cidrList = new ArrayList<CIDR>();
         for (var cidrString : value.split(",")) {
             try {
-                if (!value.isBlank()) {
+                if (!cidrString.isBlank()) {
                     cidrList.add(new CIDR(cidrString.trim()));
                 }
             } catch (UnknownHostException e) {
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
```diff
@@ -295,7 +294,9 @@ private List<CIDR> parseTrustedNetworks(String value) {
         var cidrList = new ArrayList<CIDR>();
         for (var cidrString : value.split(",")) {
             try {
-                cidrList.add(new CIDR(cidrString.trim()));
+                if (!value.isBlank()) {
```
kaikreuzer, line 297: Shouldn't this `if` be moved outside of the `for` loop?

kaikreuzer (commented):
Thanks a lot for the fix and the additional tests.

Just a very minor comment below.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/AuthFilter.java
J-N-K: It depends. If it is moved outside, something like `192.168.0.0/24,,192.168.1.0/24` (empty string in the middle) will generate a warning, if it is like it is now, this will be prevented, but we have more checks. OTOH, since this is only done when parsing the config, I believe it's ok as it is now.

kaikreuzer: I would be able to follow your argument, if the if condition would be
```suggestion
                if (! cidrString.isBlank()) {
```

J-N-K: Yes…

kaikreuzer (approved):
lgtm, thanks!

PR #3111, opened 2022-10-13
Implement command completion in console for openhab command

ccutrer:
And subcommands can implement it if they want. At the least the CommandConsoleExtension's description will now be visible in tab-completion.

I've also implemented detailed completion for openhab:items, openhab:status, openhab:send, and openhab:update.

### DIFF
--- a/bundles/org.openhab.core.io.console.karaf/src/main/java/org/openhab/core/io/console/karaf/internal/CompleterWrapper.java
+++ b/bundles/org.openhab.core.io.console.karaf/src/main/java/org/openhab/core/io/console/karaf/internal/CompleterWrapper.java
@@ -1,0 +1,98 @@
+/**
+ * Copyright (c) 2010-2022 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.io.console.karaf.internal;
+
+import java.util.Arrays;
+import java.util.List;
+
+import org.apache.karaf.shell.api.console.Candidate;
+import org.apache.karaf.shell.api.console.CommandLine;
+import org.apache.karaf.shell.api.console.Session;
+import org.apache.karaf.shell.support.completers.StringsCompleter;
+import org.eclipse.jdt.annotation.NonNull;
+import org.openhab.core.io.console.Completer;
+import org.openhab.core.io.console.extensions.ConsoleCommandExtension;
+
+/**
+ *
+ * @author Cody Cutrer - Initial contribution
+ */
+public class CompleterWrapper implements org.apache.karaf.shell.api.console.Completer {
+
+    private final Completer completer;
+    private final @NonNull String command;
+    private final @NonNull String commandDescription;
+    private final String globalCommand;
+
+    public CompleterWrapper(final @NonNull ConsoleCommandExtension command, boolean scoped) {
+        this.completer = command.getCompleter();
+        this.command = command.getCommand();
+        this.commandDescription = command.getDescription();
+        if (!scoped) {
+            globalCommand = CommandWrapper.SCOPE + ":" + this.command;
+        } else {
+            globalCommand = null;
+        }
+    }
+
+    @Override
+    public int complete(Session session, CommandLine commandLine, List<String> candidates) {
+        String localGlobalCommand = globalCommand;
+        if (commandLine.getCursorArgumentIndex() == 0) {
+            StringsCompleter stringsCompleter = new StringsCompleter();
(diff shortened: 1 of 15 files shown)


bundles/org.openhab.core.io.console/src/main/java/org/openhab/core/io/console/Completer.java
```diff
@@ -0,0 +1,37 @@
...
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
+/**
+ * Implementing this interface allows a {@link ConsoleCommandExtension} to
+ * provide completions for the user as they write commands.
+ *
+ * @author Kai Kreuzer - Initial contribution
+ */
+@NonNullByDefault
+public interface Completer {
```
J-N-K, line 26: I would suggest to rename this class so the conflict with `org.apache.karaf.shell.api.console.Completer` is avoided. WDYT about `ConsoleCommandCompleter`?

bundles/org.openhab.core.io.console.karaf/src/main/java/org/openhab/core/io/console/karaf/internal/CompleterWrapper.java
```diff
@@ -0,0 +1,98 @@
...
+import org.apache.karaf.shell.support.completers.StringsCompleter;
+import org.eclipse.jdt.annotation.NonNull;
+import org.openhab.core.io.console.Completer;
+import org.openhab.core.io.console.extensions.ConsoleCommandExtension;
+
+/**
+ *
+ * @author Cody Cutrer - Initial contribution
+ */
+public class CompleterWrapper implements org.apache.karaf.shell.api.console.Completer {
```
J-N-K, line 30: Is it possible to annotate the class with `@NonNullByDefault`? Also please add a class description in JavaDoc

bundles/org.openhab.core.io.console/src/main/java/org/openhab/core/io/console/StringsCompleter.java
```diff
@@ -0,0 +1,88 @@
...
+ * Completer for a set of strings
+ *
+ * @author Cody Cutrer - Initial contribution
+ */
+@NonNullByDefault
+public class StringsCompleter implements Completer {
+    private final SortedSet<String> strings;
+    private final boolean caseSensitive;
+
+    public StringsCompleter() {
```
J-N-K, line 33: Why do you need so many constructors? IMO the default constructor and `public StringsCompleter(final Collection<String> strings, boolean caseSensitive)` would be sufficient.

bundles/org.openhab.core.io.console/src/main/java/org/openhab/core/io/console/extensions/ConsoleCommandExtension.java
```diff
@@ -53,4 +55,11 @@ public interface ConsoleCommandExtension {
      * @return the help texts for this extension
      */
     List<String> getUsages();
+
+    /**
+     * @return a {@link Completer} object for this command
```
J-N-K, line 60: Please add a method description

J-N-K (changes requested):
In general looks good and a nice addition. Please check that you have JavaDoc for all `public` methods and add some tests.

bundles/org.openhab.core.io.console.karaf/src/main/java/org/openhab/core/io/console/karaf/internal/CompleterWrapper.java
ccutrer: I was having trouble with that. The interface is not `@NonNullByDefault`, and I couldn't figure out the proper encantation of `@Nullable` on the overridden methods to make it compatible. And given that the only methods in this are overridden from the interface, I didn't think it was worth it.

bundles/org.openhab.core.io.console/src/main/java/org/openhab/core/io/console/StringsCompleter.java
ccutrer: 🤷  this class is mostly based off of (one of the several) StringsCompleter classes from karaf, and they had a bunch of constructors too. I have no problem trimming it down.

bundles/org.openhab.core.io.console.karaf/src/main/java/org/openhab/core/io/console/karaf/internal/CompleterWrapper.java
J-N-K: Ok. Probably `@NonNullByDefault({})` would have done the trick.

ccutrer:
I've addressed the comments so far, except for writing tests. I don't have time to do that today. Hopefully within the next handful of days.

J-N-K (approved):
LGTM, thanks.

lolodomo:
This is a really cool feature, bravo.
We have console commands implemented in 14 bindings.
I am using and maintaining (or able to maintain) 7 of them. So I will extend your solution to the following bindings: hue, netatmo, linky, powermax, freebox, lgwebos, somfytahoma.

PR #3116, opened 2022-10-17
Allow sending IncreaseDecreaseType directly to DimmerItem

ccutrer:


kaikreuzer (approved):
Thanks. I am puzzled that this hasn't been there yet for all the years and whether there is any reason for that (which I'm not seeing now)?

hmerk:
Indeed, that‘s strange, as e.g. Wemo Dimmer brightness channel accepts increase/decrease command since it was implemented years ago. Same for other bindings.

ccutrer:
Rules generally use the message bus to send commands, which doesn't use these methods. The HomeKit addon for reasons I'm unsure of calls send directly on items. I've been adding more accessory types to HomeKit, and I've run into a couple of these direct-send overloads that didn't exist.

PR #3124, opened 2022-10-21
[mqtt] Properly process retained messages

ccutrer:
Only the _first_ message has the retained flag (always on MQTT3, there is a flag to forward the retain flag on all messages with MQTT5, but we don't set it), so we have to keep track of previously retained messages as well.

This shows up as when a thing comes back online and re-subscribes to a retained message that another thing subscribed to also, it won't see any changes that happened to that topic since the very first message was received.

### DIFF
--- a/bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/openhab/core/io/transport/mqtt/internal/Subscription.java
+++ b/bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/openhab/core/io/transport/mqtt/internal/Subscription.java
@@ -44,8 +44,11 @@ public class Subscription {
     public void add(MqttMessageSubscriber subscriber) {
         if (subscribers.add(subscriber)) {
             // new subscriber. deliver all known retained messages
-            retainedMessages.entrySet().parallelStream()
-                    .forEach(entry -> processMessage(subscriber, entry.getKey(), entry.getValue()));
+            retainedMessages.entrySet().parallelStream().forEach(entry -> {
+                if (entry.getValue().length > 0) {
+                    processMessage(subscriber, entry.getKey(), entry.getValue());
+                }
+            });
         }
     }
 
@@ -71,12 +74,11 @@ public class Subscription {
     }
 
     public void messageArrived(String topic, byte[] payload, boolean retain) {
-        if (retain) {
-            if (payload.length > 0) {
-                retainedMessages.put(topic, payload);
-            } else {
-                retainedMessages.remove(topic);
-            }
+        // http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc385349265
+        // Only the first message delivered will have the retain flag; subsequent messages
+        //
+        if (retain || retainedMessages.containsKey(topic)) {
+            retainedMessages.put(topic, payload);
         }
         subscribers.parallelStream().forEach(subscriber -> processMessage(subscriber, topic, payload));
     }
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/openhab/core/io/transport/mqtt/internal/Subscription.java
```diff
@@ -71,12 +74,11 @@ public void messageArrived(Mqtt5Publish message) {
...
     public void messageArrived(String topic, byte[] payload, boolean retain) {
-        if (retain) {
-            if (payload.length > 0) {
-                retainedMessages.put(topic, payload);
-            } else {
-                retainedMessages.remove(topic);
-            }
+        // http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc385349265
+        // Only the first message delivered will have the retain flag; subsequent messages
+        //
```
J-N-K, line 79: There seems to be something missing here

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/openhab/core/io/transport/mqtt/internal/Subscription.java
```diff
@@ -44,8 +44,11 @@ public class Subscription {
     public void add(MqttMessageSubscriber subscriber) {
         if (subscribers.add(subscriber)) {
             // new subscriber. deliver all known retained messages
-            retainedMessages.entrySet().parallelStream()
-                    .forEach(entry -> processMessage(subscriber, entry.getKey(), entry.getValue()));
+            retainedMessages.entrySet().parallelStream().forEach(entry -> {
+                if (entry.getValue().length > 0) {
```
J-N-K, line 48: Why is this necessary? Shouldn't the message always have a length? Otherwise it makes no sense to retain it.

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/openhab/core/io/transport/mqtt/internal/Subscription.java
```diff
@@ -71,12 +74,11 @@ public void messageArrived(Mqtt5Publish message) {
...
-        if (retain) {
-            if (payload.length > 0) {
-                retainedMessages.put(topic, payload);
-            } else {
-                retainedMessages.remove(topic);
-            }
+        // http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc385349265
+        // Only the first message delivered will have the retain flag; subsequent messages
+        //
+        if (retain || retainedMessages.containsKey(topic)) {
```
J-N-K, line 80: Maybe add the payload length check here

J-N-K (changes requested):

bundles/org.openhab.core.io.transport.mqtt/src/main/java/org/openhab/core/io/transport/mqtt/internal/Subscription.java
ccutrer: because we need to know that it's been retained in the past to continue retaining it in the future, even though it won't have the retained flag on future messages. so we still keep it in retainedMessages, but don't send it to new subscriptions if it's currently zero

J-N-K (approved):
LGTM, thanks.

PR #3129, opened 2022-10-23
update system.color-temperature-abs channel type to be a Number:Temperature

ccutrer:
refs https://github.com/openhab/openhab-addons/pull/13570#issuecomment-1288139778

now that Kelvin and mireds/mireks are easily convertible

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/DefaultSystemChannelTypeProvider.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/DefaultSystemChannelTypeProvider.java
@@ -241,7 +241,7 @@ public class DefaultSystemChannelTypeProvider implements ChannelTypeProvider {
      * Color-temperature: default system wide {@link ChannelType} which allows changing the color temperature in Kelvin
      */
     public static final ChannelType SYSTEM_COLOR_TEMPERATURE_ABS = ChannelTypeBuilder
-            .state(SYSTEM_CHANNEL_TYPE_UID_COLOR_TEMPERATURE_ABS, "Color Temperature", CoreItemFactory.NUMBER)
+            .state(SYSTEM_CHANNEL_TYPE_UID_COLOR_TEMPERATURE_ABS, "Color Temperature", "Number:Temperature")
             .withDescription("Controls the color temperature of the light in Kelvin").withCategory("ColorLight")
             .isAdvanced(true)
             .withStateDescriptionFragment(StateDescriptionFragmentBuilder.create().withMinimum(new BigDecimal(1000))


J-N-K:
Will this break linking `Number` channels in the UI to channels with this type?

ccutrer:
> Will this break linking `Number` channels in the UI to channels with this type?

No, I had no problem linking with a Number item. It just defaults to a Number:Temperature now when you created it from the channel.

J-N-K (approved):

lolodomo:
This change was very dangerous and could have even broken the hue binding (to be confirmed). 
All bindings using this channel type were certainly expecting a DecimalType as command before this change. 
With this change, they should also accept a QuantityTyoe now. 
A check of all bindings using this channel type should have been done. 
I hope a DecimalType is still provided when the item type is a simple Number... If not, several bindings are broken.

ccutrer:
Bindings using this channel type:
 * Dali
 * Hue
 * Lifx
 * Nanoleaf
 * MQTT.espmilighthub (PR pending)

This change was made in connection with the Hue binding maintainers, though it was assumed at that point that the large refactor would be done by 3.4.0 (and that this PR would be a pre-req).

It was also specifically tested that if you've already linked an item to a channel of this type, it will stay as a plain number item, and as a plain number item it will only accept and forward DecimalType, not QuantityType. Newly linked items though will default to a Number:Temperature. Any update provided by a binding will accept a DecimalType, and convert to a QuantityType of the item's unit (which is already noted as a breaking change in 3.4.0 that bindings should provide a proper state description for quantity type channels. it's arguable if this "breaking" disclaimer applies to binding authors, since it wasn't a quantity typed channel at the beginning of the cycle). As for commands... it just depends on how the user sends them. There are probably some cases yes where what was previously a DecimalType will now get sent as a QuantityType. I'll go look at the listed bindings to see how they might react to this. But that won't be tonight.

lolodomo:
Critical issue to read: https://community.openhab.org/t/oh-3-4-hue-absolute-temperature/142266/13

@J-N-K: do you understand why "K" Is added as unit to the command state in case the item type is Number? I believe you changes something about that but I do not find your PR anymore!

lolodomo:
> Newly linked items though will default to a Number:Temperature

This will be the case for anyone using config files as linked channels are built at OH startup in that case.

J-N-K:
The `CommunicationManager` automatically converts from `DecimalType` to `QuantityType` if the accepted item type has a dimension. It determines the unit from the linked item (or if that fails from the accepted item type). I believe the state description in your case contains a unit and that one is taken. The default unit for `Number:Temperature` would be `°C`.

ccutrer:
> > Newly linked items though will default to a Number:Temperature
> 
> This will be the case for anyone using config files as linked channels are built at OH startup in that case.

If you're using config files for both things and items, then your item type will _not_ change, since you didn't change your .items file. When I say "newly linked", I was referring to MainUI offering to create an item for you from a channel.

cweitkamp:
The unit is `K` because the patterns defined for the system default channel type is: `"%.0f K"`

https://github.com/openhab/openhab-core/blob/41ba3ff854797f98293c12ee0f388c1972c75bd1/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/DefaultSystemChannelTypeProvider.java#L243-L249

lolodomo:
> The unit is `K` because the patterns defined for the system default channel type is: `"%.0f K"`

Thank you for the explanation.

ccutrer:
> Bindings using this channel type:
> 
> * Dali
> * Hue
> * Lifx
> * Nanoleaf

Okay, we've got them all now. I did Dali, Lifx, and Nanoleaf, and @cweitkamp did Hue.

PR #3133, opened 2022-10-25
add helper methods for working up the ChannelUID tree

ccutrer:


bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/ChannelUID.java
```diff
@@ -149,4 +149,22 @@ public ThingUID getThingUID() {
...
+    }
+
+    /**
+     * Returns the thing type UID.
+     * 
+     * @return thing type UID
+     */
+    public ThingTypeUID getThingTypeUID() {
+        return new ThingTypeUID(getSegment(0), getSegment(1));
+    }
```
J-N-K, line 169: I don't think these are needed. You can get them via `getThingUID().getBindingId()`.

ccutrer: True, but that involves an extra intermediate object, and is more to type. And to make it uniform, you shouldn't have `getBindingId()` on `ThingUID`, only on `ThingTypeUID`, so this call would then become `getThingUID().getThingTypeUID().getBindingId()`. Quite the mouthful.

J-N-K: Since it is possible to construct a ThingUID without thing type, it‘s debatable if we can assume it’s always possible to determine the ThingTypeUID.

J-N-K: Maybe you could check if `segment[1]` equals `NO_THING_TYPE` and if this is the case return `null` (or make the return value an `Optional<ThingTypeUID>` and return `Optional.empty()`).

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/ChannelUID.java
```diff
@@ -149,4 +149,22 @@ public ThingUID getThingUID() {
         List<String> allSegments = getAllSegments();
         return new ThingUID(allSegments.subList(0, allSegments.size() - 1).toArray(new String[allSegments.size() - 1]));
     }
+
+    /**
+     * Returns the binding id.
+     * 
+     * @return binding id
+     */
+    public String getBindingId() {
```
cweitkamp, line 158: I am wondering why this is needed? This class extends `UID` and there already is a `public` method with the same name available. Or did I miss something?

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/ThingUID.java
```diff
@@ -149,6 +149,24 @@ public List<String> getBridgeIds() {
         return allSegments.subList(2, allSegments.size() - 1);
     }
 
+    /**
+     * Returns the binding id.
+     * 
+     * @return binding id
+     */
+    public String getBindingId() {
```
cweitkamp, line 157: Same here.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/ThingTypeUID.java
```diff
@@ -44,6 +44,20 @@ protected int getMinimalNumberOfSegments() {
         return 2;
     }
 
+    /**
+     * Returns the binding id.
+     * 
+     * @return binding id
+     */
+    public String getBindingId() {
```
cweitkamp, line 52: Same here.

ccutrer:
@cweitkamp : heh, thanks. I actually noticed that yesterday. When I wrote this I was thinking they inherited from AbstractUID. I need to update this PR, but I've been busy with other projects.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/ChannelUID.java
kaikreuzer: You actually cannot infer the thing type from a channel or thing UID. Although it is in most cases the second segment, it is possible that a Thing is migrated to a different type while it keeps its UID. This mechanism has been introduced for Zigbee/ZWave and is in use by a small number of other bindings. Therefore, such a method simply must not exist and there is no correct way to implement it.

J-N-K: I didn't think of that (even if I use migrate thing type quite often).

PR #3134, opened 2022-10-25
fix invertible unit conversions when sent via a binding

ccutrer:


J-N-K:
Did you check if there are other places in the code where this would be necessary?

J-N-K (approved):
Thanks, LGTM

PR #3136, opened 2022-10-27
Fix ItemStateCondition for PercentType values

J-N-K:
Dimmer items return their state as PercentType. The code did not properly handle PercentType because PercentType is a subclass of DecimalType. One value was converted by casting the value to DecimalType (which leads to 40 for 40%) while the other one used `.as` (which results in 0.4 for 40%). A comparison "state <= 50" with state = 20 is therefore executed as "20 <= 0.5"  which is not what the user would expect.

Reported on the forum.


cweitkamp (approved):
LGTM. Thank you.

Can you add a small unit test? Test class already is in place.

J-N-K:
done

cweitkamp (approved):
Thank you.

PR #3138, opened 2022-10-30
Add DateTime handling to ItemStateConditionHandler

J-N-K:
This adds the capability to handle `DateTime` items to the `ItemStateConditionHandler`. The given state can be

- the string representation of a `ZonedDateTime` (e.g. `2022-01-15T17:23:15+02:00`)
- the string representation of a `LocalDateTime` (e.g. `2022-11-06T16:05:10`), the configured time zone from the `TimeZoneProvider` is used
- a period that is applied to the current time (`ZonedDateTime.now()`). The period can contain days (`D`), hours (`H`), minutes (`M`) or seconds (`S`) or an ordered combination of these, including a sign (e.g. `1D-2H` would be the same as `ZonedDateTime.now().plusDays(1).minusHours(2)`).

I didn't find the documentation for the conditions in general, can someone point me to the correct place so that documentation for this can be added?

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/factory/CoreModuleHandlerFactory.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/factory/CoreModuleHandlerFactory.java
@@ -40,6 +40,7 @@ import org.openhab.core.automation.internal.module.handler.RunRuleActionHandler;
 import org.openhab.core.automation.internal.module.handler.SystemTriggerHandler;
 import org.openhab.core.automation.internal.module.handler.ThingStatusTriggerHandler;
 import org.openhab.core.events.EventPublisher;
+import org.openhab.core.i18n.TimeZoneProvider;
 import org.openhab.core.items.ItemRegistry;
 import org.osgi.framework.BundleContext;
 import org.osgi.service.component.annotations.Activate;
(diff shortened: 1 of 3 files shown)


kaikreuzer (approved):
lgtm, thank you.

> I didn't find the documentation for the conditions in general, can someone point me to the correct place so that documentation for this can be added?

Since you didn't find it, I am afraid the most logical answer is that such a place does not yet exist.
Afaik, all we have right now is the description field of the modules, which at least provide some short documentation at runtime.
But the more modules we have, I agree that we should have a place for them in the docs as well. Maybe @Confectrician already has an idea about a good structure?

rkoshak:
> Maybe @Confectrician already has an idea about a good structure?

I don't want to speak for @Confectrician but if I understand correctly what this is, I think the best place may be in the new sections for rules docs that I and @florian-h05 are working on at https://github.com/openhab/openhab-docs/issues/1855.

Unfortunately we haven't got very far yet and it's slow going but I would expect this to be documented either under section 1c or under section 3 (depending on whether this is a UI rule only concept of something universal to all rules. I think it's the former so I'd put it under section 3 where we plan on putting all the reference information for creating UI rules.

In the mean time I'm not sure there is a good place yet and it'll need a new home.

Our intent is to submit each major section (1, 2, 3, etc.) as separate PRs to break it up and so we don't have to wait forever to start getting the benefit. So section 3 could be started and worked on independently if that makes sense.

florian-h05:
@rkoshak 
Could you add a reference to this PR in the tracking issue https://github.com/openhab/openhab-docs/issues/1855?

PR #3139, opened 2022-11-03
[audio|voice] Add actions/commands to synthesize melodies and add configurable melody to the dialog processor

GiviMAD:
With the current capabilities the user is not aware of whetter a spot event is fired (unless is checking the listeningItem).

This pr add a small wav synthesizer to the audio utils and take advantage of it whenever is possible to add sounds to the dialog processing flow.

The <s>note</s> melody to be played on spot can be configured though the voice configuration.

Tested with the javasound and pulseaudio sinks, and should be compatible with the new WebAudio Sink.

<s>I think that adding audio actions to allow the user to manually sent notes to sinks is also interesting, but I prefer to implement that on a future PR if it's ok. </s>

Also added it to the audio manager and add action/command to use it.


GiviMAD:
I've been playing with this a lot, I'll add another commit with the audio action/command.

kaikreuzer (approved):
Lgtm and the feature sounds very nice, thanks!

PR #3143, opened 2022-11-05
Add default units for all dimensions

### LINKED ISSUE #3096: [persistence] previousState(true) returns current value when there is an unit applied on channel
I've changed persistence engine from `RRD4j` to `JDBC` (`mariadb`) sice I've read somewhere that `previousState` is not working with `RRD4j`.

Unfortunately, `previousState(true)` does not work correctly for channels with unit of measurement set.

```
function PrintCurrentAndLastUpdate(thingName)
{
  var item = itemRegistry.getItem(thingName);
  
  logger.info(thingName + " previous state:\t" + persistence.previousState(item, true));
  logger.info(thingName + " last update:\t" + persistence.lastUpdate(item));
  logger.info(thingName + " current state:\t" + item.getState());
}

PrintCurrentAndLastUpdate("Rtl433_consumed_water");
PrintCurrentAndLastUpdate("Rtl433_time");
PrintCurrentAndLastUpdate("Rtl433_rssi");
```

```
2022-09-29 12:47:14.495 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Rtl433_time' changed from 2022-09-29T12:46:44.000+0200 to 2022-09-29T12:47:14.000+0200
2022-09-29 12:47:14.496 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Rtl433_rssi' changed from -0.112511 dBm to -0.213509 dBm

==> /var/log/openhab/openhab.log <==
2022-09-29 12:47:14.521 [INFO ] [org.openhab.rule.water_volume_change] - Rtl433_time previous state:    JdbcItem [name=Rtl433_time, state=2022-09-29T12:46:44.000+0200, timestamp=2022-09-29T12:46:44.457+02:00[Europe/Warsaw]]
2022-09-29 12:47:14.522 [INFO ] [org.openhab.rule.water_volume_change] - Rtl433_time last update:       2022-09-29T12:47:14.492+02:00[Europe/Warsaw]
2022-09-29 12:47:14.522 [INFO ] [org.openhab.rule.water_volume_change] - Rtl433_time current state:     2022-09-29T12:47:14.000+0200
2022-09-29 12:47:14.524 [INFO ] [org.openhab.rule.water_volume_change] - Rtl433_rssi previous state:    JdbcItem [name=Rtl433_rssi, state=-0.213509, timestamp=2022-09-29T12:47:14.496+02:00[Europe/Warsaw]]
2022-09-29 12:47:14.525 [INFO ] [org.openhab.rule.water_volume_change] - Rtl433_rssi last update:       2022-09-29T12:47:14.496+02:00[Europe/Warsaw]
2022-09-29 12:47:14.525 [INFO ] [org.openhab.rule.water_volume_change] - Rtl433_rssi current state:     -0.213509 dBm
```

without unit set (see rssi):
```
2022-09-29 12:47:50.420 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Rtl433_time' changed from 2022-09-29T12:47:14.000+0200 to 2022-09-29T12:47:50.000+0200
2022-09-29 12:47:50.420 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Rtl433_rssi' changed from -0.213509 to -0.125042

==> /var/log/openhab/openhab.log <==
2022-09-29 12:47:50.445 [INFO ] [org.openhab.rule.water_volume_change] - Rtl433_time previous state:    JdbcItem [name=Rtl433_time, state=2022-09-29T12:47:14.000+0200, timestamp=2022-09-29T12:47:14.492+02:00[Europe/Warsaw]]
2022-09-29 12:47:50.446 [INFO ] [org.openhab.rule.water_volume_change] - Rtl433_time last update:       2022-09-29T12:47:50.424+02:00[Europe/Warsaw]
2022-09-29 12:47:50.446 [INFO ] [org.openhab.rule.water_volume_change] - Rtl433_time current state:     2022-09-29T12:47:50.000+0200
2022-09-29 12:47:50.447 [INFO ] [org.openhab.rule.water_volume_change] - Rtl433_rssi previous state:    JdbcItem [name=Rtl433_rssi, state=-0.213509, timestamp=2022-09-29T12:47:18.177+02:00[Europe/Warsaw]]
2022-09-29 12:47:50.447 [INFO ] [org.openhab.rule.water_volume_change] - Rtl433_rssi last update:       2022-09-29T12:47:50.421+02:00[Europe/Warsaw]
2022-09-29 12:47:50.447 [INFO ] [org.openhab.rule.water_volume_change] - Rtl433_rssi current state:     -0.125042
```

Using openHAB 3.3.0

### LINKED ISSUE #3121: default UoM missing
There's operational states when an UoM item (say Number:Power) has a value but no unit such as after mapdb restoreOnStartup (as persistence generally does not store the unit).
This state might go unnoticed until e.g. the item is cast to the proper type in rulesDSL. rulesDSL refuses to cast and the rule crashes. Sad but correctly so because it does not know which unit to apply (in the Number:Power example whether to apply W or kW or mW).

We need a default unit to apply whenever a UoM item is updated and does not have a unit yet. It is imho not acceptable to leave that problem to the rule writer (incredibly hard to catch and work around everywhere it could surface).
The default unit to apply can be derived from the UoM (when there's none it is reasonable to assume the base unit i.e. W in the example).

### LINKED ISSUE #3150: After sending postUpdate(0) to items, the items show inconsistently the units in the set value
In openHAB 3.3 I have four channels:

```
Thing mqtt:topic:r3 "Реле 3" (mqtt:broker:5cc410aefc) [availabilityTopic="r3/online", payloadNotAvailable="false", payloadAvailable="true"] {
  Channels:
    Type number  : r3_current     "Реле 3" [stateTopic="r3/status/switch:0", transformationPattern="JSONPATH:$.current", unit="A"]
    Type number  : r3_power       "Реле 3" [stateTopic="r3/status/switch:0", transformationPattern="JSONPATH:$.apower", unit="VA"]
    Type number  : r3_temperature "Реле 3" [stateTopic="r3/status/switch:0", transformationPattern="JSONPATH:$.temperature.tC", unit="°C"]
    Type number  : r3_voltage     "Реле 3" [stateTopic="r3/status/switch:0", transformationPattern="JSONPATH:$.voltage", unit="V"]
}
```
and four items:
```
Number:ElectricCurrent    r3_current     "Реле 3 ток"         ["Current"] {channel="mqtt:topic:r3:r3_current"}
Number:Power              r3_power       "Реле 3 мощност"     ["Power"] {channel="mqtt:topic:r3:r3_power"}
Number:Temperature        r3_temperature "Реле 3 температура" ["Temperature"] {channel="mqtt:topic:r3:r3_temperature"}
Number:ElectricPotential  r3_voltage     "Реле 3 напрежение"  ["Voltage"] {channel="mqtt:topic:r3:r3_voltage"}
```
I send to all of them over openhab-js `….postUpdate(0)`.

Then the UI shows under http://…/settings/things/mqtt:topic:r3
r3_power 0
r3_current 0
r3_temperature 0 °C
r3_voltage 0

As can be seen, except for temperature, the units are not shown.

J-N-K:
Fixes #3150 
Fixes #3121 
Fixes #3096 

It would be great if someone from the US could check if I did pick good units for the imperial measurement system or if something is missing there. 

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/i18n/I18nProviderImpl.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/i18n/I18nProviderImpl.java
@@ -25,12 +25,39 @@ import java.util.ResourceBundle;
 
 import javax.measure.Quantity;
 import javax.measure.Unit;
+import javax.measure.quantity.Acceleration;
+import javax.measure.quantity.AmountOfSubstance;
 import javax.measure.quantity.Angle;
+import javax.measure.quantity.Area;
+import javax.measure.quantity.CatalyticActivity;
 import javax.measure.quantity.Dimensionless;
+import javax.measure.quantity.ElectricCapacitance;
+import javax.measure.quantity.ElectricCharge;
+import javax.measure.quantity.ElectricConductance;
+import javax.measure.quantity.ElectricCurrent;
+import javax.measure.quantity.ElectricInductance;
+import javax.measure.quantity.ElectricPotential;
+import javax.measure.quantity.ElectricResistance;
+import javax.measure.quantity.Energy;
+import javax.measure.quantity.Force;
+import javax.measure.quantity.Frequency;
+import javax.measure.quantity.Illuminance;
 import javax.measure.quantity.Length;
+import javax.measure.quantity.LuminousFlux;
+import javax.measure.quantity.LuminousIntensity;
+import javax.measure.quantity.MagneticFlux;
+import javax.measure.quantity.MagneticFluxDensity;
+import javax.measure.quantity.Mass;
+import javax.measure.quantity.Power;
 import javax.measure.quantity.Pressure;
+import javax.measure.quantity.RadiationDoseAbsorbed;
+import javax.measure.quantity.RadiationDoseEffective;
+import javax.measure.quantity.Radioactivity;
+import javax.measure.quantity.SolidAngle;
 import javax.measure.quantity.Speed;
 import javax.measure.quantity.Temperature;
+import javax.measure.quantity.Time;
+import javax.measure.quantity.Volume;
 import javax.measure.spi.SystemOfUnits;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
@@ -40,7 +67,13 @@ import org.openhab.core.i18n.LocationProvider;
 import org.openhab.core.i18n.TimeZoneProvider;
 import org.openhab.core.i18n.TranslationProvider;
 import org.openhab.core.i18n.UnitProvider;
+import org.openhab.core.library.dimension.ArealDensity;
+import org.openhab.core.library.dimension.DataAmount;
+import org.openhab.core.library.dimension.DataTransferRate;
+import org.openhab.core.library.dimension.Density;
+import org.openhab.core.library.dimension.ElectricConductivity;
 import org.openhab.core.library.dimension.Intensity;
+import org.openhab.core.library.dimension.VolumetricFlowRate;
 import org.openhab.core.library.types.PointType;
 import org.openhab.core.library.unit.ImperialUnits;
 import org.openhab.core.library.unit.SIUnits;
@@ -107,7 +140,7 @@ public class I18nProviderImpl
     private @Nullable ZoneId timeZone;
 
     // UnitProvider
-    private static final String MEASUREMENT_SYSTEM = "measurementSystem";
+    static final String MEASUREMENT_SYSTEM = "measurementSystem";
     private @Nullable SystemOfUnits measurementSystem;
     private final Map<Class<? extends Quantity<?>>, Map<SystemOfUnits, Unit<? extends Quantity<?>>>> dimensionMap = new HashMap<>();
 
@@ -122,7 +155,7 @@ public class I18nProviderImpl
     }
 
     @Deactivate
-    protected void deactivate(ComponentContext componentContext) {
+    protected void deactivate() {
         this.resourceBundleTracker.close();
     }
 
@@ -319,8 +352,8 @@ public class I18nProviderImpl
         return text;
     }
 
-    @SuppressWarnings("unchecked")
     @Override
+    @SuppressWarnings("unchecked")
     public <T extends Quantity<T>> @Nullable Unit<T> getUnit(@Nullable Class<T> dimension) {
         Map<SystemOfUnits, Unit<? extends Quantity<?>>> map = dimensionMap.get(dimension);
         if (map == null) {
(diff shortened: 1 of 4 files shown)


rkoshak:
Use "Feet" for length unless having the default units close to the same size between SI and Imperial is important. If so use "Yard" which is closest to a meter in length.

What I don't know is if it makes sense to provide imperial units for some of the more scientific units like acceleration, density, and the like.

Everything else looks good.

J-N-K:
@rkoshak For acceleration it probably makes sense, because we also have different units for speed. I'm not so sure about density, my guess would be that the most common use-case here is absolute humidity. What is the usual unit to express that?

rkoshak:
In all my years I've never been in a situation where I've ever used or seen absolute humidity used. It's affected by so many other variables it's not all that useful (IIRC). Beyond that I'm not sure where I've ever seen these in use day-to-day. Maybe in measuring how full a propane tank. I did some looking and it looks like pounds per cubic foot is most commonly used. 

I'm not sure what makes sense for acceleration. Since feet is the base unit we use I guess feet per second squared would make sense as a default here. I can't imagine in a home automation context where we'd care about acceleration in yards or miles per second squared by default.

I've only ever encountered this sort of stuff in math and physics classes though and we used SI units even twenty-five years ago when I last took them.

J-N-K:
I guess we should leave it as we have it now. BTW: I use absolute humidity to check if it makes sense to use the bathroom fan. If the absolute humidity outside is higher than inside, don't turn it on even if the relative humidity is higher than the set value.

kaikreuzer (approved):
lgtm, thanks!

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/i18n/I18nProviderImpl.java
```diff
@@ -344,39 +377,53 @@ public SystemOfUnits getMeasurementSystem() {
...
+        addDefaultUnit(Dimensionless.class, Units.ONE);
+        addDefaultUnit(ElectricCapacitance.class, Units.FARAD);
+        addDefaultUnit(ElectricCharge.class, Units.COULOMB);
+        addDefaultUnit(ElectricConductance.class, Units.SIEMENS);
+        addDefaultUnit(ElectricConductivity.class, Units.SIEMENS_PER_METRE);
+        addDefaultUnit(ElectricCurrent.class, Units.AMPERE);
+        addDefaultUnit(ElectricInductance.class, Units.HENRY);
+        addDefaultUnit(ElectricPotential.class, Units.VOLT);
+        addDefaultUnit(ElectricResistance.class, Units.OHM);
+        addDefaultUnit(Energy.class, Units.JOULE);
```
Hilbrand, line 398: Could it be this change causes all kind of channels that are `kWh` to display in `J`, which is not expected. Is there a fix possible?

Hilbrand: Could this mean `%unit%` should not be used in binding xml definitions anymore?

J-N-K: Could be. However, there is no way to avoid that and it was always the case with the dimensions that had a default unit before.

The state description takes precedence over the default, so we could check if we need to adjust `%unit%`. I'll have a look.

kaikreuzer: @J-N-K The difference to other dimensions is that this here now is a breaking change - many of my items now show `J` instead of `KWh` and even a couple of my rules are impacted. I assume @Hilbrand (and many others) have a similar situation.
I wonder whether we should maybe define `kWh` as a default for `Energy` - after all, this is the unit that is almost always used in a smart home context, while `J` is rather a very rare exception
Wdyt?

Hilbrand: @kaikreuzer for the channels with `%unit%` my issue was solved in #3201.

spacemanspiff2007: While the `J` might be the standard unit I think in a home automation context `kWh`should absolutely be the default.
99% of the time users want to compare/measure energy consumption of devices and that's typically done in `kWh`

PR #3144, opened 2022-11-06
Fix missing stateDescriptionProvider in Group items

### LINKED ISSUE #2449: Group SUM aggregation does not inherit the member's UoM
A Group containing members with a UoM setting that differs from the 'default' will not use the members' UoM setting in the SUM aggregation. This probably also applies to the AVG aggregation since it also generates a new value (as opposed to returning a matching Item's state when aggregating with MIN or MAX).

An example is provided below:

The Buienradar binding exposes rain intensity forecasts as `Number:Speed` expressed in **mm/h**.  This is also 'inherited' by the `Item`s linked to these binding channels: the relating `Item`s are of subtype `Number:Speed` and the state is expressed in **mm/h**.

When I create a `Group` and assign all rain intensity `Item`s from the Buienradar binding, and when I set the `SUM` aggregation, the Group uses a different UoM than the individual members, even if I set the member's subtype to 'Number:Speed`. The Group reports a state in **km/h** (1 million bigger unit than the individual members' unit)

Where an `Item` can be configured to have an explicit UoM for the quantity represented in the state, this is apparently not possible with a `Group`.

For Main UI widgets you can of course use the `displayState` property which behaves similarly to what the `label="..."` construct allows in sitemaps.

See also the following discussion:
https://community.openhab.org/t/oh3-group-item-number-speed-defaults-to-km-h-how-to-set-item-uom-e-g-to-mm-h/125087

J-N-K:
Fixes #2449 

The stateDescriptionProvider was not properly passed to the base item. Because of that units set in the state description were not properly used when setting the griup state after calculating aggregation functions.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ItemRegistryImpl.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ItemRegistryImpl.java
@@ -293,14 +293,14 @@ public class ItemRegistryImpl extends AbstractRegistry<Item, String, ItemProvide
     }
 
     @Reference(cardinality = ReferenceCardinality.OPTIONAL, policy = ReferencePolicy.DYNAMIC)
-    protected void setUnitProvider(UnitProvider unitProvider) {
+    public void setUnitProvider(UnitProvider unitProvider) {
         this.unitProvider = unitProvider;
         for (Item item : getItems()) {
             ((GenericItem) item).setUnitProvider(unitProvider);
         }
     }
 
-    protected void unsetUnitProvider(UnitProvider unitProvider) {
+    public void unsetUnitProvider(UnitProvider unitProvider) {
         this.unitProvider = null;
         for (Item item : getItems()) {
             ((GenericItem) item).setUnitProvider(null);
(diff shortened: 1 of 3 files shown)


cweitkamp:
Nice catch. I am wondering if we should override the getter methods too and return the `baseItem`s `StateDescription`or `Command` provider?

J-N-K:
I guess that's not needed as they are the same. The base item is not allowed to be accessed except for instance checks (according to the JavaDoc).

cweitkamp:
Hm, are you sure? If we override the setter and do not set the internal property it will be `null`. Wouldn't it?

In is a nice place to add a test case.

https://github.com/openhab/openhab-core/blob/ad936cd83f70d9ac9732f8e1b94acc04939fe9ed/itests/org.openhab.core.tests/src/main/java/org/openhab/core/items/ItemRegistryImplTest.java#L383

cweitkamp:
Forget my comment. I missed the super calls.

J-N-K:
But we call `super.setXXX`, which sets XXX to the group itself.

cweitkamp (approved):
LGTM. Thanks.

Will you add a unit test?

cweitkamp (approved):

PR #3145, opened 2022-11-06
Add countSince/countBetween to persistence extensions

### LINKED ISSUE #944: FR: Persisitence - countSince
Use-Case:
I have an item, low_battery_alarm which is a switch. If triggered, I switch all appliances off (Solar house). 
When the alarm state clears I want to turn everything back on.

When the state is cleared, before turning everything back on, I want to check to see if the alarm has triggered n-number of times in the last 30 minutes. 

I could use a Number Item as 1/0 and sum that but that does not smell right.

Could a new method be added to the persitence CountSince which simply counts persistened instances since the time.

thank you.

J-N-K:
Closes #944

### DIFF
--- a/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
+++ b/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
@@ -16,8 +16,10 @@ import java.math.BigDecimal;
 import java.math.MathContext;
 import java.time.Instant;
 import java.time.ZonedDateTime;
+import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
+import java.util.stream.StreamSupport;
 
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.items.Item;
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
```diff
@@ -1091,6 +1093,63 @@ public static DecimalType evolutionRate(Item item, ZonedDateTime begin, ZonedDat
         return null;
     }
 
+    /**
+     * Gets the number of persisted values of a given {@link Item} from a point in time until now.
```
kaikreuzer, line 1097: I wonder if this is always correct and can be guaranteed.
The rrd4j persistence service afaik returns only changed values and due to the data densifying nature, there is simply no trace of what / how many values have been persisted and this will even change over time for the same period. Also, whenever a time-based persistence strategy is used, the results might often not be what the user expects. Wdyt about that? (I see that the issue is not really introduced by your PR, but rather by the introduction of the `getAllStatesBetween()` method already, but anyhow...)

J-N-K: I agree with rrd4j. But rrd4j will also not provide the correct "historicState", "minimumSince" or "maximumSince", because the values are probably lost. 

What you in fact get `count` is the number of values available (like for other persistence services. Maybe the wording should be improved to something like "Gets the number of available datapoints for a given ...". This is then also correct for rrd4j.

What problems do you expect with time-based persistence services (except for rrd4j, due to the nature of the binning, the result will always be the number of bins).

kaikreuzer: > Maybe the wording should be improved to something like "Gets the number of available datapoints for a given ...". This is then also correct for rrd4j.

Sounds like a good compromise, yes.

> What problems do you expect with time-based persistence services

Not time-based services, but strategies. The results of a persistence service will be very different, if you only do "onChange" or if you additionally do an "everyMinute" strategy. The use case mentioned in #944 would stop working. So a method like `countStateChangesSince()` or something like that might actually be more helpful for the mentioned use case.

J-N-K:
I would say that using `everyChange` and `everyMinute` for the same item and the same persistence service and then expecting to get a proper count for the number of changes by counting the persisted values is probably too much. For the persistence extensions there is no way to know what the used persistence strategy is (if any, you could also just use `.persist` from a rule. 

I would say that the documentation should contain a warning that the returned value may be inaccurate or unexpected if more than one strategy is applied.

jimtng:
Could you add a parameter, similar to previousState() to filter out consecutive data that are the same?

kaikreuzer (approved):
Lgtm, thanks. Although I am not really convinced that anybody will ever really use these methods as the use cases for them seem to be fairly limited.

kaikreuzer:
> Could you add a parameter, similar to previousState() to filter out consecutive data that are the same?

I guess that would be better another method like an above mentioned `countStateChangesSince()` - and that would be a separate PR.

PR #3153, opened 2022-11-08
Strip unit if NumberItem has no dimension

J-N-K:
Related to #3121 

As discussed in the linked issue: The state of a `Number` item (i.e. without dimension) is expected to be `DecimalType` without a unit. If a state update is posted or command received that is of `QuantityType` the unit should be stripped to ensure that.


kaikreuzer (approved):
lgtm, thanks!

PR #3154, opened 2022-11-08
[auth] Fix wrong parameter on AuthorizePageServlet

GiviMAD:
Found this small bug while reading the authorization code.

Edit: couple of lines after is enforcing the client id to match the refresh token which is now always true. 


lolodomo:
Any idea what kind of issues it should have triggered ?

GiviMAD:
I think it has been no impact, as the ui was sending the same value for both params, which I think was the expected but is not correctly validated. And the client id you are sending gets ignored so will be invalid in the next authorisation step, so still if you try to use a different client_id that redirect_url you are not able to get a token. So I don't think it causes any security problems in the end, just a weird behaviour because the first auth step tells you everything is ok but you won't be able to continue.

J-N-K (approved):

Issue #3128, opened 2022-10-23
Access to `org.eclipse.xtext.xbase.lib` possible from addons?


florian-h05:
Dear @openhab/core-maintainers,

is it possible to access the `org.eclipse.xtext.xbase.lib` package that is used in core in addons?
I am asking, because I need access to that package.
Adding it as a dependency of the addon might be solution, but as it is already at some place (core) it would be great if I could just use it from there.

kaikreuzer:
Hi @florian-h05,

No, this should actually not be used by any add-on. It shouldn't even be used by core bundles besides the ones that deal with DSL support.
We are trying to keep those dependencies clean, so that e.g. a future openHAB version might also work without Xtext in place (at least as an option), so if add-ons start introducing such dependencies, this wouldn't be possible.

florian-h05:
Thanks for the quick answer @kaikreuzer.

I might come up with a question related to this later, so please leave the issue open.

kaikreuzer:
Sure, feel free to come back any time!

jpg0:
I believe that the problem @florian-h05 is encountering is the need to create a Timer from Java code within an add-on (to simulate a native JS language feature). Currently `ScriptExecution` accepts a `Procedure0`, where I would expect it to accept a `Runnable`. I don't know how Xtext works though, so I guess either:

- This isn't really necessary, so should change to `Runnable`, or
- There is, or should be, another, preferred way to create a Timer from Java (add-on) code

J-N-K:
I would think that the classes and methods in `o.o.c.model.script` should not be used outside of DSL because it mixes elements of xtext with what should be plain Java code.

florian-h05:
> I don't know how Xtext works though, so I guess either:

First of all, I also have no experience with Xtext.

> This isn't really necessary, so should change to Runnable, or

Please have a look at https://github.com/florian-h05/openhab-addons/blob/jsscripting-fix-multithread-access/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/threading/ThreadsafeTimers.java. This is a replacement for the timer creation mechanism of `org.openhab.core.model.script.actions.ScriptExecution` used in the addon and it is working with `Runnable`. All I can say after some tests, it's working fine. 
I have also checked whether mutation of passed in vars works after the timer creation, it works (the argument passed is not used, but I wanted to keep the method signature).

florian-h05:
Closing as there is no need to access those.

Issue #2898, opened 2022-04-06
[DateTimeType] ZonedDateTime, Instant or neither?


jlaur:
I'm creating this issue in order to discuss how [DateTimeType](https://www.openhab.org/javadoc/latest/org/openhab/core/library/types/datetimetype) should work. It seems like currently it internally stores a [ZonedDateTime](https://docs.oracle.com/javase/8/docs/api/java/time/ZonedDateTime.html) stripped of [ZoneId](https://docs.oracle.com/javase/8/docs/api/java/time/ZoneId.html), but preserving a [ZoneOffset](https://docs.oracle.com/javase/8/docs/api/java/time/ZoneOffset.html):

https://github.com/openhab/openhab-core/blob/cabb3f73150d13f44d647f46af8362e396684b92/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java#L75-L77

This means that information about **Daylight Savings Time** is lost. Expressed as DSL rule:

```
val zdt1 = now
val zdt2 = new DateTimeType().getZonedDateTime()
logInfo("zdt1", "toString: " + zdt1.toString() + ", zone: " + zdt1.getZone() + ", dst: " + zdt1.getZone().getRules().isDaylightSavings(zdt1.toInstant()))
logInfo("zdt2", "toString: " + zdt2.toString() + ", zone: " + zdt2.getZone() + ", dst: " + zdt2.getZone().getRules().isDaylightSavings(zdt2.toInstant()))
```

Result:
```
2022-04-06 22:11:29.194 [INFO ] [org.openhab.core.model.script.zdt1  ] - toString: 2022-04-06T22:11:29.183384+02:00[Europe/Berlin], zone: Europe/Berlin, dst: true
2022-04-06 22:11:29.203 [INFO ] [org.openhab.core.model.script.zdt2  ] - toString: 2022-04-06T22:11:29.183905+02:00, zone: +02:00, dst: false
```
(side note: I have [Europe/Copenhagen] configured in openHAB, so DSL rule engine doesn't seem to use TimeZoneProvider for `now`)

This has been a problem for at least openhab/openhab-addons#12546.

I'm wondering if `ZoneId` being stripped is a feature or a bug? It seems intended as `withFixedOffsetZone()` is explicitly called, I'm just not sure why.

### Instant?
Second, a bigger topic: I'm wondering if it even makes sense to store a `ZonedDateTime` for `DateTimeType`. It's documented here:
[...]

PR #3155, opened 2022-11-09
Refactor ScriptExecution to plain Java ScriptExtension

J-N-K:
JSR-223 automation add-ons often use the script actions from the `org.openhab.core.model.script` bundle. The downside is that the actions are tied to XText, making it impossible for the automation add-ons to modify/wrap these actions without explicit dependency on XText (see #3128).

This is the first step to refactor these actions to plain Java, making them available as `ScriptExtension` via an import-preset `ScriptAction` which is not imported by default to prevent unwanted shadowing of other components. They are wrapped for DSL rules, to stay backward compatible.

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/action/ScriptExecution.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/action/ScriptExecution.java
@@ -1,0 +1,77 @@
+/**
+ * Copyright (c) 2010-2022 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.automation.module.script.action;
+
+import java.time.ZonedDateTime;
+import java.util.function.Consumer;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+
+/**
+ * The {@link ScriptExecution} is a
+ *
+ * @author Jan N. Klug - Initial contribution
+ */
+@NonNullByDefault
+public interface ScriptExecution {
+    /**
+     * Calls a script which must be located in the configurations/scripts folder.
+     *
+     * @param scriptName the name of the script (if the name does not end with
+     *            the .script file extension it is added)
+     * @return the return value of the script
+     */
+    Object callScript(String scriptName);
+
+    /**
+     * Schedules a block of code for later execution.
+     *
+     * @param instant the point in time when the code should be executed
+     * @param closure the code block to execute
+     * @return a handle to the created timer, so that it can be canceled or rescheduled
+     */
+    Timer createTimer(ZonedDateTime instant, Runnable closure);
+
+    /**
+     * Schedules a block of code for later execution.
+     *
+     * @param identifier an optional identifier
+     * @param instant the point in time when the code should be executed
+     * @param closure the code block to execute
+     * @return a handle to the created timer, so that it can be canceled or rescheduled
+     */
+    Timer createTimer(@Nullable String identifier, ZonedDateTime instant, Runnable closure);
+
+    /**
+     * Schedules a block of code (with argument) for later execution
+     *
+     * @param instant the point in time when the code should be executed
+     * @param arg1 the argument to pass to the code block
+     * @param closure the code block to execute
+     * @return a handle to the created timer, so that it can be canceled or rescheduled
+     */
+    Timer createTimerWithArgument(ZonedDateTime instant, Object arg1, Consumer<Object> closure);
+
+    /**
+     * Schedules a block of code (with argument) for later execution
+     *
+     * @param identifier an optional identifier
+     * @param instant the point in time when the code should be executed
+     * @param arg1 the argument to pass to the code block
+     * @param closure the code block to execute
+     * @return a handle to the created timer, so that it can be canceled or rescheduled
+     */
+    Timer createTimerWithArgument(@Nullable String identifier, ZonedDateTime instant, Object arg1,
+            Consumer<Object> closure);
+}
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/action/ScriptExecution.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/action/ScriptExecution.java
@@ -29,41 +29,41 @@ public interface ScriptExecution {
     /**
      * Schedules a block of code for later execution.
      *
-     * @param instant the point in time when the code should be executed
+     * @param zonedDateTime the point in time when the code should be executed
      * @param closure the code block to execute
      * @return a handle to the created timer, so that it can be canceled or rescheduled
      */
-    Timer createTimer(ZonedDateTime instant, Runnable closure);
+    Timer createTimer(ZonedDateTime zonedDateTime, Runnable closure);
 
     /**
      * Schedules a block of code for later execution.
      *
      * @param identifier an optional identifier
-     * @param instant the point in time when the code should be executed
+     * @param zonedDateTime the point in time when the code should be executed
      * @param closure the code block to execute
      * @return a handle to the created timer, so that it can be canceled or rescheduled
      */
-    Timer createTimer(@Nullable String identifier, ZonedDateTime instant, Runnable closure);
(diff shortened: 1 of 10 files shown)


J-N-K:
@florian-h05 Is that what you would expect?

florian-h05:
Wow, that was fast Jan 👍
Looks good on the first view, but I have to checkout to have a closer look.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/action/ScriptExecution.java
```diff
@@ -0,0 +1,77 @@
...
+package org.openhab.core.automation.module.script.action;
+
+import java.time.ZonedDateTime;
+import java.util.function.Consumer;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+
+/**
+ * The {@link ScriptExecution} is a
```
ccutrer, line 22: Is a what?

ccutrer (commented):
This is much nicer. I was just working with createTimer from the jrubyscripting library yesterday, and was wondering why it wasn't just using a Runnable.

florian-h05:
@J-N-K 
Could you point me in the right direction, what is the best way to access the ScriptExtension from the JS Scripting add-on?

I have seen that you export a ScriptExtension, but I am not able to import them in JS scripts. (Whilst this is technically not required, I am wondering why it is not working.)

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/action/ScriptExecution.java
```diff
@@ -0,0 +1,69 @@
...
+public interface ScriptExecution {
+
+    /**
+     * Schedules a block of code for later execution.
+     *
+     * @param instant the point in time when the code should be executed
+     * @param closure the code block to execute
+     * @return a handle to the created timer, so that it can be canceled or rescheduled
+     */
+    Timer createTimer(ZonedDateTime instant, Runnable closure);
```
jlaur, line 36: The naming of the `ZonedDateTime` as _instant_ triggers me to ask the question: Would an `Instant` be sufficient here, since the the point in time doesn't care about time-zone? If `ZonedDateTime` is more convenient, consider if parameter should be renamed for clarity, and/or if a variant taking `Instant` would be relevant.

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/actions/ScriptExecution.java
```diff
@@ -67,67 +73,23 @@ public static Object callScript(String scriptName) throws ScriptExecutionExcepti
...
-     * @throws ScriptExecutionException if an error occurs during the execution
-     */
-    public static Timer createTimer(@Nullable String identifier, ZonedDateTime instant, Procedure0 closure) {
-        Scheduler scheduler = ScriptServiceUtil.getScheduler();
-
-        return new TimerImpl(scheduler, instant, () -> {
-            closure.apply();
-        }, identifier);
+    @ActionDoc(text = "create an identifiable timer ")
+    public static Timer createTimer(@Nullable String identifier,  ZonedDateTime zonedDateTime, Procedures.Procedure0 closure) {
```
jlaur, line 82:
```suggestion
    public static Timer createTimer(@Nullable String identifier, ZonedDateTime zonedDateTime, Procedures.Procedure0 closure) {
```

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/actions/ScriptExecution.java
```diff
@@ -67,67 +73,23 @@ public static Object callScript(String scriptName) throws ScriptExecutionExcepti
...
-     * @throws ScriptExecutionException if an error occurs during the execution
-     */
-    public static Timer createTimerWithArgument(@Nullable String identifier,  ZonedDateTime instant, Object arg1, Procedure1<Object> closure) {
-        Scheduler scheduler = ScriptServiceUtil.getScheduler();
-
-        return new TimerImpl(scheduler, instant, () -> {
-            closure.apply(arg1);
-        }, identifier);
+    @ActionDoc(text = "create an identifiable timer with argument")
+    public static Timer createTimerWithArgument(@Nullable String identifier,  ZonedDateTime zonedDateTime, Object arg1, Procedures.Procedure1 closure) {
```
jlaur, line 92:
```suggestion
    public static Timer createTimerWithArgument(@Nullable String identifier, ZonedDateTime zonedDateTime, Object arg1, Procedures.Procedure1 closure) {
```

florian-h05:
@J-N-K What do you think, when does this PR get merged?

J-N-K:
I'm probably the last person with an answer to this: I can't review and merge my own PR.

florian-h05:
@kaikreuzer @cweitkamp @wborn Just asking, WDYT when can you review and merge this PR? https://github.com/openhab/openhab-addons/pull/13695 depends on this.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/action/ScriptExecution.java
rkoshak: The docs actually say it should be an `AbstractInstant`.

J-N-K: `AbstractInstant` is surely wrong because that's a class from soda-time which we don't use anymore. So the docs need to be updated.

IMO an `Instant` would also be fine (also see my comment on #2898) but I re-implemented what we have in DSL and that is a `ZonedDateTime`. If we try to solve that now, we won't get a solution before next release.

florian-h05: Out of interest: Why can't we get a solution before next release?

rkoshak: I always felt that the docs references to abstract classes like this was confusing to new users anyway. I'm not sure `Instant` is any better but it's at least consistent. When/if this ever changes here to `Instant`, we should also change the persistence actions which are documented as using `ZonedDateTime` to be consistent.

J-N-K: @florian-h05 Technically it might be possible, but it's important what @rkoshak mentions: We use `ZonedDateTime` in many places. So we should have a discussion where to go, having `Instant` for `createTimer` and `ZonedDateTime` for the persistence services is very confusing and should be avoided. 

General discussions like this tend to take some time and with less than a month until feature-freeze and then time maybe a bit short. Just "exchanging" it with `Instant` will probably break a lot of code and I feel it's not a good idea to do that for a minor release.

florian-h05: Thanks for the explanation, I understand the impact of such a decision. 

So what’s the plan now? Get this PR merged before the next release and then discuss the Instand ZonedDateTime thing later?

I’d like to get the thread-safe reimplementation of createTimer in JS Scripting into the next release.

J-N-K: That would be my proposal. IMO user-facing ZDT and Instant don‘t make much difference, so using Instant internally in DateTimeType and ZDT in rules is fine. Does it make a difference in JS?

florian-h05: > Does it make a difference in JS?

Regarding the compatibility, no.
We are using JS-Joda as date and time library, which provides the same functionality that the Joda time library had (ZonedDateTime, Instant and so on).
When core moves from ZonedDateTime to Instant, we only need to use Instant from JS-Joda instead of ZonedDateTime from JS-Joda.

J-N-K: If there is no real benefit (I don't see one for DSL or JavaRule), we should probably stay where we are.

florian-h05: The only difference I can see is, which argument and which string the `parse` function takes.
Given this little difference, but the large number of scripts that have to be updated when that’s changed to an Instant, I agree that we should probably stay on ZonedDateTime.
The question for now is: should we rename the parameter from instant to zdt or zonedDateTime or something like that?

J-N-K: Yes, I‘ll update that.

kaikreuzer (approved):
lgtm, thanks!

PR #3156, opened 2022-11-11
Add unmanaged scripts to rule registry

J-N-K:
This allows access to unmanaged DSL scripts from all rules. This is achieved by adding them as rules with tag `Script` (like managed scripts) to the `RuleRegistry`. Thy can now be called in the same way as managed scripts.

In addition the script context is made accessible as return value for `runNow` methods which will later allow calling scripts with their return value from scripted rules or other scripts.


ghys:
Excellent! this will resolve the confusion between scripts as in files in the `scripts` configuration folder and what the UI call scripts, that is rules with a single script action module. I'm sure this was your intent so you'll be aware some UI adaptations would be in order (with "lock" icons and restrictions for non-managed scripts as is in place for rules).

J-N-K:
The lock icon is already there, it appeared automatically. The only UI issue is that if you click on the script in the overview the code view opens and shows the script as expected but with an empty triggers section at the top:

Bildschirm­foto 2022-11-12 um 17 02 57

Bildschirm­foto 2022-11-12 um 17 03 12

wborn (approved):
Thanks!

Issue #2150, opened 2021-01-23
[OH3.x - feature request]: Run rules with parameters


Boby71:
Currently in OH3 it‘s possible to activate, disable and even execute rules.
As soon as the rule is driven by an event -especially those you can‘t trigger by eg tuning on a switch - the run function isn‘t helpful.

Suggestion:
There should be another „run“ button which opens a small window where it‘s possible to enter values for the following elements:

- receivedCommand - implicitly available in every rule that has at least one command event trigger.
- previousState - implicitly available in every rule that has at least one status change event trigger
- newState - implicitly available in every rule that has at least one status update or status change event trigger.
- triggeringItemName - implicitly available in every rule that has at least one status update, status change or command event trigger
- triggeringItem - implicitly available in every rule that has a "Member of" trigger
- receivedEvent - implicitly available in every rule that has a channel-based 

This would make debugging rules so much easier!

PR #3160, opened 2022-11-13
[REST] allow running rules with context

J-N-K:
This is the core part to #2150. 

An empty request body behaves like before, a body of

```
{
  "1.command" : "ON"
}
```

sets the `receivedCommand` to `ON` for DSL rules.

### DIFF
--- a/bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/RuleResource.java
+++ b/bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/RuleResource.java
@@ -312,19 +312,20 @@ public class RuleResource implements RESTResource {
     @POST
     @RolesAllowed({ Role.USER, Role.ADMIN })
     @Path("/{ruleUID}/runnow")
-    @Consumes(MediaType.TEXT_PLAIN)
+    @Consumes(MediaType.APPLICATION_JSON)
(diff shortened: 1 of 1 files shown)


wborn:
> "1.command" : "ON"

> sets the receivedCommand to ON for DSL rules.

How does one exactly know what kind of data to use as context?

J-N-K:
In fact that is not documented. The reason is that the context was internal only before and so the only existing documentation is the code itself. Essentially you have to place in the context what (in this case) would be placed there for the trigger data (`RuleEngineImpl.setTriggerOutputs`).

I don't think there is a good place in core to place that documentation.

kaikreuzer (approved):
lgtm

kaikreuzer:
> I don't think there is a good place in core to place that documentation.

Yeah, indeed - the context highly depends on the trigger modules and their defined output values. We unfortunately do not have any good overview over those in the docs.

flo-mic:
@J-N-K not sure if my issue comes from this change but I believe so. After updating to Milestone 6 I noticed that I can not more start file based rules within the UI. I always get the response code 415 with unsuported media type.

After some analysis I noticed that the ui is not sending any body with the `POST` request to `/rest/rules/example/runnow`. If the ui would send `{}` the rule would be executed as the mediatype would be `application/json`.

J-N-K:
Yes and no. The reason is that UI posts Content-Type `text/plain` instead of `application/json`. For improved backward compatibility I'll add a new endpoint for that.

PR #3162, opened 2022-11-15
Fix bug in migrateThingType to not call managed thing provider directly

Hilbrand:
The `migrateThingType` call the managed thing provider directly instead of checking if it's actually a managed thing. This check is done in the `thingUpdated` method. Changed the code so it makes the same call as in `thingUpdated`.

This bug was trigged by a change in the systeminfo binding: https://github.com/openhab/openhab-addons/issues/13708.

I've refactored the `thingUpdated` into a private method that is also called from `migrateThingType` as `thingUpdated` performs the same `update` method as was done in `migrateThingType` when the is managed. I've kept the lock and `doPrivileged` in the private method. I'm not sure if that is needed, but it seems to make sense as it's a guard in the `updateThing`, so why not in `migrateThingType`.

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
@@ -262,32 +262,7 @@ public class ThingManagerImpl
 
         @Override
         public void thingUpdated(final Thing thing) {
-            thingUpdatedLock.add(thing.getUID());
-            AccessController.doPrivileged((PrivilegedAction<@Nullable Void>) () -> {
-                Provider<Thing> provider = thingRegistry.getProvider(thing);
-                if (provider == null) {
-                    throw new IllegalArgumentException(MessageFormat.format(
-                            "Provider for thing {0} cannot be determined because it is not known to the registry",
-                            thing.getUID().getAsString()));
-                }
-                if (provider instanceof ManagedProvider) {
-                    @SuppressWarnings("unchecked")
-                    ManagedProvider<Thing, ThingUID> managedProvider = (ManagedProvider<Thing, ThingUID>) provider;
-                    managedProvider.update(thing);
-                } else {
-                    logger.debug("Only updating thing {} in the registry because provider {} is not managed.",
-                            thing.getUID().getAsString(), provider);
-                    Thing oldThing = thingRegistry.get(thing.getUID());
-                    if (oldThing == null) {
-                        throw new IllegalArgumentException(
-                                MessageFormat.format("Cannot update thing {0} because it is not known to the registry",
-                                        thing.getUID().getAsString()));
-                    }
-                    thingRegistry.updated(provider, oldThing, thing);
-                }
-                return null;
-            });
-            thingUpdatedLock.remove(thing.getUID());
+            ThingManagerImpl.this.thingUpdated(thing);
         }
 
         @Override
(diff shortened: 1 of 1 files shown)


lolodomo:
Thank you a lot @Hilbrand . 
We need that PR merged before the next 3.4 milestone is released.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
```diff
@@ -472,6 +447,35 @@ protected synchronized void deactivate(ComponentContext componentContext) {
         readyService.unregisterTracker(this);
     }
 
+    private void thingUpdated(final Thing thing) {
+        thingUpdatedLock.add(thing.getUID());
+        AccessController.doPrivileged((PrivilegedAction<@Nullable Void>) () -> {
```
J-N-K, line 452: I think that the `AccessController` can be removed, it is deprecated for removal in Java 17 anyway and I know of no code that would depend on it in openHAB. But we should probably clean-up all code that uses the `SecurityManager` after we switched to Java 17 at the end of this year.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
```diff
@@ -523,7 +527,7 @@ public void run() {
                     ((ThingImpl) thing).setThingTypeUID(thingTypeUID);
 
                     // Register the new Handler - ThingManager.updateThing() is going to take care of that
```
J-N-K, line 529: Please adjust code comment

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
```diff
@@ -262,32 +262,7 @@ private void handleBridgeChildStatusUpdate(Thing thing, ThingStatusInfo oldStatu
...
-                            thing.getUID().getAsString(), provider);
-                    Thing oldThing = thingRegistry.get(thing.getUID());
-                    if (oldThing == null) {
-                        throw new IllegalArgumentException(
-                                MessageFormat.format("Cannot update thing {0} because it is not known to the registry",
-                                        thing.getUID().getAsString()));
-                    }
-                    thingRegistry.updated(provider, oldThing, thing);
-                }
-                return null;
```
J-N-K, line 288: In fact this contradicts the description in the `ThingHandlerCallback` interface:

```
    /**
     * Informs about an update of the whole thing.
     *
     * @param thing thing that was updated (must not be null)
     * @throws IllegalStateException if the {@link Thing} is read-only.
     */
    void thingUpdated(Thing thing);
```

However, if we strictly follow the interface description this makes any dynamic properties or updates to file-based things impossible. IMO the interface desciption should be updated to reflect the current code.

I also wonder why we check if the thing exists. I would be very astonished if we got a provider from the `ThingRegistry` when the thing itself is not in the registry. I think this was introduced in #1134 to prevent the null-error for `Registry.updated`. It would probably be better to check if the thing is present in the registry before we try to get the provider and fail early if it is not in the registry (the registry fails silently if you try to update a thing that is not in the registry).

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
```diff
@@ -472,6 +447,35 @@ protected synchronized void deactivate(ComponentContext componentContext) {
...
+                            MessageFormat.format("Cannot update thing {0} because it is not known to the registry",
+                                    thing.getUID().getAsString()));
+                }
+                thingRegistry.updated(provider, oldThing, thing);
+            }
+            return null;
+        });
+        thingUpdatedLock.remove(thing.getUID());
+    }
+
```
J-N-K, line 478: While reviewing I found that the `Message.format` seven lines below contain an NBSP after the first placeholder. Can you fix that, if you are near this code anyway? Thanks.

J-N-K (changes requested):
Nice catch. I have left some comments. Also please check why the integration tests fail.

J-N-K:
It's also used in 3rd party add-ons.

Hilbrand:
@J-N-K the itests do seem to fail due to this change. But debugging these tests is a really horrible, so it will take me some time. Do you know how to enable logging in the test so it shows. The only log I get is from the iteat themselves, but not from the core code.

J-N-K:
I usually set the log messages I'm interested in to level ERROR, then they show up.

wborn:
You can easily change the log level for all loggers here:

https://github.com/openhab/openhab-core/blob/d00d14d9fcab6e614da67264bce18e91c2be1be8/itests/itest-include.bndrun#L40

lolodomo:
No chance to finish before the new milestone is generated later today?

J-N-K:
The tests failed because the thing handler was unregistered before migration but the registration of the new handler was missing.

@openhab/core-maintainers Can someone else review now that I contributed to this PR? Thanks.

andrewfg:
^
Polite reminder: can we try to get this into 3.4 RC :)

wborn (approved):
LGTM and thank you both!

PR #3164, opened 2022-11-17
[persistence] Clarify intention of FilterCriteria beginDate/endDate

jlaur:
See https://github.com/openhab/openhab-addons/issues/9906#issuecomment-1317604058

Most persistence implementations filters by `beginDate` <=date <= `endDate` which contradicts the Javadoc comments with proposed changes in this PR. The exception is JDBC persistence which lead to openhab/openhab-addons#9906.

Therefore, I would propose to either:
- Close this PR and fix persistence services RRD4j, JPA and MongoDB.
- Consider/merge this PR and fix persistence service JDBC according to mentioned issue. openhab/openhab-addons#13734 prepared.

### DIFF
--- a/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/FilterCriteria.java
+++ b/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/FilterCriteria.java
@@ -67,10 +67,10 @@ public class FilterCriteria {
     /** filter result to only contain entries for the given item */
     private String itemName;
 
-    /** filter result to only contain entries that are newer than the given date */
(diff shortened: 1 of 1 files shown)


cweitkamp (approved):
I am fine with the change of this definition. Let's ask one more @openhab/core-maintainers about it.

J-N-K (approved):
I agree. Let's wait some days if someone else differs and merge on Tuesday if no one stands up against.

kaikreuzer:
I'm also fine with it and would even assume that this was the original intention.

kaikreuzer (approved):
I'm also fine with it and would even assume that this was the original intention.

wborn (approved):

PR #3177, opened 2022-11-26
add QuantityType.toUnitRelative

ccutrer:
this is a useful helper that can eliminate some complicated code elsewhere. starting in SystemOffsetProfile. then I also want to use it in the homekit addon.

extracted from https://github.com/openhab/openhab-core/pull/3132, since the StateDescription portion of that PR is still being discussed what direction to go


J-N-K (approved):
LGTM

PR #3172, opened 2022-11-22
Fix createTimer for DSL rules

J-N-K:
Reported on the forum. Critical because it breaks timers in DSL rules.

Signed-off-by: Jan N. Klug <github@klug.nrw>

PR #3178, opened 2022-11-27
Fix transformRaw validation errors in DSL rules

### LINKED ISSUE #1644: [transformations] transformRaw() action gives validation errors.
openHAB 2.5.0
Use of 
`var blah  = transformRaw("xx","yy","zz")`
in DSL rules throws a validation error at load time in `openhab.log`

> 2020-09-13 22:47:41.089 [INFO ] [el.core.internal.ModelRepositoryImpl] - Validation issues found in configuration model 'doors.rules', using it anyway:
> The method transformRaw(String, String, String) from the type Transformation refers to the missing type Object

The function does seem to work as expected at runtime.

VSCode syntax highlighter doesn't like it either, but I expect that comes from the same root cause.

J-N-K:
Fixes #1644 

While working on #3172 I found that the same solution exists for #1644.


wborn (approved):
Thanks for the fix!

Issue #3179, opened 2022-11-27
Single JSR223 script can permanently lock up all other scripts from loading


ccutrer:
Say you have a super simple script (simplified for illustration purposes; a real world example could involve an accidental deadlock, or waiting indefinitely for an external service the script calls), like this:

```ruby
loop { sleep 1 }
```

(Yes, I shot myself in the foot doing something kind of like this trying to debug something! Luckily my condition eventually happened, so I didn't have to restart, but I realized my mistake as soon as I saved the file.)

ScriptFileWatcher loads scripts serially, so that will permanently lock up any other scripts from loading, and the only way to recover is to restart openHAB.

cweitkamp:
Let's reopen this to track the issue again.

PR #3180, opened 2022-11-27
wrap ScriptEngine executions in a SafeCaller

ccutrer:
See #3179 for more details.

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/src/test/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImplTest.java
+++ b/bundles/org.openhab.core.automation.module.script/src/test/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImplTest.java
@@ -40,8 +40,7 @@ import org.mockito.junit.jupiter.MockitoSettings;
 import org.mockito.quality.Strictness;
 import org.openhab.core.automation.module.script.ScriptDependencyTracker;
 import org.openhab.core.automation.module.script.ScriptEngineFactory;
-import org.openhab.core.common.SafeCaller;
-import org.openhab.core.common.SafeCallerBuilder;
+import org.openhab.core.internal.common.SafeCallerImpl;
 
 /**
  * The {@link ScriptEngineManagerImplTest} is a test class for the {@link ScriptEngineManagerImpl}
@@ -65,44 +64,6 @@ public class ScriptEngineManagerImplTest {
 
     private @NonNullByDefault({}) ScriptEngineManagerImpl scriptEngineManager;
 
-    private static class SafeCallerMock implements SafeCaller {
-        public <T> SafeCallerBuilder<T> create(T target, Class<T> interfaceType) {
-            return new SafeCallerBuilderMock(target);
-        }
-
-        private static class SafeCallerBuilderMock<T> implements SafeCallerBuilder<T> {
-            final T target;
-
-            SafeCallerBuilderMock(T target) {
-                this.target = target;
-            }
-
-            public T build() {
-                return target;
-            }
-
-            public SafeCallerBuilder<T> withTimeout(long timeout) {
-                return this;
-            }
-
-            public SafeCallerBuilder<T> withIdentifier(Object identifier) {
-                return this;
-            }
-
-            public SafeCallerBuilder<T> onException(Consumer<Throwable> exceptionHandler) {
-                return this;
-            }
-
-            public SafeCallerBuilder<T> onTimeout(Runnable timeoutHandler) {
-                return this;
-            }
-
-            public SafeCallerBuilder<T> withAsync() {
-                return this;
-            }
-        }
-    }
-
     @BeforeEach
     public void setup() {
         when(scriptEngineMock.getFactory()).thenReturn(internalScriptEngineFactoryMock);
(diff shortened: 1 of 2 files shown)


ccutrer:
Note that if the ScriptEngine does not implement AutoCloseable, and the script itself doesn't define scriptUnloaded, nothing will prevent the script from continuing to execute in the background. As of writing, JRuby at least does not implement AutoCloseable. But at least openHAB won't be "hung".

wborn (commented):
It LGTM! But there is a conflict which needs to be resolved.

ccutrer:
Rebase and trivial conflict resolved (dependencyTracker assignment no longer happens in the ScriptEngineManagerImpl). I did have to update the (previously non-existent) tests for ScriptEngineManagerImpl to provide a mock SafeCaller that's actually functional. The rest of the original PR is unchanged.

bundles/org.openhab.core.automation.module.script/src/test/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImplTest.java
```diff
@@ -73,7 +113,7 @@ public void setup() {
         when(scriptEngineFactoryMock.getDependencyTracker()).thenReturn(scriptDependencyTrackerMock);
         when(scriptDependencyTrackerMock.getTracker(any())).thenReturn(dependencyListenerMock);
 
-        scriptEngineManager = new ScriptEngineManagerImpl(scriptExtensionManagerMock);
+        scriptEngineManager = new ScriptEngineManagerImpl(scriptExtensionManagerMock, new SafeCallerMock());
```
wborn, line 116: Why not just use `new SafeCallerImpl(null)` instead of adding all the code for the mock?

wborn: Yes I also saw it was in an internal package after adding my comment. It's a trade off between more code and having to maintain that and staying away from internal classes. :upside_down_face:

wborn (commented):
Why did you add 43018929fee7e78f9cecbd0e0c3b826c4017a538? Seems like it should be part #3185 only.

ccutrer:
> Why did you add [4301892](https://github.com/openhab/openhab-core/commit/43018929fee7e78f9cecbd0e0c3b826c4017a538)? Seems like it should be part #3185 only.

because I'm juggling too many things at once :). I removed that commit from the stack and force-pushed

wborn (approved):
Thanks! My commits sometimes also end up on the wrong branches. :smile:

ccutrer:
> This pull request has been mentioned on **openHAB Community**. There might be relevant details there:
> 
> https://community.openhab.org/t/openhab-3-4-milestone-discussion/138093/135

Apparently 30s+ to load a script might be normal? Still need to hear back from the user, but assuming his code isn't locked up somehow, do we need to make this timeout configurable (or disable-able?). Or perhaps each script engine needs to opt-in to this behavior, indicating it can properly interrupt the user script? I don't love a configurable timeout here, since it seems like it would be very opaque for the user to know to even set it. Unless we explicitly link to the documentation from the log message?

florian-h05:
> Apparently 30s+ to load a script might be normal?

On less powerful platforms like Raspberry Pi and during openHAB startup: yes, it‘s common to have such long load times.

> Or perhaps each script engine needs to opt-in to this behavior, indicating it can properly interrupt the user script?

Would be an option. I am trying to fix the multithread issue in the JS Scripting add-on, but in the PR description, you stated:

> > Note that if the ScriptEngine does not implement AutoCloseable, and the script itself doesn't define scriptUnloaded, nothing will prevent the script from continuing to execute in the background.

Sounds like the problem that a script continues to execute in the background is only for ScriptEngines without AutoClosable. 
What do you think about only using the safe caller for script engines that do not implement AutoClosable?

> I don't love a configurable timeout here, since it seems like it would be very opaque for the user to know to even set it.

Leaving configuration of such things is no good idea IMO.
For those engines not implementing AutoClosable, I‘d rather increase the timeout to 60 or 90 or even 120 seconds.

ccutrer:
> Sounds like the problem that a script continues to execute in the background is only for ScriptEngines without AutoClosable.
> What do you think about only using the safe caller for script engines that do not implement AutoClosable?

I think that's orthogonal. With SafeCaller, on an AutoClosable ScriptEngine (JS), the intention is that a deadlocked script is interrupted and cleaned up. With SafeCaller, on a non-AutoClosable ScriptEngine without a scriptUnloaded hook defined (JRuby, without my helper libraries), openHAB will log a warning that the script took too long, but essentially take no other action to interrupt the script. I can't remember off the top of my head how rules associated with the script get unloaded, but I _think_ they'll be unloaded in this scenario. With SafeCaller, on a non-AutoClosable ScriptEngine _with_ a scriptUnloaded hook defined that knows how to interrupt the script from itself (JRuby, _with_ my helper libraries), the script will end immediately, logging some sort of interrupted exception.

If we were to _not_ use SafeCaller for any permutation of the above, then the original bug is re-introduced, regardless of language - a script that accidentally deadlocks (because of mutexes for thread safety issues; waiting indefinitely on an external resource; an accidental infinite loop; etc.) will monopolize the ScriptFileWatcher thread indefinitely, and you will not be able to unload or reload any other scripts until openHAB is restarted.

florian-h05:
Got it, thanks for the explanation. Disabling SafeCaller is therefore no option.

I think that I found a solution for the multi-thread access issues of JS Scripting by extending synchronization, but independent from this we should increase the timeout.

ccutrer:
> but independent from this we should increase the timeout.

Agreed. I just don't know what to. @J-N-K - any suggestions? My gut says it should be dynamic somehow, like based on number of CPUs or RAM or something. Or even better "don't timeout until we've reached start level 100" or something. But those both seem complicated. I'm fine with just going to 120 to mostly-ameliorate this for 3.4.0.

PR #3181, opened 2022-11-28
Typos a/an

dilyanpalauzov:


bundles/org.openhab.core.semantics/src/main/java/org/openhab/core/semantics/model/equipment/AlarmSystem.java
```diff
@@ -17,7 +17,7 @@
 import org.openhab.core.semantics.TagInfo;
 
 /**
- * This class defines a Alarm System.
```
wborn, line 20: Maybe it is possible to apply the rules in the generator as well? 

https://github.com/openhab/openhab-core/blob/af3738487c5e214165191fe316cfbee02c9f1578/bundles/org.openhab.core.semantics/model/generateTagClasses.groovy#L82

Otherwise it complicates generation and the changes may be overwritten the next time the classes are generated.

You can test generation by activating the profile by adding `-DgenerateTagClasses` to a build.

dilyanpalauzov: I am not going to do this, brcause I jave not setup a build environment.

wborn: It's very easy to do without a  build environement, see: https://github.com/dilyanpalauzov/openhab-core/pull/1 :slightly_smiling_face:

dilyanpalauzov: The change proposed by you is only relatively good, as before U and H both A and AN can be used.

I suggest my PR and your PR are handled separatedly.

wborn (changes requested):
Please apply https://github.com/dilyanpalauzov/openhab-core/pull/1 or undo the changes on the generated code.

wborn (approved):
Looks very good to me now. :+1: I'll keep it open bit in case there is **a** native speaker who still sees **an** issue. :wink:

bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityMarketplaceAddonService.java
```diff
@@ -323,7 +323,7 @@ private Addon convertTopicItemToAddon(DiscourseTopicItem topic, List<DiscourseUs
                 "posts_count", postsCount, //
                 "tags", tags.toArray(String[]::new));
 
-        // try to use an handler to determine if the add-on is installed
+        // try to use a handler to determine if the add-on is installed
```
rkoshak, line 326: The rules for a/an in words that start with "h" are complicated and inconsistent. See https://editingandwritingservices.com/a-or-an-before-words-beginning-with-h/ for one discussion.

In this case, which to use probably depends on which English accent you are thinking of. The most American English accents would pronounce the "h" so "a" would be correct. But some accents would not sound the "h" (in speech it would come across as "andler") in which case "an" would be correct.

I would use "a" here, but I speak with an American accent (western US accent to be specific). I don't know if a UK English or many of the other native speakers would choose differently.

bundles/org.openhab.core.semantics/model/generateTagClasses.groovy
```diff
@@ -69,6 +69,8 @@ def createTagSetClass(def line, String tagSet) {
     def parent = line.Parent
     def parentClass = parent ? parent : type
     def pkg = type.toLowerCase()
+    def ch = label.toLowerCase().charAt(0)
+    def article = ch == 'a' || ch == 'e' || ch == 'i' || ch == 'o' || ch == 'u' ? "an" : "a"
```
rkoshak, line 73: Given that the correct choice depends on how the word is pronounced, I'm not certain this rule, if I'm reading it correctly, will work all the time. For example, it's "an herb" but "a history". There are already lots of examples in the changes above where most words starting with "u" use "a" and not "an". 

But for these purposes it might be close enough. I'm not sure the purpose of these in relation to Groovy and such.

rkoshak (commented):
One maybe error depending on the English accent being targeted and a concern about the rule defined for Groovy.

bundles/org.openhab.core.semantics/model/generateTagClasses.groovy
hmerk: If I remember English (UK) grammer correct, there is a clear rule to use „an“ only followed by vowels in writing.

PR #3182, opened 2022-11-28
return the precise scheduled execution time for timers

ccutrer:
When writing tests and mocking time, it can be important that execution_time match _exactly_ with what was sent to reschedule(). When re-calculating based on now and the expected remaining delay, this might be off by a few milliseconds. We have the exact time that was requested already, so might as well use it.


J-N-K (approved):
LGTM

PR #3188, opened 2022-12-03
Fix CME when creating SCRIPT transformations

J-N-K:
When updating file-based script transformations used in item state-descripting a CME can occur during reload. This is fixed by using thread-safe implementation of `HashMap`.


wborn (commented):
I wonder if `compiledScripts` and `scriptCache` will also be suffering from such concurrency issues? :thinking: They are also being modified in the `clearCache` method.

wborn (approved):
Thanks!

florian-h05:
> It also prevents parallel execution of the script which may be an issue for some scripting engines.

It is in fact an issue for the JS Scripting engine, thanks for solving that. I‘ll wait for the SNAPSHOT and verify if the fix works.

Issue #1924, opened 2020-12-13
Update existing Things for definition changes


wborn:
<a alt="bounty" href="https://www.bountysource.com/issues/95151739"><img src="https://api.bountysource.com/badge/issue?issue_id=95151739"></a>

Currently you have to manually recreate Things for changes in their ThingType or Channel(Group)Definitions to be applied.

The framework should for instance automatically add new channels or configuration parameters.

See also this ESH issue: https://github.com/eclipse-archived/smarthome/issues/2555

These PRs have already been created for solving the issue:

* https://github.com/openhab/openhab-core/pull/1027
* https://github.com/openhab/openhab-core/pull/628
* https://github.com/openhab/openhab-core/pull/527
* https://github.com/eclipse-archived/smarthome/pull/6537

wborn:
As discussed I've added your awarded bounty for #888 on this issue @J-N-K. :+1:

J-N-K:
The biggest problem is that we don't know which channels were originally added from the XML and which channels where added by the user (extensible). This could probably be solved in the REST call when the channel is added. This should be step 1 (and probably should be done before OH3, so that at least all the people starting from scratch will benefit later if we solved the other problems).

We also need to find a solution for already existing things (e.g. MQTT only has extensible channels), a lot of channels would be lost if the flag cannot be added for existing things. This needs to be done during the upgrade process.

The third question is how to handle dynamically added channels. Removing and re-creating them during each startup is possible (and links should be preserved), but this is quite time-consuming and will probably impose bigger delays for larger installations until all things are configured.

Maybe a solution would be to have a field "addedBy" in the channel, which is set to either "framework" (for channels added from the thing defintion) ,"dynamic" (for channels added by the thing handler) and "user" (for channels added by the user).

If these issues are solved, it should be possible to go the route I originally proposed:

- read the stored thing
- create the same thing from scratch
- check which channels (that are not dynamically added and not extensible) are missing in the newly created thing, remove them and add those that are missing
- add (modified) thing to the thingmanager

cdjackson:
Please can you ensure that this can be disabled by bindings (I know we discussed this in the past so I think that was agreed already, but as pointed out, a lot of these discussions were a long time ago now ;) ).

Personally I use different approach - I add a version ID to the thing type, and then use the migrate method to update the thing definition. This ensures there is some control over the process. 

[...]

PR #3191, opened 2022-12-03
Fix ThingManager is missing config description during normalization

J-N-K:
Reported on the forum. In some cases the `ThingManageImpl` tried to initialize the thing before all config descriptions are loaded.

This can happen because the `OH-INF/config` directory is processed in parallel to the thing descriptions. In case the handler factory is added and the thing description parser sets the corresponding `ReadyMarker` before the config descriptions are completly parsed this results in a warning because the config description cannot be found during normalization of the thing or channel configuration.

### DIFF
--- a/bundles/org.openhab.core.thing/pom.xml
+++ b/bundles/org.openhab.core.thing/pom.xml
@@ -22,6 +22,12 @@
     </dependency>
     <dependency>
       <groupId>org.openhab.core.bundles</groupId>
+      <artifactId>org.openhab.core.config.xml</artifactId>
+      <version>${project.version}</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.openhab.core.bundles</groupId>
       <artifactId>org.openhab.core.io.console</artifactId>
       <version>${project.version}</version>
     </dependency>
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
@@ -48,6 +48,7 @@ import org.openhab.core.config.core.ConfigUtil;
 import org.openhab.core.config.core.Configuration;
 import org.openhab.core.config.core.validation.ConfigDescriptionValidator;
 import org.openhab.core.config.core.validation.ConfigValidationException;
+import org.openhab.core.config.xml.ConfigXmlConfigDescriptionProvider;
 import org.openhab.core.events.EventPublisher;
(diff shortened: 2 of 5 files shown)


splatch:
I am not sure approach in this PR is valid because relying on registering config descriptors in thing manager builds functionality which should be provided by config layer itself. Maybe a new `ConfigDescriptionManager` or `ConfigDescriptorRegistry` could help here?

Just an idea:
`OH-INF/config -> ConfigDescriptorProvider -> ConfigDescriptorRegistry -> ConfigDescriptorManager -> ConfigDescriptorListener -> ThingManagerImpl`

With this approach each step has its own responsibility and ThingManager can be free of internals of config descriptor discovery.

J-N-K:
The problem with that approach is that the `ThingManagerImpl` can't know which config descriptions are needed (while that may be possible to check for the thing itself, the config descriptions for all channels of that thing can't be easily checked in advance (e.g. in the reported case on the forum the missing description is `http:channel-config`). Iterating over all channels, getting the config description URI and checking if they are available in the config description registry seems to be quite expensive compared to the solution we have now.

All config descriptions are available, when the bundle has been processed, so this is a sort of short-cut. The same approach is used for the thing descriptions themself, so I merely extended what is already there.

splatch:
Thanks for explaining detail of issue. Still I don't think that tracking directly appearing of config descriptors within thing manager itself is fully justified. A basic fact that thing manager tracks thing types which might be provided by `OH-INF` entries, does not mean that it should keep eye on configs too.
Imperative approach taken now makes logic less clear. Making kind of a dependency set for things would not only simplify code, cause you then could just rely on ie. `thingDependencies.satisfied()` call, it would also let cover various kinds of requirements to activate thing under one interface.
Keep in mind that config descriptors both for thing types and channel types can be provisioned through `ConfigDescriptorProvider`. Tracking `OH-INF` entries might be incomplete to cover all bindings.

J-N-K:
While it is incomplete it is at least better than what we have now.

I did check if an alternative approach is possible and that requires a BIG refactoring. It is quite easy to check if all dependencies are met (we can just check for the thing and all of it's channels if the type is present in the respective registry and if a config description uri is present and the config description itself is present in the registry). 

What makes it difficult is that we need a trigger to check if a thing can be initialized. We currently use two of these: a ready marker when a bundle has finished loading thing types and the addition of a thing handler factory. This PR adds a third one - a ready marker when a bundle has finished loading config descriptions.

If we don't use the ready markers anymore, we would need to be notified of a change in `ConfigDescriptionRegistry`, `ThingTypeRegistry` and `ChannelTypeRegistry`. Unfortunately none of these implements change listeners.

J-N-K:
Closing, will be superseded by another solution that also tries to solve #1924.

PR #3194, opened 2022-12-03
Prevent NPE in SseResource

### LINKED ISSUE #3192: NPE in SseResource during startup
I just found this exception in my logs with the latest snapshot:
```
19:02:21.654 [WARN ] [ommon.WrappedScheduledExecutorService] - Scheduled runnable ended with an exception: 
        at org.openhab.core.io.rest.sse.SseResource.lambda$0(SseResource.java:127) ~[?:?]
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515) ~[?:?]
        at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:305) ~[?:?]
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305) ~[?:?]
```

kaikreuzer:
Simple attempt to fix #3192.
I am not really certain how the `sse` field could end up being `null`, but possibly the job is only executed after the `SseResource` has been destroyed already due to a service/bundle restart.

### DIFF
--- a/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/SseResource.java
+++ b/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/SseResource.java
@@ -123,11 +123,13 @@ public class SseResource implements RESTResource, SsePublisher {
         this.itemStatesEventBuilder = itemStatesEventBuilder;
 
         aliveEventJob = scheduler.scheduleWithFixedDelay(() -> {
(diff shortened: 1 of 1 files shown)


J-N-K (approved):
LGTM. Probably on shutdown the SSE resource is destroyed before the component is de-activated. Since this is a very recent addition and we did not run into issues before I guess it's fine to leave the annotation as-is instead of adding null-checks everywhere.

PR #3142, opened 2022-11-05
[voice] New voice manager dialog methods

GiviMAD:
New dialog methods for voice manager.

Related to https://github.com/openhab/openhab-core/issues/3140

Will close this one https://github.com/openhab/openhab-core/issues/3104

Signed-off-by: Miguel Álvarez <miguelwork92@gmail.com>

Issue #3140, opened 2022-11-03
[voice] Simplify dialog methods


GiviMAD:
The dialog processor constructor and the dialog related apis of the voice manager take a lot of parameters.  

I would like to introduce a new class DialogProcessorContext which encapsulates all the related dialog configuration (sink, source, stt, tts, hlis, listeningItem, ks, keyword, …) so the  addition of new parameters in the future is easier and doesn’t break things.

  Apart from that I would like to rewrite the actions and the console commands for the voice manager so they get easy to use.  
  
  This is an example of the proposed change in voice actions api:
```
// this will be filled with the default voice/audio configuration
 var dialogContext = voice.getDefaultDialogContext(); 
dialogContext.sink = voice.getSink(“customSinkId”);
 dialogContext.listeningItem = “customValue”; 
 voice.startDialog(dialogContext);
// stop
 voice.stopDialog(dialogContext); 
```

 This is an example of the proposed change in the command api (which is change the arguments by parameters): 
```
voice startdialog —listeningItem “customValue” —sink “customSinkId”  —source “customSourceId”
# stop
voice stopdialog —source “customSourceId”
```

 I think this change can be made without breaking the current APIs just marking them as deprecated. 
 WDYT @openhab/core-maintainers? I can open a PR if you think is ok.

J-N-K:
In general I would prefer that, yes. With Java 17 this context could be a `record`, WDYT?

GiviMAD:
Yes, I was not aware of records, it sound like a great idea that will simplify things.

GiviMAD:
I have created a follow up pr to fix and unwanted behavior introduced where you can not start a dialog if some default service is not available, event if you are configuring a different one.
To fix this I have introduced a builder for the dialog context.
Pinning @wborn, in case you can take a look :)

GiviMAD:
Yes, I think so, thank you

PR #3195, opened 2022-12-03
[voice] Use a builder to configure the dialog context and remove some deprecated method usages

GiviMAD:
This is a follow up PR for [#3142](https://github.com/openhab/openhab-core/pull/3142) and related to issue [#3140](https://github.com/openhab/openhab-core/issues/3140).
I realize testing the code on that pr that it introduces and unwanted behavior. If you have setup a default sink/source that is not available (because of a binding uninstall in my case) it will prevent any call to the startDialog method to work.
This PR aims to fix that by introducing a builder for the DialogContext so services availability is checked in the correct moment.
This PR also contains the removal of some deprecated voice manager method usages.

### DIFF
--- a/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/DialogContext.java
+++ b/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/DialogContext.java
@@ -22,8 +22,6 @@ import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.audio.AudioSink;
 import org.openhab.core.audio.AudioSource;
 import org.openhab.core.voice.text.HumanLanguageInterpreter;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * Describes dialog configured services and options.
@@ -103,7 +101,6 @@ public class DialogContext {
      * Allows to describe a dialog context without requiring the involved services to be loaded
      */
     public static class Builder {
-        private final Logger logger = LoggerFactory.getLogger(Builder.class);
         // services
         private @Nullable AudioSource source;
         private @Nullable AudioSink sink;
@@ -176,6 +173,12 @@ public class DialogContext {
             return this;
         }
 
+        /**
+         * Creates a new {@link DialogContext}
+         * 
+         * @return a {@link DialogContext} with the configured components and options
+         * @throws IllegalStateException if a required dialog component is missing
+         */
         public DialogContext build() throws IllegalStateException {
             KSService ksService = ks;
             STTService sttService = stt;
@@ -185,23 +188,22 @@ public class DialogContext {
             AudioSink audioSink = sink;
             List<String> errors = new ArrayList<>();
             if (sttService == null) {
-                errors.add("Missing stt service");
+                errors.add("missing stt service");
             }
             if (ttsService == null) {
-                errors.add("Missing tts service");
+                errors.add("missing tts service");
             }
             if (hliServices.isEmpty()) {
-                errors.add("Missing interpreters");
+                errors.add("missing interpreters");
             }
             if (audioSource == null) {
-                errors.add("Missing audio source");
+                errors.add("missing audio source");
             }
             if (audioSink == null) {
-                errors.add("Missing audio sink");
+                errors.add("missing audio sink");
             }
             if (!errors.isEmpty()) {
-                errors.forEach(logger::warn);
-                throw new IllegalStateException("Cannot build dialog context, services are missing");
+                throw new IllegalStateException("Cannot build dialog context: " + String.join(", ", errors) + ".");
             }
             return new DialogContext(ksService, keyword, sttService, ttsService, voice, hliServices, audioSource,
                     audioSink, locale, listeningItem);
--- a/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/DialogContext.java
+++ b/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/DialogContext.java
@@ -12,14 +12,19 @@
  */
 package org.openhab.core.voice;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.audio.AudioSink;
 import org.openhab.core.audio.AudioSource;
 import org.openhab.core.voice.text.HumanLanguageInterpreter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Describes dialog configured services and options.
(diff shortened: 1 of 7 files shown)


bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -525,43 +570,44 @@ public void startDialog(@Nullable KSService ks, @Nullable STTService stt, @Nulla
...
         }
-        startDialog(context);
+        startDialog(builder);
     }
 
     @Override
-    public void startDialog(DialogContext context) throws IllegalStateException {
+    public void startDialog(DialogContext.Builder contextBuilder) throws IllegalStateException {
         // use defaults, if null
+        DialogContext context = contextBuilder.build();
```
wborn, line 610: Why do you pass the builder and not a context in `startDialog` and `stopDialog` when the very first thing that is done is creating a `DialogContext` with the builder? :thinking:

GiviMAD: Thinking on it, maybe for the method listenAndAnswer this makes no sense, as that one is invokes an immediate execution, I can change that one and let the builder just for the stopDialog and startDialog.

wborn: > I think the way the voice manager handles the dialogs can be improved in the future, I was thinking on adding an issue after this one to gather some feedback. My idea is that the startdialog method can accept the contextBuilder and instead of failing if the services are not present it can register this dialog in a 'pending' state. 

I don't know what that means but it does not sound like it will make it into 3.4 for which the code freeze is in 2 days. So it is probably better to just pass an actual `DialogContext`. Also we are free to break and cleanup APIs with OH4. :wink:

GiviMAD: Sure, makes sense. Thank you for warning me about the code freeze I was not aware.

wborn: The time line was announced here:

https://community.openhab.org/t/openhab-3-4-milestone-builds/138092/4

So we prefer to only merge bug fixes next week. 🙂

GiviMAD: Appreciated! I read that topic once in a while but I didn't know the timelines were announced there.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/DialogContext.java
```diff
@@ -95,72 +100,153 @@ public Locale locale() {
...
+        private Supplier<@Nullable AudioSink> sink = () -> null;
+        private @Nullable String ksId = null;
+        private Supplier<@Nullable KSService> ks = () -> null;
+        private @Nullable String sttId = null;
+        private Supplier<@Nullable STTService> stt = () -> null;
+        private @Nullable String ttsId = null;
+        private Supplier<@Nullable TTSService> tts = () -> null;
+        private @Nullable Voice voice = null;
+        private List<String> hliIds = List.of();
+        private Supplier<List<HumanLanguageInterpreter>> hlis = List::of;
```
wborn, line 121: I think the builder code could be simplified and will be easier to use if it does not use all these `Supplier`s.

wborn (changes requested):

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/DialogContext.java
GiviMAD: The reason to this is that the current implementation of the dialog processor allows to pass unregistered services to it, which is something I'm relaying on in the HABSpeaker addon (marketplace), with the suppliers this is posible.

As an explanation the addon manages the dialog creation disposal for each speaker (the web client) adding some unregistered services that are bounded to that speaker, the ability of non register those services (in my case a ks and a interpreter) allow me to not display to the user services that they should not use.

I can remove the ids if you want, those are there just to give context about the desired services, which can be different than the resolved ones for ks, tts, stt and hlis. I added the ids there also thinking in a future improvement of the voice manager.

So if it's ok for you I would prefer to keep the suppliers, but let me know if you think is not ok.

GiviMAD: I think I understand it incorrectly. If you meant remove the suppliers and just reference the services, yes I can change that without any impact and seems like a good idea.
I understand you were talking about just keeping the service ids.

wborn: Yes like replacing `Supplier<@Nullable TTSService> tts` with `@Nullable TTSService tts` ... if that would work for you.

GiviMAD: The reason for me to did it like this was having some context about the desired services so the builder contain the information to rebuild the dialog later. 
But maybe I overthought it. 
Because right now you can start a dialog with a configured stt, it can fallback to a different one (because that one is not available at that moment) and the voice manager will have not clue that this dialog preference was using other stt service.
Anyway I will open a issue in the future reflecting all this thoughts sorry for the hesitation.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/DialogContext.java
```diff
@@ -95,72 +99,112 @@ public Locale locale() {
...
+        // services
+        private @Nullable AudioSource source = null;
+        private @Nullable AudioSink sink = null;
+        private @Nullable KSService ks = null;
+        private @Nullable STTService stt = null;
+        private @Nullable TTSService tts = null;
+        private @Nullable Voice voice = null;
+        private List<HumanLanguageInterpreter> hlis = List.of();
+        // options
+        private @Nullable String listeningItem = null;
```
wborn, line 116: I think all these null assignments are rudundant?

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/DialogContext.java
GiviMAD: The code looks much better now, thank you.
Seeing it in perspective I was mixing things, so really appreciated!

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/DialogContext.java
GiviMAD: Not sure about it, I will checked it.

GiviMAD: Yes there are, nice to know, I think I did this in a bunch of places, thank you!

wborn: I once [read](http://www.javapractices.com/topic/TopicAction.do?Id=14) default value assignment causes more byte code to be generated and it also adds more cognitive load because it's more code that has to be read. Also you'll encounter unassigned values all over the place, so you have to know the default values anyhow.

GiviMAD: Oh, really nice article, I was thinking about it and I guess I tend to initialize the values because I program a lot on typescript and it blame me to initialize everything (due to the dynamic types in javascript) but I don't think I forget this now, thank you!!

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/DialogContext.java
```diff
@@ -95,72 +99,112 @@ public Locale locale() {
...
+            }
+            if (audioSource == null) {
+                errors.add("Missing audio source");
+            }
+            if (audioSink == null) {
+                errors.add("Missing audio sink");
+            }
+            if (!errors.isEmpty()) {
+                errors.forEach(logger::warn);
+                throw new IllegalStateException("Cannot build dialog context, services are missing");
```
wborn, line 204: Can you add the missing parameters to the `IllegalStateException` message instead (e.g. using `String.join(..)` ?

When exceptions are thrown it should be up to the method caller what to do with them and determine if logging should occur.

You might also want to add a JavaDoc to the `.build()` method to explain when the exception occurs so devs do not have to inspect to code to see when it is thrown.

wborn (approved):
LGTM now! Thanks for building the builder. :slightly_smiling_face:

PR #3197, opened 2022-12-03
Fix DSL action classes regression

### LINKED ISSUE #3196: historicState not available on NumberItem anymore in DSL rules
Since a recent snapshot update, a couple of my DSL rules stopped working with the following message:
```
19:58:01.006 [ERROR] [.internal.handler.ScriptActionHandler] - Script execution of rule with UID 'rule-1' failed: 'historicState' is not a member of 'org.openhab.core.library.items.NumberItem'; line 6, column 62, length 49 in rule
```
The according code in the rule is 
```
MyItem.historicState(now.minusMinutes(1))
```
I couldn't yet figure out, which recent change might have introduced this regression, I still have to further investigate.

@J-N-K Do you maybe have an idea as you have a good overview of the recent changes?

J-N-K:
Fixes #3196 

Re-add a line that was accidentally removed in #3178.


kaikreuzer (approved):
Excellent, that was quick - many thanks!

PR #3201, opened 2022-12-06
Improve UoM handling

J-N-K:
Related to #3143

The logic for determining the unit is to first check if there is a state description. If the state description is present, try to parse a unit from that. In case a unit is found, use that. If no unit is found, use the default unit. Determine the unit from the state description fails if it is set to `%unit%`.

This change disables default units if the state description sets the unit to `%unit%` which effectively means "take the unit from the value" or use no unit.

@Hilbrand, do you think this will work?

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/items/NumberItem.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/items/NumberItem.java
@@ -80,7 +80,7 @@ public class NumberItem extends GenericItem {
         internalSend(command);
     }
 
-    public void send(QuantityType command) {
+    public void send(QuantityType<?> command) {
         if (dimension == null) {
             DecimalType strippedCommand = new DecimalType(command.toBigDecimal());
             internalSend(strippedCommand);
@@ -123,7 +123,7 @@ public class NumberItem extends GenericItem {
 
         // DecimalType update for a NumberItem with dimension, convert to QuantityType:
         if (state instanceof DecimalType && dimension != null) {
-            Unit<?> unit = getUnit();
+            Unit<?> unit = getUnit(dimension, false);
             if (unit != null) {
                 super.setState(new QuantityType<>(((DecimalType) state).doubleValue(), unit));
                 return;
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core/src/test/java/org/openhab/core/library/items/NumberItemTest.java
```diff
@@ -209,4 +210,61 @@ public void testKelvinToMired() {
...
+        item.setUnitProvider(unitProviderMock);
+
+        assertThat(item.getUnit(), is(SIUnits.CELSIUS));
+
+        item.setState(new QuantityType<>("329 °F"));
+        assertThat(item.getState(), is(new QuantityType<>("165 °C")));
+
+        item.setState(new QuantityType<>("100 °C"));
+        assertThat(item.getState(), is(new QuantityType<>("100 °C")));
+    }
```
Hilbrand, line 269: I've added an additional unit test to test specifically for not converting kW to Joule when placeholder is used.
```suggestion
    }
    
    @Test
    void testNoUnitWhenUnitPlaceholderUsed() {
        final UnitProvider unitProviderMock = mock(UnitProvider.class);
        when(unitProviderMock.getUnit(Energy.class)).thenReturn(Units.JOULE);

        final NumberItem item = new NumberItem("Number:Energy", ITEM_NAME);
        item.setUnitProvider(unitProviderMock);
        assertThat(item.getUnit(), is(Units.JOULE));

        item.setStateDescriptionService(stateDescriptionServiceMock);
        item.setState(new QuantityType<>("329 kWh"));
        assertThat(item.getState(), is(new QuantityType<>("329 kWh")));
        assertNull(item.getUnit());
    }
```

bundles/org.openhab.core/src/test/java/org/openhab/core/library/items/NumberItemTest.java
```diff
@@ -209,4 +210,61 @@ public void testKelvinToMired() {
 
         assertThat(item.getState().format("%.0f mired"), is("370 mired"));
     }
+
+    @Test
+    public void stateDescriptionUnitUsedWhenStateDescriptionPresent() {
```
Hilbrand, line 215: Unit tests methods name convention applied (also no public recommended)
```suggestion
    void testStateDescriptionUnitUsedWhenStateDescriptionPresent() {
```

Hilbrand (commented):
Thanks for the fix. It does work. I've added an additional unit test for the specific Joule/kWh issue. And added some comments.

Hilbrand (approved):
LGTM

PR #3202, opened 2022-12-06
Fix AVG calculation in GroupFunction

J-N-K:
Reported on the forum. Due to selection of the `RoundingMode` instead of a `MathContext` the average calculation is returning unexpected results: 19.0 °C / 19.5 °C results in 19.3 °C (instead of 19.25 °C) which might be acceptable, but 19.5 °C / 19.5 °C results in 20 °C (instead of 19.5 °C) which is obviously not what we want or expect. Setting the `MathContext` fixes this problem. I also corrected the same calculation for non-dimensional groups.


wborn (approved):
Thanks for the fix!

PR #3203, opened 2022-12-07
[Binding Skeleton] Add doc code examples

Confectrician:
This is mainly based on my workings for consistent markdown doc files, i am currently doing in the docs repo and also for the many addons.
Hopefully it will increase the consistency by providing predefined code examples.

I have adapted the templating simply from other headlines without further research and hope that it is correct already.

References:

openhab/openhab-addons#13858
openhab/openhab-docs#999


wborn (approved):
Thanks!

PR #3204, opened 2022-12-08
Fix timer cancellation in automation cache

J-N-K:
The check for scheduling a cancellation task did not include the `Timer` class of the automation bundle.

@florian-h05 Can you check if this works for you?


florian-h05:
@J-N-K Thanks for the quick fix, it works! 
FYI, seems like you have a spotless issue.

wborn (approved):
LGTM! :+1:

PR #3208, opened 2022-12-10
Fix DataAmount dimension units

### LINKED ISSUE #3207: DataAmount unit conversion not working
I am trying to convert between DataAmount units.

I created an item Number:DataAmount not linked to any channel, with state description %0.f MB.
Using the REST API, I then assign a value of 1000 B to the item. The expected outcome in would be 1 MB. What I see is 0 MB.
This is in the event log when trying this:
`
2022-12-10 15:31:00.038 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'TestItem' received command 1000 B
2022-12-10 15:31:00.052 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'TestItem' changed from 7.8125E-12 MB to 1.5625E-11 MB
`

Only when assigning a value in MB, I get the correct output.

This issue was originally detected in combination with the SystemInfo binding, see [#13709](https://github.com/openhab/openhab-addons/issues/13709).

J-N-K:
Fixes #3207 (a regression from #3108). 

I also fixed the wrong case prefix for the "kilo"-Units (`K` instead of `k`) but kept the old ones for backward compatibility.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/Units.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/Units.java
@@ -254,7 +254,7 @@ public final class Units extends CustomUnits {
         // "KiB" is wrong, but we keep it for backward compatibility
         SimpleUnitFormat.getInstance().alias(KIBIBYTE, "KiB");
         SimpleUnitFormat.getInstance().alias(KIBIBYTE, "kio");
-        // "KiO" is wrong, but we keep it for backward compatibility
+        // "Kio" is wrong, but we keep it for backward compatibility
         SimpleUnitFormat.getInstance().alias(KIBIBYTE, "Kio");
         SimpleUnitFormat.getInstance().label(KILOBIT, "kbit");
         SimpleUnitFormat.getInstance().label(KILOBIT_PER_SECOND, "kbit/s");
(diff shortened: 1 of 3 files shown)


lolodomo:
Thank you @J-N-K for the fix, this is a very important fix to merge if possible before new milestone is built.
I checked the addons database, the DataAmount dimension is used by 3 bindings: microtik, systeminfo and tr064.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/Units.java
```diff
@@ -245,8 +247,14 @@ public final class Units extends CustomUnits {
...
-        SimpleUnitFormat.getInstance().label(KILOBYTE, "KB");
-        SimpleUnitFormat.getInstance().label(KIBIBYTE, "KiB");
+        SimpleUnitFormat.getInstance().label(KILOBYTE, "kB");
+        // "KB" is wrong, but we keep it for backward compatibility
+        SimpleUnitFormat.getInstance().alias(KILOBYTE, "KB");
+        SimpleUnitFormat.getInstance().label(KIBIBYTE, "kiB");
+        // "KiB" is wrong, but we keep it for backward compatibility
+        SimpleUnitFormat.getInstance().alias(KIBIBYTE, "KiB");
+        SimpleUnitFormat.getInstance().alias(KIBIBYTE, "kio");
+        // "KiO" is wrong, but we keep it for backward compatibility
```
wborn, line 257:
```suggestion
        // "Kio" is wrong, but we keep it for backward compatibility
```

wborn (approved):
LGTM and thanks! :+1:

PR #3219, opened 2022-12-11
Fix BasicAuthentication for websockets

J-N-K:
Follow-Up to #2891

This fixes a bug in decoding the access token validation using basic authentication.


kaikreuzer:
Just for the stupid me: What difference does this code change make?

J-N-K:
I didn‘t think of that, too. `.decode(text)` returns `byte[]` and calling `.toString()` does not return the content of the array as string but the string representation of the memory address.

kaikreuzer (approved):
lgtm, thanks!

PR #3220, opened 2022-12-11
Fix nullable errors in the IDE

kaikreuzer:


wborn (approved):
Thanks! I also noticed this and was considering a similar fix.

PR #3224, opened 2022-12-12
increase ScriptEngineManager's load timeout to 2 minutes

ccutrer:
up from 30 seconds. low-powered devices such as Raspberry Pis can take a very long time to load scripts during initial boot, when many other things are going on at the same time.

see also discussion on the original PR starting at https://github.com/openhab/openhab-core/pull/3180#issuecomment-1346783426, and in the community starting at https://community.openhab.org/t/openhab-3-4-milestone-discussion/138093/135

cc @florian-h05

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java
@@ -58,7 +58,7 @@ public class ScriptEngineManagerImpl implements ScriptEngineManager {
     /**
      * Timeout for scripts in milliseconds.
      */
-    static private final long TIMEOUT = TimeUnit.SECONDS.toMillis(30);
+    static private final long TIMEOUT = TimeUnit.SECONDS.toMillis(120);
 
     private final ScheduledExecutorService scheduler = ThreadPoolManager
             .getScheduledPool(ThreadPoolManager.THREAD_POOL_NAME_COMMON);


J-N-K:
I'm not convinced that this is really a good solution. It does not prevent the issue, it just makes it more unlikely. And the script is not reloaded automatically if it failed I the first place (even though in the reported case that would be the correct solution.

florian-h05:
We have to find a solution that does not block user script from loading, I agree that if we don't find a good solution very fast, we should revert #3180 and fix the original issue after the release.

J-N-K:
I think there is a bug anyway. Why do we call `callHook` in `loadScript` when loading the script failed? That sounds weird.

The correct way would IMO be to check the result of the script loading evaluation and

- if true, `callHook` return true (instead of void)
- if false, return false (instead of void)

The `AbstractScriptFileWatcher.createAndLoad` should then return this status, too (instead of always returning `true`). `AbstractScriptFileWatcher.importFileWhenReady` should re-`enqueue` the script if loading failed.

ccutrer:
It's actually very important for a couple of reasons that you call the scriptUnloaded hook even if the script failed to load. scriptUnloaded hooks are defined by the script themselves, dynamically. If the script has loaded far enough that a scriptUnloaded hook exists at all, it likely means that it has allocated resources that it needs to clean up, regardless of if the rest of the script successfully loads. It's more like a `finally` block than anything.

Second, it's a signal to the script that loading of the script has _not_ completed. The code most able to properly interrupt the script and clean itself up is the script itself. JRuby, for example, uses the scriptUnloaded hook being called before the scriptLoaded hook to know it's being asked to be interrupted, and send a Ruby exception through the script's main thread. This allows all resources to be cleaned up properly. AFAICT, JRuby doesn't even have a mechanism for the Java code to interrupt a running JRubyEngine. But perhaps a third hook could be added `scriptInterrupted` or something to make it more clear?

As for automatically re-enqueuing failed scripts, I have mixed opinions. It would certainly help the case of a low powered system simply timing out on boot. It would not help the case of a script that has a bug deadlocking or going into an infinite loop. It would just continue to do so on each reload. Okay, so we solve that with a max number of retries. How is that different than just doing a single retry with a longer timeout?

Anyhow, I'm more and more in favor of just reverting this for the 3.4.0 release. There seems to be a lot of intricacies that need worked out. And given how it's going so far, it definitely feels like if we fix one small thing with it, two more are going to crop up. Might as well stay with the bugs we know for this release.

cweitkamp:
With respect to the near release I vote for to temporarily revert the aforementioned integration of the safe caller. In a few days there will be a lot of time to improve the situation.

wborn:
Replaced by #3227

PR #3240, opened 2022-12-19
Remove redundant ctor in DecimalType

### LINKED ISSUE #2805: Remove redundant DecimalType constructors
In #2596 the `DecimalType(Number value)` constructor was added which makes the other constructors redundant:

* `DecimalType(BigDecimal value)`
* `DecimalType(long value)`
* `DecimalType(double value)`

They haven't been removed yet to maintain compatibility with previously compiled add-ons in the Marketplace (see #2803).
We should remove these when some other breaking change is introduced which breaks add-on compatibilty.

J-N-K:
Closes #2805


wborn (approved):
Thanks!

PR #2994, opened 2022-06-02
Raise minimum JDK version to 17 

J-N-K:
- Raise source/target level to 17 (except model classes due to missing support in Xtext https://github.com/eclipse/xtext/issues/1982)
- Remove Nashorn scripting engine
- Bump spotless to allow newer language features (requires JVM options due to https://github.com/diffplug/spotless/issues/834)

Signed-off-by: Jan N. Klug <github@klug.nrw>

PR #3250, opened 2022-12-22
Update Eclipse .classpath files

wborn:
Also adds some missing files for recently added projects.

Related to #2994


J-N-K (approved):
LGTM, thanks. I believe I can't create these files from IntelliJ IDEA.

wborn:
How is the OSGi/Bnd/lastNPE support in IntelliJ nowadays?
I use the IntelliJ IDEA Ultimate edition on other (even bigger) projects but those are Spring based. :upside_down_face:

J-N-K:
There is no editor support for bnd-files, but actually I don't need that very often. 

LastNPE is more difficult, there is no support for creating EEAs and null-checking is always done with the internal annotations. But on the other hand, those are usually better than those that are available in Eclipse (e.g. our null-checker still complains about `Map.putIfAbsent(key, new Foo())` which clearly is non-null).

PR #3251, opened 2022-12-22
Fix some recently introduced SAT and NPE issues

wborn:
Signed-off-by: Wouter Born <github@maindrain.net>

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/action/ScriptExecutionImpl.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/action/ScriptExecutionImpl.java
@@ -17,8 +17,6 @@ import java.util.function.Consumer;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
-import org.openhab.core.automation.RuleManager;
-import org.openhab.core.automation.RuleRegistry;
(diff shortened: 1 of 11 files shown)


bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/action/ScriptExecutionImpl.java
```diff
@@ -37,15 +35,10 @@
 public class ScriptExecutionImpl implements ScriptExecution {
 
     private final Scheduler scheduler;
-    private final RuleManager ruleManager;
-    private final RuleRegistry ruleRegistry;
 
     @Activate
-    public ScriptExecutionImpl(@Reference RuleRegistry ruleRegistry, @Reference RuleManager ruleManager,
```
J-N-K, line 44: That's a leftover from trying to also implement `callScript` here, but that's difficult.

J-N-K (approved):
LGTM, thanks.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/action/ScriptExecutionImpl.java
wborn: I was wondering if it was needed to prevent init issues during startup, but if that is the reason it can be safely removed. :+1:

PR #3252, opened 2022-12-22
Upgrade to Karaf 4.4.3

wborn:
* Sync runtime dependencies with Karaf 4.4.3, most notably:
  * Jetty 9.4.50.v20221201
  * Pax Logging 2.2.0
  * Pax Web 8.0.15
* Use OSGi R8 as compile dependency
* Rework Servlets to use Http Whiteboard annotations in favor of proprietary `org.openhab.core.io.http.servlet` classes
* Resolve itest runbundles

---

To do:

* [x] Rebase and fix merge conflicts
* [x] Upgrade to the latest greatest Karaf 4.4 version
* [x] Upgrade dependencies to use OSGi R8


J-N-K:
There are still issues. BasicUI starts, also charts and icons are available, but MainUI does not. 

WebSocket servlet is present, but fails to create the socket, I believe I'm running into the same issue as described here: https://github.com/eclipse/jetty.project/issues/7835.

wborn:
I did have Main UI working with my changes but I can imagine with https://github.com/openhab/openhab-webui/pull/1521 it needs to be reworked. The main issue I had with main UI was that it didn't allow for any other servlets to be registered. So that's why I went with those Whiteboard changes which is also more standard compared to the proprietary OpenHABServlet classes etc. I've added you as collaborator on my webui repo so I think you can also push to my webui PR branch now.

J-N-K:
MainUI works now (mostly, there is still one issue with serving images, but I guess that can be figured out.

One issue is that we declare `service = Foo.class` on some servlet components and then they are not registered. It is mandatory they also provide `Servlet.class`.

J-N-K:
Regarding core, on startup I see:

```
Launching the openHAB runtime...
WARNING: package org.apache.karaf.specs.locator not in java.base
Listening for transport dt_socket at address: 5005
Error starting karaf activator org.apache.karaf.specs.activator.Activator: org/apache/karaf/specs/locator/OsgiLocator
```

That looks similar to https://issues.apache.org/jira/browse/KARAF-5840

But I'm not sure if this is similar to

J-N-K:
This one looks pretty good now, apart from the itests that don't resolve. Maybe you can have a look at that, @wborn.

J-N-K:
I fixed the tests (here and in addons). If you review my contributions, I'm fine with yours and we can merge if the web issues are resolved.

J-N-K:
We'll have Karaf 4.4.3 probably tomorrow: https://lists.apache.org/list.html?dev@karaf.apache.org

wborn:
If it works well that would be a nice latest greatest version to use with the PRs. :slightly_smiling_face:

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/package-info.java
```diff
@@ -0,0 +1,20 @@
...
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+@org.osgi.annotation.bundle.Header(name = org.osgi.framework.Constants.DYNAMICIMPORT_PACKAGE, value = "*")
```
wborn, line 13: Is this dynamic import really necessary?

J-N-K: Yes, even though I don't fully understand why. But I found that this is the solution for https://github.com/openhab/openhab-core/pull/3252#issuecomment-1374418213.

wborn: Did you also try adding an import for `org.eclipse.jetty.websocket.server.WebSocketServerFactory` in the code instead? That way bnd will add an import for the package to the manifest.

itests/org.openhab.core.model.script.tests/itest.bndrun
```diff
@@ -4,9 +4,8 @@ Bundle-SymbolicName: ${project.artifactId}
 Fragment-Host: org.openhab.core.model.script
 
 -runblacklist: \
-    bnd.identity;id='org.osgi.service.cm',\
-    osgi.identity;filter:='(osgi.identity=org.apache.tomcat.util.descriptor.web)'
-#
+	bnd.identity;id='org.osgi.service.cm',\
+	osgi.identity;filter:='(osgi.identity=org.apache.tomcat.util.descriptor.web)'
```
J-N-K, line 8: same here

wborn: I only changed the indentation from spaces to tabs. :slightly_smiling_face:

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/package-info.java
J-N-K: That works, too. Why is that superior to the dynamic-import?

wborn: It will make sure that the required package imports are satisfied before the bundle is started so you'll know well in advance that something is wrong. That can be long before the code is executed that looks up the class.

wborn:
Thank you very much for your help @J-N-K. :+1: 
I think we are done with this so I think it is time for the other maintainers to review this.

wborn:
> Do we need to check zwave and zigbee?

AFAIK they don't use servlets so I think they are fine. :slightly_smiling_face:

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/AudioServlet.java
```diff
@@ -36,47 +38,39 @@
...
 
 /**
  * A servlet that serves audio streams via HTTP.
  *
  * @author Kai Kreuzer - Initial contribution
  */
+@Component(service = { AudioHTTPServer.class, Servlet.class })
+@HttpWhiteboardServletName(AudioServlet.SERVLET_PATH)
+@HttpWhiteboardServletPattern(AudioServlet.SERVLET_PATH + "/*")
 @NonNullByDefault
```
J-N-K, line 56: This is more a comment about code-style (and not necessarily to be enforced here): There are dozens of different arrangements for the annotations, sometime `@Component` is first and `@NonNullByDefault` is at the end, sometime it's exactly the opposite and sometime they are together with the whiteboard annotations above or below. 

Can we decide for one convention and probably enable https://github.com/checkstyle/checkstyle/issues/11693 if it is implemented in the customizable way someday? I would agree with https://stackoverflow.com/a/5638388 and order by importance, i.e. `@Component`, other and `@NonNullByDefault` at the end as it is present on nearly every class.

wborn: I usually sort them which most of the time also goes well with the order of importance.

J-N-K (approved):

J-N-K (approved):

wborn:
I will resolve the merge conflicts after the review is done. :wink:

kaikreuzer (approved):
Thanks for the work @wborn!
I have only glanced over the code and not tested it, but this is anyhow tricky - but what I see lgtm.

> I will resolve the merge conflicts after the review is done. 😉

It looks to me as if @J-N-K already reviewed and approved it 2 weeks ago - if you waited for me to approve as well: Here you go!

J-N-K:
@kaikreuzer we did wait for you because I contributed to the code 😀

florian-h05:
It seems that the developer docs are outdated: https://next.openhab.org/docs/developer/guidelines.html#d-language-levels-and-libraries
The say that openHAB is using OSGi R7, but isn’t it using R8 now?

PR #3253, opened 2022-12-22
[CLI/voice] use parameters for cli voice commands startDialog and listenAndAnswer

GiviMAD:
This is a proposal for simplify the usage of the cli voice commands startDialog and listenAndAnswer, passing the dialog options as parameters instead of arguments.

Related to this issue: https://github.com/openhab/openhab-core/issues/3140


GiviMAD:
I can simplify this based on changes suggested for the dialog builder on the review of https://github.com/openhab/openhab-core/pull/3264.

GiviMAD:
@J-N-K I needed to add the nullability to the parameters in the voice manager interface to use the get service methods in the same way with the builder.
This is fine, right?

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -941,8 +941,8 @@ public Collection<KSService> getKSs() {
     }
 
     @Override
-    public List<HumanLanguageInterpreter> getHLIsByIds(String ids) {
-        return getHLIsByIds(Arrays.asList(ids.split(",")));
+    public List<HumanLanguageInterpreter> getHLIsByIds(@Nullable String ids) {
+        return ids == null ? new ArrayList<>() : getHLIsByIds(Arrays.asList(ids.split(",")));
```
wborn, line 945: This code is shorter and returns a constant which prevents unneccessary garbage collection.

```suggestion
        return ids == null ? List.of() : getHLIsByIds(Arrays.asList(ids.split(",")));
```

GiviMAD: I didn't know about that, makes sense. Thanks for the information!

wborn (approved):
Thanks!

PR #3255, opened 2022-12-22
Refactor AbstractScriptFileWatcher

### LINKED ISSUE #3242: Script engine not properly removed when script loading fails
In https://github.com/openhab/openhab-addons/issues/14019 I described an issue with the JavaScript (GraalJS) Automation addon where a number of attempts to load a script that has a bug (so that the loading will fail) lead to OoM.

For me, it seems like the engine is not properly cleaned up when evaluation fails which leads to a memory leak and in consequence OoM.

I will take a heap dump to closer analyse this, but I have to wait for the refactoring of the watch service because currently file based scripts don't load.

/cc @J-N-K Ping. Can you add the bug label please?

### LINKED ISSUE #3245: Fix AbstractScriptFileWatcher checks scriptType in wrong place
JSScripting scripts fail to load when Nashorn is not available because the availability of a script engine is checked too early.

J-N-K:
Fixes #3245
Fixes #3242

Additionally the methods for dependency changes or start level changes now return fast. Changes to the script file watchers in openhab-addons are also needed and need to be merged together with this one.


J-N-K:
I wonder why the tests fail. I never get them to fail locally.

florian-h05:
@openhab/core-maintainers 
These fixes are really important for JS Scripting, because at the current state it is not possible to use file based rules with the 4.0.0-SNAPSHOTs and a corrupt script file can currently also lead to OoM because of a memory leak. This fixed these two super nasty bugs.

florian-h05:
@J-N-K Can you please ping me when this one is finished, so I can test it then?

J-N-K:
Regarding the test refactoring: Since I made all actual load/unload operation async (which is great improvement for the performance of other components that interact with the script file watcher), I had to find a way to figure out when an operation was completed. Some tests then failed (e.g. remove before add) because of the async nature load condition where fulfilled too early or too late so that we missed calls (in most cases) or had extra calls (in few cases).

J-N-K:
@florian-h05 I guess it's fine now. At least all tests are reliably working now and it also works in my dev system.

florian-h05 (approved):
I have tested together with https://github.com/openhab/openhab-addons/pull/14043, works great, thanks!

Finally it is possible to load file-based JS scripts on openHAB 4, and the engines are now also properly closed when script loading fails (I added a breakpoint in the close method).

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
```diff
@@ -202,158 +206,157 @@ protected boolean watchSubDirectories() {
...
-                return true;
+        try (InputStreamReader reader = new InputStreamReader(Files.newInputStream(ref.getScriptFilePath()),
+                StandardCharsets.UTF_8)) {
+            ScriptEngineContainer container = manager.createScriptEngine(ref.getScriptType(),
+                    ref.getScriptIdentifier());
+            if (container != null) {
+                container.getScriptEngine().put(ScriptEngine.FILENAME, scriptIdentifier);
+                if (manager.loadScript(container.getIdentifier(), reader)) {
+                    return true;
+                }
```
florian-h05, line 323: @J-N-K I am wondering whether the "Script loading error, ignoring file" log message shouldn't be here in an `else` statement, because the `else` statament of `container != null` will get called when creation of a script engine fails.
The actual loading of a script fails inside the container's null check if.

J-N-K: I think the `else` around the logging should be removed. The loading fails in both cases: `container==null` or script error.

florian-h05: Better solution, thanks :+1:

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
```diff
@@ -119,73 +119,76 @@ public void activate() {
...
-                URL url = file.toURI().toURL();
-                resources.add(new ScriptFileReference(url));
-            } catch (MalformedURLException e) {
-                // can't happen for the 'file' protocol handler with a correctly formatted URI
-                logger.debug("Can't create a URL", e);
-            }
         }
-        return resources;
+
+        return StartLevelService.STARTLEVEL_RULEENGINE;
```
J-N-K, line 183: @ccutrer @florian-h05 I wonder if this is really a good choice. Start level 40 is named "rules loaded", so shouldn't they be loaded at start level 30 instead after persistence has restored item states?

florian-h05: This is a good question.
Looking at https://github.com/openhab/openhab-distro/blob/main/distributions/openhab/src/main/resources/runtime/services.cfg, start level 40 is that dsl rules and UI rules are loaded. (If I understand that right?)
So I would say that JSR223 file based rules belong to start level 40 too.

I guess we also have to look out, that the addons are loaded before their scripts are loaded by the `AbstractScriptFileWatcher`, but honestly I have no experience with the start levels and the start order in openHAB.

J-N-K: The difficult thing to figure out is if all script file watchers have finished their initial script processing because we don't know which exist. 

If we move the "automation:scriptEngingeFactories" to start level 30, add a new interface `ScriptFileWatcher` with `void boolean initialLoadingCompleted` and set another ready marker "jsr223:initialLoadingCompleted" which is needed for start level 40.

ccutrer: My ideal situation:

 * JSR223 scripts by default act as close to DSL rules as possible. Seems like loading at SL30 would mirror that.
 * a user can override the start level to be as early as possible. I have scripts that don't add rules, but extend openHAB such as adding items or profiles, or custom units (I should probably get that one added to core). So if I can get something to run as early as SL10, that would be ideal.

I haven't looked too close into how the ScriptFileWatcher action actually starts, or when automation addons start. Mostly in ensuring the ordering of the configured start levels for scripts is respected.

J-N-K: @florian-h05 @ccutrer I have improved the code so that it handles startup in a more predictable way. There might be issues if files are modified prior to the initial import (i.e. before the system reached start level 30). 

But setting the import to start level 30 in conjunction with #3275 makes sure script engines are available before we import files. I also added a new interface that allows us to later set a ready marker for "all script file watchers finished their initial import", but I postponed implementation of that because it depends on #3275.

florian-h05: Sounds good, thanks :+1:

J-N-K:
@openhab/core-maintainers Can this be reviewed so we can proceed with the more breaking changes in #3004? Thanks.

wborn:
I had planned to have a look at it already but my 30y old trusty German washing machine broke down. But at least I now have a new German machine that I can integrate with openHAB. 😉

florian-h05:
@wborn 
Any plan when you can have a look? 
This fix is required to get file based JS scripts working on 4.0 and holds me back from testing the 4.0 SNAPSHOTS.

kaikreuzer:
@wborn Let me know if you are still busy with washing machines - if so, I can also take over the review.

wborn (approved):
Thanks for the refactoring and your patience! :slightly_smiling_face:

PR #3261, opened 2022-12-23
Remove wrong DataAmount units

wborn:
These wrong DataAmount units were added in #3208 for backwards compatibility. In OH4 we can clean them up because this major version allows for breaking changes.


J-N-K:
LGTM. Two remarks:

- WDYT about `kiB` vs `kB`? In my world `1 kB = 1024 b`, not `1 kB = 1000 b`. I have never seen anyone using that (except HDD manufacturers trying to polish their numbers for marketing purposes).
- Would it make sense to track all these changes (and what is coming for UoM with the default units) somewhere so we can write a "How to prepare your system for OH4" guide?

wborn:
Did you read the Wikipedia article?

https://en.m.wikipedia.org/wiki/Kilobyte

I agree it's a bit strange but if you use them as SI prefixes as recommended by the IEC it makes sense. If you use macOS it apparently also uses SI based units while Windows uses "bibyte" bases units.

Having some docs on how to prepare/migrate is a good thing. 👍  That's why I also added the "API breaking" label so it's easy to find back breaking changes.

J-N-K (approved):

lolodomo:
Please start listing somewhere all the introduced breaking changes.
At least tag the PRs so that we can find them later.

PR #3264, opened 2022-12-23
[voice] Support managed dialogs

GiviMAD:
This PR adds a new functionality to the voice manager for register/unregister dialog descriptions.
The voice manager will try to keep a started dialog for each registered description (and stop the dialog if it's unregistered).
The dialog descriptions are persisted using a storage.

This PR do not introduce any code breaking changes, but introduces the behavior to stop the dialogs whenever a service in use became unavailable. 

Related to https://github.com/openhab/openhab-core/issues/3265

<s>This is a draft because integration tests are missing.</s>

### DIFF
--- a/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/DialogContext.java
+++ b/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/DialogContext.java
@@ -137,17 +137,23 @@ public class DialogContext {
         }
 
         public Builder withKS(@Nullable KSService service) {
-            this.ks = service;
+            if (service != null) {
+                this.ks = service;
+            }
             return this;
         }
 
         public Builder withSTT(@Nullable STTService service) {
-            this.stt = service;
+            if (service != null) {
+                this.stt = service;
+            }
             return this;
         }
 
         public Builder withTTS(@Nullable TTSService service) {
-            this.tts = service;
+            if (service != null) {
+                this.tts = service;
+            }
             return this;
         }
 
@@ -156,32 +162,44 @@ public class DialogContext {
         }
 
         public Builder withHLIs(List<HumanLanguageInterpreter> services) {
-            this.hlis = services;
+            if (!services.isEmpty()) {
+                this.hlis = services;
+            }
             return this;
         }
 
-        public Builder withKeyword(String keyword) {
-            this.keyword = keyword;
+        public Builder withKeyword(@Nullable String keyword) {
+            if (keyword != null && !keyword.isBlank()) {
+                this.keyword = keyword;
+            }
             return this;
         }
 
         public Builder withVoice(@Nullable Voice voice) {
-            this.voice = voice;
+            if (voice != null) {
+                this.voice = voice;
+            }
             return this;
         }
 
         public Builder withListeningItem(@Nullable String listeningItem) {
-            this.listeningItem = listeningItem;
+            if (listeningItem != null) {
+                this.listeningItem = listeningItem;
+            }
             return this;
         }
 
         public Builder withMelody(@Nullable String listeningMelody) {
-            this.listeningMelody = listeningMelody;
+            if (listeningMelody != null) {
+                this.listeningMelody = listeningMelody;
+            }
             return this;
         }
 
-        public Builder withLocale(Locale locale) {
-            this.locale = locale;
+        public Builder withLocale(@Nullable Locale locale) {
+            if (locale != null) {
+                this.locale = locale;
+            }
             return this;
         }
 
--- a/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/DialogRegistration.java
+++ b/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/DialogRegistration.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2010-2022 Contributors to the openHAB project
+ * Copyright (c) 2010-2023 Contributors to the openHAB project
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information.
--- a/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/DialogRegistration.java
+++ b/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/DialogRegistration.java
@@ -1,0 +1,81 @@
+/**
+ * Copyright (c) 2010-2022 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.voice;
+
+import java.util.List;
+import java.util.Locale;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+
+/**
+ * Describes dialog desired services and options.
+ *
+ * @author Miguel Álvarez - Initial contribution
+ */
+@NonNullByDefault
+public class DialogRegistration {
+    /**
+     * Dialog audio source id
+     */
+    public String sourceId;
+    /**
+     * Dialog audio sink id
+     */
+    public String sinkId;
+    /**
+     * Preferred keyword-spotting service
+     */
+    public @Nullable String ksId;
+    /**
+     * Selected keyword for spotting
+     */
+    public @Nullable String keyword;
+    /**
+     * Preferred speech-to-text service id
+     */
+    public @Nullable String sttId;
+    /**
+     * Preferred text-to-speech service id
+     */
+    public @Nullable String ttsId;
+    /**
+     * Preferred voice id
+     */
+    public @Nullable String voiceId;
+    /**
+     * List of interpreters
+     */
+    public List<String> hliIds = List.of();
+    /**
+     * Dialog locale
+     */
+    public @Nullable Locale locale;
+    /**
+     * Linked listening item
+     */
+    public @Nullable String listeningItem;
+    /**
+     * Custom listening melody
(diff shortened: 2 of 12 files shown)


bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -951,8 +1038,92 @@ private Comparator<Voice> createVoiceComparator(Locale locale) {
...
+            if (dialogsToStop.isEmpty()) {
+                return;
+            }
+            for (var dialog : dialogsToStop) {
+                stopDialog(dialog.dialogContext.source());
+            }
+        }
+    }
+
+    private void debounceBuildDialogRegistrations() {
```
J-N-K, line 1053: Why is this needed?

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/DialogRegistration.java
```diff
@@ -0,0 +1,81 @@
+/**
+ * Copyright (c) 2010-2022 Contributors to the openHAB project
```
J-N-K, line 2:
```suggestion
 * Copyright (c) 2010-2023 Contributors to the openHAB project
```

J-N-K (changes requested):
What is the use-case for that?

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
GiviMAD: It's not really needed, but I think it will skip a lot of unnecessary executions, as this needs to be called on every related add-on registration but there is not need to do it immediately, I added it thinking in the following cases:

* On the initial start of openHAB this method will be called a lot of times.
* Normally the sink and source related to the dialog will bellow to the same add-on so registration will happen mostly at same time.

Let me know if you think is better to remove/improve it.

GiviMAD:
> What is the use-case for that?

For the new methods? The final use-case is exposing a way users can configure dialogs that openHAB will keep running whenever is possible event after an openHAB reboot.

I'll explain in more detail what is my point of view after my experience and some feedback on the marketplace:

As an add-on developer the current methods startDialog and stopDialog are perfect to consume , I'm using those on the HABSpeaker add-on you can find at the marketplace, where I'm playing with audio recognition on the browser and testing other ideas.

But as an openHAB user or an UI client the startDialog and stopDialog functionality is not really useful, because what those consumer want is to be able to configure the desired services for a dialog and that the system keep it running whenever is possible (this involves the initial dialog start, and restart the dialog after a reboot or a related bundle update), be able to check its state, and unregister it whenever you don't what it any more.

So my proposal for solving this is adding these new methods that will be exposed through the console but also to the API hopping some UI contributor is interested in build a configuration page for the dialogs which I think can consist on a list of dialog registrations and a creation form that can be pretty similar to the voice options configuration page. This way the user will be able to manage its dialogs from the UI.

And thank you for the review! :)

GiviMAD:
Of course let me know if you think there is a better way for adding this functionality. If this looks ok to you I can add a more tests because right now I've just included a basic one that do not tests all the scenarios.

GiviMAD:
@J-N-K did you have a change to look at this?

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -951,8 +1039,92 @@ private Comparator<Voice> createVoiceComparator(Locale locale) {
...
+    }
+
+    private void tryBuildDialogRegistration(DialogRegistration dialogRegistration) {
+        var builder = getDialogContextBuilder().withSink(audioManager.getSink(dialogRegistration.sinkId))
+                .withSource(audioManager.getSource(dialogRegistration.sourceId));
+        String ksId = dialogRegistration.ksId;
+        if (ksId != null) {
+            builder.withKS(getKS(ksId));
+        }
+        String keyword = dialogRegistration.keyword;
```
J-N-K, line 1079: You do a lot of these null-checks in several places. IMO it would be better to accept `null` in the builder and discard the value there. You can also improve `getKS` and similar:

```
    @Override
    public @Nullable KSService getKS(@Nullable String id) {
        return id == null ? null : ksServices.get(id);
    }
```

then you can really make good use of the builder and concatenate the `.withXY` calls.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
J-N-K: Understood. Looks useful. Maybe it's better to call it `scheduleDialogRegistrations` and add JavaDoc explaining that each call re-schedules the build.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -700,6 +714,50 @@ public void listenAndAnswer(DialogContext context) throws IllegalStateException
...
+    public void unregisterDialog(DialogRegistration registration) {
+        unregisterDialog(registration.sourceId);
+    }
+
+    @Override
+    public void unregisterDialog(String sourceId) {
+        synchronized (dialogRegistrationStorage) {
+            var registrationRef = dialogRegistrationStorage.remove(sourceId);
+            if (registrationRef != null) {
+                dialogRegistrationStorage.remove(sourceId);
```
J-N-K, line 739: Why do you remove it twice?

J-N-K (changes requested):
Sorry for responding so late. I completely forgot about this. I have left some comments, otherwise LGTM.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
GiviMAD: Not needed, my fault. Thank you!

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
GiviMAD: I don't like it either, but I didn't know how to solved.
I think I have understand your solution looks great, I'll go for it.

GiviMAD: Looks much better now, thank you for proposing this change.
I can take advantage of it also on this one https://github.com/openhab/openhab-core/pull/3253.

GiviMAD:
Sorry, I saw an error and I've added a commit after asking for the review.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/DialogContext.java
```diff
@@ -137,17 +137,23 @@ public Builder withSink(@Nullable AudioSink sink) {
         }
 
         public Builder withKS(@Nullable KSService service) {
-            this.ks = service;
+            if (service != null) {
+                this.ks = service;
+            }
             return this;
```
J-N-K, line 143: I wonder if you need to do a check at all here. The field is annotated as `@Nullable` anyway, so setting it to `null` will not change anything. Or is there a place where a valid value could be overwritten?

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -951,8 +1040,73 @@ private Comparator<Voice> createVoiceComparator(Locale locale) {
...
+     * this method schedules a call to {@link #buildDialogRegistrations() buildDialogRegistrations} in five seconds
+     * and cancel the previous scheduled call if any.
+     */
+    private void scheduleDialogRegistrations() {
+        if (dialogRegistrationFuture != null) {
+            dialogRegistrationFuture.cancel(false);
+        }
+        dialogRegistrationFuture = scheduledExecutorService.schedule(this::buildDialogRegistrations, 5,
+                TimeUnit.SECONDS);
+    }
```
J-N-K, line 1066:
```suggestion
    private void scheduleDialogRegistrations() {
        ScheduledFuture<?> job = this.dialogRegistrationFuture;
        if (job != null) {
            job.cancel(false);
        }
        dialogRegistrationFuture = scheduledExecutorService.schedule(this::buildDialogRegistrations, 5,
                TimeUnit.SECONDS);
    }
```

J-N-K (commented):
Otherwise LGTM

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/DialogContext.java
GiviMAD: Yes, that is the reason, it contains the default "keyword spotting" service if any, the reason for it to be nullable is that the listenAndAnswer functionality can be used without it.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -151,7 +164,13 @@ protected void activate(BundleContext bundleContext, Map<String, Object> config)
 
     @Deactivate
     protected void deactivate() {
-        stopAllDialogs();
+        dialogProcessors.values().forEach(DialogProcessor::stop);
+        dialogProcessors.clear();
+        var dialogRegistrationFuture = this.dialogRegistrationFuture;
```
J-N-K, line 169: I'm really sorry, I didn't notice before and have another request. We avoid using `var`. Since this is only cosmetic, it should be a fast-fix. I'll merge afterwards, promised.

PR #3268, opened 2022-12-24
Initial contribution of an CLI upgrade-tool

J-N-K:
This is the initial contribution for a tool that can perform more complex upgrade tasks. The idea is to provide an easy way for the user to perform one-time upgrade operations to the databases that are optional but useful.

The first task is the migration of the unit of a `NumberItem` with dimension from the state description to the new `defaultUnit` metadata introduced in #3248. Since there might be side-effects this should not be performed automatically. For textual configuration this can easily be done with a text editor or command line-tools. For UI defined items it's more difficult since two files need to be considered.

Other tasks might be automatically executed by the update script if necessary.

### DIFF
--- a/tools/upgradetool/src/main/java/org/openhab/core/tools/UpgradeTool.java
+++ b/tools/upgradetool/src/main/java/org/openhab/core/tools/UpgradeTool.java
@@ -12,6 +12,8 @@
  */
 package org.openhab.core.tools;
 
+import java.util.Set;
+
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.DefaultParser;
 import org.apache.commons.cli.HelpFormatter;
@@ -29,15 +31,19 @@ import org.openhab.core.tools.internal.Upgrader;
  */
 @NonNullByDefault
 public class UpgradeTool {
+    private static final Set<String> LOG_LEVELS = Set.of("TRACE", "DEBUG", "INFO", "WARN", "ERROR");
     private static final String CMD_OPT_ITEM = "item";
     private static final String CMD_OPT_LINK = "link";
     private static final String CMD_OPT_DIR = "dir";
+    private static final String CMD_OPT_LOG = "log";
 
     private static Options getOptions() {
         Options options = new Options();
 
         options.addOption(
                 Option.builder().longOpt(CMD_OPT_DIR).desc("directory to process").numberOfArgs(1).required().build());
+
+        options.addOption(Option.builder().longOpt(CMD_OPT_LOG).numberOfArgs(1).desc("log verbosity").build());
 
         // add the group for available options
         OptionGroup operation = new OptionGroup();
@@ -46,7 +52,6 @@ public class UpgradeTool {
                 .desc("perform the given operation on the item database").build());
         operation.addOption(Option.builder().longOpt(CMD_OPT_LINK).numberOfArgs(1)
                 .desc("perform the given operation on the link database").build());
-
         options.addOptionGroup(operation);
 
         return options;
@@ -56,7 +61,15 @@ public class UpgradeTool {
         Options options = getOptions();
         try {
             CommandLine commandLine = new DefaultParser().parse(options, args);
-            System.setProperty(org.slf4j.impl.SimpleLogger.DEFAULT_LOG_LEVEL_KEY, "TRACE");
+
+            String loglevel = commandLine.hasOption(CMD_OPT_LOG) ? commandLine.getOptionValue(CMD_OPT_LOG).toUpperCase()
+                    : "INFO";
+            if (!LOG_LEVELS.contains(loglevel)) {
+                System.out.println("Allowed log-levels are " + LOG_LEVELS);
+                System.exit(0);
+            }
+
+            System.setProperty(org.slf4j.impl.SimpleLogger.DEFAULT_LOG_LEVEL_KEY, loglevel);
 
             String baseDir = commandLine.hasOption(CMD_OPT_DIR) ? commandLine.getOptionValue(CMD_OPT_DIR) : "";
             Upgrader upgrader = new Upgrader();
--- a/tools/upgradetool/src/main/java/org/openhab/core/tools/UpgradeTool.java
+++ b/tools/upgradetool/src/main/java/org/openhab/core/tools/UpgradeTool.java
@@ -12,13 +12,14 @@
  */
 package org.openhab.core.tools;
 
+import static org.openhab.core.tools.internal.Upgrader.*;
+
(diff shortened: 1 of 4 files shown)


stefan-hoehn:
I would support that because with 4.0 we need a mechanism to automatically convert all rules from "application/javascript" to "application/javascript;version=ECMAScript-5.1" or would there be another way to do automatic migrations?

J-N-K:
@jimtng FYI, the latest commit contains all that is needed, so it's super easy. After compiling you would need to run 
```
java -jar upgradetool-4.0.0-SNAPSHOT-jar-with-dependencies.jar --dir path/to/userdata --link upgradeJsProfile
``

File: tools/upgradetool/src/main/java/org/openhab/core/tools/UpgradeTool.java [hunk 1]
```diff
@@ -0,0 +1,190 @@
...
+
+        List.copyOf(linkStorage.getKeys()).forEach(linkUid -> {
+            ItemChannelLink link = Objects.requireNonNull(linkStorage.get(linkUid));
+            Configuration configuration = link.getConfiguration();
+            String profileName = (String) configuration.get(ItemChannelLinkConfigDescriptionProvider.PARAM_PROFILE);
+            if ("transform:JS".equals(profileName)) {
+                String function = (String) configuration.get("function");
+                if (function != null) {
+                    configuration.put("toItemScript", function);
+                    configuration.put("toHandlerScript", function);
```
jimtng, line 145: The [original behaviour](https://github.com/openhab/openhab-addons/blob/ab910cca9851ee0c3ac06625ebf1d126073dde6d/bundles/org.openhab.transform.javascript/src/main/java/org/openhab/transform/javascript/internal/profiles/JavaScriptTransformationProfile.java#L88-L90) was to pass through the command back to the handler. I'd suggest either not setting this at all (discard) or assign it to `"|input"` to preserve the same behaviour.

jimtng (changes requested):

tools/upgradetool/src/main/java/org/openhab/core/tools/UpgradeTool.java
```diff
@@ -0,0 +1,190 @@
...
+                    configuration.remove("function");
```
jimtng, line 146: Should we also remove [sourceFormat](https://github.com/openhab/openhab-addons/blob/ab910cca9851ee0c3ac06625ebf1d126073dde6d/bundles/org.openhab.transform.javascript/src/main/java/org/openhab/transform/javascript/internal/profiles/JavaScriptTransformationProfile.java#L45)?

kaikreuzer:
Thanks, the code looks good, but we might have to improve the user experience for it. At the moment, the usage says:
```
usage: upgradetool
    --dir <arg>    directory to process
    --item <arg>   perform the given operation on the item database
    --link <arg>   perform the given operation on the link database
    --log <arg>    log verbosity
```
To the user, it is not clear, which arguments for item, link (and maybe also log) are expected. It can only be found out by providing an illegal argument for these options, which is not very user friendly.
Ideally, I would say, the tool should execute all potentially required upgrade tasks without requiring passing any arguments.
(But in this case, it should not throw errors if no json db file exists, e.g. because only textual definitions exist.)

The `--dir` option should be optional - if not provided, the `$OPENHAB_USERDATA` env variable should be automatically used.

Also, it might be helpful to add scripts to `runtime/bin`, so that it is easy for users to launch the script without having to download it and call `java -jar upgradetool-4.0.0-SNAPSHOT-jar-with-dependencies.jar`.

Wdyt?

J-N-K:
I fully agree with all you say :-)

The question is (and there I'm not 100% sure about the "best" answer): Should the tool run automatically if an update is made? If so, what happens 

- If someone in the future upgrades from 3.4 to 4.1? We would require the steps from 3.4 to 4.0 (and possible additional steps from 4.0.0 to 4.1.0). 
- If someone upgrades from 4.0.0.M4 to 4.0.0? In that case we probably don't need any steps, because the updates have also been processed when the update to 4.0.0.M4 took place.

It's not that easy to determine the needed steps automatically. OTOH, requiring a manual step (yes, that should be an easy script like `bin/migrate 4.0.0`) is not that user-friendly.

What is already implemented is that already existing unit metadata is not overwritten, but if someone wants to use the system default (`°C/°F`) for internal representation (and persistence) there is no need to set unit metadata. If he then sets the state description to `K` everything will be fine (that's the whole point of the de-coupling in #3248). If the update tool is called, it'll detect a unit in the state description, and then write that unit to metadata (which is now wrong). 

This is why I implemented different steps instead of a single step that does everything, so updates can be done selectively when needed.

kaikreuzer:
> Should the tool run automatically if an update is made?

Ideally, it should be called by the "update" script that we anyhow already have in place for doing version updates.

> If someone in the future upgrades from 3.4 to 4.1? We would require the steps from 3.4 to 4.0 (and possible additional steps from 4.0.0 to 4.1.0).

This has been solved in the update script as well already by having information on which action is required between which versions (in the [update.lst](https://github.com/openhab/openhab-distro/blob/main/distributions/openhab/src/main/resources/bin/update.lst#L203-L212)). Something similar would probably be useful here.

> If someone upgrades from 4.0.0.M4 to 4.0.0? In that case we probably don't need any steps

If a new step becomes relevant between 4.0.0.M4 and 4.0.0.M5, though, it would need to be applied. "Inter-snapshot" or "inter-milestone" updates are more tricky to handle indeed.

> It's not that easy to determine the needed steps automatically.

The best situation would be, if there wouldn't be any harm to re-apply them. If they manage to check themselves, if there is something outdated that needs to be migrated, it would be smooth for the user.

> If he then sets the state description to K everything will be fine

Not sure how relevant this case is - I'd assume that users will almost never use K in state descriptions, but rather always °C or °F. Are there other examples where "re-applying" might cause issues?
Another option to make sure to only apply a step once would be to somewhere store the information that a certain step has already been applied and should not be applied anymore in future.

J-N-K:
```
usage: upgradetool [--command <arg>] --dir <arg> [--force] [--log <arg>]
    --command <arg>   command to execute
    --dir <arg>       directory to process
    --force           force execution (even if already done)
    --log <arg>       log verbosity
Available commands: [itemCopyUnitToMetadata, linkUpgradeJsProfile]
```

Do you think this is a better interface? I also implemented `--force` which allows to execute a given command a second time. Without it, it is blocked.

kaikreuzer (approved):
Looks good, let's go for this as a first version. For the final release, we can still check how we can include it into the distro to make it easy/automatic to launch the tool.

PR #3271, opened 2022-12-27
Remove AccessController.doPrivileged calls

J-N-K:
The `AccessController` and the `SecurityManager` is deprecated for removal in Java 17. We don't make use of the `SecurityManager` anyway, so we can safely remove it.


wborn (approved):
It's nice to see these deprecations fixed!

PR #3273, opened 2022-12-28
Ensure metadata available at startlevel 20

J-N-K:
This makes the `MetadataRegistry` report when the elements of the `ManagedMetadataProvider` is available via a ready marker (like other registries). This is needed to properly determine the unit for restoring values from persistence.  I labelled it as "bug" because I believe it could explain some strange behavior that has been reported in rare cases when restoring values during startup.

This is independent from the discussion in #3248.


kaikreuzer (approved):
It took me a moment to see that https://github.com/openhab/openhab-distro/pull/1451 is the other part of the fix - together it all makes sense then. 😄

PR #3277, opened 2022-12-28
Fix DateTimeTrigger not triggering on restored state

J-N-K:
Reported on the forum: https://community.openhab.org/t/trigger-based-on-datetime-item-not-firing-after-restart-server/142608

During startup most likely the item state is `UndefType.NULL` which results in no triggering. States are restored from persistence by directly setting the item state, not by an `ItemStateEvent` and were therefore missed. Since changing the trigger is not necessary when the state updates but does not change, the event listening was changed to `ItemStateChangedEvent` only.


kaikreuzer (approved):
lgtm, thanks!

PR #3278, opened 2022-12-28
Trigger immediately when start level already reached

### LINKED ISSUE #3199: JSR223 scripts and start levels
We're in the process of implementing the system start level trigger in the JRuby helper library, and a few things have come up that I noticed.

I was surprised that if you create a rule with the `core.SystemStartlevelTrigger` trigger, it will _not_ fire if the system has already passed that start level when the rule is created. Whenever using a ReadyService.ReadyTracker within core or an add-on, the tracker is notified of markers matching its filter as soon as it is registered. I think this isn't a problem for DSL and UI rules, because the rule engine won't start until all of those rules are loaded. At which point, the rule engine will execute those rules which have a start level that's already satisfied.

However, when creating rules using scripts, the ScriptFileWatcher is completely asynchronous from the start level service. Scripts have an associated start level (default 50 - rule engine), which they're not allowed to run before. But loading of all scripts is ordered, and the rest of openHAB is allowed to continue loading in the background. So it's completely possible to have many scripts, set to run at default level 50, but say one at the end registers a SystemStartlevelTrigger at level 100, but if openHAB reaches level 100 _before_ that script gets around to loading, it will never fire.

I see a couple possible solutions:
 * `SystemTriggerHandler` checks the start level when it's activated, and triggers immediately if the level has already been reached. This does mean that if a rules file is reloaded/UI rule is re-saved, it will immediately trigger. I'm not sure if this is a bad thing, but is a change in behavior.
 * StartLevelService's default config is changed requiring a ready marker from ScriptFileWatcher before progressing to start level 40 (or possibly each individual level?). This is touchy for a few reasons though. ScriptFileWatcher can hold scripts at arbitrary start levels, so you have to be careful that only scripts at lesser start levels are required to load before marking it as ready. Secondly, some languages (JS, and shortly Ruby) use their own ScriptFileWatcher, and thus would have their own ReadyMarkers. Perhaps we could dynamically change the config for StartLevelService, but that seems perilous. I relative recently fixed a bug that was deadlocking StartLevelService if its config was changed, so it's clearly not a well-trodden path. While more difficult to implement, this could have additional benefits - like if scripts are providing items and thing, and then other components (like HomeKit integration) are waiting for start level model, they could be sure that the items they depend on are available before they start.

Tangentially... what are thoughts on expanding the `core.SystemStartlevelTrigger` trigger - _not_ require a configuration for a particular start level, and then to trigger every time a new start level is reached? The triggering start level is already included in the outputs from the trigger, so the action can do further processing. Right now the way to do it is just have multiple triggers with varying start level configurations, so a reasonable work already exists. Unless of course solution 1 above is taken, and suddenly the rule is triggered 5 times, once for each trigger!

J-N-K:
Fixes #3199

The `StartLevelTrigger` did not fire when the start level was already above the desired start level. Also added some tests.

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/factory/CoreModuleHandlerFactory.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/factory/CoreModuleHandlerFactory.java
@@ -42,6 +42,7 @@ import org.openhab.core.automation.internal.module.handler.ThingStatusTriggerHan
 import org.openhab.core.events.EventPublisher;
 import org.openhab.core.i18n.TimeZoneProvider;
 import org.openhab.core.items.ItemRegistry;
+import org.openhab.core.service.StartLevelService;
 import org.osgi.framework.BundleContext;
 import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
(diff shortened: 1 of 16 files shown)


J-N-K:
An alternative approach would be similar to the rule-engine initialization: when the rule is added, the it is checked if the trigger is present and the rule is executed.

kaikreuzer (approved):
Lgtm and I prefer this over your alternative suggestion.
Could you please rebase the branch?

J-N-K:
Done

PR #3280, opened 2022-12-29
Only call getVolume when necessary

jimtng:


J-N-K (approved):
LGTM. Although I don't feel comfortable with the lack of tests for this part of the code. I would very much welcome if better test coverage would be added to the voice bundle in the future.

PR #3287, opened 2022-12-30
[core] Add DateTimeType method for returning Instant

jlaur:
Wish from forum post:
- **Original post:** https://community.openhab.org/t/openhab-4-0-wishlist/142388/78
- **Conclusion:** https://community.openhab.org/t/openhab-4-0-wishlist/142388/107

In relation to #2898 the plan would be to internally switch from `ZonedDateTime` to `Instant` while preserving backwards compatibility. This means `getZonedDateTime()` will stay, but it will eventually be changed to return a `ZonedDateTime` with the currently configured time-zone based on the internal `Instant`.

Therefore it would only be natural to also add a method for returning the `Instant` representation of the current object. This does not need to wait for the additional refactoring as it will provide user value right away, but implementation will of course change when we figure out how to make `TimeZoneProvider` accessible from `DateTimeType`.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
@@ -119,6 +119,15 @@ public class DateTimeType implements PrimitiveType, State, Command {
         return zonedDateTime;
     }
 
+    /**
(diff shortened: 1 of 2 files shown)


J-N-K (approved):
LGTM.

mhilbush:
Thank you @jlaur !

PR #3288, opened 2022-12-30
Fix Marketplace add-on range pattern matching

wborn:
With this fix it will use the correct string for pattern matching add-on version ranges introduced in #2811.


J-N-K (approved):
LGTM. Is there a reason this is a draft?

wborn:
It still acts strange for snapshot versions so I'm looking into that.

wborn:
It seems to work OK. I had forgotten 4.0.0 milestone/snapshot version are considered compatible when using a `[3.3.0;4.0.0)` range. It is also a bit confusing that it takes a while before the REST API returns topic title changes. Some additional debug logging helps with explaining what is happening.

J-N-K (approved):
Thanks!

PR #3290, opened 2022-12-31
[serial] Ability to use /dev/serial/ symlinks for usb serial device

dalgwen:
The rxtx serial port provider is now able to scan symlink in /dev/serial/by-id and use them as valid identifier.
Benefit : users won't have to fix the path of their USB serial device  with udev anymore, they can now use the explicit and stable symlink directly.

EDIT: explaining how: 

 **OLD, not relevant anymore ---------------**
~~1. First, we get as usual all the `CommPortIdentifier` by scan and properties. (I modified these methods to return list instead of stream to avoid unnecessary conversion)
2. Then if we are on linux (to do this I use `System.getProperty("os.name")` and check if "linux" is present, is there something better ?), we add the result of the new scan. What does this new scan ? : 
a - list symlinks inside `/dev/serial/by-id/`
b- trying to resolve each of them against the already scanned ports in step 1 : if it resolves to a real port, then it is valid and we can add it. I add it as a `SerialPortIdentifierImpl` with, as a name, the symlink. So it is basically the same port as those gotten in 1, but with a different name.

To do so I also needed to change `SerialPortIdentifierImpl` to allow using another name than the id of the `CommPortIdentifier` (i.e. i now use the link itself).~~
**------------ OLD**

(complete rewrite after comment)

1- The DiscoveryService signals to listener (including the RXTX provider) the serial devices found, by scanning /sys/ and now also by scanning /dev/serial-by-id
2- The RXTX provider is able to handle symlink but has a bug with long name (80+ char) (which is sometimes the case of symlink). I did a workaround by resolving the symlink before passing it to the native nrserial library.

Result :
- Now when querying to get a serial port by the symlink, it works by giving the real device.
- And also in the GUI, all parameters of context-type serial are pre-filled with plain old /dev/tty* and also symlinks.

### DIFF
--- a/bundles/org.openhab.core.io.transport.serial.rxtx/src/main/java/org/openhab/core/io/transport/serial/internal/RxTxPortProvider.java
+++ b/bundles/org.openhab.core.io.transport.serial.rxtx/src/main/java/org/openhab/core/io/transport/serial/internal/RxTxPortProvider.java
@@ -12,7 +12,10 @@
  */
 package org.openhab.core.io.transport.serial.internal;
 
+import java.io.IOException;
 import java.net.URI;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
@@ -36,6 +39,7 @@ import gnu.io.NoSuchPortException;
  *
  * @author Matthias Steigenberger - Initial contribution
  * @author Wouter Born - Fix serial ports missing when ports are added to system property
+ * @author Gwendal Roulleau - Workaround for long path issue by resolving symlink
  */
 @NonNullByDefault
 @Component(service = SerialPortProvider.class)
@@ -45,11 +49,19 @@ public class RxTxPortProvider implements SerialPortProvider {
 
     @Override
     public @Nullable SerialPortIdentifier getPortIdentifier(URI port) {
+        String portPathAsString = port.getPath();
         try {
-            CommPortIdentifier ident = SerialPortUtil.getPortIdentifier(port.getPath());
+            // Resolving symbolic link is needed because of a bug with nrjavaserial
+            // Until a new release with pull request #230 is included in openHAB,
+            // we keep resolving symbolic link here
+            Path portPath = Path.of(portPathAsString);
+            if (Files.isSymbolicLink(portPath)) {
+                portPathAsString = portPath.toRealPath().toString();
+            }
+            CommPortIdentifier ident = SerialPortUtil.getPortIdentifier(portPathAsString);
             return new SerialPortIdentifierImpl(ident);
-        } catch (NoSuchPortException e) {
-            logger.debug("No SerialPortIdentifier found for: {}", port.getPath());
+        } catch (NoSuchPortException | IOException e) {
+            logger.debug("No SerialPortIdentifier found for: {}", portPathAsString, e);
             return null;
         }
     }
--- a/bundles/org.openhab.core.config.discovery.usbserial.linuxsysfs/src/main/java/org/openhab/core/config/discovery/usbserial/linuxsysfs/internal/DeltaUsbSerialScanner.java
+++ b/bundles/org.openhab.core.config.discovery.usbserial.linuxsysfs/src/main/java/org/openhab/core/config/discovery/usbserial/linuxsysfs/internal/DeltaUsbSerialScanner.java
@@ -37,6 +37,10 @@ public class DeltaUsbSerialScanner {
         this.usbSerialScanner = usbSerialScanner;
     }
 
+    public Set<UsbSerialDeviceInformation> getLastScanResult() {
+        return lastScanResult;
+    }
+
     /**
      * Scans for USB-Serial devices, and returns the delta to the last scan result.
      * <p/>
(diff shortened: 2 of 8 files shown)


bundles/org.openhab.core.io.transport.serial.rxtx/src/main/java/org/openhab/core/io/transport/serial/internal/RxTxPortProvider.java
```diff
@@ -61,12 +70,49 @@ public Stream<ProtocolType> getAcceptedProtocols() {
...
+        List<CommPortIdentifier> comPortIdentifiers = Stream.concat(scanIds.stream(), propIds.stream())
+                .filter(distinctByKey(CommPortIdentifier::getName))
+                .filter(id -> id.getPortType() == CommPortIdentifier.PORT_SERIAL).collect(Collectors.toList());
+
+        String osName = System.getProperty("os.name");
+        if (osName != null && osName.toLowerCase().contains("linux")) {
+            // add links to the list :
+            return Stream.concat(comPortIdentifiers.stream().map(sid -> new SerialPortIdentifierImpl(sid)),
+                    findDevLink(comPortIdentifiers));
+        } else {
```
wborn, line 85: Wouldn't it be possible to add this OS specific discovery to the 
`org.openhab.core.config.discovery.usbserial.linuxsysfs'  bundle instead? That way it can be used with any serial transport and not just RXTX.

dalgwen: You are totally right.

1- I completely revert my original commit, and rewrite it using the discovery service. 

2- I was wrong, I didn't fix the issue with long path name you mentionned on the  community discussion (I didn't resolve the symlink for the real connection, only for discovery). So I now fixed that in the RXTX provider by resolving it before passing to the native provider. I could also have done it in SerialPortManagerImpl, but I don't think I should meddle with it as it also impact other providers ?

bundles/org.openhab.core.io.transport.serial.rxtx/src/main/java/org/openhab/core/io/transport/serial/internal/RxTxPortProvider.java
```diff
@@ -45,11 +49,19 @@ public class RxTxPortProvider implements SerialPortProvider {
...
+            Path portPath = Path.of(portPathAsString);
+            if (Files.isSymbolicLink(portPath)) {
+                portPathAsString = portPath.toRealPath().toString();
+            }
+            CommPortIdentifier ident = SerialPortUtil.getPortIdentifier(portPathAsString);
             return new SerialPortIdentifierImpl(ident);
-        } catch (NoSuchPortException e) {
-            logger.debug("No SerialPortIdentifier found for: {}", port.getPath());
+        } catch (NoSuchPortException | IOException e) {
+            logger.debug("No SerialPortIdentifier found for: {}", portPathAsString);
```
wborn, line 64: The exception causing this may contain useful info when debug logging is enabled:

```suggestion
            logger.debug("No SerialPortIdentifier found for: {}", portPathAsString, e);
```

bundles/org.openhab.core.io.transport.serial.rxtx/src/main/java/org/openhab/core/io/transport/serial/internal/RxTxPortProvider.java
wborn: Thanks! Hopefully there is a nrjavaserial release soon so we can remove the workaround again.

bundles/org.openhab.core.io.transport.serial.rxtx/src/main/java/org/openhab/core/io/transport/serial/internal/RxTxPortProvider.java
dalgwen: Done. Sorry for the delay.
I also did a rebase /squash and so force push.

wborn:
I gave this a test but couldn't get it working. Only after restarting `org.openhab.core.config.discovery.usbserial.linuxsysfs` it lists the new port identifiers for me.

The issue is that it does notify listeners of existing scan results which it should do according to the JavaDocs:

```java
    /**
     * Registers an {@link UsbSerialDiscoveryListener} that is then notified about discovered serial ports and USB
     * devices.
     */
    void registerDiscoveryListener(UsbSerialDiscoveryListener listener);
```

It works better if I add some logic for this in `PollingUsbSerialScanner.registerDiscoveryListener` like: 

```java
    @Override
    public void registerDiscoveryListener(UsbSerialDiscoveryListener listener) {
        discoveryListeners.add(listener);
        for (UsbSerialDeviceInformation deviceInfo:  deltaUsbSerialScanner.getLastScanResult()) {
            listener.usbSerialDeviceDiscovered(deviceInfo);
        }
    }
```

It's probably also an issue in the `Ser2NetUsbSerialDiscovery`.

dalgwen:
Thank you for the test !
It happens when the SerialConfigOptionProvider is registered within the PollingUsbSerialScanner AFTER the first scan.
As it is a delta scan, all futures scan will NOT acknowledge the already discovered option and thus will not notify the SerialConfigOptionProvider of the port.

I think I didn't saw it earlier because during my test, I put some breakpoint. And the breakpoint allowed the  SerialConfigOptionProvider to be registered before the scan...

I added your suggestion to the pull request (and I rebased it)

dalgwen:
> It's probably also an issue in the `Ser2NetUsbSerialDiscovery`.

Indeed, the Ser2NetUsbSerialDiscovery is also a diff scan.
Do you want me to push the same fix to this pull request ?

wborn:
> Do you want me to push the same fix to this pull request ?

That would be nice too!

wborn (approved):
Thanks for the enhancement and fixes! :+1:

PR #3292, opened 2023-01-02
Introduce a profile for the generic SCRIPT transformation

J-N-K:
This is a replacement for #2872 which for unknown reasons can't be re-opened. I did another check and found that a general TRANSFORM profile is not feasible without a nearly complete refactoring of the whole transformation infrastructure. It would also need extensive refactoring in the UI part which I can't do.

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/pom.xml
+++ b/bundles/org.openhab.core.automation.module.script/pom.xml
@@ -25,6 +25,12 @@
       <artifactId>org.openhab.core.transform</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.openhab.core.bundles</groupId>
+      <artifactId>org.openhab.core.test</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
 </project>
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptTransformationService.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptTransformationService.java
@@ -155,7 +155,7 @@ public class ScriptTransformationService
             try {
                 CompiledScript compiledScript = scriptRecord.compiledScript;
 
-               ScriptEngine engine = compiledScript != null ? compiledScript.getEngine()
+                ScriptEngine engine = compiledScript != null ? compiledScript.getEngine()
                         : scriptEngineContainer.getScriptEngine();
                 ScriptContext executionContext = engine.getContext();
                 executionContext.setAttribute("input", source, ScriptContext.ENGINE_SCOPE);
@@ -176,7 +176,8 @@ public class ScriptTransformationService
 
                 // compile the script here _after_ setting context attributes, so that the script engine
                 // can bind the attributes as variables during compilation. This primarily affects jruby.
-                if (compiledScript == null && scriptEngineContainer.getScriptEngine() instanceof Compilable scriptEngine) {
+                if (compiledScript == null
+                        && scriptEngineContainer.getScriptEngine()instanceof Compilable scriptEngine) {
                     // no compiled script available but compiling is supported
                     compiledScript = scriptEngine.compile(scriptRecord.script);
                     scriptRecord.compiledScript = compiledScript;
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptTransformationService.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptTransformationService.java
@@ -12,6 +12,10 @@
  */
 package org.openhab.core.automation.module.script;
 
+import java.net.URI;
+import java.util.Collection;
+import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ScheduledExecutorService;
@@ -20,6 +24,7 @@ import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import javax.script.Compilable;
 import javax.script.CompiledScript;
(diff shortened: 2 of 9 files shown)


wborn:
Can you have another look at this @kaikreuzer?
It's the last piece for reaching feature parity regarding scripting in OH4 due to the Nashorn removal (https://github.com/openhab/openhab-distro/issues/1316).

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptTransformationService.java
```diff
@@ -132,10 +148,10 @@ public void deactivate() {
             try {
                 CompiledScript compiledScript = scriptRecord.compiledScript;
 
-                if (compiledScript == null && scriptEngineContainer.getScriptEngine() instanceof Compilable) {
+                if (compiledScript == null
+                        && scriptEngineContainer.getScriptEngine()instanceof Compilable scriptCompiler) {
```
kaikreuzer, line 152:
```suggestion
                        && scriptEngineContainer.getScriptEngine() instanceof Compilable scriptCompiler) {
```

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/profile/ScriptProfile.java
```diff
@@ -0,0 +1,164 @@
...
+
+        this.scriptLanguage = ConfigParser.valueAsOrElse(profileContext.getConfiguration().get(CONFIG_SCRIPT_LANGUAGE),
+                String.class, "");
+        this.toItemScript = ConfigParser.valueAsOrElse(profileContext.getConfiguration().get(CONFIG_TO_ITEM_SCRIPT),
+                String.class, "");
+        this.toHandlerScript = ConfigParser
+                .valueAsOrElse(profileContext.getConfiguration().get(CONFIG_TO_HANDLER_SCRIPT), String.class, "");
+
+        if (scriptLanguage.isBlank()) {
+            logger.error("Script language is not defined. Profile will discard all states and commands.");
```
kaikreuzer, line 80: `warn` should do here

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/profile/ScriptProfile.java
```diff
@@ -0,0 +1,164 @@
...
+            isConfigured = false;
+            return;
+        }
+
+        if (toItemScript.isBlank() && toHandlerScript.isBlank()) {
+            logger.error(
```
kaikreuzer, line 86: `warn` should do here

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/profile/ScriptProfile.java
```diff
@@ -0,0 +1,164 @@
...
+        }
+    }
+
+    private @Nullable String executeScript(String script, Type input) {
+        if (!script.isBlank()) {
+            try {
+                return transformationService.transform(scriptLanguage + ":" + script, input.toFullString());
+            } catch (TransformationException e) {
+                if (e.getCause() instanceof ScriptException) {
+                    logger.warn("Failed to process script '{}': {}", script, e.getCause().getMessage());
```
kaikreuzer, line 155: this might rather be an `error`...

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/profile/ScriptProfile.java
```diff
@@ -0,0 +1,164 @@
...
+                } else {
+                    logger.warn("Failed to process script '{}': {}", script, e.getMessage());
```
kaikreuzer, line 157: this might rather be an `error`...

kaikreuzer (changes requested):
Thanks, looks pretty good to me.
Could you please rebase the branch in order to resolve the conflict?
Just a few small comments below for your consideration.

bundles/org.openhab.core.automation.module.script/pom.xml
```diff
@@ -25,6 +25,12 @@
       <artifactId>org.openhab.core.transform</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
```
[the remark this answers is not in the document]
J-N-K: It does only for scope "test" because it uses the asserts for log messages which are defined in that bundle.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/profile/ScriptProfile.java
J-N-K: I used `error` because that is also used in the other profiles:

https://github.com/openhab/openhab-core/blob/cb38d1936006798d4eb0f71651bd1cb7c34bd158/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/profiles/SystemOffsetProfile.java#L53-L74

kaikreuzer: ok

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptTransformationService.java
J-N-K: This is the result of spotless, if I change it, `spotless:apply` removes the space again.

kaikreuzer: Indeed, I was also just looking at the build failure and saw the same... Weird, but ok, let's leave it this way then.

kaikreuzer (approved):
Thanks!

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptTransformationService.java
J-N-K: I'll check if this is fixed in newer versions of spotless, but that's another PR.

PR #3298, opened 2023-01-03
Add method to get namespaces to MetadataRegistry and ItemResource

stefan-hoehn:
This adds a new method to ItemResource to retrieve all namespaces of an item via the path "/{itemname: [a-zA-Z_0-9]+}/metadata/namespaces" which will eventually allow us to show even the custom namespaces in the UI (the PR for that is also already in the making).
It also extends the Item-Registry to provide a getAllNamespaces()-method which could also be exposed via other languages (e.g. jsscripting)

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
@@ -330,8 +330,8 @@ public class ItemResource implements RESTResource {
     public Response getBinaryItemState(@HeaderParam("Accept") @Nullable String mediaType,
             @PathParam("itemname") @Parameter(description = "item name") String itemname) {
         List<String> acceptedMediaTypes = Arrays.stream(Objects.requireNonNullElse(mediaType, "").split(","))
-                .peek(String::trim).collect(Collectors.toList());
-        // get item
+                .map(String::trim).collect(Collectors.toList());
+
         Item item = getItem(itemname);
 
         // if it exists
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
@@ -226,6 +226,34 @@ public class ItemResource implements RESTResource {
         return Response.ok(new Stream2JSONInputStream(itemStream)).build();
     }
 
+    /**
+     *
+     * @param itemname name of the item
+     * @return the namesspace of that item
+     */
+    @GET
+    @RolesAllowed({ Role.USER, Role.ADMIN })
+    @Path("/{itemname: [a-zA-Z_0-9]+}/metadata/namespaces")
+    @Produces(MediaType.TEXT_PLAIN)
+    @Operation(operationId = "getItemNamespaces", summary = "Gets the namespace of an item.", responses = {
+            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = String.class))),
+            @ApiResponse(responseCode = "404", description = "Item not found") })
+    public Response getItemNamespaces(@PathParam("itemname") @Parameter(description = "item name") String itemname,
+            @HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) @Parameter(description = "language") @Nullable String language,
+            @QueryParam("exclude") @Parameter(description = "exclude the given namespaces") @Nullable String excludeNamespaces) {
+        Item item = getItem(itemname);
+        final Locale locale = localeService.getLocale(language);
+        final Set<String> excludedNamespaces = splitAndFilterNamespaces(excludeNamespaces, locale);
+
+        if (item != null) {
+            Set<String> allItemMetadataNamespaces = metadataRegistry.getAllNamespaces(itemname).stream()
+                    .filter(n -> !excludedNamespaces.contains(n)).collect(Collectors.toSet());
+            return Response.ok(allItemMetadataNamespaces.toString()).build();
+        } else {
+            return getItemNotFoundResponse(itemname);
+        }
+    }
+
     @GET
     @RolesAllowed({ Role.USER, Role.ADMIN })
     @Path("/{itemname: [a-zA-Z_0-9]+}")
@@ -281,36 +309,6 @@ public class ItemResource implements RESTResource {
             // we cannot use JSONResponse.createResponse() bc. MediaType.TEXT_PLAIN
             // return JSONResponse.createResponse(Status.OK, item.getState().toString(), null);
             return Response.ok(item.getState().toFullString()).build();
-        } else {
-            return getItemNotFoundResponse(itemname);
-        }
-    }
-
-    /**
-     *
-     * @param itemname
-     * @return
-     */
-    @GET
-    @RolesAllowed({ Role.USER, Role.ADMIN })
-    @Path("/{itemname: [a-zA-Z_0-9]+}/metadata/namespaces")
-    @Produces(MediaType.TEXT_PLAIN)
-    @Operation(operationId = "getItemNamespaces", summary = "Gets the namespace of an item.", responses = {
-            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = String.class))),
-            @ApiResponse(responseCode = "404", description = "Item not found") })
-    public Response getItemNamespaces(@PathParam("itemname") @Parameter(description = "item name") String itemname,
-            @HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) @Parameter(description = "language") @Nullable String language,
-            @QueryParam("metadata") @Parameter(description = "metadata selector - a comma separated list or a regular expression (suppressed if no value given)") @Nullable String namespaceSelector) {
-        Item item = getItem(itemname);
-        final Locale locale = localeService.getLocale(language);
-        final Set<String> excludedNamespaces = splitAndFilterNamespaces(namespaceSelector, locale);
-
-        if (item != null) {
-            Set<String> allItemMetadataNamespaces = metadataRegistry.getAll().stream()
-                    .filter(n -> !excludedNamespaces.contains(n.getUID().getNamespace()))
-                    .filter(n -> n.getUID().getItemName().equals(itemname)).map(n -> n.getUID().getNamespace())
-                    .collect(Collectors.toSet());
-            return Response.ok(allItemMetadataNamespaces.toString()).build();
         } else {
             return getItemNotFoundResponse(itemname);
         }
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
@@ -293,6 +293,36 @@ public class ItemResource implements RESTResource {
      */
     @GET
     @RolesAllowed({ Role.USER, Role.ADMIN })
+    @Path("/{itemname: [a-zA-Z_0-9]+}/metadata/namespaces")
+    @Produces(MediaType.TEXT_PLAIN)
+    @Operation(operationId = "getItemNamespaces", summary = "Gets the namespace of an item.", responses = {
+            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = String.class))),
+            @ApiResponse(responseCode = "404", description = "Item not found") })
+    public Response getItemNamespaces(@PathParam("itemname") @Parameter(description = "item name") String itemname,
+            @HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) @Parameter(description = "language") @Nullable String language,
+            @QueryParam("metadata") @Parameter(description = "metadata selector - a comma separated list or a regular expression (suppressed if no value given)") @Nullable String namespaceSelector) {
+        Item item = getItem(itemname);
+        final Locale locale = localeService.getLocale(language);
+        final Set<String> excludedNamespaces = splitAndFilterNamespaces(namespaceSelector, locale);
+
+        if (item != null) {
+            Set<String> allItemMetadataNamespaces = metadataRegistry.getAll().stream()
+                    .filter(n -> !excludedNamespaces.contains(n.getUID().getNamespace()))
+                    .filter(n -> n.getUID().getItemName().equals(itemname)).map(n -> n.getUID().getNamespace())
+                    .collect(Collectors.toSet());
+            return Response.ok(allItemMetadataNamespaces.toString()).build();
+        } else {
+            return getItemNotFoundResponse(itemname);
+        }
+    }
+
+    /**
+     *
+     * @param itemname
+     * @return
+     */
+    @GET
+    @RolesAllowed({ Role.USER, Role.ADMIN })
     @Path("/{itemname: [a-zA-Z_0-9]+}/state")
     @Operation(operationId = "getItemState", summary = "Gets the state of an item.", responses = {
             @ApiResponse(responseCode = "200", description = "OK"),
@@ -831,7 +861,7 @@ public class ItemResource implements RESTResource {
     /**
      * helper: Response to be sent to client if a Thing cannot be found
      *
-     * @param thingUID
+     * @param itemname
      * @return Response configured for 'item not found'
      */
     private static Response getItemNotFoundResponse(String itemname) {
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
@@ -241,14 +241,14 @@ public class ItemResource implements RESTResource {
     public Response getItemNamespaces(@PathParam("itemname") @Parameter(description = "item name") String itemname,
             @HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) @Parameter(description = "language") @Nullable String language,
             @QueryParam("exclude") @Parameter(description = "exclude the given namespaces") @Nullable String excludeNamespaces) {
-        Item item = getItem(itemname);
+        final Item item = getItem(itemname);
         final Locale locale = localeService.getLocale(language);
         final Set<String> excludedNamespaces = splitAndFilterNamespaces(excludeNamespaces, locale);
 
         if (item != null) {
-            Set<String> allItemMetadataNamespaces = metadataRegistry.getAllNamespaces(itemname).stream()
-                    .filter(n -> !excludedNamespaces.contains(n)).collect(Collectors.toSet());
-            return Response.ok(allItemMetadataNamespaces.toString()).build();
+            final String allItemMetadataNamespaces = metadataRegistry.getAllNamespaces(itemname).stream()
+                    .filter(n -> !excludedNamespaces.contains(n)).collect(Collectors.joining(","));
+            return Response.ok(allItemMetadataNamespaces).build();
         } else {
             return getItemNotFoundResponse(itemname);
         }
@@ -290,8 +290,8 @@ public class ItemResource implements RESTResource {
 
     /**
      *
-     * @param itemname
-     * @return
+     * @param itemname item name to get the state from
+     * @return the state of the item as mime-type text/plain
      */
     @GET
     @RolesAllowed({ Role.USER, Role.ADMIN })
@@ -316,8 +316,8 @@ public class ItemResource implements RESTResource {
 
     /**
      *
-     * @param itemname
-     * @return
+     * @param itemname the item from which to get the binary state
+     * @return the binary state of the item
      */
     @GET
     @RolesAllowed({ Role.USER, Role.ADMIN })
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/MetadataRegistryImpl.java
```diff
@@ -61,6 +64,18 @@ public boolean isInternalNamespace(String namespace) {
...
+    /**
+     * Provides all namespaces of a particular item
+     *
+     * @param itemname the name of the item for which the namespaces should be searched.
+     */
+    @Override
+    public Set<String> getAllNamespaces(String itemname) {
+        return getAll().stream().filter(n -> {
+            return n.getUID().getItemName().equals(itemname);
+        }).map(n -> n.getUID().getNamespace()).collect(Collectors.toSet());
```
Hilbrand, line 76:
```suggestion
        return getAll().stream().filter(n -> n.getUID().getItemName().equals(itemname))
            .map(n -> n.getUID().getNamespace()).collect(Collectors.toSet());
```

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
```diff
@@ -226,6 +226,34 @@ public Response getItems(final @Context UriInfo uriInfo, final @Context HttpHead
...
+    @RolesAllowed({ Role.USER, Role.ADMIN })
+    @Path("/{itemname: [a-zA-Z_0-9]+}/metadata/namespaces")
+    @Produces(MediaType.TEXT_PLAIN)
+    @Operation(operationId = "getItemNamespaces", summary = "Gets the namespace of an item.", responses = {
+            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = String.class))),
+            @ApiResponse(responseCode = "404", description = "Item not found") })
+    public Response getItemNamespaces(@PathParam("itemname") @Parameter(description = "item name") String itemname,
+            @HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) @Parameter(description = "language") @Nullable String language,
+            @QueryParam("exclude") @Parameter(description = "exclude the given namespaces") @Nullable String excludeNamespaces) {
+        Item item = getItem(itemname);
```
Hilbrand, line 244:
```suggestion
        final Item item = getItem(itemname);
```

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
```diff
@@ -226,6 +226,34 @@ public Response getItems(final @Context UriInfo uriInfo, final @Context HttpHead
...
+        final Locale locale = localeService.getLocale(language);
+        final Set<String> excludedNamespaces = splitAndFilterNamespaces(excludeNamespaces, locale);
+
+        if (item != null) {
+            Set<String> allItemMetadataNamespaces = metadataRegistry.getAllNamespaces(itemname).stream()
+                    .filter(n -> !excludedNamespaces.contains(n)).collect(Collectors.toSet());
+            return Response.ok(allItemMetadataNamespaces.toString()).build();
```
Hilbrand, line 251: Maybe use `Collectors.joining(",")` to put them directly in a comma separate string:
```suggestion
            final String allItemMetadataNamespaces = metadataRegistry.getAllNamespaces(itemname).stream()
                    .filter(n -> !excludedNamespaces.contains(n)).collect(Collectors.joining(","));

            return Response.ok(allItemMetadataNamespaces).build();
```

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/MetadataRegistryImpl.java
J-N-K, line 76:
```suggestion
    public Collection<String> getAllNamespaces(String itemname) {
        return stream().map(Metadata::getUID).filter(key -> key.getItemName().equals(itemname)).map(MetadataKey::getNamespace).toList();
```

You can use `stream()` from `AbstractRegistry` and by mapping to the `UID` (a `MetadataKey`) first also shorten the further processing.

bundles/org.openhab.core/src/main/java/org/openhab/core/items/MetadataRegistry.java
```diff
@@ -36,10 +38,17 @@ public interface MetadataRegistry extends Registry<Metadata, MetadataKey> {
      */
     boolean isInternalNamespace(String namespace);
 
+    /**
+     * Provides all namespaces of a particular item
+     *
+     * @param itemname the name of the item for which the namespaces should be searched.
+     */
+    public Set<String> getAllNamespaces(String itemname);
```
J-N-K, line 46:
```suggestion
    public Collection<String> getAllNamespaces(String itemname);
```

I don't think there is a need to be more specific.

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
```diff
@@ -831,7 +859,7 @@ private JsonObject buildStatusObject(String itemName, String status, @Nullable S
     /**
      * helper: Response to be sent to client if a Thing cannot be found
      *
-     * @param thingUID
+     * @param itemname
```
J-N-K, line 862: Nice catch. Can you also adjust `a Thing` two lines above to `an item`?

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
J-N-K: I agree with @Hilbrand that you use either `Collectors.joining(",")` or `String.join(",", collection)` and not use `.toString` here. But also see my general comment.

J-N-K (changes requested):
Thanks for your contribution. I have two general comments:

- What would be the use-case of excluding namespaces? And isn't that something the consumer can do (if he needs to provide the list of exclusions anyway)?
- Why is the list "text/plain" and not a JSON? This is most likely easier to process for clients (also if they need to take care of the filter).

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/MetadataRegistryImpl.java
stefan-hoehn: good catch, I had the "return" there during development and forgot to take it out.

stefan-hoehn: The reason only becomes obvious if you see the implementation of the UI. The way at least how the UI works currently is that it "knows" what the standard metadata-Namespaces are. Basically what it does, that it explicitly asks for the namespace it knows because that is what the API requires (this API here does not deliver any metadata, if you not explicitly ask for each individual namespace). 
What I basically did in the UI now, is that I ask for all names and exclude the ones that the UI already knows and then add them to the list the UI wants. Basically, and I agree here, will most likely result in the full set again. The only reason I did that was that I wasn't sure if @ghys only did this because it wasn't possible otherwise (which most likely is the case) or if he had something else in mind as well.

To summarize: I could easily take the exclude parameter out and ask for all namespaces. Worst case that could habe here that would have to do an exclusion in the frontend. So, let me know if I should delete the exclusion part. I would do so and simplify the UI to just take all namespaces it gets from the backend API.

While thinking about it, we could even go a bit further and make getItems() smart by providing the _metadata_-parameter with something like "*" which would internally grap all namespaces and deliver them automatically which would avoid one round-trip (setting _metadata_ to "" will return nothing). What about that idea?

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
stefan-hoehn: Will do, being the friendly boy scout now who cleans up the camp ;-) I went over all missing descriptions and added them.

there are more hints, though and one in particular that could be optimized:

In the following case String.trim() is actually ignored (this is what my IDE tells me). 
```
List<String> acceptedMediaTypes = Arrays.stream(Objects.requireNonNullElse(mediaType, "").split(","))
                .peek(String::trim).collect(Collectors.toList());
```             

and it also says that "collect(Collectors.toList())" can be simply changed to .toList()

Should I change that?

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/MetadataRegistryImpl.java
J-N-K: It's easy to filter a list on another list in js and also to concatenate them to a string:

```
let customNamespaces = allNamespaces.filter(ns => !knownNamespaces.includes(ns)).join(',')
```

and returning the list of namespaces as JSON array makes it much easier to re-use for other use-cases. So I would suggest to change to `@Produces(MediaType.APPLICATION_JSON)` and use `Response.ok(new Stream2JSONInputStream(namespaces.stream())).build();`

Regarding the change in `.getItems`: I never understood why it was designed that way. IMO it would have been much better to return everything by default and allow to limit it with the parameter than the other way round. @ghys, @lolodomo Would it hurt for UI to change that?

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
J-N-K: The IDE is fully correct. `.peek` should be a `.map`, please don't ask my why I did such a silly thing :-). Using `.toList` is always preferred in Java 17.

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/MetadataRegistryImpl.java
```diff
@@ -61,6 +64,17 @@ public boolean isInternalNamespace(String namespace) {
...
+    /**
+     * Provides all namespaces of a particular item
+     *
+     * @param itemname the name of the item for which the namespaces should be searched.
+     */
+    @Override
+    public Set<String> getAllNamespaces(String itemname) {
+        return getAll().stream().filter(n -> n.getUID().getItemName().equals(itemname))
+                .map(n -> n.getUID().getNamespace()).collect(Collectors.toSet());
+    }
```
J-N-K, line 76: Did you see my comment regarding the mapping to the UID as first step?

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
stefan-hoehn: It is actually interesting. I changed it to 

List<String> acceptedMediaTypes = Arrays.stream(Objects.requireNonNullElse(mediaType, "").split(","))
                .map(String::trim).collect(Collectors.toList());

and the IDE says

'collect(toList())' can be replaced with 'toList()'

So, I even let the IDE do so:

image

which results into 

image


Cannot resolve method 'toList' in 'Stream'

J-N-K: Does it compile? Sometimes my IDE also does strange things...

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/MetadataRegistryImpl.java
stefan-hoehn: "I never understood why it was designed that way. IMO it would have been much better to return everything by default and allow to limit it with the parameter than the other way round."

At least it would break the current behaviour. That's why I didn't dare to do it. On the other hand, why would I want to deliver all the namespace information in getItems when most of the time it is not needed. Therefore I do understand the design.

Ok, let's summarise (as I want to get this thing done eventually):
- I remove the exclusion parameter
- I change the return parameter to be JSON
- I adapt the UI accordingly
- I keep the getItem as it is for now.

J-N-K: Perfectly summed up.

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/MetadataRegistryImpl.java
stefan-hoehn: No sorry, I got overwhelmed with both you providing recommendations for the same line

Interesting that these two are supposedly the same :-) and too be honest I would have never come up with your way of doing it (took me quite a minute to understand even) but finally makes sense to me.


````
        Set<String> a=  stream().
                filter(n -> n.getUID().getItemName().equals(itemname))
                .map(n -> n.getUID().getNamespace()).collect(Collectors.toSet());
        
        Set<String> b=  stream().map(Metadata::getUID).
                filter(key -> key.getItemName().equals(itemname)).
                map(MetadataKey::getNamespace).collect(Collectors.toSet());
````

File: bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/MetadataRegistryImpl.java (hunk 1 above)
Comment (ghys) [reply bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/MetadataRegistryImpl.java]: > I never understood why it was designed that way. IMO it would have been much better to return everything by default and allow to limit it with the parameter than the other way round.

Not sure, but maybe performance reasons? @kaikreuzer designed this originally and might have the answer. It would probably be wise to have this answer before making changes that are harmful.

> @ghys, @lolodomo Would it hurt for UI to change that?

Probably not, it would simply return more metadata than necessary.

Note that the idea I had over 2 years ago but never implemented (https://github.com/openhab/openhab-webui/issues/397#issuecomment-717130372) to solve this user namespaces listing issue without changing the API was to have a managed [MetadataConfigDescriptionProvider](https://next.openhab.org/javadoc/latest/org/openhab/core/config/core/metadata/metadataconfigdescriptionprovider) backed by UI components (like UI-designed sitemaps, though it can have its own custom storage/syntax), along with an editor in the UI's developer tools. Then MetadataConfigDescriptionProviders declared this way along with others e.g. coming from add-ons could appear in the list (the "enter custom namespace" option would still be there). When editing metadata for a described namespace the user would be presented with a nice standard configuration sheet with parameters, not only the a raw YAML editor.
I saw this recently on a forum thread too and it probably could make it to 4.0.

Comment (stefan-hoehn) [reply bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/MetadataRegistryImpl.java]: @ghys did you see my contribution here?
https://github.com/openhab/openhab-webui/pull/1623

File: bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java (hunk 5 above)
Comment (stefan-hoehn) [reply bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java]: Yes, it does, but to be honest, I don't want to commit something that is shown red in my IDE ;-)

**Comment by stefan-hoehn** [comment]
All of the review comments should be resolved by now. @Hilbrand @J-N-K maybe you have a final look onto it?

File: bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java (hunk 5 above)
Comment (J-N-K) [reply bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java]: Which IDE is it?From the colors it looks a bit like IntelliJ. Did you set the project JDK to 17?

**Review by J-N-K — APPROVED** [review]
LGTM now, thanks. I think that also all comments from @Hilbrand are addressed. And congratulation to your first core contribution :-)

**Comment by jlaur** [comment]
@stefan-hoehn - could you also have a look at the hue emulation in addons? This test class also implements `MetadataRegistry`:

bundles/org.openhab.io.hueemulation/src/test/java/org/openhab/io/hueemulation/internal/rest/mocks/DummyMetadataRegistry.java

Sorry for not providing more details, on my way to bed, just saw a full build failing because of this interface not being fully implemented.

**Comment by spacemanspiff2007** [comment]
Just for my understanding - what is the use case for requesting the item namespace without the namespace values?
Currently requesting `/items/{itemname}` with metadata selector `.+` will return all namespaces, too.

**Comment by florian-h05** [comment]
The `getAllNamespaces` method is really useful for scripting/automation like JS Scripting.

Regarding the REST endpoint: When creating this PR, we didn't know that you can use `.+` to get all namespaces from the items endpoint. Now that it is there, I think even if the UI does not need is, there might be some third parties that use it.

**Comment by spacemanspiff2007** [comment]
> Regarding the REST endpoint: When creating this PR, we didn't know that you can use `.+` to get all namespaces from the items endpoint. Now that it is there, I think even if the UI does not need is, there might be some third parties that use it.

Wouldn't it be better to update the docs of that endpoint than to create another endpoint with duplicate functionality?

Is the `getAllNamespaces` also exposed through the RestAPI?

**Comment by florian-h05** [comment]
`getAllNamespaces` is required for scripting. 

It is exposed through the REST API as well. If the UI does not need it because you can get all metadata for an Item with your selector, we don’t need that REST endpoint.

Note that getAllNameapaces is not duplicate to the Item REST endpoint: getAllNamespaces and it’s endpoint return a Collection of the names of the metadata on that Item.

**Comment by spacemanspiff2007** [comment]
That was a misunderstanding on my side. I thought you added an endpoint to the RestAPI that returns a global overview of all used metadata namespaces over all items. I see now that's not the case.

I still think the `/items/{itemname}/metadata/namespaces` is not needed and making `/items/{itemname}`more user friendly e.g. with a flag `allMetadata`=[true|false] would be more elegant

**Comment by florian-h05** [comment]
If it is possible to get all metadata from the `/items/{itemname}` endpoint there is no need for a `/items/{itemname}/metadata/namespaces`.

IMO the `/items/{itemname}` endpoint should accept a `*` for the `metadata` parameter, but if it accepts a `.+` also makes it possible to get all metadata, we just need to extend the docs and we can remove the namespaces endpoint.

**Comment by rkoshak** [comment]
> endpoint should accept a * for the metadata parameter, but if it accepts a .+ also makes it possible to get all metadata, we just need to extend the docs and we can remove the namespaces endpoint.

It's a regex. I'd use `.*` but `.+` would be equivalent. A `*` or `+` by itself wouldn't work because that's not a meaningful regex.

**Comment by spacemanspiff2007** [comment]
`.+` is a regular expression which matches everything so this will return all metadata for the item.
I'm not super happy with that solution but since I got some regex knowledge it was not a problem, I even raised the [initial issue](https://github.com/openhab/openhab-core/issues/1692) for OH3.

Imho it would be better if by default all metadata would be returned and the metadata selector can be used to reduce the metadata. I'll add an issue so this can be discussed there

**Comment by spacemanspiff2007** [comment]
> (performance reasons?, and less data to transmit)

Both should be neglectable. Let's wait for some response in the issue.

**Comment by stefan-hoehn** [comment]
First of all, thanks for pointing that out - neither was I aware of linked issue nor that a regex would work. 

I looked into the code and I indeed "getItemData" says "a comma separated list or **a regular expression**". Though I even overlooked it so much that I eveb proposed to provide a "*" for everything. What I am trying to say is that we should at least document that retrieving all names should be done by ".*"(I like that one better) very obviously because it makes it clear to everyone directly. I am not sure if the new API endpoint really hurts but if it does I need to find time to rewrite / take it out  in some of the upcoming weeks and then open another PR. Up to you to decide.

**Comment by spacemanspiff2007** [comment]
> I am not sure if the new API endpoint really hurts but if it does I need to find time to rewrite / take it out in some of the upcoming weeks and then open another PR.

It doesn't "hurt" however it's redundant and imho clutters the API.
I'd rather have the items endpoint make it more clear.

Why would I make two requests (first the metadata namespaces and then the item with the namespaces comma separated) when it can be done with one request. It's faster **and** easier with only one request.
So I would prefer making the items endpoint more user friendly (and not have this endpoint) and this is why I created the follow up issue.


> Up to you to decide.

I don't have any say in this. I'm merely opening issues and commenting on them. :wink:

**Comment by stefan-hoehn** [comment]
One thing that came to my mind during implementation of the Blockly blocks for the metadata (I have not committed/pushed them yet because they highly depend on the jsscripting blcokly PR that is not yet reviewed but they blockls are already done) is that it would add the possibility to show the namespaces as a lookup dropdown as part of the block, similar like the ones when creating an item, thing or play sinks. Of course I could definitely retrieve the whole item for that and extract the namespaces from that but it is actually much lighter if I can just retrieve namespaces for that item. This is not intended to justify the fact I implemented this endpoint originally but it would just be nice now to keep it for that purpose.

**Comment by stefan-hoehn** [comment]
We are talking about two complete different usecases here:
1) When showing the items in the UI I already dropped the namespaces query and replaced it with .*
2) The other usecase is in Blockly where the users choses a block and selects the namespace within the block. There I (as the developer of that block!) want to implement a dropdown that allows the blockly user to chose the right namespace. It is much easier in the background to just query the namespaces which is the only thing I need here for the dropdown than querying the whole item and extracting the information. As I said, still doable but more complicated.

**Comment by J-N-K** [comment]
I agree with @stefan-hoehn here. Very often you don't need the information for all items, but only for one. Then it is a single request anyway and there is no benefit in getting the whole item including full metadata.

**Comment by spacemanspiff2007** [comment]
It's a two-liner, it's not some scientific calculation that takes long:
```python
metadata = item_response.get('metadata', {})
namespaces = list(metadata.keys())
```

Rest APIs tend to always over-fetch a little and it's not really a problem (it's not GraphQL). 
Parsing and traversing json is super fast, on every client so that's not really an issue, too.


> but you'll put higher processing burden on the client

The other way you'll put higher processing burden on the server, that's not really an argument. :wink: 
Every smartphone has more processing power than an e.g. raspberry pi.

---

Introducing a new endpoint that somebody has to maintain forever vs. a one or two liner on the client side is and should be always a no brainer.

PR #3299, opened 2023-01-03
rework GenericEventTrigger and GenericEventCondition

### LINKED ISSUE #3234: RFC: Implement event topic filtering directly in EventHandler
(summary re-written):

I've investigated this issue quite a bit at this point. Right now, a large number of `EventSubscriber`s will set the `event.topics` OSGi service property to the topic they expect to receive, but `OSGiEventManager` (who receives notice of the service registration) does absolutely nothing with it, and `EventHandler` only filters events based on the `EventSubscriber.getSubscribedEventTypes()` method, filtering against the event's type. The `EventSubscriber`s will then also implement `EventFilter`, and in their filter _actually_ check the event topic (or not... a lot of them simply rely on the event type). Confounding this, there's `GenericEventTriggerHandler` that has configuration for `eventSource`, `eventTopic`, and `eventType`, but `eventTopic` is ignored besides to set the service property (which is then ignored), and unintuitively filters the incoming event's _topic_ against the configured _source_. The docs at https://www.openhab.org/docs/configuration/jsr223.html#triggertype-objects-all-jsr223-languages are incorrect, that the event topic _cannot_ be filtered, and also claim that `eventSource` can be used for filtering based on the subject of the event. This is only (partially) true because `GenericEventTriggerHandler` is checking against a different property, as mentioned above. And it's buggy because it's simply doing a `String.contains` check, so if you configured it look for `MySwitchItem`, it will still pass through events for `MySwitchItem1`. `Event.getSource()` is actually intended to carry the name of the component that created the event, not the subject of the event. Adding additional confusion: there's also a `GenericEventConditionHandler`, that _does_ compare topic-to-topic, source-to-source, event-type-to-event-type (and payload-to-payload). But it's doing a regex match (except some special processing for payload).

Sooo... my proposal:
 * Add implementation to `OSGiEventManager` to capture the `event.topics` service param, and pass it to `EventHandler` (if present), and `EventHandler` will apply that _before_ checking for an `EventFilter` on the `EventSubscriber`, with the intention that the vast majority of `EventSubscriber`s no longer even need to implement an `EventFilter`.
 * The filter described above should be implemented as a filesystem glob syntax ( https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/file/FileSystem.html#getPathMatcher(java.lang.string) ). It's not as complicated as a full-on regex, but allows for single-level matching which is what most of the currently defined event.topics expect. A few will still need to be tweaked.
 * Remove `EventFilter` from almost all `EventSubscriber`s, simply relying on the now-built-in topic processing.
 * `GenericEventTriggerHandler` will simply forward its `eventTopic` param as the service property `events.topic`, as it already does. It will still implement an `EventFilter` to apply the `eventSource` filter, but that will filter against the event's source, exact match. `eventType` still works as before - exact match. This is a _breaking change_.
 * `GenericEventConditionHandler` will be updated to process the topic, source, and type the same as `EventHandler` and `GenericEventTriggerHandler` do. Again a _breaking change_.

Hopefully the impact of the breaking changes will be relatively small - most rules already rely on higher level triggers and conditions such as ItemStateChangeTrigger. MainUI doesn't expose GenericEventTrigger or GenericEventCondition. The vast majority of existing users are already setting up their topic and event types as if it works this way already (with some exceptions because of the vagueness of how the topic filtering (doesn't!) work).

@openhab/core-maintainers WDYT?

ccutrer:
to make their interfaces and semantics match, as well as having a well defined (and useful) way of defining topic filters

fixes #3234

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/GenericEventConditionHandler.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/GenericEventConditionHandler.java
@@ -54,18 +54,18 @@ public class GenericEventConditionHandler extends BaseConditionModuleHandler {
 
         this.source = (String) module.getConfiguration().get(CFG_SOURCE);
         String topic = (String) module.getConfiguration().get(CFG_TOPIC);
-        if (!topic.isEmpty()) {
+        if (!topic.isBlank()) {
             topicFilter = new TopicGlobEventFilter(topic);
         } else {
             topicFilter = null;
         }
-        if (module.getConfiguration().get(CFG_TOPIC) != null) {
-            this.types = Set.of(((String) module.getConfiguration().get(CFG_TOPIC)).split(","));
+        if (module.getConfiguration().get(CFG_TYPES) != null) {
+            this.types = Set.of(((String) module.getConfiguration().get(CFG_TYPES)).split(","));
         } else {
             this.types = Set.of();
         }
         String payload = (String) module.getConfiguration().get(CFG_PAYLOAD);
-        if (!payload.isEmpty()) {
+        if (!payload.isBlank()) {
             payloadPattern = Pattern.compile(payload);
         } else {
             payloadPattern = null;
@@ -74,7 +74,7 @@ public class GenericEventConditionHandler extends BaseConditionModuleHandler {
 
     @Override
     public boolean isSatisfied(Map<String, Object> inputs) {
-        Event event = inputs.get("event") != null ? (Event) inputs.get("event") : null;
+        Event event = inputs.get("event") instanceof Event ? (Event) inputs.get("event") : null;
         if (event == null) {
             return false;
         }
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/GenericEventConditionHandler.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/GenericEventConditionHandler.java
@@ -13,12 +13,15 @@
 package org.openhab.core.automation.internal.module.handler;
 
 import java.util.Map;
+import java.util.Set;
+import java.util.regex.Pattern;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.automation.Condition;
 import org.openhab.core.automation.handler.BaseConditionModuleHandler;
 import org.openhab.core.events.Event;
+import org.openhab.core.events.TopicGlobEventFilter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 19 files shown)


ccutrer:
This is very much a breaking change - filtering behavior and names of configuration items. I don't think GenerictEventTrigger and GenericEventCondition are used too much though, so the impact shouldn't be huge. I'm still open to walking back, or having a back-compat mode (depending on which configuration params are used) any particular breaking change.

In light of that pending discussion, I haven't ensured all the integration tests pass yet, and don't plan on solidifying them until the direction I'm going here is confirmed.

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/GenericEventConditionHandler.java
```diff
@@ -27,50 +30,69 @@
...
-                configValue = configValue.endsWith("*") ? configValue : configValue + ".*";
-            }
-            if (value != null) {
-                return value.matches(configValue);
-            } else {
-                return false;
-            }
+        this.source = (String) module.getConfiguration().get(CFG_SOURCE);
+        String topic = (String) module.getConfiguration().get(CFG_TOPIC);
+        if (!topic.isEmpty()) {
```
J-N-K, line 57: I would suggest to use `.isBlank` here, only whitespace characters don't make sense in the topic, too.

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/GenericEventConditionHandler.java
```diff
@@ -27,50 +30,69 @@
...
+            topicFilter = new TopicGlobEventFilter(topic);
+        } else {
+            topicFilter = null;
+        }
+        if (module.getConfiguration().get(CFG_TOPIC) != null) {
```
J-N-K, line 62: Shouldn't this be `CFG_TYPES`?

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/GenericEventConditionHandler.java
```diff
@@ -27,50 +30,69 @@
...
+            payloadPattern = Pattern.compile(payload);
         } else {
-            return true;
+            payloadPattern = null;
         }
     }
 
     @Override
     public boolean isSatisfied(Map<String, Object> inputs) {
         Event event = inputs.get("event") != null ? (Event) inputs.get("event") : null;
```
J-N-K, line 77:
```suggestion
        Event event = inputs.get("event") instanceof Event ? (Event) inputs.get("event") : null;
```

Just to make sure we don't run into a `ClassCastException` here.

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/GenericEventConditionHandler.java
```diff
@@ -27,50 +30,69 @@
...
-                    && isConfiguredAndMatches(PAYLOAD, event.getPayload())
-                    && isConfiguredAndMatches(EVENTTYPE, event.getType());
+        if (event == null) {
+            return false;
         }
-        return false;
+        if (!types.isEmpty() && !types.contains(event.getType())) {
+            return false;
+        }
+        TopicGlobEventFilter localTopicFilter = topicFilter;
```
J-N-K, line 84: I was think about that in your previous PR. Wouldn't it be better to have a `MatchAllEventFilter` that always returns `true`? That way we could omit a lot of null checks and also make `.getFilter` return a non-null value.

In general since `EventFilter` gets the `Event`, a filter could also filter not only on topic but also on source or type. That brings me again to the `EventFilterBuilder` that builds an `EventFilter` `.withTopicGlob(String glob)`, `.withTopicRegex(String regex)`, `.withType(Class<? extends Event>)`, `.withTypes(Set<Class<? extends Event>>` or `.withSource(String source)`. The implementation than collects all those filters in a `List<EventFilter> filterChain`, and does a `filterChain.stream().allMatch(filter -> filter.apply(event)` in `.apply`. But this is out of scope here. Just an idea :-)

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/GenericEventTriggerHandler.java
```diff
@@ -41,37 +43,51 @@
...
-        if (module.getConfiguration().get(CFG_EVENT_TYPES) != null) {
-            this.types = Set.of(((String) module.getConfiguration().get(CFG_EVENT_TYPES)).split(","));
+        this.source = (String) module.getConfiguration().get(CFG_SOURCE);
+        String topic = (String) module.getConfiguration().get(CFG_TOPIC);
+        if (!topic.isEmpty()) {
+            topicFilter = new TopicGlobEventFilter(topic);
+        } else {
+            topicFilter = null;
+        }
+        if (module.getConfiguration().get(CFG_TOPIC) != null) {
```
J-N-K, line 76: See above

J-N-K (changes requested):
I did not compare with the old implementation, just checked what I would expect. What exactly makes it incompatible with the old implementation (except that the topic filtering was not working at all for the trigger, which is clearly a bug)?

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/GenericEventConditionHandler.java
ccutrer: I actually really like this idea. But yeah... out of scope here.

ccutrer:
> I did not compare with the old implementation, just checked what I would expect. What exactly makes it incompatible with the old implementation (except that the topic filtering was not working at all for the trigger, which is clearly a bug)?

Behavior changes (whether it's a bug or not):
 * Config param names for GenericEventTrigger changed (from eventTopic => topic, eventSource => source, eventType => types
 * Config param names for GenericEventCondition changed (from eventType => types)
 * GenericEventTrigger topic param now enforced (as a glob)
 * GenericEventTrigger source param is now checked against event source, exact match, not against the event topic as a substring
 * GenericEventCondition topic param is now a glob, instead of a regex (though it was never documented that I could find that it was a regex)
 * GenericEventCondition source param is now an exact match, instead of a regex (again, never really documented)
 * GenericEventCondition payload param is now a simple regex, without implicit start/end anchors (using `find()` instead of `matches()`). There are also weirdness here because the old code checked if the payload config value started with `*` it would leave it alone, and then process as a regex, which then is a syntax error since regexes can't start with `*`.

I'm mostly concerned about the GenericEventTrigger topic now being enforced. Before switching some of the internal triggers to use the TopicPrefixEventFilter, they were attempting to set a filter like `openhab/*`, which wouldn't match an event like `openhab/items/myMotionItem1/command` anymore, since the glob would need to be `**` in order to match multiple levels. And itests definitely had several cases where the `source` param was being set to things like `myMotionItem1` which definitely won't match the event source anymore. Maybe I'm just being too conservative though, and these are things that didn't work at all in the past, so shouldn't really be considered breaking.

bundles/org.openhab.core/src/main/java/org/openhab/core/events/TopicGlobEventFilter.java
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * Constructs a new topic event filter.
+     *
+     * @param topicGlob the glob
+     * @see <a href=
+     *      "https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/file/FileSystem.html#getPathMatcher(java.lang.String)">Java
+     *      Glob</a>
+     */
+    public TopicGlobEventFilter(String topicGlob) {
+        this.topicMatcher = FileSystems.getDefault().getPathMatcher("glob:" + topicGlob);
```
ccutrer, line 44: I did verify that this still works on Windows, even though this is technically a file system interface, and file systems on windows use backslash not forward slash

J-N-K: Yes, it works because internally it is converted to an regex and the conversion code is the same for Windows and Unix. To make sure that this is always the case in the future, please add a test that checks it, we'll have Windows and Mac builds for CI, too.

J-N-K:
To me these changes make perfectly sense and I agree that most of this was probably never used. We need to add a breaking notice to openhab-distro.

I would like to hear comments from other @openhab/core-maintainers.

bundles/org.openhab.core/src/main/java/org/openhab/core/events/TopicGlobEventFilter.java
ccutrer: already there, as a unit test: TopicGlobEventFilterTest.

ccutrer:
Ping @openhab/core-maintainers

J-N-K (changes requested):
Sorry for letting you wait so long. One last point: can you re-generate the i18n-files? Otherwise LGTM.

J-N-K (approved):
Thanks

PR #3301, opened 2023-01-05
prevent a provider from removing an element from another provider

ccutrer:
if multiple providers provide the same element, the added callback simply ignores the duplicates. but the provider has no idea its element was ignored, so it may try to remove it in the future, and things will get all out of whack if the element is removed from the registry even if another provider thinks it's still there

this is easily reproducible by creating an item in the UI, then creating the same item via a .items file, and then removing the item from the .items file. The item will seem like it is gone to most of openHAB, but will come back after reboot when the managed provider re-adds.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/common/registry/AbstractRegistry.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/common/registry/AbstractRegistry.java
@@ -244,6 +244,14 @@ public abstract class AbstractRegistry<@NonNull E extends Identifiable<K>, @NonN
                         element.getClass().getSimpleName(), uid, provider.getClass().getSimpleName());
                 return;
             }
+            Provider<E> elementProvider = elementToProvider.get(existingElement);
+            if (!elementProvider.equals(provider)) {
+                logger.debug(
+                        "Cannot remove \"{}\" with key \"{}\" from provider \"{}\" because the instance in the registry is from provider \"{}\"!",
+                        element.getClass().getSimpleName(), uid, provider.getClass().getSimpleName(),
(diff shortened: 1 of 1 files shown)


jimtng:
why not prevent the provider from adding the duplicate in the first place?

ccutrer:
it already does. well, it ignores it. raising an exception would change the interface, and existing providers would have to be updated to handle that exception to make sure they don't persist objects that are duplicates. and then whatever other code calls add in the first place... it might end up being quite the cascade. I still think this commit is useful, since we can't guarantee that providers will not ever call remove() when add() didn't succeed.

bundles/org.openhab.core/src/main/java/org/openhab/core/common/registry/AbstractRegistry.java
```diff
@@ -244,6 +244,14 @@ public void removed(Provider<E> provider, E element) {
                         element.getClass().getSimpleName(), uid, provider.getClass().getSimpleName());
                 return;
             }
+            Provider<E> elementProvider = elementToProvider.get(existingElement);
+            if (!elementProvider.equals(provider)) {
+                logger.debug(
+                        "Cannot remove \"{}\" with key \"{}\" from provider \"{}\" because the instance in the registry is from provider \"{}\"!",
```
J-N-K, line 250:
```suggestion
                logger.error(
                        "Provider '{}' is not allowed to remove element '{}' with key '{}' from the registry because it was added by provider '{}'.",
```

and adjust the order of objects.

Otherwise LGTM.

J-N-K (changes requested):

J-N-K (approved):
Thanks!

Issue #3289, opened 2022-12-31
Centralized Crypto Library 


morph166955:
I'll preface this with, im volunteering to write this myself as it probably has a decent work effort involved and I just did a bunch of this recently for the ShieldTV binding. I am however looking for input on how best to structure it. 

Over the past few weeks of working on the ShieldTV binding I've encountered a few decent learning curves. The biggest was working with crypto functions (specifically PKI, not transport).  Several bindings need to create, store, and interact with devices using specific keys and certificates. Unfortunately, we don't have a centralized set of functions to help the developers/maintainers easily perform these functions.  This came up during a separate discussion I was having with @kaikreuzer and he suggested opening an issue here for a larger discussion. 

I propose a new addition to the core (looking for suggestions as to where it's most appropriate).  Forgive the bad pseudocode but I would envision it working something like this:

CryptoStore myCrypto = new CryptoStore(); //add a variety of different options to accept keys, filenames, etc at creation

myCrypto.generateKey(Int keyLength, String algorithm); // Generate a private key of requested length and with stated algorithm 
myCrypto.generateSelfSignedCert(String name); //use the private key and the certificate name/data to create a self signed certificate 
myCrypto.loadFromFile(String fileName);
myCrypto.writeToFile(String fileName);
myCrypto.getKeyStore() //Return a KeyStore that's properly initialized and populated.
myCrypto.setKey(String myKey); //Store a base64 encoded private key 
myCrypto.setCert(String myCert); //Store a base64 encoded certificate 

Keys and certs would be held as base64 encoded strings when not directly in the KeyStore format.  There's probably a dozen functions I'm not listing, as well as variants of those above which need to be mapped out (e.g. to return the private key as a Key instead of a String if that's what's needed).  By doing it this way you also can have multiple sets of keys being used without adding confusion. 

Assuming this gets accepted/merged, we would then need to petition the developers to make the necessary changes to remove their crypto bits and adopt this as a standard. Documentation will be obviously necessary. I'd volunteer to adapt the ShieldTV first so it can be used as an example to others. 

Thoughts?

clinique:
Very good approach and this is from my point of view the aim of core to centralize common topics.

J-N-K:
Would you suggest a keystore that is managed by each binding itself or a central keystore that is injected into (what ever service needs it)? I would say that it would better be managed centrally, that way we can even migrate something if we need it. 

[...]

PR #3302, opened 2023-01-07
[WIP] Centralized Crypto Library

morph166955:
Implements (#3289)

This is still a work in progress.  

Creates the basis for a centralized crypto storage mechanism.  Also creates several functions that are used across addon's to simplify addon creation.  

Basic usage expectation:

Key cryptoStoreKey = CryptoStore.generateEncryptionKey();
CryptoStore myCryptoStore = new CryptoStore();

cryptoStoreKey is an AES encryption key used to encrypt the private key while in memory.  This must be passed to any function which handles the private key.

Option 1) User has private key and certificate to provide (e.g. something provided by a device)
myCryptoStore.setPrivKey(String myPrivKey,cryptoStoreKey);
myCryptoStore.setCert(String cert); OR .setCert(Certificate cert);

Option 2) User needs a new keypair created.
myCryptoStore.generateNewKeyPair(Key key);

Option 3) User already has a java KeyStore stored on disk and imports it.
myCryptoStore.loadFromKeyStore(String keystoreFileName, String keystorePassword, Key key);

User can request a KeyStore via getKeyStore(String keystorePassword, Key key);

User can save the CryptoStore to a java KeyStore for persistence via loadFromKeyStore(String keystoreFileName, String keystorePassword, Key key);

Several other supporting functions exist and will be documented prior to merge.  

This is the first part of the centralized crypto pieces.  Part two will be the creation of a centralized storage location.  Centralized crypto will store the private key in it's encrypted format as well as the certificate.  It will also store an encrypted copy of the cryptoStoreKey.  Both the store and key will be recovered by the add-on as needed for operation.


morph166955:
pom.xml conflict should be fixed.  This was the addon/binding change.

morph166955:
I'm down to just a few changes at this point. Code is very stable on the androidtv binding. I need to add a few extra functions to handle CA certs.

morph166955:
Just so this doesn't go stale, this is pending additional testing once the androidtv binding is reviewed and merged.

morph166955:
@openhab/core-maintainers This is ready for review.  I'm NOT expecting this to make it in for 4.0.0.  I highly doubt any binding developer will be able to adapt their binding in the 6 days between now and the code freeze.  I see this as a 4.1.0 feature.  Also note, this is part 1 of 2.  Part 2 will be the centralized crypto storage as discussed in #3289.  In theory, CryptoStore replaces and enhances the need for a KeyStore in the bindings (or anywhere else that's relevant).  CryptoStore implements several features such as key generation, encryption/decryption of strings, and self signed certificate creation.  Code is all based off of the AndroidTV binding implementation which seems to work very well.  Once this is locked down I'll work on part 2 with the goal of having both available for 4.1.0.

morph166955:
The builds are failing due to the automation integration tests, not due to any specific issue with this PR.

morph166955:
@openhab/core-maintainers Respectfully requesting a review on this. Code is very stable on AndroidTV binding. I really want to start working on the second part of this to try and get both into 4.1. Thank you!

bundles/org.openhab.core.crypto.cryptostore/pom.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.openhab.core.bundles</groupId>
+    <artifactId>org.openhab.core.reactor.bundles</artifactId>
+    <version>4.1.0-SNAPSHOT</version>
```
wborn, line 9:
```suggestion
    <version>4.2.0-SNAPSHOT</version>
```

holgerfriedrich:
@morph166955 I think it would be helpful to integrate bcprov-jdk18on as part of core.
I tried to pull in bouncy castle into KNX addon for a proof of concept for TPMs openhab/openhab-addons#15326 - resulting in +5500 class files and +7MB size of my addon jar.
The same thing seems to happen with androidtv you mentioned.

morph166955:
Fully agreed and this PR should do that already.  My ultimate goal is to really pull as much crypto as we can into a (very) reviewed and hardened implementation to make sure we don't have security leaks/holes in bindings.  bcpkix-jdk18on, bcprov-jdk18on, and bcutil-jdk18on are all added to bom/compile/pom.xml as part of this PR.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
```diff
@@ -0,0 +1,328 @@
...
+        Key key = convertByteToKey(keyString);
+        return decrypt(this.privKey, key);
+    }
+
+    public void setCert(String cert) {
+        this.cert = cert;
+    }
+
+    public void setCert(Certificate cert) throws CertificateEncodingException {
+        this.cert = new String(Base64.getEncoder().encode(cert.getEncoded()));
```
pacive, line 162: Why is it stored as a String instead of as a Certificate object?

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
```diff
@@ -0,0 +1,328 @@
...
+            throws GeneralSecurityException, IOException, Exception {
+        this.keystoreFileName = keystoreFileName;
+        loadFromKeyStore(keystorePassword, keyString);
+    }
+
+    public void loadFromKeyStore(String keystorePassword, byte[] keyString)
+            throws GeneralSecurityException, IOException, Exception {
+        Key key = convertByteToKey(keyString);
+        KeyStore keystore = KeyStore.getInstance("JKS");
+        FileInputStream keystoreInputStream = new FileInputStream(this.keystoreFileName);
```
pacive, line 260: Is the filename an absolute path? How do the binding developers make sure that it works properly on all supported operating systems?

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    public void loadFromKeyStore(String keystoreFileName, String keystorePassword, byte[] keyString)
+            throws GeneralSecurityException, IOException, Exception {
+        this.keystoreFileName = keystoreFileName;
+        loadFromKeyStore(keystorePassword, keyString);
+    }
+
+    public void loadFromKeyStore(String keystorePassword, byte[] keyString)
+            throws GeneralSecurityException, IOException, Exception {
```
pacive, line 257: Does any of the methods called throw `Exception`? Otherwise it's best to be as specific as possible regarding the possible exceptions thrown.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
```diff
@@ -0,0 +1,328 @@
...
+        this.privKey = encrypt(new String(Base64.getEncoder().encode(byteKey)), key);
+        setCert(keystore.getCertificate(this.alias));
+        String caAlias = this.caAlias;
+        Certificate caCert = keystore.getCertificate(caAlias);
+        if (caCert != null) {
+            setCaCert(caCert);
+        }
+    }
+
+    public KeyStore getKeyStore(String keystorePassword, byte[] keyString)
```
pacive, line 272: Is this for creating a new keystore? Would be better to name it accordingly

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    public void saveKeyStore(String keystorePassword, byte[] keyString)
+            throws GeneralSecurityException, IOException, Exception {
+        saveKeyStore(this.keystoreFileName, keystorePassword, keyString);
+    }
+
+    public void saveKeyStore(String keystoreFileName, String keystorePassword, byte[] keyString)
+            throws GeneralSecurityException, IOException, Exception {
+        FileOutputStream keystoreStream = new FileOutputStream(keystoreFileName);
```
pacive, line 295: If two bindings accidentally choose the same filename, they could overwrite each other's keystore.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
```diff
@@ -0,0 +1,328 @@
...
+            throws GeneralSecurityException, OperatorCreationException, IOException, Exception {
+        Key key = convertByteToKey(keyString);
+        KeyPairGenerator kpg = KeyPairGenerator.getInstance(this.keystoreAlgorithm);
+        kpg.initialize(this.keyLength);
+        KeyPair kp = kpg.generateKeyPair();
+        Security.addProvider(new BouncyCastleProvider());
+        Signature signer = Signature.getInstance("SHA256withRSA", "BC");
+        signer.initSign(kp.getPrivate());
+        signer.update("openhab".getBytes(StandardCharsets.UTF_8));
+        signer.sign();
```
pacive, line 323: What's the purpose of this?

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    public void saveKeyStore(String keystoreFileName, String keystorePassword, byte[] keyString)
+            throws GeneralSecurityException, IOException, Exception {
+        FileOutputStream keystoreStream = new FileOutputStream(keystoreFileName);
+        KeyStore keystore = getKeyStore(keystorePassword, keyString);
+        keystore.store(keystoreStream, keystorePassword.toCharArray());
+    }
+
+    private X509Certificate generateSelfSignedCertificate(KeyPair keyPair, String distName)
```
pacive, line 300: Why is `distName` passed as an argument instead of using the instance variable?

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
```diff
@@ -0,0 +1,328 @@
...
+            return "";
+        }
+    }
+
+    public void setPrivKey(String privKey, byte[] keyString) throws Exception {
+        Key key = convertByteToKey(keyString);
+        this.privKey = encrypt(privKey, key);
+    }
+
+    public String getPrivKey(byte[] keyString) throws Exception {
```
pacive, line 152: Wouldn't it be better to have methods for sign/verify and encrypt/decrypt in this class instead of exposing the private key? Or what other use case would a binding developer have to use the key?

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
```diff
@@ -0,0 +1,328 @@
...
+        Cipher encryptionCipher = this.encryptionCipher;
+        if (encryptionCipher != null) {
+            GCMParameterSpec spec = new GCMParameterSpec(keySize, encryptionCipher.getIV());
+            decryptionCipher.init(Cipher.DECRYPT_MODE, key, spec);
+            byte[] decryptedBytes = decryptionCipher.doFinal(dataInBytes);
+            return new String(decryptedBytes);
+        } else {
+            return "";
+        }
+    }
```
pacive, line 145: I think it's a bit confusing to have a CryptoStore for managing public/private keys and certificates, but at the same time it has encrypt/decrypt functions using symmetric encryption. I don't really see the need for encrypting the keys in-memory (what threat is it supposed to protect against?), it just adds complexity for the binding developers.

pacive (changes requested):
I made a quick review, but didn't have time to look at everything in detail unfortunately. Apart from the specific comments, JavaDoc needs to be added which will make further reviewing easier. Now I had to try to deduce the functionality and intended use by analyzing the code, which makes it take more time.

But keep up the good work, I think this would be a valuable feature!

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
```diff
@@ -0,0 +1,328 @@
...
+    private int keyLength = 2048;
+    private String alias = "openhab";
+    private String caAlias = "trustedCa";
+    private String distName = "CN=openHAB, O=openHAB, L=None, ST=None, C=None";
+    private String cipher = "AES/GCM/NoPadding";
+    private String keyAlgorithm = "";
+
+    private @Nullable Cipher encryptionCipher;
+
+    public CryptoStore() {
```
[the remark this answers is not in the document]
morph166955: I thought about that but that makes the assumption that we're saving to a file at all.  There may be times (which I have had to do in ATV) where we just want to work with a KeyStore in memory and just never save it off.  I could add a second constructor that has the file name in it (or any other constructor structure for that matter) if that's what's needed.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
morph166955: I went back and forth on this several times.  It was just easier to store and manipulate as a Base64 encoded String in the end.  If you look through some older versions of the AndroidTV PR I had it in as a Certificate and ended up changing it.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
morph166955: Yes, this is absolute path.  When implemented in ATV I used "String folderName = OpenHAB.getUserDataFolder() + "/androidtv";" to create a path supported accross implementations.  I suppose looking at it now that Windows could have an issue with the / versus \.  I'll check into that.  Either way, it's absolute so the binding can decide.

https://github.com/openhab/openhab-addons/blob/06b8c73537f0c03af55b0105cdeaa521ea1e239e/bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVConnectionManager.java#L453

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
morph166955: It does actually (and unfortunately).  I'd have to go back through my notes on which.  I agree, I hate putting Exception but in this case it actually just throws that.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
morph166955: Unfortunately I'm not sure how to resolve that in this structure.  In theory, the binding review process should catch accidental use of another binding name (or something generic like my.keystore).

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
morph166955: Yes, this generates a new keystore.  What would you suggest to adjust the name too?  getKeyStore seemed clean.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
morph166955: That generates a self signed certificate.  That's the part of the function that actually signs the certificate.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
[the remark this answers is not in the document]
pacive: If that is a usecase I agree that this is better.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
morph166955: I suppose I could pull that directly.  It's only called once internally and when it's called it uses this.distName.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
pacive: What kinds of manipulations would be needed? If a certificate is modified after being signed, the signature (and therefore the entire certificate) becomes invalid.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
pacive: I think it would be better to use `OpenHAB.getUserDataFolder() + "/keystore/" + this.filename` or something like it to make it easier to use.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
pacive: `getSomething` is usually used to get a field, `createKeyStore` would better reflect the purpose.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
morph166955: I added those into this in the event someone had a use case where they needed to access their private keys.  You're absolutely right, it should be held in the class IMHO.  I don't use these functions in ATV.  I was just trying to be complete by providing access should someone have a use case.  I can pull them out if they aren't needed.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
pacive: Would a binding need more than one keystore? Otherwise it might be best to just use the binding id? (Together with my other suggestion above)

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
pacive: From what I understand of the JavaDoc, this only creates a signature for the string "openhab", and the returned signature gets discarded anyway.

morph166955:
@pacive Thank you for the review!  I believe I've replied to all of your notes above.  Absolutely happy to keep talking about any if you don't feel the answer is solid.  As I mentioned above, the goal here is two fold.  

1) To create a central set of functions to deal with crypto so that the binding developers are not required to work that (especially since crypto can be confusing to those who don't work on it).  This also helps the overall security posture of the platform by reducing the risk of a bad/improper/mistake in those implementations.  

2) To prepare for a key warehouseing function to go into the core.  Several of the functions are designed with the concept of shared storage in mind and how to secure the bindings across (e.g. encryption of the keys).  The goal would be for the warehouse to hand a CryptoStore and a key to the binding when it asks for it (which would likely be in a recall manner as opposed to a "give me new" which CryptoStore would handle directly).

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
pacive: Yes but no matter how many keys you use to encrypt other keys, one must be stored somewhere in plaintext in memory (unless you want to ask the user to input it every time the private key is needed) so if someone is in a position to retrieve the private key, they can just as easily get the symmetric key and then decrypt it. 

I think your basic idea of a central warehouse is a good one, but as you say, cryptography is very hard to get right. I would be happy to continue discussing how it can be implemented in a good way.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
pacive: I think lines 308-310 is what signs the certificate.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
morph166955: I'll have to go back and look to see what the issue was.  I believe part of it was simply the way exceptions were handled (where Certificate throws several that we would have to catch and deal with, String did not which made the code much cleaner).  Another part of it was that it's very likely that the bindings are dealing with them in base64 (ATV was for example being sent a certificate as part of something else that I had to deal with) so flipping back and forth was becoming complicated.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
morph166955: So the question I have is, is this something we want to force onto the developers.  What if there is a use case to interact with something outside of the user data folder for example?

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
morph166955: That's fair.  In my original thinking, since we were never actually storing a KeyStore, a get seemed like a good facade to the binding developer where we were returning it.  I can change this easily enough.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
morph166955: Absolutely.  Even at the individual thing level yes.  For example, ATV uses a mechanism that creates a MITM for purposes of debugs.  I have to maintain a keystore for each side of that just for that connection.  Also, ATV works with two different protocol stacks in the case of ShieldTV, each uses it's own KeyStore.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
morph166955: Yes, looking back this looks like an artifact I missed from a previous revision.  I can remove it.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
morph166955: Agreed, and it's a step along the path.  Long term goal is to utilize things like TPMs to store keys.  I believe it will be a tiered mechanism that the user will ultimately decide.  For example, we could (for basic users) have them store a password in a file that's read on start.  For more advanced, we could query the user for that password on start to unlock the system.  For expert users, maybe we go with TPMs to store everything.

pacive: But what is the actual threat here? Who is the attacker that would benefit from retrieving the key? What would they gain? The ability to control your TV? And if they can access memory on your OH server, they could probably do much more damage than that without using the key. Don't get me wrong, having a central credential storage would be a great addition to OH, but the cost of added complexity needs to be weighed against the actual risk.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
pacive: Then this would probably have to be solved with an instruction on how to name the file (e.g bindingId + identifier) that's in the javadoc so it can be (more) easily checked when doing reviews

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
pacive: Was mainly a thought because the getter for the certificate returns a Certificate object, so there are more conversions when storing as a string that it would be the other way around. Providing a string as a setter should definitely be supported, but it would probably be better to get an exception when trying to set an invalid string-encoded cert than getting it when retrieving the cert.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
pacive: After reading a bit more on this, the KeyStore already is an implementation to store the keys securely (both in memory and on disk), so instead of implementing your own security solution, just retrieve the private key and certificate(s) from the KeyStore when needed and don't store it in your class' variables.

holgerfriedrich:
I played around a bit with your PR and wrote a simple test as well.
For most of the methods the crypto algorithms seem to be provided by the JDK - not BC. If I explicitly add “BC” to the getter, it does not find provider BC.
So I think we should make sure to add the provider BC - as you did for generating keys.
This could be done in the CTOR - and maybe using an Activate tag to ensure BC is properly added during start of the OSGI bundle.
Not sure if we should play with the precedence or query explicitly for “BC”.

bundles/org.openhab.core.crypto.cryptostore/NOTICE
```diff
@@ -0,0 +1,13 @@
...
+
+== Declared Project Licenses
+
+This program and the accompanying materials are made available under the terms
+of the Eclipse Public License 2.0 which is available at
+https://www.eclipse.org/legal/epl-2.0/.
+
+== Source Code
+
+https://github.com/openhab/openhab-core
```
holgerfriedrich, line 13:
```suggestion
https://github.com/openhab/openhab-core

== Third-party Content

bouncy castle
* License: MIT license
* Project: https://www.bouncycastle.org/
* Source:  https://www.bouncycastle.org/latest_releases.html
```

bom/compile/pom.xml
```diff
@@ -18,6 +18,7 @@
 
   <properties>
     <californium.version>2.7.4</californium.version>
+    <bouncycastle.version>1.75</bouncycastle.version>
```
holgerfriedrich, line 21:
```suggestion
    <bouncycastle.version>1.77</bouncycastle.version>
```

could we switch to 1.77 - or is there a reason to stick with 1.75 for now?

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
dbadia: I agree with pacive that encrypting keys in memory adds unnecessary complexity.  It's a nice goal, but it's a chicken and egg situation.  At some point, something has to be in cleartext which will unlock everything else.  The key isolation goal a very good one.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
dbadia: Instead of taking keystoreName as a parameter, what if we determine keystore name based on the namespace of the caller?  Bindings package names don't change, and I think it's pretty easy to walk up a stack trace tree to figure out who's invoking the method.  I can work on some sample code if it would help

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
```diff
@@ -0,0 +1,328 @@
...
+    public void loadFromKeyStore(String keystoreFileName, String keystorePassword, byte[] keyString)
+            throws GeneralSecurityException, IOException, Exception {
+        this.keystoreFileName = keystoreFileName;
+        loadFromKeyStore(keystorePassword, keyString);
+    }
+
+    public void loadFromKeyStore(String keystorePassword, byte[] keyString)
+            throws GeneralSecurityException, IOException, Exception {
+        Key key = convertByteToKey(keyString);
+        KeyStore keystore = KeyStore.getInstance("JKS");
```
dbadia, line 259: Couple of suggestions with regards to keystore type:

1.   Change from JKS format (which can only store asymmetric keys) to "JCEKS" which can store both asym and symmetric keys.  
2. Write the keystore type to a file on the filesystem, perhaps alongside the keystore itself.   This allows for:
  a. migration to a different keystore type in the future if necessary.  (There's some drama in the keystore space - Oracle recently change their default keystore from JKS to PCKS12, and JKS seems to be deprecated at this point)
  b. There may be cases where a FIPS compliant keystore is required which would warrant a different keystore type.  I don't think we need to go all the way down that rabbit hole just yet, but it would be good to have some scaffolding in place

J-N-K:
We could add a new interface `CredentialConsumer` which is automatically added to a `CredentialSupplierServlet` that allows entering credentials on a web-servlet. The `CredentialConsumer` could provide a description what is needed (a password, a private/public key, ...) and after reception store it in the crypto store. If it is bound to a thing, it could also implement `ThingHandlerService` which creates an OSGi component and injects the `ThingHandler`.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
morph166955: I went back and looked at the old commits and why I changed this.  This is absolutely because of exception handling.  I was trying to avoid a try/catch inside the class (so that the binding could handle the try/catch and logs were more accurate).  By storing as a string I avoided that (and could simply put it in a throws line for the end user to catch).

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
morph166955: I'll see if I can narrow down which specific line throws Exception.  Not a bad idea.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
morph166955: I went down the road of a single keystore and there were several issues.  First, working across things made it complex.  Second, the CA path was different for each which added complexity (because they were self signed certs by the device).  Ultimately, the work involved with making a single keystore wasn't worth it when maintaining multiple keystores was easy and kept a natural divide across the keys.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
morph166955: The concern here was a "rogue binding" more than anything else.  For example, a marketplace binding.  We didn't want to risk that binding being able to get a hold of credentials or something like that.  Also, this is very much for the warehouse than really for this part of the implementation.

bom/compile/pom.xml
morph166955: Yes, we can switch.  1.77 came out after my last push I believe.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
morph166955: What would be the use case for storing symmetric keys?  In general those are generated as needed for things like IPSEC connections.  

I am tracking the Oracle JKS/PKCS12.  We could migrate to PKCS12 if we wanted to.  I had some issues with it early on and JKS just worked.

dbadia: The openhab binding has symmetric (AES) keys for secure communication with the devices.  Looks like PKCS12 has basic support for symmetric (AES) keys.

morph166955: The AES bits there are just for storing the private key in the memory. It's not used at all for interaction with the device.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
```diff
@@ -0,0 +1,328 @@
...
+        } else {
+            return "";
+        }
+    }
+
+    public String decrypt(String encryptedData, Key key) throws Exception {
+        return decrypt(encryptedData, key, this.cipher);
+    }
+
+    public String decrypt(String encryptedData, Key key, String cipher) throws Exception {
```
dbadia, line 133: You should be able to change throws Exception to throws GeneralSecurityException

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
pacive: Isn't that solved in an equally secure way by the password on the KeyStore? Also, @J-N-K had a proposal for isolating bindings from each other in [this comment](https://github.com/openhab/openhab-core/issues/3289#issuecomment-1368783070), which wouldn't complicate things for binding developers.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
```diff
@@ -0,0 +1,328 @@
...
+        byte[] dataInBytes = Base64.getDecoder().decode(encryptedData);
+        Cipher decryptionCipher = Cipher.getInstance(cipher);
+        Cipher encryptionCipher = this.encryptionCipher;
+        if (encryptionCipher != null) {
+            GCMParameterSpec spec = new GCMParameterSpec(keySize, encryptionCipher.getIV());
```
dbadia, line 138: you are using a GCMParameterSpec object here but I don't see one used in the encrypt method.  You should probably just remove it from here as the cipher you selected has NoPadding, so IV is irrelevant.  Also, invoking encryptionCipher.getIV probably just returns an empty byte array

J-N-K:
@morph166955 I'll wait with a review until you discussions with @dbadia are finished. I'm not an expert on the crypto-parts, so it's probably better to discuss those issues first.

bundles/org.openhab.core.crypto.cryptostore/src/main/java/org/openhab/core/crypto/cryptostore/CryptoStore.java
dbadia: You are referring to the AES key in the CryptoStore.  I am referring to the AES keys in the zwave binding.  If the CryptoStore is going to be adopted, it needs to support storage of symmetric keys.  JKS does not allow that and JKS format is basically deprecated by Oracle.  

Use PCKS12 or JCEKS instead.  Use <keystore_name>.pkcs12 or <keykstore_name>.jceks so the format is obvious.. in case it needs to change in the future.   Best to change the default now instead of just creating a problem to have to solve later

bundles/org.openhab.core.crypto.cryptostore/pom.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.openhab.core.bundles</groupId>
+    <artifactId>org.openhab.core.reactor.bundles</artifactId>
+    <version>4.2.0-SNAPSHOT</version>
```
wborn, line 9:
```suggestion
    <version>4.3.0-SNAPSHOT</version>
```

kaikreuzer:
@morph166955 Is that not of interest anymore?

morph166955:
> @morph166955 Is that not of interest anymore?

Looking back through the comments, and given that in the last 2 years since this was submitted not much has changed across OH in the terms of crypto implementations, it may be better just for the bindings to handle this directly as they all have different needs.  The only thing I would suggest would be to pull things like bouncycastle into the core dependencies so that the bindings don't use different versions.

PR #3304, opened 2023-01-08
Move config description URI to type-base class

J-N-K:
This allows simplifying methods dealing with config descriptions because they can be re-used for different types.


J-N-K:
Integrated in ThingManagerImpl refactoring as it is needed there anyway

PR #3307, opened 2023-01-09
Prevent thing initialization with missing bridge

J-N-K:
If a thing type has `supported-bridge-type-refs` set, a configured bridge is mandatory. In addition to the already existing config validations a check for that has been added.

Related to https://github.com/openhab/openhab-addons/issues/10802

@openhab/add-ons-maintainers If this is merged, things don't need to check whether a bridge is configured.

### DIFF
--- a/bundles/org.openhab.core.config.core/src/main/resources/OH-INF/i18n/validation.properties
+++ b/bundles/org.openhab.core.config.core/src/main/resources/OH-INF/i18n/validation.properties
@@ -11,3 +11,5 @@ min_value_numeric_violated=The value must not be less than {0}.
 pattern_violated=The value {0} does not match the pattern {1}.
 options_violated=The value {0} does not match allowed parameter options. Allowed options are: {1}
 multiple_limit_violated=Only {0} elements are allowed but {1} are provided.
+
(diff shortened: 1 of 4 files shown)


clinique:
Is the bridge supposed to be initialized also before calling initialize on child's ?

J-N-K:
Ig the bridge is not initialized, the child handler is not initialized (status `UNINITIALIZED/BRIDGE_UNINITIALIZED`), if the bridge initializes, child things are check and initialized, too.

clinique:
Thing is that getBridge is Nullable, has to be tested to get handler that is also Nullable. We could maybe improve the framework on this.

J-N-K:
That's difficult, because if we have a thing without a bridge, what should `getBridge` return? Maybe it would help to make `getBridge` and `getHandler` return `Optional<T>` instead, then you could `getBridge().map(Thing::getHandler)` and skip one null check (or more, depending on what you want to do). But that is heavy API breaking.

clinique:
Or maybe have a childThingHandler, constructed with its bridge, and being sure he gets initialized once the bridge is.

J-N-K:
Superseded by #3330

PR #3310, opened 2023-01-11
[transport.upnp] Slightly decrease risk of removing callback without calling end()

jlaur:
Call `callback.end()` after removal to avoid concurrency situation of subscription being created right between looking for callback and removing it. Implementation uses `ConcurrentHashMap` without explicit locking. In this case `end()` would not be called, but callback would still be removed.

I have no proof of this situation happening but was reviewing some of the code while working on openhab/openhab-addons#14163.


J-N-K (approved):
LGTM, Thanks.

Issue #3300, opened 2023-01-05
[REST] Return all metadata by default


spacemanspiff2007:
Currently the metadata selector for the  `/items/{itemname}` has to be used to return the item metadata.
To get all metadata one has to use a regular expression, e.g. `.+`.
However not all users know regular expressions and it's not clear for all users that metadata can be requested this way.

I propose a change in behavior:
By default the whole metadata for the item is returned. 
The metadata selector can be used to reduce the returned metadata.
That way it would be a backwards compatible change.

---

This is a follow up issue for #3298

florian-h05:
If there returning all metadata does not lead to performance drawbacks: +1

spacemanspiff2007:
@J-N-K since you're the master of metadata I'd like to hear your opinion. I think from a usability perspective it absolutely makes sense. Do you have any performance concerns?

J-N-K:
Not sure. The code for fetching metadata is quite complex due to the applied filters, this can be easily improved if we simply want "everything". 

With the current code fetching all items (360 item) from my production system without metadata takes around 350 ms (107 kB) and fetching all items with `.+` takes around 400 ms (123 kB). This is just a quick test made with the API Explorer from a remote machine, so I'm not sure how much is really related to backend processing or network transfer. 

Anyway, it seems that getting all metadata takes about 15% more time in this scenario. It's worth noting that the range without metadata was 337 to 482 ms and with metadata 389 to 437 ms (I tried ten times each). So maybe the "real life" difference is small. Maybe you can try that on your systems, too.

spacemanspiff2007:
I've written a short script that makes 50 requests and measures the time against my production system.
(140k without metadata, 150k with metadata)

````
Without metadata
Avg: 313.7ms, best: 282.0ms, worst: 438.0ms
Quantile: 50% between 297.0ms and 313.0ms

With metadata
Avg: 320.6ms, best: 297.0ms, worst: 344.0ms
Quantile: 50% between 312.0ms and 328.0ms

Difference:
Avg: +2.2%, best: +5.1%, worst: -27.3%
Quantile:+4.8% and +4.6%
````

On average the request with the metadata filter took 2.2% longer, however the 50% of the requests took ~5% longer.

I run it a couple of times and the avg response time has a +-20% jitter and the worst response time was 300% of the previous worst response time.
So it seems to takes longer, but it's not the limiting factor and the response time is absolutely not deterministic anyway.


[...]

PR #3311, opened 2023-01-12
Return item metadata by default

spacemanspiff2007:


kaikreuzer:
Why should it do this?

spacemanspiff2007:
Items and metadata are getting more and more intertwined.
Unfortunately it's not clear for every user how to receive the corresponding metadata.
This lead to e.g. implementation of #3298 which should have just been a `GET` for `/items/{itemname}`.

MainUI already sets the metadata parameter for requesting the item.
But the parameter was also wrong (see https://github.com/openhab/openhab-webui/issues/1633) but will be fixed in the next release.

It seems it's very hard for the majority of openHAB RestAPI users to set the corresponding parameter correctly,
but for almost all users it makes sense to return the metadata with the items.
This change will improve usability for all RestAPI users.

See also discussion in #3300

J-N-K:
@spacemanspiff2007 In https://github.com/openhab/openhab-core/pull/3298#issuecomment-1377803222 you argue that fetching namespace data only (which is a some bytes only) instead of the full item (which is more on the order of kB) is wrong, because JSON processing is fast and the larger amount of data doesn't matter. Here you argue exactly the opposite. What is the difference?

J-N-K:
@kaikreuzer Do you think there is a need for a deeper discussion here? I would be fine with this change, and I don't think we'll run into any performance issues with that.

kaikreuzer:
I'd love to hear the opinion of @ghys as our "main" consumer of the API and whether this might have any impact (in terms of necessary changes) on the Main UI.

spacemanspiff2007:
This is a backwards compatible change so I expect no surprises.
MainUI addopted the filter accordingly in https://github.com/openhab/openhab-webui/pull/1623/files

grafik

J-N-K:
I tend to agree with @spacemanspiff2007, it is backward compatible in the sense that only more, never less information is returned and nearly all current usages already employ a filter and are not affected at all. With https://github.com/openhab/openhab-distro/pull/1460 the network load is also considerably reduces.

@ghys, do you disagree here?

ghys:
I can see the appeal of responding with all metadata by default, but please consider the following as well:

- since MetadataProviders can be arbitrarily injected (including by add-ons installed outside the reviewed set which is part of the distribution), we can't know how computationally intensive all those they'll be, therefore a ` /rest/items` request call might involve running code that isn't necessarily needed and will effectively affect performance. If all is needed is a simple (item name, label) tuple then it's clearly not ideal
- There should be a way - maybe there is one but it's not clear - to explicitly specify that no metadata is needed (for those scenarios as the one described above)
- Probably more related to https://github.com/openhab/openhab-distro/pull/1460 but I'd be very wary of applying a blanket GZIP compression to all resources. It might for one conflict with https://github.com/openhab/openhab-webui/pull/1521 which allowed to serve pre-compressed UI assets in GZIP or Brotli format, without involving on-the-fly compression.

spacemanspiff2007:
> If all is needed is a simple (item name, label) tuple then it's clearly not ideal
> ...
> There should be a way - maybe there is one but it's not clear - to explicitly specify that no metadata is needed (for those > scenarios as the one described above)

Isn't that the `fields` parameter?

> It might for one conflict

How could one test if it really does conflict or not?

---

Note that with the current version of the new `UoM` you'll need the `unit` metadata for each Number item otherwise you don't have a full item definition.

florian-h05:
> It seems it's very hard for the majority of openHAB RestAPI users to set the corresponding parameter correctly

If a decision is made against returning all metadata by default, improving the SWAGGER docs might also be an option.
Anyway, IMO #3298 can be reverted.

J-N-K:
> Since metadata play such a crucial part and are a core functionality (`autoupdate`, `expire`, etc.) maybe it would be good to deprecate the possibility to inject arbitrary MetadataProviders.

The whole idea of OSGi and DS is that we can inject services later. And I don't see an issue with that.

Regarding a request "without metadata": If you provide `^$` as metadata filter this should not match anything. It's kind of the inverse of what we have now (`.+` to get everything).

Regarding ZIP: I guess we can also limit that to `/rest` to prevent issues with other endpoints that we don't control.

spacemanspiff2007:
> The whole idea of OSGi and DS is that we can inject services later. And I don't see an issue with that.

Me neither, but if it's e.g. "too dangerous"  then my next suggestion will of course be to not allow it :wink: 

> Regarding a request "without metadata": If you provide `^$` as metadata filter this should not match anything. It's kind of the inverse of what we have now (`.+` to get everything).

A modified implementation with a sentinel value (e.g. `"SKIP_METADATA"`) would be even nicer.
The current implementation also matches all metadata namespaces with the regex but we could just skip the match and return every entry ...

J-N-K:
@kaikreuzer  I would like to merge this. By supplying a proper REGEX metadata can be suppressed (similar to now, where a REGEX is needed to return everything), so it is still possible to get what we have now. If namespaces are provided it also behaves the same way as before. 

The only concern left is that getting the metadata might be a time-consuming task. If this turns out to be the case we can add additional indices to the `MetadataRegistry` that speed this up.

florian-h05:
> By supplying a proper REGEX metadata can be suppressed 

Can this probably be documented in the SWAGGER annotations, as well as a note that REGEX is used for selecting metadata?

J-N-K:
Currently the documentation is:

```
metadata selector - a comma separated list or a regular expression (suppressed if no value given)
```

and would be changed to

```
metadata selector - a comma separated list or a regular expression (returns all if no value given)
```

What else would you expect?

florian-h05:
Looks good, I must have overread that.

kaikreuzer (approved):
Ok, let's go for it then.

PR #3319, opened 2023-01-15
Fix nullable errors

kaikreuzer:

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/addon/AddonI18nUtil.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/addon/AddonI18nUtil.java
@@ -36,15 +36,16 @@ public class AddonI18nUtil {
         this.i18nProvider = i18nProvider;
     }
 
-    public @Nullable String getDescription(Bundle bundle, String bindingId, @Nullable String defaultDescription,
-            @Nullable Locale locale) {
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core/src/main/java/org/openhab/core/addon/AddonI18nLocalizationService.java
```diff
@@ -53,6 +53,14 @@ public AddonInfo createLocalizedAddonInfo(Bundle bundle, AddonInfo addonInfo, @N
         String name = addonI18NUtil.getName(bundle, addonInfoUID, addonInfo.getName(), locale);
         String description = addonI18NUtil.getDescription(bundle, addonInfoUID, addonInfo.getDescription(), locale);
 
+        if (name == null) {
```
J-N-K, line 56: IMO `AddonI18NUtil` should be fixed that the return values are `@NonNull`, because we supply a default value that is non-null. In `.getDescription` the parameter annotation should be changed, too.

kaikreuzer: Better now?

J-N-K (approved):
LGTM, thanks. IntelliJ didn't complain and I believe this was more an analysis error than a real world problem.

kaikreuzer:
Yes, I agree, but Eclipse shows an error, so it needed to be fixed.

PR #3320, opened 2023-01-16
Fix addon.xsd

J-N-K:
The namespace was missing, leading to problems in SAT


kaikreuzer (approved):

jlaur:
@J-N-K - since this never appeared in a milestone release, you can add the "regression" tag to avoid it being mentioned in the release notes.

Issue #3332, opened 2023-01-22
unable to install ui widget from marketplace if yaml contains default value


crnjan:
Having ui widget, for example https://community.openhab.org/t/compass-card/143589 where we use `default` as
```
...
props:
  parameters:
    - default: if:mdi:arrow-compass
      label: Arrow image
      name: arrowImage
      required: false
      type: TEXT
```
will give the following error when trying to install via marketplace from main ui:
```
2023-01-22 10:05:06.422 [ERROR] [munity.CommunityUIWidgetAddonHandler] - Unable to parse YAML: Unrecognized field "default" (class org.openhab.core.config.core.dto.ConfigDescriptionParameterDTO), not marked as ignorable (22 known properties: "readOnly", "max", "limitToOptions", "groupName", "name", "stepsize", "context", "defaultValue", "min", "label", "filterCriteria", "verify", "type", "description", "options", "required", "advanced", "unit", "multipleLimit", "pattern", "multiple", "unitLabel"])
 at [Source: (StringReader); line: 21, column: 21] (through reference chain: org.openhab.core.ui.components.RootUIComponent["props"]->org.openhab.core.config.core.dto.ConfigDescriptionDTO["parameters"]->java.util.ArrayList[3]->org.openhab.core.config.core.dto.ConfigDescriptionParameterDTO["default"])
2023-01-22 10:05:06.422 [ERROR] [munity.CommunityUIWidgetAddonHandler] - Widget from marketplace is invalid: Unable to parse YAML
```
did a quick spike and seems as `@SerializedName` [here](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/dto/ConfigDescriptionParameterDTO.java#L33) is not used (ignored) by jackson so only `defaultValue` is allowed while `default` is not  ... [this](https://stackoverflow.com/questions/72027481/make-jackson-use-gson-annotations) is maybe worth checking ...

My guess is that adding something like `@JsonAlias({"default"})` should do the trick ...

I'm not a java or OH core expert so pls take it with a grain of salt ...

PR #3334, opened 2023-01-23
Fix parsing of YAML default value for community marketplace UI Widgets

crnjan:
Unable to install ui widget from marketplace if yaml contains default value, please see https://github.com/openhab/openhab-core/issues/3332 for details.
fixes https://github.com/openhab/openhab-core/issues/3332
Signed-off-by: Boris Krivonog boris.krivonog@inova.si

### DIFF
--- a/bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityUIWidgetAddonHandler.java
+++ b/bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityUIWidgetAddonHandler.java
@@ -33,6 +33,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.introspect.AnnotationIntrospectorPair;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 
 /**
@@ -60,6 +61,8 @@ public class CommunityUIWidgetAddonHandler implements MarketplaceAddonHandler {
         this.yamlMapper = new ObjectMapper(new YAMLFactory());
         yamlMapper.findAndRegisterModules();
         this.yamlMapper.setDateFormat(new SimpleDateFormat("MMM d, yyyy, hh:mm:ss aa", Locale.ENGLISH));
+        yamlMapper.setAnnotationIntrospector(new AnnotationIntrospectorPair(new SerializedNameAnnotationIntrospector(),
+                yamlMapper.getSerializationConfig().getAnnotationIntrospector()));
     }
 
     @Override
--- a/bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/SerializedNameAnnotationIntrospector.java
+++ b/bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/SerializedNameAnnotationIntrospector.java
@@ -1,0 +1,53 @@
+/**
+ * Copyright (c) 2010-2023 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.addon.marketplace.internal.community;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import com.fasterxml.jackson.core.Version;
+import com.fasterxml.jackson.databind.AnnotationIntrospector;
+import com.fasterxml.jackson.databind.PropertyName;
+import com.fasterxml.jackson.databind.introspect.Annotated;
+import com.google.gson.annotations.SerializedName;
+
+/**
+ * This is a {@link SerializedNameAnnotationIntrospector}, which processes SerializedName annotations.
+ *
+ * @author Boris Krivonog - Initial contribution
+ *
+ */
+final class SerializedNameAnnotationIntrospector extends AnnotationIntrospector {
+    private static final long serialVersionUID = 1L;
+
+    @Override
+    public PropertyName findNameForDeserialization(Annotated annotated) {
+        Optional<SerializedName> serializedName = Optional.ofNullable(annotated.getAnnotation(SerializedName.class));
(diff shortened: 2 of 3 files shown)


J-N-K:
I don't think we should introduce a dependency to Jackson to `openhab-core-base`. Why can't the widget just use `defaultValue`? That should work from my understanding.

hmerk:
@J-N-K You are right, it works with defaultValue, already tested.
There was an issue with the widget editor not accepting defaultValue, but I think that one was resolved in the meantime.

crnjan:
@hmerk are you sure? afaik `defaultValue` is not (anymore) used by widgets ... I already have the patched core version so need to downgrade to official version to give it a try, but AFAIK `defaultValue` was renamed to `default` and widgets are ignoring `defaultValue` ... https://github.com/openhab/openhab-core/pull/2383

hmerk:
@ghys Can you give us an update on this.
When using defaultValue, widgets can be installed via marketplace, we had this lately.
So what would be the proper usage here ?

crnjan:
I can only say widgets don't care about `defaultValue` ...

hmerk:
The post (mine 😉) you are refering was using default: in the widget….

crnjan:
I wanted to publish a widget (https://community.openhab.org/t/compass-card/143589) and it took quite some time before I realised what was wrong - struck two issues, one - non published widgets are not displayed, the other, widgets with `default` are failing to install ... please let me know what/how is the correct syntax and I will try my best to help out :)

crnjan:
The problem with `default` is that you cannot publish it to marketplace @hmerk since install will fail with 
```2023-01-22 10:05:06.422 [ERROR] [munity.CommunityUIWidgetAddonHandler] - Unable to parse YAML: Unrecognized field "default" (class org.openhab.core.config.core.dto.ConfigDescriptionParameterDTO), not marked as ignorable (22 known properties: "readOnly", "max", "limitToOptions", "groupName", "name", "stepsize", "context", "defaultValue", "min", "label", "filterCriteria", "verify", "type", "description", "options", "required", "advanced", "unit", "multipleLimit", "pattern", "multiple", "unitLabel"])
 at [Source: (StringReader); line: 21, column: 21] (through reference chain: org.openhab.core.ui.components.RootUIComponent["props"]->org.openhab.core.config.core.dto.ConfigDescriptionDTO["parameters"]->java.util.ArrayList[3]->org.openhab.core.config.core.dto.ConfigDescriptionParameterDTO["default"])
2023-01-22 10:05:06.422 [ERROR] [munity.CommunityUIWidgetAddonHandler] - Widget from marketplace is invalid: Unable to parse YAML```
ofc it works if you copy paste, but you cannot promote it to marketplace ... please correct me if I'm wrong ...

hmerk:
@crnjan please see here
https://community.openhab.org/t/service-status-standalone-widget/128041/8
It confirms if using defaultValue, installation from marketplace works.

crnjan:
hm, so one needs to use `defaultValue` for marketplace to get around this issue?

hmerk:
That’s what I said.
On the other hand, changing marketplace addons parser to use default instead of defaultValue would make it consistent to core. Culprit would be that all widgets using defaultValue would need to be changes….

crnjan:
based on code, that was the intention `@SerializedName(value = "default", alternate = "defaultValue")` so both `default` (which would actually be used for saving as well) and `defaultValue` would do the trick, but jackson ignores `SerializedName` ...

J-N-K:
Looking at the annotations and the discussion in https://github.com/openhab/openhab-core/pull/2383 it seems that the desired naming is `default` and `defaultValue` is the deprecated alternate/fallback name that should have been removed for openHAB 3.2. 

We could change the field name to `default` and keep the annotations, that would probably make the Jackson parser use `default`, but as @hmerk pointed out this will require changes to all widgets that currently use `defaultValue` (because that would then be the "unsupported" name).

crnjan:
> We could change the field name to `default`

IMHO, while far from ideal, if no other way to allow fallback (`defaultValue`), that is the way to go in order to make things better (cleaner?) in the future ...

hmerk:
Just checked, 142 widget would need to be checked, but there is just a very small amount using default values at all, otherwise we would have seen more complaints in the community about not installable widgets.
I can definitely check „mine“, which would reduce about 10, we can ping Rich to check his widgets, which could be about the same number…

crnjan:
but any widget on marketplace using `default` will fail to install ...

crnjan:
it would really be unfortunate if there is no (easy) way to support `defaultValue` as well ... but on the other hand it's better to cut some loose ends and continue forward ... please let me know if I can help with anything and I will be glad to do so ...

hmerk:
I think we forgot some major points here.
Changing the parser from using defaultValue to default and correcting the widgets on the marketplace will break installation for all openHAB versions befor this change.
So we definitely need to find a solution which suits both….

rkoshak:
Does this also impact rule templates? 

Interestingly I *do* have some widgets (e.g. Battery Level Status) that are using `defaultValue` as well as some of my rule templates. However, when I look in my JSONDB they've all been converted to `default`. That makes it hard to do a quick search to find out. :-(

Out of around 20 widgets I've published, two are using `defaultValue` and one is using `default` already.

hmerk:
Don‘t know about rules templates, never used them.
But it is interesting, that defaultValue is transformed to default.
So it seems to be an UI issue reading the YAML but not accepting default instead expecting defaultValue.

ghys:
The problem is related to https://github.com/openhab/openhab-core/pull/2376#discussion_r637870924 & https://github.com/openhab/openhab-core/pull/2383.

The YAML parser which parses marketplace entries probably isn't aware of this
`@SerializedName(value = "default", alternate = "defaultValue")`

so it expects `defaultValue`, while REST API calls are returning `default` instead in JSON responses, and are then re-serialized client-side back to YAML to display in the UI.
Then when people copy-paste their widgets from the UI to the forum, those copy-pastes include `default` instead of `defaultValue`, that's why the server-side deserializing fails.

ghys:
This could maybe also work: https://stackoverflow.com/a/72030015

crnjan:
Tried to address this issue as originally suggested in https://github.com/openhab/openhab-core/issues/3332#issue-1552106322 (seems as @ghys found same stackoverflow entry :) ) ...

crnjan:
Tried to address this issue as originally suggested in https://github.com/openhab/openhab-core/issues/3332#issue-1552106322 (seems as @ghys found same stackoverflow entry :) ) ...

crnjan:
I checked with marketplace widgets containing either `default` or `defaultValue` and both cases worked with this changes. Maybe something else I can try? Change should "only" affect ui widgets and not other marketplace items. @J-N-K, dependency to Jackson was also removed from `openhab-core-base`, thx for pointing this out, was not aware.

J-N-K (approved):
Thanks, LGTM.

Since I'm no expert using Jackson, @ghys can you have a look?

ghys:
LGTM as well!

hmerk:
@openhab/core-maintainers Would make sense to cherry pick thisone for openHAB 3.4 as well.

PR #3335, opened 2023-01-24
"Cacheability" option for critical REST resources

### LINKED ISSUE #3329: Support Last-Modified or ETag headers on critical REST API endpoints
Idea from https://github.com/openhab/openhab-webui/issues/1650#issuecomment-1399446960:

The Main UI (and possibly others) makes relatively frequent REST API calls to get all things, or item, or rules, or UI components to make sure the data is up-to-date. For instance every time an item or thing picker is rendered, a call to `/rest/items` resp. `/rest/things` is performed.
`/rest/items` is also called every time the home page is shown, to compute the model cards (locations, equipment, points) with the latest data.

It could be a huge performance improvement if these critical and costly endpoints supported either the `ETag` or `Last-Modified` HTTP response headers, and their equivalent request headers (`If-None-Match` and `If-Modified-Since`).
If the preconditions don't match (see [JAX-RS - Handling 'Last-Modified' and 'ETag' headers with Request Injection](https://www.logicbig.com/tutorials/java-ee-tutorial/jax-rs/request-context.html)) then a 304 Not Modified response would be sent instead.

The UI and other clients would store the last known list of entities in a client-side cache (possibly even surviving page refreshes with IndexedDB), as well as the last sent EntityTag or modified date retrieved from the response headers, which would then be sent along with future requests. It would react to a 200 OK response by updating the cache before returning the data to the caller, and to a 304 Not Modified response by simply returning the cache.

To determine if the data has changed, resources backed by a Registry could keep the last sent EntityTag or last modified date as well, and add a [change listener](https://www.openhab.org/javadoc/latest/org/openhab/core/common/registry/registry#addRegistryChangeListener(org.openhab.core.common.registry.RegistryChangeListener)) to reset this last ETag or date when a change was notified.

For items it would surely be a little more complicated since changes to metadata would have to be handled as well.

ghys:
Closes #3329.

This implements a new optional `cacheable` parameter for these REST endpoints:
- `/rest/items`
- `/rest/things`
- `/rest/rules`

When this parameter is set, a flat list of all elements excluding non-cacheable fields (e.g. "state", "transformedState", "stateDescription", "commandDescription" for items, "statusInfo", "firmwareStatus", "properties" for things, "status" for rules) will be retrieved along with a `Last-Modified` HTTP response header. When unknown, the Last-Modified header will be set to the date of the request.

Also only when this parameter is set, and a `If-Modified-Since` header is found in the request, that header will be compared to the last known modified date for the corresponding cacheable list. The last modified date will be reset when any change is made on the elements of the underlying registry. If the `If-Modified-Since` date is equal or more recent than the last modified date, then a 304 Not Modified response with no content will be served instead of the usual 200 OK, informing the client that its cache is still valid at the provided date.

All other request parameters will be ignored except for "metadata" in the `/rest/items` endpoint. When a metadata selector is set, the resulting item list will be considered like a completely different resource, i.e. it will have its own last modified date. Regarding metadata, the approach to invalidating last modified dates is very conservative: when any metadata is changed, all cacheable lists of items will have their last modified date reset even if the change was in a metadata namespace that wasn't requested.

This also implements the abovedescribed behavior for the `/rest/ui/components/{namespace}` endpoint, but no `cacheable` parameter is necessary. The last modified date is tracked by namespace.


ghys:
Examples (using [HTTPie](https://httpie.io/docs/cli)):

The first request at 08:24:14 GMT responds with a flat list of items retrieved with the cacheable parameter. Note the Last-Modified response header:
```
$ http 'http://localhost:8080/rest/items?cacheable=true'
HTTP/1.1 200 OK
Content-Type: application/json
Date: Tue, 24 Jan 2023 08:24:14 GMT
Last-Modified: Tue, 24 Jan 2023 08:24:14 GMT
Server: Jetty(9.4.46.v20220331)
Transfer-Encoding: chunked

[
    {
        "editable": false,
        "groupNames": [
            "DemoSwitchGroup"
        ],
        "link": "http://localhost:8080/rest/items/DemoSwitch",
        "name": "DemoSwitch",
        "tags": [],
        "type": "Switch"
    },
    ...
]
```

Supposing the client has cached this list, a second request is made with the If-Modified-Since set to the Last-Modified of the previous request (could also be the current date):
```
$ http 'http://localhost:8080/rest/items?cacheable=true' 'If-Modified-Since: Tue, 24 Jan 2023 08:24:14 GMT'
HTTP/1.1 304 Not Modified
Content-Length: 0
Date: Tue, 24 Jan 2023 08:25:26 GMT
Server: Jetty(9.4.46.v20220331)
```

Of course, requests without the header still return the list (with or without Last-Modified depending on the cacheable parameter):
```
$ http HEAD 'http://localhost:8080/rest/items?cacheable=true'
HTTP/1.1 200 OK
Content-Type: application/json
Date: Tue, 24 Jan 2023 08:27:32 GMT
Last-Modified: Tue, 24 Jan 2023 08:24:14 GMT
Server: Jetty(9.4.46.v20220331)
Transfer-Encoding: chunked

$ http HEAD 'http://localhost:8080/rest/items'
HTTP/1.1 200 OK
Content-Type: application/json
Date: Tue, 24 Jan 2023 08:27:56 GMT
Server: Jetty(9.4.46.v20220331)
Transfer-Encoding: chunked
```

If any item (or metadata) was changed, then the list will be sent again, with a new Last-Modified header:
```
$ http 'http://localhost:8080/rest/items?cacheable=true' 'If-Modified-Since: Tue, 24 Jan 2023 08:24:14 GMT'
HTTP/1.1 200 OK
Content-Type: application/json
Date: Tue, 24 Jan 2023 08:29:03 GMT
Last-Modified: Tue, 24 Jan 2023 08:29:03 GMT
Server: Jetty(9.4.46.v20220331)
Transfer-Encoding: chunked

[
    {
        "editable": false,
        "groupNames": [
            "DemoSwitchGroup"
        ],
        "link": "http://localhost:8080/rest/items/DemoSwitch",
        "name": "DemoSwitch",
        "tags": [],
        "type": "Switch"
    },
    ...
]
```
(note: the last modified date is the date when the first request was made after the actual change, not the real date of the change; since it's only precise to the second in those headers, it's to mitigate race conditions if changes and requests are made within the same second).

Requests for UI components are cacheable by default:
```
$ http 'http://localhost:8080/rest/ui/components/ui:page'
HTTP/1.1 200 OK
Content-Type: application/json
Date: Tue, 24 Jan 2023 09:13:39 GMT
Last-Modified: Tue, 24 Jan 2023 08:26:32 GMT
Server: Jetty(9.4.46.v20220331)
Transfer-Encoding: chunked

[
    {
        "component": "oh-layout-page",
        "config": {
            "label": "Overview"
        },
    ...
]

$ http 'http://localhost:8080/rest/ui/components/ui:page' 'If-Modified-Since: Tue, 24 Jan 2023 09:00:00 GMT'
HTTP/1.1 304 Not Modified
Content-Length: 0
Date: Tue, 24 Jan 2023 09:15:05 GMT
Server: Jetty(9.4.46.v20220331)
```

If used carefully with client-side caching by UIs and other clients, this has the potential of saving a ton of bandwidth.

bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/RuleResource.java
```diff
@@ -138,15 +147,47 @@ public RuleResource( //
...
     @GET
     @Produces(MediaType.APPLICATION_JSON)
     @Operation(operationId = "getRules", summary = "Get available rules, optionally filtered by tags and/or prefix.", responses = {
             @ApiResponse(responseCode = "200", description = "OK", content = @Content(array = @ArraySchema(schema = @Schema(implementation = EnrichedRuleDTO.class)))) })
-    public Response get(@QueryParam("prefix") final @Nullable String prefix,
+    public Response get(@Context Request request, @QueryParam("prefix") final @Nullable String prefix,
             @QueryParam("tags") final @Nullable List<String> tags,
-            @QueryParam("summary") @Parameter(description = "summary fields only") @Nullable Boolean summary) {
+            @QueryParam("summary") @Parameter(description = "summary fields only") @Nullable Boolean summary,
+            @DefaultValue("false") @QueryParam("cacheable") @Parameter(description = "provides a cacheable list and honors the If-Modified-Since header, all other parameters are ignored") boolean cacheable) {
```
J-N-K, line 166: I don't think that's a good naming. It is not clear what is actually returned if `cachable=true` is used. It would be better to add a parameter to return structural/static content only (i.e. without status/state). If that is set (which means, this is data that is likely not going to change), the `If-Modified-Since` header should be processed.

J-N-K (changes requested):

bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/RuleResource.java
ghys: > It would be better to add a parameter to return structural/static content only (i.e. without status/state)

That's basically that this parameter does, I'll admit it was not my first choice either, but after thinking long and hard this is the best I could come up with. It conveys that it will only output information that is ready to be cached (and nothing more) along with the headers to let the browsers do it.

I'm open to suggestions to improve the name 🙂

J-N-K: What about "suppressRuntimeData"? Or "limitToStaticData"?

florian-h05:
@ghys @J-N-K What‘s the state here? This looks like a great improvement and I‘d really like to have it for openHAB 4, so sorry for pinging!

florian-h05:
Okay, thanks for the status. 
No hurries, openHAB 4 is coming closer but I think there is enough time left to finish this PR and the UI part of it.

Regarding the naming, something like `staticDataOnly` would IMO be a good option.

J-N-K:
@ghys I like the naming suggested by @florian-h05. Would you update this PR and rebase, so we can merge?

florian-h05:
DCO is failing, but other then this it seems finished to me, ping @J-N-K.

J-N-K (approved):
LGTM

J-N-K:
@ghys Can you rebase to solve the merge conflict and check the sign-off of you last commit?

ghys:
Yeah please bear with me, I'm using this PR and made the changes "in the blind" to check out the GitHub Web IDE (https://github.dev/openhab/openhab-core/pull/3335) as a contributor.
It didn't sign-off that one commit, maybe you can set it to pass this one time, the sign-off line is the same as the others, you know the commit is from me.

As for the rebase this is also a good opportunity to test https://github.blog/changelog/2022-02-03-more-ways-to-keep-your-pull-request-branch-up-to-date/, I've activated it on the webui repo for a while (it's under "General" now, they've designed the settings) so you can rebase yourself if needed, it's always nice to have the option instead of asking the contributor when merging stale PRs.

J-N-K (approved):

bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/RuleResource.java
```diff
@@ -141,20 +150,52 @@ public RuleResource( //
...
     }
 
     @GET
     @RolesAllowed({ Role.USER, Role.ADMIN })
     @Produces(MediaType.APPLICATION_JSON)
     @Operation(operationId = "getRules", summary = "Get available rules, optionally filtered by tags and/or prefix.", responses = {
             @ApiResponse(responseCode = "200", description = "OK", content = @Content(array = @ArraySchema(schema = @Schema(implementation = EnrichedRuleDTO.class)))) })
-    public Response get(@Context SecurityContext securityContext, @QueryParam("prefix") final @Nullable String prefix,
+    public Response get(@Context SecurityContext securityContext, @Context Request request, @QueryParam("prefix") final @Nullable String prefix,
             @QueryParam("tags") final @Nullable List<String> tags,
```
J-N-K, line 168:
```suggestion
    public Response get(@Context SecurityContext securityContext, @Context Request request,
            @QueryParam("prefix") final @Nullable String prefix, @QueryParam("tags") final @Nullable List<String> tags,
```

bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/RuleResource.java
```diff
@@ -567,4 +608,26 @@ public Response setModuleConfigParam(@PathParam("ruleUID") @Parameter(descriptio
             return null;
         }
     }
+
+    private void resetStaticListLastModified() {
+        staticListLastModified = null;
```
J-N-K, line 613:
```suggestion
        cacheableListLastModified = null;
```

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/thing/ThingResource.java
```diff
@@ -291,13 +307,34 @@ public Response create(
...
+                    return responseBuilder.build();
+                }
+            } else {
+                cacheableListLastModified = Date.from(Instant.now().truncatedTo(ChronoUnit.SECONDS));
+            }
+
+            CacheControl cc = new CacheControl();
+            cc.setMustRevalidate(true);
+            cc.setPrivate(true);
+            thingStream = dtoMapper.limitToFields(thingStream, "UID,label,bridgeUID,thingTypeUID,location,editable");
```
florian-h05, line 333: @ghys @J-N-K Is is possible that channels change without a notification to the registry? 

If not, we should add channels here to fix https://github.com/openhab/openhab-webui/issues/1996 - the alternative would be to not use the cached list there.

florian-h05: @J-N-K Sorry for pinging again, but I think you can answer my question abovey

J-N-K: Technically it might be possible, but I would consider it a bug in the binding. You might be able to edit the thing, but if you do it right (using `ThingBuilder thingBuilder = editThing();` and `updateThing(thingBuilder.build);`) you'll get a notification about an updated thing.

ghys: > @ghys @J-N-K Is is possible that channels change without a notification to the registry? 
> 
> If not, we should add channels here to fix https://github.com/openhab/openhab-webui/issues/1996 - the alternative would be to not use the cached list there.

it's IMO way better to just change the channeltype-picker in the UI to not use cached lists. There is no need to add channels (a huge chunk of the response payload, some things have dozens if not hundreds of channels), they're not used apart from that picker and there is only one instance of them in the entire UI.

PR #3342, opened 2023-01-28
Fix AbstractScriptFileWatcherTest

J-N-K:
The test only succeeded due to bad timing. This is a regression from #3255.


cweitkamp (approved):
LGTM. Thank you.

PR #3345, opened 2023-01-30
Add log SSE endpoint

ghys:
This adds a `/rest/events/log` SSE endpoint which will stream the Pax logs, it opens the possibility of having a developer tool in the main UI to show and "follow" the last logs.

The endpoint is restricted to the admin role, so a polyfill to `EventSource` should be used when using in a browser (since the native EventSource doesn't support adding headers).
An optional `loggers` parameters is available to filter by logger name, it uses the same syntax as the `topics` parameter in the `/rest/events` endpoint.

Example:
```
$ curl -H 'Authorization: Bearer oh.logtest.GYy6K780...' http://localhost:8080/rest/events/log
event: message
data: {"timestamp":"Jan 30, 2023, 12:52:58 PM","loggerName":"org.openhab.core.model.core.internal.ModelRepositoryImpl","level":"INFO","message":"Loading model \u0027demo.rules\u0027"}

event: message
data: {"timestamp":"Jan 30, 2023, 12:53:05 PM","loggerName":"org.openhab.core.automation.internal.RuleEngineImpl","level":"INFO","message":"Rule engine started."}

event: message
data: {"timestamp":"Jan 30, 2023, 12:53:05 PM","loggerName":"openhab.event.ItemStateChangedEvent","level":"INFO","message":"Item \u0027DemoLocation\u0027 changed from NULL to 52,9"}

event: message
data: {"timestamp":"Jan 30, 2023, 12:53:05 PM","loggerName":"openhab.event.ItemStateChangedEvent","level":"INFO","message":"Item \u0027DemoContact\u0027 changed from NULL to OPEN"}

event: message
data: {"timestamp":"Jan 30, 2023, 12:53:05 PM","loggerName":"openhab.event.ItemStateChangedEvent","level":"INFO","message":"Item \u0027DemoString\u0027 changed from NULL to Hello SmartHome!"}

event: message
data: {"timestamp":"Jan 30, 2023, 12:53:05 PM","loggerName":"openhab.event.ItemStateChangedEvent","level":"INFO","message":"Item \u0027DemoDateTime\u0027 changed from NULL to 2023-01-30T12:53:05.096890200+0100"}

event: message
data: {"timestamp":"Jan 30, 2023, 12:53:05 PM","loggerName":"openhab.event.ItemStateChangedEvent","level":"INFO","message":"Item \u0027DemoNumber\u0027 changed from NULL to 12.34"}

...
```

### DIFF
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -52,9 +52,9 @@
       <!--<scope>compile</scope> -->
     </dependency>
     <dependency>
-        <groupId>org.ops4j.pax.logging</groupId>
-        <artifactId>pax-logging-api</artifactId>
-        <version>${pax.logging.version}</version>
+      <groupId>org.ops4j.pax.logging</groupId>
+      <artifactId>pax-logging-api</artifactId>
+      <version>${pax.logging.version}</version>
     </dependency>
 
     <!-- Annotation API -->
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -20,6 +20,7 @@
     <californium.version>2.7.4</californium.version>
     <jetty.version>9.4.46.v20220331</jetty.version>
     <swagger.version>2.1.9</swagger.version>
+    <pax.logging.version>2.0.16</pax.logging.version>
   </properties>
 
   <dependencies>
@@ -49,6 +50,11 @@
       <artifactId>org.eclipse.jdt.annotation</artifactId>
       <version>2.2.600</version>
       <!--<scope>compile</scope> -->
+    </dependency>
+    <dependency>
+        <groupId>org.ops4j.pax.logging</groupId>
+        <artifactId>pax-logging-api</artifactId>
+        <version>${pax.logging.version}</version>
     </dependency>
 
     <!-- Annotation API -->
--- a/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/SseResource.java
+++ b/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/SseResource.java
@@ -240,7 +240,10 @@ public class SseResource implements RESTResource, SsePublisher, PaxAppender {
     }
 
     @Override
-    public void doAppend(PaxLoggingEvent loggingEvent) {
+    public void doAppend(@Nullable PaxLoggingEvent loggingEvent) {
+        if (loggingEvent == null) {
+            return;
+        }
         LogEntryDTO logEntry = new LogEntryDTO(loggingEvent);
         if (lastLogEntries.size() >= MAX_LOG_HISTORY) {
             lastLogEntries.remove();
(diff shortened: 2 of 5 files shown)


J-N-K:
In general looks like a nice addition. I have two questions:

- Wouldn't it make sense to keep the timestamp as UNIX time? It seems to me that this is much easier to consume than parsing a string.
- Why is the test failing it never failed for me in the past.

ghys:
> Wouldn't it make sense to keep the timestamp as UNIX time? It seems to me that this is much easier to consume than parsing a string.

It does but when viewing the raw messages (like with curl above) without parsing then it'll become less evident for the human eye. If that format is like the above, then it's no problem in JS to parse it:
```
> new Date("Jan 30, 2023, 12:52:58 PM")
Mon Jan 30 2023 12:52:58 GMT+0100 (Central European Standard Time)
```

> Why is the test failing it never failed for me in the past.

I have no clue actually. Jenkins doesn't actually fails but reports the build as "unstable" (https://ci.openhab.org/job/PR-openHAB-Core/5352/console). The GitHub builds fails with:
```
Error:  Errors: 
Error:    JavaTestTest.interceptedLoggerShouldLogAboveMinLevel:84 » ClassCast class org.ops4j.pax.logging.slf4j.Slf4jLogger cannot be cast to class ch.qos.logback.classic.Logger (org.ops4j.pax.logging.slf4j.Slf4jLogger and ch.qos.logback.classic.Logger are in unnamed module of loader 'app')
Error:    JavaTestTest.interceptedLoggerShouldNotLogBelowAboveMinLevel:71 » ClassCast class org.ops4j.pax.logging.slf4j.Slf4jLogger cannot be cast to class ch.qos.logback.classic.Logger (org.ops4j.pax.logging.slf4j.Slf4jLogger and ch.qos.logback.classic.Logger are in unnamed module of loader 'app')
[INFO] 
Error:  Tests run: 5, Failures: 0, Errors: 2, Skipped: 0
```
I assume it's due to the addition of `pax-logging-api` to the compile dependencies but I'm not sure how to solve it.

J-N-K:
You could try to add the pax-logging dependency to the `openhab.core.io.rest.sse` bundle instead of the compile-bom. 

But IMO it would be better to inject `org.osgi.service.log.LogReaderService` into the component and then implement the `org.osgi.service.log.LogListener` interface. That makes it independent of the actual implementation used (also I doubt that we'll move way from pax logging).

ghys:
> You could try to add the pax-logging dependency to the openhab.core.io.rest.sse bundle instead of the compile-bom.

I was under the impression that direct dependencies in individual bundles were discouraged/an anti-pattern in core and it was better to add everything to the BOM. But in `pax-logging-api`'s case it might be different since we are de facto dependent on it anyways.

> But IMO it would be better to inject org.osgi.service.log.LogReaderService into the component and then implement the org.osgi.service.log.LogListener interface. That makes it independent of the actual implementation used (also I doubt that we'll move way from pax logging).

I'll try that.

kaikreuzer:
@ghys What's the status here?

ghys:
While the no-frills SSE read-only scheme was a good candidate for logs, I think we are right to move towards WebSockets, and I completely support https://github.com/openhab/openhab-core/pull/3859 instead of a SSE endpoint - SSE is largely abandoned by browsers, don't support sending headers natively (which is important for info that needs to be authorized like logs, there are polyfills that can do it but still).

This is a first step, then we can at last talk about a single WebSocket that would handle everything.

PR #3348, opened 2023-02-02
Fix wrong .historicState call in .changedSince

J-N-K:
When calculating `.changedSince` the "then" state is queried via the `.historicState` method. Regardless of the `serviceId` used when calling `.changedSince` the historic state was always queried from the default persistence service.


cweitkamp (approved):
LGTM. Thanks.

jlaur:
@openhab/core-maintainers, would this qualify to be cherry-picked into 3.4.x?

PR #3352, opened 2023-02-04
Actions: Add method overloads for `PercentType` to Audio & Voice

### LINKED ISSUE #3331: [actions] Accept float for Audio and Voice actions in addition to PercentType
Coming from https://github.com/openhab/openhab-js/issues/98#issuecomment-1367278509, I wonder if the Audio and Voice actions could be extended to also accept float for the volume in addition to the PercentType.

This addition would be useful to all scripting addons, because script engines can convert e.g. JS numbers to float, but can’t convert a JS number to an openHAB PercentType. In theory, it would be possible to wrap those actions inside the scripting libraries, but then every script language would need to wrap. 

IMO it would be instead better to enable core to accept float as well for all Audio and Voice actions. For some it is already implemented:

https://github.com/openhab/openhab-core/blob/673827733ae870f13937e58ccd87ab04c02673ec/bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/actions/Audio.java#L105-L111

@J-N-K What’s your opinion here? I can implement this myself, but I’d like to get some feedback here so that I don’t start working on something that won’t get accepted.

florian-h05:
Closes #3331.
Related to https://github.com/openhab/openhab-js/issues/98.

This PR adds method overloads for Audio and Voice actions to accept float as volume in addition to `PercentType`.
This is especially useful when accessing these actions from the JSR223 languages because you do not need to import and construct a `PercentType` then.

I also aligned the params for the methods to increase readability of code and corrected the mathematical notation of intervals in strings (correct is `;` instead of `,`).

### DIFF
--- a/bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/actions/Audio.java
+++ b/bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/actions/Audio.java
@@ -128,7 +128,7 @@ public class Audio {
     @ActionDoc(text = "increases the master volume")
     public static void increaseMasterVolume(@ParamDoc(name = "percent") final float percent) throws IOException {
         if (percent <= 0 || percent > 100) {
-            throw new IllegalArgumentException("Percent must be in the range (0,100]!");
+            throw new IllegalArgumentException("Percent must be in the range (0;100]!");
         }
         Float volume = getMasterVolume();
         if (volume == 0) {
@@ -150,7 +150,7 @@ public class Audio {
     @ActionDoc(text = "decreases the master volume")
     public static void decreaseMasterVolume(@ParamDoc(name = "percent") final float percent) throws IOException {
         if (percent <= 0 || percent > 100) {
-            throw new IllegalArgumentException("Percent must be in the range (0,100]!");
+            throw new IllegalArgumentException("Percent must be in the range (0;100]!");
         }
         float volume = getMasterVolume();
         float newVolume = volume * (1f - percent / 100f);
--- a/bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/actions/Audio.java
+++ b/bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/actions/Audio.java
@@ -47,7 +47,7 @@ public class Audio {
 
     @ActionDoc(text = "plays a sound with the given volume from the sounds folder to the default sink")
     public static void playSound(@ParamDoc(name = "filename", text = "the filename with extension") String filename,
-            @ParamDoc(name = "volume", text = "the volume to be used") PercentType volume) {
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/actions/Audio.java
```diff
@@ -47,17 +47,23 @@ public static void playSound(@ParamDoc(name = "filename", text = "the filename w
 
     @ActionDoc(text = "plays a sound with the given volume from the sounds folder to the default sink")
     public static void playSound(@ParamDoc(name = "filename", text = "the filename with extension") String filename,
-            @ParamDoc(name = "volume", text = "the volume to be used") PercentType volume) {
+                                 @ParamDoc(name = "volume", text = "the volume to be used") PercentType volume) {
```
J-N-K, line 50: What happened to the formatting here?

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/actions/Audio.java
```diff
@@ -96,18 +109,15 @@ public static synchronized void playStream(@ParamDoc(name = "sink", text = "the
...
     }
 
     @ActionDoc(text = "sets the master volume")
     public static void setMasterVolume(
-            @ParamDoc(name = "volume", text = "volume in the range [0,1]") final float volume) throws IOException {
-        if (volume < 0 || volume > 1) {
-            throw new IllegalArgumentException("Volume value must be in the range [0,1]!");
-        }
-        setMasterVolume(new PercentType(new BigDecimal(volume * 100f)));
+            @ParamDoc(name = "volume", text = "volume in the range [0;1]") float volume) throws IOException {
```
J-N-K, line 119: Can you re-insert the `final` for consistency here?

J-N-K (changes requested):
In general LGTM, the comment regarding the formatting applies to different locations.

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/actions/Audio.java
florian-h05: I aligned all param docs for better readability, I can revert if you want.

J-N-K: no, fine then.

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/actions/Audio.java
florian-h05: Done.

J-N-K (approved):
Thanks!

PR #3355, opened 2023-02-05
Add support for things with generic channels

J-N-K:
Some bindings (like MQTT, HTTP and some 3rd party code) provide "generic channels", i.e. channels which can be used to provide content to a certain item-type. The channels share the code for converting to a `Command` or `State` and also code for pre-processing the raw content. Since it does not make sense to have the same code in different places, this should be handled in core.

Also necessary for https://github.com/openhab/openhab-addons/pull/14345.

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/generic/converter/AbstractTransformingChannelHandler.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/generic/converter/AbstractTransformingChannelHandler.java
@@ -1,0 +1,115 @@
+/**
+ * Copyright (c) 2010-2023 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.thing.binding.generic.converter;
+
+import java.util.Optional;
+import java.util.function.Consumer;
+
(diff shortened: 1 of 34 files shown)


bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/generic/converter/AbstractTransformingChannelHandler.java
```diff
@@ -0,0 +1,115 @@
...
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.thing.binding.generic.converter;
```
kaikreuzer, line 13: I assume this class is not meant to be publically exposed - should it be moved to an `internal` package?

kaikreuzer (commented):
Lgtm, thank you! 
Just one small question before merging.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/generic/converter/AbstractTransformingChannelHandler.java
J-N-K: Done. Since all channel types are covered, I don't think that we need the consumer to extend this class.

kaikreuzer (changes requested):
Imho, the correct internal package would be
```
org.openhab.core.thing.internal.binding.generic.converter
```
instead of
```
org.openhab.core.thing.binding.internal.generic.converter
```
since the bundle name is `org.openhab.core.thing`.

J-N-K:
done

kaikreuzer (approved):
Thanks!

PR #3363, opened 2023-02-11
[State/Command Metadata] Allow using '=' in a double quoted value/label

GiviMAD:

Related to UI issue https://github.com/openhab/openhab-webui/issues/1641.

The current implementation for parsing the stateDescription and commandDescription metadata value could fail if the characters = or , are used on the value or label.

This PR introduces the use of a regex to split the stateDescription and commandDescription metadata value in the same way but discarding the split characters wrapped by double quotes.

This implementation still could fail if a single " character gets introduced in a label or value, but as this have not syntactic sense I think the solution is good enough.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/MetadataCommandDescriptionProvider.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/MetadataCommandDescriptionProvider.java
@@ -63,8 +63,18 @@ public class MetadataCommandDescriptionProvider implements CommandDescriptionPro
                 if (metadata.getConfiguration().containsKey("options")) {
                     Stream.of(metadata.getConfiguration().get("options").toString().split(",")).forEach(o -> {
                         if (o.contains("=")) {
-                            commandDescription.addCommandOption(
-                                    new CommandOption(o.split("=")[0].trim(), o.split("=")[1].trim()));
+                            String command;
+                            String label;
+                            if (o.startsWith("\"")) {
+                                String[] parts = o.trim().split("\"=\"");
+                                command = removeSurroundingQuotes(parts[0]);
+                                label = removeSurroundingQuotes(parts[1]);
+                            } else {
+                                String[] parts = o.trim().split("=");
+                                command = parts[0];
+                                label = parts[1];
+                            }
+                            commandDescription.addCommandOption(new CommandOption(command.trim(), label.trim()));
                         } else {
                             commandDescription.addCommandOption(new CommandOption(o.trim(), null));
                         }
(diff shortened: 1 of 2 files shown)


J-N-K:
Instead of using this complex logic, wouldn't it be sufficient to use `.indexOf("=")` and split at that point?

GiviMAD:
> Instead of using this complex logic, wouldn't it be sufficient to use `.indexOf("=")` and split at that point?

Hi @J-N-K, I do not understand your suggestion, can you elaborate it?

The problem is that currently values without quotes are also allowed. This change tries to make the following things equivalent:
A=B
"A"=B
A="B"
"A"="B"

That way if you have a label/value that includes an equal character you can submit it between quotes.

Are you suggesting doing looking for '.indexOf("\"=\"")' before looking for '.indexOf("=")'? That will just allow the following combinations. 
A=B
"A"="B"

For me it will be enough, also solves the problem and the code will be less complex as you said.

Just want to be sure if I have understood it correctly.

J-N-K:
Why do we need the quotes at all?

GiviMAD:
@J-N-K , It's just a solution to allow using the '=' character inside the label or value of the item state/command metadata.

Could be other solutions, like allow passing the value with a scape character and then do a replace from '\=' to '=', let me know what you think is the best solution for these, I don't have any preference.

GiviMAD:
To give more context (but it's already on the issue https://github.com/openhab/openhab-webui/issues/1641 because I though this was an UI issue at first):

I faced this when I tried to use a url with querystring as the value for the command description of a String item. I put some like this in the UI:
```
http://localhost:3000/something?apiKey=sdasdasdasdasdasdasd=SomeOpenTvChannelUrl
```

The command description was incorrectly generated and gets splitter by the equal character of the query param. So the label end up been "sdasdasdasdasdasdasd=SomeOpenTvChannelUrl" and the value "http://localhost:3000/something?apiKey".

So I though, should be a way to delimite the value, and tried to changed it to this:
```
"http://localhost:3000/something?apiKey=sdasdasdasdasdasdasd"=SomeOpenTvChannelUrl
```
But it didn't changed nothing. 

So it thing it could be a good idea to allow quoted values to cover this case.

GiviMAD:
@J-N-K, but after this chatting I think that what I commented here is best best option and avoid using the regex:

```
Are you suggesting looking for '.indexOf(""="")' before looking for '.indexOf("=")'? That will just allow the following combinations.
A=B
"A"="B"
```

So detecting if both are using quotes, split in that index, and remove the surrounding quotes.

Let me know how you see it.

J-N-K:
Yes, either uses quotes on both or never. You can even use 

```
if (something.startswith("\"")) {
  // strip quotes at endif beginning and split on "="
} else {
  // split on =
}
```

GiviMAD:
Done, thanks for the suggestion.

J-N-K (approved):
Thanks!

PR #3377, opened 2023-02-16
Change dto generation to generate into source folder

kaikreuzer:
This PR makes two small changes to the DTO generation through JAXB from https://github.com/openhab/openhab-core/pull/3330:

- instead of generating only to the target folder, a `src/generated/java` source folder is introduced to which the java files are generated. This makes the bundle compile without errors within the Eclipse IDE.
- The xsd folder is moved to the bundle root, so that those files are not packaged into the final jar (we do not have the other xsds packaged either, so this aligns that).

@J-N-K There is still one problem with the generation, though: It produces java files in packages `https.openhab_org.schemas.config_description.v1_0` and `https.openhab_org.schemas.thing_description.v1_0`, since the update-description references those schemas. The generated files do not seem to be necessary at all. I didn't yet figure out how to configure the plugin so that it stops generating those files. Maybe you have an idea?


J-N-K:
I have never used them, but I guess episodes (https://github.com/highsource/maven-jaxb2-plugin/wiki/Using-Episodes) should be what you are looking for. Unfortunately most of the documentation for that at Oracle's site seems to be no longer available (or I didn't find the new place).

Side-note 1: If that works, we could also think about migrating the other parsing step-by-step - that would solve the issue that XSD and implementation sometimes differ.

Side-note 2: We still need the `thingTypeVersion` property. Even though I excluded unmanaged things, newly created manage things (which follow the correct new schema) will also be upgraded. Even though I don't think that hurts, it's unnecessary.

J-N-K:
I gave it a try and it looks like it's an easy thing to do. The only drawback is that we generate classes in other bundles that we do not need ATM (but it'll make switching to JAXB from xstream later a very easy thing as the classes are already there and we only need to adjust the bundle readers).

J-N-K:
After thinking a while about it: I would suggest two things:

- Move the update-XML reading to the `org.openhab.core.thing.xml` bundle, rename the `ThingUpdateInstructionReader` to `XMLThingUpdateInstructionReader`, make it a `Component` that implements a new interface `ThingUpdateInstructionProvider` and inject it to the `ThingManagerImpl`. That keeps the `org.openhab.core.thing` bundle free of any XML related stuff. 
- Add a new `jaxb-profile.marker` (similar to the already existing `model-profile.marker` and move the generating stuff to a profile in the root POM. That way we don't have to duplicate it over and over again.

If you are busy, just let me know, I can prepare that.

kaikreuzer:
> Why is the XML bundle there in the first place?

Well, it was separated from the start to keep the code modular - if systems would have liked to do thing type descriptions without XML, they could have left out this bundle. With ESH being reintegrated into openHAB, this has become a rather theoretical option and reviewing your PR, I was already thinking that we should maybe merge the xml bundles to the main/core bundles. Which would mean that it makes your suggestion (1) obsolete. If you feel there is still value to keep this separation, I'm also fine keeping it.

> If you are busy, just let me know, I can prepare that.

I indeed won't have time for this the next days, so feel free to take over this PR!

PR #3398, opened 2023-02-24
add sitemap Input type

mherwege:
So far, it has not been possible to have text or numeric input in a sitemap. This can be useful to do simple updating of items, for instance for manual meter readings.
The workaround so far has been to use a Webview element and some javascript, see forum: https://community.openhab.org/t/input-field-for-number-free-text-for-openhab-uis/12461.

This PR adds an sitemap input widget. A second PR will add this to Basic UI.

If these PR's would be accepted, I will also update the main UI sitemap configuration pages to include the input widget. I have not done any app development (Android, IOS) so far, but would hope this is interesting enough for someone to also consider adding it to the apps.

Some configuration options could definitely be added in follow-up PR's for the input field, but my testing shows it is already working well in Basic UI for String, Number and QuantityType items.

### DIFF
--- a/bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/Sitemap.xtext
+++ b/bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/Sitemap.xtext
@@ -15,7 +15,7 @@ Widget:
     (LinkableWidget | NonLinkableWidget);
 
 NonLinkableWidget:
-    Switch | Selection | Slider | List | Setpoint | Video | Chart | Webview | Colorpicker | Mapview | Default;
+    Switch | Selection | Slider | List | Setpoint | Video | Chart | Webview | Colorpicker | Mapview | Input | Default;
 
 LinkableWidget:
     (Text | Group | Image | Frame)
@@ -118,6 +118,12 @@ Setpoint:
 Colorpicker:
     'Colorpicker' (('item=' item=ItemRef) & ('label=' label=(ID | STRING))? & ('icon=' icon=Icon)? & ('sendFrequency='
     frequency=INT)? &
+    ('labelcolor=[' (LabelColor+=ColorArray (',' LabelColor+=ColorArray)* ']'))? &
+    ('valuecolor=[' (ValueColor+=ColorArray (',' ValueColor+=ColorArray)* ']'))? &
+    ('visibility=[' (Visibility+=VisibilityRule (',' Visibility+=VisibilityRule)* ']'))?);
+
+Input:
+    'Input' (('item=' item=ItemRef) & ('label=' label=(ID | STRING))? & ('icon=' icon=Icon)? &
     ('labelcolor=[' (LabelColor+=ColorArray (',' LabelColor+=ColorArray)* ']'))? &
     ('valuecolor=[' (ValueColor+=ColorArray (',' ValueColor+=ColorArray)* ']'))? &
     ('visibility=[' (Visibility+=VisibilityRule (',' Visibility+=VisibilityRule)* ']'))?);


lolodomo:
I am a little surprised that nothing else than the sitemap syntax needs to be updated. 
But if that is really the case, that us great.

mherwege:
> I am a little surprised that nothing else than the sitemap syntax needs to be updated.

I was surprised as well. But all the required code is automatically generated from the syntax, and that just works. There is more additional code for Basic UI in the webui repository, but still not a lot.

lolodomo:
An update of the documentation will of course be required.

mherwege:
> 

Of course.
Looking at the documentation, I can do the minimal change, or go through the sitemap documentation and do a more thourough update. E.g. documentation for the chart element talks about begin and end parameters that do not exist, there are links to wiki's on archived repositories for examples, and I feel the warning about preference for main UI is maybe worded a bit too strong. I think it is irrelevant it cannot be shown inside main UI, but it still can be configured there.

lolodomo:
> I think it is irrelevant it cannot be shown inside main UI

You can access BasicUI (and other UIs) from MainUI through the "icon at top right" (don't know how it is properly described).

mherwege:
I have created a minimal documentation update, and will keep a broader update for later.

seime:
Great @mherwege !

If you could add support for DateTime input as well, I think many users will be happy as it then allows for simple scheduling of tasks (in combination with a rule).

mherwege:
> support for DateTime input

Next on my list, but will be in another PR, and not sure how to do it yet. Material Design Lite does not have a pre-defined calendar and time input widget, and the default html data/time entry is not very good. In the main UI, that could be solved with  a vue datepicker, but that's not an option here. I think it will have to be another input type to make it easier to handle in the mobile apps, where better support is available.

J-N-K (approved):
LGTM, thanks.

maniac103:
> I think it will have to be another input type to make it easier to handle in the mobile apps, where better support is available.

No reason to do that, we (as in: mobile apps people) can check the item type for choosing the correct widget.

mherwege:
> No reason to do that, we (as in: mobile apps people) can check the item type for choosing the correct widget.
That’s great. What about a parameter for date/time? In some cases, the time is the only thing to be updated (e.g. to  update a daily schedule). Other cases you may only want to update the date. So I would thing there needs to be a parameter for that. What do you think?

maniac103:
> What about a parameter for date/time? In some cases, the time is the only thing to be updated (e.g. to update a daily schedule). Other cases you may only want to update the date.

That may be a good idea. In the case of the Android app, there's no date/time picker in Material Design 3 anyway, just [date pickers](https://m3.material.io/components/date-pickers/overview) and [time pickers](https://m3.material.io/components/date-pickers/overview), but still, with an appropriate hint one could show either one, or both.
WHen introducing such a thing, my suggestion would be to keep the parameter as a generic string (`inputHint` or something like that?), so it's possible to have different types of hints for different types of input fields if needed at some later point.

J-N-K:
Since that is essentially content description of the item it should be handled in item metadata.

mueller-ma:
Should it be possible to change the unit of measurement of `Number` items, e.g. should the user be able to change `20 °C` to `30 F`? If not mobile apps could display a numeric keyboard which makes it easier to enter numbers, but makes it impossible to enter anything else, like units.

mherwege:
The implementation in Basic UI allows changing the UOM. And Basic UI does not impact keyboard layout on mobile. This might be another candidate for an inputHint value. If set to number or no UOM present in the numeric value, it would present a numeric keyboard, if not, present a default keyboard.

lolodomo:
For information
@openhab/webui-maintainers
@openhab/android-maintainers
@openhab/ios-maintainers

PR #3401, opened 2023-02-25
Allow configurable access per metadata namespace for user role

florian-h05:
Related to https://github.com/openhab/openhab-webui/pull/1725.

This:
- adds a boolean `userAccessAllowed` property to the Metadata class (defaults to `false`)
- updates the metadata create/update REST endpoint to allow the user role to update an existing metadata namespace if `userAccessAllowed` is true
- updates the tests for the added REST endpoint logic


florian-h05:
/cc @stefan-hoehn

@J-N-K Would be great to get your opinion and possibly your review here, since you are involved in the MainUI PR as well.

J-N-K:
See my comment in openhab-webui. We should keep the discussion in one place.

PR #3404, opened 2023-02-25
Prevent duplicate events in WatchService

### LINKED ISSUE #3400: Rules/Sitemaps/Items files refreshed twice when saved from Samba share
This old bug is back.

https://github.com/eclipse-archived/smarthome/issues/4824

Example:
```
2023-02-25 08:38:40.555 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model 'default.sitemap'
2023-02-25 08:38:40.558 [WARN ] [el.core.internal.ModelRepositoryImpl] - Configuration model 'default.sitemap' is either empty or cannot be parsed correctly!
2023-02-25 08:38:40.583 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model 'default.sitemap'
```

J-N-K:
Closes #3400 

On Samba shares a single MODIFY event may result in different sequences of CREATE/DELETE/MODIFY events. Unfortunately this is not properly handled by the underlying library and we have to re-introduce our ugly workaround. The solution itself is a bit improved over the old one as it not only keeps the last event but determines the correct event by examining the sequence of received events.

### DIFF
--- a/bundles/org.openhab.core.config.dispatch/src/main/java/org/openhab/core/config/dispatch/internal/ConfigDispatcherFileWatcher.java
+++ b/bundles/org.openhab.core.config.dispatch/src/main/java/org/openhab/core/config/dispatch/internal/ConfigDispatcherFileWatcher.java
@@ -53,7 +53,7 @@ public class ConfigDispatcherFileWatcher implements WatchService.WatchEventListe
 
         this.watchService = watchService;
 
-        watchService.registerListener(this, Path.of(servicesFolder));
+        watchService.registerListener(this, Path.of(servicesFolder), false);
         configDispatcher.processConfigFile(Path.of(OpenHAB.getConfigFolder(), servicesFolder).toFile());
     }
 
--- a/bundles/org.openhab.core.model.core/src/main/java/org/openhab/core/model/core/internal/folder/FolderObserver.java
+++ b/bundles/org.openhab.core.model.core/src/main/java/org/openhab/core/model/core/internal/folder/FolderObserver.java
@@ -12,7 +12,6 @@
  */
 package org.openhab.core.model.core.internal.folder;
 
-import static java.nio.file.StandardWatchEventKinds.*;
 import static org.openhab.core.service.WatchService.Kind.CREATE;
 import static org.openhab.core.service.WatchService.Kind.MODIFY;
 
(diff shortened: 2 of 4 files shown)


wborn:
> Unfortunately this is not properly handled by the underlying library and we have to re-introduce our ugly workaround.

Is this a known issue or should one be created for it? If it is fixed in the library, the ugly workaround will not be part of our code and other library users will also benefit.

J-N-K:
See linked issue. We should merge this now (even if we remove it later) because it seems to be a real-world problem for a lot of users (see e.g. https://community.openhab.org/t/openhab-4-0-snapshot-discussion/142322/189).

J-N-K:
After some more investigation we have at least a bug in the FolderObserver which might be the root cause of a lot of our issues, that's why I converted this PR to draft.

I did not find any hint if sub-directories should be allowed (i.e. `/etc/openhab/rules/sub/foo.rules`), but this only works as long as the file name is unique because `rules/sub/foo.rules` and `foo.rules` result in the same model. This is also the issue for me when editing files remotely, because TextMate creates a sub-directory  `/etc/openhab/rules/(A Document Being Saved By TextMate)` and saves a copy of the original file with the original file name in that directory `/etc/openhab/rules/(A Document Being Saved By TextMate)/foo.rules`. The `FolderObserver` resolves the copy and the original file to the original file (`/etc/openhab/rules/foo.rules`) - it simply takes the filename, determines the directory for the extension and then creates a file reference for the filename in the directory, sub-directories are discarded during that process.

@kaikreuzer, do you know if it was ever intended to work or was the original design a flat structure like `rules/*.rules` without sub-directories?

J-N-K:
@rkoshak I know that you use (mostly) UI, but can you tell if you ever came across someone successfully using sub-directories in the "classic" config directories like `items` or `rules`?

jimtng:
I wonder if this is related https://github.com/openhab/openhab-addons/issues/13922
Changes were made locally on the linux filesystem, not over samba.

J-N-K:
Not sure. If you can reproduce it, can you please set `io.methvin` to `DEBUG` and make it happen? If it is easy to reproduce, please let me know what to do (i.e. what scripts need to be added and which files I should edit).

J-N-K:
This is probably not the best solution but it effectively prevents duplicate events and ensures the correct event is passed to the listeners. The erratic behavior of the `FolderObserver` for files in sub-directories has also been fixed.

@openhab/core-maintainers I would consider this as blocking for a milestone release as the duplicate file reloading impacts performance significantly.

J-N-K:
Yes. But anyway, if it‘s easy to reproduce let me know how.

wborn (approved):
LGTM :+1:

PR #3406, opened 2023-02-26
Add Eclipse Add-on Service

wborn:
* This service provides add-on information when debugging in Eclipse so it is possible to add Things based on the installed bindings and configure installed add-ons in the UI.
* Replaces the Sample Add-on Service.
* Also shows the "Add-on Management" configuration which allows for configuring if incompatible add-ons are included.


bundles/org.openhab.core.addon.eclipse/src/main/java/org/openhab/core/addon/eclipse/internal/EclipseAddonService.java
```diff
@@ -0,0 +1,188 @@
...
+            new AddonType(ADDON_TYPE_UI, "User Interfaces"));
+
+    private static final Map<String, String> ADDON_BUNDLE_TYPE_MAP = Map.ofEntries(
+            entry(ADDON_TYPE_AUTOMATION, "automation"), //
+            entry(ADDON_TYPE_BINDING, "binding"), //
+            entry(ADDON_TYPE_MISC, "io"), //
+            entry(ADDON_TYPE_PERSISTENCE, "persistence"), //
+            entry(ADDON_TYPE_TRANSFORMATION, "transform"), //
+            entry(ADDON_TYPE_VOICE, "voice"));
+
```
J-N-K, line 74: This is duplicated in almost all add-on services. Wouldn't it make sense to provide these as "default" from `AddonType`?

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/addons/AddonResource.java
```diff
@@ -389,8 +391,13 @@ private void postFailureEvent(String addonId, @Nullable String msg) {
...
     private @Nullable AddonService getDefaultService() {
-        return addonServices.stream().filter(addonService -> DEFAULT_ADDON_SERVICE.equals(addonService.getId()))
-                .findFirst().orElse(addonServices.stream().findFirst().orElse(null));
+        Optional<AddonService> karafAddonService = addonServices.stream()
+                .filter(addonService -> KARAF_ADDON_SERVICE_ID.equals(addonService.getId())).findFirst();
+        Optional<AddonService> eclipseAddonService = addonServices.stream()
+                .filter(addonService -> ECLIPSE_ADDON_SERVICE_ID.equals(addonService.getId())).findFirst();
+        Optional<AddonService> firstAddonService = addonServices.stream().findFirst();
+        return Stream.of(karafAddonService, eclipseAddonService, firstAddonService).filter(Optional::isPresent)
+                .findFirst().get().orElse(null);
```
J-N-K, line 400: Why is this needed? If the eclipse add-on service prefixes the ids with it's own id (like the other add-on services do) it should work without this strong coupling.

wborn: Yes that was something already on my todo list and it seems to work.

Though it could probably use some UI tweaks as well:

image

bundles/org.openhab.core.addon.eclipse/src/main/java/org/openhab/core/addon/eclipse/internal/EclipseAddonService.java
wborn: Yes I don't like all the duplication either so that might be something to cleanup as well.
I first wanted to quickly check if this approach would turn into something usable. :wink:

bom/openhab-core/pom.xml
```diff
@@ -514,6 +514,12 @@
       <version>${project.version}</version>
       <scope>compile</scope>
     </dependency>
+    <dependency>
```
J-N-K, line 517: Is this still needed?

wborn: AFAIK the BOMs are required for generating the indices used by Bnd for resolving the "runbundles" in .bndrun files (itests/Eclipse).

J-N-K: For me it is sufficient to add the bundle to the [demo-app POM](https://github.com/openhab/openhab-distro/blob/main/launch/app/pom.xml), at least the resolver fails if I remove the bundle from the BOM and succeeds after I add it to the POM.

J-N-K (changes requested):
See my comment regarding the BOM, otherwise LGTM.

bom/openhab-core/pom.xml
wborn: Perhaps the index also improves performance?

Besides that the BOMs are used for managing versions and to make sure all bundles are build before feature validation starts.

J-N-K (approved):
LGTM, unfortunately I can't merge in openhab-distro

PR #3411, opened 2023-02-28
Fix overriding Jetty's User-Agent in HttpUtil

jimtng:
[request.header](https://github.com/eclipse/jetty.project/blob/4563f7a90e6ca11efeb1a48b5475e9853e6e2bcf/jetty-client/src/main/java/org/eclipse/jetty/client/HttpRequest.java#L309-L316) just _adds_ the header, resulting in multiple 'User-Agent' headers, whereas [request.agent](https://github.com/eclipse/jetty.project/blob/4563f7a90e6ca11efeb1a48b5475e9853e6e2bcf/jetty-client/src/main/java/org/eclipse/jetty/client/HttpRequest.java#L286-L290) _replaces_ it.

Looking at the implementation of headers.add, a workaround could be done (without this PR) by giving a 'User-Agent' of nil, and the second 'User-Agent' header containing the intended value. The nil entry would cause headers.add to remove existing headers of the same name.

### DIFF
--- a/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/HttpUtil.java
+++ b/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/HttpUtil.java
@@ -213,7 +213,11 @@ public class HttpUtil {
 
         if (httpHeaders != null) {
             for (String httpHeaderKey : httpHeaders.stringPropertyNames()) {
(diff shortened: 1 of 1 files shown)


jimtng:
@openhab/core-maintainers could this also be included in 3.4.x? It has been [tested to work](https://community.openhab.org/t/sendhttpgetrequest-does-not-respect-user-agent-in-header/144838/7)

wborn:
Done! :slightly_smiling_face:

PR #3418, opened 2023-03-02
Add inputHint param to sitemap Input element

mherwege:
See discussion in https://github.com/openhab/openhab-core/pull/3398.
See implementation of usage in BasicUI https://github.com/openhab/openhab-webui/pull/1787.

This is a first step of work to be done in the respective sitemap based UI's.

This PR adds a new inputHint parameter that can take a string that describes behaviour of the UI widget.
Right now, I propose to develop support for following values in the UI's:
-  `text`: default for non numeric item types, would show a standard keyboard on touch devices with on screen keyboard
-  `number`: show a numeric only keyboard on touch devices, default if item type is Numeric
-  `date`: show a date picker
-  `time`: show a time picker
-  `datetime`: show a combined date/time picker (or sequence of two pickers), default for DateTime type items

Using `number` on QuantityType items would still show a numeric keypad. Therefore the Unit cannot be changed in the input and the Unit of the current value is kept.

### DIFF
--- a/bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/Sitemap.xtext
+++ b/bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/Sitemap.xtext
@@ -129,6 +129,7 @@ Colorpicker:
 
 Input:
     'Input' (('item=' item=ItemRef) & ('label=' label=(ID | STRING))? & ('icon=' icon=Icon)? &
+    ('inputHint=' inputHint=(STRING))? &
     ('labelcolor=[' (LabelColor+=ColorArray (',' LabelColor+=ColorArray)* ']'))? &
     ('valuecolor=[' (ValueColor+=ColorArray (',' ValueColor+=ColorArray)* ']'))? &
     ('iconcolor=[' (IconColor+=ColorArray (',' IconColor+=ColorArray)* ']'))? &
--- a/bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/Sitemap.xtext
+++ b/bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/Sitemap.xtext
@@ -129,7 +129,7 @@ Colorpicker:
 
 Input:
     'Input' (('item=' item=ItemRef) & ('label=' label=(ID | STRING))? & ('icon=' icon=Icon)? &
-    ('inputHint=' inputHint=INPUT_HINT)? &
+    ('inputHint=' inputHint=STRING)? &
     ('labelcolor=[' (LabelColor+=ColorArray (',' LabelColor+=ColorArray)* ']'))? &
     ('valuecolor=[' (ValueColor+=ColorArray (',' ValueColor+=ColorArray)* ']'))? &
     ('iconcolor=[' (IconColor+=ColorArray (',' IconColor+=ColorArray)* ']'))? &
@@ -147,7 +147,7 @@ Mapping:
     cmd=Command '=' label=(ID | STRING);
 
 VisibilityRule:
-    (item=ID) (condition=("==" | ">" | "<" | ">=" | "<=" | "!=")) (sign=('-' | '+'))? (state=XState);
+    (item=ID) (condition=('==' | '>' | '<' | '>=' | '<=' | '!=')) (sign=('-' | '+'))? (state=XState);
 
 ItemRef:
     ID;
@@ -159,7 +159,7 @@ Icon returns ecore::EString:
     STRING | ID;
 
 ColorArray:
-    ((item=ID)? (condition=("==" | ">" | "<" | ">=" | "<=" | "!="))? (sign=('-' | '+'))? (state=XState) '=')?
+    ((item=ID)? (condition=('==' | '>' | '<' | '>=' | '<=' | '!='))? (sign=('-' | '+'))? (state=XState) '=')?
     (arg=STRING);
 
 Command returns ecore::EString:
@@ -170,10 +170,6 @@ Number returns ecore::EBigDecimal:
 
 XState returns ecore::EString:
     INT | ID | STRING | FLOAT;
-
-terminal INPUT_HINT returns ecore::EString:
-    ('"' ("text" | "number" | "date" | "time" | "datetime") '"') |
-    ("'" ("text" | "number" | "date" | "time" | "datetime") "'");
 
 @Override 
 terminal ID:
--- a/bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/Sitemap.xtext
+++ b/bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/Sitemap.xtext
@@ -172,8 +172,8 @@ XState returns ecore::EString:
     INT | ID | STRING | FLOAT;
 
 terminal INPUT_HINT returns ecore::EString:
-            ('"' ("text" | "number" | "date" | "time" | "datetime") '"') |
-            ("'" ("text" | "number" | "date" | "time" | "datetime") "'");
+    ('"' ("text" | "number" | "date" | "time" | "datetime") '"') |
+    ("'" ("text" | "number" | "date" | "time" | "datetime") "'");
 
 @Override 
 terminal ID:
--- a/bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/Sitemap.xtext
+++ b/bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/Sitemap.xtext
@@ -54,7 +54,7 @@ Image:
 
 Video:
     'Video' (('item=' item=ItemRef)? & ('label=' label=(ID | STRING))? & ('icon=' icon=Icon)? &
-    ('url=' url=(STRING)) & ('encoding=' encoding=(STRING))? &
+    ('url=' url=STRING) & ('encoding=' encoding=STRING)? &
     ('labelcolor=[' (LabelColor+=ColorArray (',' LabelColor+=ColorArray)* ']'))? &
     ('valuecolor=[' (ValueColor+=ColorArray (',' ValueColor+=ColorArray)* ']'))? &
     ('iconcolor=[' (IconColor+=ColorArray (',' IconColor+=ColorArray)* ']'))? &
@@ -62,7 +62,7 @@ Video:
 
 Chart:
     'Chart' (('item=' item=ItemRef) & ('label=' label=(ID | STRING))? & ('icon=' icon=Icon)? &
-    ('service=' service=(STRING))? & ('refresh=' refresh=INT)? & ('period=' period=ID) &
+    ('service=' service=STRING)? & ('refresh=' refresh=INT)? & ('period=' period=ID) &
     ('legend=' legend=BOOLEAN_OBJECT)? & ('forceasitem=' forceAsItem=BOOLEAN_OBJECT)? &
     ('labelcolor=[' (LabelColor+=ColorArray (',' LabelColor+=ColorArray)* ']'))? &
     ('valuecolor=[' (ValueColor+=ColorArray (',' ValueColor+=ColorArray)* ']'))? &
@@ -72,7 +72,7 @@ Chart:
 
 Webview:
     'Webview' (('item=' item=ItemRef)? & ('label=' label=(ID | STRING))? & ('icon=' icon=Icon)? &
-    ('height=' height=INT)? & ('url=' url=(STRING)) &
+    ('height=' height=INT)? & ('url=' url=STRING) &
     ('labelcolor=[' (LabelColor+=ColorArray (',' LabelColor+=ColorArray)* ']'))? &
     ('valuecolor=[' (ValueColor+=ColorArray (',' ValueColor+=ColorArray)* ']'))? &
     ('iconcolor=[' (IconColor+=ColorArray (',' IconColor+=ColorArray)* ']'))? &
@@ -129,7 +129,7 @@ Colorpicker:
 
 Input:
     'Input' (('item=' item=ItemRef) & ('label=' label=(ID | STRING))? & ('icon=' icon=Icon)? &
-    ('inputHint=' inputHint=(STRING))? &
+    ('inputHint=' inputHint=STRING)? &
     ('labelcolor=[' (LabelColor+=ColorArray (',' LabelColor+=ColorArray)* ']'))? &
     ('valuecolor=[' (ValueColor+=ColorArray (',' ValueColor+=ColorArray)* ']'))? &
     ('iconcolor=[' (IconColor+=ColorArray (',' IconColor+=ColorArray)* ']'))? &
@@ -171,6 +171,7 @@ Number returns ecore::EBigDecimal:
 XState returns ecore::EString:
     INT | ID | STRING | FLOAT;
 
+@Override 
 terminal ID:
     ('^'? ('a'..'z' | 'A'..'Z' | '_') ('a'..'z' | 'A'..'Z' | '_' | '0'..'9')*) |
     (('0'..'9')+ ('a'..'z' | 'A'..'Z' | '_') ('0'..'9' | 'a'..'z' | 'A'..'Z' | '_')*);
--- a/bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/Sitemap.xtext
+++ b/bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/Sitemap.xtext
@@ -129,7 +129,7 @@ Colorpicker:
 
 Input:
     'Input' (('item=' item=ItemRef) & ('label=' label=(ID | STRING))? & ('icon=' icon=Icon)? &
-    ('inputHint=' inputHint=STRING)? &
+    ('inputHint=' inputHint=INPUT_HINT)? &
     ('labelcolor=[' (LabelColor+=ColorArray (',' LabelColor+=ColorArray)* ']'))? &
     ('valuecolor=[' (ValueColor+=ColorArray (',' ValueColor+=ColorArray)* ']'))? &
     ('iconcolor=[' (IconColor+=ColorArray (',' IconColor+=ColorArray)* ']'))? &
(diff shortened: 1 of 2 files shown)


lolodomo:
For information
@openhab/webui-maintainers
@openhab/android-maintainers
@openhab/ios-maintainers

bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/Sitemap.xtext
```diff
@@ -137,6 +137,7 @@ Colorpicker:
 
 Input:
     'Input' (('item=' item=ItemRef) & ('label=' label=(ID | STRING))? & ('icon=' icon=Icon)? &
+    ('inputHint=' inputHint=(STRING))? &
```
mueller-ma, line 140: Maybe limit the possible values in the Sitemap syntax? Otherwise syntax issues aren't visible in the server log.

mherwege: This is possible, but also not done for any of the other parameters. I would like to see some broader opinion on this before making an exception for this parameter.

lolodomo: Good remark. For what other existing parameters should have been expected?

mherwege: In general parser only checks if any parameter is an int, ID, string, float or boolean. The parser itself does not check if e.g. an item is valid, although that would show up in other places when trying to update the non existing item from a sitemap UI.
Likewise, there is no check if icons are valid, colors are valid.
The chart period parameter has a limited set of allowed values, no check in the parser.
The video encoding parameter as a limited list of options.
There is no validation for the url parameters.

All of these will fail at runtime, Not when saving the sitemap.

lolodomo: Is the parenthesis around STRING required ?
There are sometimes there, sometimes not.
For example the url attribute has no parenthesis in the Image element but the url attribute has parenthesis in the Video element.
The two syntax are identical or is there a real difference ?

mherwege: > Is the parenthesis around STRING required ?
> There are sometimes there, sometimes not.
> For example the url attribute has no parenthesis in the Image element but the url attribute has parenthesis in the Video element.
> The two syntax are identical or is there a real difference ?

I don’t think the parentheses are required, but didn’t test.

lolodomo: Please also note that there are some controls added outside the .xtext file. They are defined in package "validation".

mherwege: > For example the url attribute has no parenthesis in the Image element but the url attribute has parenthesis in the Video element.

I removed the extra parentheses from in a few places in the file.

When doing that, I noticed that the url parameter for the Image and Video parameter are different. It is a required parameter for the Video, but not for the Image. Is that correct? There is no difference in the documentation.

mueller-ma: The image can be set as:
* base64 encoded item state
* URL

The video must be provided by URL.

lolodomo: Here is what is written in the documentation:

> `url` is the default URL from which to retrieve the image, if there is no associated Item or if the associated item's state is not a URL.

In practice, BasicUI is properly handling the case url is not set or not set to a valid URL.
@mueller-ma : can you confirm that it is also handled in the Android App ?

So yes, `url` is optional for image while mandatory for video.

Maybe the documentation could be slightly updated.

lolodomo: Thinking more about it, I really think the syntax should consider/check the allowed values when it is clear that only few values are accepted. It could be done either directly in the xtext file or in "validation" class.
It could help users to identify some mistakes in their sitemaps (because they used an unexpected value).

I am not asking to fix all the existing syntax in that PR but I would like we do it in a next PR.
Maybe we can at least start with the `inputHint`. Are you ok @mherwege ?

maniac103: > can you confirm that it is also handled in the Android App ?

Yes, see [here](https://github.com/openhab/openhab-android/blob/main/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt#L676-682).

mueller-ma: The Android app checks:
1. is widget state a base64 encoded image => decode and use it
2. is widget.url set => use it
3. else => don't display an image

https://github.com/openhab/openhab-android/blob/main/mobile/src/main/java/org/openhab/habdroid/ui/WidgetAdapter.kt#L676-L683

bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/Sitemap.xtext
```diff
@@ -170,6 +171,7 @@ Number returns ecore::EBigDecimal:
 XState returns ecore::EString:
     INT | ID | STRING | FLOAT;
 
+@Override
```
lolodomo, line 174: As I am terribly fighting with the update of the xtext definition for icons, I am curious to understand why this `@override` is suddenly required. I need to learn about XText.

bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/Sitemap.xtext
lolodomo: Thank @mueller-ma  and @maniac103 , it confirms `url` is really optional for image.

bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/Sitemap.xtext
mherwege: It is not strictly required, but it did give a warning in the Eclipse IDE if it was not there. iD is defined by default for xtext, and his definition definitely overrides that default definition.

mherwege: Note that the grammar also gives a warning in the IDE for the Image item, as there is no required parameter (both url and item are optional), it complains about potentially not instantiating an object. So that grammar may have to be refined well.

bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/Sitemap.xtext
lolodomo: Yes, I believe it is very easy to do it directly in the grammar. Look at BOOLEAN_OBJECT rule as an example.

mherwege: I added a validation in the grammar. I got this in the log when using an illegal value in a sitemap, and the sitemap was not accepted:
```
Configuration model 'thuis.sitemap' has errors, therefore ignoring it: [13,67]: mismatched input '"test"' expecting RULE_INPUT_HINT
```

lolodomo: Looks good.

bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/Sitemap.xtext
```diff
@@ -170,6 +171,11 @@ Number returns ecore::EBigDecimal:
 XState returns ecore::EString:
     INT | ID | STRING | FLOAT;
 
+terminal INPUT_HINT returns ecore::EString:
+            '"' ("text" | "number" | "date" | "time" | "datetime") '"' |
+            "'" ("text" | "number" | "date" | "time" | "datetime") "'";
```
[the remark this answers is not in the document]
lolodomo: Can we also ask a third alternative without quotes ?
```
terminal INPUT_HINT returns ecore::EString:
            ('"' ("text" | "number" | "date" | "time" | "datetime") '"') |
            ("'" ("text" | "number" | "date" | "time" | "datetime") "'") |
            ("text" | "number" | "date" | "time" | "datetime");
```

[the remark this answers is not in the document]
mherwege: > Wouldn't it be more secure to add parenthesis?
> 
> ```
> terminal INPUT_HINT returns ecore::EString:
>             ('"' ("text" | "number" | "date" | "time" | "datetime") '"') |
>             ("'" ("text" | "number" | "date" | "time" | "datetime") "'");
> ```

Would make it clearer for the reader, but is not required. I will change it. This is how the default terminal STRING is defined:

```
terminal STRING:
    '"' ( '\\' . /* 'b'|'t'|'n'|'f'|'r'|'u'|'"'|"'"|'\\' */ | !('\\'|'"') )* '"' |
    "'" ( '\\' . /* 'b'|'t'|'n'|'f'|'r'|'u'|'"'|"'"|'\\' */ | !('\\'|"'") )* "'"
```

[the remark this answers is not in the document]
mherwege: > Can we also add a third alternative without quotes ?
> 
> ```
> terminal INPUT_HINT returns ecore::EString:
>             ('"' ("text" | "number" | "date" | "time" | "datetime") '"') |
>             ("'" ("text" | "number" | "date" | "time" | "datetime") "'") |
>             ("text" | "number" | "date" | "time" | "datetime");
> ```

I am reluctant to do this. The period parameter for the Chart widget that cannot take quotes around it caused a lot of trouble in the mainUI sitemap parser. The JSON representation has quotes around it. It required special handling in a few places, see https://github.com/openhab/openhab-webui/pull/1745 and following PR's https://github.com/openhab/openhab-webui/pull/1781 and https://github.com/openhab/openhab-webui/pull/1783. So I would like to avoid it.
I actually think it was bad practice to mix this in the grammar in the first place. Booleans should be true/false without quotes, numbers should be without quotes and any other parameter should have quotes. The item parameter is OK as an exception. But at this point, I don't think we should change that, just not introduce new confusing ones.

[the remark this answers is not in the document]
lolodomo: Ok with your arguments.

mherwege:
Please wait for merging. There still seems to be something not quite right with quotes around the inputHint parameter.

mherwege:
@lolodomo Can you have another look? The previous validation of the inputHint parameter did not work as expected. The match with the STRING terminal sometimes took precedence. I needed to move the check in a custom validator to make things work. It should provide a similar message in the log, but will not stop the sitemap from loading.

mherwege:
@lolodomo I add the draft label yesterday to avoid merging, but cannot remove it anymore. I lick permissions. The PR is in a good state again.

bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/validation/SitemapValidator.xtend
```diff
@@ -102,4 +107,14 @@ class SitemapValidator extends AbstractSitemapValidator {
                 SitemapPackage.Literals.SETPOINT.getEStructuralFeature(SitemapPackage.SETPOINT__MIN_VALUE));
         }
     }
+    
+    @Check
+    def void checkInputHintParameter(Input i) {
+        if (i.inputHint !== null && !ALLOWED_HINTS.contains(i.inputHint)) {
+            val node = NodeModelUtils.getNode(i)
+            val line = node.getStartLine()
+            error("Input on item '" + i.item + "' has invalid inputHint '" + i.inputHint + "' at line " + line,
```
lolodomo, line 116: Did you test what happens if the item is not defined ?

mherwege: It will not even run this code and fail earlier, refusing to load the sitemap. That's because it is already an error in the grammar xtext.

mherwege:
> Does it mean that we have to use STRING and a custom validator as soon as we want to restrict values ?

It looks like it. I needed a terminal at the end and these cannot overlap, or one will hide the other.

mherwege:
> Is it now well tested ?

Better than before for sure. I have been working on the BasicUI implementation, and so far what I get from the inputHint is as expected.

mherwege:
> Sorry, I have no particular permission on this repo.
> And as I have never used the "draft" feature, I can't tell you how to remove it.

As long as it can review and merged when everything is OK, I don't care.

bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/validation/SitemapValidator.xtend
lolodomo: Sorry, I did not see that you made the item mandatory.

lolodomo:
> > Sorry, I have no particular permission on this repo.
> > And as I have never used the "draft" feature, I can't tell you how to remove it.
> 
> As long as it can review and merged when everything is OK, I don't care.

Normally, in front of "This pull request is still a work in progress", you should find a button named "Ready for review". I think this will remove the "draft".

mherwege:
See the https://github.com/openhab/openhab-webui/pull/1787 implementation for BasicUI to test this.

J-N-K:
@lolodomo Did you test this? Is this complete or are there extensions needed to be used in UI? I would like to prevent having tons of follow-up PR just because it was not complete.

lolodomo:
No, not tested by myself and to be very honest with you, I am not sure when this code is used) run. But looking at the code of the existing method, the change makes sense. 
I let @mherwege as author of the PR comment.

mherwege:
@J-N-K I tested this with various inputHints, correct and wrong. The wrong ones did log a validation error, as expected.
I used sitemaps with inputHint parameters in a further development of BasicUI, see https://github.com/openhab/openhab-webui/pull/1787.

This is part of a larger set of PR’s to support input elements in sitemap based UI’s. I am only adding the functionality to BasicUI myself at the moment. Here is the full set:

Basic Input element support (only text input element):
- https://github.com/openhab/openhab-core/pull/3398
- https://github.com/openhab/openhab-webui/pull/1729
- https://github.com/openhab/openhab-core/pull/3431 (config through main UI)
- https://github.com/openhab/openhab-webui/pull/1735 (config through main UI)

The second series of PR’s extends that to more advanced support with an inputHint parameter to be used to suggest the sitemap based UI to use a specific input widget:
- https://github.com/openhab/openhab-core/pull/3418 (this PR)
- https://github.com/openhab/openhab-webui/pull/1787
- PRs for supporting configuration in main UI (in openhab-core and openhab-webui repositories) to be created, waiting on acceptance of https://github.com/openhab/openhab-core/pull/3431 and this PR

I don’t have plans to further extend this beyond what is proposed here. I do believe there is interest to also use this in the Android app.

bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/Sitemap.xtext
```diff
@@ -146,7 +147,7 @@ Mapping:
     cmd=Command '=' label=(ID | STRING);
 
 VisibilityRule:
-    (item=ID) (condition=("==" | ">" | "<" | ">=" | "<=" | "!=")) (sign=('-' | '+'))? (state=XState);
+    (item=ID) (condition=('==' | '>' | '<' | '>=' | '<=' | '!=')) (sign=('-' | '+'))? (state=XState);
```
J-N-K, line 150: Why did you change this?

J-N-K (commented):
Otherwise LGTM.

bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/Sitemap.xtext
mherwege: Purely for consistency, so cleanup. Xtext will allow single or double quotes. All other matches in the file use single quotes. There is no specific reason for the functionality of this PR.

lolodomo: Agree with @mherwege

J-N-K (approved):

lolodomo:
> No, not tested by myself and to be very honest with you, I am not sure when this code is used) run. But looking at the code of the existing method, the change makes sense.
> I let @mherwege as author of the PR comment.

I am sorry, my comment was not applicable to that PR, I mismatched with another PR!!!

Now, to answer to @J-N-K , this code looks good to me but there is probably missing code to return this new property through the  REST API. 

But I see it is too late.

bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/Sitemap.xtext
mherwege: @J-N-K @lolodomo That is correct. I intended to do that in a PR related to sitemap config in mainUI, because my understanding is that is where it is used. See https://github.com/openhab/openhab-core/pull/3431 for the base one (without inputHint). My intention was to create a separate PR for this, but I can include it in the existing open one if preferred.

mherwege:
> Now, to answer to @J-N-K , this code looks good to me but there is probably missing code to return this new property through the REST API.

@J-N-K @lolodomo That is correct. I intended to do that in a PR related to sitemap config in mainUI, because my understanding is that is where it is used. See https://github.com/openhab/openhab-core/pull/3431 for the base one (without inputHint). My intention was to create a separate PR for this, but I can include it in the existing open one if preferred.

lolodomo:
It should be separated from MainUI stuff, the REST API is used by the Android app for example. So it would have make sense to cover that part in this PR.

J-N-K:
If it is needed to make it work please include it.

mherwege:
@lolodomo @J-N-K I clearly misunderstood and missed the property in the REST API

I now added all to the still open https://github.com/openhab/openhab-core/pull/3431. That one now should cover REST API and the core part of configuration (with inputHint) in the main UI.

PR #3428, opened 2023-03-06
WatchService should send the last notification kind

jimtng:
Fix an issue introduced in #3404

Scenario that caused the problem:
* Registered a listener for a directory within a configWatcher
* Created a file
* CREATE notification sent to the listener
* Unregistered the listener
* Deleted the file. 
* Registered a new listener
* Created another file with the same name
* Now `scheduledEventKinds` contains two events: DELETE (from the previous event not yet removed from scheduledEventKinds) and CREATE
* MODIFY was sent, despite it never having registered a MODIFY event.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/service/WatchServiceImpl.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/service/WatchServiceImpl.java
@@ -228,12 +228,13 @@ public class WatchServiceImpl implements WatchService, DirectoryChangeListener {
             future = scheduler.schedule(() -> notifyListeners(path), PROCESSING_TIME, TimeUnit.MILLISECONDS);
             scheduledEventKinds.computeIfAbsent(path, k -> new CopyOnWriteArrayList<>()).add(kind);
             scheduledEvents.put(path, future);
-
         }
     }
 
     private void notifyListeners(Path path) {
         List<Kind> kinds = scheduledEventKinds.remove(path);
+        logger.trace("notifyListeners({}) kinds: {}", path, kinds);
+
         if (kinds == null || kinds.isEmpty()) {
             logger.debug("Tried to notify listeners of change events for '{}', but the event list is empty.", path);
             return;
@@ -251,14 +252,14 @@ public class WatchServiceImpl implements WatchService, DirectoryChangeListener {
         // determine final event
         if (lastElement == Kind.DELETE) {
             if (firstElement == Kind.CREATE) {
-                logger.debug("Discarding events for '{}' because file was immediately deleted bafter creation", path);
+                logger.debug("Discarding events for '{}' because file was immediately deleted after creation", path);
                 return;
             }
             doNotify(path, Kind.DELETE);
         } else if (firstElement == Kind.CREATE) {
             doNotify(path, Kind.CREATE);
         } else {
-            doNotify(path, Kind.MODIFY);
+            doNotify(path, lastElement);
         }
     }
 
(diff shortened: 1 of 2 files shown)


J-N-K:
This behavior was introduced intentionally and is correct IMO. In what kind of situation would the listener be unregistered in between?

jimtng:
I came across this problem whilst running a series of tests. Each test case would register a listener, perform the tests, and remove the listener. The next test case registers another listener on the same directory, created the same file, but testing a different scenario, but it stumbled across this issue.

In the situation that #3404 fixed, wouldn't the last notification kind still be MODIFY anyway, therefore won't be affected by this PR?

jimtng:
Come to think of it, I think the code should be like this:

```java
        Kind firstElement = kinds.get(0);
        Kind lastElement = kinds.get(kinds.size() - 1);

        if (firstElement == Kind.CREATE) {
            doNotify(path, Kind.CREATE);
        }

        if (kinds.size() > 1 && firstElement != lastElement)
            doNotify(path, lastElement);
        }
    }
```

I just don't have a way to test it and it doesn't affect my scenario.

J-N-K:
No, DELETE - CERATE - DELETE - CREATE would result in CREATE which is wrong, it should be MODIFY. I still fail to see where in a real world application this should lead to an issue. It is far more likely that a CREATE results in an issue (because clean-up is not done) than a MODIFY (which probably unnecessarily cleans something up that is not there).

jimtng:
> No, DELETE - CERATE - DELETE - CREATE would result in CREATE which is wrong, it should be MODIFY.

A CREATE at the end of the series of notifications should be treated as a fresh, EMPTY file, so the code that handles the change notification is the one that would need to change to handle this scenario, I would think?

The WatchService itself should be faithfully conveying the actual events, and not mask DELETE/CREATE/DELETE/CREATE and change it to a completely different event MODIFY (which didn't occur), just for the sake of tricking some specific event handling code to do what it should be doing.

I understand the need to buffer and deduplicate the events, but I disagree with completely changing the actual events.

J-N-K:
But that is exactly what is happening. I modify a file and get that event sequence on a Samba share.

jimtng:
OK, how about If DELETE-CREATE-DELETE-CREATE would be reduced to DELETE (to let the listener discard its buffer / remove the entry) + CREATE (to let it create a new, empty entry), wouldn't that achieve the same as MODIFY?

jimtng:
@J-N-K I've updated this to send DELETE first then CREATE. Would you mind testing to see if that works for you?

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/service/WatchServiceImpl.java
```diff
@@ -248,17 +249,17 @@ private void notifyListeners(Path path) {
...
+            logger.debug("Discarding events for '{}' because file was immediately deleted after creation", path);
+            return;
+        }
+
+        if (firstElement == Kind.CREATE || firstElement == Kind.DELETE) {
+            doNotify(path, firstElement);
+        }
+
+        if (kinds.size() > 1 && firstElement != lastElement) {
+            doNotify(path, lastElement);
```
J-N-K, line 262: That's not correct. CREATE - MODIFY should result in only one event CREATE. There is no need for a second reload.

I still don't understand what real-life use-case can cause an issue with the way it is implemented now.

jimtng: I've updated the code again, and also added a test to clarify the input vs output sequence. CREATE - MODIFY should now only result in one event CREATE.

> I still don't understand what real-life use-case can cause an issue with the way it is implemented now.

MODIFY - DELETE - CREATE - DELETE - MODIFY (or a combination thereof) should send DELETE - CREATE so the listeners are aware that the files were deleted and recreated. This should still make existing listeners / handlers to work because they will still reload the file once, so it still solves the multiple reload problem, but at the same time improves the visibility of what truly happened.

jimtng:
Sending DELETE - CREATE instead of MODIFY may cause double loading for things such as dependencyTracker. So for that reason, I'm closing this PR for now and will find a workaround in the cases where it's necessary to detect a DELETE - CREATE sequence.

Issue #3317, opened 2023-01-14
[jetty] OSGI does not load HpackFieldPreEncoder module


andrewfg:
As reported on this [Jetty Issue](https://github.com/eclipse/jetty.project/issues/9168), during my work on the [Hue Clip 2 Binding](https://github.com/openhab/openhab-addons/pull/13570), when the binding sends a client HTTP/2.0 PUT request, it throws an 'HpackException$SessionException', but the same code succeeds when it sends the exact same client HTTP/2.0 request with GET instead of PUT. It seems from the Jetty issue thread discussion that the problem is due to the OH OSGI service loader is not loading the `Http1FieldPreEncoder` module.

EDIT: actually the cause is that OH core is not loading the `org.eclipse.jetty.http2.hpack.HpackFieldPreEncoder` module.

J-N-K:
Please try and report again after #3252 has been merged. Thanks.

andrewfg:
> Please try and report again after https://github.com/openhab/openhab-core/pull/3252 has been merged. Thanks.

@ J-N-K as I learned from the ongoing thread in the Jetty issue, this OH issue will not be solved by simply updating the OH core Jetty Http/1.1 version from v9.4.46 to v9.4.50 (indeed it would also not be solved by upgrading Jetty Http/1.1 to v10.x or even v11.x)

HOWEVER this issue would be solved by including the Jetty HTTP/2 modules in #3252 as well as the HTTP/1.1 modules..

To be specific, my Hue CLIP2 binding PR depends on the following modules, and I think all of them could/should be added with value to the OH core. However for a minimalist approach, just the `org.eclipse.jetty.http2/http2-hpack` is the minimum required to solve this current issue.

```
org.eclipse.jetty.http2/http2-client
org.eclipse.jetty.http2/http2-common
org.eclipse.jetty.http2/http2-hpack
org.eclipse.jetty/jetty-alpn-java-client
org.eclipse.jetty/jetty-alpn-client
```

Furthermore it might also need us to add the following SPI file to OH core as well (??)

**./META-INF/service/org.eclipse.jetty.http.HttpFieldPreEncoder**
```
org.eclipse.jetty.http.Http1FieldPreEncoder
org.eclipse.jetty.http2.hpack.HpackFieldPreEncoder
```

J-N-K:
Yes, but we should go for that after #3252, otherwise we'll create merge conflicts.

andrewfg:
Ok to that.

[...]

PR #3433, opened 2023-03-07
[oh.core.io.net] Add support for Jetty HTTP/2 clients

andrewfg:
### Description
This PR adds support for Jetty HTTP/2 clients (as discussed in the thread of https://github.com/openhab/openhab-core/issues/3317).

Specifically it does the following..

1. Add dependencies on Jetty HTTP2 modules, plus Jetty ALPN modules that are required for HTTP2 implementation.
2. Adds `createHttp2Client` methods to the existing `HttpClientFactory` interface.
3. Adds implementations of the `createHttp2Client` methods to the `WebClientFactoryImpl` component.

The added Jetty (direct plus transitive) dependencies are as follows..
```
Jetty :: HTTP2 :: Client
Jetty :: HTTP2 :: Common
Jetty :: HTTP2 :: Hpack
Core :: ALPN :: Client
Core :: ALPN :: Java Client
```
### Notes

- Resolves https://github.com/openhab/openhab-core/issues/3317
- I am not a fully fluent expert on OH core development, so maintainers please check this one carefully!


andrewfg:
@J-N-K ping for info..

bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/Http2ClientFactory.java
```diff
@@ -0,0 +1,58 @@
...
+import org.eclipse.jetty.http2.client.HTTP2Client;
+import org.eclipse.jetty.util.ssl.SslContextFactory;
+
+/**
+ * Factory class to create Jetty HTTP/2 clients
+ *
+ * @author Andrew Fiddian-Green - Initial contribution
+ */
+@NonNullByDefault
+public interface Http2ClientFactory {
```
J-N-K, line 26: I think you should rather add the methods to `HttpClientFactory` instead of creating a new interface here.

J-N-K (changes requested):
You probably also need to adjust the openhab-tp feature.

bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/Http2ClientFactory.java
andrewfg: Ok.

andrewfg:
> You probably also need to adjust the openhab-tp feature.

@J-N-K you are correct. I discovered that also myself when the CI build failed. :)

wborn:
Why not depend on the pax-web-jetty-http2 and pax-web-jetty-http2-jdk9 features? It will prevent having to micromanage these bundles and use the correct start levels.

andrewfg:
> You probably also need to adjust the openhab-tp feature
> You probably also need to make sure that the pax-web-jetty-http2 and pax-web-jetty-http2-jdk9 features are installed

@J-N-K / @wborn guys I need your help here: I installed the dependencies in the `openhab.tp-httpclient` and `openhab-transport-http` features, but it still won't build; so I guess it is because the `pax-web-jetty-http2` and `pax-web-jetty-http2-jdk9` features are still not installed (??) .. but unfortunately I have no idea how (or in which feature file) to do that => can either of you kindly give me a tip?

.. I tried for a few hours adding stuff by trial and error, but I have an incredibly slow PC where the Maven build takes ages, and the error occurs on file 102 of 131, which is somewhat frustrating .. perhaps it is telling me that I need to buy a new PC ;)

wborn:
You could for instance try adding the following to the `openhab.tp-httpclient` feature:

```
<feature dependency="true">pax-web-jetty-http2</feature>
<feature dependency="true">pax-web-jetty-http2-jdk9</feature>
```

For reference what's in them, see this Pax Web [features.xml](https://github.com/ops4j/org.ops4j.pax.web/blob/cffd894d95325faed94a74c6d9e9e9e31db3c0ce/pax-web-features/src/main/resources/features.xml)

When using these predefined features, we won't have to deal with any bundle name changes, versions, adding missing bundles or messing around with start levels.

We should also look into using the Pax Web features for all the websocket bundles.

andrewfg:
@wborn I added the  `pax-web-jetty-http2` and `pax-web-jetty-http2-jdk9` features. It was necessary to add them to both karaf 'core' and karaf 'tp' feature.xml files. And doing the full build on my local PC (via `mvn clean install -U -DskipTests -DskipChecks` ) the build was successful. Nevertheless the GitHub CI build fails, and I have no idea why. => So perhaps you can advise?

wborn:
I had a look and created https://github.com/andrewfg/openhab-core/pull/1.

wborn:
Most likely you need to resolve the runbundles:

https://github.com/openhab/openhab-addons#integration-tests

It does not use Karaf features when running itests.

andrewfg:
> Most likely you need to resolve the runbundles:

@wborn OK I have done that. I did note however that a side effect of running `mvn -DwithResolver` was that `org.apache.felix.http.servlet-api;version='[1.2.0,1.2.1)'` got downgraded to `org.apache.felix.http.servlet-api;version='[1.1.2,1.1.3)'` (??) .. => I manually reverted it change back to v1.2 => but you may want to look into this yourself.

wborn:
You should never have to manually edit the runbundles and just stick to what is resolved by the resolver.
If you have good reasons for something else to be resolved, you need to fix this by either changing POMs or the runrequires/runblacklist in bndrun files.

andrewfg:
> You should never have to manually edit the runbundles

@wborn so why do the openhab-core/main existing run bundles contain felix v1.2 yet running the resolver on my PR causes a downgrade to a lower felix version? No matter what you may argue, I am convinced that this has nothing to do with the current. PR.

bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/HttpClientFactory.java
```diff
@@ -64,4 +66,35 @@ public interface HttpClientFactory {
...
+     * Creates a new Jetty HTTP/2 client.
+     * The returned client is not started yet. You have to start it yourself before using.
+     * Don't forget to stop a started client again after its usage.
+     * The client lifecycle should be the same as for your service.
+     * DO NOT CREATE NEW CLIENTS FOR EACH REQUEST!
+     *
+     * @param consumerName for identifying the consumer in the Jetty thread pool.
+     *            Must be between 4 and 20 characters long and must contain only the following characters [a-zA-Z0-9-_]
+     * @return the Jetty HTTP/2 client
+     * @throws NullPointerException if {@code consumerName} is {@code null}
```
wborn, line 80:
```suggestion
```

I think it is better to remove all these `@throws NullPointerException` from the interface. I see it was already part of the existing methods on the interface. Because the interface is annotated with `@NonNullByDefault` the "contract" is violated when `null` is passed as  `consumerName` so then all bets are off. Also when code is properly annotated with `@NonNullByDefault` the compiler will yell at the programmer to stop passing `null` values.

wborn:
> why do the openhab-core/main existing run bundles contain felix v1.2 yet running the resolver on my PR causes a downgrade to a lower felix version? No matter what you may argue, I am convinced that this has nothing to do with the current. PR

I don't see a good reason either why it would previously resolve to 1.2.
But then again I'm only human so I am badly suited for this [according to Bnd](https://bnd.bndtools.org/chapters/250-resolving.html):

> Establishing the perfect class path for an application is a process for which the human mind is extremely badly suited and the transitive model of Maven has too many sharp edges. The resolver provides an alternative model that focuses on looking at the whole solution space instead of having to live with the (sometimes arbitrary) decisions of developers or of deeply nested transitive dependencies.

bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/HttpClientFactory.java
andrewfg: Ok

bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/internal/WebClientFactoryImpl.java
```diff
@@ -362,4 +366,42 @@ private SslContextFactory createSslContextFactory() {
...
+        try {
+            logger.debug("creating HTTP/2 client for consumer {}", consumerName);
+
+            try {
+                PreEncodedHttpField field = new PreEncodedHttpField(HttpHeader.C_METHOD, "PUT");
+                ByteBuffer bytes = ByteBuffer.allocate(32);
+                field.putTo(bytes, HttpVersion.HTTP_2);
+            } catch (Exception e) {
+                throw new HttpClientInitializationException("Jetty HTTP/2 hpack module not loaded", e);
+            }
```
wborn, line 392: Does this really need to be checked every time? I think when the hpack bundle is stopped after this check would it still throw some exception when creating or using the client?

andrewfg: The issue is that the lack of HTTP/2 hpack encoding is not detected when CREATING the client. Indeed it is not normally detected when doing HTTP/ GET calls with commonly used headers (because these common cases are pre-encoded without needin hpack). The problem only actually arises when the client makes more 'exotic' calls such as PUT, or uses less common headers, which require encoding on the fly; and at this point the lack cayses an exception.

So my intention is to force throw the exception on client creation. In theory the check could be done one time only on creating the first client. The only problem with that is that you could see the exception originating from only the first binding that creates a client (whichever it might be) and not from any subsequent binding creating a client; so you might conclude that the issue was randomly related to that one first binding, rather than being a common issue in the core.

andrewfg: PS I just pushed a minor optimisation..

wborn (approved):
LGTM now, thanks!

wborn:
I've created https://github.com/ops4j/org.ops4j.pax.web/issues/1842 which may help with simplifying the management of all those Jetty bundles.

andrewfg:
@J-N-K / @wborn .. Oops! Sorry but I missed one dependency in this PR; when I run it now on the latest OH 4.0 Snapshot I get the error shown below, which I can manually dropping `jetty-alpn-java-client-9.4.46.v20220331.jar` in my addons folder. => How shall we proceed, shall I open a new PR? Or reopen this one?

```
2023-03-13 18:41:41.017 [WARN ] [org.eclipse.jetty.io.ManagedSelector] - java.lang.IllegalStateException: No Client ALPNProcessors!
```

andrewfg:
@wborn / @J-N-K the missing dependency was already declared in the runtime pom.xml, the feature.xml, and the itest.bndrun files; but not declared in the compile time pom.xml (because it is not needed at compile time). So I don’t really understand why it is not being imported at runtime. ?? Do you guys have any suggestions?

wborn:
Sometimes the resolver has no clue bundles are required, e.g. if they provide some implemention of an interface defined in another bundle. In that case you add a `runrequires` for that bundle.

PR #3434, opened 2023-03-07
Add ColorUtil for better support of xyY conversion

J-N-K:
This has been refactored to align with the usually used conversion by a lot of ZigBee products like Hue or deconz.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
@@ -1,0 +1,293 @@
+/**
+ * Copyright (c) 2010-2023 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.util;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.openhab.core.library.types.HSBType;
+import org.openhab.core.library.types.PercentType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The {@link ColorUtil} is responsible for converting HSB to CIE
+ *
+ * The class is based work from Erik Baauw for the <a href="https://github.com/ebaauw/homebridge-lib">Homebridge</a>
+ * project
+ *
+ * @author Jan N. Klug - Initial contribution
+ */
+@NonNullByDefault
+public class ColorUtil {
+    private static final Logger LOGGER = LoggerFactory.getLogger(ColorUtil.class);
+    public static final Gamut DEFAULT_GAMUT = new Gamut(new double[] { 0.9961, 0.0001 }, new double[] { 0, 0.9961 },
+            new double[] { 0, 0.0001 });
+
+    private ColorUtil() {
+        // prevent instantiation
+    }
+
+    /**
+     * Transform <a href="https://en.wikipedia.org/wiki/SRGB">sRGB</a> based {@link HSBType} to
+     * <a href="https://en.wikipedia.org/wiki/CIE_1931_color_space">CIE 1931</a> `xy` format.
+     *
+     * See <a href=
+     * "https://developers.meethue.com/develop/application-design-guidance/color-conversion-formulas-rgb-to-xy-and-back/">Hue
+     * developerportal</a>.
+     *
+     * @param hsbType a {@link HSBType} value
+     * @return double array with the closest matching CIE 1931 colour, x, y between 0.0000 and 1.0000.
+     */
+    public static double[] hsbToXY(HSBType hsbType) {
+        return hsbToXY(hsbType, DEFAULT_GAMUT);
+    }
+
+    /**
+     * Transform <a href="https://en.wikipedia.org/wiki/SRGB">sRGB</a> based {@link HSBType} to
+     * <a href="https://en.wikipedia.org/wiki/CIE_1931_color_space">CIE 1931</a> `xy` format.
+     *
+     * See <a href=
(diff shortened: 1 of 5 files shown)


bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/HSBTypeTest.java
```diff
@@ -129,14 +129,14 @@ public void testConversionToPointType() {
     public void testConversionToXY() {
         HSBType hsb = new HSBType("220,90,50");
         PercentType[] xy = hsb.toXY();
-        assertEquals(new PercentType("16.969364"), xy[0]);
-        assertEquals(new PercentType("12.379659"), xy[1]);
+        assertEquals(14.65, xy[0].doubleValue(), 0.01);
+        assertEquals(11.56, xy[1].doubleValue(), 0.01);
```
J-N-K, line 133: These colors are nearly the same. Using e.g. https://www.easyrgb.com/en/convert.php#inputFORM to convert both colors to HTML color code results in `#021645` and `#001849` which is very close.

bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/HSBTypeTest.java
```diff
@@ -129,14 +129,8 @@ public void testConversionToPointType() {
     public void testConversionToXY() {
         HSBType hsb = new HSBType("220,90,50");
         PercentType[] xy = hsb.toXY();
-        assertEquals(new PercentType("16.969364"), xy[0]);
-        assertEquals(new PercentType("12.379659"), xy[1]);
-    }
-
-    @Test
-    public void testCreateFromXY() {
```
J-N-K, line 137: I have removed this test because it was invalid. The JavaDoc of `HSBType.fromXY` states that the parameters should be between 0 and 1.

holgerfriedrich:
Thanks @J-N-K, I was preparing a PR as well, but of course you were faster and already included the HSBType.... 👍

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -0,0 +1,293 @@
...
+        public Point(double x, double y) {
+            this.x = x;
+            this.y = y;
+        }
+
+        /**
+         * distance between this point and another point
+         *
+         * @param other the other point
+         *
```
holgerfriedrich, line 207:
```suggestion
```

remove this line to avoid checkstyle warning

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -0,0 +1,293 @@
...
+     * @return double array with the closest matching CIE 1931 colour, x, y, Y between 0.0000 and 1.0000.
+     */
+    public static double[] hsbToXY(HSBType hsbType, Gamut gamut) {
+        double r = inverseCompand(hsbType.getRed().doubleValue() / PercentType.HUNDRED.doubleValue());
+        double g = inverseCompand(hsbType.getGreen().doubleValue() / PercentType.HUNDRED.doubleValue());
+        double b = inverseCompand(hsbType.getBlue().doubleValue() / PercentType.HUNDRED.doubleValue());
+
+        double X = r * 0.664511 + g * 0.154324 + b * 0.162028;
+        double Y = r * 0.283881 + g * 0.668433 + b * 0.047685;
+        double Z = r * 0.000088 + g * 0.072310 + b * 0.986039;
```
holgerfriedrich, line 73: this seems to checkstyle warning LocalVariableNameCheck

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -0,0 +1,293 @@
...
+     * @return the corresponding {@link HSBType}.
+     */
+    public static HSBType xyToHsv(double[] xy, Gamut gamut) {
+        Point p = gamut.closest(new Point(xy[0], xy[1]));
+        double x = p.x;
+        double y = p.y == 0.0 ? 0.000001 : p.y;
+        double z = 1.0 - x - y;
+        double Y = (xy.length == 3) ? xy[2] : 1.0;
+        double X = (Y / y) * x;
+        double Z = (Y / y) * z;
```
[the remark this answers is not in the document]
J-N-K: True, but this seems to be the common naming scheme for these values. If you have a good suggestion, please let me know.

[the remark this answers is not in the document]
holgerfriedrich: It is much clearer to keep the commonly used names Y, X, Z. I tried to add a ```SuppressWarnings``` tag, but did not succeed. Maybe adding a prefix is a quick solution, e.g. ```tmpX```...

holgerfriedrich:
@kaikreuzer It would be helpful to include this PR in the first milestone build, as we would like to use it in addons....

kaikreuzer:
@holgerfriedrich Ok. But it seems that you did some review comments, so I'll first wait for @J-N-K to resolve those.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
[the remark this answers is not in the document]
J-N-K: I have added a suppression for that.

holgerfriedrich (approved):
LGTM

Just a remark: I am not sure if checking the range of x and y to be in 0..1 will break a binding implementation. Looking at the usage of ```HSBType.fromXY``` it is hard to tell. EspMilightHubHandler looks a bit suspicious, as it scales up the values by 100.

J-N-K:
@kaikreuzer I think everything is addressed now.

kaikreuzer (approved):
Great, thank you.

florian-h05:
@J-N-K 
Probably related to this PR: Hue Binding Tests are broken, see https://github.com/openhab/openhab-addons/actions/runs/4454698602/jobs/7824057913#step:9:219

NorbertHD:
Fixes for color test are missing for MQTT Things and Channels and TRÅDFRI Binding too.

PR #3435, opened 2023-03-08
Fix WatchQueueReader on new multi-level subdirectories

jimtng:
WatchQueueReader ignored subdirectories created through atomic `mkdir -p` operation involving multiple level subdirectories.

e.g.:
When `/openhab/conf/automation/ruby/a` didn't exist, `mkdir -p /openhab/conf/automation/ruby/a/b/c` would cause only `/openhab/conf/automation/ruby/a` (the direct subdirectory under the existing directory) to be monitored.

Changes inside `/openhab/conf/automation/ruby/a/b` and any further subdirectories won't be monitored.

This PR fixed that by registering the new directory recursively.

Obviously this only applies to openhab 3.x since 4.x had refactored this class out.


jimtng:
Closing this since I'm no longer using openhab 3 and noone's looked at it.

PR #3439, opened 2023-03-09
[persistence] fix time weighting in averageSince and averageBetween

rimago:
The `averageSince` and `averageBetween` methods now use the followed timespan instead of the previous timespan for weighting. 
While this was not so important when using it with regular timeseries, such as a per minute persistence strategy, it is necessary to work with with irregular timeseries, such as the on-change persistence strategy.

I have added the test `testAverageSinceDecimalTypeIrregularTimespans` to illustrate this.

See also [averageSince not working as expected](https://community.openhab.org/t/averagesince-not-working-as-expected/143127).

Signed-off-by: Florian Binder <fb@java4.info>


bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
```diff
@@ -1257,7 +1248,56 @@ private static Iterable<HistoricItem> getAllStatesBetween(Item item, ZonedDateTi
             }
             filter.setItemName(item.getName());
             filter.setOrdering(Ordering.ASCENDING);
-            return qService.query(filter);
+
```
J-N-K, line 1251: Can you shortly describe what you are doing here and why? It looks to me like you are creating "artificial" data points.

rimago: Yes, that's right. In the case that there is no HistoricItem at the beginning of the period, the historical state at the beginning (the last HistoricItem before the beginning) is taken and appended to the beginning as "artificial" data to be used in the calculation.

Let's assume we have the following data:
01:00 - 1
04:00 - 2
05:00 - 1
07:00 - 0

To calculate averageBetween(02:00, 08:00), we use the following data:
02:00 - 1  <-- this one is "artificial"
04:00 - 2
05:00 - 1
07:00 - 0

So we get: (1*2h+2*1h+1*2h+0*1h)/6h = 1

J-N-K: For discrete values (like switch states) this may be correct, but what about things like temperature. I don't think it's correct to assume that the state at 2:00 is the same as at 1:00. If the temperature is 18 °C at 01:00 and 19 °C at 04:00, do you think 18 °C would be the correct value at 03:30?

I must admit that I don't have a good solution for that but given that it was the way it is since a very long time, we should try to find the best solution if we change it now.

rimago: Yes, that is a question of interpolation and there might be some cases where other interpolation algorithms might be better. However, taking the previous value best fits the "on change" persistence strategy. In the case of regular persistence (e.g. every minute) the effect of the chosen interpolation algorithm is small and the quality of the average estimate mainly depends on the sampling rate. So if you need a better estimate you should increase the sampling rate instead of tuning the interpolation.

I think the best solution would be to let the user select the interpolation algorithm by a parameter. For example as it is there:
https://www.mathworks.com/help/matlab/ref/interp1.html#btwp6lt-1-method
However, this might be exaggerated. Since openhab is not a statistical tool, I would suggest the implemented "previous" strategy that best fits the "on change" persistence strategy.

J-N-K: Would it be an option to calculate the integral for 01:00 - 04:00 and then scale that by (04:00 - 02:00)/(04:00 - 01:00) = 2/3 if the requested start time is 02:00?

rimago: Yes, that would be an option. However, it would make the calculation even more complicated. 

How about moving the creation of the starting item to another function, such as getAllStatesBetweenWithStartingState(...)?
In this way, we could keep the getAllStatesBetween as it was and we do not need the ugly `addBeginningHistoricItem` flag.

rimago: Okay, I have moved it to the function `getAllStatesBetweenWithStartingState(...)`.

J-N-K: If we look at `PersistenceResource` where a similar issue is needed: It would probably better to call `historicState` to get the state at the beginning of the interval and the `getAllStatesBetween` in a second step.

rimago: That is actually what happens in the `getAllStatesBetweenWithStartingState(...)`-method. The advantage of having it in a separate method is that it can be used in all other methods, such as `internalMinimum`, `internalVariance`, ... .

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
```diff
@@ -1265,6 +1257,60 @@ private static Iterable<HistoricItem> getAllStatesBetween(Item item, ZonedDateTi
         }
     }
 
+    private static Iterable<HistoricItem> getAllStatesBetweenWithStartingState(Item item, ZonedDateTime begin,
```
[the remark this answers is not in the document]
rimago: Ok, sounds good. I like the idea of reusing the code for `PersistenceResource`. 
What do you think about creating a new function `getAllStatesBetweenWithBoundaries` instead of extending the `getAllStatesBetween` with the flag? Of course, it would use the `getAllStatesBetween` and just add the boundaries, as it is currently in `getAllStatesBetweenWithStartingState`. It  would replace the `getAllStatesBetweenWithStartingState` and could be used in both `PersistenceResource` and `PersistenceExtensions`.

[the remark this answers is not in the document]
J-N-K: I'm not totally against a new method, but where you see the benefit for that (compared to a parameter)?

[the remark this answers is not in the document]
rimago: I see mainly two benefits:
1. The methods are shorter and simpler.
2. The code is easier to understand. For example: The call `getAllStatesBetweenWithBoundaries(...)` is more informative than `getAllStatesBetween(..., true)`. In the latter case, you must always check the name of the parameter to understand its meaning.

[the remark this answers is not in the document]
J-N-K: Can you do the necessary changed to `getAllStatesBetweenWithBoundaries`? Otherwise I think we are good to merge.

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/RetimedHistoricItem.java
```diff
@@ -0,0 +1,54 @@
...
+        return originItem.getState();
+    }
+
+    @Override
+    public String getName() {
+        return originItem.getName();
+    }
+
+    @Override
+    public String toString() {
```
J-N-K, line 51: Shouldn't the state also be contained in the `.toString`?

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/RetimedHistoricItem.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.openhab.core.persistence.HistoricItem;
+import org.openhab.core.types.State;
+
+/**
+ * This class wraps a HistoricItem to reset the timestamp.
+ *
+ * @author Florian Binder - Initial contribution
+ */
+public class RetimedHistoricItem implements HistoricItem {
```
J-N-K, line 25: Please make this a `private static` inner class of `PersistenceExtensions`. We should not expose such "helper" classes to keep flexibility when doing refactoring in the future and avoid API breaking changes.

J-N-K (changes requested):
Thanks unfortunately I now have another request related to your latest changes.

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/RetimedHistoricItem.java
rimago: Sure, did not know that.
Done.

J-N-K (approved):
Thanks!

PR #3445, opened 2023-03-11
Fix race condition in AbstractScriptFileWatcher

jimtng:
Some script file watcher implementations may need to access their rule engine's configurations. The rule engines need to be initialized before ScriptFileWatcher performs the initial import.

Specifically, this issue caused an NPE in jrubyscripting addon because AbstractScriptFileWatcher initialImport() calls getScriptType() down the line but still within AbstractScriptFileWatcher's constructor. The derived class hasn't finished initializing itself.  In jrubyscripting's implementation, to decide whether the given path should be watched, it needs access to its scriptengine, which at this point hasn't been initialized yet.

### DIFF
--- a/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
+++ b/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
@@ -111,7 +111,7 @@ public abstract class AbstractScriptFileWatcher implements WatchService.WatchEve
         this.scheduler = getScheduler();
 
         currentStartLevel = startLevelService.getStartLevel();
-        if (currentStartLevel > StartLevelService.STARTLEVEL_MODEL) {
+        if (currentStartLevel >= StartLevelService.STARTLEVEL_RULEENGINE) {
             initialImport();
         }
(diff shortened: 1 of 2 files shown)


J-N-K:
Do I understand it correctly that the issue is that `initialImport` is called in the constructor and that initializing code in the sub-classes constructor is needed for initial-import to succeed? 

In that case I don't think that the way you chose is correct, it prevents early script execution.

We could add a second constructor which has an addition boolean parameter `skipInitialImport` that does not call `initialImport` and leaves that to the sub-class. By doing so the standard way (like it is now) can stay and all implementations that work with that will need to adaption and special cases could be handled by calling `super(..., true);` and then do the initial import after they have finished their initialization.

WDYT?

jimtng:
> Do I understand it correctly that the issue is that `initialImport` is called in the constructor and that initializing code in the sub-classes constructor is needed for initial-import to succeed?

Yes, indeed.

> 
> In that case I don't think that the way you chose is correct, it prevents early script execution.

Indeed, I ran into some problems, hence why I've marked this as Draft.

> 
> We could add a second constructor which has an addition boolean parameter `skipInitialImport` that does not call `initialImport` and leaves that to the sub-class. By doing so the standard way (like it is now) can stay and all implementations that work with that will need to adaption and special cases could be handled by calling `super(..., true);` and then do the initial import after they have finished their initialization.
> 

I think that's a great idea. Let me try that out and see.

jimtng:
Superseded by #3451

PR #3446, opened 2023-03-12
Upgrade XStream to 1.4.20

wborn:
This addresses CVE-2022-40151 and CVE-2022-41966, see:

https://x-stream.github.io/changes.html#1.4.20

This version also [fixes](https://github.com/x-stream/xstream/pull/287) an issue with closing streams so the workaround in GenerateDefaultTranslationsMojoTest is no longer needed.


J-N-K (approved):
LGTM.

521 seems to be an cloudflare error.

PR #3448, opened 2023-03-12
Channels once more displayed in correct order in Main UI

### LINKED ISSUE #3442: OH 4.0 UI changes the sort order of channels
Is it just me, or does OH v4.0 now change the sort order of channels in the UI? Previously they would sort according to the order in the thing.xml but now they seem to display in a random order..

```
<channels>
    <channel id="position" typeId="shade-position"/>
    <channel id="secondary" typeId="shade-position"/>
    <channel id="vane" typeId="shade-vane"/>
    <channel id="lowBattery" typeId="system.low-battery"/>
    <channel id="batteryLevel" typeId="system.battery-level"/>
    <channel id="signalStrength" typeId="system.signal-strength"/>
</channels>
```

image

andrewfg:
Fixes #3442

As described in #3442 _"we switched from a `List<Channel>` to a `Map<ChannelUID, Channel>` which does not necessarily preserve the order"_ -- with the consequence that in the Main UI, Thing Channels were no longer displayed in the order in which they were declared in the `thing.xml` file.

This PR uses `LinkedHashMap<ChannelUID, Channel>` instead of `HashMap<ChannelUID, Channel>` so that Thing Channels are once more displayed in the order in which they were declared in the `thing.xml` file.

### DIFF
--- a/itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/ThingChannelsTest.java
+++ b/itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/ThingChannelsTest.java
@@ -31,7 +31,8 @@ import org.openhab.core.thing.binding.builder.ThingBuilder;
 @NonNullByDefault
 public class ThingChannelsTest extends JavaOSGiTest {
 
-    private static final List<String> CHANNEL_IDS = List.of("polarBear", "alligator", "hippopotamus", "aardvark");
+    private static final List<String> CHANNEL_IDS = List.of("polarBear", "alligator", "hippopotamus", "aardvark",
+            "whiteRabbit", "redHerring", "orangutan", "kangaroo", "rubberDuck", "timorousBeastie");
 
     @Test
     public void testThingChannelOrder() {
--- a/itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/ThingChannelsTest.java
+++ b/itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/ThingChannelsTest.java
@@ -22,6 +22,7 @@ import org.junit.jupiter.api.Test;
 import org.openhab.core.test.java.JavaOSGiTest;
 import org.openhab.core.thing.binding.builder.ChannelBuilder;
 import org.openhab.core.thing.binding.builder.ThingBuilder;
+import org.openhab.core.thing.dto.ThingDTOMapper;
 
 /**
  * Test for thing channel creation and display order.
@@ -42,20 +43,26 @@ public class ThingChannelsTest extends JavaOSGiTest {
         // create and fill the list of origin channels
         List<Channel> originChannels = new ArrayList<>();
         CHANNEL_IDS.forEach(channelId -> originChannels
-                .add(ChannelBuilder.create(new ChannelUID(thingUID, channelId), null).withLabel(channelId).build()));
-
-        // build a thing with the origin channels and get its result channels
-        Thing thing = ThingBuilder.create(thingTypeUID, thingUID).withChannels(originChannels).build();
-        List<Channel> resultChannels = thing.getChannels();
-
-        // run the tests
+                .add(ChannelBuilder.create(new ChannelUID(thingUID, channelId), null).build()));
         assertEquals(CHANNEL_IDS.size(), originChannels.size());
+
+        // build a thing with the origin channels
+        Thing thing = ThingBuilder.create(thingTypeUID, thingUID).withChannels(originChannels).build();
+
+        List<Channel> resultChannels;
+
+        // test #1: read the channels from the thing, and compare the resulting channel order
+        resultChannels = thing.getChannels();
         assertEquals(CHANNEL_IDS.size(), resultChannels.size());
         for (int i = 0; i < CHANNEL_IDS.size(); i++) {
-            String expectId = CHANNEL_IDS.get(i);
-            Channel resultChannel = resultChannels.get(i);
-            assertTrue(expectId.equals(resultChannel.getUID().getId()));
-            assertTrue(expectId.equals(resultChannel.getLabel()));
+            assertTrue(CHANNEL_IDS.get(i).equals(resultChannels.get(i).getUID().getId()));
+        }
+
+        // test #2: serialize/deserialize the thing via a DTO, and compare the resulting channel order
+        resultChannels = ThingDTOMapper.map(ThingDTOMapper.map(thing), false).getChannels();
+        assertEquals(CHANNEL_IDS.size(), resultChannels.size());
+        for (int i = 0; i < CHANNEL_IDS.size(); i++) {
+            assertTrue(CHANNEL_IDS.get(i).equals(resultChannels.get(i).getUID().getId()));
         }
     }
 }
--- a/itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/ThingChannelsTest.java
+++ b/itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/ThingChannelsTest.java
@@ -1,0 +1,60 @@
+/**
+ * Copyright (c) 2010-2023 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.thing;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.junit.jupiter.api.Test;
+import org.openhab.core.test.java.JavaOSGiTest;
+import org.openhab.core.thing.binding.builder.ChannelBuilder;
+import org.openhab.core.thing.binding.builder.ThingBuilder;
+
+/**
+ * Test for thing channel creation and display order.
+ *
+ * @author Andrew Fiddian-Green - Initial contribution
+ */
+@NonNullByDefault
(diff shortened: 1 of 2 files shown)


J-N-K (approved):
LGTM

wborn:
A unit test for checking the order would also be nice to prevent such regressions.

andrewfg:
> A unit test for checking the order would also be nice to prevent such regressions.

Ok. Coming up.

andrewfg:
Oops. I realise that the current version of my test has 4 channels only so it has a 1 in 24 change of passing by random. So I will add some more channels to reduce that risk.

itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/ThingChannelsTest.java
```diff
@@ -0,0 +1,61 @@
...
+        ThingUID thingUID = new ThingUID(thingTypeUID, "thingLabel");
+
+        // create and fill the list of origin channels
+        List<Channel> originChannels = new ArrayList<>();
+        CHANNEL_IDS.forEach(channelId -> originChannels
+                .add(ChannelBuilder.create(new ChannelUID(thingUID, channelId), null).withLabel(channelId).build()));
+
+        // build a thing with the origin channels and get its result channels
+        Thing thing = ThingBuilder.create(thingTypeUID, thingUID).withChannels(originChannels).build();
+        List<Channel> resultChannels = thing.getChannels();
```
J-N-K, line 49: I think the real issue is one step deeper. 

You should use `ThingDTOMapper` to convert the thing to the stored format and back and then do the test for correct ordering with that "restored" thing.

wborn:
I think `ThingImpl`s are also stored in the `JsonStorage` using a `OrderingMapSerializer` (see [code](https://github.com/wborn/openhab-core/blob/3ce1738adae6ca558ac66baf7dc127fa72945841/bundles/org.openhab.core.storage.json/src/main/java/org/openhab/core/storage/json/internal/JsonStorage.java#L102-L113))
So it could still be that the channels have a different order when `ThingImpl`s are deserialized that way.

itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/ThingChannelsTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+        // run the tests
+        assertEquals(CHANNEL_IDS.size(), originChannels.size());
+        assertEquals(CHANNEL_IDS.size(), resultChannels.size());
+        for (int i = 0; i < CHANNEL_IDS.size(); i++) {
+            String expectId = CHANNEL_IDS.get(i);
+            Channel resultChannel = resultChannels.get(i);
+            assertTrue(expectId.equals(resultChannel.getUID().getId()));
+            assertTrue(expectId.equals(resultChannel.getLabel()));
+        }
```
J-N-K, line 59: I don't think you need to check the labels. You can simplify this to

```
List<String> resultChannelIds = resultChannels.stream().map(Channel:getId).toList();
assertThat(resultChannelIds, contains(CHANNEL_IDS.toArray()));
```

J-N-K (changes requested):

J-N-K:
> I think `ThingImpl`s are also stored in the `JsonStorage` using a `OrderingMapSerializer` (see [code](https://github.com/wborn/openhab-core/blob/3ce1738adae6ca558ac66baf7dc127fa72945841/bundles/org.openhab.core.storage.json/src/main/java/org/openhab/core/storage/json/internal/JsonStorage.java#L102-L113))
> So it could still be that the channels have a different order when `ThingImpl`s are deserialized that way.

The `ThingImpl` is not stored, the `ThingStorageEntity` is stored and that uses a `List`. The `Map` is generated when mapping `ThingStorageEntity` to `ThingImpl`.

itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/ThingChannelsTest.java
andrewfg: Hmm. It’s not obvious to me that ‘contains’ is checking the order of entries.

andrewfg:
> The ThingImpl is not stored, the ThingStorageEntity is stored and that uses a List. The Map is generated when mapping ThingStorageEntity to ThingImpl.

Yes that is my impression too. And in my actual testing, I can confirm that (using LinkedHashMap) the ordering does indeed persist over a restart (because LinkedHashMap properly serialises to, and deserialises from, a List..)

itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/ThingChannelsTest.java
andrewfg: .. but I agree that the labels are not strictly necessary, so I will push a compromise between my prior code and your suggestion..

itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/ThingChannelsTest.java
andrewfg: ^
Probably the test should check both i.e. both of the round trips..
- List => LinkedHashMap => List, AND also 
- List => Serialize => Deserialize => List

wborn:
Will it also reapply the order according to how they are ordered in the Thing Type after deserialization? Channels could get rearranged in newer OH versions or new channels may get added in between existing channels.

mhilbush:
Makes me also wonder what happens wrt new channels during a binding update when the update instructions are applied. Will the update instructions result in the new channels being added to the end of the list of channels versus a new installation of the binding which will put the new channels wherever the developer put them in the thing xml?

andrewfg:
> Will the update instructions result in the new channels being added to the end of the list of channels

AFAIK, if you call ‘editThing’ you get the original channels as a List, and recreate the Thing via ‘.withChannels’ you supply the.new channels as a List. So maintaining the sequence of entries in the List is the critical point. And if new channels are added via ‘List.add’ they are added at then end of the List, etc.

andrewfg:
> Will it also reapply the order according to how they are ordered in the Thing Type after deserialization?

AFAIK the channel order is defined by thing.xml when the thing is first instantiated. Thereafter the channel order should be serialised and deserialised in the same order on each restart. I am not 100% sure but I think that if a new binding version changes the order in thing.xml, that change is NOT directly applied to already defined things, so you have to delete and recreate the thing to get the new order. And it’s probably different for things defined via ‘.things’ file.

mhilbush:
>  AFAIK, if you call ‘editThing’ you get the original channels as a List, and recreate the Thing via ‘.withChannels’ you supply the.new channels as a List

Sorry, I was referring to a new feature in 4.0 that let's you (the developer) provide instructions that indicate, new, changed, or deleted channels, which are applied to the thing without the need to delete/recreate the thing.

For example, see [this PR](https://github.com/openhab/openhab-addons/pull/14460).

In this case, the new channels were added to the end of the channels in thing-types.xml.

However, what if the new channels were inserted into the middle of the list of channels in thing-types.xml? In the case of a newly created thing, the order would be as the channels are defined in thing-types.xml. But in the case of an existing thing that's updated as a consequence of the update instructions, I assume the new channels would be added at the end of the list.

I don't think this is a really big deal. But it is behavior that we want to understand so that we know what to expect.

So for example, let's say I have two things of the same thing type. One thing was newly created and one was an existing thing that was updated according to the update instructions. I'm assuming the order of the channels shown in the UI might be different depending on where the new channels appear in the channel list in thing-types.xml.

andrewfg:
> I was referring to a new feature in 4.0 

@mhilbush thanks for putting me on the right track.

I think your concern is not directly impacting the PR. This PR is essentially just about ensuring that whatever channels are created in a thing shall remain consistently in the same order in which they were originally added. i.e. both 
- transient consistency via `.withChannels()` => `getChannels()`, and 
- persistent consistency via `.withChannels()` => `ThingDTOMapper.map(Thing)` => `ThingDTO` => `ThingDTOMapper.map(ThingDTO)` => `Thing` => `getChannels()`.

mhilbush:
> I think your concern is not directly impacting the PR

Yes, I agree.

Since your original issue had to do with the order of channels in the UI, I just wanted to raise awareness that there may still be some cases where the order of channels presented in the UI *may* not exactly match the order of channels in the thing xml.

andrewfg:
> may still be some cases where the order of channels presented in the UI may not exactly match the order of channels in the thing xml

@mhilbush digging into the OH core channel modification instruction code, it looks like there are (currently) two supported instructions, namely `add` resp. `remove`. So currently there is nothing that could change the channel sort order. Of course I do admit that the core might add support for an `insertAt` instruction in future, so the situation that you describe might indeed occur in future. But I suggest to focus on the current rather than the future.

https://github.com/openhab/openhab-core/blob/0894d25448356cc4c4624fcefc3e279f8b1808f4/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/update/ThingUpdateInstructionReader.java#L84

mhilbush:
> An add instruction will append the new channel to the existing list. It will not insert into the middle of an existing list.

I'm sorry for not being clear. Yes, that's exactly what I'm trying to say.

In my example above, the channels of a newly created thing will be in the order in which the channels appear in the thing.xml. In that example, the developer inserted new channels `a1` and `a2` into the middle of the list (presumably because they had some relationship to channel `a`).

The channels of a thing updated by the update instructions will not be in the order in which the channels appear in the thing.xml because the new channels will be appended.

So for those two things, won't the order of the channels in the UI be different?

J-N-K:
Yes, that would be the case. But I don't think there is an easy way around it. While it may be possible to add something like `before="channelId"` to the add instruction, it'll become very complex for channel-groups. What if the developer just re-orders the channels? It is nice to preserve the order (especially when it is so easy like the change of the collection type) but it's not absolutely necessary.

J-N-K (approved):
LGTM, thanks.

mhilbush:
@J-N-K Yep, I agree completely. Like I said earlier, I just wanted to bring some awareness to the behavior for when someone asks why their channels are "out of order." 😉

PR #3451, opened 2023-03-12
Fix NPE caused by AbstractScriptFileWatcher initialization order

jimtng:
See the discussion in #3445

Another problem I encountered was the onReadyMarkerAdded got called out of order. It jumped from 80 -> 20 -> 100 -> 40 -> 50 -> 70 -> 30. Here's an example trace:
```
023-03-13 01:42:22.428 [TRACE] [ort.loader.AbstractScriptFileWatcher] - onReadyMarkerAdded(startlevel=80) 100 => 100 for /openhab/conf/automation/ruby
2023-03-13 01:42:22.428 [TRACE] [ort.loader.AbstractScriptFileWatcher] - onReadyMarkerAdded(startlevel=20) 100 => 100 for /openhab/conf/automation/ruby
2023-03-13 01:42:22.429 [TRACE] [ort.loader.AbstractScriptFileWatcher] - onReadyMarkerAdded(startlevel=100) 100 => 100 for /openhab/conf/automation/ruby
2023-03-13 01:42:22.430 [TRACE] [ort.loader.AbstractScriptFileWatcher] - onReadyMarkerAdded(startlevel=40) 100 => 100 for /openhab/conf/automation/ruby
2023-03-13 01:42:22.430 [TRACE] [ort.loader.AbstractScriptFileWatcher] - onReadyMarkerAdded(startlevel=50) 100 => 100 for /openhab/conf/automation/ruby
2023-03-13 01:42:22.431 [TRACE] [ort.loader.AbstractScriptFileWatcher] - onReadyMarkerAdded(startlevel=70) 100 => 100 for /openhab/conf/automation/ruby
2023-03-13 01:42:22.431 [TRACE] [ort.loader.AbstractScriptFileWatcher] - onReadyMarkerAdded(startlevel=30) 100 => 100 for /openhab/conf/automation/ruby
```

This caused AbstractScriptFileWatcher to stop loading the files because the last startLevel it received through onReadyMarkerAdded() was startLevel 30, so it never end up instructing importFileWhenReady to run.

### DIFF
--- a/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
+++ b/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
@@ -260,20 +260,21 @@ public abstract class AbstractScriptFileWatcher implements WatchService.WatchEve
         return CompletableFuture.runAsync(() -> {
             Lock lock = getLockForScript(scriptIdentifier);
             try {
-                ScriptFileReference ref = scriptMap.remove(scriptIdentifier);
+                scriptMap.computeIfPresent(scriptIdentifier, (id, ref) -> {
+                    if (ref.getLoadedStatus().get()) {
+                        manager.removeEngine(scriptIdentifier);
+                        logger.debug("Unloaded script '{}'", scriptIdentifier);
+                    } else {
+                        logger.debug("Dequeued script '{}'", scriptIdentifier);
+                    }
 
-                if (ref == null) {
-                    logger.warn("Failed to unload script '{}': script reference not found.", scriptIdentifier);
-                    return;
-                }
-
-                if (ref.getLoadedStatus().get()) {
-                    manager.removeEngine(scriptIdentifier);
-                    logger.debug("Unloaded script '{}'", scriptIdentifier);
-                } else {
-                    logger.debug("Dequeued script '{}'", scriptIdentifier);
-                }
+                    return null;
+                });
             } finally {
+                if (scriptMap.containsKey(scriptIdentifier)) {
+                    logger.warn("Failed to unload script '{}'", scriptIdentifier);
+                }
+
                 scriptLockMap.remove(scriptIdentifier);
                 lock.unlock();
             }
--- a/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
+++ b/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
@@ -104,16 +104,9 @@ public abstract class AbstractScriptFileWatcher implements WatchService.WatchEve
         this.readyService = readyService;
         this.watchSubDirectories = watchSubDirectories;
         this.watchPath = watchService.getWatchPath().resolve(fileDirectory);
-
-        manager.addFactoryChangeListener(this);
-        readyService.registerTracker(this, new ReadyMarkerFilter().withType(StartLevelService.STARTLEVEL_MARKER_TYPE));
-
         this.scheduler = getScheduler();
-
-        currentStartLevel = startLevelService.getStartLevel();
-        if (currentStartLevel > StartLevelService.STARTLEVEL_MODEL) {
-            initialImport();
-        }
+        // Start with the lowest level to ensure the code in onReadyMarkerAdded runs
+        this.currentStartLevel = StartLevelService.STARTLEVEL_OSGI;
     }
 
     /**
@@ -144,6 +137,9 @@ public abstract class AbstractScriptFileWatcher implements WatchService.WatchEve
         } else if (!Files.isDirectory(watchPath)) {
             logger.warn("Trying to watch directory {}, however it is a file", watchPath);
         }
+
+        manager.addFactoryChangeListener(this);
+        readyService.registerTracker(this, new ReadyMarkerFilter().withType(StartLevelService.STARTLEVEL_MARKER_TYPE));
         watchService.registerListener(this, watchPath, watchSubDirectories);
     }
 
@@ -353,20 +349,6 @@ public abstract class AbstractScriptFileWatcher implements WatchService.WatchEve
         return false;
     }
 
-    private void initialImport() {
-        File directory = watchPath.toFile();
-
-        if (!directory.exists()) {
-            if (!directory.mkdirs()) {
-                logger.warn("Failed to create watched directory: {}", watchPath);
-            }
-        } else if (directory.isFile()) {
-            logger.warn("Trying to watch directory {}, however it is a file", watchPath);
-        }
-
-        addFiles(listFiles(directory.toPath(), watchSubDirectories)).thenRun(() -> initialized.complete(null));
-    }
-
     @Override
     public void onDependencyChange(String scriptIdentifier) {
         logger.debug("Reimporting {}...", scriptIdentifier);
@@ -381,13 +363,16 @@ public abstract class AbstractScriptFileWatcher implements WatchService.WatchEve
         int previousLevel = currentStartLevel;
         currentStartLevel = Integer.parseInt(readyMarker.getIdentifier());
 
+        logger.trace("Added ready marker {}: start level changed from {} to {}. watchPath: {}", readyMarker,
+                previousLevel, currentStartLevel, watchPath);
+
         if (currentStartLevel < StartLevelService.STARTLEVEL_STATES) {
             // ignore start level less than 30
             return;
         }
 
-        if (currentStartLevel == StartLevelService.STARTLEVEL_STATES) {
-            initialImport();
+        if (!initialized.isDone()) {
+            addFiles(listFiles(watchPath, watchSubDirectories)).thenRun(() -> initialized.complete(null));
         } else {
             scriptMap.values().stream().sorted()
                     .filter(ref -> needsStartLevelProcessing(ref, previousLevel, currentStartLevel))
--- a/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
+++ b/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
@@ -219,27 +219,22 @@ public abstract class AbstractScriptFileWatcher implements WatchService.WatchEve
 
     @Override
     public void processWatchEvent(WatchService.Kind kind, Path path) {
+        if (!initialized.isDone()) {
+            // discard events if the initial import has not finished
+            return;
+        }
+
         Path fullPath = watchPath.resolve(path);
         File file = fullPath.toFile();
-        if (!file.isHidden()) {
-            if (kind == DELETE) {
-                if (file.isDirectory()) {
-                    if (watchSubDirectories) {
-                        synchronized (this) {
-                            String prefix = fullPath.getParent().toString();
-                            Set<String> toRemove = scriptMap.keySet().stream().filter(ref -> ref.startsWith(prefix))
-                                    .collect(Collectors.toSet());
-                            toRemove.forEach(this::removeFile);
-                        }
-                    }
-                } else {
-                    removeFile(ScriptFileReference.getScriptIdentifier(file.toPath()));
-                }
-            }
 
-            if (file.canRead() && (kind == CREATE || kind == MODIFY)) {
-                addFiles(listFiles(file.toPath(), watchSubDirectories));
+        // Subdirectory events are filtered out by WatchService, so we only need to deal with files
+        if (kind == DELETE) {
+            String scriptIdentifier = ScriptFileReference.getScriptIdentifier(fullPath);
+            if (scriptMap.containsKey(scriptIdentifier)) {
+                removeFile(scriptIdentifier);
             }
+        } else if (!file.isHidden() && file.canRead() && (kind == CREATE || kind == MODIFY)) {
+            addFiles(listFiles(fullPath, watchSubDirectories));
         }
     }
 
--- a/bundles/org.openhab.core.automation.module.script.rulesupport/src/test/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcherTest.java
+++ b/bundles/org.openhab.core.automation.module.script.rulesupport/src/test/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcherTest.java
@@ -620,8 +620,10 @@ class AbstractScriptFileWatcherTest extends JavaTest {
     @Test
     public void testIfInitializedForLateInitialization() {
         scriptFileWatcher = createScriptFileWatcher(true);
-        when(startLevelServiceMock.getStartLevel()).thenReturn(StartLevelService.STARTLEVEL_RULEENGINE);
         AbstractScriptFileWatcher watcher = createScriptFileWatcher();
+        watcher.activate();
+        watcher.onReadyMarkerAdded(new ReadyMarker(StartLevelService.STARTLEVEL_MARKER_TYPE,
+                Integer.toString(StartLevelService.STARTLEVEL_STATES)));
 
         waitForAssert(() -> assertThat(watcher.ifInitialized().isDone(), is(true)));
 
--- a/bundles/org.openhab.core.automation.module.script.rulesupport/src/test/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcherTest.java
+++ b/bundles/org.openhab.core.automation.module.script.rulesupport/src/test/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcherTest.java
@@ -622,6 +622,7 @@ class AbstractScriptFileWatcherTest extends JavaTest {
     @Test
     public void testIfInitializedForLateInitialization() {
         scriptFileWatcher = createScriptFileWatcher(true);
+        when(startLevelServiceMock.getStartLevel()).thenReturn(StartLevelService.STARTLEVEL_RULEENGINE);
         AbstractScriptFileWatcher watcher = createScriptFileWatcher();
         watcher.activate();
         watcher.onReadyMarkerAdded(new ReadyMarker(StartLevelService.STARTLEVEL_MARKER_TYPE,
--- a/bundles/org.openhab.core.automation.module.script.rulesupport/src/test/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcherTest.java
+++ b/bundles/org.openhab.core.automation.module.script.rulesupport/src/test/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcherTest.java
@@ -554,7 +554,7 @@ class AbstractScriptFileWatcherTest extends JavaTest {
     }
 
     @Test
-    public void testDirectoryRemoved() {
+    public void testFileRemoved() {
         scriptFileWatcher = createScriptFileWatcher(true);
         when(scriptEngineManagerMock.isSupported("js")).thenReturn(true);
         ScriptEngineContainer scriptEngineContainer = mock(ScriptEngineContainer.class);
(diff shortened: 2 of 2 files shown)


jimtng:
@J-N-K I'd appreciate your review on this

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
```diff
@@ -272,7 +277,7 @@ private CompletableFuture<Void> removeFile(String scriptIdentifier) {
                 ScriptFileReference ref = scriptMap.remove(scriptIdentifier);
 
                 if (ref == null) {
-                    logger.warn("Failed to unload script '{}': script reference not found.", scriptIdentifier);
+                    logger.debug("Failed to unload script '{}': script reference not found.", scriptIdentifier);
```
J-N-K, line 280: Why did you reduce this logger? If unloading fails it is very likely we create memory or thread leaks.

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
```diff
@@ -353,18 +358,14 @@ private boolean createAndLoad(ScriptFileReference ref) {
         return false;
     }
 
-    private void initialImport() {
-        File directory = watchPath.toFile();
```
J-N-K, line 357: Why did you remove the directory creation?

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
```diff
@@ -353,18 +358,14 @@ private boolean createAndLoad(ScriptFileReference ref) {
...
-
-        if (!directory.exists()) {
-            if (!directory.mkdirs()) {
-                logger.warn("Failed to create watched directory: {}", watchPath);
-            }
-        } else if (directory.isFile()) {
-            logger.warn("Trying to watch directory {}, however it is a file", watchPath);
+    protected void performInitialImport() {
```
J-N-K, line 361: I don't think this should be a separate method. If a sub-class wants to handle the import one it's own, it's free to do so and nesting methods only adds complexity without a real benefit.

J-N-K (changes requested):
I think we should better investigate why the ready markers arrive unordered. Or use the ready marker as trigger only to get the actual start level from the `StartLevelService`.

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
jimtng: The problem is that this would happen when some random, untracked files got deleted.
Example:
- /conf/automation/ruby/ is the main watchPath
- /conf/automation/ruby/a_file is excluded and not added to scriptMap (for whatever reason - e.g. it is excluded by AbstractScriptFileWatcher itself, or excluded by child class)
- that file got deleted, triggering processWatchEvent
- Now people see this warning

It happens consistently with Ruby because .gem subdir is excluded and during addon initialisation, some lock files were created and then deleted, triggering a watchEvent.

Secondly - I couldn't see a memory / thread leak risk here.

In all places that call `removeFile`, they checked whether the entry they're removing exists inside scriptMap first before calling it. One place this didn't happen is inside processWatchEvent for DELETE event - and that's exactly when this happened.

Would you prefer that a check is done in processWatchEvent too, and leave the WARN logging there? I'd be happy with that solution as well, although I think it's redundant. Are there any other reasons that scriptMap.remove would return null other than the fact that the given element didn't exist?

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
jimtng: I thought it's redundant because:
- The directory creation is also done in WatchServiceImpl
- The directory creation is also done in `activate()` - this is also redundant here, but happy to leave it there to ensure this behaviour in case it changed in WatchServiceImpl.
- Prior to this PR, the directory creation is checked/performed inside `initialImport`

There is only a tiny window between activate() and the time the final call to `initialImport()` was made upon reaching startlevel 100 that an existing directory ensured by `activate()` got deleted. However, I'm happy to restore it if you think it's essential.

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
jimtng: The only reason I made it a separate method, is so that there was this check inside the constructor:
```
        if (currentStartLevel > StartLevelService.STARTLEVEL_MODEL) {
```
and I don't want the child class to need to know this implementation detail, and instead it should [just call `performInitialImport()`](https://github.com/openhab/openhab-addons/pull/14590/files#diff-0391e0939589717a418a22a3138a32b9fe3b36a31091bb5bfbb8cd5bd9938101R66).

I couldn't move it into `initialImport` because it's not always checked with this exact condition everywhere, just inside the constructor.

jimtng:
> I think we should better investigate why the ready markers arrive unordered.

I agree. However I just wanted my jruby addon to load so I can continue using 4.0-snapshot at this point. We could dig into the root cause of the out-of-order start level events later and remove the check here once we've fixed that.



> Or use the ready marker as trigger only to get the actual start level from the `StartLevelService`

I will try this.

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
```diff
@@ -225,25 +230,25 @@ private List<Path> listFiles(Path path, boolean includeSubDirectory) {
...
+                String scriptIdentifier = ScriptFileReference.getScriptIdentifier(fullPath);
+                if (scriptMap.containsKey(scriptIdentifier)) {
+                    removeFile(scriptIdentifier);
                 }
             }
-
-            if (file.canRead() && (kind == CREATE || kind == MODIFY)) {
-                addFiles(listFiles(file.toPath(), watchSubDirectories));
-            }
+        } else if (!file.isHidden() && file.canRead() && (kind == CREATE || kind == MODIFY)) {
```
jimtng, line 250: I moved the hidden file check here, because they should be checked during deletion (above), in case they got added previously. Otherwise they'd never be deleted.

jimtng:
> think we should better investigate why the ready markers arrive unordered. Or use the ready marker as trigger only to get the actual start level from the `StartLevelService`.

I've investigated this further. 

Ready markers do arrive in order - IF you registered your tracker early on, before the actual startlevels were reached. However, if you registered your tracker very late, say when startlevel was already at 80, ReadyServiceImpl::registerTracker() will catch you up on it, and send all the already reached startlevels all in one go. This is also why previously the import seems to have gotten done twice.

Because ReadyService doesn't know that it's a "startlevel" and supposed to be sorted, it just sends it unordered. This is the first time I looked at ReadyService, but I presumed there could be other Ready marker stuff that makes no sense to be sorted? If it needs to be sorted, it can be done in a separate PR. It doesn't affect this PR.

This PR just makes sure that it doesn't execute the code when it's out of order. Also checking that the import hasn't been done before calling initialImport stops us from doing initialimport twice here.

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
jimtng: I realised that the two checks are essentially the same, and I've combined the two functions into one and use the same check.

J-N-K:
> Ready markers do arrive in order - IF you registered your tracker early on, before the actual startlevels were reached. However, if you registered your tracker very late, say when startlevel was already at 80, ReadyServiceImpl::registerTracker() will catch you up on it, and send all the already reached startlevels all in one go. This is also why previously the import seems to have gotten done twice.
> 
> Because ReadyService doesn't know that it's a "startlevel" and supposed to be sorted, it just sends it unordered. This is the first time I looked at ReadyService, but I presumed there could be other Ready marker stuff that makes no sense to be sorted? If it needs to be sorted, it can be done in a separate PR. It doesn't affect this PR.

I checked and will provide a fix for that. The set that stores the received ready markers should retain the order (which it currently does not).

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
jimtng: I've tested deleting conf/automation/jsr223 then starting openhab up. It got recreated, and it monitors changes in the directory. I tried creating a script in there and it got (re)loaded. So I believe it's fine to remove it from initialImport.

jimtng:
Sorry for the late force-pushes. I believe it should be ready now.

> > Ready markers do arrive in order - IF you registered your tracker early on, before the actual startlevels were reached. However, if you registered your tracker very late, say when startlevel was already at 80, ReadyServiceImpl::registerTracker() will catch you up on it, and send all the already reached startlevels all in one go. This is also why previously the import seems to have gotten done twice.
> > Because ReadyService doesn't know that it's a "startlevel" and supposed to be sorted, it just sends it unordered. This is the first time I looked at ReadyService, but I presumed there could be other Ready marker stuff that makes no sense to be sorted? If it needs to be sorted, it can be done in a separate PR. It doesn't affect this PR.
> 
> I checked and will provide a fix for that. The set that stores the received ready markers should retain the order (which it currently does not).

Awesome, thanks! I don't think it makes a difference to this PR though. The checks here should still remain.

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
```diff
@@ -225,25 +230,25 @@ private List<Path> listFiles(Path path, boolean includeSubDirectory) {
...
-                    if (watchSubDirectories) {
-                        synchronized (this) {
-                            String prefix = fullPath.getParent().toString();
-                            Set<String> toRemove = scriptMap.keySet().stream().filter(ref -> ref.startsWith(prefix))
-                                    .collect(Collectors.toSet());
-                            toRemove.forEach(this::removeFile);
-                        }
+
+        if (kind == DELETE) {
+            if (file.isDirectory()) {
```
J-N-K, line 235: I wonder if this check is necessary. Did you check if we receive events for each file in the deleted directory? In that case we could ignore directory changes at all.

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
```diff
@@ -144,6 +146,9 @@ public void activate() {
         } else if (!Files.isDirectory(watchPath)) {
             logger.warn("Trying to watch directory {}, however it is a file", watchPath);
         }
+
+        manager.addFactoryChangeListener(this);
+        readyService.registerTracker(this, new ReadyMarkerFilter().withType(StartLevelService.STARTLEVEL_MARKER_TYPE));
```
J-N-K, line 151: Why did you move these?

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
J-N-K: Thanks, can you keep the original name?

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
```diff
@@ -378,16 +375,25 @@ public void onDependencyChange(String scriptIdentifier) {
 
     @Override
     public synchronized void onReadyMarkerAdded(ReadyMarker readyMarker) {
+        int newLevel = Integer.parseInt(readyMarker.getIdentifier());
+
+        if (newLevel <= currentStartLevel) {
+            logger.trace("onReadyMarkerAdded({}) - ignored {}", readyMarker, watchPath);
```
J-N-K, line 381:
```suggestion
            logger.trace("Ignored ready marker {} because it is for a lower start-level", readyMarker);
```

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
```diff
@@ -378,16 +375,25 @@ public void onDependencyChange(String scriptIdentifier) {
...
         if (currentStartLevel < StartLevelService.STARTLEVEL_STATES) {
             // ignore start level less than 30
             return;
         }
 
-        if (currentStartLevel == StartLevelService.STARTLEVEL_STATES) {
-            initialImport();
+        if (!ifInitialized().isDone()) {
+            performInitialImport();
         } else {
```
J-N-K, line 397: Nice

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
jimtng: Another update on this. Although adding the check to make sure that the given scriptIdentifier exists in scriptMap inside processWatchEvent DELETE does indeed avoid triggering that warning, I came across another case where that warning was erroneously issued despite it being perfectly fine:

During bundle termination (e.g. bundle:stop), removeFile was called twice: once by `factoryRemoved` and another time by `deactivate`. And because removeFile is async, the same scriptIdentifier gets included twice in the removal. The first one removed the file normally, unloading the scriptEngine and all. The second one would find that the scriptIdentifier no longer exists in scriptMap and then issued this warning.

Instead, I've changed the warning to be issued when the scriptIdentifier is still there when it shouldn't be.

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
jimtng: I've changed it to the original name.

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
jimtng: You are absolutely right! I did some tests and deleting a directory with scripts inside it would correctly issue a DELETE event for each file inside that directory, but not the directory itself. This is because it's filtered by WatchServiceImpl so processWatchEvent would never receive events about directories anyway, so yes this piece of code will never be reached.

I have removed it and also adjusted the corresponding test.

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
jimtng: Remember that readyService.registerTracker will catch you up on all the startlevel markers that had already been received/reached? Well, that would trigger initialImport right there if I left it in the constructor.

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
jimtng: I've changed it as suggested. However, I added the watchPath because when you have multiple instances of this class, you can't tell which is which otherwise.

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
```diff
@@ -378,16 +375,25 @@ public void onDependencyChange(String scriptIdentifier) {
...
+        if (newLevel <= currentStartLevel) {
+            logger.trace("onReadyMarkerAdded({}) - ignored {}", readyMarker, watchPath);
+            return;
+        }
+
         int previousLevel = currentStartLevel;
-        currentStartLevel = Integer.parseInt(readyMarker.getIdentifier());
+        currentStartLevel = newLevel;
+
+        logger.trace("onReadyMarkerAdded({}) {} => {} {}", readyMarker, previousLevel, currentStartLevel, watchPath);
```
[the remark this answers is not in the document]
jimtng: I've changed it as suggested. However, I added the watchPath because when you have multiple instances of this class, you can't tell which is which otherwise.

jimtng:
I've just realised, that we could avoid having two constructors by simply moving `initialImport` off the constructor and into `activate()`.

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
```diff
@@ -378,15 +358,26 @@ public void onDependencyChange(String scriptIdentifier) {
...
+        logger.trace("Added ready marker {}: start level changed from {} to {}. watchPath: {}", readyMarker,
+                previousLevel, currentStartLevel, watchPath);
 
         if (currentStartLevel < StartLevelService.STARTLEVEL_STATES) {
             // ignore start level less than 30
             return;
         }
 
-        if (currentStartLevel == StartLevelService.STARTLEVEL_STATES) {
+        if (!ifInitialized().isDone()) {
```
J-N-K, line 380:
```suggestion
        if (initialized.isDone()) {
```

J-N-K (commented):
Thanks, code looks good now. I'll test it in the next days and if I don't encounter problems we can merge it.

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
jimtng: Indeed. At first I only wanted to avoid the NPE and didn't want to change the existing code much. But you're correct, there's no need to do initialImport twice. I've removed it.

bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java
jimtng: changed as requested

jimtng:
> Thanks, code looks good now. I'll test it in the next days and if I don't encounter problems we can merge it.

Thanks for your help!

J-N-K (approved):

J-N-K:
Seems to work fine for me.

florian-h05:
Big thanks for that fix @jimtng, the error also affected JS Scripting!

PR #3455, opened 2023-03-14
Add dependency on jetty-alpn-java-client

andrewfg:
The `jetty-alpn-java-client` dependency was missing in #3433

### DIFF
--- a/features/karaf/openhab-tp/src/main/feature/feature.xml
+++ b/features/karaf/openhab-tp/src/main/feature/feature.xml
@@ -75,6 +75,7 @@
 		<feature dependency="true">pax-web-jetty-http2-jdk9</feature>
 		<bundle dependency="true">mvn:javax.servlet/javax.servlet-api/3.1.0</bundle>
 		<bundle dependency="true">mvn:org.eclipse.jetty/jetty-alpn-client/${jetty.version}</bundle>
+		<bundle dependency="true">mvn:org.eclipse.jetty/jetty-alpn-java-client/${jetty.version}</bundle>
 		<bundle dependency="true">mvn:org.eclipse.jetty/jetty-client/${jetty.version}</bundle>
 		<bundle dependency="true">mvn:org.eclipse.jetty/jetty-http/${jetty.version}</bundle>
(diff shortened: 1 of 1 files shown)


andrewfg:
@J-N-K / @wborn it is hard to know if this PR will actually deploy the missing Jar, without actually seeing a running snapshot, so do you think I should add 'runrequires' to make double sure?

wborn:
This PR will probably fix it in the Karaf distro. If you have issues in integration tests or when debugging in Ecliipse you'll probably need to add a `runrequires` for that.

Since I think you removed the integration tests for your Hue PR, you'll probably also not detect any future bundle dependency issues regarding HTTP/2 dependencies. E.g. an itest would be useful to find dependency regressions whenever Jetty refactors their bundles.

andrewfg:
> This PR will probably fix it in the Karaf distro

Ok. Good to know.

> I think you removed the integration tests for your Hue PR

Indeed. I think the problem is that the itest.bndrun changes in openhab-addons and the openhab-core snapshot distribution have not caught up with each other on jfrog yet (or something like that). So I am going to wait until some stability has arrived before I try again to re-integrate the hue itests in my PR.

> an itest would be useful to find dependency regressions whenever Jetty refactors their bundles

Can you please be more specific? I am not sure if you are suggesting a new itest in openhab-core, or one in the hue binding? I guess that one in the core would make more sense. And anyway I am not sure how to detect such a regression?

wborn:
I think it would be nice to have some integration tests for org.openhab.core.io.net. E.g. we could test that the client factories get properly injected and they can be used to create HTTP, HTTP/2 and Websocket clients and use each client on a server instantiated in the test.

wborn (approved):

andrewfg:
Ok, I will try to write some itests.

> test that the client factories get properly injected 

Is there a 'correct' way to test for proper injection? Or does one simply test the correct function of the tests below?

> can be used to create HTTP, HTTP/2 and Websocket clients

This sounds easy enough.

> use each client on a server instantiated in the test

The server being created via mockito, right?

PR #3457, opened 2023-03-14
Fix file processing in FileTransformationProvider

### LINKED ISSUE #3456: [SCRIPT] Attempting to use transform on Channel requires restart to see .script file
Forum thread: https://community.openhab.org/t/executing-the-script-transformation-failed-could-not-get-script-for-uid-filename-script/145251

I've seen this reported on 3.2.4 (https://community.openhab.org/t/scaling-of-voltage-reading-in-mqtt-reading/145242/) and experienced this on OH 4.0 #3368.

I only tried this with `graaljs` but it seems to be a problem with `dsl` too and likely all languages.

I only tried this with the MQTT binding but given where the error is coming from I suspect it affects all bindings (if I'm wrong this needs to move to the addons repo).

When adding a new transformation `.script` file, the SCRIPT transform will not see the file until OH is restarted. It reports the following error.

    2023-03-14 11:35:51.142 [WARN ] [t.generic.ChannelStateTransformation] - Executing the SCRIPT-transformation failed: Could not get script for UID 'logtrans.script'.

J-N-K:
Fixes #3456 

Relative and absolute path were not properly handled. The watch service provides relative paths to the root watch folder. The implementation expected absolute path (which were provided in the initial import).

This was not discovered by the tests because the tests wrongly modeled the watch service calls to `processWatchEvent`.

### DIFF
--- a/bundles/org.openhab.core.transform/src/main/java/org/openhab/core/transform/FileTransformationProvider.java
+++ b/bundles/org.openhab.core.transform/src/main/java/org/openhab/core/transform/FileTransformationProvider.java
@@ -14,6 +14,7 @@ package org.openhab.core.transform;
 
 import static org.openhab.core.service.WatchService.Kind.CREATE;
 import static org.openhab.core.service.WatchService.Kind.DELETE;
+import static org.openhab.core.service.WatchService.Kind.MODIFY;
 import static org.openhab.core.transform.Transformation.FUNCTION;
 
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.transform/src/main/java/org/openhab/core/transform/FileTransformationProvider.java
```diff
@@ -75,7 +75,7 @@ public FileTransformationProvider(
         watchService.registerListener(this, transformationPath);
         // read initial contents
         try {
-            Files.walk(transformationPath).filter(Files::isRegularFile)
+            Files.walk(transformationPath).filter(Files::isRegularFile).map(transformationPath::relativize)
```
wborn, line 78: Can you also use try-with-resources with it?

See the JavaDocs for `Files.walk`

```java
     * This method must be used within a try-with-resources statement or similar
     * control structure to ensure that the stream's open directories are closed
     * promptly after the stream's operations have completed.
```

It might also be an issue in other places where `Files.walk` is used and the stream is not closed.

J-N-K: Done. I have also improved the tests a little bit to make them more realistic (the mocked `WatchService` is now used in the same way as it is in the real-world application).

wborn (approved):
Thanks for the fix!

jimtng:
@openhab/core-maintainers could this be applied to 3.4.x too please?

PR #3458, opened 2023-03-15
[voice] Fix default interpreter selection.

GiviMAD:
This fixes the selection of the default human language interpreter.

By error it's loading the whole list of interpreters.

Should this be back-ported to the 3.4.x branch? I think the issue is also there.

Regards!


wborn (approved):
Thanks for the fix!

wborn:
> Should this be back-ported to the 3.4.x branch? I think the issue is also there.

I've :cherries:-picked it!

PR #172, opened 2017-07-29
HomeBuilder - Initial contribution

kubawolanin:
Here's my attempt to port [generator-openhab](https://github.com/kubawolanin/generator-openhab) functionality to the openHAB UI.

Closes #178

Discussion that led me to this can be found [here](https://community.openhab.org/t/openhab-generator-design-your-smart-home-in-no-time/).

~~#### *Very early stage*. ***Not for merging***~~
~~The code is ugly and doesn't work currently - submitting a PR for visibility.~~
~~Note: "openHAB Home Designer" title is temporary.~~

### Preview:
Here's a little demo for you to play around: **[home generator](http://kubawolanin.com/home-test/)**
It doesn't reflect the TODO list underneath. It's temporarily hosted on my server if that's ok. Will remove it after it's all merged.

generator

### TODO:

- [x] i18n
- [x] Styling + layout
- [x] Swap Bootstrap 4 stylesheet with [Material](http://daemonite.github.io/material/)
- [ ] Syntax highlight with [highlight.js](https://github.com/isagalaev/highlight.js) ([for vue](https://github.com/metachris/vue-highlightjs))
- [ ] Ability to dynamically add new rooms (per @ThomDietrich's request)
- [x] Quick 'Copy' button
- [x] REST API request construction for non-textual config ( `PUT /items` - requires https://github.com/eclipse/smarthome/issues/4060 to be implemented first)
- [x] Update documentation and README
- [x] Move changes to dashobard bundle to another PR
- [x] Create a separate bundle
- [x] Get rid of plural `Group`-items names in favor of `gSomething` (meh...)
- [x] Make the dynamic form produce real home structure
- [x] Sort out templating (from EJS to Vue templates)
- [x] Create a webpack-based build chain
- [x] Move all dependencies to `package.json`
- [x] Dynamic column formatting for Items (ascii-table)
- [x] Items preview
- [x] Sitemap preview
- [x] Ability to toggle `{ channel="" }` suffixes for non-Group items
- [x] Ability to toggle icons
- [x] `checklists` replaced with [vue-multiselect](http://monterail.github.io/vue-multiselect/)
- [x] Better Sitemaps design (a frame per floor with the room groups)
- [x] HABPanel JSON generator
- [x] Add optional HomeKit/HueEmulation/Alexa tags (see [Item tagging](http://docs.openhab.org/addons/io/homekit/readme.html#item-configuration))

And stuff that I  will need someone's help with:
[...]

PR #3460, opened 2023-03-16
Integration tests for org.openhab.core.io.net

andrewfg:
See https://github.com/openhab/openhab-core/pull/3455#issuecomment-1471634036


andrewfg:
@wborn you wanted the tests to verify if the http and web clients could connect to an actual server. But if we would try to connect to a real server on the Internet, this would make the tests potentially unreliable. An I could not figure out a way to simulate a server within an itest in any meaningful manner. 

However if you want to add to this test suite please feel free to do so ;)

wborn:
Thanks for having a go at this!

> I could not figure out a way to simulate a server within an itest in any meaningful manner.

There is an example of how to create a local Jetty HTTP server and use it for requests in the [WWNThingHandlerOSGiTest](https://github.com/wborn/openhab-addons/blob/5a104e4b5433fcf1af7ef5ef73d8ca384e16b17c/itests/org.openhab.binding.nest.tests/src/main/java/org/openhab/binding/nest/internal/wwn/handler/WWNThingHandlerOSGiTest.java#L131-L133).
I think some similar approach could be used to also create a server that supports HTTP/2 and WebSocket connections.

andrewfg:
> example of how to create a local Jetty HTTP server

I presume that the CI build is run on a virtual machine in the cloud? And I guess that such a local Jetty server would want to listen on 'localhost' or '127.0.0.1'? So is it given that a virtual machine in the cloud would allow an itest to connect to itself in that way?

andrewfg:
@wborn in order to itest the newly added HTTP/2 client we need an additional dependency on `org.eclipse.jetty.http2.server`. So given my past bad experience of messing with pom.xml, feature.xml, and itest.bndrun files, can you kindly advise me where this dependency should be added? Note it only needs to be added for compiling and running the itests, but probably should not be included in the OH runtime distribution.

wborn:
>  advise me where this dependency should be added?

It's similar to the `jetty-server` and `websocket-server` dependencies, so it should be fine to also add it to `bom/compile/pom.xml` and `bom/runtime/pom.xml`.

andrewfg:
@wborn I have been playing around with the itest code for most of today. So far I have been able to write the client code (which is actually what we want to test). And based on your example I was able to get the server code working to handle http/1.1 GET requests (via an HttpServlet). However I got really blocked in getting the server to handle http/2 GET requests (since the HttpServlet seems never to be called). Furthermore I was unable to get the server to handle WebSocket connection requests (which presumably need some other kind of Servlet). As Jetty is not well documented for such things, and since my work is not finally testing OH itself, (but rather testing some code that might simulate an external server that OH connect to), I am wondering if it is worth investing more time on this? => WDYT? (Perhaps I can submit the current code as WIP, in case others can fix the server code…)

wborn:
If you push the changes I can have a look at it. Jetty is very customizable and you can usually make it do anything you want. But I've also often ran into that it's not always clear how to use and customize it.

andrewfg:
> If you push the changes I can have a look at it

Ok @wborn I pushed my current work. If I modify the code to manually create the clients and run it as a local Junit test, then it passes the test. But if I run it as an OSGI test, (either locally or in CI tests), then it fails magnificently. .. and I don’t know why.

andrewfg:
@wborn the issue seems indeed to be due to missing dependencies. But I am unable to resolve it via `runrequires`; where there are two possible failure cases below..

---
### Failure Case 1
If `runrequires` contains only `org.eclipse.jetty.http2.hpack` then it fails due to missing `org.eclipse.jetty.websocket.server.WebSocketServerFactory`..

**itest.bndrun**
```
-runrequires: \
	bnd.identity;id='org.openhab.core.io.net.tests',\
	bnd.identity;id='org.eclipse.jetty.http2.hpack'
```
**Failure log**
```
org.eclipse.jetty.util [org.eclipse.jetty.server.handler.ContextHandler] WARN : unavailable
[1386](https://github.com/openhab/openhab-core/actions/runs/4462427379/jobs/7836956284?pr=3460#step:8:1387)
java.lang.RuntimeException: Unable to load org.eclipse.jetty.websocket.server.WebSocketServerFactory
```
---
### Failure Case 2
If `runrequires` contains both `org.eclipse.jetty.http2.hpack` and ``org.eclipse.jetty.websocket.server` then it fails due to missing `org.eclipse.jetty.http2.hpack`..

**itest.bndrun**
```
-runrequires: \
	bnd.identity;id='org.openhab.core.io.net.tests',\
	bnd.identity;id='org.eclipse.jetty.http2.hpack',\
	bnd.identity;id='org.eclipse.jetty.websocket.server'
```
**Failure log**
```
[INFO] Testing G:\git\andrewfg\openhab-core\itests\org.openhab.core.io.net.tests\itest.bndrun:
TEST org.openhab.core.io.net.tests.ClientFactoryTest#testHttp2Client() <<< ERROR: Jetty HTTP/2 hpack module not loaded
org.openhab.core.io.net.http.HttpClientInitializationException: Jetty HTTP/2 hpack module not loaded
```

joakime:
Note that Jetty 9.x is now at End of Community Support as well

* https://github.com/eclipse/jetty.project/issues/7958

andrewfg:
> It seems to work for me when adding the spifly bundle and then reresolving the runbundles

@wborn yes, it seems to work for me also -- but only when I run the tests on my local machine; but it seems to have failed in the CI build though..

```
public class ClientFactoryTest extends JavaOSGiTest {
..
line 107:  ContentResponse response = client.GET(httpUrl);
=> throws java.util.concurrent.ExecutionException: java.net.ConnectException: Connection refused
```

This makes me suspect that the VM where the CI build is running, probably has some kind of firewall blocking traffic to/from its `127.0.0.1` address. Or something like that...

andrewfg:
> When [KARAF-7544](https://issues.apache.org/jira/browse/KARAF-7544) gets resolved we should be able to hop on the Jetty 10 train

But that means these tests will once more not work again @wborn ...

> WebSocketExampleTest no longer works, as it uses `org.eclipse.jetty.websocket.api.annotations.WebSocket` and `org.eclipse.jetty.websocket.servlet.WebSocketServlet` which are not available in Jetty 10

And you have to keep in mind that any existing bindings with adapter classes that rely on thos `org.eclipse.jetty.websocket.api.annotations.WebSocket` annotations (e.g. my NeoHub binding) will be broken by this upgrade.

wborn:
It also sometimes fails for me locally. It's probably better if it doesn't the server in a new thread. Looks like some leftover from an [older TestServer implementation](https://github.com/eclipse-archived/smarthome/blob/56df0fa09d84912b96c7c8d1f229ca59938d4858/bundles/test/org.eclipse.smarthome.test/src/main/java/org/eclipse/smarthome/test/TestServer.java#L64-L103). It'll cause timing issues when the tests run and the server isn't fully started yet.

This might help fixing it: https://github.com/andrewfg/openhab-core/pull/2

> But that means these tests will once more not work again
> And you have to keep in mind that any existing bindings with adapter classes that rely on those org.eclipse.jetty.websocket.api.annotations.WebSocket annotations (e.g. my NeoHub binding) will be broken by this upgrade.

I think the upgrader will do a good job and make it all working again. :crossed_fingers:

andrewfg:
@wborn .. it still failed 1 out of 2 CI tests :(

wborn:
There is probably some kind of timing issue because it's succeeding now.

andrewfg:
> probably some kind of timing issue

Yeah. A hypothesis may be that we need to wait until the server is fully online. Perhaps try to run the GET test a couple of times with a delay until it either succeeds or the retry count is exceeded?

wborn:
If the server is not started it will throw:

```
java.util.concurrent.ExecutionException: java.net.ConnectException: Connection refused
	at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:396)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2096)
	at org.openhab.core.io.net.tests.ClientFactoryTest.testHttp2Client(ClientFactoryTest.java:149)
```

Increasing `WAIT_SECONDS` could also help because it may not always complete in 3 seconds in parallel builds because the CPU is busy doing something else.

andrewfg:
^
I have increased `WAIT_SECONDS` and applied a specific connect timeout to the tests. However the issue is probably not really about the connect timeout on the client, but rather that the server sends Connection Refused too fast (before it is fully initialized). So perhaps there is some server setting to delay sending Connection Refused??

itests/org.openhab.core.io.net.tests/src/main/java/org/openhab/core/io/net/tests/internal/TestServer.java
```diff
@@ -61,6 +65,17 @@ public void startServer() throws Exception {
...
+            public void lifeCycleStarted(@Nullable LifeCycle event) {
+                if (event != null && event.isStarted()) {
+                    isStarted.complete(true);
+                } else {
+                    isStarted.completeExceptionally(new IllegalStateException("Server not started"));
+                }
+            }
+        });
+
         server.start();
```
joakime, line 79: `Server.start()` won't return until the server is started.
This `LifeCycleListener` isn't doing anything.

What you need to be paying attention to is the WebApp deployed / started.

itests/org.openhab.core.config.core.tests/itest.bndrun
```diff
@@ -62,3 +62,4 @@ Fragment-Host: org.openhab.core.config.core
 	org.mockito.junit-jupiter;version='[4.11.0,4.11.1)',\
 	org.mockito.mockito-core;version='[4.11.0,4.11.1)',\
 	org.objenesis;version='[3.3.0,3.3.1)'
+
```
wborn, line 65:
```suggestion
```

features/karaf/openhab-core/src/main/feature/feature.xml
```diff
@@ -130,6 +130,12 @@
 		<bundle>mvn:org.openhab.core.bundles/org.openhab.core.io.http.auth/${project.version}</bundle>
 	</feature>
 
+	<feature name="openhab-core-io-net-test" version="${project.version}">
+		<feature>openhab-core-base</feature>
+		<bundle>mvn:org.eclipse.jetty.http2/http2-server/${jetty.version}</bundle>
+		<bundle>mvn:org.eclipse.jetty.http2/http2-hpack/${jetty.version}</bundle>
+	</feature>
+
```
wborn, line 138: This is not used in tests.

```suggestion
```

itests/org.openhab.core.io.net.tests/src/main/java/org/openhab/core/io/net/tests/internal/TestServer.java
andrewfg: > What you need to be paying attention to is the WebApp deployed / started.

@joakime thanks for the feedback. I guess I misunderstood that the lifecycle notification is only sent after everything in the server and all its nested connectors and handlers etc. are running. So can you please be more specific about 'checking if the WebApp is running'? In this case the 'applications' are servlets that are nested in a handler which is nested in the server (which is why I was looking at the lifecycle notification test). i.e. in other words, is it possible that something might be 'started' but eventually not actually 'running' resp. not yet processing incoming data??

andrewfg: .. or to put it another way: How can it be that we get 'connection refused' after the 'start()` method has returned? The server is obviously running, since otherwise the connection would passively timeout rather than being actively refused..

joakime: That would mean that the test is starting before the `Server.start()` has completed.
Should be a simple enough thing to double check.

Do you add your `Connectors` after `Server.start()`? (if yes, then you need to wait for those to reach "started" in their LifeCycle)

In Jetty, we test the real Jetty `Server` everywhere in our own unit tests.
We rely on the `LifeCycle.start()` behavior to behave consistently and predictably for this to work. (and it has, for 22 years)

If we deploy a webapp after the `Server.start()`, then we have to test for if that webapp is started before we can send requests to that webapp.

If we have a `DeploymentManager` doing the deployment, then we can add a `AppLifeCycle.Binding` to a binding target of `"deployed"` to trigger our tests to start making requests to that webapp.

If we add the `WebAppContext` or `ServletContextHandler` or `ContextHandler` to the server handler tree manually (not via a DeploymentManager), then we can either use a LifeCycleListener to tell us about the `lifeCycleStarted` event for that webapp, or we can periodically test the `LifeCycle.isStarted()` method against that webapp to know if it's started yet. (We usually use the LifeCycleListener on webapp, and a `java.util.concurrent.CountDownLatch` to trigger off the started event in our test cases)

andrewfg: Hmm. All connectors and servlets are added to the server before server.start() gets called. And (at least according to the itest @BeforeAll annotation) the client tests (annotated @Test) can (in theory) only start after the server.start() method (in @BeforeAll) has completed. However in reality sometimes we see a 'connection refused' result from the first test. .. I suppose there could be a threading issue whereby an @Test method is called on one thread, before the @BeforeAll method on another thread has fully completed. Or domething like that. I will try adding a synchronized lock to prevent that.

joakime: It could also be a test that's still running when your `@AfterAll` runs and shuts down/stops the Server.

andrewfg: ^
Good point.

andrewfg: @joakime many thanks for the guidance. Through it I have learned a lot about Jetty (albeit not even approaching your own knowledge). I added some Completable's to ensure correct pipelining of @BeforeAll => @Test => @AfterAll methods.

andrewfg:
@wborn two things concerning CI build errors..
- This time the Java 17 test passed, but Jenkins failed. So at least that is different from prior 1 out of 2 failures. But unfortunately the Jenkins console output gives no clue about why.
- I accepted two of your proposed changes directly in Git UI, so the commits are effectively without my DSig, so the DCO test also failed. Hopefully you can override that.

andrewfg:
^
PS I tested my local build with `mvn clean test -DskipChecks -T 2C` ..

andrewfg:
.. but this last time around, Jenkins passed, and the Java 17 build failed; I think the CI build environment is seriously munged, because the build results are essentialy unpredictable.

andrewfg:
^
I have been doing some Googling and it might be that the `@NotThreadSafe` class annotation might fix this. I will try it tomorrow..

(Alternative might be to concatenate all test code and before/after code into a single long method, which would not look elegant, but would probably work..)

wborn:
There's really nothing special about the computers used for CI. :slightly_smiling_face: If you'd run the tests on a computer with similar specs (or another OS) you'd also more likely run into these timing issues. You'll also often run into such issues on your own computer (with better specs) if you run the tests while putting more load on your CPU.

> I have been doing some Googling and it might be that the @NotThreadSafe class annotation might fix this. 
> (Alternative might be to concatenate all test code and before/after code into a single long method, which would not look elegant, but would probably work..)

I don't think any of this will help.

The `Completable`s also  have a [failed](https://www.eclipse.org/jetty/javadoc/jetty-9/org/eclipse/jetty/util/Promise.Completable.html#failed(java.lang.Throwable)) method which is called whenever exceptions occur. Maybe you can use that to figure out the root cause?

andrewfg:
@wborn the solution -- albeit as the Germans say _'shooting at sparrows with cannons'_ -- seems to be as follows..

I have pipelined the initialization method `beforeAll()`, the three test methods, and the shutdown method `afterAll()`, in series through a single `@Test` method call, and synchronized this method so that only one such test series method can run at a time, either within a single class instance, or across multiple class instances running within the same JVM.

Note: even this cannot synchronize across multiple JVMs running on the same (virtual) machine! i.e. if there are multiple JVMs running on one VM, they could still try to access the machine's common `localHost` IP stack at the same time.

So (in other words) if..
1. Neither the Jenkins test nor the Java 17 build use multiple JVM's, AND 
2. Neither the Jenkins test nor the Java 17 build are running on the same VM, then..

=> Probably the tests should reliably succeed.. :)

wborn:
> Probably the tests should reliably succeed.. :)

Let's test it another time. 😄

wborn:
I've added some debug info and done some cleanup (https://github.com/wborn/openhab-core/commit/09196328069b94f3b6fde67f4eada63f2705dba4) and will check what happens in GHA CI builds.

itests/org.openhab.core.io.net.tests/src/main/java/org/openhab/core/io/net/tests/ClientFactoryTest.java
```diff
@@ -0,0 +1,224 @@
...
+            assertNotNull(session);
+            assertFalse(session.isClosed());
+
+            // open stream
+            session.newStream(headers, streamCompletable, streamAdapter);
+            Stream stream = streamCompletable.get(WAIT_SECONDS, TimeUnit.SECONDS);
+            assertNotNull(stream);
+
+            // confirm stream is open
+            assertFalse(stream.isClosed());
```
wborn, line 172: Why would the stream always be open here?

The `HeadersFrame` is created such that it should end the stream:

`HeadersFrame headers = new HeadersFrame(request, null, true);`

```java
    /**
     * <p>Creates a new {@code HEADERS} frame with an unspecified stream {@code id}.</p>
     * <p>The stream {@code id} will be generated by the implementation while sending
     * this frame to the other peer.</p>
     *
     * @param metaData the metadata containing HTTP request information
     * @param priority the PRIORITY frame associated with this HEADERS frame
     * @param endStream whether this frame ends the stream
     */
    public HeadersFrame(MetaData metaData, PriorityFrame priority, boolean endStream)
```

Now it isn't closed immediately either. But that may be an implementation detail. :wink: 

If I wait for it to be closed using:

`waitForAssert(() -> assertTrue(stream.isClosed()));`

it does get closed eventually.

andrewfg:
> Why would the stream always be open here?
> The HeadersFrame is created such that it should end the stream:

@wborn thanks for finding this. Yes, this is indeed the cause of the problem! .. But perhaps not for the reason that you imply. The issue is NOT that the client's request HeadersFrame causes the client write channel of the stream to be closed (sooner or later). The issue IS that the server has already quickly written its response to the server write channel of the stream, and thus already quickly closed it.

On a PC with a real network card, the round trip of sending a client request to the hardware and getting a server response from it, takes an appreciable period of time, during which the assertion at line #172 WILL pass, because the stream is indeed still open.  By contrast on a VM the round trip between client request and server response apparently takes less time, so the response will already have been received, and the stream therefore closed, BEFORE line #172 will have been encountered.

andrewfg:
@wborn the Java 17 build failed yet again.. but the error was `[INFO] openHAB Core :: Integration Tests :: Voice Test .... FAILURE [ 27.192 s]` .. so that is not my problem.

joakime:
> It's using `localhost` so I would expect the OS networking logic to be smart enough not to bother any networking hardware with this.

If you test on a newer JVM with SSL/TLS, the use of "localhost" is unsupported for anything related to Certificates, TLS/1.2, ALPN, SNI, etc.
eg: Your Certificates cannot contain localhost (in any form), or IP literals.
TLS/1.2+ will not negotiate with various important extensions if you use localhost or IP literals (such as SNI)
ALPN will reject use of localhost or IP literals for various protocol negotiations as well.

I really don't know how you test, but this is a common "gotcha".
Also, if your CI tests on a docker/containerd/kubernetes, you'll have issues if you let your JVM have IPv6 enabled, or you want to test anything that uses UDP (eg: HTTP/3).

andrewfg:
@joakime many thanks for the advice; in this test suite we only use plaintext, so fortunately its not an issue here; (but will bear it in mind for the future..)

itests/org.openhab.core.config.core.tests/itest.bndrun
```diff
@@ -61,4 +61,4 @@ Fragment-Host: org.openhab.core.config.core
 	net.bytebuddy.byte-buddy-agent;version='[1.12.19,1.12.20)',\
 	org.mockito.junit-jupiter;version='[4.11.0,4.11.1)',\
 	org.mockito.mockito-core;version='[4.11.0,4.11.1)',\
-	org.objenesis;version='[3.3.0,3.3.1)'
+	org.objenesis;version='[3.3.0,3.3.1)'
```
wborn, line 64: Can you revert this change?

wborn (commented):
>  But if you have a better explanation, please feel to share it..

I don't, but if build succeeds without these assertions that's fine with me. :+1: 

> build failed yet again

There's #3477 for that now.

wborn (approved):
LGTM now and it's nice to have some tests for these often used clients now! :+1: 
Since it also some of my own contributions, it would be nice if one of the other @openhab/core-maintainers can have a second look.

itests/org.openhab.core.io.net.tests/src/main/java/org/openhab/core/io/net/tests/ClientFactoryTest.java
```diff
@@ -0,0 +1,189 @@
...
+@NonNullByDefault
+public class ClientFactoryTest extends JavaOSGiTest {
+    public static final String RESPONSE = "response";
+
+    private static final String CONSUMER = "consumer";
+    private static final String HOST = "127.0.0.1";
+    private static final int WAIT_SECONDS = 10;
+    private static final long CONNECTION_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(WAIT_SECONDS);
+
+    private static @Nullable TestServer server;
```
J-N-K, line 63: In other tests we use `@NonNullByDefault({})` for fields that are initialized in `@BeforeEach` and `@BeforeAll`. You also don't need the `getServer` method then. I there is nothing else that speaks against using that, I would suggest to do the same here for consistency. Otherwise LGTM.

wborn: > @NonNullByDefault({})

I tried that but it does not work with static fields.

J-N-K (approved):

itests/org.openhab.core.io.net.tests/src/main/java/org/openhab/core/io/net/tests/ClientFactoryTest.java
andrewfg: Yup. And I tried it before you. .. :)

Issue #3257, opened 2022-12-23
AutomationIntegrationJsonTest unstable


wborn:
This test failed in a Jenkins build: https://ci.openhab.org/job/openHAB-Core/1107/

```
TEST org.openhab.core.automation.integration.test.AutomationIntegrationJsonTest#assertThatARuleFromJsonFileIsExecutedCorrectly() <<< ERROR: 
Expected: is not null
     but: was null
java.lang.AssertionError: 
Expected: is not null
     but: was null
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
	at org.openhab.core.automation.integration.test.AutomationIntegrationJsonTest.lambda$17(AutomationIntegrationJsonTest.java:383)
	at org.openhab.core.test.java.JavaTest.waitForAssert(JavaTest.java:245)
	at org.openhab.core.test.java.JavaTest.waitForAssert(JavaTest.java:213)
	at org.openhab.core.test.java.JavaTest.waitForAssert(JavaTest.java:166)
	at org.openhab.core.automation.integration.test.AutomationIntegrationJsonTest.assertThatARuleFromJsonFileIsExecutedCorrectly(AutomationIntegrationJsonTest.java:382)
```

wborn:
Another test in this class failed today, see: https://ci.openhab.org/job/openHAB-Core/1135

```
TEST org.openhab.core.automation.integration.test.AutomationIntegrationJsonTest#assertThatARuleFromJsonFileIsAddedAutomaticallyAndTheRuntimeRuleHasResolvedModuleReferences() <<< ERROR: 
Expected: is not null
     but: was null
java.lang.AssertionError: 
Expected: is not null
     but: was null
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
	at org.openhab.core.automation.integration.test.AutomationIntegrationJsonTest.lambda$15(AutomationIntegrationJsonTest.java:325)
	at org.openhab.core.test.java.JavaTest.waitForAssert(JavaTest.java:245)
	at org.openhab.core.test.java.JavaTest.waitForAssert(JavaTest.java:213)
	at org.openhab.core.test.java.JavaTest.waitForAssert(JavaTest.java:177)
[...]

PR #3461, opened 2023-03-17
[audio] More capabilities for AudioSink using the AudioServlet

dalgwen:
Hello,

Following discussions with @lolodomo, I try here to enhance some aspect of the audio framework

- This proposal extracts `ClonableAudioStream` from `FixedLengthAudioStream`. We can have clonable Stream but without the fixed length property (example : `URLAudioStream`).

- Some (`AudioServlets` ?) sink can only work with `FixedLengthAudioStream `(at least squeezebox and webaudio sinks). Because they need to clone a stream to serve several client, or a client several times. To lift this limitation, this proposal make the `AudioServlet` able to cache the whole stream in memory (if < 1MB)  or in file (if <5 MB) to serve the sink a fitting stream by wrapping it with an implementation of `FixedLenghtAudioStream` (either `ByteArrayAudioStream` or `FileAudioStream`)

- Some AudioSink services can work asynchronously. It means that the process method returns before the sound is played entirely. It is an issue because of the volume restoration functionality that doesn't wait. For example, if we play a sound in the `ChromecastAudioSink` with a specific volume :  the volume is fixed at the desired level, the sound is served on the `AudioServlet`, the request is sent to the Chromecast device, and the volume level is immediately restored to the previous value **without waiting the stream consumption**.
~~This proposal adds a `Runnable` argument to the process method. This `Runnable` has the volume restoration code inside. It should be the sink's responsibility to execute this code when it has done playing.~~ This proposal adds a return value to the process method : a CompletableFuture which should be completed when the sound is fully played. It is the responsibility of the sink to complete this Future.
To not break compatibility or change behaviour, there is now a default process method in the `AudioSink` interface : it plays the sound in the sink and ~~launches the `Runnable`~~ complete the `CompletableFuture` immediately after the process method returns (exactly as before).
To change this behaviour, the `AudioSinks` has to override the new process method to handle the ~~`Runnable`~~ `CompletableFuture` at the proper time.
To ease further developement, I added two abstract classes : `AudioSinkSync` and `AudioSinkAsync`. I also provided two examples with the openHAB included sinks. ~~The `JavaSoundAudioSink` now extends the synced one~~, the `JavaSoundAudioSink` and the `WebAudioSink` (with the help of a now aware `AudioServlet`) extends the asynchronous one. 
~~The `AudioServlet` implementation is not very good, as it considers that a stream is ended when all bytes have been transfered and it is timed out, but it is still better than before. To further enhance this, we could maybe add some kind of calculation to compute the sound duration (as it is actually done by the `PulseAudioSink`, for this purpose). But it should be another PR.~~
An utility method to compute sound duration is included, with no guarantee of success (better than nothing). The AudioServlet makes use of it to know when to expire a stream and complete the `CompletableFuture`

- Some TTS services (at least pico tts, probably others), create temporary files to store TTS result before sending it to the framework. Some even makes temporary file to work around the issue with `AudioServlet` and `FixedLengthAudioStream`, mainly MimicTTS (This should not be necessary and could be deleted if this PR is merged) It also puts a need to delete file after use. We can use the java temporary file method, as some already do, but it delays the deletion to the end of the JVM, and it is not fool proof. This proposal adds a callback method for deleting file when they are not needed, with a Disposable interface.
Deleting the temporary file is the sink's responsibility.
To ease the development and mutualize code, the `AudioSinkSync` and `AudioSynkAsync` classes has, for one, this capability (Sync), and, for the other, an utility method (Async).

- Adding an identifyier to audio stream for further development. This will allow, for example, audio sink to cache data related to a stream, and not compute it everytime. I have a use case in mind with some sinks (`PulseAudio` and `Doorbird`) that do some encoding.

### DIFF
--- a/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioSink.java
+++ b/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioSink.java
@@ -87,12 +87,16 @@ public interface AudioSink {
      * interface, the sink should hereafter get rid of it by calling the dispose method.
      *
      * @param audioStream the audio stream to play or null to keep quiet
-     * @throws UnsupportedAudioFormatException If audioStream format is not supported
-     * @throws UnsupportedAudioStreamException If audioStream is not supported
+     * @return A future completed when the sound is fully played. The method can instead complete with
+     *         UnsupportedAudioFormatException if the audioStream format is not supported, or
+     *         UnsupportedAudioStreamException If audioStream is not supported
      */
-    default CompletableFuture<@Nullable Void> processAndComplete(@Nullable AudioStream audioStream)
-            throws UnsupportedAudioFormatException, UnsupportedAudioStreamException {
-        process(audioStream);
+    default CompletableFuture<@Nullable Void> processAndComplete(@Nullable AudioStream audioStream) {
+        try {
+            process(audioStream);
+        } catch (UnsupportedAudioFormatException | UnsupportedAudioStreamException e) {
+            return CompletableFuture.failedFuture(e);
+        }
         return CompletableFuture.completedFuture(null);
     }
 
--- a/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioSink.java
+++ b/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioSink.java
@@ -58,8 +58,8 @@ public interface AudioSink {
      *
      * In case the audioStream is null, this should be interpreted as a request to end any currently playing stream.
      *
-     * If you call this method and if the sink is synchronous, you should thereafter get rid of a stream implementing
-     * the {@link org.openhab.core.common.Disposable} interface by calling the dispose method.
+     * When the process method ends, if the stream implements the {@link org.openhab.core.common.Disposable} interface,
+     * the sink should hereafter get rid of it by calling the dispose method.
      *
      * @param audioStream the audio stream to play or null to keep quiet
      * @throws UnsupportedAudioFormatException If audioStream format is not supported
@@ -67,6 +67,35 @@ public interface AudioSink {
      */
     void process(@Nullable AudioStream audioStream)
             throws UnsupportedAudioFormatException, UnsupportedAudioStreamException;
+
+    /**
+     * Processes the passed {@link AudioStream}, and executes the whenFinished Runnable (or stores for later execution
+     * if the sink is asynchronous).
+     *
+     * If the passed {@link AudioStream} is not supported by this instance, an {@link UnsupportedAudioStreamException}
+     * is thrown.
+     *
+     * If the passed {@link AudioStream} has an {@link AudioFormat} not supported by this instance,
+     * an {@link UnsupportedAudioFormatException} is thrown.
+     *
+     * In case the audioStream is null, this should be interpreted as a request to end any currently playing stream.
+     *
+     * When the process method ends, if the stream implements the {@link org.openhab.core.common.Disposable} interface,
+     * the sink should hereafter get rid of it by calling the dispose method.
+     *
+     * @param audioStream the audio stream to play or null to keep quiet
+     * @param whenFinished A Runnable to run when the sound is finished playing.
+     * @throws UnsupportedAudioFormatException If audioStream format is not supported
+     * @throws UnsupportedAudioStreamException If audioStream is not supported
+     */
+    default void process(@Nullable AudioStream audioStream, Runnable whenFinished)
+            throws UnsupportedAudioFormatException, UnsupportedAudioStreamException {
+        try {
+            process(audioStream);
+        } finally {
+            whenFinished.run();
+        }
+    }
 
     /**
      * Gets a set containing all supported audio formats
@@ -97,17 +126,4 @@ public interface AudioSink {
      * @throws IOException if the volume can not be set
      */
     void setVolume(PercentType volume) throws IOException;
-
-    /**
-     * Tell if the sink is synchronous.
-     * If true, caller may dispose of the stream immediately after the process method.
-     * On the contrary, if in the process method, the sink returns before the input stream is entirely consumed,
-     * then the sink should override this method and return false.
-     * Please note that by doing so, the sink should then take care itself of the InputStream implementing the
-     * {@link org.openhab.core.common.Disposable} interface by calling the dispose method when finishing
-     * reading it.
-     */
-    default boolean isSynchronous() {
-        return true;
-    }
 }
--- a/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioSink.java
+++ b/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioSink.java
@@ -58,6 +58,9 @@ public interface AudioSink {
      *
      * In case the audioStream is null, this should be interpreted as a request to end any currently playing stream.
      *
+     * If you call this method and if the sink is synchronous, you should thereafter get rid of a stream implementing
+     * the {@link org.openhab.core.common.Disposable} interface by calling the dispose method.
+     *
      * @param audioStream the audio stream to play or null to keep quiet
      * @throws UnsupportedAudioFormatException If audioStream format is not supported
      * @throws UnsupportedAudioStreamException If audioStream is not supported
@@ -94,4 +97,17 @@ public interface AudioSink {
      * @throws IOException if the volume can not be set
      */
     void setVolume(PercentType volume) throws IOException;
+
+    /**
+     * Tell if the sink is synchronous.
+     * If true, caller may dispose of the stream immediately after the process method.
+     * On the contrary, if in the process method, the sink returns before the input stream is entirely consumed,
+     * then the sink should override this method and return false.
+     * Please note that by doing so, the sink should then take care itself of the InputStream implementing the
+     * {@link org.openhab.core.common.Disposable} interface by calling the dispose method when finishing
+     * reading it.
+     */
+    default boolean isSynchronous() {
+        return true;
+    }
 }
--- a/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioSink.java
+++ b/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioSink.java
@@ -15,6 +15,7 @@ package org.openhab.core.audio;
 import java.io.IOException;
 import java.util.Locale;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
@@ -58,19 +59,21 @@ public interface AudioSink {
      *
      * In case the audioStream is null, this should be interpreted as a request to end any currently playing stream.
      *
-     * When the process method ends, if the stream implements the {@link org.openhab.core.common.Disposable} interface,
-     * the sink should hereafter get rid of it by calling the dispose method.
+     * When the stream is not needed anymore, if the stream implements the {@link org.openhab.core.common.Disposable}
+     * interface, the sink should hereafter get rid of it by calling the dispose method.
      *
      * @param audioStream the audio stream to play or null to keep quiet
      * @throws UnsupportedAudioFormatException If audioStream format is not supported
      * @throws UnsupportedAudioStreamException If audioStream is not supported
+     * @deprecated Use {@link AudioSink#processAndComplete(AudioStream)}
      */
+    @Deprecated
     void process(@Nullable AudioStream audioStream)
             throws UnsupportedAudioFormatException, UnsupportedAudioStreamException;
 
     /**
-     * Processes the passed {@link AudioStream}, and executes the whenFinished Runnable (or stores for later execution
-     * if the sink is asynchronous).
+     * Processes the passed {@link AudioStream}, and returns a CompletableFuture that should complete when the sound is
+     * fully played. It is the sink responsibility to complete this future.
      *
      * If the passed {@link AudioStream} is not supported by this instance, an {@link UnsupportedAudioStreamException}
      * is thrown.
@@ -80,23 +83,17 @@ public interface AudioSink {
      *
      * In case the audioStream is null, this should be interpreted as a request to end any currently playing stream.
      *
-     * When the process method ends, if the stream implements the {@link org.openhab.core.common.Disposable} interface,
-     * the sink should hereafter get rid of it by calling the dispose method.
+     * When the stream is not needed anymore, if the stream implements the {@link org.openhab.core.common.Disposable}
+     * interface, the sink should hereafter get rid of it by calling the dispose method.
      *
      * @param audioStream the audio stream to play or null to keep quiet
-     * @param whenFinished A Runnable to run when the sound is finished playing.
      * @throws UnsupportedAudioFormatException If audioStream format is not supported
      * @throws UnsupportedAudioStreamException If audioStream is not supported
      */
-    default void process(@Nullable AudioStream audioStream, @Nullable Runnable whenFinished)
+    default CompletableFuture<@Nullable Void> processAndComplete(@Nullable AudioStream audioStream)
             throws UnsupportedAudioFormatException, UnsupportedAudioStreamException {
-        try {
-            process(audioStream);
-        } finally {
-            if (whenFinished != null) {
-                whenFinished.run();
-            }
-        }
+        process(audioStream);
+        return CompletableFuture.completedFuture(null);
     }
 
     /**
--- a/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioSink.java
+++ b/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioSink.java
@@ -88,12 +88,14 @@ public interface AudioSink {
      * @throws UnsupportedAudioFormatException If audioStream format is not supported
      * @throws UnsupportedAudioStreamException If audioStream is not supported
      */
-    default void process(@Nullable AudioStream audioStream, Runnable whenFinished)
+    default void process(@Nullable AudioStream audioStream, @Nullable Runnable whenFinished)
             throws UnsupportedAudioFormatException, UnsupportedAudioStreamException {
         try {
             process(audioStream);
         } finally {
-            whenFinished.run();
+            if (whenFinished != null) {
+                whenFinished.run();
+            }
         }
     }
 
--- a/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioSinkAsync.java
+++ b/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioSinkAsync.java
@@ -1,0 +1,86 @@
+/**
+ * Copyright (c) 2010-2023 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.audio;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+import org.openhab.core.common.Disposable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Definition of an audio output like headphones, a speaker or for writing to
+ * a file / clip.
+ * This version is asynchronous: when the process() method returns, the {@link AudioStream}
+ * may or may not be played, and we don't know when the delayed task will be executed.
+ * CAUTION : It is the responsibility of the implementing AudioSink class to call the runDelayedTask
+ * method when playing is done.
+ *
+ * @author Gwendal Roulleau - Initial contribution
+ */
+@NonNullByDefault
+public abstract class AudioSinkAsync implements AudioSink {
+
+    private final Logger logger = LoggerFactory.getLogger(AudioSinkAsync.class);
+
+    private final Map<AudioStream, Runnable> runnableByAudioStream = new HashMap<>();
+
+    @Override
+    public void process(@Nullable AudioStream audioStream, Runnable whenFinished)
+            throws UnsupportedAudioFormatException, UnsupportedAudioStreamException {
+
+        try {
+            if (audioStream != null) {
+                runnableByAudioStream.put(audioStream, whenFinished);
+            }
+            process(audioStream);
+        } finally {
+            if (audioStream == null) {
+                // No need to delay the post process task
+                whenFinished.run();
+            }
+        }
+    }
+
+    /**
+     * Will run the delayed task stored previously.
+     *
+     * @param audioStream The AudioStream is the key to find the delayed Runnable task in the storage.
+     */
+    protected void runDelayed(AudioStream audioStream) {
+        Runnable delayedTask = runnableByAudioStream.remove(audioStream);
+
+        if (delayedTask != null) {
+            delayedTask.run();
+        }
+
+        // if the stream is not needed anymore, then we should call back the AudioStream to let it a chance
+        // to auto dispose:
+        if (audioStream instanceof Disposable disposableAudioStream) {
+            try {
+                disposableAudioStream.dispose();
+            } catch (IOException e) {
+                String fileName = audioStream instanceof FileAudioStream file ? file.toString() : "unknown";
+                if (logger.isDebugEnabled()) {
+                    logger.debug("Cannot dispose of stream {}", fileName, e);
+                } else {
+                    logger.warn("Cannot dispose of stream {}, reason {}", fileName, e.getMessage());
+                }
+            }
+        }
+    }
+}
--- a/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioSinkAsync.java
+++ b/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioSinkAsync.java
@@ -40,16 +40,16 @@ public abstract class AudioSinkAsync implements AudioSink {
     private final Map<AudioStream, Runnable> runnableByAudioStream = new HashMap<>();
 
     @Override
-    public void process(@Nullable AudioStream audioStream, Runnable whenFinished)
+    public void process(@Nullable AudioStream audioStream, @Nullable Runnable whenFinished)
             throws UnsupportedAudioFormatException, UnsupportedAudioStreamException {
 
         try {
-            if (audioStream != null) {
+            if (audioStream != null && whenFinished != null) {
                 runnableByAudioStream.put(audioStream, whenFinished);
             }
             process(audioStream);
         } finally {
-            if (audioStream == null) {
+            if (audioStream == null && whenFinished != null) {
                 // No need to delay the post process task
                 whenFinished.run();
             }
--- a/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioSinkAsync.java
+++ b/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioSinkAsync.java
@@ -42,7 +42,6 @@ public abstract class AudioSinkAsync implements AudioSink {
     @Override
     public void process(@Nullable AudioStream audioStream, @Nullable Runnable whenFinished)
             throws UnsupportedAudioFormatException, UnsupportedAudioStreamException {
-
         try {
             if (audioStream != null && whenFinished != null) {
                 runnableByAudioStream.put(audioStream, whenFinished);
--- a/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioSinkAsync.java
+++ b/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioSinkAsync.java
@@ -15,6 +15,7 @@ package org.openhab.core.audio;
 import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.CompletableFuture;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
@@ -25,10 +26,10 @@ import org.slf4j.LoggerFactory;
 /**
  * Definition of an audio output like headphones, a speaker or for writing to
  * a file / clip.
- * This version is asynchronous: when the process() method returns, the {@link AudioStream}
- * may or may not be played, and we don't know when the delayed task will be executed.
- * CAUTION : It is the responsibility of the implementing AudioSink class to call the runDelayedTask
- * method when playing is done.
+ * Helper class for asynchronous sink : when the process() method returns, the {@link AudioStream}
+ * may or may not be played. It is the responsibility of the implementing AudioSink class to
+ * complete the CompletableFuture when playing is done. Any delayed tasks will then be performed, such as volume
+ * restoration.
  *
  * @author Gwendal Roulleau - Initial contribution
  */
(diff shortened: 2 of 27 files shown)


wborn:
Does this still allow for mulitple clients to consume the same audio stream?
That might happen for instance when playing audio on the webaudio sink.

dalgwen:
> Does this still allow for mulitple clients to consume the same audio stream? That might happen for instance when playing audio on the webaudio sink.

Indeed.
Previously, the webaudio sink could only consume FixedLengthAudioStream served by the AudioServlet
With this pull request, the AudioServlet can receive a generic AudioStream, will build a FixedLengthAudioStream from it, and serve it. The FixedLengthAudioStream are either FileAudioStream or ByteArrayAudioStream (depending on size, ByteArrayAudioStream is best to avoid writing on disk/sdcard/whatever), and both of them implement the getClonedStream() method wich allow several client to consume the same source stream.

dalgwen:
The **PR-openHAB-Core** task is failing but I don't know why. The build is successfull on my environment.
I don't see anything strange in the log.
Could someone point me the issue ?

wborn:
This is the issue https://github.com/openhab/openhab-core/issues/3257

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/AudioServlet.java
```diff
@@ -195,13 +228,42 @@ public String serve(AudioStream stream) {
...
+        FixedLengthAudioStream fixedLengthAudioStreamResult;
+        if (dataBytes.length <= ONETIME_STREAM_BUFFER_MAX_SIZE) {
+            // we will use an in memory buffer to avoid disk operation
+            fixedLengthAudioStreamResult = new ByteArrayAudioStream(dataBytes, stream.getFormat());
+        } else {
+            // in memory max size exceeded, sound is too long, we will use a file
+            File tempFile = File.createTempFile(streamId, ".snd");
+            tempFile.deleteOnExit();
+            try (OutputStream outputStream = new FileOutputStream(tempFile)) {
+                outputStream.write(dataBytes);
```
lolodomo, line 258: This will only write the first ONETIME_STREAM_BUFFER_MAX_SIZE + 1 bytes and not all ?

lolodomo:
What about all our existing audio sinks implemented through bindings ? Are they all synchronous (the new default) ? Will they require a change ?

lolodomo:
I believe we will have to update all the audio sinks. For example, the Sonos audio sink is only supporting URLAudioStream and FixedLengthAudioStream. FixedLengthAudioStream is then served through our HTTP audio server.
I understand that we have now to support AudioStream, the HTTP audio server will create the appropriate FixedLenghtAudioStream.
The Sonos audio sink then call either playURI for URLAudioStream or playNotificationSoundURI for FixedLengthAudioStream. playURI is asynchronous while playNotificationSoundURI  is synchronous. So it looks like isSynchronous should then return true or false depending on the kind of handled audio streams ?

lolodomo:
Or maybe `isSynchronous` is only important when using the HTTP audio server ?

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/AudioServlet.java
dalgwen: To avoid using too much memory, I indeed set a limit. But only for the in-memory buffer.

If the limit is reached, then the 1MB memory buffer is copied to a file and the remaining of the stream is also copied to it.

lolodomo: Ok, this is done by the next line (stream.transferTo)

dalgwen:
You are right, things are getting complicated.
I thought the synchronous way of doing was nearly everywhere. For example, in the core AudioManagerImpl, after the call to the audiosink process method, there is some code to reset the audio volume to the previous value.
I always thought this implied that the AudioSink has to be preferentially synchronous. I even made a patch for the pulseaudiosink to be synchronous to handle this use case.

But it is indeed more complex than that. 
I'm now thinking that asynchronous sinks should make a callback to the core framework for it to handle stream disposal and volume reset (at least).

lolodomo:
We can check all existing audio sink implementations. Here are the list:

- allplay
- amplipi
- chromecast
- doorbird
- freebox
- heos
- kodi
- onkyo
- pulseaudio
- sonos
- squeezeboc
- upnpcontrol

When the sink is handling a URLAudioStream, this is certainly always asynchronously and there is certainly nothing special to do (no file to delete). I have the feeling your enhancement only concerns other audio strreams when the HTTP audio server is used ?

lolodomo:
> But it is indeed more complex than that.
> I'm now thinking that asynchronous sinks should make a callback to the core framework for it to handle stream disposal and volume reset (at least).
> It would be a breaking change. Should I work in this direction ?

I am not sure.
Maybe we should first analyse the different existing audio sinks. Their implementation is probably very similar.
Doorbird and pulseaudio may be different from others.

If I have correctly searched, URLAudioStream is only used when stream (audio manager) is called and in that case there is no volume management. So it is not a problem if sink process is asynchronous.
https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/AudioManagerImpl.java#L194
We have probably nothing to do for such audio streams.
But it may be strange that doorbird and pulseaudio are not checking these streams.

dalgwen:
I don't understand why you are focusing on URLAudioStream ? They are only a source of audio data, with some code from openHAB to easily handle the data retrieval. Sinks should be agnostic and handle them the same way ? They are not related to the HTTP audio server (or there is something I don't understand).
They are neither synchronous nor asynchronous in nature (except for the getUrl method wich is a shortcut and is clearly meaning that an aware audiosink will play then asynchronously)

By the way, just to be sure, when I use the word "asynchronous", I mean that the sink will return BEFORE consuming/playing all the data from the AudioStream)

My view is : 

In this PR. if the audiosink can play something that can be Disposable (as they shoud be, because this should not be a blocking point unless very bad design) : 
-----  if it is synchronous : **OK**, nothing more to do, handled by the post process code in this PR.
-----  if it is asynchronous  : 
---------- if it uses the HTTP audio server : **OK**, nothing more to do, handled by the HTTP audio server itself in this PR.
---------- if it uses something else : **there is something to do if this pull request is merged.** <---- is there some sinks in this case ?

I propose to : 
- Check all audiosinks included in openhab with this checklist, and make them implements isSynchronous to false and make them handle the Disposable resource on their own. This is not a breaking change (but if some audiosinks outside the openHAB release are in this case and not fixed before the 4.0 version,  they would have issues, because the file can then be deleted before then read it)

OR

- A step further, I can make a breaking change :
---- Remove the default implementation of isSynchronous, to force all audiosink to implement the method. I then will make a PR in openhab addon for all of them. This way, the audiosink developers will all be aware.
---- (optional) Take this opportunity to make a rewrite to the AudioStream interface, for all asynchronous sinks to be able to make a callback when they are done playing (in order to fix the very old issue with the volume reseting to early)

lolodomo:
I started to check few audio sinks and we have finally different implementations, some are synchronous, some are asynchronous.

> I don't understand why you are focusing on URLAudioStream ? 

Just because most of audio sinks use that to know if they have to use or not the HTTP audio server.
Some bindings (not all) even have a different behaviour in that case (Sonos for example).

> ---------- if it uses something else : there is something to do if this pull request is merged. They should declare themsselves as asynchronous and call the dispose method when finishing <---- is there some sinks in this case ?

Yes, most of them in the case of URLAudioStream because the HTTP audio server is not used in that case. But there is nothing to dispose in that case, we are just using an external URL.

lolodomo:
By the way, URLAudioStream will not implement Disposable, so I agree we don't have to care about it.

The problem is more that an audio sink is not always either fully synchronous or fully asynchronous. It can be both depending on the audio stream. For example, it should be asynchronous for an URLAudioStream (the squeezebox binding implementation is probably wrong in that case) as the stream can have an inifnite duration (a WEB radio for example).

lolodomo:
> * Check all audiosinks included in openhab with this checklist, and make them implements isSynchronous to false and make them handle the Disposable resource on their own. This is not a breaking change (but if some audiosinks outside the openHAB release are in this case and not fixed before the 4.0 version,  they would have issues, because the file can then be deleted before then read it)

Yes, the problem is for the audio sinks which are asynchronous. This may be a majority.

lolodomo:
If I am not wrong, it would be fully unchanged compared to today if by default isSynchronous was returning false ?
In that case, this would be a better choice and we will just update audio sinks implementing a synchronous handling of audio stream to take advantage of automatic file deletion.
WDYT ?

lolodomo:
@dalgwen : any feedback?

dalgwen:
@lolodomo  Many thanks your for your feedback.
I took some time to fully take it into acount it and to come back with something that I hope will make a good use of it.

You are right, some sinks are neither fully synchronous, neither asynchronous, so I think it is not a good idea to force a isSynchonize method.
So I switched for another solution : giving the sink the possibility to handle the post process as it sees fit, with the help of a Runnable argument in a new process method.
I try to design it with the goal of keeping exactly the same behaviour as before, with no breaking changes, but with the possibility to handle it properly with a minimal effort.
It also tackles the issue mentionned before (volume restoration).
You can read details in the updated open post.

lolodomo (changes requested):

lolodomo:
I have the feeling it is good.
Do you agree that it will require then to update all existing audio sinks to extend either AudioSinkAsync or AudioSinkSync to benefit all of the new features ?

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioSink.java
```diff
@@ -58,13 +58,45 @@ public interface AudioSink {
...
+     * @throws UnsupportedAudioStreamException If audioStream is not supported
+     */
+    default void process(@Nullable AudioStream audioStream, Runnable whenFinished)
+            throws UnsupportedAudioFormatException, UnsupportedAudioStreamException {
+        try {
+            process(audioStream);
+        } finally {
+            whenFinished.run();
+        }
+    }
```
lolodomo, line 98: You assume a synchronous behaviour by default here. But as you are not handling Closable (deleting source) here, that should be OK, meaning no change compared to the current behaviour.

lolodomo:
So, except my comment about limit checking when copying creating a temportry file, I think the code is fine and I understand what should be the next steps for audio sinks and TTS services.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/FileAudioStream.java
```diff
@@ -125,4 +133,11 @@ public synchronized void reset() throws IOException {
     public InputStream getClonedStream() throws AudioException {
         return getInputStream(file);
     }
+
+    @Override
+    public void dispose() throws IOException {
```
lolodomo, line 138: Note that it will require a change the pico and mac TTS services to build a FileAudioStream rather than a FixedLengthAudioStream to have the temporary file deleted.

lolodomo:
I can't wait for this PR to be merged so that we can start enhance the audio sinks (audio restoration) and the TTS services (using the TTS cache).
Your review will be welcome @J-N-K

lolodomo:
@dalgwen : your PR does not built apparently and you have conflicts to resolve.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/FileAudioStream.java
lolodomo: Or to implement Closable in the existing FixedLengthAudioStream.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/AudioManagerImpl.java
```diff
@@ -140,20 +140,23 @@ public void play(@Nullable AudioStream audioStream, @Nullable String sinkId, @Nu
...
-            } finally {
-                if (volume != null && oldVolume != null) {
+
+            final PercentType oldVolumeFinal = oldVolume;
+            Runnable toRunWhenProcessFinished = () -> {
+                if (volume != null && oldVolumeFinal != null) {
                     // restore volume only if it was set before
                     try {
-                        sink.setVolume(oldVolume);
+                        sink.setVolume(oldVolumeFinal);
```
lolodomo, line 149: Some audio sinks will throw `UnsupportedOperationException`, for example the freebox audio sink.
Maybe this exception should be caught and ignored ?

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -291,19 +291,22 @@ public void say(String text, @Nullable String voiceId, @Nullable String sinkId,
...
-            } finally {
-                if (volume != null && oldVolume != null) {
+
+            final PercentType oldVolumeFinal = oldVolume;
+            Runnable toRunWhenProcessFinished = () -> {
+                if (volume != null && oldVolumeFinal != null) {
                     // restore volume only if it was set before
                     try {
-                        sink.setVolume(oldVolume);
+                        sink.setVolume(oldVolumeFinal);
```
lolodomo, line 300: Some audio sinks will throw `UnsupportedOperationException`, for example the freebox audio sink.
Maybe this exception should be caught and ignored ?

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/AudioServlet.java
```diff
@@ -159,58 +164,156 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
-        streamTimeouts.put(streamId, System.nanoTime() + TimeUnit.SECONDS.toNanos(seconds));
+        ClonableAudioStream clonableAudioStream = null;
+        if (stream instanceof ClonableAudioStream clonableAudioStreamDetected) {
+            clonableAudioStream = clonableAudioStreamDetected;
+        } else { // we prefer a ClonableAudioStream, but we can try to make one
+            try {
+                clonableAudioStream = createClonableInputStream(stream, streamId);
+            } catch (AudioException | IOException e) {
+                logger.warn("Cannot precache the audio stream to serve it", e);
+                return getRelativeURL(streamId);
```
lolodomo, line 264: You are storing neither a streamToServe nor a timeout in that case. Isn't it a problem ?

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/AudioServlet.java
```diff
@@ -159,58 +164,156 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
+            File tempFile = File.createTempFile(streamId, ".snd");
+            tempFile.deleteOnExit();
+            try (OutputStream outputStream = new FileOutputStream(tempFile)) {
+                // copy already read data to file :
+                outputStream.write(dataBytes);
+                // copy the remaining stream data to a file.
+                byte[] buf = new byte[8192];
+                int length;
+                // but with a limit
+                int fileSize = 0;
```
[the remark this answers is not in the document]
dalgwen: Indeed !

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioSink.java
dalgwen: Yes, exactly, no change.

> Asynchronous audio sinks will have to move to AudioSinkAsync to have the volume restored after the stream is fully played.

Yes, sinks that choose to extends AudioSinkAsync are then responsible to execute the Runnable.  To get it, they have to either overriding the process with Runnable arg, or use the runDelayed method that stores it for them.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/FileAudioStream.java
dalgwen: > Or simply to implement Closable in the existing FixedLengthAudioStream.

Yes, this one seems easier.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/AudioManagerImpl.java
dalgwen: Seems legit. Done.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
dalgwen: Seems legit. Done.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/AudioServlet.java
dalgwen: The `serve` method doesn't have any Exception in its signature that I can use, and I tried to avoid any breaking changes.
I'm not satisfied with that either. The issue is then delayed to when the device will try to get the sound. It will get a 404, and it's far from ideal, indeed.
Should I launch a RuntimeException ? But it seems a little bit overkill.

I've added a IOException to the new serve method, at least this one will be OK.
But not to the legacy one... WDYT ? Should we make a Breaking Change ?
Or do you have another idea ?

dalgwen:
> Do you agree that it will require then to update all existing audio sinks to extend either AudioSinkAsync or AudioSinkSync to benefit all of the new features ?

Yes, or if it is not suitable for them, they can override the process method with the Runnable argument.

> And all calls by the audio sinks to the HTTP audio server (serve method) should then be updated to add the runnable parameter like you did in WebAudioAudioSink when the sink is asynchronous to benefit from the Closable feature and the delayed restoration of volume ?

It is not mandatory, if the sink knows when the sound is fully played (example, with the possible monitoring of a playing state ?)
If the sink can't know by itself when the sound is finished, then yes, they should let the AudioServlet try to guess.
I must again add that the "guess" from the AudioServlet is not good : it checks if the AudioStream is fully consumed and the timeout is in the past, and the next pass of the cleaning method will remove it. It is better than before, but it is not right.
We should maybe add some method to guess the length of an audio stream, but it is trickier than seems (example in PulseAudioSink, ConvertedInputStream, getDuration)

> For the Sonos audio sink for example, as it is synchronous for audio served by the HTTP audio server, I think the change should be as simple as what you did for JavasoundAudioSink.

If it returns from process when the sound is played, then yes.

dalgwen:
I made some little changes in the last commit
- some null handling
- some checks in the volume changing process to avoid setting / resetting the volume if it is the same as before.
- mutualize the volume changing method in an utility method for VoiceManager and AudioManager

dalgwen:
(I see that you are eager to tackle the other AudioSinks.
If this is merged and you want to make according changes for other sinks, I used for my test the Chromecast sink and [the code is here](https://github.com/dalgwen/openhab-addons/tree/chromecast_audiosinkasync). (It was more tedious than anticipated because the AudioSink registered class was already extending something else and I had to sever it to use the abstract AudioSinkAsync)
I can make a PR or let you take it in a more global PR, as you want)

lolodomo:
> I can make a PR or let you take it in a more global PR, as you want)

I propose to do it binding by binding and to share the work. So if you already enhanced few audio sinks, you are welcome to push a PR once this PR is merged. I will review it and so this will be merged relatively fast.

dalgwen:
I added some utility method to compute sound duration.
OK, now it's done, I won't touch it again, I promise  😅

dalgwen:
The build failed, but again, I don't know why : I see nothing in the log.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/utils/AudioSinkUtils.java
```diff
@@ -0,0 +1,83 @@
...
+
+        if (volume == null) {
+            return null;
+        }
+
+        // set notification sound volume
+        try {
+            // get current volume
+            oldVolume = sink.getVolume();
+        } catch (IOException e) {
```
lolodomo, line 52: You should consider UnsupportedOperationException  for sinks that are not able to manage volume.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/utils/AudioSinkUtils.java
```diff
@@ -0,0 +1,83 @@
...
+            logger.debug("An exception occurred while getting the volume of sink '{}' : {}", sink.getId(),
+                    e.getMessage(), e);
+        }
+
+        if (!volume.equals(oldVolume) || oldVolume == null) {
+            try {
+                sink.setVolume(volume);
+                volumeChanged = true;
+            } catch (IOException e) {
```
lolodomo, line 61: You should consider UnsupportedOperationException  for sinks that are not able to manage volume.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/AudioServlet.java
```diff
@@ -159,58 +163,171 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
+                int fileSize = ONETIME_STREAM_BUFFER_MAX_SIZE + 1;
+                while ((length = stream.read(buf)) != -1 && fileSize < ONETIME_STREAM_FILE_MAX_SIZE) {
+                    outputStream.write(buf, 0, length);
+                    fileSize += length;
+                }
+            }
+            try {
+                clonableAudioStreamResult = new FileAudioStream(tempFile, stream.getFormat(), true);
+            } catch (AudioException e) { // this is in fact a FileNotFoundException and should not happen
+                throw new IOException("Cannot found the cache file we just created.", e);
```
lolodomo, line 315: Cannot find

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/AudioServlet.java
```diff
@@ -159,58 +163,171 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
+                outputStream.write(dataBytes);
+                // copy the remaining stream data to a file.
+                byte[] buf = new byte[8192];
+                int length;
+                // but with a limit
+                int fileSize = ONETIME_STREAM_BUFFER_MAX_SIZE + 1;
+                while ((length = stream.read(buf)) != -1 && fileSize < ONETIME_STREAM_FILE_MAX_SIZE) {
+                    outputStream.write(buf, 0, length);
+                    fileSize += length;
+                }
```
lolodomo, line 310: With this code, the file can finally have a size bigger than ONETIME_STREAM_FILE_MAX_SIZE

lolodomo (changes requested):

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/AudioServlet.java
dalgwen: Fixed

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/AudioServlet.java
dalgwen: Done

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/utils/AudioSinkUtils.java
dalgwen: Done

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/utils/AudioSinkUtils.java
dalgwen: Done

lolodomo:
Code looks good to me, than you @dalgwen .
I hope this will be review & merge very soon, I have a lot of free time the coming week and I would like to help with the next steps (audio sinks and TTS services).

lolodomo (approved):
LGTM

lolodomo:
Maybe @kaikreuzer or @J-N-K could have a look ?

lolodomo:
@dalgwen : I am preparing/thinking about the next steps. The most common pattern in audio sinks (process method) is this one (taken from Onkyo binding):
```
        String url = null;
        if (audioStream instanceof URLAudioStream) {
            // it is an external URL, the speaker can access it itself and play it.
            URLAudioStream urlAudioStream = (URLAudioStream) audioStream;
            url = urlAudioStream.getURL();
        } else {
            if (callbackUrl != null) {
                String relativeUrl;
                if (audioStream instanceof FixedLengthAudioStream) {
                    // we serve it on our own HTTP server
                    relativeUrl = audioHTTPServer.serve((FixedLengthAudioStream) audioStream, 20);
                } else {
                    relativeUrl = audioHTTPServer.serve(audioStream);
                }
                url = callbackUrl + relativeUrl;
            } else {
                logger.warn("We do not have any callback url, so onkyo cannot play the audio stream!");
                return;
            }
        }
```
Should we now simply remove the test on `instanceof FixedLengthAudioStream` and just call serve with the 20 timeout parameter?

What is exactly the use case where the audio stream would be neither a `URLAudioStream` nor `FixedLengthAudioStream`? For a notification from a TTS service for example , we will always have `FixedLengthAudioStream`. In which case can it happen ?

In this use case, `serve` is called without a timeout parameter, don't you change something by adding an implicit timeout to 10 seconds ?

dalgwen:
In fact, if we want simplicity, the whole block you quote can be simplified in :

`
String url = null;
String relativeUrl = audioHTTPServer.serve(audioStream, 20);`
if (relativeUrl != null) {
    url = callbackUrl +relativeUrl;
} else {
        logger.warn("We do not have any callback url, so onkyo cannot play the audio stream!");
        return;
}
`

This is because there should be no need to test for an URLAudioStream instance and acts differently : as it now implements the Cloneable method, without the need to be FixedLength, it can be served by the AudioServlet in all cases. OpenHAB would then acts as a proxy, reading the stream and sending it without modification. Which is not optimal (unneeded processing) but can be useful in some situation. (1- no need for the physical device to access internet, 2- openHAB can know when the playing is ended). It is the sink choice.
Note that I don't remember if I already tested this situation (an URLAudioStream served by openHAB as a proxy), but it should work theorically.

> Should we now simply remove the test on instanceof FixedLengthAudioStream and just call serve with the 20 timeout parameter?

(First, I think that we should not test against FixedLengthAudioStream, but against the ClonableAudioStream instead)
Indeed, the paradigm shifts here a little.
Before  this PR, we tested for the instance type because it would determine if we could serve it multiple times or juste one time.
Now, it is not needed anymore because even with a bare InputStream we can serve it multiple times. The only remaining use of the serve method with no timeout argument is to state that the stream will be served only once, and thus _don't need to be_ wrapped in a Clonable implementation (if not already Clonable). This is the only case we should want to use it.
But the serve with timeout will now work in every case, indeed. The sink is the only one able to choose if calling to serve once is a good idea.

> What is exactly the use case where the audio stream would be neither a URLAudioStream nor FixedLengthAudioStream? For a notification from a TTS service for example , we will always have FixedLengthAudioStream. In which case can it happen ?

We can always imagine some weird audio source generating sound in real time. Reencoding something on the fly is another use case.
Note that I plan to make MimicTTS returning a bare AudioInputStream, direct from the mimic server. It is an advantage of this new process : simplifying TTS, which should not have to take this kind of constraint and twist their implementation to store/transform/accomodate for AudioSink. We could imagine the same for the other TTS getting their audio data from the web : no need for to store it in a file. GoogleTTS and other like it could be greatly simplified (by removing their cache implementation)

> In this use case, serve is called without a timeout parameter, don't you change something by adding an implicit timeout to 10 seconds ?

Yes, indeed, there is a little change with this implicit timeout. I choose to add it because we can imagine this situation : 
- a serve method with no timeout, counting on the doGet method to clean the stream after use
- the physical device doesn't work and never get the stream. No doGet called.
- an so ... no volume restoration !
Now that the audiosink is responsible for executing some delayed code, I think that this a timeout to ensure execution is necessary.
Note that in line 195 ->203 I add some code to execute as soon as possible the delayed code in case of a one time stream.

lolodomo:
> This is because there should be no need to test for an URLAudioStream instance and acts differently : as it now implements the Cloneable method, without the need to be FixedLength, it can be served by the AudioServlet in all cases. OpenHAB would then acts as a proxy, reading the stream and sending it without modification.

As you introduced a timeout of 10 seconds, it would not work anymore for example for the use case the user requests the streaming of a WEB radio.
So I believe it is better to keep the URLAudioStream case apart in audio sinks, without proxying it.

> I just had a thought to simplify methods signature : we could put the Runnable inside AudioStream, as a https://github.com/nullable field.
> Is it a good idea ?

Maybe but I have difficulties to understand the implications.
This would simplify audio sinks and move a part of the logic to the service that builds the audio stream ?
I let you decide.

I just know that we have to stabilize the proposal if we expect the PR to be merged one day ;)

dalgwen:
> As you introduced a timeout of 10 seconds, it would not work anymore for example for the use case the user requests the streaming of a WEB radio.

There are two conditions for a stream to be removed : timeout **AND** not transfering anymore in the doGet method (this is why I count the currentlyServedStream). So it should be OK.
In fact,  I just test a radio mp3 stream with a modified chromecast (this chromecast version treats an URLAudioStream and any other AudioStream in the same way) and it is OK (tested for 3 minutes).

> I just know that we have to stabilize the proposal if we expect the PR to be merged one day ;)

Yes, this is why I asked you before trying it 😅
But in fact my proposal of including the Runnable inside the AudioStream is not so simple, so no need to bother.

lolodomo:
> In fact, I just test a radio mp3 stream with a modified chromecast (this chromecast version treats an URLAudioStream and any other AudioStream in the same way) and it is OK (tested for 3 minutes).

Fine. 
So we just have to decide if this is a good idea or not to proxy any audio by the OH server.

dalgwen:
> So we just have to decide if this is a good idea or not to proxy any audio by the OH server.

Should we do it here ? As you said, this PR is already a lot 😅
We don't have to make this choice here for the other sinks: it is a per sink choice.
For example, for the WebAudioSink, since we can't manage volume, and since a browser have internet access, I didn't even really consider switching to a proxyfied version and so didn't change the code.

Anyway, for the record, regarding proxification:
Pro: 1 - OH AudioServlet can manage volume restoration at the right time ; the AudioSink don't have to restore it by itself 2 - No need for an internet connection
Cons: 1 - Unnecessary burden for OH server. 2 - Just a guess, may have compatibility issue (http header, something else ?)

lolodomo:
> Should we do it here ? As you said, this PR is already a lot 😅
> We don't have to make this choice here for the other sinks: it is a per sink choice.

Not here, I agree.
Waiting for the merge of the PR, we are just discussing here the next steps and one of them will be the update/simplification of existing audio sinks.

dalgwen:
(Note : I force pushed to do a rebase/resolve conflict)

lolodomo:
@openhab/core-maintainers : no chance to have a review of this PR?

J-N-K (changes requested):

lolodomo:
@dalgwen : did you see @J-N-K comments?

lolodomo:
@dalgwen : no feedback from you?

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioSinkAsync.java
```diff
@@ -0,0 +1,86 @@
...
+public abstract class AudioSinkAsync implements AudioSink {
+
+    private final Logger logger = LoggerFactory.getLogger(AudioSinkAsync.class);
+
+    private final Map<AudioStream, Runnable> runnableByAudioStream = new HashMap<>();
+
+    @Override
+    public void process(@Nullable AudioStream audioStream, @Nullable Runnable whenFinished)
+            throws UnsupportedAudioFormatException, UnsupportedAudioStreamException {
+
```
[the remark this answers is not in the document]
dalgwen: Thanks for you review and sorry for the delay.
Done

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioSink.java
```diff
@@ -58,13 +58,47 @@ public interface AudioSink {
...
+     *
+     * When the process method ends, if the stream implements the {@link org.openhab.core.common.Disposable} interface,
+     * the sink should hereafter get rid of it by calling the dispose method.
+     *
+     * @param audioStream the audio stream to play or null to keep quiet
+     * @param whenFinished A Runnable to run when the sound is finished playing.
+     * @throws UnsupportedAudioFormatException If audioStream format is not supported
+     * @throws UnsupportedAudioStreamException If audioStream is not supported
+     */
+    default void process(@Nullable AudioStream audioStream, @Nullable Runnable whenFinished)
```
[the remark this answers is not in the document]
dalgwen: OK, I rewrote it using CompletableFuture.
Summary of changes around this :
- In one place I had to return a record (ServedStream) instead of just a String in order to return two informations : the CompletableFuture and the stream url, instead of only the stream URL. ServedStream was previously private and is now public.
- Returning a CompletableFuture instead of passing a Runnable forced me to change the method name to avoid collision.
And changing the method signature/name made me think that the old process method should not be used and returning a CompletableFuture should be encouraged by deprecating the old one. Is it OK ?

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/utils/AudioSinkUtils.java
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * Handle a volume command change and returns a Runnable to restore it.
+     *
+     * @param volume The volume to set
+     * @param sink The sink to set the volume to
+     * @param logger to log error to
+     * @return A runnable to restore the volume to its previous value, or null if no change is required
+     */
+    public static @Nullable Runnable handleVolumeCommand(@Nullable PercentType volume, AudioSink sink, Logger logger) {
```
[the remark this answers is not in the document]
dalgwen: > Why do you pass a logger here?

openHAB has a SAT rule of non static logger. As I didn't want to change this simple class with static method to a full service/component (I wanted to keep it static), I thought of passing logger instance as an argument. But I now see that static logger are used elsewhere and it was not necessary.
Anyway, I moved handleVolumeCommand in AudioManagerImpl, so it has access to the class logger.

> And wouldn't this whole class unnecessary if .setVolume would return the old volume? Restoring the old volume would then be a one-liner.

The whole point of this method, unlike the previous way of doing, is to allow us to do no nothing if not needed (and mutualize code and exception handling). In the previous implementation, volume was always set back to the old value, even if it's the same value. And for some sinks, it is an operation that can take some non negligible time and/or unneeded side effect)
That said, I could have done otherwise, because a Runnable is not the only way to state a no-op operation. A Nullable or Optional old volume value would have worked. But then it wouldn't be a one liner, as it would be mandatory to test for null. As this method is used only in a couple of place in the same way (with a delayed task in mind), a Runnable seems to me a good choice to make it a one-liner. Is also has the added value of managing exception inside. I added some comment in the method to explain this. But it is not very important, so if you want me to change it, no problem, just tell me to.

Another point to mention:
I had to move the transferAndAnalyzeLength method to a service component, to avoid a dependancy issue with the javazoom library used for parsing mp3.

dalgwen:
> @dalgwen : no feedback from you?

Sorry for the delay. I rebased and force pushed another commit with the changes suggested by J-N-K.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioSinkAsync.java
```diff
@@ -0,0 +1,114 @@
...
+            throws UnsupportedAudioFormatException, UnsupportedAudioStreamException {
+        CompletableFuture<@Nullable Void> completableFuture = new CompletableFuture<@Nullable Void>();
+        try {
+            if (audioStream != null) {
+                runnableByAudioStream.put(audioStream, completableFuture);
+            }
+            processAsynchronously(audioStream);
+            return completableFuture;
+        } finally {
+            if (audioStream == null) {
```
J-N-K, line 54: If this is true, no `CompletableFuture` was added, so it doesn't need to be removed. In fact, it would be better to do the null check above and return a `CompletableFuture.completedFuture(null)`. You can skip the try/catch block then.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/AudioManagerImpl.java
```diff
@@ -122,38 +122,13 @@ public void play(@Nullable AudioStream audioStream, @Nullable String sinkId) {
...
-                    // restore volume only if it was set before
-                    try {
-                        sink.setVolume(oldVolume);
-                    } catch (IOException e) {
-                        logger.debug("An exception occurred while setting the volume of sink '{}' : {}", sink.getId(),
-                                e.getMessage(), e);
-                    }
-                }
             }
         } else {
```
J-N-K, line 133: You can collapse all that by using `.handle((t, v) -> {...})` instead of `.thenRun()`. If an exception occured, `t != null` and contains the exception (can be used to log the warning). There is no need for the `Runnable` variable then.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -272,44 +272,17 @@ public void say(String text, @Nullable String voiceId, @Nullable String sinkId,
...
-                    // restore volume only if it was set before
-                    try {
-                        sink.setVolume(oldVolume);
-                    } catch (IOException e) {
-                        logger.debug("An exception occurred while setting the volume of sink '{}' : {}", sink.getId(),
-                                e.getMessage(), e);
-                    }
-                }
-            }
+            volumeRestauration = audioManager.handleVolumeCommand(volume, sink);
```
J-N-K, line 275: `restoreVolume` instead of `volumeRestauration`

J-N-K (changes requested):
Thanks, I have left some last comments.

@lolodomo Can you also have a final look?

lolodomo (approved):
I have done a new full review.
It looks good to me (even if all this is not trivial).

lolodomo:
@dalgwen : do you have time to answer to @J-N-K 's last comments? 
We approach OH4 release and we must conclude quickly regarding this PR and following coming changes in addons and TTS services.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioSinkAsync.java
dalgwen: Indeed, the remove in the null case is an error.

But, correct me if I'm wrong, as we have to call the processAsynchronously method -even if the stream is null, in order to ask the sink to stop current playing-, then if I remove the try / finally, I cannot be sure that the delayed volume restauration will occur (if there is an exception during the processAsynchronously(null) method).

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/AudioManagerImpl.java
dalgwen: Your proposal highlight the fact that exceptions were not handled in the  'CompletableFuture' way.
So, I rewrited it to do so. Please note that I use the method `exceptionally` instead of `handle` because Void return value makes it more adequate.

But that said, I still need the Runnable variable, because the  volume value to set back has to be computed before the processAndComplete method. Or is there something I didn't understand ?

lolodomo:
@dalgwen : I do not see any new commit since  @J-N-K last review, did you forget to push your last commit ?

dalgwen:
> @dalgwen : I do not see any new commit since @J-N-K last review, did you forget to push your last commit ?

Woops, indeed. Sorry ! It's done.

lolodomo:
Cool, now we can enhance TTS services and audio sinks.

Can you please trigger a new build ?

PR #3464, opened 2023-03-18
Fix variable binding in ScriptTransformationService

jimtng:
...by setting all attributes before compiling.

In JRuby, because an identifier can either be a variable or a method call, the compiler needs to know variables ahead of time  in order to recognise them as variables. Because scripts are compiled first, and then the attributes set, this information is missing, therefore [a "hack" was needed in jruby transformations](https://github.com/openhab/openhab-docs/blob/ad1e66c440e7a6150b11aeaef31a132c389d552b/addons/transformations.md?plain=1#L113) to give the compiler the needed hint.

This PR simply reverses the order so that attributes were set before compiling, so the hack in jruby will no longer be needed. I have tested this on my jrubyscripting addon installation.

For more info, see https://github.com/jruby/jruby/issues/5876


J-N-K (approved):
LGTM, thanks!

jimtng:
Thanks for merging! Could this please be applied to 3.4.x too? `patch` tag I think?

Issue #3321, opened 2023-01-16
Upgrade Gson to 2.10


jlaur:
Now that we have Java 17 support, it would be nice to upgrade Gson to the latest version (currently 2.10.1). Version 2.10 supports Java records: https://github.com/google/gson/releases/tag/gson-parent-2.10. Records are particularly useful when used in combination with Gson for serializing/deserializing from/to DTO's.

I'm working on a new binding (openhab/openhab-addons#14222) where I have initially taken advantage of this:

```java
@NonNullByDefault
public record ElspotpriceRecord(@SerializedName("HourUTC") Instant hour,
        @SerializedName("SpotPriceDKK") double spotPrice) {
}
```

However, including this Gson dependency within the binding adds more than 250 kB, which is probably not acceptable:
```xml
  <dependencies>
    <dependency>
      <groupId>com.google.code.gson</groupId>
      <artifactId>gson</artifactId>
      <version>2.10.1</version>
    </dependency>
  </dependencies>
```

I tried upgrading from 2.9.1 to 2.10.1 in core, but ended up with a dependency problem:

`[ERROR] Failed to execute goal on project org.openhab.core.bom.compile-model: Could not resolve dependencies for project org.openhab.core.bom:org.openhab.core.bom.compile-model:pom:4.0.0-SNAPSHOT: Failed to collect dependencies at org.eclipse.xtext:org.eclipse.xtext.xbase.ide:jar:2.29.0 -> org.eclipse.xtext:org.eclipse.xtext.ide:jar:2.29.0 -> org.eclipse.lsp4j:org.eclipse.lsp4j:jar:0.19.0 -> org.eclipse.lsp4j:org.eclipse.lsp4j.generator:jar:0.19.0 -> org.eclipse.lsp4j:org.eclipse.lsp4j.jsonrpc:jar:0.19.0 -> com.google.code.gson:gson:jar:[2.9.1,2.10): No versions available for com.google.code.gson:gson:jar:[2.9.1,2.10) within specified range -> [Help 1]`

I don't know how to fix this, but I'm wondering if it would be possible and acceptable to have both 2.9.1 (because of this dependency) and 2.10.1 in core? At least 2.10.1 would only be included once and all bindings could take advantage of it.

J-N-K:
Xtext 2.30 is expected at end of February and that updates to GSON 2.10.

wborn:
[...]

PR #3467, opened 2023-03-20
Adding "category" property to things

clinique:
Following the suggestion I made [here](https://community.openhab.org/t/openhab-4-0-wishlist/142388/403?u=glhopital).

Thing-Type and Thing XML schemas where already holding a category field.
This adds the ability to add it in `.things` files, with the same syntax than `.items` file (eg:

`magic:color-light:magicColor @"theplace" <lightbulb>
`

I'm missing the ability to inherit thing category from thing-type category (did not find where it happens).

Using the API explorer, I get visibility on category as a field of the thing. If this gets approved (and probably enhanced), it could then be added in web-ui.

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/thing/EnrichedThingDTO.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/thing/EnrichedThingDTO.java
@@ -46,7 +46,7 @@ public class EnrichedThingDTO extends AbstractThingDTO {
     EnrichedThingDTO(ThingDTO thingDTO, List<EnrichedChannelDTO> channels, ThingStatusInfo statusInfo,
             FirmwareStatusDTO firmwareStatus, boolean editable) {
         super(thingDTO.thingTypeUID, thingDTO.UID, thingDTO.label, thingDTO.bridgeUID, thingDTO.configuration,
-                thingDTO.properties, thingDTO.location);
+                thingDTO.properties, thingDTO.location, thingDTO.category);
         this.channels = channels;
         this.statusInfo = statusInfo;
         this.firmwareStatus = firmwareStatus;
--- a/bundles/org.openhab.core.model.thing/src/org/openhab/core/model/thing/Thing.xtext
+++ b/bundles/org.openhab.core.model.thing/src/org/openhab/core/model/thing/Thing.xtext
@@ -22,6 +22,7 @@ ModelBridge returns ModelThing:
 	bridge?='Bridge' (id=UID | thingTypeId=UID_SEGMENT thingId=UID_SEGMENT)
     (label=STRING)?
     ('@' location=STRING)?
+    ('<' icon=(ID|STRING) '>')?    
 	('['
 		properties+=ModelProperty? (',' properties+=ModelProperty)*
 	']')?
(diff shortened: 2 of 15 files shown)


lolodomo:
We are currently refactoring/extending what is a category for an item, so I would prefer to wait for the change for items.
I already started.

lolodomo:
> Where is your work around items ?

In fact, my plans were to do it in two steps: first step was to update the icon attribute in sitemap (PR #3378). Second step was to apply a similar change for item.
I am still fighting with the way to update the sitemap syntax. I will try again during the weekend.

lolodomo:
The thing icon will then be used only when setting up the system, e.g. in the page showing all the things ?
I am just not sure it is worth the effort ?

clinique:
> The thing icon will then be used only when setting up the system, e.g. in the page showing all the things ? I am just not sure it is worth the effort ?

As said in the comment in the wishlist, it's a start to make the UI more fancy. Bindings also has icons that are only displayed in documentation. The provided icon could be used for things representing the group. I also would like to think on how could binding provide additional icon sets to streamline the user experience when creating items.

lolodomo:
> I also would like to think on how could binding provide additional icon sets to streamline the user experience when creating items.

I believe it should not be too much difficult. The binding should implement an `IconProvider` with the binding id as icon set id and that icon provider would then search icons in binding folder "resources/OH-INF/icon". Then, the icon could be acceded by using "oh:bindingId:icon", for example "oh:netatmo:welcome_camera" assuming you provided the icon "welcome_camera.svg" or "welcome_camera.png".
The only difficulty I can guess is how to define the icon provider priority.
You can look at class ClassicIconProvider.

clinique:
Yes, I know about the IconProvider but I'm not sure it's the correct approach if we want to promote it accross all bindings. That would mean one in each binding ? It then should be good to add it in the binding skeleton.

lolodomo:
By the way, your idea to have a binding providing a dedicated icons is a really good idea. I remember we adviced to download few netatmo icons and create custom iconswith that, for the wifi status for example. This could be simplified if the binding was providing these icons.
No change in core framework is required, this can already be implemented in any binding.
These icons will then be usable by the user and will render properly in Basic UI and in the Android app, they will also in MainUI after a bug is fixed.

lolodomo:
> Yes, I know about the IconProvider but I'm not sure it's the correct approach if we want to promote it accross all bindings. That would mean one in each binding ? It then should be good to add it in the binding skeleton.

We have a discovery service in most of bindings, so why not an icon service. 
I just imagine it will be less commonly used by developers than discovery services. So maybe not necessarily needed to include that in the binding skeleton.

clinique:
Yes, I'm doing a POC with the MeteoAlerte binding.

J-N-K:
If it turns out that a lot of code is shared for every `IconProvider` we can also add that to an `AbstractIconProvider` in core so that the code needed in the add-ons is very small (probably only namespace and location of the items, if they are included in the bundle).

lolodomo:
There is already an abstract class on core.

clinique:
Pushed PR [here](https://github.com/openhab/openhab-addons/pull/14811)

J-N-K:
One problem with `record` is that our current GSON is not able to handle them. We have to find a solution for stream first (see #3321)

J-N-K:
Do we still need these thing categories when we have custom iconsets and icon providers in add-ons?

clinique:
Yes, delivering on purpose iconset providers give sense to this PR. I just want to be active on an idea I proposed in the wishlist.

andrewfg:
> I'm missing the ability to inherit thing category from thing-type category (did not find where it happens).

If you look at the changes I made in #4617 you will see all the classes where your new field needs to be added. There are over 30 classes which need to be changed.

PR #3470, opened 2023-03-21
[de] Fix article of "token"

felixschndr:
The word `token` is male in German. In `auth.createapitoken.success.footer` the word was gender neutral


florian-h05:
AFAIK translations should not be edited outside of CrowdIn because they will be overwritten with those from CrowdIn. You can edit the German translations here: https://crowdin.com/translate/openhab-core/508/en-de?filter=basic&value=0.

felixschndr:
If so I will make the change there

felixschndr:
Changed on crowdin.com

Issue #3473, opened 2023-03-22
Failed to normalize configuration for thing


morph166955:
I'm on snapshot 3360.  I updated a jar from the marketplace and I'm now getting this error when installing the new version.  I haven't made any changes to the channels on that binding in a while so not sure if this is something I did or if it was something new in the core.

2023-03-22 14:47:30.727 [WARN ] [core.thing.internal.ThingManagerImpl] - Failed to normalize configuration for thing 'androidtv:shieldtv:theater': {thing/channel=Type description for {0} not found although we checked the presence before.}
2023-03-22 14:47:30.729 [WARN ] [core.thing.internal.ThingManagerImpl] - Failed to normalize configuration for thing 'androidtv:shieldtv:livingroom': {thing/channel=Type description for {0} not found although we checked the presence before.}

J-N-K:
The reason is that a description is missing and the thing was initialized via the 120s timeout introduced in #3397 as temporary solution to #3394. The permanent solution is #3407 and related PR in openhab-addons.

J-N-K:
Is the code available somewhere? It's most likely a binding issue.

morph166955:
https://github.com/openhab/openhab-addons/pull/14282 is the pending PR

J-N-K:
Does it only happen when you update the bundle or does it also happen when you restart openHAB after you updated the bundle? In that case it's probably a concurrency issue when updating the bundle.

morph166955:
I noticed it when I removed and then installed it from the marketplace earlier.  It was doing it every time.  I haven't been able to replicate it since updating from snapshot 3360 to 3382.

J-N-K:
So what is the current state? Do you still see it on latest snapshot? Otherwise it's probably fixed. If you still see it: when exactly and how an I reproduce it?

morph166955:
I can't reproduce anymore on the current snapshot. I'll close until this pops up again. Thanks for looking into it!

PR #3474, opened 2023-03-22
Fix ConfigValidationMessage in ThingManagerImpl

J-N-K:
Seen in the message in #3473: The placeholder was not properly replaced.


cweitkamp (approved):
LGTM. Thanks.

Issue #3465, opened 2023-03-18
SCRIPT transformation syntax and file naming


jimtng:
I wish that the SCRIPT transformation's syntax uses the underlying scripting engine's extension, i.e. just `.js`, `.rb`, etc. instead of `.script`. I understand that the reason it's `.script` now is because it's the name of the transformation itself (i.e. `SCRIPT` transformation -> `.script` file extension, `MAP` transformation -> `.map` file extension). As a result of this choice, an extra field must be supplied to indicate the script type (rb, graaljs, dsl, etc).

But it would be so much nicer to have `SCRIPT(myscript.rb)` and `SCRIPT(myscript.js)` rather than `SCRIPT(rb:myscript.script)` and `SCRIPT(graaljs:myscript.script)`:
* It looks nicer. The scriptType is deduced from the file extension just like the scripts in conf/automation
* The syntax is simpler and more straight forward. One less field. If necessary, this first field can be provided to avoid ambiguity e.g. graaljs vs nashorn, if that's still relevant. This way it can be backwards compatible with the current syntax. When the first field is provided, it overrides the scriptType, but if it's absent, deduce the scriptType from the given file extension
* We get the correct syntax highlighting and editor behaviour while editing myscript.rb instead of myscript.script because the editor recognises the file extension
* When looking through the files inside OPENHAB_CONF/transform, when we see myscript.rb and myscript.js, we'd know exactly which language they're written in, instead of myscript.script

J-N-K:
I can't find the discussion immediately but this has been discussed before and we agreed that files in the transformation folder should have an extension that is their transformation type. My vote would be "won't fix".

wborn:
A link to the previous discussion would be appreciated!

> If necessary, this first field can be provided to avoid ambiguity e.g. graaljs vs nashorn, if that's still relevant.

With jsr223 rules files the .graaljs and .nashornjs extensions can already be used for avoiding ambiguity. Usually you can customize your OS and editors to add additional extensions for opening editors and syntax highlighting.

J-N-K:
I‘ll see if I can find it. Removing the `language:` prefix would bring us into trouble with the inline scripts where we don‘t have an extension to derive the language from.

jimtng:
No, don't remove it completely. Make it optional, both for compatibility and for the cases when it's needed. So `SCRIPT(graaljs:myscript.script)` `SCRIPT(graaljs:myscript.js)` and  `SCRIPT(myscript.js)` are all supported.

[...]

Issue #3383, opened 2023-02-18
ConfigDispatcherFileWatcherTest always fails on Windows


wborn:
These tests now always fail on Windows builds so it might be a real bug e.g. caused by #3004.

See this GHA build: https://github.com/wborn/openhab-core/actions/runs/4210674830/jobs/7308736469

### configurationFileCreated

```
[ERROR] Tests run: 6, Failures: 3, Errors: 0, Skipped: 0, Time elapsed: 1.562 s <<< FAILURE! - in org.openhab.core.config.dispatch.internal.ConfigDispatcherFileWatcherTest
[ERROR] org.openhab.core.config.dispatch.internal.ConfigDispatcherFileWatcherTest.configurationFileCreated  Time elapsed: 1.426 s  <<< FAILURE!
Wanted but not invoked:
configDispatcherMock.processConfigFile(
    myPath.cfg
);
-> at org.openhab.core.config.dispatch.internal.ConfigDispatcherFileWatcherTest.configurationFileCreated(ConfigDispatcherFileWatcherTest.java:50)

However, there was exactly 1 interaction with this mock:
configDispatcherMock.processConfigFile(
    conf\services
);
-> at org.openhab.core.config.dispatch.internal.ConfigDispatcherFileWatcher.<init>(ConfigDispatcherFileWatcher.java:58)


	at org.openhab.core.config.dispatch.internal.ConfigDispatcherFileWatcherTest.configurationFileCreated(ConfigDispatcherFileWatcherTest.java:50)
```

### configurationFileRemoved

```
[ERROR] org.openhab.core.config.dispatch.internal.ConfigDispatcherFileWatcherTest.configurationFileRemoved  Time elapsed: 0.004 s  <<< FAILURE!
Wanted but not invoked:
configDispatcherMock.fileRemoved(
    "D:\a\openhab-core\openhab-core\bundles\org.openhab.core.config.dispatch\myPath.cfg"
);
-> at org.openhab.core.config.dispatch.internal.ConfigDispatcherFileWatcherTest.configurationFileRemoved(ConfigDispatcherFileWatcherTest.java:82)

However, there was exactly 1 interaction with this mock:
configDispatcherMock.processConfigFile(
    conf\services
);
-> at org.openhab.core.config.dispatch.internal.ConfigDispatcherFileWatcher.<init>(ConfigDispatcherFileWatcher.java:58)


[...]

PR #3476, opened 2023-03-23
Support a cleaner syntax for SCRIPT transformation

jimtng:
Support `SCRIPT(file.rb)` syntax as an alternative, not a replacement, to `SCRIPT(rb:file.script)`
Also `SCRIPT(file.rb?param1=x)` works like `SCRIPT(rb:file.script?param1=x)`

The original syntax is still fully supported, so this is not a breaking change.

This makes changing from JS transformation to SCRIPT transformation as easy as changing `JS(myscript.js)` to `SCRIPT(myscript.js)`

Discussed in #3465

I understand that @J-N-K disagreed with this proposal. However, I feel very strongly about making the script transformation syntax more elegant and intuitive for file-based scripts and I have created this PR as a PoC to demonstrate what I wish to achieve. 

Unrelated to the matter, some extra error checks were added to the regex pattern to avoid errors, e.g. when an inline transformation script had a `?` character as part of the code, it would've been extracted as  parameters, which shouldn't be the case.

If this doesn't negatively affect anything else, I'd appreciate further considerations for this PR. If there are any potential issues that I'm not aware of, I'll try to address them also.

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptTransformationService.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptTransformationService.java
@@ -68,8 +68,12 @@ public class ScriptTransformationService
     private static final String PROFILE_CONFIG_URI = "profile:transform:SCRIPT";
     public static final String SUPPORTED_CONFIGURATION_TYPE = "script";
 
+    // exclude `?` in scriptType so "script.js?param=:|" doesn't match inline pattern
+    private static final Pattern INLINE_SCRIPT_CONFIG_PATTERN = Pattern
+            .compile("(?<scriptType>[^?]*?):\\|(?<inlineScript>.+)");
+
     private static final Pattern SCRIPT_CONFIG_PATTERN = Pattern
-            .compile("(?<scriptType>.*?):(?<scriptUid>.*?)(\\?(?<params>.*?))?");
+            .compile("((?<scriptType>.+):)?(?<scriptUid>.+?)(\\?(?<params>.*?))?");
 
     private final Logger logger = LoggerFactory.getLogger(ScriptTransformationService.class);
 
@@ -101,30 +105,53 @@ public class ScriptTransformationService
 
     @Override
     public @Nullable String transform(String function, String source) throws TransformationException {
-        Matcher configMatcher = SCRIPT_CONFIG_PATTERN.matcher(function);
-        if (!configMatcher.matches()) {
-            throw new TransformationException("Script Type must be prepended to transformation UID.");
+        String scriptType;
+        String scriptUid;
+        String inlineScript = null;
+        String params = null;
+
+        Matcher configMatcher = INLINE_SCRIPT_CONFIG_PATTERN.matcher(function);
+        if (configMatcher.matches()) {
+            scriptType = configMatcher.group("scriptType");
+            inlineScript = configMatcher.group("inlineScript");
+
+            if (scriptType == null || scriptType.isBlank()) {
+                throw new TransformationException(
+                        "Script type must be specified for the inline transformation '" + inlineScript + "'");
+            }
+
+            scriptUid = "|" + Integer.toString(inlineScript.hashCode()); // prefix with | to avoid clashing with a real
+                                                                         // filename
+        } else {
+            configMatcher = SCRIPT_CONFIG_PATTERN.matcher(function);
+            if (!configMatcher.matches()) {
+                throw new TransformationException("Invalid syntax for the script transformation: '" + function + "'");
+            }
+            scriptUid = configMatcher.group("scriptUid");
+            scriptType = configMatcher.group("scriptType");
+            params = configMatcher.group("params");
         }
-        String scriptType = configMatcher.group("scriptType");
-        String scriptUid = configMatcher.group("scriptUid");
 
         ScriptRecord scriptRecord = scriptCache.computeIfAbsent(scriptUid, k -> new ScriptRecord());
         scriptRecord.lock.lock();
         try {
             if (scriptRecord.script.isBlank()) {
-                if (scriptUid.startsWith("|")) {
-                    // inline script -> strip inline-identifier
-                    scriptRecord.script = scriptUid.substring(1);
+                scriptRecord.scriptType = scriptType;
+                if (inlineScript != null) {
+                    scriptRecord.script = inlineScript;
                 } else {
                     // get script from transformation registry
                     Transformation transformation = transformationRegistry.get(scriptUid);
                     if (transformation != null) {
-                        if (!SUPPORTED_CONFIGURATION_TYPE.equals(transformation.getType())) {
-                            throw new TransformationException("Configuration does not have correct type 'script' but '"
-                                    + transformation.getType() + "'.");
-                        }
                         scriptRecord.script = transformation.getConfiguration().getOrDefault(Transformation.FUNCTION,
                                 "");
+                        if (scriptType == null || scriptType.isBlank()) {
+                            scriptRecord.scriptType = transformation.getType();
+                            if (scriptRecord.scriptType.equals(SUPPORTED_CONFIGURATION_TYPE)) {
+                                throw new TransformationException(
+                                        "Script Type must be prepended to transformation UID.");
+                            }
+                        }
                     }
                 }
                 if (scriptRecord.script.isBlank()) {
@@ -133,18 +160,18 @@ public class ScriptTransformationService
                 scriptCache.put(scriptUid, scriptRecord);
             }
 
-            if (!scriptEngineManager.isSupported(scriptType)) {
+            if (!scriptEngineManager.isSupported(scriptRecord.scriptType)) {
                 // language has been removed, clear container and compiled scripts if found
                 if (scriptRecord.scriptEngineContainer != null) {
                     scriptEngineManager.removeEngine(OPENHAB_TRANSFORMATION_SCRIPT + scriptUid);
                 }
                 clearCache(scriptUid);
-                throw new TransformationException(
-                        "Script type '" + scriptType + "' is not supported by any available script engine.");
+                throw new TransformationException("Script type '" + scriptRecord.scriptType
+                        + "' is not supported by any available script engine.");
             }
 
             if (scriptRecord.scriptEngineContainer == null) {
-                scriptRecord.scriptEngineContainer = scriptEngineManager.createScriptEngine(scriptType,
+                scriptRecord.scriptEngineContainer = scriptEngineManager.createScriptEngine(scriptRecord.scriptType,
                         OPENHAB_TRANSFORMATION_SCRIPT + scriptUid);
             }
             ScriptEngineContainer scriptEngineContainer = scriptRecord.scriptEngineContainer;
@@ -160,7 +187,6 @@ public class ScriptTransformationService
                 ScriptContext executionContext = engine.getContext();
                 executionContext.setAttribute("input", source, ScriptContext.ENGINE_SCOPE);
 
-                String params = configMatcher.group("params");
                 if (params != null) {
                     for (String param : params.split("&")) {
                         String[] splitString = param.split("=");
@@ -279,6 +305,7 @@ public class ScriptTransformationService
         public String script = "";
         public @Nullable ScriptEngineContainer scriptEngineContainer;
         public @Nullable CompiledScript compiledScript;
+        public String scriptType = "";
 
         public final Lock lock = new ReentrantLock();
     }
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptTransformationService.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptTransformationService.java
@@ -13,6 +13,7 @@
 package org.openhab.core.automation.module.script;
 
 import java.net.URI;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Locale;
@@ -84,6 +85,7 @@ public class ScriptTransformationService
 
     private final TransformationRegistry transformationRegistry;
     private final Map<String, String> supportedScriptTypes = new ConcurrentHashMap<>();
+    private final List<String> supportedTransformationTypes = new ArrayList<>(List.of(SUPPORTED_CONFIGURATION_TYPE));
 
     private final ScriptEngineManager scriptEngineManager;
 
@@ -275,7 +277,7 @@ public class ScriptTransformationService
         if (PROFILE_CONFIG_URI.equals(uri.toString())) {
             if (ScriptProfile.CONFIG_TO_HANDLER_SCRIPT.equals(param)
                     || ScriptProfile.CONFIG_TO_ITEM_SCRIPT.equals(param)) {
-                return transformationRegistry.getTransformations(List.of(SUPPORTED_CONFIGURATION_TYPE)).stream()
+                return transformationRegistry.getTransformations(supportedTransformationTypes).stream()
                         .map(c -> new ParameterOption(c.getUID(), c.getLabel())).collect(Collectors.toList());
             }
             if (ScriptProfile.CONFIG_SCRIPT_LANGUAGE.equals(param)) {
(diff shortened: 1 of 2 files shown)


J-N-K:
This breaks the `ConfigOptionProvider` and some scripts will show up when creating profiles (`.script` and managed) and some will not (`.js`, `.rb`, ...). This is far more confusing than prefixing the script language.

jimtng:
> This breaks the `ConfigOptionProvider` and some scripts will show up when creating profiles (`.script` and managed) and some will not (`.js`, `.rb`, ...). This is far more confusing than prefixing the script language.

Thanks for the feedback. I believe I can fix that.

jimtng:
> This breaks the `ConfigOptionProvider` and some scripts will show up when creating profiles (`.script` and managed) and some will not (`.js`, `.rb`, ...)

I've added some extra code to include all the supported scriptTypes loaded in the transformation registry.

If we are going to infer the scriptengine / language from the extension here too, we can remove the support for .script completely, although it will be a breaking change, but it is one that actually improves usability as it simplifies the UI by reducing errors. Previously, the user can select the wrong language especially in the UI when all they see was `myscript.script` they don't have a hint as to the real language inside that file. If we removed the `.script` and only support the actual languages' extensions, this wouldn't happen.

I am merely mentioning this as an observation. It is not the goal / purpose / scope of this PR to make such a drastic change.

image

J-N-K:
I don't understand where you see the "extra burden", selecting the language is a common task when configuring openHAB. If you add a rule with a script action, you have to select the language, too. As I already said: I believe this is wrong and complicates things all the way for the benefit of a few.

jimtng:
> I don't understand where you see the "extra burden", selecting the language is a common task when configuring openHAB.

It seems quite clear to me:
- You decide which language to use when creating the .script or managed transformation
- You are given a number of *.script files and managed transformations as an option
- Then again you have to select which language to execute that file/managed transformation on this screen and each and every time you need to select a transformation.
- User is allowed to make the wrong language choice here. At the very least, if you've created those transformation a long time ago, you'd have to look into the file again to check which language it is before making the choice here. It is a completely unnecessary step. Why make things harder?

versus:
- You decide which language to use, at the creation of the script, whether it is a file, or a managed transformation.
- You are given a number of *.js, *.rb, *.groovy, etc, + managed transformations, all of which already carries their own association to the corresponding language, so the user doesn't have to choose the language again when selecting the transformation

It simplifies the UI and eliminates one possible error. 

To me .script is the complicated way.

However, as I said, I'm not here to advocate for changing the use of .script in this PR. People can continue to use .script exclusively and have no trace of per-language extensions showing in their system should they choose to do so.

> I believe this is wrong and complicates things all the way for the benefit of a few.

Indeed this is a matter of opinion and I'm at the opposite end of that opinion. As I understand it, openHAB is a community project, so is there a democratic process in making a decision or is it a case of the decision is made only by certain individuals? Should we create a poll or something to see whether this only benefits "a few" or the "majority"?

Again, I'm quite happy to leave `.script` as is if that's your preferred way of working with the transformation, as long as .js / .rb etc is also supported.

jimtng:
> If you add a rule with a script action, you have to select the language, too.

Yes, but _only_ at the time of creation of that rule. Not when you are merely selecting from a list of existing rules such as demonstrated in the following screenshots

image

Here you're given a list of existing rules. You do not need to specify in which language the rule should be executed, because indeed that is irrelevant here.
image

J-N-K:
Do what you want, I won't veto that. But I'm pretty sure that this will end in a full nightmare for @rkoshak when support is needed.

ccutrer:
> Do what you want, I won't veto that. But I'm pretty sure that this will end in a full nightmare for @rkoshak when support is needed.

You keep saying that inferring the script type is somehow more complicated than having to specify it, without giving any actual examples of what is more complicated. Up to this point I thought you meant that it would make the implementation and interfaces complicated, but this comment makes me think maybe you have something in mind that _not_ having to specify the script language will somehow be more complicated for the user to use. Do you have a specific example in mind where you think a user might be confused?

J-N-K:
Every time there are several ways to do the same thing this ends up with people mixing instructions from different places. The result is that they start complaining something is "not working".

ccutrer:
Wow, that argument basically reads like "we shouldn't make simpler ways to do things, because sometimes people will get confused." Why was Main UI (and Paper UI before it) even written if you could already administer openHAB completely via files?!

@jimtng even posited that we could entirely get rid of having to choose the scripting language at all (when referencing transformations, not when writing new ones). That would mean there are no additional ways to do the same thing (though there would be a small period of additional support burden if there needs to be manual steps when updating openHAB, but the transformation could still undocumented support the old way), not adding to the support burden.

I still haven't seen an argument I understand for _not_ making the user repetitively specify the scripting language violates "simple design principles".

ccutrer:
> [file based configuration] should be removed completely

Got it. You believe file based configuration should be removed completely, so anything that improves the file-based experience isn't worth doing. I'm not being facetious or sarcastic here. It helps me to understand from the point of view your arguments come.

> What I'm saying is that "map transformation uses .map", "scale transformation uses .scale" and "script transformation uses .script" is much easier than "map transformation uses .map", "scale transformation uses .scale" and "script transformation uses .script or whatever scripting language you want".

I think our argument is that people don't ask "what extension does a script transformation use? obviously it's .script! now I'm going throw some javascript in there", they say "I have a javascript transformation. that's implicitly a .js file. now which transformation do I use for this? oh, script, okay!"

> Except when you configure via UI, then you have to use script.

Why are you thinking about the file extension at all when creating a transformation in the UI??

> If you want to go that route, do it fully: remove .script

Yes, @jimtng suggested that, since it would also simplify the UI way.

J-N-K:
> > [file based configuration] should be removed completely
> 
> Got it. You believe file based configuration should be removed completely, so anything that improves the file-based experience isn't worth doing.

I didn‘t say that and this really marks the end of the discussion for me.

rkoshak:
I've watched this thread and the other one.

From my perspective, the `.script` file extension has always been unsatisfactory for all the reasons mentioned here already and a few more reasons. Yes, using the `.script` extension is consistent with the other transformations but it's inconsistent with rule file extensions which each have their own file extensions even given that in addition they have separate folders to further separate and distinguish them. There's also overlap with [Rules DSL Scripts](https://www.openhab.org/docs/configuration/actions.html#openhab-subsystem-actions). So it's consistent with the other transformations but in OH writ large it poses it has inconsistencies.

Also, when users create a SCRIPT transformation, I'm not sure they are thinking about it as a SCRIPT transform so much as a jRuby transform or a Rules DSL transform. I know that's the mental model I fall into when I think about it. Whether or not that is reasonable I can't say, but it's what I do. @J-N-K, I'm sure you think about it as one thing given you wrote it that way and have the most familiarity with it. The SCRIPT part isn't as important to the end user as the language the transform is written in because it's the language that is going to dictate the syntax. The SCRIPT transformation itself, mentally, is like the engine that the scripts run in.

It also really seems odd and redundant, in the profile case, to have to choose the language. It seems like the language the transform is written in should be somehow inferred from the script itself. Why does the user need to choose the language when creating the script and then have to remember and choose it again when applying it? (Not to mention that screen seems to indicate that you can't have the incoming and outgoing transform script for a profile written in different languages). But even if we kept it where the user selects the language, wouldn't it be nice if upon selecting the language only those scripts written in that language were shown to select from.

So I am sympathetic to the arguments made by @jimtng and @ccutrer. However, I don't really know what the scripts will look like when defined in the UI. I don't understand the implications of what's being proposed and how that is different (from the end user's perspective) so I can't really weigh in on the counter argument without more of an idea how that will look. As far as I can tell, at least on build #3383, we don't have a UI for creating transformation scripts yet. But from what I can gather, the filename is part of the scripts' UID and therefore it: 1. must conform to the UID restrictions for allowable characters (meaning we can't have `Myfile.script.js` for example) and we want to make sure the ID follows the same pattern whether the transform is defined by file or through the UI.

It's already somewhat of a problem that applying it on a label follows `SCRIPT(js:mytransform.script):%s` for labels, `SCRIPT(js:mytransform.script)` for file based Profiles (I'm guessing), `transform('SCRIPT', 'js:mytransform.script`, someVariable)` in a rule (Rules DSL example) and `SCRIPT:js:mytransform.script` for at least the HTTP and MQTT bindings (maybe more?). Doubling these by having different IDs for file based transforms and UI based transforms would indeed be a nightmare.

At a high level:

- yes, it should look and mostly work the same no matter how the scripts are defined; we should not have one type of ID for file based and another for UI based.
- I'd personally like to see the feature somehow infer or know what the language the transform is written in so the user doesn't have to select it twice, once when creating the script and again when applying it (file extension? 🤷,  value in the transformation script JSONDB?, something else? ) or, if the user does select it twice once selected the second time when choosing it it only offers those .scripts that are written in that language.

That's my perspective, for what it's worth.

ccutrer:
> I didn‘t say that and this really marks the end of the discussion for me.

:(. I hope I didn't offend you. My personality is that I'm a direct and frank person, but I'm honestly not trying to be offensive, insulting, sarcastic, etc.

Now I'm brainstorming here:

**WHAT IF?**

What if the script transformation, instead of being a proper transformation service itself was simply a meta-transformation service that listens for new instances of script engines to be registered, and when they are it dynamically registers a new transformation service instance specifically for that engine? Then we could properly say "the JS transformation" or the "Ruby transformation." They would be referenced as "JS(mytransform.js)" from file-based configuration, which has the double-effect of making the proper extension match the transformation's name, which seems to be very important to @J-N-K, as well as restoring some semblance of backwards compatibility with openHAB 3.4's JS transform (though you'd have to install the JS automation addon, instead of the JS transformation provider; honestly I see this as a good change, because then it's up to each addon to document its support of transformations, as opposed to the SCRIPT transformation which has poor discoverability in the docs because it's not actually an addon at all - it's part of core). The UI would automatically change to simply choosing "JS transformation" instead of "script transformation" _and then_ "JS". Other transformation providers wouldn't need any change to satisfy any (now unnecessary) interface changes to support a single service supporting multiple file types. There are probably other pitfalls to this approach that I'm unaware of, and would be happy to have them pointed out to me.

jimtng:
> Do what you want, I won't veto that.

> If you want to go that route, do it fully: remove `.script`, extend the `TransformationService` interface with a method `.getSupportedTransformations`, implement that in all transformation add-ons and refactor `TransformationResource` to use that method so that UI and files-based transformations behave the same way.

Thanks for the pointers. I think that would be the ideal scenario, which would address all the issues raised here.

J-N-K:
> What if the script transformation, instead of being a proper transformation service itself was simply a meta-transformation service that listens for new instances of script engines to be registered, and when they are it dynamically registers a new transformation service instance specifically for that engine? Then we could properly say "the JS transformation" or the "Ruby transformation." They would be referenced as "JS(mytransform.js)" from file-based


This won't happen. We would in a situation even worse than before when we had only "JS Transformation": "Why can I use JS, DSL, Groovy and Ruby but not Python?". The transformation is universal and this will not change.

jimtng:
@rkoshak if we were to change the `SCRIPT` transformation syntax to `script.js` syntax, do you think we should:
a. completely drop the `js:file.script` syntax in 4.0, or
b. silently support it so as not to break people's existing installation, but not mention it in the documentation,

While (b) sounds nice, eventually we'll have to drop it anyway (when?) and we'll simply delay the inevitable. You know that if we maintain support, people will not go out of their way to change it until it breaks.

rkoshak:
> @rkoshak if we were to change the SCRIPT transformation syntax to script.js syntax, do you think we should:
> a. completely drop the js:file.script syntax in 4.0, or
> b. silently support it so as not to break people's existing installation, but not mention it in the documentation,

I don't really have a position either way really largely because I don't know whether/how either choice impacts the UI created script transformations. 

If we did "b", I would not leave it out of the docs. If it's supported, it should be documented, even if it's marked as deprecated in the docs. I really don't like undocumented features. 

If we do a, we should make sure it's done before OH 4.0 release. It's a breaking change for sure.

ccutrer:
> > What if the script transformation, instead of being a proper transformation service itself was simply a meta-transformation service that listens for new instances of script engines to be registered, and when they are it dynamically registers a new transformation service instance specifically for that engine? Then we could properly say "the JS transformation" or the "Ruby transformation." They would be referenced as "JS(mytransform.js)" from file-based
> 
> This won't happen. We would be in a situation even worse than before when we had only "JS Transformation": "Why can I use JS, DSL, Groovy and Ruby but not Python?". The transformation is universal and this will not change. A required metadata for scripting addons contradicts that.

I didn't say any required metadata. I said meta-transformation service. One in core, that's not actually a concrete transformation service, but would enumerate script engines as they're registered, and dynamically create a new transformation service for that script engine. So Python would automatically get one, even though it's not maintained.

jimtng:
> UI created script transformations.

How do you create a UI script transformation? AFAIK it isn't implemented yet.

If/when it's implemented, it should contain information about the scriptType / engine that it uses. The `scriptType` in `SCRIPT(scriptType:scriptUID)` syntax is needed only for identifying scriptType because of the chosen generic extension in "filename.script" currently. That would be fixed by simply allowing the actual language's extension `filename.js` / `filename.rb` etc. 

UI transformation being created in the UI will be retrieved through its scriptUID as a `Transformation` object and which contains `scriptType` information. It can be made to return scriptType not as `script` (which currently pertains to the name of the transformation) but as its actual script language, and then they will be checked against the currently loaded script engines, in exactly the same manner as file-based transformation scripts will be.

jimtng:
> The UI would automatically change to simply choosing "JS transformation" instead of "script transformation" and then "JS". Other transformation providers wouldn't need any change to satisfy any (now unnecessary) interface changes to support a single service supporting multiple file types. There are probably other pitfalls to this approach that I'm unaware of, and would be happy to have them pointed out to me.

> meta-transformation service. One in core, that's not actually a concrete transformation service, but would enumerate script engines as they're registered, and dynamically create a new transformation service for that script engine. So Python would automatically get one, even though it's not maintained.

This sounds very cool, although the implementation would be a bit "more complex" than the currently "static" `SCRIPT` transformation. 

Thinking out loud here: once implemented, the `SCRIPT()` transformation itself would cease to exist as far as the user is concerned, and we would simply refer to the "script transformation" as being done using the language's identifier `JS()`, `RB()`, `GROOVY()`, `JAVA()`, `DSL()`. Then in the UI, if you've got one or more of the scripting language addons installed, it would list them as separate types of transformation and when selected, it will list only transformations written in that selected language.

I think practically, from the end user's point of view, this is the smoothest and _easiest_ to understand transition from the old JS transformation.

image

rkoshak:
> How do you create a UI script transformation? AFAIK it isn't implemented yet.

There is an open PR created for it to add a transformations registry and the other stuff required in core. I'm not sure the status of it beyond a lot of work has been put into it. Once that's done there will need to be changes to MainUI to support it too which I don't think exists.

jimtng:
> There is an open PR created for it to add a transformations registry and the other stuff required in core. I'm not sure the status of it beyond a lot of work has been put into it.

Ah yes, I found that PR. I believe it can be adapted.

jimtng:
Latest update:

`SCRIPT` transformation itself is removed completely.
Instead, it creates the transformation for whatever language is registered in the system: 
- DSL
- JS
- RB
- GROOVY
- PY

It will only show the ones that are actually installed. Here, I have Ruby, Jython and JS Scripting installed and Groovy not installed, so GROOVY is not showing up on the list

image

Also note that it only shows the scripts for the selected language. Above shows only Ruby scripts, and below, only JS scripts:

image

The profile URI for Ruby transformation is `profile:transform:RB` which corresponds to the .rb extension
The profile URI for JS transformation is `profile:transform:JS` which corresponds to the .js extension
... and so on.

Furthermore, the configuration options can also be localised, although I can't seem to get anything (including MAP, REGEX, etc) to show any language other than English on my system, even after changing my language regional setting in openhab.

It should be compatible with the old JS transformation too. Here's my sitemap:
```
		Switch item=Black_Alert label="Black Alert [JS(bro.js):%s]"
```
bro.js:
```
(function (i) {
  return i + " Bro!";
})(input)
```

image

Even inline JS transformation works like it used to:

```
		Switch item=Black_Alert label="Black Alert [JS(| input + \" - Inline Bro!\"):%s]"
```


And now Ruby inline transformation works too:

```
		Switch item=Black_Alert label="Black Alert [RB(| \" #{input} - Ruby Inline Bro!\"):%s]"
```

jimtng:
And here's the result of the REST `rest/transformation/services`

```
[
  "JSONPATH",
  "REGEX",
  "MAP",
  "DSL",
  "RB",
  "JS",
  "PY"
]
```

And the result of `rest/profile-types` (with other profile entries removed for brevity)

```
[
  {
    "uid": "transform:RB",
    "label": "RB",
    "kind": "STATE",
    "supportedItemTypes": []
  },
  {
    "uid": "transform:JS",
    "label": "JS",
    "kind": "STATE",
    "supportedItemTypes": []
  },
  {
    "uid": "transform:PY",
    "label": "PY",
    "kind": "STATE",
    "supportedItemTypes": []
  },
  {
    "uid": "transform:DSL",
    "label": "DSL",
    "kind": "STATE",
    "supportedItemTypes": []
  }
]
```

jimtng:
superseded by #3487

PR #3479, opened 2023-03-24
Reduce rounding errors of RGB/HSB conversion and enhance ColorUtil

holgerfriedrich:
RGB colors are converted into HSBType, and sometimes also need to be transformed back to RGB. Whenever a binding uses  integer values to represent RGB or HSB, a lossless conversion is not possible without further measures. ```HSBType::fromRGB``` rounds to integer numbers. This results in a small difference when you receive an RGB value, store it in OH HSBType representation, and then sent it out as RGB. This is not unexpected.

But looking at openhab/openhab-addons#12643, I found larger deviations than expected.

This PR improves the rounding of results when values are converted to PercentType.
A few parameterized tests have been added to check if this improves the behavior. Some old tests could be deleted and included in the parameterized tests, but I wanted to keep them in for reference.

There is a risk that merging this PR beaks tests of some bindings.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/HSBType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/HSBType.java
@@ -113,9 +113,9 @@ public class HSBType extends PercentType implements ComplexType, State, Command 
     }
 
     /**
-     * @deprecated Use {@link ColorUtil#rgbToHsv(int[])} instead
+     * @deprecated Use {@link ColorUtil#rgbToHsb(int[])} instead
      *
-     * Create HSB from RGB
+     *             Create HSB from RGB
      *
      * @param r red 0-255
      * @param g green 0-255
@@ -124,12 +124,12 @@ public class HSBType extends PercentType implements ComplexType, State, Command 
      */
     @Deprecated
     public static HSBType fromRGB(int r, int g, int b) throws IllegalArgumentException {
-        return ColorUtil.rgbToHsv(new int[] { r, g, b });
+        return ColorUtil.rgbToHsb(new int[] { r, g, b });
     }
 
     /**
-     * @deprecated Use {@link ColorUtil#xyToHsv(double[])} or {@link ColorUtil#xyToHsv(double[], ColorUtil.Gamut)}
-     *             instead
+     * @deprecated Use {@link ColorUtil#xyToHsb(double[])} or {@link ColorUtil#xyToHsb(double[], ColorUtil.Gamut)}
+     *             instead.
      *
      *             Returns a HSBType object representing the provided xy color values in CIE XY color model.
      *             Conversion from CIE XY color model to sRGB using D65 reference white
@@ -141,7 +141,7 @@ public class HSBType extends PercentType implements ComplexType, State, Command 
      */
     @Deprecated
     public static HSBType fromXY(float x, float y) throws IllegalArgumentException {
-        return ColorUtil.xyToHsv(new double[] { x, y });
+        return ColorUtil.xyToHsb(new double[] { x, y });
     }
 
     @Override
@@ -165,29 +165,36 @@ public class HSBType extends PercentType implements ComplexType, State, Command 
         return new PercentType(value);
     }
 
+    /** @deprecated Use {@link ColorUtil#hsbToRgb(HSBType)} instead */
+    @Deprecated
     public PercentType getRed() {
         return toRGB()[0];
     }
 
+    /** @deprecated Use {@link ColorUtil#hsbToRgb(HSBType)} instead */
+    @Deprecated
     public PercentType getGreen() {
         return toRGB()[1];
     }
 
+    /** @deprecated Use {@link ColorUtil#hsbToRgb(HSBType)} instead */
+    @Deprecated
     public PercentType getBlue() {
         return toRGB()[2];
     }
 
     /**
-     * Returns the RGB value representing the color in the default sRGB
-     * color model.
-     * (Bits 24-31 are alpha, 16-23 are red, 8-15 are green, 0-7 are blue).
+     * @deprecated Use {@link ColorUtil#hsbTosRgb(HSBType)} instead.
+     *
+     *             Returns the RGB value representing the color in the default sRGB
+     *             color model.
+     *             (Bits 24-31 are alpha, 16-23 are red, 8-15 are green, 0-7 are blue).
      *
      * @return the RGB value of the color in the default sRGB color model
      */
+    @Deprecated
     public int getRGB() {
-        PercentType[] rgb = toRGB();
-        return ((0xFF) << 24) | ((convertPercentToByte(rgb[0]) & 0xFF) << 16)
-                | ((convertPercentToByte(rgb[1]) & 0xFF) << 8) | ((convertPercentToByte(rgb[2]) & 0xFF) << 0);
+        return ColorUtil.hsbTosRgb(this);
     }
 
     @Override
@@ -239,58 +246,10 @@ public class HSBType extends PercentType implements ComplexType, State, Command 
                 && getBrightness().equals(other.getBrightness());
     }
 
+    /* @deprecated Use {@link ColorUtil#hsbToRgb(HSBType)} instead */
+    @Deprecated
     public PercentType[] toRGB() {
-        PercentType red = null;
-        PercentType green = null;
-        PercentType blue = null;
-
-        BigDecimal h = hue.divide(BIG_DECIMAL_HUNDRED, 10, RoundingMode.HALF_UP);
-        BigDecimal s = saturation.divide(BIG_DECIMAL_HUNDRED);
-
-        int hInt = h.multiply(BigDecimal.valueOf(5)).divide(BigDecimal.valueOf(3), 0, RoundingMode.DOWN).intValue();
-        BigDecimal f = h.multiply(BigDecimal.valueOf(5)).divide(BigDecimal.valueOf(3), 10, RoundingMode.HALF_UP)
-                .remainder(BigDecimal.ONE);
-        PercentType a = new PercentType(value.multiply(BigDecimal.ONE.subtract(s)));
-        PercentType b = new PercentType(value.multiply(BigDecimal.ONE.subtract(s.multiply(f))));
-        PercentType c = new PercentType(
-                value.multiply(BigDecimal.ONE.subtract((BigDecimal.ONE.subtract(f)).multiply(s))));
-
-        switch (hInt) {
-            case 0:
-            case 6:
-                red = getBrightness();
-                green = c;
-                blue = a;
-                break;
-            case 1:
-                red = b;
-                green = getBrightness();
-                blue = a;
-                break;
-            case 2:
-                red = a;
-                green = getBrightness();
-                blue = c;
-                break;
-            case 3:
-                red = a;
-                green = b;
-                blue = getBrightness();
-                break;
-            case 4:
-                red = c;
-                green = a;
-                blue = getBrightness();
-                break;
-            case 5:
-                red = getBrightness();
-                green = a;
-                blue = b;
-                break;
-            default:
-                throw new IllegalArgumentException("Could not convert to RGB.");
-        }
-        return new PercentType[] { red, green, blue };
+        return ColorUtil.hsbToRgbPercent(this);
     }
 
     /**
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/HSBType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/HSBType.java
@@ -34,6 +34,7 @@ import org.openhab.core.util.ColorUtil;
  *
  * @author Kai Kreuzer - Initial contribution
  * @author Chris Jackson - Added fromRGB
+ * @author Andrew Fiddian-Green - closeTo (copied from binding)
  */
 @NonNullByDefault
 public class HSBType extends PercentType implements ComplexType, State, Command {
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/HSBType.java
```diff
@@ -274,7 +275,7 @@ public PercentType[] toRGB() {
         BigDecimal h = hue.divide(BIG_DECIMAL_HUNDRED, 10, RoundingMode.HALF_UP);
         BigDecimal s = saturation.divide(BIG_DECIMAL_HUNDRED);
 
-        int hInt = h.multiply(BigDecimal.valueOf(5)).divide(BigDecimal.valueOf(3), 10, RoundingMode.HALF_UP).intValue();
+        int hInt = h.multiply(BigDecimal.valueOf(5)).divide(BigDecimal.valueOf(3), 0, RoundingMode.DOWN).intValue();
```
holgerfriedrich, line 278: Using 10 decimal places and the dropping all decimal places afterwards can be replaced by using 0 ddecimal places and roundmode DOWN

andrewfg:
^
FWIW I added the following method to my binding JUnit tests for comparing HSB colours; so you might consider adding it to ColorUtil as well ;)

```
    /**
     * Helper method for checking if expected and actual HSBType color parameters lie within a given percentage of each
     * other. This method is required in order to eliminate integer rounding artifacts in JUnit tests when comparing HSB
     * values. Asserts that the color parameters of expected and actual are within delta percent of each other.
     *
     * @param expected an HSBType containing the expected colour.
     * @param actual an HSBType containing the actual colour.
     * @param delta the maximum allowed percentage difference between the two (0..99 percent).
     */
    private static void assertHSBEqual(HSBType expected, HSBType actual, float delta) {
        if (delta <= 0f || delta > 99f) {
            throw new IllegalArgumentException("'delta' out of bounds");
        }
        double[] exp = ColorUtil.hsbToXY(expected);
        double[] act = ColorUtil.hsbToXY(actual);
        double max = delta / 100.0f;
        assertTrue((Math.abs(exp[0] - act[0]) < max) && (Math.abs(exp[1] - act[1]) < max));
    }
```

bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/HSBTypeTest.java
```diff
@@ -198,4 +206,109 @@ public void testConstructorWithIllegalSaturationValue() {
...
+    @ArgumentsSource(HsbRgbProvider.class)
+    public void paremeterizedTestRgbToRgbConversion(int[] hsb, int[] rgb) {
+        final HSBType hsbType = HSBType.fromRGB(rgb[0], rgb[1], rgb[2]);
+
+        final PercentType[] rgbPercent = hsbType.toRGB();
+        compareRgbToRgbValues(rgb, rgbPercent, 0);
+    }
+
+    // compare two rgb representations based on squared sum of differences, prepare a readable test message on fail
+    private void compareRgbToRgbValues(final int[] rgb, final PercentType[] rgbPercent, int maxSquaredSum) {
```
J-N-K, line 282:
```suggestion
    private void assertEqualsRGB(final int[] rgb, final PercentType[] rgbPercent, int maxSquaredSum) {
```

J-N-K (changes requested):
Thanks, I fully agree with your code changes.

Regarding the tests: Please group different parts: 

- all value providers
- all test methods
- internal helpers

Please also check @andrewfg's comment.

bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/HSBTypeTest.java
holgerfriedrich: Thanks @J-N-K , I will have a look tomorrow. 

Should I move the whole RGB/HSV conversion to ```ColorUtil```` together with the suggested function above?

Maybe it would be useful to deprecate the getter functions for single color channels like ```HSBType::getRed()```. The implementation is highly inefficient (converting to RGB and picking just one channel), especially when the users of this function calls the getters for all channels separately. getRed() is used ~50x in bindings (w/o test).

holgerfriedrich:
This PR has become a restructuring of HSBType RGB conversion.
- The complete RGB/HSB conversion has moved to ```ColorUtil```.
- Several functions in ```HSBType``` have been deprecated.
- I renamed ...hsv to ...hsb to be consistent with the naming of HSBType.

```ColorUtil``` API now defaults to integer data types where possible. I have noticed that several bindings use their own way to convert percentages to bytes, etc... Though, PercentType can also be obtained (but is not recommended).

@J-N-K maybe you can give some feedback on the current state.
@andrewfg  thanks, I added the helper function to the API (but as a boolean function without using the assert call, as it is only available in the test classes).

**Review by J-N-K — CHANGES_REQUESTED** [review]

File: bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java [hunk 3]
```diff
@@ -295,7 +455,61 @@ public Point closest(Point p) {
...
     private static boolean inRange(double val) {
         return val >= 0.0 && val <= 1.0;
     }
+
+    private static int convertColorPercentToByte(PercentType percent) {
+        return percent.toBigDecimal().multiply(BigDecimal.valueOf(255))
+                .divide(BIG_DECIMAL_HUNDRED, 0, RoundingMode.HALF_UP).intValue();
+    }
+
+    /* Helper functions */
```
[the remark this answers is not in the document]
holgerfriedrich: the helper functions are intended to be used in bindings or binding tests.... I could move them if you wish....

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -112,7 +272,7 @@ public static HSBType xyToHsv(double[] xy) {
      * @return the corresponding {@link HSBType}.
      * @throws IllegalArgumentException when input array has wrong size or exceeds allowed value range
      */
-    public static HSBType xyToHsv(double[] xy, Gamut gamut) {
+    public static HSBType xyToHsb(double[] xy, Gamut gamut) throws IllegalArgumentException {
```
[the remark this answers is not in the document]
holgerfriedrich: I created a change for knx and deconz, see openhab/openhab-addons#14724

holgerfriedrich:
I think there is no need to mark this as an API breaking change, renaming of functions has been done only for ```xyToHs...``` which got recently introduced for the knx / deconz backport.

Deprecation will cause warnings in other bindings as well, which may need to be fixed later.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -12,23 +12,32 @@
...
- * The class is based work from Erik Baauw for the <a href="https://github.com/ebaauw/homebridge-lib">Homebridge</a>
- * project
+ * The implementation of HSB/CIE conversion is based work from Erik Baauw for the
+ * <a href="https://github.com/ebaauw/homebridge-lib">Homebridge</a>
+ * project.
  *
  * @author Jan N. Klug - Initial contribution
+ * @author Holger Friedrich - Transfer RGB color conversion from HSBType, improve RGB conversion, restructuring
+ * @author Chris Jackson - Added fromRGB (moved from HSBType)
+ * @author Andrew Fiddian-Green - assertHSBEqual/hsbEquals (copyied from binding)
```
jlaur, line 35:
```suggestion
 * @author Andrew Fiddian-Green - assertHSBEqual/hsbEquals (copied from binding)
```

jlaur:
> Deprecation will cause warnings in other bindings as well, which may need to be fixed later.

I have a small question about these deprecations. I'm wondering if it would make sense to still have some static factory methods for creating `HSBType` conveniently, e.g. `HSBType valueOf(int r, int g, int b)` and `HSBType valueOf(double x, double y)` - or even with more explicit naming (like `ofRGB`)? Are the methods being deprecated because of the naming, or is it a general migration from static factory methods to constructors, or some specific issue with `HSBType`?

holgerfriedrich:
> I have a small question about these deprecations. I'm wondering if it would make sense to still have some static factory methods for creating `HSBType` conveniently, e.g. `HSBType valueOf(int r, int g, int b)` and `HSBType valueOf(double x, double y)` - or even with more explicit naming (like `ofRGB`)? Are the methods being deprecated because of the naming, or is it a general migration from static factory methods to constructors, or some specific issue with `HSBType`?

Right now, the ```HSBType``` can be created using ```ColorUtil.rgbToHsb(int[])``` or ```ColorUtil.rgbToHsb(new int[]{r, g, b})```.
The conversion itself moved to ColorUtil, but easy to use valueOf methods may be useful. I think there I just followed what @J-N-K did with ```fromXY``` when he introduced ColorUtil to core.

Actually I was more thinking of the deprecation of getters for single colors, ```getRed()``` etc. I tried to deprecate them because they internally did the conversion from HSB to RGB (for 3 colors), and then only returned one. In the bindings, typically all colors are requested via the getter functions, resulting in 3 HSB to RGB conversions - where one would be sufficient.

I do not mind removing the deprecation note from ```fromRGB``` or adding ```valueOf``` methods (for int[], and r,g,b) - WDYT?

jlaur:
> Right now, the `HSBType` can be created using `ColorUtil.rgbToHsb(int[])` or `ColorUtil.rgbToHsb(new int[]{r, g, b})`. The conversion itself moved to ColorUtil, but easy to use valueOf methods may be useful. I think there I just followed what @J-N-K did with `fromXY` when he introduced ColorUtil to core.

Sure, moving the implementation to `ColorUtil` is totally fine, it's only the deprecation of the static factory methods I'm questioning. I know @J-N-K deprecated `HSBType fromXY(float x, float y)` before your changes, so I think it's important to understand his reasons - there could be valid reasons I'm not seeing. I noticed it when some addons starting giving deprecation warnings when doing:
- openhab/openhab-addons#14629
- openhab/openhab-addons#14630
- openhab/openhab-addons#14631

> I do not mind removing the deprecation note from `fromRGB` or adding `valueOf` methods (for int[], and r,g,b) - WDYT?

Let's wait for @J-N-K. If we do this, we should probably also bring back (remove deprecation note for) the convenience method for XY.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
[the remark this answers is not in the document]
J-N-K: In general bindings should not do tests that are already present in core, i.e. they should not test color conversion itself.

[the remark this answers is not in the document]
holgerfriedrich: I assumed that comparing two color values is a helpful functionality. Maybe I am wrong here, but if we do not offer, there will be many different implementations in the bindings. Not necessarily in the test, but whereever colors are compared.
The functions below are just boolean functions and do not call the assert... functions (anywhere only available in the test folder)

holgerfriedrich:
> I think that especially the factory method should stay, we could think of adding a ```.fromXYY(double[])``` or at least ```.fromXYY(double x, double y, double Y)```.

> Regarding the getters for R, G, B: We use HSB model internally, so getters for H,S,B are fine, for everything else the ColorUtil is the correct place. We could also remove them completely (together with fromXY), but I'm not sure how common they are for use in rules.

They seem to be used in many places, so I would go for the deprecation here....

Proposal:
- deprecate ```getRed```, ```getGreen```, ```getBlue```, ```getRGB```, ```toRGB```.
- keep ```fromRGB(r,g,b)```

holgerfriedrich:
@J-N-K could you please have a look again? Shall I remove the helpers for comparing color values? Can the deprecated functions stay as implemented?

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -295,7 +455,64 @@ public Point closest(Point p) {
...
+     * Helper method for checking if two RGB colors (int[3], 0..255) are close to each other. A maximum distance is
+     * specified
+     * as sum of squared distances of the integer components.
+     *
+     * @param first an int array containing the first RGB color.
+     * @param second an int array containing the second RGB color.
+     * @param maxSquaredSum the maximum allowed squared sum of differences.
+     * @throws IllegalArgumentException if array size is not 3.
+     */
+    public static boolean closeToRgb(final int[] firstRgb, final int[] secondRgb, int maxSquaredSum)
```
J-N-K, line 502: This should be removed. Bindings should neither test color conversion (as that is part of core), nor should they handle RGB in any other way than converting not to `HSBType`. We are much more consistent if we try to use as much `HSBType` as possible. If we want to make RGB a first-class citizen then we should discuss`RGBType`, but that is far out of scope here.

J-N-K (changes requested):
Otherwise LGTM.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
andrewfg: ;) not to mention XYType ..

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -295,7 +455,64 @@ public Point closest(Point p) {
...
+     */
+    public static boolean closeToHsb(HSBType expected, HSBType actual, float delta) throws IllegalArgumentException {
+        if (delta <= 0f || delta > 99f) {
+            throw new IllegalArgumentException("'delta' out of bounds");
+        }
+        double[] exp = ColorUtil.hsbToXY(expected);
+        double[] act = ColorUtil.hsbToXY(actual);
+        double max = delta / 100.0f;
+        return ((Math.abs(exp[0] - act[0]) < max) && (Math.abs(exp[1] - act[1]) < max));
+    }
```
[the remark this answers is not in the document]
holgerfriedrich: done, moved to ```HSBType```, changed percentage range to 0..1 instead of 1..100, added test.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
holgerfriedrich: done, moved content of function to ```ColorUtilTest::assertRgbEquals``` (still needed for the test)

J-N-K (approved):
LGTM, thanks.

PR #3481, opened 2023-03-25
[uom] Add `unit` metadata for NumberItem

J-N-K:
This is a new approach to solving #3282. It keeps the dimension when the item is created but it separates the state description from the internal unit.

With this PR:

- The unit for the item is either the system default (provided by the `UnitProvider`, depends on dimension and system of measurement) or the unit configured in the new `defaultUnit` metadata (takes precedence if present) 
- The state description is ONLY used for displaying the values (before it was also used to determine the item's unit)
- States are converted to the item unit before the state is set, for plain `Number` items units are stripped, `DecimalType` updates to dimension items are considered to be in the configured unit, incompatible units are rejected.
- Commands are handled the same as states, the unit of `QuantityType` to a plain item is stripped, the item unit is added to `DecimalType` if there is a dimension, incompatible units are rejected. In contrast to state updates, the unit for commands is NOT converted.
- The `CommunicationManager` strips the unit and sends `DecimalType` if the channel has the accepted item-type `Number`, it adds the item-unit if the dimension of the item and the channel are the same. Commands with incompatible dimension are rejected. "Guessing" a unit based on the channel type when the item is dimensionless has been removed as it is confusing. 
- A new UoM profile is introduced. If a unit is configured for that profile, it is added to a `DecimalType` value, and a `QuantityType` value is converted to that unit, if no unit is configured, the unit is stripped from a `QuantityType` value. 

To fully solve #3282 we also need to merge #3141 and check where `ItemStateEvent` was used before.


spacemanspiff2007:
> * The unit for the item is either the system default

Do I understand this correctly that this will keep e.g. `Number:Temperature`

> * or the unit configured in the new `defaultUnit` metadata (takes precedence if present)

Wasn't it `unit` before? `defaultUnit` sounds like a fall back yet you write this takes precedence and overrides all other settings.
I think `unit` would be a much better wording (if I understand this correctly).

J-N-K:
Yes, `Number:Temperature` is kept. It's less invasive and we can still workaround the limitations we found before. One benefit is that the UI now can propose a unit, because it already knows the dimension.

Regarding the naming: it was always `defaultUnit` in this branch (which is the direct successor of #3248). It can be easily changed to `unit` if that is the result of the discussion.

One other thing that could be implemented (independently) is changing the system default unit (which is currently hard-coded, depending on dimension and system of measurement).

spacemanspiff2007:
> Yes, `Number:Temperature` is kept. It's less invasive and we can still workaround the limitations we found before. One benefit is that the UI now can propose a unit, because it already knows the dimension.

So if I select `Number:Temperature` the UI can suggest `unit="°C` instead of selecting `unit="°C"` directly?
I get that this makes migration easier but it obfuscates the whole UoM concept again and provides no real benefit.
The item unit can't and shouldn't be changed automatically (after initial creation) so what do we use `Number:Temperature` for?
The current dimension can also be derived from `°C` so when changing the locale the displayed value can still be changed accordingly.



> It can be easily changed to `unit` if that is the result of the discussion.

Depends on how it works. 
If it's the fallback that is used when nothing other is there then `defaultUnit` is good.
If it overrides everything else and takes precedence it should not have `default` in the name since my understanding of default is more of a fallback in case nothing else is specified.

J-N-K:
Regarding `defaultUnit/unit` it probably depends on the POV. It is the unit that is assumed for values that don't carry a unit (it's therefore the "default"), but it is also the unit that is used to store the value internally.

lolodomo:
What would be the syntax in items file?

J-N-K:
@lolodomo Like any other metadata:

```
Number:Temperature Freezer_Temp "Freezer [%.1f %unit%]" { channel="foo:bar:....", defaultUnit="K" }
```

J-N-K:
@spacemanspiff2007 

If nothing is configured, the system default (in my case `°C`) is used. By knowing the dimension, the UI can "validate" anything I configure as unit (e.g. it could reject `kg` for a `Number:Temperature` and prevent setting a unit that will later result in issues.

spacemanspiff2007:
> It is the unit that is assumed for values that don't carry a unit (it's therefore the "default"), but it is also the unit that is used to store the value internally.

But if it carries a unit (e,g, `°F`) it will still be normalized to `°C` so default does not apply there.
Default would mean the new unit would be `°F` yet the item will still carry the unit `°C`.
The field will configure the item unit so it should be named accordingly `unit`.



> If nothing is configured, the system default (in my case `°C`) is used. By knowing the dimension, the UI can "validate" anything I configure as unit (e.g. it could reject `kg` for a `Number:Temperature` and prevent setting a unit that will later result in issues.

Imho you're mixing two things - how the GUI presents the setting and how it actually is configured.
I think it would be nice to have some kind of tree where you move down (e.g. first you select "temperature" which reduces the available values to `°F`, `°C` and `°K`. Then the user selects `°F` which will internally set `unit="°F"`.

However there is no need for a `Number:Temperature` item because the unit can be inferred from the item metadata.
This will introduce two places where the user has to configure the same thing: first the item type and then the item metadata.
There should be one place to define this and one place only. There should be at maximum two number item types: one with a uom set and one without a unit and even those two types can be hidden from the user.

Also if I as a user don't explicitly set the unit it might lead to the impression that the unit will change when I e.g. change my system locale which is something we explicitly don't want to do. We neither want to change the normalization value automatically for existing items nor do we want to create the impression that openHAB will behave that way.

J-N-K:
As I said earlier: if it's `unit` or `defaultUnit` is a minor issue and can be changed in seconds. The more important discussion is the general UoM handling.

If the user creates the item from a thing channel with "Add link to item  -> Create new item" then the item type is automatically set from the accepted item type of the channel. Nothing that can go wrong there. If the user doesn't want anything special and the system default unit is fine, then nothing else needs to be configured. If he wants to use a different unit, he can configure the unit metadata to use another unit. In this case UI can make suggestions (a list of common units for this dimension) and verify the compatibility of the selected unit.

It's debatable if the unit metadata should be automatically added when the item is created and initialized with the system default unit. I would be open for that, this would also solve the "change system of measurement". But this is IMO a rare case and we'll probably have a 50/50 chance that someone wants to change the item units in that case or not.

spacemanspiff2007:
I think I understand what you are describing.

> from the accepted item type of the channel.

But why doesn't the channel type say `Number` and provide an additional optional field "unit" which would say "Temperature".
It can be easily included e.g. in the CannelTypeDTO, That way the UI still can provide the functionality you described but there is only once place where the unit is configured for the item and that is the item metadata.

> But this is IMO a rare case and we'll probably have a 50/50 chance that someone wants to change the item units in that case or not.

I think explicit is better than implicit and in this case I'd stick to that. It takes a very short time to configure it but it's such an important field that the user should be aware of it. So from the current perspective I'd vote against that.

J-N-K:
Strictly typed item-types (liked it is now) have some advantages. The channel defines `Number:Temperature` and exactly that item is created. The user can then change from `°C` to `K` or `°F` but not to `kg`, because `kg` is not compatible with the dimension `Temperature`. 

This also applies to textual configuration, if you define an incompatible unit, the item will still work and provide values. Breaking that requires deliberately changing the item-type, not just a typo in the unit (e.g. `kW` vs. `kWh`). If you determine the dimension from the unit the dimension will change from `Power` to `Energy` and future state-updates fail.

I don't think there is a benefit of using `Number` with additional information `Temperature` and unit `°C` over `Number:Temperature` and unit `°C`. It's just another way of defining the same and it is not backward compatible.

spacemanspiff2007:
> If you determine the dimension from the unit the dimension will change from `Power` to `Energy` and future state-updates fail.

But would that fail quickly - e.g. if I change the unit from `kW` to `kWh` it would be the next item update that would generate an error and the user would see that it was a mistake. Since the normalization unit is something that normally is set up once I don't see the need for that kind of error check.

Also if I am not super fluent in English I'll quickly get confused with `Number:Energy` and `Number:Power` (or more exotic ones) but I'll most definitely know that I'll want `kWh`. The use case for most users will be to think of the unit, then look up the corresponding item type in the docs and blindly use that making that kind of error check obsolete. 

> I don't think there is a benefit of using `Number` with additional information `Temperature` and unit `°C` over `Number:Temperature` and unit `°C`. It's just another way of defining the same and it is not backward compatible.

I think this is a misunderstanding. My suggestion is that the item only defines the unit and scale and the channel defines the item type (e.g. String/Number) and optionally an UoM (in case of Number).
E.g. `Number` and `Temperature` is defined for the Channel, `°C` is defined for the item.
That way there is no redundant information and all the error checks are still possible.

There should be one way to define the uom and one way only. It just makes no sense to say: "This item should have the temperature in °C and yes - I'm really sure it's a temperature". You're also not saying "Today it'll be 15°C and what I just told you was a temperature". Everyone know that already from the °C.
Also you're not saving me from typos in the item name and state either. :wink:

lolodomo:
Just one remark as I see that some of you want to remove the number dimension: we cannot break the setup of 100% of users by making the item syntax not backward compatible...

J-N-K:
For managed things #3268 would solve the issue. This will also be needed for the approach taken here. Textual configuration will in any case require manual action, but that is exactly what those users want.

spacemanspiff2007:
> Just one remark as I see that some of you want to remove the number dimension: we cannot break the setup of 100% of users by making the item syntax not backward compatible...

We had a huge breaking changes from OH1 to OH2 and from OH2 to OH3 and people figured it out, too.
For a major version number increase people will expect breaking changes and I think it's okay to have those.
Moving forward without change is impossible.
There will be tools to help with the migration, e.g. the CLI upgrade tool @J-N-K mentioned and I offered to write a python script that will help with the migration of the *.items files.
So I think the migration path should be relatively straight forward and more or less painless.
If it's properly documented I see no issue.

spacemanspiff2007:
@J-N-K what are your current thoughts on dropping e.g. 
`Number:Temperature    MyItem   {unit="°C"}`
in favor of
`Number   MyItem   {unit="°C"}`

If you're not sold can you share your doubts so I can better understand them.
I'll try to improve my arguments accordingly.

J-N-K:
We have discussed that before (see the attached issue) and there have been concerns regarding clarity what the item actually stores. That is why I decided to go on with this PR.

lolodomo:
I see a big advantage to @J-N-K 's solution, it is (almost fully) backward compatible. @J-N-K : please correct me if I am wrong. Even if the user does not update their items definition in config files, it will continue to work as before. In case the user was not using the default system unit for an item, the only impacts would be the values displayed in charts. No change in UIs. Not sure if rules could be impacted ?

J-N-K:
@lolodomo The breaking change is that currently we use the state description for determining the unit and if that is not set the system default unit. With this PR the state description is not considered, but the `defaultUnit` (or `unit`) metadata and if that is not set the system default unit. That is one of the main points here: decouple internal representation of the value and display value.

So if someone currently uses `%.2f K` for a temperature item, the item's unit is considered to be `K`. In the future it is `°C` (or `°F`, depending on the locale) if he does not set `defaultUnit="K"`. I have a PR (#3268) that fixes these issues for managed items (by setting the `defaultUnit` when a unit is found in the state description and prints a warning during upgrade if the unit cannot be determined), but there is not much we can do for textual configuration, these user will need to do it by hand - but they chose to do so.

lolodomo:
> So if someone currently uses `%.2f K` for a temperature item, the item's unit is considered to be `K`. In the future it is `°C` (or `°F`, depending on the locale) if he does not set `defaultUnit="K"`.

Yes, I clearly understand that point but I have the feeling this change will be almost fully transparent for the user ? The values displayed in UIs will be the same as before. As you say his rules will not be impacted, the only impact I can guess for the user is that now in charts the item values will be in °C rather than in Kelvin before.

> bindings might be affected depending on my last question in https://github.com/openhab/openhab-core/issues/3282.

Not sure what question you are referring. In case we have to update all the bindings, this is critical problem !

J-N-K:
IMO bindings that can be linked to `Number` items with values that MIGHT have a unit (i.e. all measured values, not integers that represent different states) should support `DecimalType` and `QuantityType`, but I'm aware of the fact that this is not the case.

That is why I proposed a profile that strips units if necessary (effectively converting `QuantityType` to `DecimalType`. If a binding absolutely needs `DecimalType`, the user can apply that profile and everything works without code changes.

J-N-K:
The migration tool picks the unit from the state description, so when the unit is correct now, it'll also be correct after the migration. If the state description is not set, the system default will be used (now and in future).

lolodomo:
Except very advanced users, I believe any normal user does not care about the unit used to store internally the item state. What is important for a normal user is the value displayed in UIs and having his rules are working (potentially comparing states with value). A normal user will certainly never use any REST API.

spacemanspiff2007:
> A normal user will certainly never use any REST API.

But a normal user will most certainly use persistence and this is the mechanism how to pick the persisted unit and scale.

florian-h05 (commented):
After following both the discussion in this PR and the linked issue #3282 for some time, I decided to **signal my support for this PR** by approving it (or commenting my approval, I am somehow unable to approve) 👍
Decoupling the internal representation (or name it the backend-representation, since persistence, REST API etc. are affected by it as well) and the UI representation of a UoM Item is an important step forward!

It seems like a good compromise to me between doing nothing (which is not an option) and the initial proposal to me, and regarding that users would require migration for DB values in any case: I would veto that, with this PR, I think there are many users which won‘t have to migrate their DB values. And even if this is required, if it is documented, I don‘t see a problem with it.

clinique:
I also do support. All these discussions that happened, plus the work involved - we're in the best moment (pre OH4) to onboard this. It would be a pitty to let it apart.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/CommunicationManager.java
```diff
@@ -427,118 +423,37 @@ private <T extends Type> void handleEvent(String itemName, T type, @Nullable Str
...
-                        && State.class.isAssignableFrom(typeClass) && originalType instanceof State state) {
-                    T ret = (T) state.as((Class<? extends State>) typeClass);
-                    if (logger.isDebugEnabled()) {
-                        logger.debug("Converted '{}' ({}) to accepted type '{}' ({}) for channel '{}' ", originalType,
-                                originalType.getClass().getSimpleName(), ret, ret.getClass().getName(),
-                                channel.getUID());
-                    }
-                    return ret;
-                }
-            }
```
J-N-K, line 483: This is interesting. I removed the code and replaced it by the `Dimmer/HSBType` check, because it did only affect the return type in two cases:

```
[ERROR] Failures: 
[ERROR]   CommunicationManagerConversionTest.testCommand:114 CommunicationManager did a conversion for target item Rollershutter from class org.openhab.core.library.types.HSBType to class org.openhab.core.library.types.PercentType
[ERROR]   CommunicationManagerConversionTest.testCommand:114 CommunicationManager did a conversion for target item Dimmer from class org.openhab.core.library.types.HSBType to class org.openhab.core.library.types.PercentType    
```

One is covered by the simple check that I added and the other one is IMO wrong. Sending `HSBType` to a `RollershutterItem` doesn't make any sense.

spacemanspiff2007:
> e.g. for temperature most will use the future `defaultUnit`. 

Just a heads up - we already settled on `unit`, because `defaultUnit` would imply that it's taken when no unit is supplied but the unit transforms the value in any case so `default` is very misleading.

kaikreuzer:
I second the opinion that this solution is smoother than #3248 and I support it.

But as it was discussed, there is a need for adding the unit metadata for backward compatibility, especially for users that do not use the same unit as defined by the measurement system - if this isn't done with the migration, persistence data (e.g. int he default rrd4j) might quickly be messed up.
Hence my question: Why did you close the PR for the CLI tool, @J-N-K?

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/CommunicationManager.java
```diff
@@ -427,118 +423,37 @@ private <T extends Type> void handleEvent(String itemName, T type, @Nullable Str
...
-
-        // DecimalType command sent to a NumberItem with dimension defined:
-        if (originalType instanceof DecimalType type && hasDimension(item, acceptedItemType)) {
-            @Nullable
-            QuantityType<?> quantityType = convertToQuantityType(type, item, acceptedItemType);
-            if (quantityType != null) {
-                return (T) quantityType;
-            }
-        }
+        String channelAcceptedItemType = channel.getAcceptedItemType();
```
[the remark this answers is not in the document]
kaikreuzer: Yes, I would agree - this feels too early here - adapting to the channel type must only happen after the profile has been applied. But this can probably be tackled in a separate PR, since it is like this since the beginning.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/CommunicationManager.java
```diff
@@ -427,118 +425,42 @@ private <T extends Type> void handleEvent(String itemName, T type, @Nullable Str
...
-        if (acceptedItemType == null) {
+        if (channelAcceptedItemType == null) {
             return originalType;
         }
 
-        List<Class<? extends T>> acceptedTypes = acceptedTypesFunction.apply(acceptedItemType);
-        if (acceptedTypes == null) {
-            return originalType;
+        // handle Number-Channels for backward compatibility
+        if ("Number".equals(channelAcceptedItemType) && originalType instanceof QuantityType<?> quantityType) {
```
kaikreuzer, line 435: Could this string be replaced by `CoreItemFactory.NUMBER`?

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/CommunicationManager.java
kaikreuzer: Yes, sending `HSBType` to rollershutters does not need to be supported.
The check seems to cover hierarchical types in a very generic way, but since we are unlikely to add further hierarchical types in future, I agree that this complex check isn't needed here.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/CommunicationManager.java
```diff
@@ -427,118 +425,42 @@ private <T extends Type> void handleEvent(String itemName, T type, @Nullable Str
...
-    private @Nullable QuantityType<?> convertToQuantityType(DecimalType originalType, Item item,
-            @Nullable String acceptedItemType) {
-        if (!(item instanceof NumberItem)) {
-            // PercentType command sent via DimmerItem to a channel that's dimensioned
-            // (such as Number:Dimensionless, expecting a %).
-            // We can't know the proper units to add, so just pass it through and assume
-            // The binding can deal with it.
-            return null;
+        // handle HSBType/PercentType
+        if ("Dimmer".equals(channelAcceptedItemType) && originalType instanceof HSBType hsb) {
```
kaikreuzer, line 451: `CoreItemFactory.DIMMER`?

bundles/org.openhab.core/src/main/java/org/openhab/core/items/MetadataAwareItem.java
```diff
@@ -0,0 +1,47 @@
...
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.items;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
+/**
+ * The {@link MetadataAwareItem} is a
```
kaikreuzer, line 18: is a ... ?

bundles/org.openhab.core/src/main/java/org/openhab/core/library/items/NumberItem.java
```diff
@@ -43,26 +49,41 @@
  * @author Kai Kreuzer - Initial contribution
  */
 @NonNullByDefault
-public class NumberItem extends GenericItem {
-
+public class NumberItem extends GenericItem implements MetadataAwareItem {
+    public static final String TYPE = "Number";
```
kaikreuzer, line 53: This does not seem to be used (and duplicating `CoreItemFactory.NUMBER`).

itests/org.openhab.core.auth.oauth2client.tests/itest.bndrun
```diff
@@ -68,4 +68,5 @@ Fragment-Host: org.openhab.core.auth.oauth2client
 	net.bytebuddy.byte-buddy;version='[1.12.19,1.12.20)',\
 	net.bytebuddy.byte-buddy-agent;version='[1.12.19,1.12.20)',\
 	org.mockito.mockito-core;version='[4.11.0,4.11.1)',\
-	org.objenesis;version='[3.3.0,3.3.1)'
+	org.objenesis;version='[3.3.0,3.3.1)',\
+	org.osgi.service.cm;version='[1.6.0,1.6.1)'
```
kaikreuzer, line 72: Are all these changes of adding `org.osgi.service.cm` intended? It does not seem to be related to this PR, so it's probably better to revert them.

kaikreuzer (changes requested):
Thanks @J-N-K, this looks pretty good to me.
Just a few small comment below, if you allow.

J-N-K:
@kaikreuzer I think I addressed your comments. I would like to move the uom profile to a separate PR. It is not really needed at the moment and I would like to do check if one profile with one setting really works for all use-cases or if we need something different. I re-opened the CLI-tool (#3268), it should probably be merged together.

kaikreuzer (approved):
Thank you!

kaikreuzer:
@J-N-K Could you please resolve the conflict?

lolodomo:
As this PR is about to be merged, @J-N-K could you please confirm that it is without any impact on bindings (no need to update them) ?

Regarding user rules, can you please explain what will be returned by `item.state` **if you don't update the items definition** ? Is it the new internal representation with default unit or the same value as before using the format from the label pattern ? The answer is important to explain to users and depending on this answer the changes to apply to items definition (adding or not the new unit field) are not the same.

As the default units will now become very important, do we have a page in the documentation showing what is the default unit for each number dimension ? I guess users will ask this question. If the default unit is identical to the unit I set in the label pattern, I understand no change is required and I guess it could be the case for a majority of my items.

J-N-K:
@lolodomo 

- Bindings: I don't think there is a need to change anything (except from rare corner cases where removed methods are used). I'll provide a PR for that if necessary.
- item.state returns the internal state (which it also did before). The difference is that the state description is no longer considered when setting the internal state. If no state description is set, there is no change at all. If a state description is set AND the state description specifies a unit other than the default unit (if the default unit is `°C` and you set `%.1f °C` it behaves as before), then the state is returned in the default unit (instead of the state description unit). This can be fixed by running the upgrade tool once.
- When `%unit%` is set, the migration tool prints a warning that you should consider setting `unit` metadata. `%unit` displays the internal state, i.e. the default unit.
- The list is here: https://www.openhab.org/docs/concepts/units-of-measurement.html. But it is incomplete. I'll prepare a PR for that.

Where would we add documentation for manual operation of the upgrade tool? This is also missing, but needed.

lolodomo:
> * Bindings: I don't think there is a need to change anything (except from rare corner cases where removed methods are used). I'll provide a PR for that if necessary.

Fine

> * item.state returns the internal state (which it also did before). The difference is that the state description is no longer considered when setting the internal state. If no state description is set, there is no change at all. If a state description is set AND the state description specifies a unit other than the default unit (if the default unit is `°C` and you set `%.1f °C` it behaves as before), then the state is returned in the default unit (instead of the state description unit). This can be fixed by running the upgrade tool once.

Ok, thank you for confirming that.
For users defining items in file, their rules will be impacted if they don't add the new `defaultUnit` field on item when the label pattern included another unit than the default.
Either they update their rules or they add the `defaultUnit` on the item item and no need to update the rules.

> * When `%unit%` is set, the migration tool prints a warning that you should consider setting `unit` metadata. `%unit` displays the internal state, i.e. the default unit.

Then it is a change ! `%unit%` was to display in the unit the item was updated (by a binding or a user command).
It has to be documented and usage of %unit% in label pattern should be discouraged (as you did with your warning the tool).

> * The list is here: https://www.openhab.org/docs/concepts/units-of-measurement.html. But it is incomplete. I'll prepare a PR for that.

In addition, we should add explanations to what to do for users having items in config files.

rkoshak:
> Where would we add documentation for manual operation of the upgrade tool? This is also missing, but needed.

I think an "Upgrade" section could be added to https://www.openhab.org/docs/installation/ unless there are differences between the various OS's in how the tool works in which case a section in each OS page. I could also see a new upgrade page being added to the docs under installation. 

With the upgrade tool, it's not clear if we will need an official migration tutorial this time. We probably should have had one for OH 3 but it never got written. 

 > In addition, we should add explanations to what to do for users having items in config files.

Agreed. I don't imagine that the upgrade tool will be able to do much for text based config users anyway so what they need to do manually needs to be documented somewhere.

J-N-K:
Please see https://github.com/openhab/openhab-docs/pull/2072 and https://github.com/openhab/openhab-docs/pull/2073

J-N-K:
@kaikreuzer For better user-experience I would suggest to merge  https://github.com/openhab/openhab-distro/pull/1528 and the two docs PR before and https://github.com/openhab/openhab-distro/pull/1529 together with this one.

kaikreuzer:
Yes, absolutely - thanks for working on all of these!

lolodomo:
Regarding adjustments to apply to config files for items, we need to emphase this for users migrating to OH4.
@J-N-K already updated the UoM page in documentation so no need to explain the concepts. We just need short instructions to either use the upgrade tool for managed items or to provide the clear actions to do for unmanaged items.
Where can we provide that ? Is it in the release notes ?
@kaikreuzer : I believe it should also be included in your message announcing the new milestone (and later the new OH4) to be sure no one will loose these informations.

WDYT about "For your items managed with Main UI, it is recommended to use the upgrade tool with the command parameter `itemCopyUnitToMetadata`. For your items defined in configuration files, and in particular for those that are persisted or used in rules, it is recommended to add the new `unit` metadata in case you are not using the default unit in state pattern." ?

J-N-K:
This is already part of https://github.com/openhab/openhab-distro/pull/1529.

lolodomo:
> This is already part of [openhab/openhab-distro#1529](https://github.com/openhab/openhab-distro/pull/1529).

That is good but unfortunately you can expect 90% of users not reading these pages while I expect 98% of users reading @kaikreuzer annoucement.

kaikreuzer:
> Where can we provide that ? Is it in the release notes ?

Imho, it should definitely be explained in some detail in the "breaking changes" section of the release notes.
It might also worth to spend it a section in the release announcement blog post, so that everybody is aware of it.

Btw, the blog post is not "mine" - if any of you would be interested to compile it for OH4, it would be awesome. 😎

J-N-K:
Do we wait for something from me here before we can merge?

lolodomo:
> Imho, it should definitely be explained in some detail in the "breaking changes" section of the release notes.
> It might also worth to spend it a section in the release announcement blog post, so that everybody is aware of it.

Fine. 
And as this will not be available for milestone 3, please mention something in your M3 announcement for early adopters.

kaikreuzer:
> Do we wait for something from me here before we can merge?

Maybe for the [upgrade tool docs to appear](https://github.com/openhab/openhab-docs/pull/2073#issuecomment-1537338394)?

J-N-K:
https://github.com/openhab/openhab-docs/pull/2075

lolodomo:
@J-N-K : do we have now a method on the `Item` to get its internal unit (coming either from the unit metadata if defined or from the default unit) ?
@mherwege  for information

spacemanspiff2007:
Shouldn't the default unit be only used for visualization purposes?
I was under the impression that during item creation the metadata `unit` will always be created (depending on the system locale).

lolodomo:
> > Do we wait for something from me here before we can merge?
> 
> Maybe for the [upgrade tool docs to appear](https://github.com/openhab/openhab-docs/pull/2073#issuecomment-1537338394)?

Looks like the page with the upgrade tool is now published. 

Note that it would be more reasonable to have this change in a snapshot before it is included in a milestone so that we can make few tests including using the new upgrade tool.

J-N-K:
@kaikreuzer Documentation is online.

J-N-K:
> @J-N-K : do we have now a method on the `Item` to get its internal unit (coming either from the unit metadata if defined or from the default unit) ?
> @mherwege for information

If you call `.getUnit` or `.getUnitSymbol` on the item and get what the item currently uses internally. This of course can change if add/update/remove the `unit` metadata.

lolodomo:
> > @J-N-K : do we have now a method on the `Item` to get its internal unit (coming either from the unit metadata if defined or from the default unit) ?
> > @mherwege for information
> 
> If you call `.getUnit` or `.getUnitSymbol` on the item and get what the item currently uses internally. This of course can change if add/update/remove the `unit` metadata.

Excellent, thank you @J-N-K . Certainly what you should use @mherwege .

jlaur:
Is this change included in build 3457?

lolodomo:
> Is this change included in build 3457?

No because the build of core framework failed this morning. But will be in next snapshot successfully built.

jlaur:
> No because the build of core framework failed this morning. But will be in next snapshot successfully built.

Thanks. That explains why my tests were very disappointing. 😄

rkoshak:
https://ci.openhab.org/ shows there was a successful build 11 hours ago. Is the failure not showing up there?

lolodomo:
There was a successful build but not with the last core changes and not this PR in particular,

lolodomo:
@kaikreuzer : I believe that the snapshot 3458 does not include webui repo built with the most recent core framework (and so without this new UoM stuff).

lolodomo:
I just tried with few Number items defined in a config file, covering all the possible cases I can imagine, and it works as claimed.
I also tried with managed items, with or without the "unit" metadata, and it works.

Main UI has to be enhanced to propose the new "unit" metadata in that pop-up:
image
I used "custom namespace" to fill it.

The only problem I encountered (and it is major for users using managed items), the upgrade tool is doing apparently nothing in my case. No error in the console when I run it but no metadata added. It was run in a Windows Terminal window.

jlaur:
@lolodomo - my first tests also looks promising. I tried also to differentiate display unit from persisted unit:
```javascript
Number:Temperature TestTemperature "Temperature [%.10f °C]" <temperature> { unit="°F" }
```

with a few updates:
```shell
openhab:send TestTemperature "10 °C"
openhab:send TestTemperature "10 °F"
```

All looks good. If I change state description to "%d °C" there are some rounding issues, but that's expected when internally storing a different unit. For example, when updated to 10 °C, stored in °F and displayed in °C, then it will be  10.00000000000000000000000000000002 °C.

I'm using this to find items I have not yet applied unit to:
```shell
grep "Number:" items/*.items | grep -v "unit="
```

lolodomo:
@jlaur : did you succeed using the upgrade tool?

jlaur:
> did you succeed using the upgrade tool?

I have not tried that yet. But I plan to add unit metadata to my items now on my 3.4 production system so that I'm ready for upgrading to 4.0. I guess the upgrade tool is to be used during or after the upgrade?

J-N-K:
The upgrade tool only works for managed items, because it's difficult to fix these manually.

lolodomo:
It has to be run before starting a OH version containing the new UoM code if you have managed items.

J-N-K:
It'll will be called during the update process from the update script. We need a minor modification before that works.

jlaur:
So for users having unmanaged items, it's very important to update them before starting an upgrade. Which means that the distro breaking notes are too late since they are shown during the upgrade. I guess it has all been mentioned, but we need to make sure this information is made very visible in the announcement etc. to increase the chance of users actually reacting before starting the upgrade.

lolodomo:
> So for users having unmanaged persisted items, it's very important to update them before starting an upgrade. Which means that the distro breaking notes are too late since they are shown during the upgrade. I guess it has all been mentioned, but we need to make sure this information is made very visible in the announcement etc. to increase the chance of users actually reacting before starting the upgrade.

Yes, it was discussed with @kaikreuzer in this topic, he will certainly mention that in his release announcement (even in the milestone announcement).
For unmanaged persisted items, it is very important to apply manual changes before starting the new OH version.
I did it yesterday, in practice, I added the unit metadata only on few items because I was using the default unit for most of my items. I also replaced all my usages of `%unit%` by a real unit, even if probably not required.

What is important to note is that in the future we should really avoid changing default units.

jlaur:
> I also replaced all my usages of `%unit%` by a real unit, even if probably not required.

I'm wondering if this is redundant now when you want to display the same unit as stored, so I would even consider going the other way, i.e. from:

```javascript
Number:Temperature TestTemperature "Temperature [%.0f °C]" <temperature>
```

to:

```javascript
Number:Temperature TestTemperature "Temperature [%.0f %unit%]" <temperature> { unit="°C" }
```

since I guess `%unit%` should now automatically pick the defined unit?

J-N-K:
@jlaur correct

lolodomo:
Yes, that looks like an even better option, I may do the same thing, even if it means more changes to my items file.
The advantage to fix the internal unit is that even if the default unit is changed in the future, you will not be impacted.

lolodomo:
> to:
> 
> ```js
> Number:Temperature TestTemperature "Temperature [%.0f %unit%]" <temperature> { unit="°C" }
> ```

@jlaur : take care to not run an OH server with the old UoM with such config as usign `%unit%` in state description was a problem for persisted items (for restore on startup if I correctly remember).
At least, this PR fixed that problem ;)

lolodomo:
> At least, this PR fixed that problem ;)

That let me think we have certainly several issues to close ;)

lolodomo:
> I was checking Basic UI JavaScript code and its handling when a new item state arrives, and I am almost certain that a different unit between internal representation and state description could break the behaviour of the setpoint and slider widgets !
> I need to confirm that with a real test case.

Good news, the sitemap SSE event contains two values, the value in display unit and the value in the internal unit when they are different. Basic UI is using in priority the value in display unit. As consequence, the setpoint widget is finally working as expected.
I still have to check the slider widget.

Note that I discovered a new problem with ERROR logged form core framework when the label pattern does not contain unit but only the value ("Setpoint Temp [%f]"). I don't know if this is something new with this PR or not:
```
11:05:28.874 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'SetpointTemp' changed from NULL to 19 °C
11:05:28.889 [ERROR] [org.openhab.core.types.util.UnitUtils] - Unknown unit from pattern: %f
11:05:28.893 [ERROR] [org.openhab.core.types.util.UnitUtils] - Unknown unit from pattern: %f
11:05:28.894 [ERROR] [org.openhab.core.types.util.UnitUtils] - Unknown unit from pattern: %f
11:05:28.895 [ERROR] [org.openhab.core.types.util.UnitUtils] - Unknown unit from pattern: %f
11:05:28.896 [ERROR] [org.openhab.core.types.util.UnitUtils] - Unknown unit from pattern: %f
11:05:28.896 [ERROR] [org.openhab.core.types.util.UnitUtils] - Unknown unit from pattern: %f
```
It is logged by method `parseUnit` from class `UnitUtils` used in two places in `ItemUIRegistryImpl`. Either there is a bug or the log level is inappropriate.

J-N-K:
This seems to be the same issue as reported by @rkoshak on the forum. Probably I increased the log level, I'll check that.

lolodomo:
> This seems to be the same issue as reported by @rkoshak on the forum. Probably I increased the log level, I'll check that.

Yes, you changed the log level in this PR. It was at DEBUG level before.

lolodomo:
@J-N-K : when no space is found in the pattern, shouldn't we directly return null rather than continuing with the label pattern as unit symbol ?
```
        String unitSymbol = pattern;
        int lastBlankIndex = pattern.lastIndexOf(" ");
        if (lastBlankIndex >= 0) {
            unitSymbol = pattern.substring(lastBlankIndex).trim();
        }
```

J-N-K:
Probably yes.

lolodomo:
> Good news, the sitemap SSE event contains two values, the value in display unit and the value in the internal unit when they are different. Basic UI is using in priority the value in display unit. As consequence, the setpoint widget is finally working as expected.
> I still have to check the slider widget.

Second good news: the slider widget in Basic UI is not impacted too.

maniac103:
> please also check that the separation of unit between internal representation (item state) and display representation (using unit from state pattern) has no wrong impact on Main UI, Android app and iOS app.

@lolodomo Can you please share your test case?

lolodomo:
> > please also check that the separation of unit between internal representation (item state) and display representation (using unit from state pattern) has no wrong impact on Main UI, Android app and iOS app.
> 
> @lolodomo Can you please share your test case?

You can try with these items
```
Number:Temperature CurrentTemp "Current Temp [%.1f °C]" <none> (GTest1,GTest) [ "Measurement", "Temperature" ]
Number:Length TestNumberDim1 "Test number dim 1 [%.1f cm]" (GTest1) { unit="mm" }
Number:Length TestNumberDim2 "Test number dim 2 [%.1f cm]" (GTest1)
```
and this in sitemap:
```
	Frame label="Text and unit" {
		Text item=SetpointTemp label="Setpoint Temp [%f °C]"
		Text item=SetpointTemp label="Setpoint Temp [%f %unit%]"
		Text item=SetpointTemp label="Setpoint Temp [%f °F]"
		Text item=SetpointTemp label="Setpoint Temp [%f]"
		Text item=SetpointTemp label="Setpoint Temp []"
		Text item=TestNumberDim1 label="Test number dim 1 [%f mm]"
		Text item=TestNumberDim1 label="Test number dim 1 [%f %unit%]"
		Text item=TestNumberDim1 label="Test number dim 1 [%f cm]"
		Text item=TestNumberDim1 label="Test number dim 1 [%f]"
		Text item=TestNumberDim1 label="Test number dim 1 []"
		Text item=TestNumberDim2 label="Test number dim 2 [%f mm]"
		Text item=TestNumberDim2 label="Test number dim 2 [%f %unit%]"
		Text item=TestNumberDim2 label="Test number dim 2 [%f cm]"
		Text item=TestNumberDim2 label="Test number dim 2 [%f]"
		Text item=TestNumberDim2 label="Test number dim 2 []"
	}
	Frame label="Setpoint and unit" {
		Default item=SetpointTemp label="Setpoint Temp [%f °C]"
		Default item=SetpointTemp label="Setpoint Temp [%f %unit%]"
		Default item=SetpointTemp label="Setpoint Temp [%f °F]"
		Default item=SetpointTemp label="Setpoint Temp [%f]"
		Default item=SetpointTemp label="Setpoint Temp []"
	}
	Frame label="Slider and unit" {
		Slider item=TestNumberDim1 label="Test number dim 1 [%f mm]" minValue=0 maxValue=100 step=5
		Slider item=TestNumberDim1 label="Test number dim 1 [%f %unit%]" minValue=0 maxValue=100 step=5
		Slider item=TestNumberDim1 label="Test number dim 1 [%f cm]" minValue=0 maxValue=10 step=0.5
		Slider item=TestNumberDim1 label="Test number dim 1 [%f]" minValue=0 maxValue=100 step=5
		Slider item=TestNumberDim1 label="Test number dim 1 []" minValue=0 maxValue=100 step=5
		Slider item=TestNumberDim2 label="Test number dim 2 [%f mm]" minValue=0 maxValue=1000 step=100
		Slider item=TestNumberDim2 label="Test number dim 2 [%f %unit%]" minValue=0 maxValue=1 step=0.1
		Slider item=TestNumberDim2 label="Test number dim 2 [%f cm]" minValue=0 maxValue=100 step=10
		Slider item=TestNumberDim2 label="Test number dim 2 [%f]" minValue=0 maxValue=1 step=0.1
		Slider item=TestNumberDim2 label="Test number dim 2 []" minValue=0 maxValue=1 step=0.1
	}
```

lolodomo:
I just tested again the upgrade tool on a very simple case and it worked as expected, meaning unit metadata was added on the expected item.
@J-N-K : maybe it would better to add an INFO log for each item updated ?

jlaur:
I have one more question/uncertainty from the testing of openhab/openhab-addons#15002. The binding provides frequency in **rpm** and system default unit is **Hz**. I had to change state description from "rpm" to "%unit%" because otherwise it would show for example 74580 rpm while actually being 1243 rpm. This is because it was considered 1243 Hz.

My question is: The user will have to manually set unit to "rpm" when linking an item, since there is no way to "enforce" this from the binding, right? Probably my doubt is just because both Hz and rpm are commonly used, but for different purposes.

J-N-K:
Is the provided value a `DecimalType` or a `QuantityType`? If it is a `QuantityType`, it should be fine with `rpm`, if not, then the fault unit `Hz` is added and the converted to `rpm` for displaying.

jlaur:
It's provided as `QuantityType`. I'll test again, it's probably some corner case.

jlaur:
Okay, at least reproducible. After thing is freshly created and new item created and linked:
```
2023-05-18 14:14:27.131 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'danfossairunit:airunit:-1062731542' changed from UNKNOWN to OFFLINE (COMMUNICATION_ERROR): Connection timed out: connect
2023-05-18 14:14:30.707 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Danfoss_Air_Unit_Indblaesning_blaeserhastighed' changed from NULL to 1251 Hz
2023-05-18 14:14:30.707 [INFO ] [openhab.event.ItemUpdatedEvent      ] - Item 'Danfoss_Air_Unit_Indblaesning_blaeserhastighed' has been updated.
```

image

The binding has received no state updates (as being offline), but the previous value must exist somewhere, just without unit.

J-N-K:
Maybe it is the result of a REFRESH command? Does that properly set the value?

lolodomo:
I discovered this warning. The message is really not clear at all, I do not understand the last part:
```
09:02:09.622 [WARN ] [.ui.internal.items.ItemUIRegistryImpl] - Exception while formatting value '88.31899999999999 kWh' of item ConsoSemaineDerniere with format '%.0f kWh [%.3f %unit%': Format specifier '%.3f'
```
A "]" is missing in the message. Where is coming the `%.3f` from ?

Here is how is defined my item:
```
Number:Energy ConsoSemaineDerniere "Conso semaine dernière [%.0f kWh" <energy> (GLinky) [ "Measurement", "Energy" ] { channel="linky:linky:local:weekly#lastWeek", listWidget="" [action=""] }
```

lolodomo:
I just discovered that I lost dynamic icon when my item is Number:Dimensionless (humidities or battery levels).
Here is an example of item no more working:
```
Number:Dimensionless UpsChargeBatterie "Batterie onduleur [%d %%]" <battery> (GOnduleur) [ "Measurement", "Energy" ] { channel="networkupstools:ups:eaton:batteryCharge", listWidget="" [iconUseState=true, action=""] }
```
I am sure it was working before the UoM change.
What could be wrong ?

J-N-K:
Missing metadata? Dimensionless has default unit ONE, not %.

lolodomo:
> Missing metadata? Dimensionless has default unit ONE, not %.

But as there is a state description, shouldn't it work ?
It does not work neither in MainUI nor in BasicUI/Android app.

J-N-K:
No, the state description is no longer considered, that is the whole point of the refactoring and what the upgrade tool fixes for managed items.

lolodomo:
But for UI, the state description is used ???

I can see that 1 (instead of 100) is sued when requesting the icon:
image

How can fix my item definition to have the proper behaviour ?

lolodomo:
Should I add `unit="%"` ?

lolodomo:
That would mean that item state passed to icon request is the internal state. That is probable.

J-N-K:
What version did you use for that test?

maniac103:
Snapshot 3389.

J-N-K:
That's very old (the last snapshot is 3479) and I believe way before this PR was merged, can you please try with a recent snapshot?

maniac103:
Just did just that, but AFAICT, the sitemap response is the same as with 3389:

```
                    {
                        "widgetId": "0200",
                        "type": "Slider",
                        "visibility": true,
                        "label": "Test number dim 1 [60,000000 mm]",
                        "icon": "number",
                        "mappings": [

                        ],
                        "switchSupport": false,
                        "sendFrequency": 0,
                        "minValue": 0,
                        "maxValue": 100,
                        "step": 5,
                        "state": "60",
                        "item": {
                            "link": "http://127.0.0.1:8080/rest/items/TestNumberDim1",
                            "state": "6 cm",
                            "stateDescription": {
                                "pattern": "%.1f cm",
                                "readOnly": false,
                                "options": [

                                ]
                            },
                            "type": "Number:Length",
                            "name": "TestNumberDim1",
                            "label": "Test number dim 1",
                            "tags": [

                            ],
                            "groupNames": [

                            ]
                        },
                        "widgets": [

                        ]
                    },
```

... so the points from above (particularly the missing widget format string) still apply.

J-N-K:
IMO the response is fully valid, the state is `6 cm` because there is no `unit` metadata and the default for length is `cm`. The sitemap instructs openHAB to use `mm` 
```
Text item=TestNumberDim1 label="Test number dim 1 [%f mm]"
```

(and that overrides the state description, which is also correct). Also the label is formatted accordingly, as well as the state. So I guess the only thing that is missing is the unit. 

@lolodomo WDYT?

maniac103:
Well, the state is `6 cm`, yes (as visible via item -> state), but
- How does the client know `maxValue=100` means `mm`, not `cm`?
- How does the client know `state=60` means `mm`, not `cm`?
- How is the client know the widget format string (`%f mm`)?

As I said, it's all there, it just needs to be put in the sitemap widget response (and SSE event, in case of the state) ;-)
IMO, it would be ideal to add an (optional) `unit` field (which would contain `mm` here) and another (optional) `formatPattern` field (which would contain `%f mm` here).

lolodomo:
Default unit for length is meters, not centimeters.

maniac103:
> I am a bit puzzled how the result above could be produced by recent code.

'Result above' as in: mine? I'll retry with a new install (as opposed to unpacking the zip over the old install) tomorrow.
Regarding the `metadata` part of your sitemap response: I assume the `stateDescription` part automatically derives from that if no other state description is given? 

> This should be discussed in a new issue.

Will open a new open for that.

lolodomo:
Regarding min/max/step values, there is am open issue to add unit. Currently, the default unit for these attributes is the display unit if one is defined. 
For the state field in the sitemap event, it is the value in the display unit. For slider widget, the unit is stripped IIRC
The label field contains the value formaated according to state description.

lolodomo:
I have not read everything in ultra details but I see nothing wrong in the events.
 
I agree that it could be better to explain the units for min/max(step or even to be able to provide an unit.

ssalonen:
This seems to have broken profile that converts unit-aware-number commands to plain numbers

https://community.openhab.org/t/modbus-thermostat-stepper-dont-work/149836/20?u=ssalonen

PR #3486, opened 2023-03-26
Bump commons-fileupload

J-N-K:
It is not needed in openhab-core or openhab-addons.

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -392,7 +392,7 @@
     <dependency>
       <groupId>commons-fileupload</groupId>
       <artifactId>commons-fileupload</artifactId>
-      <version>1.4</version>
+      <version>1.5</version>
       <scope>compile</scope>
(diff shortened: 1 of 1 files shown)


wborn:
I think it's needed for being able to resolve the deps of the demo app in the distro, see: https://github.com/openhab/openhab-core/pull/1458

J-N-K:
If it's only needed there, why don't we add it to the POM of the demo app?

wborn:
Historically this BOM has been used for both itest and demo app deps. That was probably done because those runtime environments are very similar and it's easier to manage these deps in place. It contains more deps that are only used by the Demo app such as all those "Web Console" and Gogo shell deps. So if the commons-fileupload dep is moved those should go as well. :wink:

J-N-K:
I reverted and decided to only bump to 1.5 instead

kaikreuzer (approved):

PR #3488, opened 2023-03-26
Fix OOBE in item metadata descriptions providers

GiviMAD:
The previous PR #3363 introduces possible out of bound exception on the item metadata descriptions provider. Sorry for the inconvenience.


GiviMAD:
I have changed the condition a little and moved the duplicated code into a function, hope that is ok.
Also I've updated the tests for both metadata services to cover this case.

wborn (approved):
Thanks for the fix, cleanup and test!

PR #3491, opened 2023-03-28
Sort profile-type resources by their label

jimtng:
Sort the profile type json result from the REST service. This makes the UI a bit more predictable. Helpful for repetitive operations so your eyes don't go around hunting for the wanted item.

Current:
image

Sorted:
image


J-N-K (approved):
LGTM, thanks.

Issue #3478, opened 2023-01-07
Energy management/calculations


jlaur:

jlaur:
@kaikreuzer - great to see some concrete proposals for the architecture, thanks! Since I have mostly been working within the price (including calculations) domain so far, I would be interested into digging a bit more into the `EnergyProvider` interface. My only power source is the grid, so more specifically `getPrice(interval)`.

Do you imagine `EnergyManagementService` requesting all prices within a given interval, so that it can perform calculations, such as the cost of a given consumption within a given period, or do you expect each `EnergyProvider` implementation to perform such calculations? Having different sources of prices (bindings implementing `EnergyProvider`) with the same "calculation engine" seems desirable to avoid duplication? In this case the provider would either need to provide a sum of the price elements for the available intervals, or provide the elements themselves.

In the openhab/openhab-addons#14376 I decided to provide the elements individually and let the user create a `Group` item or similar for calculating the sum: [channels](https://github.com/openhab/openhab-addons/blob/2581fc3855dfd3743355262885cecf1ea3c4cf6f/bundles/org.openhab.binding.energidataservice/README.md#channel-group-electricity). This provides a certain level of flexibility as a base price can be added without the binding needing to support this, or elements could even come from other external sources. For example, one binding could provide the spot price and another one the relevant tariffs for a specific grid company or area.

So I'm thinking, if the `EnergyManagementService` should perform the calculations, some degree of configuration is probably needed for picking relevant sources. Perhaps each implementation could expose which kind of elements it supports (e.g. `energidataservice:spot-price`), so that this could be requested specifically when asked to provide prices for a given interval.

Calculations can be done without VAT internally to find lowest relative price etc. or the engine could support transformations, e.g. openhab/openhab-addons#14529

A few additional unstructured thoughts for which I currently have no idea where they fit in (or if they do at all).
- Energy is not only electricity, it could also be gas. Energi Data Service also has datasets for that.
- In Denmark you have to pay net tariffs when delivering electricity (i.e. from PV) to the grid. When spot prices are low, you might have to pay more than you receive for delivering. From a pure economical perspective, this could incentivize to use excess power by any means.
- Energi Data Service also provides datasets related to CO2 emissions and how "green" (renewable) the electricity will be at given hours. This does not necessarily map directly to the prices. So if one would prioritize Planet Earth over price, this data could be a driver for decision-making rather than prices.

kaikreuzer:
[...]

PR #3503, opened 2023-03-28
[UoM] Add currency handling 

### LINKED ISSUE #3408: Currency type UoM
I noticed there seems not to be any UoM type to reflect currencies or "money" the dimension is.
With dynamic power tariffs and such raising, these should be added fast before we have binding devs and users embark on any type of workaround in rules.

Related: #983
Related: #3282

J-N-K:
Closes #3408 
Related to #3478 

This enables currency handling for UoM. You can define prices (e.g. `0.336 €/kWh`) and do calculations based on that like any other unit.

Currently you can define a "system currency" (e.g. `€`, defaults zu `@` if not set at all) which is then available as any other unit. It is also possible to define a `CurrencyProvider` which adds additional currencies and their exchange rates (relative to the system currency). Switching between different currency providers is not yet implemented.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/i18n/I18nProviderImpl.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/i18n/I18nProviderImpl.java
@@ -14,7 +14,6 @@ package org.openhab.core.internal.i18n;
 
 import static org.openhab.core.library.unit.MetricPrefix.HECTO;
 
-import java.math.BigDecimal;
 import java.text.MessageFormat;
 import java.time.DateTimeException;
 import java.time.ZoneId;
@@ -25,7 +24,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.ResourceBundle;
 import java.util.Set;
-import java.util.function.Function;
 
 import javax.measure.Quantity;
 import javax.measure.Unit;
@@ -82,8 +80,6 @@ import org.openhab.core.library.dimension.Intensity;
 import org.openhab.core.library.dimension.RadiationSpecificActivity;
 import org.openhab.core.library.dimension.VolumetricFlowRate;
 import org.openhab.core.library.types.PointType;
-import org.openhab.core.library.unit.CurrencyProvider;
-import org.openhab.core.library.unit.CurrencyUnit;
 import org.openhab.core.library.unit.CurrencyUnits;
 import org.openhab.core.library.unit.ImperialUnits;
 import org.openhab.core.library.unit.SIUnits;
@@ -124,8 +120,8 @@ import org.slf4j.LoggerFactory;
         "service.config.category=system", //
         "service.config.description.uri=system:i18n" })
 @NonNullByDefault
-public class I18nProviderImpl implements TranslationProvider, LocaleProvider, LocationProvider, TimeZoneProvider,
-        UnitProvider, CurrencyProvider {
+public class I18nProviderImpl
+        implements TranslationProvider, LocaleProvider, LocationProvider, TimeZoneProvider, UnitProvider {
 
     private final Logger logger = LoggerFactory.getLogger(I18nProviderImpl.class);
 
@@ -133,7 +129,6 @@ public class I18nProviderImpl implements TranslationProvider, LocaleProvider, Lo
 
     // LocaleProvider
     public static final String LANGUAGE = "language";
-    public static final String BASE_CURRENCY = "baseCurrency";
     public static final String SCRIPT = "script";
     public static final String REGION = "region";
     public static final String VARIANT = "variant";
@@ -174,7 +169,6 @@ public class I18nProviderImpl implements TranslationProvider, LocaleProvider, Lo
     @Modified
     protected synchronized void modified(Map<String, Object> config) {
         final String language = toStringOrNull(config.get(LANGUAGE));
-        this.currencyCode = toStringOrNull(config.get(BASE_CURRENCY));
         final String script = toStringOrNull(config.get(SCRIPT));
         final String region = toStringOrNull(config.get(REGION));
         final String variant = toStringOrNull(config.get(VARIANT));
@@ -456,30 +450,5 @@ public class I18nProviderImpl implements TranslationProvider, LocaleProvider, Lo
             Map<Class<? extends Quantity<?>>, Map<SystemOfUnits, Unit<? extends Quantity<?>>>> dimensionMap,
             Class<T> dimension, Unit<T> unit) {
         dimensionMap.put(dimension, Map.of(SIUnits.getInstance(), unit, ImperialUnits.getInstance(), unit));
-    }
-
-    @Override
-    public Unit<Currency> getBaseCurrency() {
-        String currencyCode = this.currencyCode;
-        if (currencyCode == null && locale != null) {
-            currencyCode = java.util.Currency.getInstance(locale).getCurrencyCode();
-        }
-        if (currencyCode != null) {
-            // either the currency was set or determined from the locale
-            String symbol = java.util.Currency.getInstance(currencyCode).getSymbol();
-            return new CurrencyUnit(currencyCode, symbol);
-        } else {
-            return new CurrencyUnit("DEF", null);
-        }
-    }
-
-    @Override
-    public Collection<Unit<Currency>> getCurrencies() {
-        return Set.of();
-    }
-
-    @Override
-    public Function<Unit<Currency>, @Nullable BigDecimal> getExchangeRateFunction() {
-        return unit -> null;
     }
 }
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/i18n/I18nProviderImpl.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/i18n/I18nProviderImpl.java
@@ -14,6 +14,7 @@ package org.openhab.core.internal.i18n;
 
 import static org.openhab.core.library.unit.MetricPrefix.HECTO;
 
+import java.math.BigDecimal;
 import java.text.MessageFormat;
 import java.time.DateTimeException;
 import java.time.ZoneId;
@@ -24,6 +25,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.ResourceBundle;
 import java.util.Set;
+import java.util.function.Function;
 
 import javax.measure.Quantity;
 import javax.measure.Unit;
@@ -70,14 +72,19 @@ import org.openhab.core.i18n.TimeZoneProvider;
 import org.openhab.core.i18n.TranslationProvider;
 import org.openhab.core.i18n.UnitProvider;
 import org.openhab.core.library.dimension.ArealDensity;
+import org.openhab.core.library.dimension.Currency;
 import org.openhab.core.library.dimension.DataAmount;
 import org.openhab.core.library.dimension.DataTransferRate;
 import org.openhab.core.library.dimension.Density;
 import org.openhab.core.library.dimension.ElectricConductivity;
+import org.openhab.core.library.dimension.EnergyPrice;
 import org.openhab.core.library.dimension.Intensity;
 import org.openhab.core.library.dimension.RadiationSpecificActivity;
 import org.openhab.core.library.dimension.VolumetricFlowRate;
 import org.openhab.core.library.types.PointType;
+import org.openhab.core.library.unit.CurrencyProvider;
+import org.openhab.core.library.unit.CurrencyUnit;
+import org.openhab.core.library.unit.CurrencyUnits;
 import org.openhab.core.library.unit.ImperialUnits;
 import org.openhab.core.library.unit.SIUnits;
 import org.openhab.core.library.unit.Units;
@@ -117,8 +124,8 @@ import org.slf4j.LoggerFactory;
         "service.config.category=system", //
         "service.config.description.uri=system:i18n" })
 @NonNullByDefault
-public class I18nProviderImpl
-        implements TranslationProvider, LocaleProvider, LocationProvider, TimeZoneProvider, UnitProvider {
+public class I18nProviderImpl implements TranslationProvider, LocaleProvider, LocationProvider, TimeZoneProvider,
+        UnitProvider, CurrencyProvider {
 
     private final Logger logger = LoggerFactory.getLogger(I18nProviderImpl.class);
 
@@ -126,10 +133,12 @@ public class I18nProviderImpl
 
     // LocaleProvider
     public static final String LANGUAGE = "language";
+    public static final String BASE_CURRENCY = "baseCurrency";
     public static final String SCRIPT = "script";
     public static final String REGION = "region";
     public static final String VARIANT = "variant";
     private @Nullable Locale locale;
+    private @Nullable String currencyCode;
 
     // TranslationProvider
     private final ResourceBundleTracker resourceBundleTracker;
@@ -165,6 +174,7 @@ public class I18nProviderImpl
     @Modified
     protected synchronized void modified(Map<String, Object> config) {
         final String language = toStringOrNull(config.get(LANGUAGE));
+        this.currencyCode = toStringOrNull(config.get(BASE_CURRENCY));
         final String script = toStringOrNull(config.get(SCRIPT));
         final String region = toStringOrNull(config.get(REGION));
         final String variant = toStringOrNull(config.get(VARIANT));
@@ -395,6 +405,7 @@ public class I18nProviderImpl
         addDefaultUnit(dimensionMap, Area.class, SIUnits.SQUARE_METRE, ImperialUnits.SQUARE_FOOT);
         addDefaultUnit(dimensionMap, ArealDensity.class, Units.DOBSON_UNIT);
         addDefaultUnit(dimensionMap, CatalyticActivity.class, Units.KATAL);
+        addDefaultUnit(dimensionMap, Currency.class, CurrencyUnits.BASE_CURRENCY);
         addDefaultUnit(dimensionMap, DataAmount.class, Units.BYTE);
         addDefaultUnit(dimensionMap, DataTransferRate.class, Units.MEGABIT_PER_SECOND);
         addDefaultUnit(dimensionMap, Density.class, Units.KILOGRAM_PER_CUBICMETRE);
@@ -420,6 +431,7 @@ public class I18nProviderImpl
         addDefaultUnit(dimensionMap, Mass.class, SIUnits.KILOGRAM, ImperialUnits.POUND);
         addDefaultUnit(dimensionMap, Power.class, Units.WATT);
         addDefaultUnit(dimensionMap, Pressure.class, HECTO(SIUnits.PASCAL), ImperialUnits.INCH_OF_MERCURY);
+        addDefaultUnit(dimensionMap, EnergyPrice.class, CurrencyUnits.BASE_ENERGY_PRICE);
         addDefaultUnit(dimensionMap, RadiationDoseAbsorbed.class, Units.GRAY);
         addDefaultUnit(dimensionMap, RadiationDoseEffective.class, Units.SIEVERT);
         addDefaultUnit(dimensionMap, Radioactivity.class, Units.BECQUEREL);
@@ -444,5 +456,30 @@ public class I18nProviderImpl
             Map<Class<? extends Quantity<?>>, Map<SystemOfUnits, Unit<? extends Quantity<?>>>> dimensionMap,
             Class<T> dimension, Unit<T> unit) {
         dimensionMap.put(dimension, Map.of(SIUnits.getInstance(), unit, ImperialUnits.getInstance(), unit));
+    }
+
+    @Override
+    public Unit<Currency> getBaseCurrency() {
+        String currencyCode = this.currencyCode;
+        if (currencyCode == null && locale != null) {
+            currencyCode = java.util.Currency.getInstance(locale).getCurrencyCode();
+        }
+        if (currencyCode != null) {
+            // either the currency was set or determined from the locale
+            String symbol = java.util.Currency.getInstance(currencyCode).getSymbol();
+            return new CurrencyUnit(currencyCode, symbol);
+        } else {
+            return new CurrencyUnit("DEF", null);
+        }
+    }
+
+    @Override
+    public Collection<Unit<Currency>> getCurrencies() {
+        return Set.of();
+    }
+
+    @Override
+    public Function<Unit<Currency>, @Nullable BigDecimal> getExchangeRateFunction() {
+        return unit -> null;
     }
 }
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/library/unit/CurrencyService.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/library/unit/CurrencyService.java
@@ -83,7 +83,6 @@ public class CurrencyService {
 
     @Reference(cardinality = ReferenceCardinality.MULTIPLE, policy = ReferencePolicy.DYNAMIC)
     public void addCurrencyProvider(CurrencyProvider currencyProvider) {
-        logger.error("Added {}", currencyProvider.getName());
         currencyProviders.put(currencyProvider.getName(), currencyProvider);
         if (configuredCurrencyProvider.equals(currencyProvider.getName())) {
             enableProvider(currencyProvider);
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/library/unit/CurrencyService.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/library/unit/CurrencyService.java
@@ -1,0 +1,86 @@
+/**
+ * Copyright (c) 2010-2023 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.internal.library.unit;
+
+import static org.openhab.core.library.unit.CurrencyUnits.BASE_CURRENCY;
+
+import java.math.BigDecimal;
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.function.Function;
+
+import javax.measure.Unit;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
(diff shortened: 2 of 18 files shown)


bundles/org.openhab.core/src/main/java/org/openhab/core/internal/i18n/I18nProviderImpl.java
```diff
@@ -430,4 +439,19 @@ private <T extends Quantity<T>> void addDefaultUnit(Class<T> dimension, Unit<T>
     private <T extends Quantity<T>> void addDefaultUnit(Class<T> dimension, Unit<T> unit) {
         dimensionMap.put(dimension, Map.of(SIUnits.getInstance(), unit, ImperialUnits.getInstance(), unit));
     }
+
+    @Override
+    public Unit<Currency> getSystemCurrency() {
+        return new CurrencyUnit("@", "Default Currency");
```
jlaur, line 445: Can we default the currency based on configured `Locale`? Example:

```
String currencyCode = Currency.getInstance(LocaleProvider.getLocale()).getCurrencyCode();
String currencySymbol = Currency.getInstance(LocaleProvider.getLocale()).getSymbol();
```

Also, I'm wondering if we should/can encapsulate the Java [Currency](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Currency.html) rather than reinventing too much. This would give us both the currency code (EUR) and the symbol (€) (if any) for display purposes.

J-N-K: locale based default: maybe, that‘s why it is still WIP and one of the thing I will look into.

Encapsulate `Currency`: difficult. I already tried that some time ago, but failed. I think we don‘t need to duplicate that much, we can use a lot of what is already there for UoM.

bundles/org.openhab.core/src/test/java/org/openhab/core/library/unit/CurrencyUnitTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+        assertThat(price, is(notNullValue()));
+        assertThat(price.getUnit(), is(TestCurrencyProvider.EURO));
+        assertThat(price.doubleValue(), closeTo(1.25, 1E-4));
+    }
+
+    private static class TestCurrencyProvider implements CurrencyProvider {
+        public static final Unit<Currency> EURO = new CurrencyUnit("€", "Euro");
+        public static final Unit<Currency> DKK = new CurrencyUnit("DKK", "Danske Kroner");
+        public static final Unit<Currency> USD = new CurrencyUnit("$", "US Dollar");
```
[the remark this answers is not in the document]
J-N-K: True, it‘s up to the currency provider to handle that in a good and understandable way. $ could also be Australian or Canadian dollar.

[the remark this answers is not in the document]
jlaur: Ah, this is **name** and **symbol** in `CurrencyUnit`?

bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/CurrencyUnit.java
```diff
@@ -0,0 +1,287 @@
...
+ * @author Jan N. Klug - Initial contribution
+ */
+@NonNullByDefault({ PARAMETER, RETURN_TYPE, FIELD, TYPE_BOUND })
+public final class CurrencyUnit implements Unit<Currency>, Comparable<Unit<Currency>>, PrefixOperator<Currency>,
+        Nameable, Serializable, SymbolSupplier {
+
+    private static final long serialVersionUID = -1L;
+    private static final Dimension DIMENSION = UnitDimension.parse('$');
+    private String name;
+    private String symbol;
```
jlaur, line 69: I assume name is "Euro" and symbol is "€", as an example? Should we also add currency code ("EUR"), which seems more important than the symbol? And probably symbol should be optional since I believe not all currencies have one.

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/i18n/I18nProviderImpl.java
jlaur: Sure, just giving some early feedback for possibly triggering ideas. I assume that's the purpose of draft/WIP. 🙂

J-N-K:
Maybe we should make three parameters for the `CurrencyUnit`? ISO-Code ("DKK"), symbol ("Kr") and name ("Danske Kroner")?

The problem is that especially the symbols are not necessarily unique, so it may be difficult to determine the correct currency if there is something like `15 $`.

jlaur:
> The problem is that especially the symbols are not necessarily unique, so it may be difficult to determine the correct currency if there is something like `15 $`.
> 
> Edit: I think we can omit the name and use the `name` field for the three letter ISO code. We can also validate that when instantiating a new unit.

Yes, the unique key should definitely be the [currency code](https://en.wikipedia.org/wiki/ISO_4217).

The [symbol](https://en.wikipedia.org/wiki/Currency_symbol) is probably subject to some kind of localization, since it can be placed before or after the amount, for example "$ 15" in the US, and "15 €" in Europe. And the currency name (if included) should support I18N (e.g. "den danske krone" in Danish or "Danish krone" in English).

On top of that, this could be formatted also as the sub unit (1/100) "øre", e.g. "1 krone og 25 øre", although this would usually be written "1,25 kr.". Just listing everything I can think of now. 🙂 But again, the most important start is to use the ISO currency code.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/CurrencyProvider.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * The {@link CurrencyProvider} is a
+ *
+ * @author Jan N. Klug - Initial contribution
+ */
+@NonNullByDefault
+public interface CurrencyProvider {
+
+    Unit<Currency> getSystemCurrency();
```
splatch, line 33: System currency may not always apply as from exchange/conversion point of view you always get pair of currencies. Separating `CurrencyProvider` into i18n/l10n specific and `ExchangeProvider` / `CurrencyConversionProvider` will make it clear.

splatch (approved):
Small note on naming. `CurrencyProvider` suggest that it is something to declare currencies while it seems to provide also exchange rate. The scale factor might also be a bit unfortunate term, cause it might imply a more static nature than it actually is.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/CurrencyProvider.java
J-N-K: `SystemCurrency` may not be the best naming. It is a sort of "base currency", that is a used for conversion between different currencies. Separating that into an `ExchangeProvider` and `CurrencyProvider` makes it much more difficult.

splatch: Its easier to introduce complete design now than later. Once this code is part of codebase you will need to look for multiple compromises. Same thing apply to naming.

J-N-K: True, but the UoM system requires a base unit for conversions, so we have to set one in any case. UoM also requires that we can convert any unit to any other compatible unit, so if we have "CurrencyA" and "CurrencyB" with an exchange rate "AB" and a "CurrencyC" and "CurrencyC" with an exchange rate "CD" but no exchange rate for "CurrencyC" or "CurrencyD" to "CurrencyA" to "CurrencyB" we have units that we can't convert. Even if we have say "BD" it is very hard to determine the correct exchange rate for "AC", because we would have to check every possible combination if we find one that allows conversion (in this case A->B->D->C). It is far easier (also for the implementation of the exchange rate provider) to have exchanges rates relative to a given "default" (the base unit), in which case at most two conversions (A->base->C) are required.

That said: probably this is already over-engineered, since we are talking about a home automation system where probably no one needs more than one currency.

splatch: My earlier nitpick was mainly about naming, not functionality. I see you made some adjustments in naming which makes new API is less confusing. I'm happy with currency handling will become part of codebase.
My knowledge of indriya internals is fairly limited, thus I can't comment cleverness of algorithms applied there.

jlaur (changes requested):
Some minor comments (I know it's still a draft).

I plan to refactor openhab/openhab-addons#14376 quickly to use this when finished and merged.

bundles/org.openhab.core/src/test/java/org/openhab/core/library/unit/CurrencyUnitTest.java
```diff
@@ -0,0 +1,110 @@
...
+        QuantityType<Currency> convertedAmount = amount.toUnit("$");
+
+        assertThat(convertedAmount, is(notNullValue()));
+        assertThat(convertedAmount.getUnit(), is(TestCurrencyProvider.USD));
+        assertThat(convertedAmount.doubleValue(), closeTo(1.8845, 1e-4));
+    }
+
+    @Test
+    public void testPriceCalculation() {
+        QuantityType<?> unitPrice = new QuantityType<>("0.25 €/kWh");
```
jlaur, line 77: Shouldn't this only work for "0.25 EUR/kWh", since the symbol is ambiguous and we're using the currency code as key?

J-N-K: The `CurrencyProvider` has to take care that the symbol is unique. For the system provider there is only one currency, so there can't be ambiguity. For (future) extensions, the provider has to make sure that e.g. `Kr` or `$` is only used for one currency and either not set a symbol for other currencies or prefix them (like `AUS$`).

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/library/unit/CurrencyService.java
```diff
@@ -0,0 +1,86 @@
...
+import org.osgi.service.component.annotations.Activate;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Reference;
+import org.osgi.service.component.annotations.ReferenceCardinality;
+import org.osgi.service.component.annotations.ReferencePolicy;
+
+import tech.units.indriya.format.SimpleUnitFormat;
+
+/**
+ * The {@link CurrencyService} is allows to register and switch {@link CurrencyProvider}s and provides exchange rates
```
jlaur, line 40:
```suggestion
 * The {@link CurrencyService} allows to register and switch {@link CurrencyProvider}s and provides exchange rates
```

bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/CurrencyProvider.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.function.Function;
+
+import javax.measure.Unit;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+import org.openhab.core.library.dimension.Currency;
+
+/**
+ * The {@link CurrencyProvider} can be implemented by services that supply currencies and their
```
jlaur, line 26: and their?

bundles/org.openhab.core/src/test/java/org/openhab/core/library/unit/CurrencyUnitTest.java
jlaur: Hmm, it's quite possible (even likely) I don't fully understand it yet. Let's say you have two currencies:
- Currency code: NOK, symbol: kr
- Currency code: SEK, symbol: kr

I then receive an energy price in NOK, e.g. 1 NOK/kWh, and would like that converted (through an exchange rate service) to SEK. How would I then define that quantity, if not like `new QuantityType<>("1 NOK/kWh")`? I imagined that symbols would only be used for something like state patterns, i.e. only for display purposes.

jlaur (commented):
Submitting review to publish comment made after unrelated minor code comments.

J-N-K:
The `CurrencyService` is responsible for injecting a set of one or more currencies to the UoM system. It always injects the whole set from a `CurrencyProvider`, previously selected sets from other `CurrencyProvider`s are removed. It is the responsibility of a `CurrencyProvider` that all symbols are unique, so a `CurrencyProvider` that defines DKK, NOK and SEK may assign the `kr` symbol to only one of these currencies. Using the ISO codes is always possible (the symbols are just aliases for the ISO code). Besides the set of currency it also provides a "system currency" (a "reference" currency) and an exchange rate for each currency relative to that "system currency". 

The system provider (which is part of the `I18NProviderImpl` only provides one currency, that is the one defined by the locale. Currently no other implementations exist.

jlaur (changes requested):
Two minor comments.

bundles/org.openhab.core/src/test/java/org/openhab/core/library/unit/CurrencyUnitTest.java
jlaur: I didn't mean to test that, only to ensure coverage for using a currency code rather than currency symbol. So simply the same testcase but using for example "EUR" instead of "€".

jlaur:
@openhab/core-maintainers - any chance to have this PR reviewed? 🙂

kaikreuzer:
I can have a look at it asap.
@jlaur As you are deep in the topic yourself: Are you fine with the current state here? The "a CurrencyProvider that defines DKK, NOK and SEK may assign the kr symbol to only one of these currencies" doesn't seem to have your approval, if I get your comments right?

jlaur:
> I can have a look at it asap.

Awesome, thanks! 👍

> @jlaur As you are deep in the topic yourself: Are you fine with the current state here?

From reading the code I don't see any issues. But I have to admit that I scratched only the surface and tried to ask a few questions from my somewhat limited understanding of this area (I read only the new code, not the current code). Everything else I'm leaving for core maintainers, I'm just a passer-by with some interest in the topic. 😄

I did not yet get my hands dirty with it, i.e. tried to build the core project, implement currency support in Energi Data Service and/or implement a currency provider. Only then would I probably really understand what is being offered here. 🙂 I am eager to try it out though, and the concept looks promising.

> The "a CurrencyProvider that defines DKK, NOK and SEK may assign the kr symbol to only one of these currencies" doesn't seem to have your approval, if I get your comments right?

Correct, but that's only a slight difference of opinion and should not prevent a merge. Since currency symbols like "kr." are ambiguous I don't see the benefit of using them for anything other than display/formatting purposes. Currency codes like DKK are ISO standard and unambiguous.

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/library/unit/CurrencyService.java
```diff
@@ -30,57 +33,110 @@
...
+                DefaultCurrencyProvider.getInstance().getName());
+        CurrencyProvider currencyProvider = currencyProviders.getOrDefault(configuredCurrencyProvider,
+                DefaultCurrencyProvider.getInstance());
         enableProvider(currencyProvider);
     }
 
     @Reference(cardinality = ReferenceCardinality.MULTIPLE, policy = ReferencePolicy.DYNAMIC)
     public void addCurrencyProvider(CurrencyProvider currencyProvider) {
-        currencyProviders.add(currencyProvider);
+        logger.error("Added {}", currencyProvider.getName());
```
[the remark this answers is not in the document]
jlaur: The error logging is still here?

jlaur:
> I did not yet get my hands dirty with it, i.e. tried to build the core project, implement currency support in Energi Data Service and/or implement a currency provider. Only then would I probably really understand what is being offered here. 🙂 I am eager to try it out though, and the concept looks promising.

Until I can try this out, maybe let me run through an example to verify if my understanding is correct.

New channel type definition:
```xml
	<channel-type id="spot-price">
		<item-type>Number:EnergyPrice</item-type>
		<label>Spot Price</label>
		<description>Spot price.</description>
		<category>Price</category>
		<state readOnly="true" pattern="%.9f %unit%"></state>
	</channel-type>
```

I would assume the channel can be updated like this:
```java
updateState(CHANNEL_SPOT_PRICE, new QuantityType<>(spotPrice + " DKK/kWh");
```
(here I'm not sure if it can be provided as value and unit separately, but that's a minor thing)

For the `spotPrice` value 1, with a linked item without unit metadata and for base unit configured as "DKK", I would now assume the state pattern description will make sure to display the item state as "1.000000000 kr./kWh".

Is that correct?

If I now change either base unit configuration to "EUR" or item metadata unit to "EUR/kWh", I'm not sure what would then be displayed, since I don't have any `CurrencyProvider` implementation yet that can convert DKK to EUR. Will it still show DKK/kr.?

If I implement a currency exchange converter (i.e. a `CurrencyProvider`) that is able to convert DKK to EUR, I would then assume that the previous channel update example and unit configuration of "EUR/kWh" would then result in a state similar to "0.130000000 €/kWh"?

Last, in a JavaScript rule I would now assume I could do something like this:
```javascript
var spotPrice = items.SpotPrice.quantityState; // Unit is "EUR/kWh"
var consumption = Quantity("2 kWh");
var totalPrice = consumption.multiply(spotPrice);
console.log("Total price: " + totalPrice);
```

and I would see this logged:
```
Total price: 0.26 €
```

@J-N-K - can you confirm or correct my assumptions here? @kaikreuzer - to me this would seem logical and correct, so if confirmed, I'm very happy with the result. If not, let's figure out where I'm wrong and synchronize.

jlaur:
> > The "a CurrencyProvider that defines DKK, NOK and SEK may assign the kr symbol to only one of these currencies" doesn't seem to have your approval, if I get your comments right?
> 
> Correct, but that's only a slight difference of opinion and should not prevent a merge. Since currency symbols like "kr." or "$" are ambiguous I don't see the benefit of using them for anything other than display/formatting purposes. Currency codes like DKK are ISO standard and unambiguous.

In relation to this, I'm now wondering how to override/configure display unit. Let's say channel `electricity#spot-price` here is updated with energy price in DKK/kWh:
```
Number:EnergyPrice EnergiDataService_SpotPrice "Spotpris [%.2f %unit%]" <price> (EnergiDataService_TotalPrice) { unit="DKK/kWh", channel="energidataservice:service:energidataservice:electricity#spot-price" [profile="transform:VAT"] }
```

I assume I can trigger a currency conversion for display purposes:
```
Number:EnergyPrice EnergiDataService_SpotPrice "Spotpris [%.2f EUR/kWh]" <price> (EnergiDataService_TotalPrice) { unit="DKK/kWh", channel="energidataservice:service:energidataservice:electricity#spot-price" [profile="transform:VAT"] }
```

But what if I wanted to not only trigger a currency conversion, but also currency symbol display?
```
Number:EnergyPrice EnergiDataService_SpotPrice "Spotpris [%.2f kr./kWh]" <price> (EnergiDataService_TotalPrice) { unit="DKK/kWh", channel="energidataservice:service:energidataservice:electricity#spot-price" [profile="transform:VAT"] }
```
(note: symbol for DKK is "kr." with dot... for Swedish Kroner it would be just "kr" without dot)

This would be ambiguous, so what would actually happen - is it possible at all? I agree that unit symbol is often preferable for display purposes.

jlaur:
See also https://github.com/openhab/openhab-core/pull/3597#issuecomment-1783865866.

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/library/unit/CurrencyService.java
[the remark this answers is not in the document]
J-N-K: done

jlaur:
Thanks for rebasing, @J-N-K. I have now built `org.openhab.core.config.core` and `org.openhab.core` from your fork and replaced the JARs in my installation from latest snapshot.

Configured provider:
image

(I'm wondering if this should be part of regional settings rather than a separate unit settings - do you expect more unit settings since you put it separately?)

Created an item:
```yaml
label: Currency Item
type: Number:Currency
category: price
groupNames: []
groupType: None
function: null
tags: []
```

image

And first tests:
```shell
openhab> openhab:update CurrencyItem "100 kr."
Update has been sent successfully.
openhab> openhab:update CurrencyItem "200 DKK"
Error: State '200 DKK' is not valid for item 'CurrencyItem'
Valid data types are: ( DecimalType QuantityType UnDefType )
openhab> openhab:update CurrencyItem "300 $"
Error: State '300 $' is not valid for item 'CurrencyItem'
Valid data types are: ( DecimalType QuantityType UnDefType )
openhab> openhab:update CurrencyItem "400 €"
Error: State '400 €' is not valid for item 'CurrencyItem'
Valid data types are: ( DecimalType QuantityType UnDefType )
openhab> openhab:update CurrencyItem "500 kr"
Error: State '500 kr' is not valid for item 'CurrencyItem'
Valid data types are: ( DecimalType QuantityType UnDefType )
```

Result:
image

Besides not being able to use currency code for the update, I think it behaves as I would expect. I assume the error when trying unit symbols $, € and kr. is because I have configured DKK indirectly through `LocaleBasedCurrencyProvider` and because I don't have any `CurrencyProvider` that can convert DKK to any of those?

When eventually adding a new `CurrencyProvider` implementing currency exchange, how do I select in the unit settings that my global/system currency is DKK and at the same time select which currency provider to use? Or do you need to define the currency on each item in this case?

kaikreuzer:
> Since currency symbols like "kr." or "$" are ambiguous I don't see the benefit of using them for anything other than display/formatting purposes. 

This is correct, but for display/formatting purposes you would want SEK and NOK both use "kr" as a symbol - and if only one of both is allowed to use it, that would be a problem, wouldn't it?

jlaur:
> > Since currency symbols like "kr." or "$" are ambiguous I don't see the benefit of using them for anything other than display/formatting purposes.
> 
> This is correct, but for display/formatting purposes you would want SEK and NOK both use "kr" as a symbol - and if only one of both is allowed to use it, that would be a problem, wouldn't it?

Indeed. Perhaps my misconception is that I somehow assumed that display/formatting could have an asymmetric relation to "real" unit, so that both SEK/NOK would be displayed as "kr" but still maintain their currency code for all other matters. But of course this wouldn't work the other way, so for example a sitemap `Input` would have to accept unit "kr" as otherwise the user would have to correct the content from e.g. "1 kr" to "1 SEK" in order to provide the quantity in the opposite direction.

After running the tests in my previous post I'm still a bit confused about the roles of `FixedCurrencyProvider` vs. `LocaleBasedCurrencyProvider` in relation to currency code vs. currency symbol.

If it's possible for a binding to publish SEK, and for a user to use a currency exchange service (through a `CurrencyProvider` I assume) to convert to NOK (after all, Norway and Sweden are neighbors 🙂), then it should be fine. If that would not work, I think we have a problem to solve.

J-N-K:
If a `CurrencyProvider` provides SEK and NOK and their exchange rate, it's perfectly possible to convert between them, and it is done in the same way like meter to foot.

Both, `LocaleBasedCurrencyProvider` and `FixedCurrencyProvider` provide exactly one currency. So they essentially allow using that unit in price calculations. The difference is that the `LocaleBasedCurrencyProvider` uses the configured locale to determine the currency, while the `FixedCurrencyProider` can be configured to any currency the user thinks is a good choice. Neither of them allow conversion of currencies, as they only know about one currency.

jlaur:
> What is the state here? Any change request or can this be merged?

From my perspective, here's a shortened summary of concerns from my testing session:
- When using the `FixedCurrencyProvider`, currency **code** can only be used as unit when updating state, and is also shown in the UI, whereas when using `LocaleBasedCurrencyProvider`, currency **symbol** is used for both. This seems inconsistent, is this expected?
- Is it expected that the state **value** is kept, but the **unit** is changed, when another currency is configured? Unit metadata doesn't seem to have much impact. See details above.
- UI: I'm wondering if this should be part of regional settings rather than a separate unit settings - do you expect more unit settings since you put it separately? It it should have its own unit section, I think the existing unit settings from regional settings should be moved there.

J-N-K:
- The `FixedCurrencyProvider` does provide exactly the currency that is configured. You can also define something like `JLC` for "Jacob Laursen's currency". Maybe we could add an option to add the symbol. The `LocaleBasedCurrencyProvider` selects a "real" (i.e. existing) currency and (if available) a symbol for that. 
- If you switch the currency, you essentially say that "what was called x before is now y". The unit stays the same, it just gets a new name (and symbol if available). I believe that this happens only on rare occasions, I cannot think of a use-case of a frequently switched currency. 
- Currently the metadata has no effect, because with the two providers we have, the whole UoM system knows only exactly one unit (=currency), the one provided by the selected `CurrencyProvider`. The `CurrencyProvider` provides a "system unit" and a set of additional units and their conversion factors to the system unit. If we had the same for length, a `LengthProvider` could set a system unit of `m` and also provider `ft` and `Å`, because it knows how to convert between those. If you set a value of `5 parsec` it would not work, because neither the unit `parsec` nor the conversion factor to `m` is known. 
- IMO the whole settings page needs to be reworked for localized settings. Why do we have "Ephemeris" as a section and need to set "Country" again, if we have the same setting under "Regional Settings". IMO it would be much better to move the UoM settings to a new section. I can also imagine that we have configurable "default" units for other dimensions (like length).

jlaur:
Thank you for your detailed answer, @J-N-K. That removes my last doubts. Regarding the settings page, this can always be reworked later without breaking anything. I just wondered why a new page was introduced when we can already configure the measurement system in regional settings.

If by any chance this PR would be ready for a merge soon, I should be able to implement currency support in Energy Data Service binding for 4.1. Currency exchange probably not. 😄

jlaur:
@kaikreuzer - gentle ping. 🙂

kaikreuzer (approved):
lgtm, thanks @J-N-K.
@jlaur Sorry for the delay, I didn't follow your discussion here and thus didn't notice that you approved the PR. You still have one day left for adapting your binding. 😄

jlaur:
> Sorry for the delay, I didn't follow your discussion here and thus didn't notice that you approved the PR. You still have one day left for adapting your binding. 😄

Thank you. 😄 I'll give it a go, I have a few hours left. Any chance of a full build now so that a PR in addons will build successfully?

J-N-K:
I have triggered a core build, once it's done I'll trigger an add-ons build and a distro-build.

jlaur:
> I have triggered a core build, once it's done I'll trigger an add-ons build and a distro-build.

Thanks! Meanwhile I started looking into the implementation. Is it correct that a state update for kr./kWh can/should now be done like this?
```java
BigDecimal price = new BigDecimal("1");
updateState("price", new QuantityType<EnergyPrice>(price + " DKK/kWh"));
```

kaikreuzer:
I had already triggered a core build and all is available with https://ci.openhab.org/job/openHAB-Core/1310/.

kaikreuzer:
With the addons build I am waiting for https://github.com/openhab/openhab-addons/pull/16060 to succeed first.

PR #3504, opened 2023-03-28
Use try-with-resources on Files.walk/Files.list

J-N-K:
It is recommended to use try-with-resources when using `File.list` or `Files.walk` to ensure streams are properly closed.


wborn (approved):
Thanks! I also had checking all the usages on my virtual todo list. :wink:

PR #3508, opened 2023-03-29
Code cleanup: Use Java17 features

holgerfriedrich:
Applied new language features, mainly instanceof pattern matching. Just on bundles/org.openhab.core.


wborn (approved):
Thanks for the massive cleanup @holgerfriedrich!

PR #3511, opened 2023-03-31
Fix Semantics default synonyms

jimtng:
Fix incorrect Property id


bundles/org.openhab.core.semantics/src/main/java/org/openhab/core/semantics/Property.java
```diff
@@ -20,6 +20,6 @@
  * @author Kai Kreuzer - Initial contribution
  */
 @NonNullByDefault
-@TagInfo(id = "MeasurementProperty")
```
jimtng, line 23: @kaikreuzer was this an unintentional typo?

kaikreuzer: I have no clue. It might be that `MeasurementProperty` was the original name and we missed to update that line when we simplified the name to `Property`.

kaikreuzer (approved):
lgtm

Issue #3380, opened 2023-02-17
ItemChannelLinkOSGiTest unstable


wborn:
This test failed in a GHA build: https://github.com/openhab/openhab-core/actions/runs/4201120561/jobs/7287810287

```
TEST org.openhab.core.thing.link.ItemChannelLinkOSGiTest#assertThatAllLinksForItemCanBeDeleted() <<< ERROR: 
Expected: a collection with size <27>
     but: collection size was <21>
java.lang.AssertionError: 
Expected: a collection with size <27>
     but: collection size was <21>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
	at org.openhab.core.thing.link.ItemChannelLinkOSGiTest.lambda$12(ItemChannelLinkOSGiTest.java:230)
```

J-N-K:
How can this test fail? It just checks if all links have been added to the registry...

wborn:
Perhaps the test depends on some logic being executed in another thread which sometimes has not yet finished running.

PR #3513, opened 2023-04-01
Fix resource leak in SysfsUsbSerialScanner

J-N-K:
This is a regression from #3290 and was reported on the community forum and results in "too many open files" errors after some days.


wborn (approved):
LGTM!

wborn:
The GHA failure is just another occurence of #3380.

PR #3519, opened 2023-04-02
Add dynamic creation of semantic tags

jimtng:
Add the ability to dynamically add new semantic tags

For example, to add a new tag called `MainBedroom` with `Bedroom` as the parent:
`SemanticTags.add("MainBedroom", Bedroom.class)` or
`SemanticTags.add("MainBedroom", "Bedroom")` 

The new tags can then be used in items and rules.

A text file in the following format can be easily loaded and parsed with a simple rule on startup in any scripting language.

```
NewTag(Label,Synonym1,Synonym2): ParentTag A long description can go here
SecretRoom: Room
SwimmingPool: Location
Motion: Status
Color: Property
Brightness: Property
```

or go full YAML... the syntax is up to the rule writer.

Currently the MainUI can recognise and display the added tags:

image

However, because MainUI uses a static list of tags (loaded from a file), it doesn't list the new tags in the tag selection dialog. More work is needed to provide a REST end point which the MainUI can then query rather than having to duplicate the list of tags in a static file like it currently does.

image

### DIFF
--- a/bundles/org.openhab.core.semantics/model/generateTagClasses.groovy
+++ b/bundles/org.openhab.core.semantics/model/generateTagClasses.groovy
@@ -134,6 +134,10 @@ public class Locations {
     public static Stream<Class<? extends Location>> stream() {
         return LOCATIONS.stream();
     }
+
+    public static boolean add(Class<? extends Location> tag) {
+        return LOCATIONS.add(tag);
+    }
 }
 """)
     file.close()
@@ -171,6 +175,10 @@ public class Equipments {
 
     public static Stream<Class<? extends Equipment>> stream() {
         return EQUIPMENTS.stream();
+    }
+
+    public static boolean add(Class<? extends Equipment> tag) {
+        return EQUIPMENTS.add(tag);
     }
 }
 """)
@@ -210,6 +218,10 @@ public class Points {
     public static Stream<Class<? extends Point>> stream() {
         return POINTS.stream();
     }
+
+    public static boolean add(Class<? extends Point> tag) {
+        return POINTS.add(tag);
+    }
 }
 """)
     file.close()
@@ -247,6 +259,10 @@ public class Properties {
 
     public static Stream<Class<? extends Property>> stream() {
         return PROPERTIES.stream();
+    }
+
+    public static boolean add(Class<? extends Property> tag) {
+        return PROPERTIES.add(tag);
     }
 }
 """)
--- a/bundles/org.openhab.core.semantics/pom.xml
+++ b/bundles/org.openhab.core.semantics/pom.xml
@@ -20,6 +20,12 @@
       <artifactId>org.openhab.core</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.ow2.asm</groupId>
+      <artifactId>asm</artifactId>
+      <version>9.2</version>
+      <scope>provided</scope>
+    </dependency>
   </dependencies>
 
   <profiles>
--- a/bundles/org.openhab.core.semantics/src/main/java/org/openhab/core/semantics/SemanticTags.java
+++ b/bundles/org.openhab.core.semantics/src/main/java/org/openhab/core/semantics/SemanticTags.java
@@ -26,6 +26,9 @@ import java.util.stream.Stream;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
+import org.objectweb.asm.AnnotationVisitor;
+import org.objectweb.asm.ClassWriter;
+import org.objectweb.asm.Opcodes;
 import org.openhab.core.items.Item;
 import org.openhab.core.semantics.model.equipment.Equipments;
 import org.openhab.core.semantics.model.location.Locations;
@@ -33,12 +36,15 @@ import org.openhab.core.semantics.model.point.Measurement;
 import org.openhab.core.semantics.model.point.Points;
 import org.openhab.core.semantics.model.property.Properties;
 import org.openhab.core.types.StateDescription;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * This is a class that gives static access to the semantic tag library.
  * For everything that is not static, the {@link SemanticsService} should be used instead.
  *
  * @author Kai Kreuzer - Initial contribution
+ * @author Jimmy Tanagra - Add the ability to add new tags at runtime
  */
 @NonNullByDefault
 public class SemanticTags {
@@ -46,6 +52,9 @@ public class SemanticTags {
     private static final String TAGS_BUNDLE_NAME = "tags";
 
     private static final Map<String, Class<? extends Tag>> TAGS = new TreeMap<>();
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(SemanticTags.class);
+    private static final SemanticClassLoader CLASS_LOADER = new SemanticClassLoader();
 
     static {
         Locations.stream().forEach(location -> addTagSet(location));
@@ -203,6 +212,117 @@ public class SemanticTags {
         return null;
     }
 
+    /**
+     * Adds a new semantic tag with inferred label, empty synonyms and description.
+     * 
+     * The label will be inferred from the tag name by splitting the CamelCase with a space.
+     * 
+     * @param name the tag name to add
+     * @param parent the parent tag that the new tag should belong to
+     * @return the created semantic tag class, or null if it was already added.
+     */
+    public static @Nullable Class<? extends Tag> add(String name, String parent) {
+        return add(name, parent, null, null, null);
+    }
+
+    /**
+     * Adds a new semantic tag.
+     * 
+     * @param name the tag name to add
+     * @param parent the parent tag that the new tag should belong to
+     * @param label an optional label. When null, the label will be inferred from the tag name,
+     *            splitting the CamelCase with a space.
+     * @param synonyms a comma separated list of synonyms
+     * @param description the tag description
+     * @return the created semantic tag class, or null if it was already added.
+     */
+    public static @Nullable Class<? extends Tag> add(String name, String parent, @Nullable String label,
+            @Nullable String synonyms, @Nullable String description) {
+        Class<? extends Tag> parentClass = getById(parent);
+        if (parentClass == null) {
+            LOGGER.warn("Adding semantic tag '{}' failed because parent tag '{}' is not found.", name, parent);
+            return null;
+        }
+        return add(name, parentClass, label, synonyms, description);
+    }
+
+    /**
+     * Adds a new semantic tag with inferred label, empty synonyms and description.
+     * 
+     * The label will be inferred from the tag name by splitting the CamelCase with a space.
+     * 
+     * @param name the tag name to add
+     * @param parent the parent tag that the new tag should belong to
+     * @return the created semantic tag class, or null if it was already added.
+     */
+    public static @Nullable Class<? extends Tag> add(String name, Class<? extends Tag> parent) {
+        return add(name, parent, null, null, null);
+    }
+
+    /**
+     * Adds a new semantic tag.
+     * 
+     * @param name the tag name to add
+     * @param parent the parent tag that the new tag should belong to
+     * @param label an optional label. When null, the label will be inferred from the tag name,
+     *            splitting the CamelCase with a space.
+     * @param synonyms a comma separated list of synonyms
+     * @param description the tag description
+     * @return the created semantic tag class, or null if it was already added.
+     */
+    public static @Nullable Class<? extends Tag> add(String name, Class<? extends Tag> parent, @Nullable String label,
+            @Nullable String synonyms, @Nullable String description) {
+        if (getById(name) != null) {
+            return null;
+        }
+
+        if (!name.matches("[A-Z][a-zA-Z0-9]+")) {
+            throw new IllegalArgumentException(
+                    "The tag name '" + name + "' must start with a capital letter and contain only alphanumerics.");
+        }
+
+        String parentId = parent.getAnnotation(TagInfo.class).id();
+        String type = parentId.split("_")[0];
+        String className = "org.openhab.core.semantics.model." + type.toLowerCase() + "." + name;
+
+        // Infer label from name, splitting up CamelCaseALL99 -> Camel Case ALL 99
+        label = Optional.ofNullable(label).orElseGet(() -> name.replaceAll("([A-Z][a-z]+|[A-Z][A-Z]+|[0-9]+)", " $1"))
+                .trim();
+        synonyms = Optional.ofNullable(synonyms).orElse("").replaceAll("\\s*,\\s*", ",").trim();
+
+        // Create the tag interface
+        ClassWriter classWriter = new ClassWriter(0);
+        classWriter.visit(Opcodes.V11, Opcodes.ACC_PUBLIC + Opcodes.ACC_ABSTRACT + Opcodes.ACC_INTERFACE,
+                className.replace('.', '/'), null, "java/lang/Object",
+                new String[] { parent.getName().replace('.', '/') });
+
+        // Add TagInfo Annotation
+        classWriter.visitSource("Status.java", null);
+
+        AnnotationVisitor annotation = classWriter.visitAnnotation("Lorg/openhab/core/semantics/TagInfo;", true);
+        annotation.visit("id", parentId + "_" + name);
+        annotation.visit("label", label);
+        annotation.visit("synonyms", synonyms);
+        annotation.visit("description", Optional.ofNullable(description).orElse("").trim());
+        annotation.visitEnd();
+
+        classWriter.visitEnd();
+        byte[] byteCode = classWriter.toByteArray();
+        Class newTag = null;
+        try {
+            newTag = CLASS_LOADER.defineClass(className, byteCode);
+        } catch (Exception e) {
+            LOGGER.warn("Failed creating a new semantic tag '{}': {}", className, e.getMessage());
+            return null;
+        }
+        addToModel(newTag);
+        addTagSet(newTag);
+        if (LOGGER.isTraceEnabled()) {
+            LOGGER.trace("'{}' semantic {} tag added.", className, type);
+        }
+        return newTag;
+    }
+
     private static void addTagSet(Class<? extends Tag> tagSet) {
         String id = tagSet.getAnnotation(TagInfo.class).id();
         while (id.indexOf("_") != -1) {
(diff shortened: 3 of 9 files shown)


jimtng:
A text file in the format of
```
NewTag: ParentTag
SecretRoom: Room
SwimmingPool: Location
Motion: Status
Color: Property
Brightness: Property
```

can be easily loaded and parsed with a simple rule on startup in any scripting language.

rkoshak:
Does this break HABot? That's where the semantic model came from in the first place.

jimtng:
> Does this break HABot? That's where the semantic model came from in the first place.

This merely lets you _add_ more tags, not delete/alter existing ones. I don't see how it would break anything, although to be sure it would need to be tested by someone familiar with habot.

rkoshak:
My concern with HABot is it actually uses the semantic model to parse arbitrary text to understand the intent. Each semantic tag has a specific meaning to HABot. My worry is that arbitrarily adding them could potentially confuse HABot's human language processor if the new tags are not properly integrated into the ontology.

jimtng:
> My concern with HABot is it actually uses the semantic model to parse arbitrary text to understand the intent. Each semantic tag has a specific meaning to HABot. My worry is that arbitrarily adding them could potentially confuse HABot's human language processor if the new tags are not properly integrated into the ontology.

While I haven't looked closely into HABot:
- Adding more tags can most likely help it improve, if the intention of adding / tweaking was to improve it
- The onus is on each individual user who added more tags to know that it may affect things
- For the rest of us who don't use habot, it is of zero concern, meanwhile the benefit for better organisation is tremendous to the rest (items / rules structure). So many people, myself included, have been begging to have this ability for years.

rkoshak:
Only if adding the tags actually works and doesn't break it. It would even be OK if the added tags are simply ignored by HABot. But based on past experiences working with ontologies (granted that was a decade ago), it's almost never as simple as just adding a new tag and away we go.

> The onus is on each individual user who added more tags to know that it may affect things
> For the rest of us who don't use habot, it is of zero concern, meanwhile the benefit for better organisation is tremendous to the rest (items / rules structure). So many people, myself included, have been begging to have this ability for years.

I don't think it's right to add a feature that can render a preexisting feature inoperable with the only mitigation being "well, you should have been more careful". 

And while breaking HABot is of zero concern to you, it's not to me and everyone else who uses HABot (see https://community.openhab.org/t/self-hosted-notifications-with-nextcloud-talk/144090 for just one use case).

"I want it really badly" isn't justification for breaking things IMO.

@ghys, I hate to tag you but you are the resident expert on the ontology and developer of HABot. Are my concerns out of line and there really is no issue here, or does this have the potential to break HABot as I fear?

J-N-K:
OTOH, using "it'll break something" can't be the reason for not going forward. It seems that HABot is not maintained anymore, so it should not be the reason to stop development in other areas.

That aside: I don't care at all, I don't use semantic tags.

jimtng:
While I understand your concern regarding its potential impact to habot, I don't understand why it shouldn't be added because of it. This doesn't do anything unless _deliberately_ used.

I am sure there are many other things that fall in the similar category. One that instantly came to mind is `sudo rm -rf /`. One can type gibberish in a script, or drag and drop random blocks in blockly. Just because the `rm` command or the `-rf` option can potentially do damage, doesn't mean people shouldn't have it.

rkoshak:
I'm not saying not to develop this. I would like to be able to add new semantic tags too.

What I am saying is let's make sure that we understand the impacts to other parts of OH and mitigate those impacts as much as possible. 

"I don't use it so who cares" is a pretty hostile attitude towards the end users. That's in fact the exact stated reason for why a number of OH users abandoned HomeAssistant in favor of OH. Things break without warning because the developers don't care how it impacts the end users. I'd hate to see us heading in the same direction. 

And if this does break HABot and if HABot is truly abandoned than let's just remove it (which would be a shame because one of the most magical things with OH is the ability to tell it to do stuff using natural language) rather than setting up a  bunch of land mines for end users to find which blows it up. And this one will be found right away because those who use HABot are definitely going to want to use these new custom defined tags too.

Again, to summarize, I'm not saying "no don't do it!" I'm not even saying "no don't do it if it breaks HABot!" I am saying lets spend maybe just a couple of minutes or so to find out how it impacts HABot and if it does, how to mitigate any problems it may cause so end users don't just wake up one day and find HABot doesn't work because they added a custom semantic tag. I really don't think that's too much to ask.

jimtng:
> I am saying lets spend maybe just a couple of minutes or so to find out how it impacts HABot and if it does, how to mitigate any problems it may cause so end users don't just wake up one day and find HABot doesn't work because they added a custom semantic tag.

Thanks for clarifying. Unfortunately I am not familiar with habot. The last time I "played" with it was uhm... when I first came across openhab in 2017-18 maybe. Waiting to hear from @ghys but would you also be able to test?

J-N-K:
@rkoshak You should know that I always try to keep things as backward compatible as possible and to have smooth migration path when this is not possible. "Thing that break without warning" is really something that should be avoided and I don't think it is common in openHAB to have that. We have made a big step forward in avoiding such situations with #3330, especially for UI users.

What I wanted to express is: I don't use HABot and I don't use semantic tags. I have no preference for "not extending tags in favor of HABot" or "extending tags and dropping HABot".

Regarding the maintenance state: HABot can't be build on current Apple machines since more than two years because of outdated NodeJS (which also brings a ton of outdated dependencies).

jimtng:
> I am saying lets spend maybe just a couple of minutes or so to find out how it impacts HABot

@rkoshak I was just thinking, is there a similar concern each time a new tag is added, or removed from the official list of tags in core? If so, is there a process in place to verify / test / check against habot? Perhaps we can utilise the same process.

rkoshak:
> Waiting to hear from @ghys but would you also be able to test?

Yes as long as I don't have to build. I don't have a dev environment and I understand it's a heavy lift to do so.

But the test should be relatively simple I would think:

1. make a query/command to HABot for something semantically tagged using the built in tags and verify that works as expected
2. add a custom semantic tag
3. repeat 1
4. add the custom tag to an Item
5. repeat 1
6. create a query to HABot for the new custom tag

Obviously watch the logs for errors.

I think that would show if ti breaks and whether the new tag is available to HABot (which is less of a concern to me than steps 1-5 failing for some reason).

> @rkoshak You should know that I always try to keep things as backward compatible as possible and to have smooth migration path when this is not possible.

I know you do and greatly appreciate that. It just seemed in this particular case that breaking HABot didn't rise to a level where it was even considered. I am happy that isn't the case.

> Regarding the maintenance state: HABot can't be build on current Apple machines since more than two years because of outdated NodeJS (which also brings a ton of outdated dependencies).

I'd really hate to lose it but if it's slowly dying out maybe now's the time to just remove it. 

> @rkoshak I was just thinking, is there a similar concern each time a new tag is added, or removed from the official list of tags in core? If so, is there a process in place to verify / test / check against habot? Perhaps we can utilise the same process.

That's a really good question and I don't know the answer to it. I'll see if I can dig up an old PR that added a new tag. It's been a good while since the last tag was added but I seem to recall that it was more than just adding the tag.

jimtng:
# Testing

Thanks for wanting to test.

1. Make a copy of this file somewhere like ~/ or in openhab/conf/ `runtime/system/org/openhab/core/bundles/org.openhab.core.semantics/4.0.0-SNAPSHOT/org.openhab.core.semantics-4.0.0-SNAPSHOT.jar`

2. Put this file in place of the above. This file is actually a .jar file. I appended .zip to the name so I can upload it here. Don't unzip it, just remove the extension off the end.
[org.openhab.core.semantics-4.0.0-SNAPSHOT.jar.zip](https://github.com/openhab/openhab-core/files/11134189/org.openhab.core.semantics-4.0.0-SNAPSHOT.jar.zip)

3. Stop openhab
4. `rm -rf ~/userdata/cache/org.eclipse.osgi/ ~/userdata/tmp/*` (assuming `~` = the openhab's main directory, otherwise adjust accordingly
5. Start openhab

Then proceed with your tests that you've outlined.

In order to add a tag, use jsscripting or jruby. I'm not familiar with how to do this in jsscripting, so here's a jruby version:

```
org.openhab.core.semantics.SemanticTags.add("YourCustomTag", "ExistingParent")

# or:
org.openhab.core.semantics.SemanticTags.add("YourCustomTag", "ExistingParent", "Custom Label", "Synonym1,Synonym2", "Description")
```

Existing parent can be `Property`, `Location`, `Equipment`, `Point`, or any one of the other valid tags, e.g. `Indoor`, `Room`, `Door`, `Control`, `Light`, etc.

It won't let you use an invalid parent tag nor an invalid tag name, etc.

The only way to remove the custom tags is by making sure your script doesn't run (e.g. comment them out) then restart openhab.



> I'd really hate to lose it but if it's slowly dying out maybe now's the time to just remove it.
> 
I think HABot is really cool. I just really haven't had the "time" to play with it / use it. It would be a shame to kill it. It probably just needs some TLC?

rkoshak:
> org.openhab.core.semantics.SemanticTags

Given the name I'm guess that you are working with the actual Java Class here? In that case it's pretty straight forward. I just need to use `Java.type` to import it and the calls should be the same.

I'm currently upgrading to the latest snapshot. When that's done I'll give it a go.

jimtng:
I wonder if the label is what's actually being used by the language interpreter, e.g. when your label is "Foo tag", you might have to say "Turn on Foo tag" instead of "Turn on Foo".

rkoshak:
OK, let the testing resume!

I installed the new jar file pre instructions.

1. worked as expected. 
2. No error in the rule so I'm assuming it worked
```
var SemanticTags = Java.type('org.openhab.core.semantics.SemanticTags');
SemanticTags.add("Foo", "Point");
SemanticTags.add('FooBar", "Property");
```
3. worked as expected, so we have confirmation that the new tag doesn't break HABot. Now to see if HABot can see the new tag.
4. I created a new Point Item through the model screen. The "Foo" tag was not listed as an option in the UI which is unsurprising. So I manually added the tags as regular old tags. I gave it the "Foo" point tag and "FooBar" property tag. I placed it into an existing "Lightbulb" Equipment which is the same equipment I've been using for the "control" tests.
5. and 6. 
  a. The good news is it sees the Point tag but it doesn't see/understand the Property tag. I can ask it to "what's the Foo in the living room" and it finds the Item. However, if I ask "what's the FooBar in the living room" it says it can't find it.
  b. When I say "show me the lights" the Foo Item appears along with all the other lights. That is probably because of the Equipment it belongs to more than anything else. 

In summary, merely creating the custom tags does not break HABot. However, it appears that more needs to be done to make HABot aware of custom Property tags. I didn't test Equipment nor Location yet but given that the custom Point worked, I expect those will work too. I'll test that though as well as testing what happens if I use something like "Light" as the parent Point tag.

jimtng:
Try calling it Foobar not FooBar.

FooBar creates a label "Foo Bar" with a space and that might confuse habot.

Regarding mainui not displaying it, I am aware of that. Already opened an issue to track and will work on that next after this is merged. Lots of work needed for that to work. Mainui currently keeps it's own hardcoded list of tags instead of asking core for the list.

rkoshak:
> Try calling it Foobar not FooBar.

Neither "Foobar" nor "foobar" worked but "foo bar" did work. So it uses camelcase to determine where spaces should go? This is kind of exciting! It looks like it might work after all, but we will have to document pretty closely how it works when using camelcase like that.

> Also, can habot understand it when you used an existing property tag in the same manner eg LightTemperature or even a single word property?

Yes, all the built in tags work. But you are right, the tag is "LivingRoom" but in the query I have to use "living room". Interestingly it doesn't seem to use the Item label in the query. The Item I have tagged with "LivingRoom" is labeled "Family Room" and HABot can't find it. But that's a concern for a different issue. It just cost me a bit of time thinking it wasn't working when it was and my query was wrong. 

I created a single word "Dog" tag with "Property" as the parent and applied that to my test Item and "show the dog in the living room" worked. 

So maybe it is working after all and the problem was I used a multiword tag and didn't know it.

Question, I've made a couple of assumptions:

1. when restarting OH these tags go away, at least for now
2. if I call to add a custom tag that already exists, it changes it; for example, if I wanted to change my "Dog" tag to use "Light" as it's parent I can just update and run my script again

1 is understandable at the current state of the PR. If my assumption is wrong, as it appears to be, 2 needs to throw an exception (similar to how it does when trying to use a tag that doesn't start with a capital) or at least a warning log statement if it's really just ignoring the change.

If my assumption for 2 was correct, after changing the "Dog" tag to inherit from "Light" I was unable to query for and get it to show up along side all the other Items tagged with "Light". However, looking at the Item seems to indicate that it never changed:

image

The class still shows Property_Dog, not Light_Dog.

Creating a new "Cat" tag that inherits from Light shows:

image

And indeed queries for lights now show the test Item as expected.

Tl;dr, it seems all the problems I've encountered are my own and not the code. This is awesome!

jimtng:
At it's current incarnation, you cannot re-add an existing tag. It checks against existing tags (built-in or otherwise) and will simply return null and do nothing.

It is documented that it returns null in this case, but I see that a warning logger is probably needed as well.

jimtng:
> It is documented that it returns null in this case, but I see that a warning logger is probably needed as well.

@rkoshak I go back and forth about this, but I prefer not to issue a warning log. Instead let the caller handle it. The method will return null to indicate this, so if the caller wants to issue warning they can do so themselves. This gives people the freedom to choose the behaviour rather than forcing it on them. 

```
if (Semantics.add("newtag", "parent") == null) {
  logger.warn("Adding failed - newtag already exists")
}
```

rkoshak:
> This is also documented.

Just to be clear though, by documented I mean the behavior needs to be described in the OH docs, not just in the code. End users are not going to go read comments in the code nor refer to the JavaDocs for the most part. Because this is a completely new capability from the end user's perspective, nothing about creating new semantic tags appears anywhere in the user guide.

It probably also makes sense that the code that adds/removes?/modifies? semantic tags be exposed as an Action if we are thinking this is something a user may want to do from a rule. I'm thinking that is probably not the best idea and tags should only really be manipulated through the REST API/UI/Karaf Console, but I don't always see all the use cases.

> Instead let the caller handle it.

As long as the caller isn't a rule I think I'm OK with that, so long as it's clearly documented at the level of the caller (e.g. if at the REST API it should be clear in the swagger docs what indicates success, ignored, or problem).

ghys:
> Again, to summarize, I'm not saying "no don't do it!" I'm not even saying "no don't do it if it breaks HABot!" I am saying lets spend maybe just a couple of minutes or so to find out how it impacts HABot and if it does, how to mitigate any problems it may cause so end users don't just wake up one day and find HABot doesn't work because they added a custom semantic tag. I really don't think that's too much to ask.

I agree fully with this. The request to let users expand the semantic model seems to come every other year or so, including when the feature was discussed at the time when it was merely in the plans (https://github.com/eclipse-archived/smarthome/issues/1093) and had yet to be implemented.

If HABot hasn't received much attention, this isn't because it's not maintained anymore, it's rather because "it works" and I moved on to other things - but I would be sad if it broke for no good reason. As the author I'll admit I rarely use the dedicated HABot UI these days but I still regularly use the text box on the main UI home page, mainly to recall controls from keywords ("kettle", "kitchen lights"...) that I haven't properly put on a page so they're accessible easily, and would be 3 to 4 clicks away if browsing through the semantics cards on the other tabs of the home page.

HABot is kind of test-driven, there are extensive tests to exercise the bundled skills in every language HABot supports - you can see the output about 3/4 of the way in the build logs (https://ci.openhab.org/job/openHAB-WebUI/611/consoleText)

```
----
"change the corridor lights to 70%"
{0.9491474334451203=[set-value], 0.022335667955037122=[get-status], 0.01294666887947513=[deactivate-object], 0.010717567115524455=[activate-object], 0.001731931115492209=[get-history-last-changes], 0.0012824703029145064=[get-history-monthly], 7.640074801369204E-4=[get-history-daily], 4.94159508736749E-4=[get-history-weekly], 3.533538500641698E-4=[get-history-hourly], 2.2674034749846687E-4=[create-rule]}
Intent [name=set-value, entities={location=corridor, value=70%, object=lights}]
----
```
If those didn't pass, for any reason, then the web UI builds would fail. I would expect any author of a change that has an impact in this area to own up to the regression they introduced and strive to fix it themselves ;)

ghys:
> In summary, merely creating the custom tags does not break HABot. However, it appears that more needs to be done to make HABot aware of custom Property tags. I didn't test Equipment nor Location yet but given that the custom Point worked, I expect those will work too. I'll test that though as well as testing what happens if I use something like "Light" as the parent Point tag.

And I'm grateful to @rkoshak for doing all this testing! Having some more Properties added could indeed cause HABot to behave slightly differently (because its training would be different) and could have an impact on the tests, this would be solved with some more [training material](https://github.com/openhab/openhab-webui/tree/main/bundles/org.openhab.ui.habot/src/main/resources/train).

rkoshak:
> Having some more Properties added could indeed cause HABot to behave slightly differently

I've still a bunch of tests to do but it does appear that my initial assessment was caused by using "FooBar" as the Property tag and querying HABot for "foobar" (single word). Once I got that working HABot was able to see and return the Item and control it based on the "foo bar" tag. And it worked with "dog" and "cat" too as single word tags. I made "cat" have "light" as the parent and HABot even recognized "cat" as a light which was really cool.

It seems like HABot might be treating these new tags as synonyms perhaps? It would be interesting to custom build a whole hierarchy and see how far it goes. If I add a tag that has "cat" as a parent, will that tag be recognized as a "cat" and a "light"?

ghys:
It's hard to say, but there is some code to feed any "attributes" available (meaning, tags & synonyms including those by the user) to the trainer to improve the recognition of those entities.

https://github.com/openhab/openhab-webui/blob/6cbc0a021811412d0d86fa0f2a103ae054a0a211/bundles/org.openhab.ui.habot/src/main/java/org/openhab/ui/habot/nlp/internal/OpenNLPInterpreter.java#L135-L156

If OpenNLP is involved at all (remember, it isn't if you just state the label of an item in which case it will merely return a card with the item's members if it's a Group), then any deviation in the training might lead to different results.

The outcome of these tests can be relied upon because the set of training is predetermined, but in practice it isn't because of the code above.

HABot is no ChatGPT, of course, but the basic principle remains the same: the more you train it, the better it becomes.
The balance to find is to give it sufficient training so it is useful, but not too much so it still runs on Raspberry Pis.

J-N-K:
> If those didn't pass, for any reason, then the web UI builds would fail. I would expect any author of a change that has an impact in this area to own up to the regression they introduced and strive to fix it themselves ;)

@jimtng Did you check if web still builds with this changes? I can't test that because as I said above: HABot can't be build on Apple Silicon processors.

@rkoshak @ghys If the build is ok, is there anything that prevents this from being reviewed and merged?

rkoshak:
I'm good. There are still some tests I want to run just to find the limits of what works and what doesn't but I've been able to confirm:

- custom tags do not break HABot
- HABot surprisingly can see and recognize the custom tags
- HABot can see and use the custom tags logically based on the parent tag (e.g. if the new tag uses "Light" as the parent, HABot will recognize that tag as a light)

There should be no surprises for the end users unless they really go deep and create some big new hierarchy of tags (I still don't know what will happen in that case). 

So as far as I'm concerned it's all good and I'm looking forward to adding some custom tags. :-D

jimtng:
> @jimtng Did you check if web still builds with this changes? I can't test that because as I said above: HABot can't be build on Apple Silicon processors.

I'm not sure how to do this, but I've built the entire openhab-core repo and openhab-webui repo on linux on the same machine but I didn't combine them into a running instance. Does that count, or is there a special step to verify? I've never done a full distro build.

J-N-K:
That should work.

lolodomo:
Why do we need a script/rule to extend semantics? Until now, everything was doable through REST APIs. Why not in this case? The choice looks strange and unusable. 
But I probably miss the good reason leading to this solution.

jimtng:
This API is needed as the primary mechanism to create the tags. Using a script is the easiest method to invoke this API.

REST interface to create the tag can be added later on + its corresponding MainUI user interface. Hence the two additional PRs linked to this to lay the ground work towards that:

https://github.com/openhab/openhab-core/pull/3559

https://github.com/openhab/openhab-webui/pull/1850

Once those two are merged, then we could add the support to create the tags from the UI.

lolodomo:
What we need is a REST API to add/remove semantics tags. They need to be stored somewhere to survive a OH restart. 
To be clear, being consistent with everything else in OH.

jimtng:
> What we need is a REST API to add/remove semantics tags. They need to be stored somewhere to survive a OH restart. To be clear, being consistent with everything else in OH.

Yes, I understand and agree. There needs to be a semantic tag provider.

Removing semantic tags without restarting openhab would be great too.

PR #3523, opened 2023-04-04
EventLogger minor refactor

jimtng:
Just a very minor change, but since this gets called a lot, might be worth it.
- Cleaner code
- Avoid doing string concat every time

### DIFF
--- a/bundles/org.openhab.core.io.monitor/src/main/java/org/openhab/core/io/monitor/internal/EventLogger.java
+++ b/bundles/org.openhab.core.io.monitor/src/main/java/org/openhab/core/io/monitor/internal/EventLogger.java
@@ -17,6 +17,7 @@ import java.util.Map;
 import java.util.Set;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.events.Event;
 import org.openhab.core.events.EventSubscriber;
 import org.openhab.core.service.ReadyMarker;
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.io.monitor/src/main/java/org/openhab/core/io/monitor/internal/EventLogger.java
```diff
@@ -72,14 +73,8 @@ public void receive(Event event) {
...
 
-    private Logger getLogger(String eventType) {
-        String loggerName = "openhab.event." + eventType;
-        Logger logger = eventLoggers.get(loggerName);
-        if (logger == null) {
-            logger = LoggerFactory.getLogger(loggerName);
-            eventLoggers.put(loggerName, logger);
-        }
-        return logger;
+    private @Nullable Logger getLogger(String eventType) {
```
J-N-K, line 76: That's not correct. It's `@NonNull`, the checker is unable to detect it due to limitations in ECJ. Please revert that and use `Objects.requireNonNull` around `eventLoggers.computeIfAbsent(...)`.

J-N-K (changes requested):

bundles/org.openhab.core.io.monitor/src/main/java/org/openhab/core/io/monitor/internal/EventLogger.java
jimtng: Thanks! I didn't know how to resolve that issue.

J-N-K (approved):
LGTM, thanks

PR #3526, opened 2023-04-06
Fix thing reloading from things file

### LINKED ISSUE #3527: Things from textual configuration are not properly reloaded
Unfortunately, this issue persists. The file change was noticed all the time and is still noticed (`[el.core.internal.ModelRepositoryImpl] - Loading model 'knx.things'`), but the KNX devices/Things don't update.

I am not also encountering another weird behaviour: my `-control` channels have stopped working, the `handleCommand` method of `DeviceThingHandler` is not called for control channels. I have just checked, with SNAPSHOT 3390 from 31th of March, they work.

_Originally posted by @florian-h05 in https://github.com/openhab/openhab-addons/issues/14534#issuecomment-1497943762_

J-N-K:
Fixes #3527 

It was found that things from textual configuration are not properly updated if changes are only made in configuration or label of a channel. The reason is that for assuming equality only uid and accepted item-type of the channel were checked. This is obviously wrong and also confusing because for the thing itself configuration and other fields are also compared.

Another bug in the `GenericThingProvider`: The things are cached in the provider to determine changes. However, this list was never updated, resulting in changes back to the first version (v1 -> v2 -> v1) not resulting in an update notification, because the new thing and the cached thing are equal.

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/Channel.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/Channel.java
@@ -14,6 +14,7 @@ package org.openhab.core.thing;
 
 import java.util.LinkedHashSet;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
@@ -177,5 +178,26 @@ public class Channel {
 
     public @Nullable AutoUpdatePolicy getAutoUpdatePolicy() {
         return autoUpdatePolicy;
+    }
+
+    @Override
+    public boolean equals(@Nullable Object o) {
+        if (this == o)
+            return true;
+        if (o == null || getClass() != o.getClass())
+            return false;
+        Channel channel = (Channel) o;
+        return Objects.equals(acceptedItemType, channel.acceptedItemType) && kind == channel.kind
+                && Objects.equals(uid, channel.uid) && Objects.equals(channelTypeUID, channel.channelTypeUID)
+                && Objects.equals(label, channel.label) && Objects.equals(description, channel.description)
+                && Objects.equals(configuration, channel.configuration)
+                && Objects.equals(properties, channel.properties) && Objects.equals(defaultTags, channel.defaultTags)
+                && autoUpdatePolicy == channel.autoUpdatePolicy;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(acceptedItemType, kind, uid, channelTypeUID, label, description, configuration, properties,
+                defaultTags, autoUpdatePolicy);
     }
 }
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/Channel.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/Channel.java
@@ -182,10 +182,12 @@ public class Channel {
 
     @Override
     public boolean equals(@Nullable Object o) {
-        if (this == o)
+        if (this == o) {
             return true;
-        if (o == null || getClass() != o.getClass())
+        }
+        if (o == null || getClass() != o.getClass()) {
             return false;
+        }
         Channel channel = (Channel) o;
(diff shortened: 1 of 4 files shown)


florian-h05:
I‘ve tested this, and it is very weird: sometimes when you change channels, the thing is updated, but sometimes the thing is not updated.

J-N-K:
If you enable debug logging for `org.openhab.core.model.thing`, do you see this message in both cases?

```
logger.debug("Updating thing '{}' from model '{}'.", newThing.UID, modelName);
```

florian-h05:
No, `Updating thing 'knx:device:bridge:Kueche_Glastaster' from model 'knx.things'.` is only logged when it works.

`Read things from model 'knx.things'` is always logged.

J-N-K:
Ok, so for some reason the code still thinks they are equals even if they are not. Do you have something that always or at least mostly fails (i.e. editing the label fails, editing the configuration works)? Can you share the `.things` code?

florian-h05:
It seems like it mostly happens when:
1. openHAB starts with a given .things file
2. I change anything of the channel cfg -> Thing updated
3. I change the channel cfg back -> Thing not updated

Here the relavent part of my `*.things` file (I have removed the Things I don't modify):

```
Bridge knx:ip:bridge [  
    type="ROUTER", 
    ipAddress="", // in TUNNEL Mode: 192.168.180.5
    portNumber=3671, // for TUNNEL mode
    localIp="192.168.180.4",
    readingPause=50, 
    responseTimeout=10, 
    readRetriesLimit=4, 
    autoReconnectPeriod=30,
    localSourceAddr="0.0.0",
    useNAT=false
] {

    Thing device Kueche_Glastaster "Küche Glastaster" @ "Küche" [
        address="1.1.24",
        fetch=true,
        pingInterval=300,
        readInterval=0
    ] {
        Type number             : Temperatur                 "Temperatur"              [ ga="9.001:3/3/130" ]
        Type switch             : Verstaerker_Radio          "Verstärker Radio"        [ ga="2/3/150" ]
        Type switch-control     : Verstaerker_Radio_Status   "Verstärker Radio Status" [ ga="1.011:2/3/151"]
        Type switch             : Verstaerker_BT             "Verstärker BT"           [ ga="2/3/155" ]
        Type switch-control     : Verstaerker_BT_Status      "Verstärker BT Status"    [ ga="1.011:2/3/156" ]
        
    }
    
}
```

J-N-K:
I think I found it. However, I'm not 100% if my fix is correct, it's in the XTend code and that is always a mystery for me. Currently testing if it works for me.

J-N-K:
Please check this version, it also contains an updated `org.openhab.core.model.thing` bundle.

florian-h05 (approved):
Works fine, thanks for the fix!

jlaur:
> Another bug in the `GenericThingProvider`: The things are cached in the provider to determine changes. However, this list was never updated, resulting in changes back to the first version (v1 -> v2 -> v1) not resulting in an update notification, because the new thing and the cached thing are equal.

Great finding, thanks for this fix. This is something that has annoyed me for a long time, but never took the time to reproduce or investigate this further. I just had this feeling that it did not work correctly, as it usually happened for me during tests where I was changing configuration back and forth.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/Channel.java
```diff
@@ -178,4 +179,25 @@ public Set<String> getDefaultTags() {
     public @Nullable AutoUpdatePolicy getAutoUpdatePolicy() {
         return autoUpdatePolicy;
     }
+
+    @Override
+    public boolean equals(@Nullable Object o) {
+        if (this == o)
+            return true;
+        if (o == null || getClass() != o.getClass())
+            return false;
```
jlaur, line 188:
```suggestion
        if (this == o)
        {
            return true;
        }
        if (o == null || getClass() != o.getClass())
        {
            return false;
        }
```

J-N-K: Done

wborn (approved):
LGTM! :+1:

PR #3528, opened 2023-04-06
Fix IllegalStateException in CommunityKarafAddonHandler

J-N-K:
This is a regression from #3504: the stream is used as return value in a method but is closed because the try-with-resources was contained in the method itself. The stream is now collected to a list and then returned as a stream.

This results in continuous re-installs for kars from the community market place or JSON 3rd party addon services


kaikreuzer (approved):
lgtm, thanks

PR #3532, opened 2023-04-08
Fix hidden files showing up in TransformationRegistry

J-N-K:
This is especially important because it clutters the config options for profiles that rely on transformation services.


wborn (approved):
Thank you!

PR #3533, opened 2023-04-09
Fix wrong event subscription in GroupStateTrigger and ItemStateTrigger

J-N-K:
The `GroupStateTrigger` subscibed to `ItemStateEvent` instead of `ItemStateUpdatedEvent` (while `ItemStateTrigger` uses `ItemStateUpdatedEvent`), this is confusiing and results in different behavior of both triggers.

The `ItemStateTrigger` did not subscribe to `GroupStateUpdatedEvent` which results in the trigger not firing for group state UPDATES. This is wrong because it fires for item UPDATES and for both, group and item CHANGES.


wborn (approved):
Thanks!

PR #3534, opened 2023-04-09
Add a Range Filter profile

jimtng:
Add a `range-filter` profile to allow/discard values inside a given range.

Example .items file:
```java
// Simple range - allows 0 - 30
Number MyNumber { channel="xxxx" [profile="system:range-filter", range="[0 .. 30]" }
// exclusive boundaries with ( and ) discards 0 and 30
Number MyNumber { channel="xxxx" [profile="system:range-filter", range="(0 .. 30)" }
// exclusive boundaries only on beginning/end: accepts 0, discards 30
Number MyNumber { channel="xxxx" [profile="system:range-filter", range="[0 .. 30)" }

// Supports "inverted" range: discards what's in the given range
Number MyNumber { channel="xxxx" [profile="system:range-filter", range="[0 .. 30]", action="discard" }

// Supports begin-less and end-less range, i.e. for a simple "greater than / less than" filter
//   allow anything up to and including 30
Number MyNumber { channel="xxxx" [profile="system:range-filter", range="[ .. 30]" }
//   allow anything 30 or above
Number MyNumber { channel="xxxx" [profile="system:range-filter", range="[30..]" }

// multiple ranges
Number MyNumber { channel="xxxx" [profile="system:range-filter", range="[0 .. 30], [60..80]" }

// Dimensioned item/channel
Number:Temperature MyTemp { channel="xxxx" [profile="system:range-filter", range="[-50 °C..60 °C]" ] }
// mix units in the range
Number:Temperature MyTemp { channel="xxxx" [profile="system:range-filter", range="[-50 °C..140 °F]" ] }
```

Config in the Main UI:
image

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/profiles/SystemProfileFactory.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/profiles/SystemProfileFactory.java
@@ -67,7 +67,7 @@ public class SystemProfileFactory implements ProfileFactory, ProfileAdvisor, Pro
     private final ChannelTypeRegistry channelTypeRegistry;
 
     private static final Set<ProfileType> SUPPORTED_PROFILE_TYPES = Set.of(DEFAULT_TYPE, FOLLOW_TYPE, HYSTERESIS_TYPE,
-            OFFSET_TYPE, RANGE_TYPE, RAWBUTTON_ON_OFF_SWITCH_TYPE, RAWBUTTON_TOGGLE_PLAYER_TYPE,
+            OFFSET_TYPE, RANGE_TYPE, RANGE_FILTER_TYPE, RAWBUTTON_ON_OFF_SWITCH_TYPE, RAWBUTTON_TOGGLE_PLAYER_TYPE,
             RAWBUTTON_TOGGLE_ROLLERSHUTTER_TYPE, RAWBUTTON_TOGGLE_SWITCH_TYPE, RAWROCKER_DIMMER_TYPE,
             RAWROCKER_NEXT_PREVIOUS_TYPE, RAWROCKER_ON_OFF_TYPE, RAWROCKER_PLAY_PAUSE_TYPE,
             RAWROCKER_REWIND_FASTFORWARD_TYPE, RAWROCKER_STOP_MOVE_TYPE, RAWROCKER_UP_DOWN_TYPE,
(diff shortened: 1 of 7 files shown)


jimtng:
I wonder if other types of filter are needed, for example: outlier filter, moving average filter, etc. If they are, perhaps it would be better to have all these filters in a single "Filter Profile" add-on, and the profile id would be `filter:range`, `filter:outlier`, `filter:moving-average`, etc. instead of `system:range-filter`.

rkoshak:
I suspect if you make it someone somewhere will use it. The `range` as implemented can handle most common use cases. I'm not sure how one would implemented `outlier` except though a `moving-average`, though maybe I'm just not 100% back mentally after being ill all last week.

Thinking through some use cases:

- reject values that are obviously out of range (e.g. temp sensor reporting 5000 °C)
- reject values that are a jump beyond what's reasonable from the last value (e.g. temp sensor jumps from 5 °C to 30 °C)
- reject values that are too small of a change from the last one (smooth out values from a noisy sensor)
- ignore a discrete state always (e.g. maybe someone has a Switch Item linked to a channel that provides ON, OFF, STANDBY and they want to just ignore STANDBY)

It makes a lot of sense to have one profile that can handle all these (and any others I've not thought of.

crnjan:
Will something like https://www.home-assistant.io/integrations/filter/#time-simple-moving-average also be supported? Currently I'm using something like
```
  var item = items.getItem(name);
  var from = time.ZonedDateTime.now().minusMinutes(10);
  var average = item.history.averageSince(from);
```
to smoothen my temperature readings but having a profile to do the job would be imho nicer ...

jimtng:
> Will something like https://www.home-assistant.io/integrations/filter/#time-simple-moving-average also be supported? Currently I'm using something like
> 
> ```
>   var item = items.getItem(name);
>   var from = time.ZonedDateTime.now().minusMinutes(10);
>   var average = item.history.averageSince(from);
> ```
> 
> to smoothen my temperature readings but having a profile to do the job would be imho nicer ...

I think that belongs in a separate profile (to be created) called `smoothing-filter` with the option of various ways of achieving that, e.g. https://en.wikipedia.org/wiki/Moving_average

jimtng:
Closing this because [it needs to go into an addon](https://github.com/openhab/openhab-core/pull/3547#issuecomment-1516756612)

maisun:
This is a very useful profile - has it been merged to the recent release?

jimtng:
No, but check out the basic profiles addon

maisun:
> No, but check out the basic profiles addon

Thanks! Do you mean [this](https://www.openhab.org/addons/transformations/basicprofiles/) add-on? I read the doc but don’t find something that filters value based on defined range. Or do you mean your change will be merged there? Thanks

jimtng:
It was recently added and only available in 4.3 (milestone+)
https://next.openhab.org/addons/transformations/basicprofiles/#state-filter-profile

See the discussion here: https://community.openhab.org/t/state-filter-range-filter-profile/158025

There's one more pending PR which adds even more features:
https://github.com/openhab/openhab-addons/pull/17362

maisun:
Fantastic news! Can’t wait for the 4.3 release and remove my proxy items!

PR #3535, opened 2023-04-09
Improve log messages in ItemStateConditionHandler

### LINKED ISSUE #3370: ItemStateConditionHandler warning logs should include the ruleUID
It can be a challenge sometimes to determine what rule an error or a warning comes from. In this specific case ItemStateConditionHandler does not log out the ruleUID when one of the operands in the condition is a QuantityType and the other is not.

https://github.com/openhab/openhab-core/blob/8be306a5bc794821a15d7e95534efd2a4e135ab0/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ItemStateConditionHandler.java#L172

https://github.com/openhab/openhab-core/blob/8be306a5bc794821a15d7e95534efd2a4e135ab0/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ItemStateConditionHandler.java#L211

https://github.com/openhab/openhab-core/blob/8be306a5bc794821a15d7e95534efd2a4e135ab0/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ItemStateConditionHandler.java#L247

https://github.com/openhab/openhab-core/blob/8be306a5bc794821a15d7e95534efd2a4e135ab0/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ItemStateConditionHandler.java#L290 (maybe?)

The same should probably also be the case for errors:

https://github.com/openhab/openhab-core/blob/8be306a5bc794821a15d7e95534efd2a4e135ab0/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ItemStateConditionHandler.java#L129

https://github.com/openhab/openhab-core/blob/8be306a5bc794821a15d7e95534efd2a4e135ab0/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ItemStateConditionHandler.java#L152

I did a couple of spot checks on a few of the other handlers and they seem to include the ruleID but a thorough review should probably be made. In other words, even though I mention ItemStateConditionHandler, don't feel constrained to only that one class if the ID is missing from other warning and error logs for other rule related handlers.

J-N-K:
Fixes #3370 

This improves the log messages in the `ItemStateConditionHandler` by adding the rule UID. This makes it easier to identify the condition that creates the message.

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ItemStateConditionHandler.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ItemStateConditionHandler.java
@@ -126,12 +126,13 @@ public class ItemStateConditionHandler extends BaseConditionModuleHandler implem
         String state = (String) module.getConfiguration().get(STATE);
         String operator = (String) module.getConfiguration().get(OPERATOR);
         if (operator == null || state == null || itemName == null) {
-            logger.error("Module is not well configured: itemName={}  operator={}  state = {}", itemName, operator,
-                    state);
+            logger.error("Module is not well configured: itemName={}  operator={}  state = {} for rule {}", itemName,
+                    operator, state, ruleUID);
             return false;
         }
         try {
-            logger.debug("ItemStateCondition '{}' checking if {} {} {}", module.getId(), itemName, operator, state);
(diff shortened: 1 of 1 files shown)


wborn (approved):
That will be handy indeed!

PR #3536, opened 2023-04-09
Add group name to GroupStateTriggerHandler

J-N-K:
When triggering on state change/update of group member, the group is not available in the rule context (because the event is the `ItemStateChanged/UpdatedEvent` that caused the group to change/update. This adds a new element `triggeringGroup` to the rule context.

This is useful for generalized rules that trigger on different groups.


kaikreuzer (approved):
lgtm, thanks.

PR #3537, opened 2023-04-10
[oauth] Add support for custom deserialization of AccessTokenResponse

### LINKED ISSUE #1888: [oauth] scope in AccessTokenResponse

jlaur:
This reverts #1891 and fixes #1888 by providing a new way to customize deserialization of `AccessTokenResponse` which has been tested with the Netatmo binding. See openhab/openhab-addons#14755 for a description of the concrete Netatmo problem that so far prevented Netatmo from using the core OAuth implementation.

In this new approach a custom `GsonBuilder` can be provided which can be preconfigured with custom type adapters etc.

This fixes the issue with #1891:
```
[oauth2client.internal.OAuthConnector] - Unable to construct custom deserializer 'org.openhab.binding.netatmo.internal.deserialization.AccessTokenResponseDeserializer'
java.lang.ClassNotFoundException: org.openhab.binding.netatmo.internal.deserialization.AccessTokenResponseDeserializer cannot be found by org.openhab.core.auth.oauth2client_4.0.0.202304080304
```

but also adds more flexibility, for example by allowing multiple type adapters, not only a single fixed deserialization class.

### DIFF
--- a/bundles/org.openhab.core.auth.oauth2client/src/main/java/org/openhab/core/auth/oauth2client/internal/OAuthClientServiceImpl.java
+++ b/bundles/org.openhab.core.auth.oauth2client/src/main/java/org/openhab/core/auth/oauth2client/internal/OAuthClientServiceImpl.java
@@ -35,7 +35,7 @@ import org.openhab.core.io.net.http.HttpClientFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.gson.JsonDeserializer;
+import com.google.gson.GsonBuilder;
 
 /**
  * Implementation of OAuthClientService.
@@ -68,16 +68,19 @@ public class OAuthClientServiceImpl implements OAuthClientService {
     private final String handle;
     private final int tokenExpiresInSeconds;
     private final HttpClientFactory httpClientFactory;
+    private final @Nullable GsonBuilder gsonBuilder;
     private final List<AccessTokenRefreshListener> accessTokenRefreshListeners = new ArrayList<>();
 
     private PersistedParams persistedParams = new PersistedParams();
 
     private volatile boolean closed = false;
 
-    private OAuthClientServiceImpl(String handle, int tokenExpiresInSeconds, HttpClientFactory httpClientFactory) {
+    private OAuthClientServiceImpl(String handle, int tokenExpiresInSeconds, HttpClientFactory httpClientFactory,
+            @Nullable GsonBuilder gsonBuilder) {
         this.handle = handle;
         this.tokenExpiresInSeconds = tokenExpiresInSeconds;
         this.httpClientFactory = httpClientFactory;
+        this.gsonBuilder = gsonBuilder;
     }
 
     /**
@@ -103,7 +106,7 @@ public class OAuthClientServiceImpl implements OAuthClientService {
             return null;
         }
         OAuthClientServiceImpl clientService = new OAuthClientServiceImpl(handle, tokenExpiresInSeconds,
-                httpClientFactory);
+                httpClientFactory, null);
         clientService.storeHandler = storeHandler;
         clientService.persistedParams = persistedParamsFromStore;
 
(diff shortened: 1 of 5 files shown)


bundles/org.openhab.core.auth.oauth2client/src/main/java/org/openhab/core/auth/oauth2client/internal/OAuthClientServiceImpl.java
```diff
@@ -400,10 +413,9 @@ private String createNewState() {
     }
 
     @Override
-    public <T extends JsonDeserializer<?>> OAuthClientService withDeserializer(Class<T> deserializerClass) {
+    public OAuthClientService withGsonBuilder(GsonBuilder gsonBuilder) {
```
jlaur, line 416: This is not the factory pattern, just a method named similarly, but actually creating a new object once again. This could probably be moved to `OAuthFactory`.

bundles/org.openhab.core/src/main/java/org/openhab/core/auth/client/oauth2/OAuthClientService.java
```diff
@@ -291,10 +291,10 @@ AccessTokenResponse getAccessTokenByImplicit(@Nullable String redirectURI, @Null
...
     /**
-     * Adds a personalized deserializer to a given oauth service.
+     * Adds a custom GsonBuilder to be used with the OAuth service instance.
      *
-     * @param deserializeClass the deserializer class that should be used to deserialize AccessTokenResponse
-     * @return the oauth service
+     * @param gsonBuilder the custom GsonBuilder instance
+     * @return the OAuth service
      */
-    <T extends JsonDeserializer<?>> OAuthClientService withDeserializer(Class<T> deserializerClass);
```
jlaur, line 299: There are no usages of this method in the addons repository.

clinique (approved):
LGTM

lolodomo:
@J-N-K @kaikreuzer : there might be not real risk to merge that PR inside milestone 2.
@jlaur : do you think there is any risk to break something?

jlaur:
> do you think there is any risk to break something?

Not really, I'm removing something which is not used and which doesn't work. The new approach was on purpose introduced in the same fashion as the previous one which was already accepted and thought through. The client service objects are cached without the custom builder, which is always added afterwards in the newly created object, so it should not break the main path. It would say it doesn't change anything radically for better or worse.

I have tested the new method with openhab/openhab-addons#14780 and it works as intended. It would be cool to have it, so I can finish that PR the day after when a new snapshot is available.

Anyway, I will leave it to the @openhab/core-maintainers to assess the risk. 🙂

J-N-K:
Why do we use `GsonBuilder` instead of adding multiple `JsonDeserializer`s? E.g. in the `ThingBuilder` we can call `.withChannel` several times and all of them get added.

jlaur:
> Why do we use `GsonBuilder` instead of adding multiple `JsonDeserializer`s? E.g. in the `ThingBuilder` we can call `.withChannel` several times and all of them get added.

Please see the description. The approach of providing a deserializer directly didn't work. I went with the `GsonBuilder` approach to work around that issue, but also to provide a more flexible approach giving the caller much more control since the builder can be freely configured with both serializers, deserializers etc.

J-N-K (approved):

PR #3542, opened 2023-04-12
[ColorUtil] Add fine precision methods

### LINKED ISSUE #3540: ColorUtil proposed extensions
Following are proposed extensions to ColorUtil to enable higher precision conversion XY <=> HSB without intermediate transition via integer RGB values. 

NOTE: this proposal is DRAFT for discussion purposes. It uses blunt clones of some methods to use `double` rather than `int` internally. So it is not code efficient. Efficiency could be improved either by using inner/outer methods, or perhaps generics.

```
    // convert double precision HSB to double precision XY
    public static double[] hsbToXY(HSBType hsbType, Gamut gamut) {
        PercentType[] rgb = hsbToRgbPercent(hsbType);
        double r = inverseCompand(rgb[0].doubleValue() / PercentType.HUNDRED.doubleValue());
        double g = inverseCompand(rgb[1].doubleValue() / PercentType.HUNDRED.doubleValue());
        double b = inverseCompand(rgb[2].doubleValue() / PercentType.HUNDRED.doubleValue());

        double X = r * 0.664511 + g * 0.154324 + b * 0.162028;
        double Y = r * 0.283881 + g * 0.668433 + b * 0.047685;
        double Z = r * 0.000088 + g * 0.072310 + b * 0.986039;

        double sum = X + Y + Z;
        Point p = sum == 0.0 ? new Point() : new Point(X / sum, Y / sum);
        Point q = gamut.closest(p);

        double[] xyY = new double[] { q.x, q.y, Y };

        LOGGER.trace("HSB: {} - XYZ: {} {} {} - xyY: {}", hsbType, X, Y, Z, xyY);

        return xyY;
    }

    // convert double precision XY to double precision HSB
    public static HSBType xyToHsb(double[] xy, Gamut gamut) throws IllegalArgumentException {
        if (xy.length < 2 || xy.length > 3 || !inRange(xy[0]) || !inRange(xy[1])
                || (xy.length == 3 && !inRange(xy[2]))) {
            throw new IllegalArgumentException("xy array only allowes two or three values between 0.0 and 1.0.");
        }
        Point p = gamut.closest(new Point(xy[0], xy[1]));
        double x = p.x;
        double y = p.y == 0.0 ? 0.000001 : p.y;
        double z = 1.0 - x - y;
        double Y = (xy.length == 3) ? xy[2] : 1.0;
        double X = (Y / y) * x;
        double Z = (Y / y) * z;
        double r = X * 1.656492 + Y * -0.354851 + Z * -0.255038;
        double g = X * -0.707196 + Y * 1.655397 + Z * 0.036152;
        double b = X * 0.051713 + Y * -0.121364 + Z * 1.011530;

        // Correction for negative values is missing from Philips' documentation.
        double min = Math.min(r, Math.min(g, b));
        if (min < 0.0) {
            r -= min;
            g -= min;
            b -= min;
        }

        // rescale
        double max = Math.max(r, Math.max(g, b));
        if (max > 1.0) {
            r /= max;
            g /= max;
            b /= max;
        }

        r = compand(r);
        g = compand(g);
        b = compand(b);

        // rescale
        max = Math.max(r, Math.max(g, b));
        if (max > 1.0) {
            r /= max;
            g /= max;
            b /= max;
        }

        HSBType hsb = rgbToHsb(new double[] { 255.0 * r, 255.0 * g, 255.0 * b });
        LOGGER.trace("xy: {} - XYZ: {} {} {} - RGB: {} {} {} - HSB: {} ", xy, X, Y, Z, r, g, b, hsb);

        return hsb;
    }

    // convert double precision RGB to double precision HSB
    public static HSBType rgbToHsb(double[] rgb) throws IllegalArgumentException {
        if (rgb.length != 3 || !inByteRange(rgb[0]) || !inByteRange(rgb[1]) || !inByteRange(rgb[2])) {
            throw new IllegalArgumentException("RGB array only allows values between 0 and 255");
        }
        final double r = rgb[0];
        final double g = rgb[1];
        final double b = rgb[2];

        double max = (r > g) ? r : g;
        if (b > max) {
            max = b;
        }
        double min = (r < g) ? r : g;
        if (b < min) {
            min = b;
        }
        double tmpHue;
        final double tmpBrightness = max / 2.55f;
        final double tmpSaturation = (max != 0 ? (max - min) / max : 0) * 100.0;
        // smallest possible saturation: 0 (max=0 or max-min=0), other value closest to 0 is 100/255 (max=255, min=254)
        // -> avoid float comparision to 0
        // if (tmpSaturation == 0) {
        if (max == 0 || (max - min) == 0) {
            tmpHue = 0;
        } else {
            double red = (max - r) / (max - min);
            double green = (max - g) / (max - min);
            double blue = (max - b) / (max - min);
            if (r == max) {
                tmpHue = blue - green;
            } else if (g == max) {
                tmpHue = 2.0f + red - blue;
            } else {
                tmpHue = 4.0f + green - red;
            }
            tmpHue = tmpHue / 6.0f * 360;
            if (tmpHue < 0) {
                tmpHue = tmpHue + 360.0f;
            }
        }

        return new HSBType(new DecimalType(tmpHue), new PercentType(BigDecimal.valueOf(tmpSaturation)),
                new PercentType(BigDecimal.valueOf(tmpBrightness)));
    }

    // helper method
    private static boolean inByteRange(double val) {
        return val >= 0.0 && val <= 255.0;
    }
```

andrewfg:
Resolves #3540

Add fine precision methods and tests for XY -> RGB -> HSB -> RGB -> XY to the ColorUtil class.


andrewfg:
Ping @holgerfriedrich / @J-N-K for info..

J-N-K:
I'm not convinced that this is the way to go. 

- We already have `int` and `PercentType`, we should not add `double` as third option. 
In my opinion you should  use `PercentType` instead and internally `BigDecimal`. I already had a look at that yesterday and it's possible to clean-up the RGB/HSB conversion quite a bit by doing so: https://github.com/J-N-K/openhab-core/commit/39127560c9cb5728460ac84548211e8aa225676c. If necessary (e.g. for re-use in the XY methods), you could also refactor it to a private method that accepts `BigDecimal[]` (or `BigDecimal, BigDecimal, BigDecimal`.

- Truncating values from the HSB/XY conversions should just be removed instead of duplicating the methods. We should try to use the best precision that is useful. The only exception (as I already stated in the issue) is the RGB-`int[]` method because when using 0-255 in almost any case you want to use a byte/three bytes.

andrewfg:
> use PercentType instead and internally BigDecimal. I already had a look at that yesterday and it's possible to clean-up the RGB/HSB conversion quite a bit by doing so: https://github.com/J-N-K/openhab-core/commit/39127560c9cb5728460ac84548211e8aa225676c

Yes that looks good.

> The only exception is the RGB-int[] method because when using 0-255 in almost any case you want to use a byte/three bytes

I am not convinced. One might want better RGB precision than 1 part in 255, so for example a real RGB value of `254.5` being half way between 254 and 255. (I think this was what the OP of the original issue was asking for). Your are 'half' doing this in `convertByteToColorPercent()`, so I don't see why you would not go the full way and changing `rgbToHsb(int[] rgb)` to `rgbToHsb(double[] rgb)` or `rgbToHsb(BigDecimal[] rgb)` .. ??

J-N-K:
I really don't see a use-case for that. In what application would you need a precision higher than 0.25 % for RGB values? And if there really is a use-case for that, you can always use the method that returns `PercentType[]` which has `BigDecimal` precision internally. OTOH we have a lot of use-cases for 0-255 `int[]`, because that is what most external system expect when RGB is used.

andrewfg:
```
public static HSBType xyToHsb(double[] xy, Gamut gamut) throws IllegalArgumentException {
..
// I would change this..
    HSBType hsb = rgbToHsb(new int[] { (int) Math.round(255.0 * r), (int) Math.round(255.0 * g), (int) Math.round(255.0 * b) });

// to this..
    PercentType[] rgbParameters = new PercentType[] { //
        new PercentType(BigDecimal.valueOf(r * 255.0 / 100.0)),
        new PercentType(BigDecimal.valueOf(r * 255.0 / 100.0)),
        new PercentType(BigDecimal.valueOf(r * 255.0 / 100.0)) };
    HSBType hsb = rgbToHsb(rgbParameters);

    return hsb;
```

andrewfg:
^
I will commit a new proposal..

J-N-K:
I made a test with more precision and disabled rounding (see https://github.com/J-N-K/openhab-core/commit/201a868c31dffab71e84c3779971e98e5d812ab0) and the issue is the round-trip conversion for "white" (and IIRC that is also why I introduced the rounding):

```
[main] ERROR org.openhab.core.util.ColorUtil - HSB: 0,0,100 - RGB: [100, 100.0000000000, 100] - XYZ: 0.980863 0.9999990000000001 1.058437 - xyY: [0.32272672086556803, 0.32902290955907926, 0.9999990000000001]
[main] ERROR org.openhab.core.util.ColorUtil - xy: [0.32272672086556803, 0.32902290955907926, 0.9999990000000001] - XYZ: 0.9808630000000002 0.9999990000000001 1.058437 - RGB: 0.9999996235179819 0.9999988133819798 0.9999999999999999 - HSB: 280.96361196495535,0.0000118661802010000000,99.99999999999999 
```

As I already said in the issue: hue as no meaning, when saturation is zero. So (0,0,100) is the same color as (280.96361196495535,0.000118661802010000000,99.99999999999999). The reason for the difference is that the RGB values are not all the same (which is just a result of rounding errors). If you apply rounding (like we did before), then they all will be exactly 1 and not RGB: 0.9999996235179819 0.9999988133819798 0.9999999999999999. 

In real world application that does not matter. No-one can see the difference between magenta with a saturation of 0.001% and white. For math comparison it does. I don't know of any formula that can convert between xy and HSB without any rounding issue. Maybe the situation improves bit when we use `BigDecimal` also in the xy methods, but I doubt it'll be sufficient to eliminate the test failure.

andrewfg:
> the issue is the round-trip conversion for "white" 

Yes I saw that in the Junit tests.

holgerfriedrich:
@andrewfg could you also pick the changes to ColorUtilTest.java I did in https://github.com/openhab/openhab-core/pull/3546/commits/b12d5bf0a07a41373c946bf32a9efa543b791055 ?

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -63,7 +63,7 @@ public static int[] hsbToRgb(HSBType hsb) {
      * Transform <a href="https://en.wikipedia.org/wiki/HSL_and_HSV">HSV</a> based {@link HSBType} to
      * <a href="https://en.wikipedia.org/wiki/SRGB">sRGB</a>.
      *
-     * This function not round the components to integer values. Please consider using {@link #hsbToRgb(HSBType)}
+     * This function does not round the components to integer values. Please consider using {@link #hsbToRgb(HSBType)}
      * whenever integer values are required.
```
J-N-K: Maybe this should be rewritten to something like `This method does not round the components. For conversion to integer values in the range of 0-255 use ...`,

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -182,20 +183,17 @@ public static double[] hsbToXY(HSBType hsbType, Gamut gamut) {
         double[] xyY = new double[] { ((int) (q.x * 10000.0)) / 10000.0, ((int) (q.y * 10000.0)) / 10000.0,
                 ((int) (Y * 10000.0)) / 10000.0 };
 
-        LOGGER.trace("HSB: {} - RGB: {} - XYZ: {} {} {} - xyY: {}", hsbType, hsbType.toRGB(), X, Y, Z, xyY);
-
+        LOGGER.trace("HSB: {} - XYZ: {} {} {} - xyY: {}", hsbType, X, Y, Z, xyY);
```
J-N-K, line 186: Why did you remove the intermediate RGB step from the logging?

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -205,6 +203,14 @@ public static HSBType rgbToHsb(int[] rgb) throws IllegalArgumentException {
                 convertByteToColorPercent(rgb[2]) });
     }
 
+    /**
+     * Transform <a href="https://en.wikipedia.org/wiki/SRGB">sRGB</a> color format to
+     * <a href="https://en.wikipedia.org/wiki/HSL_and_HSV">HSV</a> based {@link HSBType}.
+     *
+     * @param rgb PercentType[] array of length 3, comprising integer or real values, constrained to 0-100.
```
J-N-K, line 210:
```suggestion
     * @param rgb {@link PercentType} array of length 3
```

`PercentType` is always 0-100, you can't set any other values.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -285,7 +292,18 @@ public static HSBType xyToHsb(double[] xy, Gamut gamut) throws IllegalArgumentEx
...
+     * See <a href=
+     * "https://developers.meethue.com/develop/application-design-guidance/color-conversion-formulas-rgb-to-xy-and-back/">Hue
+     * developer portal</a>.
+     *
+     * @param xy the CIE 1931 xy color, x,y[,Y] between 0.0000 and 1.0000. <code>Y</code> value is optional.
+     * @param gamut the gamut supported by the light.
+     * @return the corresponding {@link HSBType} created from non rounded RGB values.
+     * @throws IllegalArgumentException when input array has wrong size or exceeds allowed value range.
      */
     public static HSBType xyToHsbFine(double[] xy, Gamut gamut) throws IllegalArgumentException {
```
J-N-K, line 308: I don't think we should do that. Why don't we always use the exact calculation (i.e. the one you call `fine`?

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -296,9 +314,19 @@ public static HSBType xyToHsbFine(double[] xy, Gamut gamut) throws IllegalArgume
...
+     * "https://developers.meethue.com/develop/application-design-guidance/color-conversion-formulas-rgb-to-xy-and-back/">Hue
+     * developer portal</a>.
+     *
+     * @param xy the CIE 1931 xy color, x,y[,Y] between 0.0000 and 1.0000. <code>Y</code> value is optional.
+     * @param gamut the gamut supported by the light.
+     * @return an array of doubles containing the RGB elements.
+     * @throws IllegalArgumentException when input array has wrong size or exceeds allowed value range.
      */
-    private static double[] xyToRgb(double[] xy, Gamut gamut) throws IllegalArgumentException {
+    public static double[] xyToRgb(double[] xy, Gamut gamut) throws IllegalArgumentException {
```
J-N-K, line 329: Please don't introduce a third way of representing RGB. We have `int[]` (for rounded 0-255) and `PercentType[]` for 0-100 with `BigDecimal` precision. I don't see any benefit in cluttering API by adding a `double[]`.

J-N-K (changes requested):

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
andrewfg: Because the method is deprecated

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
andrewfg: Ok.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
andrewfg: Ok

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
andrewfg: > Why don't we always use the exact calculation

Umm. I thought it was you insisting on non exact. Or??

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
andrewfg: If you allow the 'fine' method to exist, then this method can be private. But if you want 'fine' to be removed, then making this public allows users to do the 'fine' themselves externally.

J-N-K: Why? I don't see a use-case for `xyToRgb` on the openHAB context, but if you absolutely want such a method, it should return `PercentType[]` like the `hsbToRgbPercent`, which also returns the "exact" values.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
J-N-K: No, that must be a mis-understanding. I want a method that returns 0..255 as `int` because that is a very common use-case. And I don't want hundreds of methods that essentially return the same values, just with different types. For precise values we can use `PercentType` and whoever wants a double can call `rgb[0].doubleValue()`.

That aside: I don't see a real-world use-case for color values with a precision better than what we had before. openHAB is about home automation and I doubt that there is a single device out there that can reliably produce light with an RGB color of 255,127,98 in a way that a human eye can see a difference to 254,127,98. I also don't see why it is important that a conversion "xyY - something - xyY" results in the same values as long as the color that is associated with these values is so similar that a human eye can't detect a difference. I'm willing to accept such code only because it does not add additional processing.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
J-N-K: `HSBType.toRGB` is deprecated, `ColorUtil.hsbToRgbPercent` is not. And please put that in a `if (LOGGER.isTraceEnabled()) { ... }` block, `hsbToRgbPercent` is quite complex and we don't need to calculate it all the time if we don't log it (that was already a problem before, it's not something new).

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
andrewfg: @J-N-K I am sorry but we seem to be dancing in circles. So let me summarise below the method signatures that you do, resp. do not want in ColorUtil. => Is this correct?

### Wanted methods
```
public static HSBType xyToHsb(double[] xy) // see question 1. below
public static HSBType xyToHsb(double[] xy, Gamut gamut) // see question 1. below

public static int[] hsbToRgb(HSBType hsb)
public static PercentType[] hsbToRgbPercent(HSBType hsb) // see question 2. below
public static int hsbTosRgb(HSBType hsb)

public static double[] hsbToXY(HSBType hsb)
public static double[] hsbToXY(HSBType hsb, Gamut gamut)

public static HSBType rgbToHsb(int[] rgb) // see questions 1. and 3. below
public static HSBType rgbToHsb(PercentType[] rgb) // see questions 1. and 3. below
```
### Unwanted methods
```
private static HSBType xyToHsbFine(double[] xy, Gamut gamut)
private static double[] xyToRgb(double[] xy, Gamut gamut)
```
### Qustions
And if the above is the case, then I have two further questions for clarification..
1. Do you accept that it is OK for `blahblahToHsb()` to produce an HSB with non integer internal parameters?
2. Should we rename `hsbToRgbPercent()` as `hsbToRgb()` to be internally self consistent?
3. Or, alternatively should we rename `rgbToHsb()` as `rgbToHsbPercent()` or `rgbPercentToHsb()` to be internally self consistent?

J-N-K: 1. Sure. Why should there be a limitation on the internal representation in `HSBType`?
2. That would be nice, but it's not possible, because we can't have to methods with the same name and parameters that return different types.
3. Like you, I'm not so happy with the naming of `hsbToRgbPercent` and `hsbTosRgb` but I think we have to live with that because of the limitations of Java. I would prefer to keep the shorter names, they are more intuitive.

andrewfg: Ok.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
andrewfg: Ok.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
andrewfg: But, I don't think there is any reason why the Java doc should not say that.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
andrewfg: Ok.

andrewfg:
> could you also pick the changes to ColorUtilTest.java I did in https://github.com/openhab/openhab-core/commit/b12d5bf0a07a41373c946bf32a9efa543b791055

Yes I can do that. And `inversionTest(HSBType hsb)` will 'cop out' if hue becomes meaningless at zero saturation.

holgerfriedrich:
@andrewfg thank you, overall the change looks good to me.
It eliminates the unnecessary errors during conversion, and we can see the round-trip without differences.
It will likely fix openhab/openhab-addons#12643, as we have a lossless RGB(int)->HSB(float)->RGB(int) conversion.


One thing still concerns me a bit: efficiency. Couldn't we restrict the BigDecimal to a reasonable number of scales?
Especially for the int valued RGB -> HSB -> int valued RGB above, the full accuracy is not necessary. But maybe I am a bit to pessimistic - those color conversions will not be done too often and computing power is cheap.

holgerfriedrich (approved):

andrewfg:
> Couldn't we restrict the BigDecimal to a reasonable number of scales?

I don't really know enough about BigDecimal to be able to comment.

J-N-K (approved):
That was difficult :-) Thanks!.

J-N-K:
Unfortunately openhab-addons does not build after these changes:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.0.0-M7:test (default-test) on project org.openhab.binding.knx: There are test failures.
[ERROR] 
[ERROR] Please refer to /Users/jan/Documents/git/openhab-addons/bundles/org.openhab.binding.knx/target/surefire-reports for the individual test results.
[ERROR] Please refer to dump files (if any exist) [date].dump, [date]-jvmRun[N].dump and [date].dumpstream.
[ERROR] -> [Help 1]
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.0.0-M7:test (default-test) on project org.openhab.binding.tradfri: There are test failures.
[ERROR] 
[ERROR] Please refer to /Users/jan/Documents/git/openhab-addons/bundles/org.openhab.binding.tradfri/target/surefire-reports for the individual test results.
[ERROR] Please refer to dump files (if any exist) [date].dump, [date]-jvmRun[N].dump and [date].dumpstream.
[ERROR] -> [Help 1]
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.0.0-M7:test (default-test) on project org.openhab.binding.webthing: There are test failures.
[ERROR] 
[ERROR] Please refer to /Users/jan/Documents/git/openhab-addons/bundles/org.openhab.binding.webthing/target/surefire-reports for the individual test results.
[ERROR] Please refer to dump files (if any exist) [date].dump, [date]-jvmRun[N].dump and [date].dumpstream.
[ERROR] -> [Help 1]
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.0.0-M7:test (default-test) on project org.openhab.binding.mqtt.generic: There are test failures.
[ERROR] 
[ERROR] Please refer to /Users/jan/Documents/git/openhab-addons/bundles/org.openhab.binding.mqtt.generic/target/surefire-reports for the individual test results.
[ERROR] Please refer to dump files (if any exist) [date].dump, [date]-jvmRun[N].dump and [date].dumpstream.
```

Can you have a look? Thanks.

holgerfriedrich:
@J-N-K I will have a look....

holgerfriedrich:
@J-N-K I will have a look....

holgerfriedrich:
Issue is resolved by openhab/openhab-addons#14808

holgerfriedrich:
@J-N-K regarding failing tests in addons:

```
        HSBType t = ColorUtil.xyToHsb(new double[] {0.3f, 0.6f});
        c.processMessage("state", "0.3,0.6,100".getBytes());
        assertThat(value.getChannelState(), is(t)); // HSB
```

This one fails now in mqtt.generic. Can you give a hint how to fix this?

```
Expected: is <91.77373046366659,73.7842616508531700,99.99999999999999>
     but: was <91.77373046366659,73.7842616508531700,100>

```

andrewfg:
@holgerfriedrich if I understand correctly there were four bindings whose tests failed due to the extra precision - namely `knx`, `tradfri`, `mqtt` and `webthing` whereby the former is resolved by #14808 and the latter three by #14810. Is that correct? Many thanks for those fixes. And/or do you need anything more from me on this?

holgerfriedrich:
@andrewfg correct. openhab/openhab-addons#14810 works but still needs some improvement (I agree to the comments jnk made). I don't know when I can spent time, so in case you want take over.....

andrewfg:
>  in case you want take over..

@holgerfriedrich here you go https://github.com/holgerfriedrich/openhab-addons/pull/3

holgerfriedrich:
@J-N-K ready to merge? will it get into M2?

J-N-K:
No, I see the risk that it breaks something because something now has decimals that didn‘t have it before. Overall code is pretty stable at the moment, I don’t want I put that at risk. I’ll merge immediately after M2.

andrewfg:
Apropos the discussion about precision, below is typical JSON verbatim from the Philips Hue API v2. So we may conclude that the Hue Bridge does process to 4 places of decimals on XY (1 part in 10000), resp. 2 places of decimals on Brightness (also 1 part in 10000). Although we assume obviously that the actual light bulbs probably only see a lesser precision. As mentioned earlier, I don't know the inner workings of BigDecimal, so I don't know what is the current precision level in the current PR code.  But if it is actually higher than that, and if you so wish, we could formally reduce it now to 1 part in 10000 in this PR at this time. => WDYT?

```
                {
                    "color": {
                        "xy": {
                            "x": 0.4731,
                            "y": 0.3723
                        }
                    },
                    "dimming": {
                        "brightness": 44.09
                    }
                },
```

J-N-K:
This BTW is the reason why I postponed this PR. I'm not sure how `hue` (and `deconz` and probably others) format the values when they output them. This is also the reason why we has the `((int) (value*10000))/10000.0` code: before it was part of core, the conversion was part of `deconz` which - like `hue` uses 4 decimal places.

holgerfriedrich:
I am a bit concerned about events.log.... Could be unreadable due to all the decimals.

andrewfg:
> This BTW is the reason why I postponed this

@J-N-K so shall I try to modify the PR to consume/produce values at 1 in 10k precision?

PS I think the binding Junit test changes should survive such a change because they mostly now check 'closeTo' 1% -- which is not really testing the **accuracy** of the conversion (the accuracy tests should be part of core) but rather testing its **plausibility**..

J-N-K:
@andrewfg It's difficult to say. IMO truncating to `int` is too less precision, but a 1/10000 would probably more than sufficient, so probably using the same precision that hue does is a good compromise.

J-N-K:
@andrewfg Did you see my last comment? Shall we merge as-is or are you working on improvements?

andrewfg:
> .. or are you working on improvements?

I have been a bit busy. Can you hold on for a couple of days?

J-N-K:
No need to hurry. If it takes some time, that's fine. Just ping me when you are ready, simple pushed don't create a notification and I might miss your updates.

andrewfg:
@J-N-K I just pushed a commit with a new `MathContext` as you suggested. Please note that the methods `hsbToXY()` and `xyToHsb()` are still using `double`s internally, so I was wondering if we should change those to `BIGDECIMAL` as well? I guess probably not..

J-N-K (approved):
LGTM, thanks.

J-N-K:
DCO probably fails because of the merge commit.

J-N-K:
Regarding the double/BigDecimal: If necessary we can refactor that later. We should get this merged now. I confirmed the tests in openhab-addons are also ok with the corrections from the PR. I'll wait for CI build and merge then.

PR #3546, opened 2023-04-12
ColorUtil: rgbToHsb precision

holgerfriedrich:
```rgbToHsb``` no longer rounds resulting HSB values to next integer.
This allows lossless conversion from integer valued RGB representation
to ```HSBType``` and back integer valued RGB.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
@@ -50,7 +50,7 @@ public class ColorUtil {
      *
      * This function does rounding to integer valued components. It is the preferred way of doing HSB to RGB conversion.
      *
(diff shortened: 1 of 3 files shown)


holgerfriedrich:
@J-N-K @andrewfg
When I started, the goal was to fix deviations seen in the round tip RGB(int) -> HSB (double/decimal) -> RGB(int). This is possible if we remove the unnecessary rounding in ```rgbToHsb```. Unit tests are fine and due to the integer representation we have no issues even if we cut the BigDecimals to 1 or 2 scales. This might be better to read when the HSB value is printed.

Maybe before rewriting xy conversion and introducing higher precision conversions, let us do this simple fix the remove the unnecessary rounding to integer valued HSB components. When I took over this function from smarthome/j, my understanding was that it is required to have integer valued components. Sorry for that.

holgerfriedrich:
obsoleted by current state of #3540

PR #3547, opened 2023-04-13
Add multiplier option to offset profile

### LINKED ISSUE #3538: [profiles] Add existing profile "Gain-Offset" from ModBus binding to core to allow value scaling
When it is necessary to adjust incoming or outgoing values it is currently possible to define an offset on the original value using profile "Offset".

Unfortunately some devices deliver float values scaled to integers, e.g. ModBus energy meters do so.
For that reason there is a [profile contained in the ModBus binding called "Gain-Offset"](https://next.openhab.org/addons/bindings/modbus/#profiles) that allows not only an offset but to provide a gain factor plus adding a UoM as well and works on incoming and outbound numeric values.

Clear advantage: No need to involve scripts to scale the original value, complete setup is possible from the UI.

See [this community post](https://community.openhab.org/t/item-number-change-decimal-point/145873/7) for more details on how this would simplify to end users life.

Either the existing profile "Offset" should be enhanced with a gain factor or it could even be superseded by the "Gain-Offset" profile.

OH Version 3.4, similar in OH 4 M1.
OS version independent.

jimtng:
Resolve #3538

Add multiplier option to offset profile so people don't need to install the modbus addon just to get such a profile.

There are some differences compared to the modbus `gain-offset` profile, primarily:
- modbus gain-offset profile applies its offset parameter before calculating the gain (pre-gain offset), i.e. output = (input + offset) * gain
- This PR calculates it in the convention of a linear equation y = mx + b, so the offset is added after applying the gain (post-gain offset).
- The multiplier is a simple decimal type. No unit injection is done by this profile.

This PR should maintain backwards compatibility against existing users who only configured the `offset` parameter. In such circumstances, the multiplier defaults to `1` (no gain/attenuation). However:
- Incompatible units between the offset and the incoming value, previously resulted in an `UNDEF` being passed on. This is changed so nothing gets passed on.

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/profiles/SystemOffsetProfile.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/profiles/SystemOffsetProfile.java
@@ -13,12 +13,11 @@
 package org.openhab.core.thing.internal.profiles;
 
 import java.math.BigDecimal;
-
-import javax.measure.UnconvertibleException;
-import javax.measure.quantity.Temperature;
+import java.util.Optional;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
+import org.openhab.core.config.core.ConfigParser;
 import org.openhab.core.library.types.DecimalType;
 import org.openhab.core.library.types.QuantityType;
 import org.openhab.core.library.unit.Units;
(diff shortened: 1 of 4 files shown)


clinique (approved):
LGTM. Maybe worth mentionning that this PR can also be used to apply a unit to an item coming from a channel without unit.

jimtng:
> LGTM. Maybe worth mentionning that this PR can also be used to apply a unit to an item coming from a channel without unit.

A comprehensive list of what it does / doesn't do will go into the documentation PR, if this PR is accepted.

jimtng:
@Oggerschummer apart from the difference in how the offset is calculated, would this fulfil your requirement? Do you use unit injection with the gain-offset profile currently?

Oggerschummer:
> @Oggerschummer apart from the difference in how the offset is calculated, would this fulfil your requirement? Do you use unit injection with the gain-offset profile currently?

Yes, I use the unit injection as ModBus only knows pure vaues but no units.

Looks good besides one point.
From handler to item, case 2:
Wouldn't it make more sense to take the multipliers unit here?
Because if the user scales by the gain factor the unit changes. E.g. incoming 1000 W, gain 0.001, multiplier unit "kW"
I'd expect the output being 1 kW, not 1 W.
In the end the user should be able to define the resulting unit as the profile itself can only guess in that case.

jimtng:
> Because if the user scales by the gain factor the unit changes. E.g. incoming 1000 W, gain 0.001, multiplier unit "kW"
> I'd expect the output being 1 kW, not 1 W.

If modbus doesn't give you units, would this apply to you still?

Technically 1000W * 0.001kW = 1 W*kW.

We can't just ignore the input value's unit. Example:

If incoming is 1000W gain 0.001 kW - if you want the output to be 1 kW (by multiplying 1000 * 0.001, dropping input's unit, and adopting multiplier's unit), what if it gets an input of 1 kW? Will it produce 0.001 kW? Or do we say, all input values will first be converted to their "base unit" (which is W in this case) first? What about Temperature, what's the base unit of C vs F vs K? 

This is why when encountering input value with unit + multiplier with unit, it will issue a warning because the result may not be what you expect.

Now, I understand that you want something that converts to a known unit, e.g. whatever comes in, you want the output to be in `kW`. Perhaps this can be done using an additional parameter `force_unit="kW"` - which will only work if the incoming unit is convertible to it. So with this:
- an input of 1000W, multiplier 1, output: 1kW
- an input of 1kW, multiplier 1, output: 1kW

That would make much more sense than using the multiplier's unit.

spacemanspiff2007:
With the new UoM concept changing unit and scale should be trivial.
In the example the only thing that has to be done is attach `W` as a unit to the unit less value from the binding via profile.
The item then has to be defined with unit=`kW`. 
The value `1000W` will then automatically result in `1kW` on the item.

Imho setting/stripping units should be done through a separate profile and this profile should keep the unit (if available) as is.

jimtng:
> Imho setting/stripping units should be done through a separate profile and this profile should keep the unit (if available) as is.

I totally agree with this. Changing units is awkward and attaching it to the multiplier is wrong.



> With the new UoM concept changing unit and scale should be trivial.

Is that going to be adopted / merged?

spacemanspiff2007:
> Is that going to be adopted / merged?

It's scheduled for 4.0

jimtng:
In that case, I'll gladly remove the unit injection from here.

Oggerschummer:
> In that case, I'll gladly remove the unit injection from here.

Agreed, if this is part of 4.0 we should drop changing units here and stick to a simple value adjustment.
Much cleaner approach then.

jimtng:
I've removed the unit from the multiplier. It is now just a plain multiplier.

J-N-K:
I don't question that this extension (and also #3534) is useful, I just wonder if it would be better to create a new add-on for profiles, so that they are optional. The reason is that with increasing profile count the list will get confusing and it might be a good idea to have the ability to chose. This also already applies to at least the offset-profile we have now. I don't think that we would need a separate bundle for each profile, but a group of connected profiles (e.g. one for mathematical manipulation with offset/round/invert, one for limiting with debounce/coerce/threshold, one for type conversion with open-closed-2-on-off/ play-pause-2-on-off, ...).

jimtng:
> I just wonder if it would be better to create a new add-on for profiles, so that they are optional.

I was thinking about this too but not sure which way is better. I'm in favour of putting them into addons.

So in the case of the offset profile, should we remove it from core? That would cause a breaking change.

PR #3549, opened 2023-04-13
Fix IndexOutOfBoundsException in ScriptEngineFactoryHelper

jimtng:
https://community.openhab.org/t/openhab-4-0-snapshot-discussion/142322/319

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineFactoryHelper.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineFactoryHelper.java
@@ -61,7 +61,8 @@ public class ScriptEngineFactoryHelper {
     public static String getPreferredMimeType(ScriptEngineFactory factory) {
         List<String> scriptTypes = factory.getScriptTypes();
         if (scriptTypes.isEmpty()) {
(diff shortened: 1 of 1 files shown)


J-N-K:
I don't think this is the right way. IMO it's a programming error if we run into that situation and this should not be hidden. I would prefer to throw an `IllegalStateException`.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineFactoryHelper.java
```diff
@@ -59,9 +59,13 @@ private ScriptEngineFactoryHelper() {
     }
 
     public static String getPreferredMimeType(ScriptEngineFactory factory) {
-        List<String> mimeTypes = new ArrayList<>(factory.getScriptTypes());
+        List<String> scriptTypes = factory.getScriptTypes();
+        if (scriptTypes.isEmpty()) {
+            throw new IllegalStateException(factory.getClass().getName() + " does not support any scriptTypes.");
```
J-N-K, line 64:
```suggestion
            throw new IllegalStateException(factory.getClass().getName() + " does not support any scriptTypes. Please report it as a bug.");
```

jimtng: added

J-N-K (approved):
Thanks

PR #3550, opened 2023-04-13
Remove GenericScriptEngineFactory

jimtng:
It seems that this is not used. It returns an empty list in getScriptTypes() that caused a problem (like in #3549) sometimes due to assumption that it should always returns something.


J-N-K (approved):
LGTM. I would like to wait two weeks (i.e. until 04/28/23) if someone steps up that would like to keep it.

wborn (approved):

PR #3553, opened 2023-04-15
Add methods to get Semantic tag synonyms and description

jimtng:
Add methods for getting the label and description of a semantic tag.

In addition to enriching the ability to work with semantic tags, this should also help in building a rest endpoint that provides the semantic tags to mainui, as discussed here: https://github.com/openhab/openhab-webui/issues/1822

Although at this point I'm not yet sure exactly what's needed for the rest endpoint. More methods would probably be needed when we tackle the implementation of that feature.


J-N-K (approved):
LGTM, Thanks.

PR #3559, opened 2023-04-16
[REST] List semantic tags

jimtng:
This is to enable the use of dynamically added tags (#3519) in the main ui (https://github.com/openhab/openhab-webui/pull/1850)

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/tag/TagResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/tag/TagResource.java
@@ -64,16 +64,16 @@ import io.swagger.v3.oas.annotations.responses.ApiResponse;
  */
 @Component
 @JaxrsResource
-@JaxrsName(TagResource.PATH_ITEMS)
+@JaxrsName(TagResource.PATH_TAGS)
 @JaxrsApplicationSelect("(" + JaxrsWhiteboardConstants.JAX_RS_NAME + "=" + RESTConstants.JAX_RS_NAME + ")")
 @JSONRequired
-@Path(TagResource.PATH_ITEMS)
-@io.swagger.v3.oas.annotations.tags.Tag(name = TagResource.PATH_ITEMS)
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/tag/TagResource.java
```diff
@@ -0,0 +1,106 @@
...
+@JaxrsName(TagResource.PATH_ITEMS)
+@JaxrsApplicationSelect("(" + JaxrsWhiteboardConstants.JAX_RS_NAME + "=" + RESTConstants.JAX_RS_NAME + ")")
+@JSONRequired
+@Path(TagResource.PATH_ITEMS)
+@io.swagger.v3.oas.annotations.tags.Tag(name = TagResource.PATH_ITEMS)
+@NonNullByDefault
+public class TagResource implements RESTResource {
+
+    /** The URI path to this resource */
+    public static final String PATH_ITEMS = "tags";
```
kaikreuzer, line 76: Shouldn't this constant rather be called `PATH_TAGS`?

kaikreuzer (changes requested):

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/tag/TagResource.java
jimtng: oopsie! :) fixed

jimtng: Should it be `/semantics` instead of `/tags` ?

kaikreuzer: No, the name is [`tag`](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core/src/main/java/org/openhab/core/items/Item.java#L116), so that's ok.

kaikreuzer (approved):
lgtm

PR #3362, opened 2023-02-08
Remove author from addon xsd

kaikreuzer:
See https://github.com/openhab/openhab-core/pull/3050#discussion_r1071603658

Signed-off-by: Kai Kreuzer <kai@openhab.org>

PR #3563, opened 2023-04-18
Redefine connection in addon.xsd

### LINKED ISSUE #3555: Refine addon.xml `connection`
Discussion extracted from openhab/openhab-addons#14813

@J-N-K posted:
> If cloud is used for discovery it's `cloudDiscovery`, if it accesses the cloud (or a remote service) during normal operation, it's `cloud` and otherwise it's `local`.

Sorry if the following can be perceived as pedantic. My intention is only to try to get this right, now that we have this box opened.

The original issue definition by @kaikreuzer (openhab/openhab-core#2058):
- `local` (for add-ons that) interacts locally
- `cloud` if it connects to an online service (seldom or continuously)

The [documented](https://next.openhab.org/docs/developer/addons/addon.html#xml-structure-for-add-on-definitions) definition (highlighting important words) by you (openhab/openhab-docs#1872):
- `local` for add-ons that **only** interact locally
- `cloud` for add-ons that **require** a cloud connection
- `cloudDiscovery` for add-ons that **require** a cloud connection for set-up

And what you just wrote:
- `cloudDiscovery` if cloud is used for discovery
- `cloud` if it accesses the cloud (or a remote service) during **normal operation**
- `local` otherwise (i.e. if it doesn't access the cloud during **normal operation**)

There are slight contradictions here, even between your own definitions. If it doesn't **require** cloud, but **accesses** cloud during **normal operation**, then according to your first definition it's inconclusive - it doesn't **require** cloud, but it also doesn't interact locally **only**. According to your new definition it would be `cloud`, even if it can run perfectly well locally without cloud.

For me as a user, the questions I would expect immediate answers for are:
- Can it run locally without being dependent on an internet connection or any external service?
- Even if it can, will it still call some external services, for example to provide extended functionality?

We cannot clearly answer that with the current options.

I would propose the following (explanations, not descriptions to be used):
- `none` if it runs locally in the openHAB instance without any network dependencies, even local.
- `local` if it runs entirely locally (LAN) and will never call external services.
- `hybrid` is it runs locally, but can optionally also use external (cloud) services for added functionality. This includes what is now called "cloud discovery" which cannot also be seen as "added functionality".
- `cloud` if entirely dependent on external cloud services.

_Originally posted by @jlaur in https://github.com/openhab/openhab-addons/issues/14813#issuecomment-1509734325_

jlaur:
Redefines `connection` in addon.xsd by:
- Adding new member `none` for no interactions with external systems at all (e.g. astro)
- Replacing member `cloudDiscovery` by `hybrid` which is more generic and can be used when internet access is not required, but can provide optional extended functionality (for example discovery).

Also includes a commit a very small cleanup after #3362.

Resolves #3555

### DIFF
--- a/bundles/org.openhab.core.addon/schema/addon-1.0.0.xsd
+++ b/bundles/org.openhab.core.addon/schema/addon-1.0.0.xsd
@@ -51,9 +51,26 @@
 
     <xs:simpleType name="connectionType">
         <xs:restriction base="xs:string">
-            <xs:enumeration value="local"/>
-            <xs:enumeration value="cloud"/>
-            <xs:enumeration value="cloudDiscovery"/>
+            <xs:enumeration value="none">
+                <xs:annotation>
+                    <xs:documentation>No interaction with external systems at all</xs:documentation>
+                </xs:annotation>
+            </xs:enumeration>
+            <xs:enumeration value="local">
+                <xs:annotation>
+                    <xs:documentation>Interaction with external systems, without internet access</xs:documentation>
+                </xs:annotation>
+            </xs:enumeration>
+            <xs:enumeration value="hybrid">
+                <xs:annotation>
+                    <xs:documentation>Interaction with external systems, internet access required only for extended functionality (such as discovery)</xs:documentation>
+                </xs:annotation>
+            </xs:enumeration>
+            <xs:enumeration value="cloud">
+                <xs:annotation>
+                    <xs:documentation>Interaction with external systems, internet access required for normal operation</xs:documentation>
+                </xs:annotation>
+            </xs:enumeration>
         </xs:restriction>
     </xs:simpleType>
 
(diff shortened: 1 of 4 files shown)


jlaur:
I have prepared openhab/openhab-addons#14845 in case this pull request will be accepted, since `cloudDiscovery` will then no longer be valid. Therefore that PR should be merged right after this one.

lsiepel:
Shouldn't there also be a  PR to the UI, adding these options to filter the bindings?

jlaur:
> Shouldn't there also be a PR to the UI, adding these options to filter the bindings?

You might be right, I did not know anything was implemented in the UI yet, didn't notice that.

wborn (commented):
Can you also create a PR to update https://github.com/openhab/website/blob/main/.vuepress/public/schemas/addon-1.0.0.xsd ?

jlaur:
@wborn - thanks for the reference! Done.

kaikreuzer (approved):

PR #3564, opened 2023-04-21
[voice] Add interface to represent a remote keyword spotter

GiviMAD:
Add interface that allows to register a keyword spotting service that is not running on the oh server, so the dialog processing can be triggered remotely instead of processing an audio stream.

The reason for having this interface is that the DialogProcessor can avoid the creation of an audio stream from the source that will not be used.

Hope it is an acceptable solution. Best regards.

### DIFF
--- a/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/KSEdgeService.java
+++ b/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/KSEdgeService.java
@@ -1,0 +1,44 @@
+/**
+ * Copyright (c) 2010-2023 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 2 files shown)


GiviMAD:
Hello @openhab/core-maintainers, do you think these PR can be merged before the 4.1.0? First of all, do you think it is an acceptable solution?

For the context, I would like to have a way of register a remote keyword spotter instance, because I would like to add the functionality in the rustpotter add-on to handle a websocket connection from the rustpotter-cli and expose it in the server as an available keyword spotter so, having the rustpotter-cli running in a device as a service and with pulseaudio setup, I can register a dialog using it and the sink and source from the same device.

I have the idea since a while ago, but with the functionality I got after fine-tuning a whisper model to use with the add-on on these PR https://github.com/openhab/openhab-addons/pull/15166 , I would like to add at least two more speakers around the house for daily usage, and I would prefer to run rustpotter at the remote device, at least having the option for it in case I start to see the server is overloaded.

kaikreuzer (approved):
Hey @GiviMAD, sorry that this PR hang around unnoticed for so long.
I think it is a valid approach and I appreciate your work on these features!

GiviMAD:
Thank you!

PR #3565, opened 2023-04-21
Backward compatibility for jsprofile from OH3

jimtng:
https://community.openhab.org/t/preparation-for-oh-4-0-how-to-find-my-js-transformations/146233/4

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/profile/ScriptProfile.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/profile/ScriptProfile.java
@@ -44,6 +44,11 @@ public class ScriptProfile implements StateProfile {
     public static final String CONFIG_TO_ITEM_SCRIPT = "toItemScript";
     public static final String CONFIG_TO_HANDLER_SCRIPT = "toHandlerScript";
 
+    /**
+     * @deprecated Use `toItemScript` instead.
+     */
+    public static final String CONFIG_FUNCTION = "function";
+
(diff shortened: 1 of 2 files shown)


J-N-K:
I understand your intention, but I think this is wrong. We should not add something that will be wrong in most cases (which script would be able to to bidirectional conversion?) and confuse people who are configuring the profile now.

jimtng:
The jsscripting profile passes on the command from item to handler, hence why I set it to `|input`. The user script from `function` is only used for `toItemScript`. Wouldn't that work?

J-N-K:
It still shows up in the UI and allows to add new "wrong" configurations. And if we want to add that, then we should:

1. check for `toItemScript`, if present use it
2. check for `function`, if present use it and print a warning that support for `function will be removed`.
3. remove it from the config description, so it doesn't show up in the UI

jimtng:
I originally didn't want to add it to the config description, but if it wasn't there, how can the user "transfer" their existing configuration? I thought that by showing it there, with the warning, they could at least copy paste.

How would you propose we do this?

jimtng:
I haven't investigated it, but perhaps there is a way to automatically migrate the config Param?

stefaanbolle:
I have a few 100 changes to make. It would be nice that the system can change the configs automatically during a next upgrade or that a script can be provided to do it manually if you would chose to not maintain backwards compatibility (for which I also understand your argumentation).

jimtng:
> I have a few 100 changes to make

@stefaanbolle Were your items created through the MainUI or file based configs (i.e. .items file)?

stefaanbolle:
> 

All via Main UI. No text configuration files in use

jimtng:
Closing this. The upgrade process will be done through cli #3268

PR #3566, opened 2023-04-21
Remove orbit bundles from runtime-pom

J-N-K:
They are not available on Maven Central and generate warning during the build.

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -436,9 +436,9 @@
       </exclusions>
     </dependency>
     <dependency>
-      <groupId>org.eclipse.orbit.bundles</groupId>
-      <artifactId>net.i2p.crypto.eddsa</artifactId>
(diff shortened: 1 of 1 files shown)


wborn:
That would be nice but as you can see the artifacts on Maven Central do not work as well as the Orbit bundles.
To get rid of the warnings just create some POMs and Kai should be able to upload those to Artifactory as well.
There are also also some missing POMs for some add-on deps, see https://github.com/openhab/openhab-addons/issues/11524.

J-N-K:
It's essentially the same problem we saw in #3321: The xtext guys use `Require-Bundle` headers instead of `Import-Package`, and they require e.g. `io.github.classgraph` while the non-orbit bundle has a BSN `io.github.classgraph.classgraph`.

Maybe we should really re-package the xtext bundles and use `Import-Package` instead of `Require-Bundle` headers. This will also help us with the equninox/felix issue.

PR #3572, opened 2023-04-23
Fix marketplace add-on services fail after upgrade

J-N-K:
A format change in the database (related to the introduction of addon.xml) leads to a non-migrateable and permanent error that floods the log with exceptions and malfunctions.

If such an error is detected, the database is purged and an error logged.

Also reported on the forum several times.

### DIFF
--- a/bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/AbstractRemoteAddonService.java
+++ b/bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/AbstractRemoteAddonService.java
@@ -97,8 +97,8 @@ public abstract class AbstractRemoteAddonService implements AddonService {
         } catch (JsonSyntaxException e) {
             List.copyOf(installedAddonStorage.getKeys()).forEach(installedAddonStorage::remove);
             logger.error(
-                    "Failed to read JSON database, trying to purge it. You might need to re-install the bindings from the '{}' service.",
-                    getId());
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/AbstractRemoteAddonService.java
```diff
@@ -90,8 +91,16 @@ public void refreshSource() {
...
         }
         List<Addon> addons = new ArrayList<>();
-        installedAddonStorage.stream().map(e -> Objects.requireNonNull(gson.fromJson(e.getValue(), Addon.class)))
-                .forEach(addons::add);
+        try {
+            installedAddonStorage.stream().map(e -> Objects.requireNonNull(gson.fromJson(e.getValue(), Addon.class)))
+                    .forEach(addons::add);
+        } catch (JsonSyntaxException e) {
+            List.copyOf(installedAddonStorage.getKeys()).forEach(installedAddonStorage::remove);
+            logger.error(
```
jlaur, line 99: Perhaps it would be useful to also log the removed entries?

J-N-K: Done

J-N-K:
@kaikreuzer For better user-experience this should also be merged before M3.

kaikreuzer (approved):

Issue #3544, opened 2023-04-12
[ephemeris] Upgrade/replace Jollyday


jlaur:
Jollyday used by Ephemeris seems abandoned years ago:
https://github.com/svendiedrichsen/jollyday

Fortunately the project was picked up and continued here:
https://github.com/focus-shift/jollyday

See svendiedrichsen/jollyday#72

This seems almost like a drop-in replacement.

Replacing the library would give several benefits, for example updated holidays for 70 countries that would be bundled with the distribution. The current situation is that users need to maintain national holidays by themselves. As an example, one of the Danish holidays has been eliminated, starting next year (**GENERAL_PRAYER_DAY** in the XML below). So if I understand this correctly, I must now create my own file:

/etc/openhab/services/Holidays_dk.xml:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<tns:Configuration hierarchy="dk" description="Denmark"
        xmlns:tns="http://www.example.org/Holiday" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:schemaLocation="http://www.example.org/Holiday /Holiday.xsd">
        <tns:Holidays>
                <tns:Fixed month="JANUARY" day="1" descriptionPropertiesKey="NEW_YEAR" />
                <tns:Fixed month="DECEMBER" day="25"
                        descriptionPropertiesKey="CHRISTMAS" />
                <tns:Fixed month="DECEMBER" day="26"
                        descriptionPropertiesKey="STEPHENS" />
                <tns:ChristianHoliday type="EASTER" />
                <tns:ChristianHoliday type="MAUNDY_THURSDAY" />
                <tns:ChristianHoliday type="GOOD_FRIDAY" />
                <tns:ChristianHoliday type="EASTER_MONDAY" />
                <tns:ChristianHoliday type="GENERAL_PRAYER_DAY" validTo="2023" />
                <tns:ChristianHoliday type="ASCENSION_DAY" />
                <tns:ChristianHoliday type="WHIT_MONDAY" />
                <tns:ChristianHoliday type="PENTECOST" />
        </tns:Holidays>
</tns:Configuration>
```

and directly reference this file in all rules:

```java
var dt1 = ZonedDateTime.of(2023, 5, 5, 0, 0, 0, 0, ZoneId.systemDefault())
var dt2 = ZonedDateTime.of(2024, 4, 26, 0, 0, 0, 0, ZoneId.systemDefault())
[...]

PR #3573, opened 2023-04-23
[ephemeris] Add support for overriding holiday definitions

jlaur:
The version of Jollyday we are using was abandoned years ago, so the holiday definitions we are using are getting outdated as well.

This pull request introduces the possibility to override holiday definitions in core, and as an example contains updated Danish definitions, ending Great Prayer Day this year, since our government has [abolished](https://www.thelocal.dk/20230228/great-prayer-day-abolished-by-danish-parliament) it from next year.

The obvious disadvantage is that we would then take over the maintenance. But on the other hand, they are not maintained upstream anymore, so at least it would improve the user experience. Additionally, it would not change the API, so upgrading to **focus-shift** later will be seamless for users.

Related to #3544

### DIFF
--- a/bundles/org.openhab.core.ephemeris/src/main/java/org/openhab/core/ephemeris/internal/EphemerisManagerImpl.java
+++ b/bundles/org.openhab.core.ephemeris/src/main/java/org/openhab/core/ephemeris/internal/EphemerisManagerImpl.java
@@ -59,6 +59,7 @@ import de.jollyday.Holiday;
 import de.jollyday.HolidayManager;
 import de.jollyday.ManagerParameter;
 import de.jollyday.ManagerParameters;
+import de.jollyday.parameter.CalendarPartManagerParameter;
 import de.jollyday.util.ResourceUtil;
 
 /**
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.core.ephemeris/src/main/java/org/openhab/core/ephemeris/internal/EphemerisManagerImpl.java
```diff
@@ -231,9 +234,15 @@ private URL getUrl(String filename) throws FileNotFoundException {
     private HolidayManager getHolidayManager(Object managerKey) {
         HolidayManager holidayManager = holidayManagers.get(managerKey);
         if (holidayManager == null) {
-            final ManagerParameter parameters = managerKey.getClass() == String.class
-                    ? ManagerParameters.create((String) managerKey)
-                    : ManagerParameters.create((URL) managerKey);
+            final ManagerParameter parameters;
+            if (managerKey.getClass() == String.class) {
```
clinique, line 238: Wouldn't `if (managerKey instanceof String stringKey) {` be better ? This avoids below typecasts

clinique (changes requested):

bundles/org.openhab.core.ephemeris/src/main/java/org/openhab/core/ephemeris/internal/EphemerisManagerImpl.java
jlaur: Right - fixed. Do you by any chance know how to run the integration tests locally? I would like to refactor a bit more, but first would need to get the tests running.

clinique (approved):

clinique (approved):

bundles/org.openhab.core.ephemeris/src/main/java/org/openhab/core/ephemeris/internal/EphemerisManagerImpl.java
```diff
@@ -107,7 +111,6 @@ public class EphemerisManagerImpl implements EphemerisManager, ConfigOptionProvi
     public EphemerisManagerImpl(final @Reference LocaleProvider localeProvider) {
         this.localeProvider = localeProvider;
 
-        final Bundle bundle = FrameworkUtil.getBundle(getClass());
```
J-N-K, line 110: I would suggest to add `BundleContext bc` to the constructor parameters and use `Bundle bundle = bundleContext.getBundle();` here. I don't know why it was done in this strange way.

jlaur: Done.

J-N-K (approved):
LGTM

PR #3575, opened 2023-04-24
Consider auto-update policy from channel types

kaikreuzer:
When the Main UI creates things/channels through the REST API, it does not set autoUpdatePolicy for the channels, which makes the core use "RECOMMEND" as a policy without looking at a potential information at the channel type definition.

This PR leaves the autoUpdatePolicy for the channels unset, so that the AutoUpdateManager then checks for the information in the channel type.

See also https://github.com/openhab/openhab-addons/pull/14809#issuecomment-1510455008


J-N-K (approved):
LGTM, thanks

PR #3580, opened 2023-04-27
Code cleanup: Use Java 17 features

holgerfriedrich:
Use new language features, mainly instanceof pattern matching.
This change covers itests.

Relates to #3508, #3522, and #3585.

### DIFF
--- a/itests/org.openhab.core.config.discovery.tests/src/main/java/org/openhab/core/config/discovery/inbox/DynamicThingUpdateOSGiTest.java
+++ b/itests/org.openhab.core.config.discovery.tests/src/main/java/org/openhab/core/config/discovery/inbox/DynamicThingUpdateOSGiTest.java
@@ -108,7 +108,7 @@ public class DynamicThingUpdateOSGiTest extends JavaOSGiTest {
     }
 
     private ThingHandlerFactory createThingHandlerFactory() {
-        return new BaseThingHandlerFactory() {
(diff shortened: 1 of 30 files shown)


itests/org.openhab.core.config.discovery.tests/src/main/java/org/openhab/core/config/discovery/inbox/DynamicThingUpdateOSGiTest.java
```diff
@@ -121,7 +121,6 @@ public boolean supportsThingType(ThingTypeUID thingTypeUID) {
                 return thingHandler;
             }
         };
-        return thingHandlerFactory;
```
J-N-K, line 124: I would prefer to keep the old version here, it look cleaner.

J-N-K (changes requested):
Thanks for the huge cleanup. I have only one comment.

itests/org.openhab.core.config.discovery.tests/src/main/java/org/openhab/core/config/discovery/inbox/DynamicThingUpdateOSGiTest.java
holgerfriedrich: ok, rolled back

J-N-K (approved):
Thanks!

PR #3583, opened 2023-04-29
[core] Switch `DateTimeType` to `Instant` internally for consistent time-zone handling

### LINKED ISSUE #2898: [DateTimeType] ZonedDateTime, Instant or neither?

jlaur:
Proposal for switching `DateTimeType` to use `Instant` internally rather than `ZonedDateTime`.

This will effectively discard time-zone information from the _source_ of the `DateTimeType` and leave it to consumers to apply time-zone at the moment of _presentation_, according to the configured time-zone in openHAB regional settings or in the browser.

Configured time-zone is now applied for:
- REST API
- SSE item state events
- Item UI registry (affecting e.g. sitemaps)

Consequential changes because `DateTimeType` is now "zone-less":
- Parameter `timezone` for `TimestampOffsetProfile` has been removed. openhab/openhab-docs#2403 has been prepared.
- Methods `toZone` and `toLocaleZone` have been marked as deprecated and all usages have been refactored. I would suggest to remove them in 5.0 (or maybe even before when all bindings are refactored, see openhab/openhab-addons#17725).
- All usages of `getZonedDateTime` have been refactored.

The following demonstrates the difference between old and new behavior.

REST API calls are for `http://localhost:8080/rest/items/DateTimeTest` and the returned `state` is shown in the comparison table:
```json
{
    "link": "http://localhost:8080/rest/items/DateTimeTest",
    "state": "2024-10-31T21:00:00.000+0100",
    "stateDescription": {
        "pattern": "%1$tY-%1$tm-%1$td %1$tH:%1$tM:%1$tS",
        "readOnly": false,
        "options": []
    },
    "editable": true,
    "type": "DateTime",
    "name": "DateTimeTest",
    "label": "",
    "category": "",
    "tags": [],
    "groupNames": []
}
```

Updates are made in Karaf using commands like:
```shell
openhab:update DateTimeTest 2024-10-31T21:00:00+01:00
```

| Step                                     | Before                       | Now                              |
|------------------------------------------|------------------------------|----------------------------------|
| System time-zone: Europe/Copenhagen      |                              |                                  |
| Set openHAB time-zone: Europe/Copenhagen |                              |                                  |
| Update to: 2024-10-01T20:00:00Z          |                              |                                  |
| Observe REST API result                  | 2024-10-01T20:00:00.000+0000 | 2024-10-01T**22:00:00.000+0200** |
| Observe /settings/items/                 | 2024-10-01T20:00:00.000+0000 | 2024-10-01T**22:00:00.000+0200** |
| Observe /settings/items/DateTimeTest     | 2024-10-01 22:00:00          | 2024-10-01 22:00:00              |
| Observe sitemap                          | 2024-10-01 22:00:00          | 2024-10-01 22:00:00              |
| Update to: 2024-10-31T20:00:00Z          |                              |                                  |
| Observe REST API result                  | 2024-10-31T20:00:00.000+0000 | 2024-10-31T**21:00:00.000+0100** |
| Observe /settings/items/                 | 2024-10-31T20:00:00.000+0000 | 2024-10-31T**21:00:00.000+0100** |
| Observe /settings/items/DateTimeTest     | 2024-10-31 21:00:00          | 2024-10-31 21:00:00              |
| Observe sitemap                          | 2024-10-31 21:00:00          | 2024-10-31 21:00:00              |
| Update to: 2024-10-31T21:00:00+01:00     |                              |                                  |
| Observe REST API result                  | 2024-10-31T21:00:00.000+0100 | 2024-10-31T21:00:00.000+0100     |
| Observe /settings/items/                 | 2024-10-31T21:00:00.000+0100 | 2024-10-31T21:00:00.000+0100     |
| Observe /settings/items/DateTimeTest     | 2024-10-31 21:00:00          | 2024-10-31 21:00:00              |
| Observe sitemap                          | 2024-10-31 21:00:00          | 2024-10-31 21:00:00              |
| Set time-zone: US/Alaska                 |                              |                                  |
| Observe REST API result                  | 2024-10-31T21:00:00.000+0100 | 2024-10-31T**12:00:00.000-0800** |
| Observe /settings/items/                 | 2024-10-31T21:00:00.000+0100 | 2024-10-31T**12:00:00.000-0800** |
| Observe /settings/items/DateTimeTest     | 2024-10-31 21:00:00          | 2024-10-31 **12:00:00**          |
| Observe sitemap                          | 2024-10-31 21:00:00          | 2024-10-31 **12:00:00**          |
| Update to: 2024-10-31T**12:00:00-0800**  |                              |                                  |
| Observe REST API result                  | 2024-10-31T12:00:00.000-0800 | 2024-10-31T12:00:00.000-0800     |
| Observe /settings/items/                 | 2024-10-31T12:00:00.000-0800 | 2024-10-31T12:00:00.000-0800     |
| Observe /settings/items/DateTimeTest     | 2024-10-31 21:00:00          | 2024-10-31 **12:00:00**          |
| Observe sitemap                          | 2024-10-31 21:00:00          | 2024-10-31 **12:00:00**          |

As it can be seen from the table, we now have consistent results.

Example DSL rule:

```java
var dt = ZonedDateTime.of(2024, 4, 26, 0, 0, 0, 0, ZoneId.systemDefault())
DateTimeTest1.postUpdate(new DateTimeType(dt))
DateTimeTest2.postUpdate(new DateTimeType(dt.toInstant().atZone(ZoneId.of("UTC"))))
DateTimeTest3.postUpdate(new DateTimeType(dt.toInstant()))
```

Previous result with system time-zone CET/DST:

```
[openhab.event.ItemStateChangedEvent ] - Item 'DateTimeTest1' changed from NULL to 2024-04-26T00:00:00.000+0200
[openhab.event.ItemStateChangedEvent ] - Item 'DateTimeTest2' changed from NULL to 2024-04-25T22:00:00.000+0000
```

New result with same system time-zone:

```
[openhab.event.ItemStateChangedEvent ] - Item 'DateTimeTest1' changed from NULL to 2024-04-26T00:00:00.000+0200
[openhab.event.ItemStateChangedEvent ] - Item 'DateTimeTest2' changed from NULL to 2024-04-26T00:00:00.000+0200
[openhab.event.ItemStateChangedEvent ] - Item 'DateTimeTest3' changed from NULL to 2024-04-26T00:00:00.000+0200
```

Summary of benefits for developers:
- In cases where `TimeZoneProvider` is used only to generate a `DateTimeType` with the correct time-zone, this can now be completely removed. This simplifies the code. See openhab/openhab-addons#17725 for examples.
- The default `DateTimeType` constructor can now be used (without losing correct time-zone).
- In cases where the configured time-zone was not correctly applied, this will now work properly without any changes.
- The change is fully backwards compatible, although two methods have been deprecated: `toZone` and `toLocaleZone`. I will go through usages and remove them, since they no longer make sense. See openhab/openhab-addons#17725.
- Pull request reviews should be a bit smoother, since add-on maintainers will no longer have to assist contributors in constructing a proper `DateTimeType` using `TimeZoneProvider`.

And for users:
- There will now be consistent behavior across all bindings, where previously time-zone could differ between bindings and even within same binding: System time-zone or openHAB time-zone.
- There will now be consistent behavior across different parts of the UI, where previously some parts used the time-zone contained in the `DateTimeType` (which could be either depending on binding) and other parts used system time-zone.
- A change in time-zone will have immediate effect for the UI without requiring state updates for the items.

And finally limitations:
- It is no longer possible to send a `DateTimeType` command to a binding with a specified time-zone. The bindings will have to provide a time-zone if needed (e.g. time-zone from device, openHAB configuration or system time-zone). I have not found any bindings using this possibility, so I don't think this will cause any issues.

Resolves #2898

### DIFF
--- a/bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/RuleResource.java
+++ b/bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/RuleResource.java
@@ -441,7 +441,7 @@ public class RuleResource implements RESTResource {
             return defaultSupplier.get();
         }
         final DateTimeType dateTime = new DateTimeType(sTime);
-        return dateTime.getInstant().atZone(timeZoneProvider.getTimeZone());
+        return dateTime.getZonedDateTime(timeZoneProvider.getTimeZone());
     }
 
     @GET
--- a/bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/RuleResource.java
+++ b/bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/RuleResource.java
@@ -76,6 +76,7 @@ import org.openhab.core.common.registry.RegistryChangedRunnableListener;
 import org.openhab.core.config.core.ConfigUtil;
 import org.openhab.core.config.core.Configuration;
 import org.openhab.core.events.Event;
+import org.openhab.core.i18n.TimeZoneProvider;
 import org.openhab.core.io.rest.DTOMapper;
 import org.openhab.core.io.rest.JSONResponse;
 import org.openhab.core.io.rest.RESTConstants;
@@ -133,6 +134,7 @@ public class RuleResource implements RESTResource {
     private final RuleManager ruleManager;
     private final RuleRegistry ruleRegistry;
     private final ManagedRuleProvider managedRuleProvider;
+    private final TimeZoneProvider timeZoneProvider;
     private final RegistryChangedRunnableListener<Rule> resetLastModifiedChangeListener = new RegistryChangedRunnableListener<>(
             () -> lastModified = null);
 
@@ -144,11 +146,13 @@ public class RuleResource implements RESTResource {
             final @Reference DTOMapper dtoMapper, //
             final @Reference RuleManager ruleManager, //
             final @Reference RuleRegistry ruleRegistry, //
-            final @Reference ManagedRuleProvider managedRuleProvider) {
+            final @Reference ManagedRuleProvider managedRuleProvider, //
+            final @Reference TimeZoneProvider timeZoneProvider) {
         this.dtoMapper = dtoMapper;
         this.ruleManager = ruleManager;
         this.ruleRegistry = ruleRegistry;
         this.managedRuleProvider = managedRuleProvider;
+        this.timeZoneProvider = timeZoneProvider;
 
         this.ruleRegistry.addRegistryChangeListener(resetLastModifiedChangeListener);
     }
@@ -431,9 +435,9 @@ public class RuleResource implements RESTResource {
         return Response.ok(ruleExecutions.toList()).build();
     }
 
-    private static ZonedDateTime parseTime(String sTime) {
+    private ZonedDateTime parseTime(String sTime) {
         final DateTimeType dateTime = new DateTimeType(sTime);
-        return dateTime.getZonedDateTime();
+        return dateTime.getInstant().atZone(timeZoneProvider.getTimeZone());
     }
 
     private static long daysBetween(ZonedDateTime d1, ZonedDateTime d2) {
--- a/bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/RuleResource.java
+++ b/bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/RuleResource.java
@@ -25,6 +25,7 @@ import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Predicate;
+import java.util.function.Supplier;
 import java.util.stream.Stream;
 
 import javax.annotation.security.RolesAllowed;
@@ -423,10 +424,10 @@ public class RuleResource implements RESTResource {
                     + DateTimeType.DATE_PATTERN_WITH_TZ_AND_MS + "]") @QueryParam("from") @Nullable String from,
             @Parameter(description = "End time of the simulated rule executions. Will default to 30 days after the start time. Must be less than 180 days after the given start time. ["
                     + DateTimeType.DATE_PATTERN_WITH_TZ_AND_MS + "]") @QueryParam("until") @Nullable String until) {
-        final ZonedDateTime fromDate = from == null || from.isEmpty() ? ZonedDateTime.now() : parseTime(from);
-        final ZonedDateTime untilDate = until == null || until.isEmpty() ? fromDate.plusDays(31) : parseTime(until);
+        final ZonedDateTime fromDate = parseTime(from, ZonedDateTime::now);
+        final ZonedDateTime untilDate = parseTime(until, () -> fromDate.plusDays(31));
 
-        if (daysBetween(fromDate, untilDate) >= 180) {
+        if (ChronoUnit.DAYS.between(fromDate, untilDate) >= 180) {
             return JSONResponse.createErrorResponse(Status.BAD_REQUEST,
                     "Simulated time span must be smaller than 180 days.");
         }
@@ -435,13 +436,12 @@ public class RuleResource implements RESTResource {
         return Response.ok(ruleExecutions.toList()).build();
     }
 
-    private ZonedDateTime parseTime(String sTime) {
+    private ZonedDateTime parseTime(@Nullable String sTime, Supplier<ZonedDateTime> defaultSupplier) {
+        if (sTime == null || sTime.isEmpty()) {
+            return defaultSupplier.get();
+        }
         final DateTimeType dateTime = new DateTimeType(sTime);
         return dateTime.getInstant().atZone(timeZoneProvider.getTimeZone());
-    }
-
-    private static long daysBetween(ZonedDateTime d1, ZonedDateTime d2) {
-        return ChronoUnit.DAYS.between(d1, d2);
     }
 
     @GET
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ItemStateConditionHandler.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ItemStateConditionHandler.java
@@ -161,7 +161,7 @@ public class ItemStateConditionHandler extends BaseConditionModuleHandler implem
         State itemState = item.getState();
         if (itemState instanceof DateTimeType dateTimeState) {
             Instant itemTime = dateTimeState.getInstant();
-            Instant compareTime = getCompareTime(state).toInstant();
+            Instant compareTime = getCompareTime(state);
             return itemTime.compareTo(compareTime) <= 0;
         } else if (itemState instanceof QuantityType qtState) {
             if (compareState instanceof DecimalType type) {
@@ -198,7 +198,7 @@ public class ItemStateConditionHandler extends BaseConditionModuleHandler implem
         State itemState = item.getState();
         if (itemState instanceof DateTimeType dateTimeState) {
             Instant itemTime = dateTimeState.getInstant();
-            Instant compareTime = getCompareTime(state).toInstant();
+            Instant compareTime = getCompareTime(state);
             return itemTime.compareTo(compareTime) >= 0;
         } else if (itemState instanceof QuantityType qtState) {
             if (compareState instanceof DecimalType type) {
@@ -253,36 +253,36 @@ public class ItemStateConditionHandler extends BaseConditionModuleHandler implem
         eventSubscriberRegistration.unregister();
     }
 
-    private ZonedDateTime getCompareTime(String input) {
+    private Instant getCompareTime(String input) {
         if (input.isBlank()) {
             // no parameter given, use now
-            return ZonedDateTime.now();
+            return Instant.now();
         }
         try {
-            return ZonedDateTime.parse(input);
+            return ZonedDateTime.parse(input).toInstant();
         } catch (DateTimeParseException ignored) {
         }
         try {
             return LocalDateTime.parse(input, DateTimeFormatter.ISO_LOCAL_DATE_TIME)
-                    .atZone(timeZoneProvider.getTimeZone());
+                    .atZone(timeZoneProvider.getTimeZone()).toInstant();
         } catch (DateTimeParseException ignored) {
         }
         try {
             int dayPosition = input.indexOf("D");
             if (dayPosition == -1) {
                 // no date in string, add period symbol and time separator
-                return ZonedDateTime.now().plus(Duration.parse("PT" + input));
+                return Instant.now().plus(Duration.parse("PT" + input));
             } else if (dayPosition == input.length() - 1) {
                 // day is the last symbol, only add the period symbol
-                return ZonedDateTime.now().plus(Duration.parse("P" + input));
+                return Instant.now().plus(Duration.parse("P" + input));
             } else {
                 // add period symbol and time separator
-                return ZonedDateTime.now().plus(Duration
+                return Instant.now().plus(Duration
                         .parse("P" + input.substring(0, dayPosition + 1) + "T" + input.substring(dayPosition + 1)));
             }
         } catch (DateTimeParseException e) {
             logger.warn("Couldn't get a comparable time from '{}', using now", input);
         }
-        return ZonedDateTime.now();
+        return Instant.now();
     }
 }
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ItemStateConditionHandler.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ItemStateConditionHandler.java
@@ -13,6 +13,7 @@
 package org.openhab.core.automation.internal.module.handler;
 
 import java.time.Duration;
+import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
@@ -158,9 +159,9 @@ public class ItemStateConditionHandler extends BaseConditionModuleHandler implem
         Item item = itemRegistry.getItem(itemName);
         State compareState = TypeParser.parseState(item.getAcceptedDataTypes(), state);
         State itemState = item.getState();
-        if (itemState instanceof DateTimeType type) {
-            ZonedDateTime itemTime = type.getZonedDateTime();
-            ZonedDateTime compareTime = getCompareTime(state);
+        if (itemState instanceof DateTimeType dateTimeState) {
+            Instant itemTime = dateTimeState.getInstant();
+            Instant compareTime = getCompareTime(state).toInstant();
             return itemTime.compareTo(compareTime) <= 0;
         } else if (itemState instanceof QuantityType qtState) {
             if (compareState instanceof DecimalType type) {
@@ -195,9 +196,9 @@ public class ItemStateConditionHandler extends BaseConditionModuleHandler implem
         Item item = itemRegistry.getItem(itemName);
         State compareState = TypeParser.parseState(item.getAcceptedDataTypes(), state);
         State itemState = item.getState();
-        if (itemState instanceof DateTimeType type) {
-            ZonedDateTime itemTime = type.getZonedDateTime();
-            ZonedDateTime compareTime = getCompareTime(state);
+        if (itemState instanceof DateTimeType dateTimeState) {
+            Instant itemTime = dateTimeState.getInstant();
+            Instant compareTime = getCompareTime(state).toInstant();
             return itemTime.compareTo(compareTime) >= 0;
         } else if (itemState instanceof QuantityType qtState) {
             if (compareState instanceof DecimalType type) {
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/item/EnrichedItemDTOMapper.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/item/EnrichedItemDTOMapper.java
@@ -12,9 +12,7 @@
  */
 package org.openhab.core.io.rest.core.item;
 
-import java.time.Instant;
 import java.time.ZoneId;
-import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.LinkedHashSet;
@@ -56,10 +54,6 @@ public class EnrichedItemDTOMapper {
 
     private static final Pattern EXTRACT_TRANSFORM_FUNCTION_PATTERN = Pattern.compile("(.*?)\\((.*)\\):(.*)");
 
-    private static final String DATE_FORMAT_PATTERN_WITH_TZ_RFC = "yyyy-MM-dd'T'HH:mm[:ss[.SSSSSSSSS]]Z";
-    private static final DateTimeFormatter FORMATTER_TZ_RFC = DateTimeFormatter
-            .ofPattern(DATE_FORMAT_PATTERN_WITH_TZ_RFC);
-
     private static final Logger LOGGER = LoggerFactory.getLogger(EnrichedItemDTOMapper.class);
 
     /**
@@ -100,7 +94,7 @@ public class EnrichedItemDTOMapper {
             if (dateTime == null) {
                 state = item.getState().toFullString();
             } else {
-                state = formatDateTime(dateTime.getInstant(), zoneId);
+                state = dateTime.toFullString(zoneId);
             }
         } else {
             state = item.getState().toFullString();
@@ -153,24 +147,6 @@ public class EnrichedItemDTOMapper {
         }
 
         return enrichedItemDTO;
-    }
-
-    private static String formatDateTime(Instant instant, ZoneId zoneId) {
-        String formatted = instant.atZone(zoneId).format(FORMATTER_TZ_RFC);
-        if (formatted.contains(".")) {
-            String sign = "";
-            if (formatted.contains("+")) {
-                sign = "+";
-            } else if (formatted.contains("-")) {
-                sign = "-";
-            }
-            if (!sign.isEmpty()) {
-                // the formatted string contains 9 fraction-of-second digits
-                // truncate at most 2 trailing groups of 000s
-                return formatted.replace("000" + sign, sign).replace("000" + sign, sign);
-            }
-        }
-        return formatted;
     }
 
     private static @Nullable StateDescription considerTransformation(@Nullable StateDescription stateDescription) {
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/item/EnrichedItemDTOMapper.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/item/EnrichedItemDTOMapper.java
@@ -12,6 +12,9 @@
  */
 package org.openhab.core.io.rest.core.item;
 
+import java.time.Instant;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.LinkedHashSet;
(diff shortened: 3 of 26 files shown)


jlaur:
I hit a wall with this when realizing that `toFullString()` and `toString()` should probably format as UTC for consistency. However, I believe one of these is used by Main UI for display purposes, so to improve the display with correct time-zone rather breaking it by raw UTC format, we would need UI changes - see also https://github.com/openhab/openhab-core/issues/2898#issuecomment-1370230276.

I don't know if any @openhab/webui-maintainers can throw some light at what may be a possible way forward?

jlaur:
@mherwege - seeing your work in #4169 I'm wondering if you can advise how to accomplish decoupling the timezone + formatting of `DateTimeType`. The general idea is to reduce it to be only an `Instant` without time-zone information. Therefore the UI should no longer display the raw output of `toString()` / `toFullString()`. Instead it must use the configured time-zone from `TimeZoneProvider` for formatting the `DateTime` in local time.

Currently it's formatted like this: `2024-04-10T20:55:01.000+0200`. The problem is that "+0200" is the time-zone provided when updating the value, not the current time-zone. I don't know how this formatting works in Main UI.

mherwege:
@jlaur Main UI works with what it gets from the REST API.
On the Items screen (items list), it will show the state exactly as it is received from the REST API `/items` endpoint.
Here is an extract of such a response:
```
    {
        "link": "http://openhabian.local:8080/rest/items/PMD",
        "state": "2024-04-11T00:00:00.000+0200",
        "stateDescription": {
            "pattern": "%1$td-%1$tm-%1$tY",
            "readOnly": true,
            "options": []
        },
        "editable": true,
        "type": "DateTime",
        "name": "PMD",
        "label": "PMD",
        "category": "calendar",
        "tags": [],
        "groupNames": [
            "Afval"
        ]
    }
```
`state` is what gets shown in the items list.

The formatted state comes from the `states` event stream in the `displayState` field. It is a result of applying the stateDescription pattern, but that is done in core.
Event stream extract:
```
data: {"Afval":{"state":"2024-04-10T00:00:00.000+0200","displayState":"10-04-2024","type":"DateTime"},"Papier":{"state":"2024-04-10T00:00:00.000+0200","displayState":"10-04-2024","type":"DateTime"},"Restafval":{"state":"2024-04-18T00:00:00.000+0200","displayState":"18-04-2024","type":"DateTime"},"PMD":{"state":"2024-04-11T00:00:00.000+0200","displayState":"11-04-2024","type":"DateTime"}}
```
`displayState` gets used in widgets when available.

So, to understand where `state` and `displayState` come from, one needs to look at the REST API code. According to https://github.com/openhab/openhab-core/blob/cbb458e0c3c35f353954f82031c5cd1b3943759c/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java#L953, `state` is added through the `EnrichedItemDTOMapper` here: https://github.com/openhab/openhab-core/blob/cbb458e0c3c35f353954f82031c5cd1b3943759c/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/item/EnrichedItemDTOMapper.java#L81. So it is simply using the `toFullString()` method on the state of the item.

jlaur:
@mherwege - many thanks for your insights. I will have a closer look at this soon.

jlaur:
@openhab/core-maintainers - I don't have permission to remove the "work in progress" label since I'm not in the contributors team, so perhaps someone can do that for me? I consider this PR ready for review now after finally fixing the returned `state` in the REST API. In parallel I will run some more tests and also see if it is possible to use `TimeZoneProvider` in more places rather than converting to system time-zone.

@mherwege - perhaps you'd also like to have a look?

lolodomo:
What about bindings setting a DateTimeType state with a particular time zone?
Is it really backward compatible with all bindings?

jlaur:
> What about bindings setting a DateTimeType state with a particular time zone?

They can still do that, but the time-zone will be stripped, so it's really not needed anymore.

> Is it really backward compatible with all bindings?

It should be, but there can be unit tests failing if they expect a certain time-zone:
```java
DateTimeType dt = new DateTimeType(ZonedDateTime.now());
ZonedDateTime zdt = dt.getZonedDateTime();
```

Previously this would return a `ZonedDateTime` with the same zone as provided when creating the object. But since the zone is now stripped, it will return a `ZonedDateTime` with system time-zone.

jlaur:
@joerg1985 - perhaps you would also like to have a look? I saw some of your recent related comments and enhancements, e.g. #4384.

jlaur:
This will call `DateTimeType.toFullString`:

https://github.com/openhab/openhab-core/blob/3bc9ae3e1469dad525cd7b01a4caff5cf3e0c456/bundles/org.openhab.core/src/main/java/org/openhab/core/items/events/ItemEventFactory.java#L300

which is probably the reason for system time-zone being used in some places. I don't know how to improve this, since `TimeZoneProvider` is required for properly formatting this in the configured local time-zone. One solution might be to change the implementation of `DateTimeType.toFullString` to format in UTC as `Instant`. However, this would however require all receivers to reformat by applying local time-zone from `TimeZoneProvider`, so I think it would be a bigger change. Any suggestions will be much appreciated here.

Keeping it "as is", in current version of the code I have provided, will be correct in all cases where system time-zone is the same as the time-zone configured in openHAB. When they are different and a state is provided using the correct time-zone from `TimeZoneProvider` (i.e. from bindings correctly implementing this), it would have been correctly displayed in local time-zone previously, but will now be in system time-zone.

jlaur:
@wborn - can you help me by removing the "work in progress" label to signal this is now ready for the review process? 🙂

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/item/EnrichedItemDTOMapper.java
```diff
@@ -51,6 +56,10 @@ public class EnrichedItemDTOMapper {
 
     private static final Pattern EXTRACT_TRANSFORM_FUNCTION_PATTERN = Pattern.compile("(.*?)\\((.*)\\):(.*)");
 
+    private static final String DATE_FORMAT_PATTERN_WITH_TZ_RFC = "yyyy-MM-dd'T'HH:mm[:ss[.SSSSSSSSS]]Z";
+    private static final DateTimeFormatter FORMATTER_TZ_RFC = DateTimeFormatter
```
joerg1985, line 60: Is this pattern identical to DateTimeFormatter.ISO_OFFSET_DATE_TIME? In any case the .SSSSSSSSS is incorrect, should be .nnnnnnnnn in my mind.

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/item/EnrichedItemDTOMapper.java
```diff
@@ -134,6 +155,24 @@ private static EnrichedItemDTO map(Item item, ItemDTO itemDTO, boolean drillDown
...
+            String sign = "";
+            if (formatted.contains("+")) {
+                sign = "+";
+            } else if (formatted.contains("-")) {
+                sign = "-";
+            }
+            if (!sign.isEmpty()) {
+                // the formatted string contains 9 fraction-of-second digits
+                // truncate at most 2 trailing groups of 000s
+                return formatted.replace("000" + sign, sign).replace("000" + sign, sign);
```
joerg1985, line 170: With the formater mentioned above this should not be needed.

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/item/EnrichedItemDTOMapper.java
jlaur: I have to admit this was duplicated from `DateTimeType` itself, thinking I would later remove it there - which didn't happen. I'll see if it can be refactored while maintaining backwards compatiblity, and also get rid of the duplication.

jlaur: In `DateTimeType` I have now exposed a method overload of `toFullString` taking a `ZoneId` - and added full test coverage for this. As a consequence, this duplicated method and `DateTimeFormatter` was removed.

For now, I have left the method and `DateTimeFormatter` as it was before this PR, but we now have test coverage to make sure refactoring will not break the backwards compatibility.

jlaur: It seems `DateTimeFormatter.ISO_OFFSET_DATE_TIME` does not provide the same output:
```
[ERROR] Failures: 
[ERROR]   DateTimeTypeTest.createDate:360 expected: <2014-03-30T10:58:47.033+0000> but was: <2014-03-30T10:58:47.033Z>
[ERROR]   DateTimeTypeTest.createDate:360 expected: <2014-03-30T12:58:47.033+0200> but was: <2014-03-30T12:58:47.033+02:00>
[ERROR]   DateTimeTypeTest.createDate:360 expected: <2014-03-30T12:58:47.033+0200> but was: <2014-03-30T12:58:47.033+02:00>
[ERROR]   DateTimeTypeTest.createDate:360 expected: <2014-03-30T05:58:47.033-0200> but was: <2014-03-30T05:58:47.033-02:00>
[ERROR]   DateTimeTypeTest.createDate:360 expected: <2014-03-30T12:58:47.033-0200> but was: <2014-03-30T12:58:47.033-02:00>
[ERROR]   DateTimeTypeTest.createDate:360 expected: <1970-01-01T10:58:47.000+0000> but was: <1970-01-01T10:58:47Z>
[ERROR]   DateTimeTypeTest.createDate:360 expected: <1970-01-01T10:58:00.000+0000> but was: <1970-01-01T10:58:00Z>
[ERROR]   DateTimeTypeTest.createDate:360 expected: <1970-01-01T10:58:47.000+0100> but was: <1970-01-01T10:58:47+01:00>
[ERROR]   DateTimeTypeTest.createDate:360 expected: <1970-01-01T10:58:00.000+0100> but was: <1970-01-01T10:58:00+01:00>
[ERROR]   DateTimeTypeTest.createDate:360 expected: <2014-03-30T00:00:00.000+0000> but was: <2014-03-30T00:00:00Z>
[ERROR]   DateTimeTypeTest.createDate:360 expected: <2014-03-30T10:58:47.033+0000> but was: <2014-03-30T10:58:47.033Z>
[ERROR]   DateTimeTypeTest.createDate:360 expected: <2014-03-30T08:58:47.033+0000> but was: <2014-03-30T08:58:47.033Z>
[ERROR]   DateTimeTypeTest.createDate:360 expected: <2014-03-30T10:58:47.230+0000> but was: <2014-03-30T10:58:47.23Z>
[ERROR]   DateTimeTypeTest.createDate:360 expected: <2014-03-30T10:58:47.000+0000> but was: <2014-03-30T10:58:47Z>
[ERROR]   DateTimeTypeTest.createDate:360 expected: <2014-03-30T12:58:47.033+0200> but was: <2014-03-30T12:58:47.033+02:00>
[ERROR]   DateTimeTypeTest.createDate:360 expected: <2014-03-30T10:58:47.033+0200> but was: <2014-03-30T10:58:47.033+02:00>
[ERROR]   DateTimeTypeTest.createDate:360 expected: <2014-03-30T10:58:47.000+0200> but was: <2014-03-30T10:58:47+02:00>
[ERROR]   DateTimeTypeTest.createDate:360 expected: <2014-03-30T15:58:47.000+0500> but was: <2014-03-30T15:58:47+05:00>
[ERROR]   DateTimeTypeTest.createDate:360 expected: <2014-03-30T10:58:47.000+0200> but was: <2014-03-30T10:58:47+02:00>
[ERROR]   DateTimeTypeTest.toFullStringWithZone:412 
Expected: is "2024-11-11T20:39:00.000+0000"
     but: was "2024-11-11T20:39:00Z"
[ERROR]   DateTimeTypeTest.toFullStringWithZone:412 
Expected: is "2024-11-11T20:39:00.000+0000"
     but: was "2024-11-11T20:39:00Z"
[ERROR]   DateTimeTypeTest.toFullStringWithZone:412 
Expected: is "2024-11-11T20:39:00.000000001+0000"
     but: was "2024-11-11T20:39:00.000000001Z"
[ERROR]   DateTimeTypeTest.toFullStringWithZone:412 
Expected: is "2024-11-11T20:39:00.123+0000"
     but: was "2024-11-11T20:39:00.123Z"
[ERROR]   DateTimeTypeTest.toFullStringWithZone:412 
Expected: is "2024-11-11T20:39:00.123456+0000"
     but: was "2024-11-11T20:39:00.123456Z"
[ERROR]   DateTimeTypeTest.toFullStringWithZone:412 
Expected: is "2024-11-11T20:39:00.123456789+0000"
     but: was "2024-11-11T20:39:00.123456789Z"
[ERROR]   DateTimeTypeTest.toFullStringWithZone:412 
Expected: is "2024-11-11T21:39:00.123+0100"
     but: was "2024-11-11T21:39:00.123+01:00"
[ERROR]   DateTimeTypeTest.toFullStringWithZone:412 
Expected: is "2024-11-10T23:59:59.999-0500"
     but: was "2024-11-10T23:59:59.999-05:00"
```

joerg1985: The parser test does accept the format above as valid input, so why should this be invalid output?

jlaur:
> which is probably the reason for system time-zone being used in some places. I don't know how to improve this, since `TimeZoneProvider` is required for properly formatting this in the configured local time-zone. One solution might be to change the implementation of `DateTimeType.toFullString` to format in UTC as `Instant`. However, this would require all receivers to reformat by applying local time-zone from `TimeZoneProvider`, so I think it would be a bigger change. Any suggestions will be much appreciated here.

This was not the case. I finally figured it out and fixed this inconsistent behavior, so that configured time-zone rather than system time-zone is now used for SSE item events and sitemaps as well.

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/item/EnrichedItemDTOMapper.java
jlaur: With da2b50a2e71da3152248b349a4fc77a11db81ead this duplicated method was removed, so what is left is the original implementation of `toString()`/`toFullString()`. This is unchanged by this PR to remain backwards compatible. The `DateTimeType` String constructor supports many different inputs, but `toFullString()` can only provide one specific output.

jlaur:
> > Is it really backward compatible with all bindings?
> 
> It should be, but there can be unit tests failing if they expect a certain time-zone:
> 
> ```java
> DateTimeType dt = new DateTimeType(ZonedDateTime.now());
> ZonedDateTime zdt = dt.getZonedDateTime();
> ```
> 
> Previously this would return a `ZonedDateTime` with the same zone as provided when creating the object. But since the zone is now stripped, it will return a `ZonedDateTime` with system time-zone. Generally `getInstant()` should be preferred over `getZonedDateTime()` now.

I have now checked all bindings by recompiling **main** against changes of this PR and fixed those tests having that assumption: openhab/openhab-addons#17764.

mherwege:
@jlaur I looked at the code, and there is nothing I can add to this. I agree this will simplify addon development and make things more consistent.
The one worry I have is that we are getting close to the 4.3 release. I am not sure we will have enough milage in a milestone for this. It doesn't look like there is much impact, but touching a core type always carries some risk. But I let maintainers be the judge on this.

From my perspective LGTM.

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/item/EnrichedItemDTOMapper.java
jlaur: See https://github.com/openhab/openhab-core/pull/3583#discussion_r1838883285

jlaur:
@openhab/core-maintainers - WDYT?

J-N-K (changes requested):

bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/DateTimeGroupFunctionTest.java
```diff
@@ -36,36 +37,36 @@ public class DateTimeGroupFunctionTest {
 
     @Test
     public void testLatestFunction() {
-        ZonedDateTime expectedDateTime = ZonedDateTime.now();
+        Instant expectedDateTime = Instant.now();
```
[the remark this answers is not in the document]
jlaur: It was not necessary to change the tests, they would still pass. It was a consequence of https://github.com/openhab/openhab-core/pull/3583#discussion_r1865703710. If you prefer to keep `getZonedDateTime()` **and** prefer to use this in the tests, I can change it back. Let me know.

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/module/handler/ItemStateConditionHandler.java
```diff
@@ -158,9 +159,9 @@ private boolean lessThanOrEqualsToItemState(String itemName, String state) throw
         Item item = itemRegistry.getItem(itemName);
         State compareState = TypeParser.parseState(item.getAcceptedDataTypes(), state);
         State itemState = item.getState();
-        if (itemState instanceof DateTimeType type) {
-            ZonedDateTime itemTime = type.getZonedDateTime();
-            ZonedDateTime compareTime = getCompareTime(state);
+        if (itemState instanceof DateTimeType dateTimeState) {
+            Instant itemTime = dateTimeState.getInstant();
+            Instant compareTime = getCompareTime(state).toInstant();
```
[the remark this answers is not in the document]
jlaur: Done.

bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/RuleResource.java
```diff
@@ -431,9 +435,9 @@ public Response simulateRules(
         return Response.ok(ruleExecutions.toList()).build();
     }
 
-    private static ZonedDateTime parseTime(String sTime) {
+    private ZonedDateTime parseTime(String sTime) {
```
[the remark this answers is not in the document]
jlaur: Done. I introduced a `Supplier` though to avoid instantiating the default objects when they are not needed.

bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/DateTimeGroupFunctionTest.java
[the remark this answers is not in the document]
jlaur: Just to be 100% sure, I reverted and ran the tests, and as expected, they still passed. I would prefer to stick with this simplified version, because it's more to the point without unnecessary conversions:
- `ZonedDateTime.now()` creates a `ZonedDateTime` using `Instant.now()` with system default time-zone.
- `DateTimeType(ZonedDateTime)` strips the time-zone.
- `DateTimeType.getZonedDateTime()` return a `ZonedDateTime` using internal `Instant` with system default time-zone applied.

These conversions are already covered by `DateTimeTypeTest` methods, so I don't see the value of including them here also when testing the group function logic.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
```diff
@@ -73,70 +73,91 @@ public class DateTimeType implements PrimitiveType, State, Command, Comparable<D
...
-        zonedDateTime = date.withFixedOffsetZone();
     }
 
+    /**
+     * Get object represented as a {@link ZonedDateTime} with system
+     * default time-zone applied
+     *
+     * @return a {@link ZonedDateTime} representation of the object
+     */
     public ZonedDateTime getZonedDateTime() {
```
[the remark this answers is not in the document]
jlaur: Would this be an acceptable compromise, i.e. introducing the overload you suggested, but also marking the existing one as deprecated?
```java
    /**
     * @deprecated
     *             Get object represented as a {@link ZonedDateTime} with system
     *             default time-zone applied
     *
     * @return a {@link ZonedDateTime} representation of the object
     */
    @Deprecated
    public ZonedDateTime getZonedDateTime() {
        return getZonedDateTime(ZoneId.systemDefault());
    }

    /**
     * Get object represented as a {@link ZonedDateTime} with the
     * the provided time-zone applied
     *
     * @return a {@link ZonedDateTime} representation of the object
     */
    public ZonedDateTime getZonedDateTime(ZoneId zoneId) {
        return instant.atZone(zoneId);
    }
```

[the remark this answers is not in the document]
jlaur: To speed things up, please see 36c02028d144d34e9440d10250f02064636a9d95. I can easily revert or rework it.

jlaur:
@J-N-K - thanks for picking this up for review. I have tried to address all your review comments. I have force-pushed after rebasing towards main, but all today's changes are in the last two commits.

kaikreuzer (approved):
Many thanks!

PR #3588, opened 2023-05-01
Allow OnOffType as valid conversion types for HSBType and PercentType

kaikreuzer:
This fixes a regression that has been recently introduced (most likely through the refactoring of the CommunicationManager).
I couldn't figure out what change exactly causes the regression, but at least this PR fixes it.

The bug is the following: I have a Color item with a Switch channel attached using the Follow profile - this controls an LED on my KNX wall switch to show whether the light is on or off. In the past, a HSBType state update was correctly converted to ON/OFF and sent out to the channel, which is no longer the case with the current code base.

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/CommunicationManager.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/CommunicationManager.java
@@ -46,6 +46,8 @@ import org.openhab.core.items.events.ItemCommandEvent;
 import org.openhab.core.items.events.ItemStateUpdatedEvent;
 import org.openhab.core.library.items.NumberItem;
 import org.openhab.core.library.types.DecimalType;
+import org.openhab.core.library.types.OnOffType;
+import org.openhab.core.library.types.PercentType;
 import org.openhab.core.library.types.QuantityType;
 import org.openhab.core.thing.Channel;
 import org.openhab.core.thing.ChannelUID;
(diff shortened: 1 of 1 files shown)


J-N-K:
I think the reason is that the `CommunicationManager` now uses the `ItemStateUpdatedEvent` instead of the `ItemStateEvent`. The `ItemStateUpdatedEvent` contains the new state of the item, which is a `HSBType`, not an `OnOffType` which was used to update the item.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/CommunicationManager.java
```diff
@@ -467,6 +469,9 @@ && channelTypeDefinesDimension(channel.getChannelTypeUID())) {
 
         if (acceptedTypes.contains(originalType.getClass())) {
             return originalType;
+        } else if (acceptedTypes.contains(OnOffType.class) && originalType instanceof State state
+                && PercentType.class.isAssignableFrom(originalType.getClass())) {
+            return (@Nullable T) state.as(OnOffType.class);
         } else {
```
J-N-K, line 475: Should we also add code here for `PercentType.class` so that we can link a `Color` item to a `Dimmer` channel in the same way?

kaikreuzer:
Hm, the failing tests have a point - I'll have to further look into this.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/CommunicationManager.java
kaikreuzer: Ok, I think I now understood what you meant - and it should fix the tests.

J-N-K (approved):
LGTM, thanks!

PR #3591, opened 2023-05-03
Make --dir param optional for upgradetool

kaikreuzer:
Almost all openHAB installations should have the $OPENHAB_USERDATA env variable set. If this is the case, the user does not have to provide a directory, so this parameter can be made optional.


J-N-K (approved):
Thanks

PR #3592, opened 2023-05-05
Fix non-unique module handler identifiers

### LINKED ISSUE #3505: Wrong Rule triggered by item
when I have a rule with more than 10 elements that can trigger the rule, the wrong rule is executed.
See also: https://community.openhab.org/t/rule-triggered-by-wrong-item/125894/14?u=pmpfm

J-N-K:
Fixes #3505 

The cause is the naming of the module handlers. Internally the `BaseModuleHandlerFactory` stores the modules in a `Map` where the key is `ruleUid+moduleId`. 

Textual defined rules have `ruleUid` that consists of the filename, a dash and a number that denotes the position of the rule in the file, so in a file `foo.rules`, the first rule gets a `foo-1`, the 11th gets `foo-11`. Within a rule, the modules are numbered from 0 to n. So the if the first rule has 10 triggers, the last module handler is stored with a key `foo-110`. When the 11th rule is added and has a trigger of the same type as first trigger, the `BaseModuleHandlerFactory` generates the same key and returns the already created handler for 10th trigger of rule 1 instead creating a new handler.

This PR changes the key to `<ruleUid>$<moduleId>`. This should be safe because `$` is not an allowed character for the module id and collisions are prevented.

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/handler/BaseModuleHandlerFactory.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/handler/BaseModuleHandlerFactory.java
@@ -60,7 +60,7 @@ public abstract class BaseModuleHandlerFactory implements ModuleHandlerFactory {
     @Override
     @SuppressWarnings("null")
     public @Nullable ModuleHandler getHandler(Module module, String ruleUID) {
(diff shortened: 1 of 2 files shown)


kaikreuzer (approved):
Thanks, interesting effect.

PR #3593, opened 2023-05-05
Improve message when rule action fails

### LINKED ISSUE #3468: Include a more meaningful error when a script action fails
## Background 

I have a rule that gets pounded sometimes, keeping up with a bunch of sensors to generate an alert when one fails to report after a time. When it does need to alert, it calls another rule to handle the alert (which may or may not be relevant here) and that rule is called from a timer. 

Occasionally I'll see the following error in my logs:

```
2023-03-20 13:18:18.274 [ERROR] [e.automation.internal.RuleEngineImpl] - Failed to execute rule 'sensor-status-detection': Fail to execute action: 1
2023-03-20 13:18:18.292 [WARN ] [b.automation.openhab-js.sensor alert] - ALERT: Family Room Wave plus is back online
```

The last significant line of code run in `sensor-status-detection` is the call to this rule. After that a couple of flags are set and the looping timer is rescheduled (or not as appropriate). You can find the full rule at https://raw.githubusercontent.com/rkoshak/openhab-rules-tools/main/rule-templates/thresholdAlert/newThresholdAlert.yaml (it's big and complicated but I also think it's not actually relevant here except as background).

**Even when the error appears, the rule appears to execute correctly.**

## Source Code

So I figured I'd try to figure out what the root cause of the error is. I traced it down to the following location 

https://github.com/openhab/openhab-core/blob/8d64ecfd8db99844d79a776fef37cbe527c1f0c6/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleEngineImpl.java#L996 

which is just a kind of catch-all. Digging deeper leads to 

https://github.com/openhab/openhab-core/blob/8d64ecfd8db99844d79a776fef37cbe527c1f0c6/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleEngineImpl.java#L1186

Here's more context

```
                try {
                    Map<String, ?> outputs = aHandler.execute(Collections.unmodifiableMap(context));
                    if (outputs != null) {
                        context = getContext(ruleUID, null);
                        updateContext(ruleUID, action.getId(), outputs);
                    }
                } catch (Throwable t) {
                    String errMessage = "Fail to execute action: " + action.getId();
                    if (stopOnFirstFail) {
                        RuntimeException re = new RuntimeException(errMessage, t);
                        throw re;
                    } else {
                        logger.warn(errMessage, t);
                    }
                }
```

There's not a lot that can go wrong here as far as I can tell. `ActionHandler.execute` doesn't declare to throw any exceptions and there doesn't appear to be one in `WrappedActionHandler`. 

`updateContext()` doesn't appear to have anything in it that could generate an exception.

`getContext(ruleUID, null)` could throw an exception: see https://github.com/openhab/openhab-core/blob/8d64ecfd8db99844d79a776fef37cbe527c1f0c6/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleEngineImpl.java#L1097 and in fact if I'm seeing that exception the message clearly indicates a bug that should be reported. I didn't go any deeper than this so maybe the exception is coming from the `ReferenceResolver`.

## Proposed Change

Can we update the error message from above to include the message from the exception thrown rather than simply suppressing it with a generic "Failed to execute action".

    String errMessage = "Failed to execute action: " + action.getId() + " - " + t.getMessage();

Especially if the only exception I found in this block has the message "please report a bug" we can't do that if we don't see the message.

J-N-K:
Fixes #3468 

Also logs the full stacktrace at DEBUG level

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleEngineImpl.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleEngineImpl.java
@@ -1183,8 +1183,9 @@ public class RuleEngineImpl implements RuleManager, RegistryChangeListener<Modul
                         updateContext(ruleUID, action.getId(), outputs);
                     }
                 } catch (Throwable t) {
-                    String errMessage = "Fail to execute action: " + action.getId();
+                    String errMessage = "Failed to execute action: " + action.getId() + "(" + t.getMessage() + ")";
                     if (stopOnFirstFail) {
+                        logger.debug("Action {}-{} threw an exception: ", ruleUID, action.getId(), t);
                         throw new RuntimeException(errMessage, t);
                     } else {
                         logger.warn(errMessage, t);


kaikreuzer (approved):
lgtm, thanks.

PR #3598, opened 2023-05-09
Fix storing of StateDescriptionFragment in AbstractStorageBasedTypeProvider

J-N-K:
@lolodomo FYI.

Since `StateDescriptionFragment` is an interface it can't be restored from the storage without adding type specific converters. Since `JsonStorage` does not provide a solution for that, the `StateDescriptionFragment` is now mapped to a storage entity.


kaikreuzer (approved):

lolodomo:
Thank you, I will try tomorrow with the new snapshot.

PR #3599, opened 2023-05-09
Upgrade SAT to 0.14.0

holgerfriedrich:
New static code analysis tools package has been released, lifting the versions of 
* PMD from 6.39.0 to 6.53,
* SpotBugs from 4.7.0 to 4.7.3,
* Checkstyle from 8.45.1 to 10.6.0;
and to adapt to schema changes in OH 4.0.

Looking at the check results, there are no substantial differences between 0.13 and 0.14, see the comparison below:
grafik


holgerfriedrich:
Remark: Unfortunately, we still see false positives "Self comparison of field with itself (SA)" for the java17 instancof with name pattern. This is still unresolved in upstream repo spotbugs (spotbugs/spotbugs#1992).

wborn (approved):
Thanks!

PR #3609, opened 2023-05-13
Reduce loglevel in UnitUtils

J-N-K:
The log level was increased during the UoM refactoring but it turns out that a lot of code depends on failing silently. Therefore the log level is again reduced to DEBUG.


lolodomo:
@kaikreuzer : to be merged before milestone 3 please ;)

kaikreuzer (approved):

Issue #3562, opened 2023-04-17
During startup the system applies transforms before transform addons are loaded


mhilbush:
During startup, it appears that the system tries to apply transforms to items before the transform addons are loaded.

For example, if I define an item like this to use a map transform.
```
Switch  PresenceMark  "Mark [MAP(presence.map):%s]"  {channel="unifi:wirelessClient:mark:online"}
```

At startup, I see the following in the log.
```
2023-04-17 11:43:27.402 [WARN ] [rest.core.item.EnrichedItemDTOMapper] - Failed transforming the state 'NULL' on item 'PresenceMark' with pattern 'MAP(presence.map):%s': Couldn't transform value because transformation service of type 'MAP' is not available.
```

It also occurs with my Javascript transforms.

I'm currently running build 3417. However, I've been seeing this for quite a while in the 4.0 snapshots, but only now getting around to reporting it. I suspect something in the startup order has changed such that the system is attempting to apply the map transform on the item, but the map transform addon has not been loaded yet. I seem to recall there were some changes made to the startup order / run levels (persistence, others?), but I don't recall the exact change(s).

rkoshak:
Is it every boot or only after a clearing of the cache. 

After clearing the cache, I see that and even complaints that Js Scripting isn't available for rules that get triggered. However, after giving it a chance to settle and install all the add-ons, a restart of OH causes all the errors to go away. Because it takes some time to download and install the add-ons, this behavior makes a little bit of sense. 

I've seen this almost since the beginning of OH 4 but it's not consistent. Sometimes I need to restart after an update and sometimes not.

mhilbush:
Every boot except for the very first one after clearing cache.

rkoshak:
> Every boot except for the very first one after clearing cache.

So your behavior is exactly the opposite of what I'm seeing. 🤣

mhilbush:
> So your behavior is exactly the opposite of what I'm seeing. 

Yeah that's very strange.

> rest.core.item.EnrichedItemDTOMapper as logger seems to indicate that "someone" is accessing the REST API during startup

That's very interesting.

I have HABpanel running on a tablet, which I typically don't exit when doing a stop/start. This time I exited HABpanel before starting OH and the WARNs didn't appear. So I guess we can assume that it's HABpanel making REST API calls.

What's odd is that the WARNs always get logged at exactly the same place during startup process - after items, persistence and things files, but before rules files.

J-N-K:
[...]

PR #3611, opened 2023-05-13
[rest] Add endpoint for UoM information

J-N-K:
Currently the list of available dimensions for `Number` items is hard-coded in Main UI. The reason is that there is no way to determine the available dimensions at runtime. This endpoint adds that, so that the list can be created dynamically and will automatically be up-to-date if new dimensions (like #3608 or #3408) are added.

### DIFF
--- a/bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/internal/resources/SystemInfoResource.java
+++ b/bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/internal/resources/SystemInfoResource.java
@@ -23,9 +23,11 @@ import javax.ws.rs.core.UriInfo;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.openhab.core.auth.Role;
+import org.openhab.core.i18n.UnitProvider;
 import org.openhab.core.io.rest.RESTConstants;
 import org.openhab.core.io.rest.RESTResource;
 import org.openhab.core.io.rest.internal.resources.beans.SystemInfoBean;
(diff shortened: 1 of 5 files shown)


J-N-K:
@ghys @florian-h05 FYI.

florian-h05:
Looks like a great addition, the ability to get the default unit is something that could be really helpful for the user to choose the right dimension. Often you know the unit, but don‘t know the dimension.

kaikreuzer (approved):
lgtm, thanks!

florian-h05:
@J-N-K Is there any way to register a unit system change listener? I’d like to implement caching like in https://github.com/openhab/openhab-core/pull/3335.

florian-h05:
@J-N-K Can you please check whether you can access that endpoint on current snapshot?

J-N-K:
I have checked with the snapshot before OH4 release and there it works.

J-N-K:
Regarding the unit change listener: My recommendation would be to add a big red warning that changing the unit system change requires a restart. There are other pitfalls when doing so which probably can't be easily fixed.

florian-h05:
For me it doesn't work on openHAB 4.1.0 (build Build #3562):

```
florianh@zephyrus-fh211 ~> http http://localhost:8080/rest/uom
HTTP/1.1 404 Not Found
Allow: DELETE,POST,GET,PUT,OPTIONS,HEAD
Content-Length: 0
Date: Mon, 24 Jul 2023 15:49:52 GMT
Server: Jetty(9.4.50.v20221201)
```

J-N-K:
It's `rest/systeminfo/uom`.

PR #3612, opened 2023-05-13
Keep channel properties in thing updates

### LINKED ISSUE #3610: thing upgrades: missing support to add/change/remove channel properties
I am using channel properties and as part of a change I also need to add some properties as part of a migration. See [#14866 ](https://github.com/openhab/openhab-addons/pull/14866) for details. Changing channel type worked fine for me but this is only a part of the migration. Please add support for this.

J-N-K:
Fixes #3610

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/update/UpdateChannelInstructionImpl.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/update/UpdateChannelInstructionImpl.java
@@ -123,6 +123,7 @@ public class UpdateChannelInstructionImpl implements ThingUpdateInstruction {
             if (oldChannel != null) {
                 channelBuilder.withConfiguration(oldChannel.getConfiguration());
                 channelBuilder.withDefaultTags(oldChannel.getDefaultTags());
+                channelBuilder.withProperties(oldChannel.getProperties());
             }
         }
 


lolodomo:
Thank you @J-N-K .

I have the feeling you could also fix the problem for label/description with that code ?
```
        if (label != null) {
            channelBuilder.withLabel(Objects.requireNonNull(label));
        } else {
            channelBuilder.withLabel(oldChannel.getLabel());
        }

        if (description != null) {
            channelBuilder.withDescription(Objects.requireNonNull(description));
        } else {
            channelBuilder.withDescription(oldChannel.getDescription());
        }
```

J-N-K:
No. That would most probably be wrong, because the new channel would inherit from the old one instead of the channel-type.

lolodomo:
> No. That would most probably be wrong, because the new channel would inherit from the old one instead of the channel-type.

Oh yes, you're right but that is exactly the same thing for properties, we need to find properties from the new channel, not the old one, that is the use case from @alexf2015, he added channel properties.

J-N-K:
We could probably add a new section similar to label that adds new properties.

kaikreuzer:
@alexf2015 You mentioned [here](https://github.com/openhab/openhab-core/issues/3610#issuecomment-1546948150) that you don't think that this PR fixes your issue. Shall it be merged nonetheless or do you expect further changes?

kaikreuzer (approved):
Thanks!

PR #3616, opened 2023-05-15
[UoM] add unit rpm for rotational frequency

boehan:
Adds unit `rpm` (revolutions per minute) for rotational frequency. Even though I'd prefer `min⁻¹`, `rpm` seems to be more common.
Please let me know, if I missed something. I have to admit that this is untested as core does currently not build for me locally.


clinique (approved):
LGTM

J-N-K (approved):
LGTM, thanks

PR #3618, opened 2023-05-18
Fix only first group config considered in PersistenceManager

### LINKED ISSUE #3617: Item value no more persisted => problem in persistence strategy loading ?
I have a persisted switch item (RRD4j).
This item is linked to nothing, it is used as a parameter for my rules.
When I restart OH, its value is not restored (I see nothing in events.log).
It was set to OFF using UI before stopping OH.

It is with snapshot 3470 but I believe the problem appeared in previous snapshots.

As there were only one change regarding RRD4j (making the service writing asynchronously to DB), I suspect that change.

J-N-K:
Fixes #3617


lolodomo:
Tested and I can confirm it fixes the regression I noticed yesterday.
I believe it should be tagged "Critical".
@openhab/core-maintainers : please merge ASAP and at least before the release of milestone 3.

lolodomo:
@kaikreuzer : there was no new snapshot built this morning while core was properly built. Can you please trigger a new snapshot manually? This will allow testing this fix inside a real snapshot.

kaikreuzer:
@lolodomo Done.

PR #3621, opened 2023-05-19
Fix Upgradetool does not remember last executed step

J-N-K:
The JSON database that stores the already executed steps was not always properly flushed. This could result in the last step not being properly persisted, resulting in a new execution on next update.


kaikreuzer (approved):
thanks

Issue #3305, opened 2023-01-09
Add Role Bases Access Control


hmerk:
See community discussion and proposal
https://community.openhab.org/t/rbac-model-in-openhab-and-potential-security-vulnerability-found/136419

hmerk:
@openhab/core-maintainers Would you please add this to the project list for openHAB 4.0.
This would be a very good addition.

hmerk:
Could add it myself ;-)

splatch:
@hmerk while not being a core contributor I might be able to work on this in my spare time based on earlier work I did with Open Smart House.

I outlined some of concepts in forums https://community.openhab.org/t/rbac-model-in-openhab-and-potential-security-vulnerability-found/136419/4. They are based on initial implementation of eclipse/smarthome#6034 which covered first step of design made under eclipse/smarthome#579. Long story short - we got authentication API which was adjusted/stripped for openHAB 3.0, we never reached step 2 which would involve wider support for user roles and permissions.

hmerk:
@splatch 
To be honest, I don't care who is doing this. Just asked in the community, as it sounded like ready, just needs some code cleaning. So why not join forces with OP ?

splatch:
@hmerk I could argue on some of statements made there, ie.:

> The openHAB community has been visited 93 days and 257 topics and 1 800 posts have been viewed with a total read time of 22 hours. The activity in this community demonstrates the difficulty of incorporating the RBAC model into openHAB and the complexity of the platform.

;-)

hmerk:
@splatch  Sorry, i might have been a bit unclear.
I have asked the OP if he could provide a PR for this, to have it implement in oh4.
https://community.openhab.org/t/rbac-model-in-openhab-and-potential-security-vulnerability-found/136419/12
He answered that the code needs some cleanup and he is going to provide a PR in de next weeks.
https://community.openhab.org/t/rbac-model-in-openhab-and-potential-security-vulnerability-found/136419/13

So under the bottom line, I only wanted to prevent anyone from double work...

splatch:
Going over code made by @gennartn I found several troubles. There are multiple commits, including merges, which make moving it into current code base troublesome as we would need to make a "back merge" changes which are started on 3.2.x state of auth packages and UI. Both core and ui moved forward since then. I attempted to rebase/squash commits but its not going straight as there are conflicting changes between commits.
[...]

PR #3622, opened 2023-05-21
[WebSocket] Allow register websocket handlers

GiviMAD:
Hi openHAB community, I would like to merge this PR which allows to register multiple WebSocket handlers in to a common servlet which handles the authentication.

I have refactored the already present implementation into a separate handler (EventWebSocketHandler) that is served in the root path '/ws' so I think this is not a breaking change.

Also I have made some methods in the AuthFilter (package org.openhab.core.io.rest.auth) public so they can be reused here, so the "Trusted Network" and "Allow Basic Auth" rest api options apply also to the WebSocket and this can be accessed with the same credentials than the rest api (before the commit it allowed user tokens and basic auth but no JWT tokens).

Hope everything is in place. Best regards!


GiviMAD:
Seems like these changes break the core karaf feature, I'm trying to understand why, I don't have experience with these definitions. If someone can give me some some help will be appreciated.

GiviMAD:
Luckily i was able to fix it :)

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/AnonymousUserSecurityContext.java
```diff
@@ -10,7 +10,7 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  */
-package org.openhab.core.io.rest.auth.internal;
+package org.openhab.core.io.rest.auth;
```
splatch, line 13: This security context was introduced in OH 3.0 and it does work around other security related API/SPI which was present earlier - `AuthenticationSecurityContext`. If you decide to go with jaxrs API as a main driver for security in websockets in servlets you shall deprecate other classes.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/AuthFilter.java
```diff
@@ -266,29 +268,33 @@ public void filter(@Nullable ContainerRequestContext requestContext) throws IOEx
                             }
                         }
                     }
-                } else if (isImplicitUserRole(requestContext)) {
+                } else if (isImplicitUserRole(servletRequest)) {
                     requestContext.setSecurityContext(new AnonymousUserSecurityContext());
                 }
             } catch (AuthenticationException e) {
-                logger.warn("Unauthorized API request from {}: {}", getClientIp(requestContext), e.getMessage());
+                logger.warn("Unauthorized API request from {}: {}", getClientIp(servletRequest), e.getMessage());
```
splatch, line 275: Is there any particular reason why we do not log stack trace here?

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/AuthFilter.java
```diff
@@ -181,7 +183,7 @@ protected void deactivate() {
         }
     }
 
-    private SecurityContext authenticateBearerToken(String token) throws AuthenticationException {
+    public SecurityContext authenticateBearerToken(String token) throws AuthenticationException {
         if (token.startsWith(API_TOKEN_PREFIX)) {
             UserApiTokenCredentials credentials = new UserApiTokenCredentials(token);
             Authentication auth = userRegistry.authenticate(credentials);
```
splatch, line 189: While it is not related to your changes, this call should call `org.openhab.core.auth.AuthenticationManager` and not user registry itself.

bundles/org.openhab.core.io.websocket/pom.xml
```diff
@@ -20,6 +20,11 @@
       <artifactId>org.openhab.core</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.openhab.core.bundles</groupId>
+      <artifactId>org.openhab.core.io.rest.auth</artifactId>
```
splatch, line 25: This dependency brings you a lot of stuff, making websocket part embedding a lot of jaxrs dependencies. From this point of view, if you collapse dependencies you could try to do websockets at jaxrs level. it will be slower, but will keep things consistent.

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/CommonWebSocketServlet.java
```diff
@@ -0,0 +1,149 @@
...
+    public static final String DEFAULT_HANDLER_ID = EventWebSocketHandler.HANDLER_ID;
+
+    private final Map<String, WebSocketHandler> connectionHandlers = new HashMap<>();
+    private final AuthFilter authFilter;
+
+    @SuppressWarnings("unused")
+    private @Nullable WebSocketServerFactory importNeeded;
+
+    @Activate
+    public CommonWebSocketServlet(@Reference AuthFilter authFilter) throws ServletException, NamespaceException {
```
splatch, line 70: Use `org.openhab.core.auth.AuthenticationManager` instead of filter, filter itself is an entry point to authentication process, not its driver.

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/CommonWebSocketServlet.java
```diff
@@ -0,0 +1,149 @@
...
+                        servletUpgradeRequest.getRemoteAddress());
+            }
+            return null;
+        }
+
+        private boolean isAuthorizedRequest(ServletUpgradeRequest servletUpgradeRequest) {
+            Map<String, List<String>> parameterMap = servletUpgradeRequest.getParameterMap();
+            List<String> accessToken = parameterMap.getOrDefault("accessToken", List.of());
+            SecurityContext securityContext = null;
+            if (accessToken.isEmpty() && authFilter.isImplicitUserRole(servletUpgradeRequest.getHttpServletRequest())) {
```
splatch, line 130: This code builds assumptions on a top of another component configuration. It would be better to centralize that part elsewhere.

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/EventWebSocketHandler.java
```diff
@@ -0,0 +1,81 @@
...
+import com.google.gson.Gson;
+
+/**
+ * The {@link EventWebSocketHandler} allows subscription to oh events over WebSocket
+ *
+ * @author Jan N. Klug - Initial contribution
+ */
+@NonNullByDefault
+@Component(immediate = true, service = { EventSubscriber.class, WebSocketHandler.class })
+public class EventWebSocketHandler implements EventSubscriber, WebSocketHandler {
```
splatch, line 39: Just a small suggestion - maybe it would sound better as `EventWebSocketBridgeHandler` or `EventWebSocketAdapterHandler`?

features/karaf/openhab-core/src/main/feature/feature.xml
```diff
@@ -181,6 +181,8 @@
 
 	<feature name="openhab-core-io-websocket" version="${project.version}">
 		<feature>openhab-core-base</feature>
+		<feature dependency="true">openhab-core-io-rest-auth</feature>
```
splatch, line 184: You can stick with `openhab-core-base` if you decide to go with `AuthenticationManager`, cause it will be provided earlier than rest filter.

splatch (commented):
I have some remarks on direction in which use of security goes. Please consider them. Driving further adoption of rest filter as a main access control mechanism will make it difficult to adopt different authentication mechanisms requested in #3305.

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/CommonWebSocketServlet.java
GiviMAD: The reason to use use the AuthFilter over the AuthenticationManager is to make it behave in the same way than the rest api and respect its configuration (allowBasicAuth, implicitUserRole and trustedNetworks).
Do you think is it better to refactor that functionality into a new class and make use of it in both places?

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/CommonWebSocketServlet.java
GiviMAD: Yes, that answers the question on the previous comment.

I also saw this comment:

> This dependency brings you a lot of stuff, making websocket part embedding a lot of jaxrs dependencies. From this point of view, if you collapse dependencies you could try to do websockets at jaxrs level. it will be slower, but will keep things consistent.

Do you think that creating a new bundle with just the required stuff could be a good idea?

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/AnonymousUserSecurityContext.java
GiviMAD: I do not understand this comment, nether which are the classes that should be deprecated :(
Could you elaborate it?
Or we can let it for the end if it's not important.

splatch: Sorry, I did not intend to confuse you. There are old parts in code which use `AuthenticationManager`, we used to have a "secured servlets" which didn't get adopted in 2.5, so they are lingering. However they still keep a reference on how to interact with auth manager:
https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.io.http.auth/src/main/java/org/openhab/core/io/http/auth/internal/AuthenticationHandler.java#L80L99

See line 86 in above example, it gives you back `Authentication` which is OH specific representation of authenticated user/token etc. Going with this path you are free of jaxrs or http specifics - you provide credentials to authentication manager and authentication manager will seek for `AuthenticationProvider` which will be capable of using it. Since there might be multiple providers for same credentials manager will fail only if all of them fail to authenticate user.

Its a bit more complex than filter, but it was designed way earlier for variety of scenarios.

My point is - if there is no further use of `AuthenticationManager` and providers, and instead we go with `AuthFilter` we shall deprecate earlier and surrounding infrastructure (Authentication, AuthenticationProvider). If intention is to keep above then we shall abstain from exposing internal implementation details of jaxrs to other places.

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/CommonWebSocketServlet.java
splatch: Its a design question - you can think of `trustedNetwork` as another level of credentials (caller IP address), which has nothing to do with username/password. Basic auth is matter of extracting the credentials from headers, thus it is not an authentication provisioning mechanism, it is credential source. For implicit user in role - I am not sure what's purpose of that thing is, hence I cant comment on it.

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/CommonWebSocketServlet.java
splatch: You construct a security context in the same way as `AuthFilter` would do thus its better to hide this detail there.

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/CommonWebSocketServlet.java
GiviMAD: The "implicitUserRole" option makes the token not required and grant any request without token with user credentials (gives an AnonymousUserContext).

The "trustedNetworks" option allows to disable the "implicitUserRole" but to preserve this behavior in for some networks filtering the request ip by CIDR.

Yes both three options are not authentication provisioning mechanisms but affect to whetter this runs.

Another thing to highlight, the JWT token is only handled in this bundle so not supported by the org.openhab.core.auth.AuthenticationManager.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/AnonymousUserSecurityContext.java
GiviMAD: Thank you for the explanation, now it all makes sense.

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/CommonWebSocketServlet.java
GiviMAD: Sure, It will look better, I can refactor the code to handle it all in the AuthFilter class.

The initial reason to have it split is that the AccessToken for the websocket can be passed on the query.

For context, the reason for this is that the WebSocket interface exposed in browsers does not allow you to set other headers than "Sec-WebSocket-Protocol". So sending the token in the authentication header is not possible from a browser.

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/CommonWebSocketServlet.java
splatch: JWT token can be handled through `AuthenticationProvider` too, see:

JWTCredentials:
https://github.com/opensmarthouse/opensmarthouse-core/commit/947e116347df8f813322319d8224ebb698f167fe#diff-cdb1dc31859fa2b30a4312f8a9c5faba7b2594a311cffae0a4fcb7c3dabbfa09

JwtAuthenticationProvider (an external JWT)
https://github.com/opensmarthouse/opensmarthouse-core/commit/947e116347df8f813322319d8224ebb698f167fe#diff-30f808a93ce7a81e41fa6d0335d763c3d8f319577854db8ad5a9eb3523eaf6a4

LocalJwtAuthenticationProvider:
https://github.com/opensmarthouse/opensmarthouse-core/commit/947e116347df8f813322319d8224ebb698f167fe#diff-8100f63b9cbf2eb2971e7f074f9054c6c2691b0a114920ab0218d92a562e64e0

Anyhow, these are samples I made in a fork which do not apply here. Also your PR is not about revisiting security. Lets focus on not making too many assumptions on top of `AuthFilter` so eventual re-organization of this layer after 4.0 release will not result in a complete rewrite of logic you supply.

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/CommonWebSocketServlet.java
splatch: Right, in such case you could try to get it around with cookie, but not sure if it will not complicate things even more. :/

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/CommonWebSocketServlet.java
GiviMAD: Sure, thank you for the samples anyway.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/AuthFilter.java
GiviMAD: No, I think that maybe avoid emitting too much log in case of an attack? That makes sense for me.

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/CommonWebSocketServlet.java
GiviMAD: Yes, normally try to avoid using cookies.

I have added some changes to the AuthFilter so the logic is hidden there in the 'getSecuryContext' method.

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/CommonWebSocketServlet.java
GiviMAD: With the changes done in the AuthFilter it should be easier to refactor the code, just need the 'getSecuryContext' functionality to be provided from the core. But I think that is better to let that for a future refactor as you said.

GiviMAD:
Tests failure seems unrelated to the PR.

features/karaf/openhab-core/src/main/feature/feature.xml
J-N-K: I think this is a good point and the `AuthenticationManager` is the way to go.

GiviMAD: This can not be done right now, because the JWT verification and the APISecurity options are both handled in the openhab.core.io.rest.auth package instead than in the core.

@splatch mentioned in another comment that is better to let that change for a future reorganization, which seems better for me, but let me know if this comment still applies.

splatch: Agree, there is more than just `AuthenticationManager` to do, thus as long as we avoid leakage of `AuthFilter` logic beyond it it should be fine to do further updates after 4.0.

bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/AuthFilter.java
J-N-K: I tend to agree with @GiviMAD. Do you see any additional information from the stack trace that is lost otherwise?

J-N-K (changes requested):

J-N-K:
I think that's the idea of the `AuthenticationManager`. But we should probably move that to another PR.

J-N-K (changes requested):
Otherwise LGTM

J-N-K (approved):
Thanks!

jlaur:
@GiviMAD - do you think this could be related?
https://community.openhab.org/t/oh4-sign-in-fails-repeteadly-on-snapshot/147207

GiviMAD:
> @GiviMAD - do you think this could be related?
> https://community.openhab.org/t/oh4-sign-in-fails-repeteadly-on-snapshot/147207

Seems like @J-N-K just found the problem, thank you for handled it and sorry.

PR #3625, opened 2023-05-23
Move RuleHLI service from voice to system

J-N-K:
This is needed to clean-up UI. The categories should be used for add-ons while core services should use `system`. This is non-breaking, configurations ill be preserved.


kaikreuzer (approved):
lgtm, thanks.

PR #3626, opened 2023-05-23
Move LSP service from misc to system

J-N-K:
This is needed to clean-up UI. The categories should be used for add-ons while core services should use `system`.


kaikreuzer (approved):
lgtm, thanks.

PR #3627, opened 2023-05-24
Fix exception in TransformationHelper

J-N-K:
Fixes https://github.com/openhab/openhab-webui/issues/1867

Runtime exceptions thrown by transformation services propagated from the `TransformationHelper` to the `EnrichtedItemDTOMapper` to the `ItemResource`. In general transformation services should throw `TransformationException` and this should be fixed there but in this case safe-guarding the `TransformationHelper` makes sense because of the severity of issues created by fault add-ons.


kaikreuzer (approved):
Thanks!

PR #3629, opened 2023-05-25
[ColorUtil] optimise constants; hue overflow check

andrewfg:
Two improvements to ColorUtils..
- define static constants for code optimisation
- hue value overflow check and correction

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
@@ -38,7 +38,17 @@ import org.slf4j.LoggerFactory;
 public class ColorUtil {
     private static final Logger LOGGER = LoggerFactory.getLogger(ColorUtil.class);
     private static final MathContext COLOR_MATH_CONTEXT = new MathContext(5, RoundingMode.HALF_UP);
-    protected static final BigDecimal BIG_DECIMAL_HUNDRED = BigDecimal.valueOf(100);
+
+    protected static final BigDecimal BIG_DECIMAL_360 = BigDecimal.valueOf(360);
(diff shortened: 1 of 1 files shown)


J-N-K (changes requested):
Otherwise LGTM

J-N-K (approved):

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -254,12 +263,14 @@ public static HSBType rgbToHsb(PercentType[] rgb) throws IllegalArgumentExceptio
...
-            hue = BigDecimal.valueOf(120).add(redAngle).subtract(blueAngle);
+            hue = BIG_DECIMAL_120.add(redAngle).subtract(blueAngle);
         } else {
-            hue = BigDecimal.valueOf(240).add(greenAngle).subtract(redAngle);
+            hue = BIG_DECIMAL_240.add(greenAngle).subtract(redAngle);
         }
         if (hue.compareTo(BigDecimal.ZERO) < 0) {
-            hue = hue.add(BigDecimal.valueOf(360));
+            hue = hue.add(BIG_DECIMAL_360);
+        } else if (hue.compareTo(BIG_DECIMAL_360) > 0) {
```
andrewfg, line 272: @J-N-K just re-reading this line, I think it should be >= rather than > to cover the exact edge case where hue is 360. Or??

andrewfg (commented):
After thoughts

andrewfg (commented):
After thoughts

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
J-N-K: No, I think it's fine. If that is the case 360-360=0 which is the same color.

PR #3630, opened 2023-05-25
Fix NPE in UpgradeTool when no pattern present

J-N-K:
Reported on the forum. State descriptions do no necessarily contain a pattern and thus `.get("pattern")` can return null.


kaikreuzer (approved):
Thanks!

PR #3634, opened 2023-05-27
Remove package support from FeatureInstaller

J-N-K:
Packages are no longer needed since we have the setup wizard in place.

See https://github.com/openhab/openhab-distro/pull/1543


kaikreuzer (approved):
lgtm, thanks!

PR #3635, opened 2023-05-27
Improve performance for state update handling 

### LINKED ISSUE #3620: (Many) Issues with Sitemap Event Subscription
Sitemap Event Subscription is incredible unreliable. There are many scenarios where events are not sent or subscriptions stall. This occurs and can be reproduced within the browser (basic UI) and the android app (sitemaps).

## Expected Behavior
All changes to items should be reflected in real time for the sitemap which is currently opened.

## Current Behavior
There are many issues where this is not the case:
1. Subscription is stalled completely after creation (e.g. android app refresh swipe down, or browser reload). I managed to reproduce it reliably with this script, though I'm not 100% sure that this is what is happening when it occurs on the android app. 
``` javascript
const res = await got.post(`${openhabUrl}/rest/sitemaps/events/subscribe`).json();
const location = res.context.headers.Location;
// choosing any time >= 2000 here will reproduce this issue reliably. 
await new Promise(resolve => setTimeout(resolve, 2000));
const stream = await got.stream(`${location}?sitemap=main&pageid=04`);
for await (const data of stream) {
    console.log(data.toString());
}
```
This will create the following log output (grepping for this subscription id):
```
cat openhab.log | grep -a '2c365d7d-9581-4f64-87dd-54941213e343'
2023-05-19 16:39:27.197 [DEBUG] [t.sitemap.SitemapSubscriptionService] - Created new subscription with id 2c365d7d-9581-4f64-87dd-54941213e343 (7 active subscriptions for a max of 50)
2023-05-19 16:39:27.200 [DEBUG] [est.sitemap.internal.SitemapResource] - Client from IP 192.168.10.134 requested new subscription => got id 2c365d7d-9581-4f64-87dd-54941213e343.
2023-05-19 16:39:29.228 [DEBUG] [est.sitemap.internal.SitemapResource] - Subscription id 2c365d7d-9581-4f64-87dd-54941213e343 does not exist.
```
Also very important to notice is that the connection is kept open! It is not closed as I think was desired [here.](https://github.com/openhab/openhab-core/blob/769aa562a10f9e007ff08723b31b60da240dc8c4/bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/internal/SitemapResource.java#L339) This tricks the client to think the connection will send events but it never will. You can also easily reproduce it with curl and manually copy the subscription id from request 1 to request 2, as this will always take longer than 2 seconds. 

2. Often if the subscription is successfully subscribed by the client, updates are sent not fast enough. I have a sitemap which updates items multiple times a second, but often I can discover this on the logs:
```
2023-05-19 15:45:50.761 [DEBUG] [t.sitemap.SitemapSubscriptionService] - Created new subscription with id c009170c-0537-45d6-b1f8-b00639e622b8 (6 active subscriptions for a max of 50)
2023-05-19 15:45:50.763 [DEBUG] [est.sitemap.internal.SitemapResource] - Client from IP 192.168.10.55 requested new subscription => got id c009170c-0537-45d6-b1f8-b00639e622b8.
2023-05-19 15:45:50.773 [DEBUG] [t.sitemap.SitemapSubscriptionService] - Subscription c009170c-0537-45d6-b1f8-b00639e622b8 changed to page electricity of sitemap electricity (6 active subscriptions}
2023-05-19 15:45:50.774 [DEBUG] [est.sitemap.internal.SitemapResource] - Client from IP 192.168.10.55 requested sitemap event stream for subscription c009170c-0537-45d6-b1f8-b00639e622b8.
2023-05-19 15:46:07.694 [DEBUG] [est.sitemap.internal.SitemapResource] - Sent sitemap event for widget 0100 to subscription c009170c-0537-45d6-b1f8-b00639e622b8.
2023-05-19 15:46:19.553 [DEBUG] [est.sitemap.internal.SitemapResource] - Sent sitemap event for widget 0100 to subscription c009170c-0537-45d6-b1f8-b00639e622b8.
2023-05-19 15:46:28.000 [DEBUG] [est.sitemap.internal.SitemapResource] - Sent sitemap event for widget 0001 to subscription c009170c-0537-45d6-b1f8-b00639e622b8.
2023-05-19 15:46:45.879 [DEBUG] [est.sitemap.internal.SitemapResource] - Sent sitemap event for widget 0001 to subscription c009170c-0537-45d6-b1f8-b00639e622b8.
2023-05-19 15:47:16.174 [DEBUG] [est.sitemap.internal.SitemapResource] - Sent sitemap event for widget 0100 to subscription c009170c-0537-45d6-b1f8-b00639e622b8.
2023-05-19 15:47:40.237 [DEBUG] [est.sitemap.internal.SitemapResource] - Sent sitemap event for widget 0100 to subscription c009170c-0537-45d6-b1f8-b00639e622b8.
```
Here somehow only a few events are sent within minutes, while in reality multiple events should be sent per second. In this state, whenever I refresh the page/app manually it will update to the latest correct values which clearly indicates the current values were outdated. 

I thought that openhab might be overloaded as indeed I have a setup with 10-30 item updates per second, but I don't think it is as CPU load is <= 50% and also everything loads incredible fast when refreshing manually and also all other services (like persistence) don't encounter any problems.

3. Items with `autoupdate="false"` are never (event-based) updated on the UI when changed on the UI. To reproduce, create such an item, include it to a sitemap as `Setpoint`, write a script/rule which listens for commands of this item and update the state of the item. Now the logs look like this when changing it with the setpoint element:
```
2023-05-19 16:32:18.332 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'My_Item' received command 7
2023-05-19 16:32:18.418 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'My_Item' changed from 8 to 7
```
but the sitemap event stream never contains this change even for a working subscription (where point 1 and 2 did not occur). 

## Possible Solution
For me it looks like the sitemap event subscription could need a rework, maybe replacing it entirely with websockets as openhab supports this by now already. 

provided in the description

## Context
I'm using sitemaps to show the state of my home. Sitemaps are awesome and in my opinion far superior to the OH3 UI (e.g. they are blazing fast, no cumbersome login setups required on android, native app navigation works, video streams work through openhab cloud, way easier to setup, etc.). I think it's a pity they don't see a lot of love since OH3 UI. 

## Your Environment
* Version used: openhab 3.4.4 release build
* Environment name and version android 13.0 with 3.6.1 app version, chrome 113
* Operating System and version: raspbian setup

J-N-K:
Fixes #3620 

There are two improvements in this PR:

- The `EventHandler` now uses a "one-thread-per-subscriber` policy. This prevents single "slow" consumers from blocking others but still preserves event order.
- The `PageChangeListener` and the `SitemapResource` have been refactored to use events instead implementing `StateChangeListener`.

### DIFF
--- a/bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/SitemapSubscriptionService.java
+++ b/bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/SitemapSubscriptionService.java
@@ -42,6 +42,8 @@ import org.openhab.core.model.sitemap.sitemap.Sitemap;
 import org.openhab.core.model.sitemap.sitemap.Widget;
 import org.openhab.core.thing.events.ChannelDescriptionChangedEvent;
 import org.openhab.core.ui.items.ItemUIRegistry;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.ServiceRegistration;
 import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Deactivate;
@@ -72,6 +74,7 @@ public class SitemapSubscriptionService implements ModelRepositoryChangeListener
     private static final int DEFAULT_MAX_SUBSCRIPTIONS = 50;
 
     private final Logger logger = LoggerFactory.getLogger(SitemapSubscriptionService.class);
+    private final BundleContext bundleContext;
 
     public interface SitemapSubscriptionCallback {
 
@@ -94,15 +97,17 @@ public class SitemapSubscriptionService implements ModelRepositoryChangeListener
     private final Map<String, Instant> creationInstants = new ConcurrentHashMap<>();
 
     /* sitemap+page -> listener */
-    private final Map<String, PageChangeListener> pageChangeListeners = new ConcurrentHashMap<>();
+    private final Map<String, ListenerRecord> pageChangeListeners = new ConcurrentHashMap<>();
 
     /* Max number of subscriptions at the same time */
     private int maxSubscriptions = DEFAULT_MAX_SUBSCRIPTIONS;
 
     @Activate
-    public SitemapSubscriptionService(Map<String, Object> config, final @Reference ItemUIRegistry itemUIRegistry) {
-        applyConfig(config);
+    public SitemapSubscriptionService(Map<String, Object> config, final @Reference ItemUIRegistry itemUIRegistry,
+            BundleContext bundleContext) {
         this.itemUIRegistry = itemUIRegistry;
+        this.bundleContext = bundleContext;
+        applyConfig(config);
     }
 
     @Deactivate
@@ -110,9 +115,7 @@ public class SitemapSubscriptionService implements ModelRepositoryChangeListener
         pageOfSubscription.clear();
         callbacks.clear();
         creationInstants.clear();
-        for (PageChangeListener listener : pageChangeListeners.values()) {
-            listener.dispose();
-        }
+        pageChangeListeners.values().forEach(l -> l.serviceRegistration.unregister());
         pageChangeListeners.clear();
     }
 
@@ -150,7 +153,7 @@ public class SitemapSubscriptionService implements ModelRepositoryChangeListener
      * Creates a new subscription with the given id.
      *
      * @param callback an instance that should receive the events
-     * @returns a unique id that identifies the subscription or null if the limit of subscriptions is already reached
+     * @return a unique id that identifies the subscription or null if the limit of subscriptions is already reached
      */
     public @Nullable String createSubscription(SitemapSubscriptionCallback callback) {
         if (maxSubscriptions >= 0 && callbacks.size() >= maxSubscriptions) {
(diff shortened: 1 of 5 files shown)


TheTrueRandom (approved):
Nice improvements, tested on my raspberry pi with > 50 item updates/s which are handled with ease now. Before it was struggling with 30 updates/s.

lolodomo:
In my TODO list for the weekend to review this PR.

lolodomo:
This code is probably correct even if I don't understand why it is better to use an event subscriber rather than implementing a `StateChangeListener`.
As this change updates the core behaviour of any sitemap UI, a very detailed review from @kaikreuzer is really expected.

@J-N-K : I assume you checked that live update in BasicUI and Android app is still working as expected after your changes ?

TheTrueRandom:
For me everything on BasicUI and Android app still works as before and I have quite some elements in it.

lolodomo:
@kaikreuzer and all @openhab/core-maintainers : as it is said that performance are greatly improved (for users considering OH as a real-time system), I believe this PR should be considered. As it impacts a sensible part of sitemap UIs, I would suggest to not wait for the last OH4 minute to merge it but rather include it in coming milestone to have few weeks feedback form users.
I already reviewed the code and it looks OK to me. But I am honest, I do not understand why certain changes lead to improved performance. To be reviewed by someone who knows this part of code if possible.

kaikreuzer (approved):
Thanks. My assumption is that the multi-threaded eventhandler brings the performance (well it actually shows that the event subscribers are not implemented well, since they are expected to return immediately...).
I do not see why moving to events for the page listeners is relevant here, but I am fine with that change as well. It is the riskier part wrt potential regressions, but I cannot think of a situation where this might be an issue. The only cases that would now differ in behavior are the ones where an item state changes without an event being created; this might only be the case at the system startup, but at that moment we do not care too much about the page update notifications either.

TL;DR: Lgtm, let's merge and have 2 more milestones for testing before GA.

lolodomo:
I believe the title of the PR could be enhanced to mention that it concerns sitemap UIs.

PR #3642, opened 2023-06-01
Improve existing and add new persistence filters

J-N-K:
#2871 introduced working persistence filters base on time and threshold, but since they were never documented, I doubt they have been used at all.

This PR changes is API breaking because it changes the filter definitions in `persistence.xtext` but this will probably not be noticed at all. It also adds two new persistence filters: "Equals" and "NotEquals" that can be used to allow/disallow a certain set of values.

Filters can be defined by

```
Filters {
  tenseconds : T 10 s
  fivepercent : > % 5
  hundredmilliamps : > 100 mA
  equals : = "only","these","values"
  notequals : ! "not","these","values"
```

and be applied with

```
Items {
  item1, item2 : strategy = everyChange filter = fivepercent
  group1 : strategy = everyUpdate filter = tenseconds
```
I'll add documentation in the next few days.


rkoshak:
While you are at it, does it make sense to add a range filter? Ignore all values that are above/below a min/max range for filtering out obviously bogus sensor readings. That's a common use case.

One could argue that would best be applied in a profile I'll concede.

kaikreuzer:
@J-N-K What do you think about @rkoshak suggestion above?

J-N-K:
It's easy from code, the question is as @rkoshak already pointed out: is this something that should be done for persistence or is it more something for a profile? If you think it's useful as persistence filter, then I can add it.

kaikreuzer:
I personally would consider it useful, because I often see the situation that bogus sensor values get persisted and there is no easy way to clean up persisted data afterwards. So if it is easy to implement and nobody vetos - why not add it.

kaikreuzer:
> That's not an easy change I suspect but is it something worth opening another issue for?

Feel free to open one. 😄

kaikreuzer (approved):
lgtm, thanks!

Issue #3624, opened 2023-05-23
[REST API] Add widget state format to sitemap response


maniac103:
Discussion split from https://github.com/openhab/openhab-core/pull/3481 :

Consider the sitemap and item configuration from [here](https://github.com/openhab/openhab-core/pull/3481#issuecomment-1550869514). The sitemap REST API will yield the following response for the first slider widget:
```
                    {
                        "widgetId": "0200",
                        "type": "Slider",
                        "visibility": true,
                        "label": "Test number dim 1 [60,000000 mm]",
                        "icon": "number",
                        "mappings": [

                        ],
                        "switchSupport": false,
                        "sendFrequency": 0,
                        "minValue": 0,
                        "maxValue": 100,
                        "step": 5,
                        "state": "60",
                        "item": {
                            "link": "http://127.0.0.1:8080/rest/items/TestNumberDim1",
                            "state": "6 cm",
                            "stateDescription": {
                                "pattern": "%.1f cm",
                                "readOnly": false,
                                "options": [

                                ]
                            },
                            "type": "Number:Length",
                            "name": "TestNumberDim1",
                            "label": "Test number dim 1",
                            "tags": [

                            ],
                            "groupNames": [

                            ]
                        },
                        "widgets": [

                        ]
                    },
```

... Which raises a number of questions:

- How does the client know `maxValue=100` means mm, not cm? Since the widget format specifier specifies mm, it's probably what the user intends, but it's not visible to the client.
- Likewise, how does the client know state=60 means mm, not cm?
- How does the client get to know the widget format string (`%f mm`)?

Therefore I'd like to propose the widget state format pattern (as given in the label in the sitemap definition) to the sitemap REST API widget DTO.

lolodomo:
> * Likewise, how does the client know state=60 means mm, not cm?

It is I believe something very specific to the slider widget. The unit is apparently stripped in that case. 
In other widgets, you will get a unit in that field.

lolodomo:
> Discussion split from #3481 :
> 
> Consider the sitemap and item configuration from [here](https://github.com/openhab/openhab-core/pull/3481#issuecomment-1550869514). The sitemap REST API will yield the following response for the first slider widget:
> 
> ```
>                     {
>                         "widgetId": "0200",
>                         "type": "Slider",
>                         "visibility": true,
>                         "label": "Test number dim 1 [60,000000 mm]",
>                         "icon": "number",
>                         "mappings": [
> 
>                         ],
[...]

PR #3644, opened 2023-06-05
[rest] Add widget state pattern and default unit to ItemUIRegistry

mherwege:
Resolves https://github.com/openhab/openhab-core/issues/3624

When developing the Input widget type for BasicUI and the Android app, I ran into a number of constraints to properly support an input widget in all cases. Having better access to the defined pattern and default unit would make this a lot simpler for BasicUI (using the methods of ItemUIRegistry) and make things possible that are not possible at this time in the Android app.

The fundamental problem is that format patterns are meant to be used for visualisation, but not as an input mask. Still, also in the input widget, one wants to use the formatted value (otherwise, you end up with the wrong unit, or digits beyond what is needed). Also, if a value is undefined, it would still be required to show the unit from the state pattern if available (or the default unit if not).

This PR proposes to extend ItemUIRegistry methods by providing a `getFormatPattern` method. This method will get the format pattern from the widget label, resp. item label, resp. item state description.

The sitemap REST response has following additional fields:
- pattern: widget format pattern (using the value from the `getFormatPattern` listed above
- unit: unit of the widget, also available when the value is NULL or UNDEF

In addition, the item REST part of the sitemap response contains an extra field, `unitSymbol` with the default unit or defined unit from meta data, so this can be shown for NULL or UNDEF value if no format or no state description is provided.

All existing tests for ItemUIRegistry run through fine. A modified version of BasicUI is also a lot simpler with these changes added and has been tested to work. I will amend https://github.com/openhab/openhab-webui/pull/1787 and push a version making use of the functionality in this PR if this PR would be accepted.

While this will simplify things in the UI's, this is not an absolute must for BasicUI. For the Android app, a user would need to make sure to provide a simplified format in the label to be able to use the input field consistently. This PR would allow to work around this. It does provide all info in the REST API to avoid ambiguity.

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/item/EnrichedItemDTOMapper.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/item/EnrichedItemDTOMapper.java
@@ -27,6 +27,7 @@ import org.openhab.core.items.GroupItem;
 import org.openhab.core.items.Item;
 import org.openhab.core.items.dto.ItemDTO;
 import org.openhab.core.items.dto.ItemDTOMapper;
+import org.openhab.core.library.items.NumberItem;
 import org.openhab.core.transform.TransformationException;
 import org.openhab.core.transform.TransformationHelper;
 import org.openhab.core.types.StateDescription;
@@ -93,6 +94,10 @@ public class EnrichedItemDTOMapper {
 
         EnrichedItemDTO enrichedItemDTO;
 
+        String unitSymbol = null;
+        if (item instanceof NumberItem numberItem) {
+            unitSymbol = numberItem.getUnitSymbol();
+        }
         if (item instanceof GroupItem groupItem) {
             EnrichedItemDTO[] memberDTOs;
             if (drillDown) {
@@ -111,10 +116,10 @@ public class EnrichedItemDTOMapper {
                 memberDTOs = new EnrichedItemDTO[0];
             }
             enrichedItemDTO = new EnrichedGroupItemDTO(itemDTO, memberDTOs, link, state, transformedState,
-                    stateDescription);
+                    stateDescription, unitSymbol);
         } else {
             enrichedItemDTO = new EnrichedItemDTO(itemDTO, link, state, transformedState, stateDescription,
-                    item.getCommandDescription(locale));
+                    item.getCommandDescription(locale), unitSymbol);
         }
 
         return enrichedItemDTO;
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/item/EnrichedItemDTOMapper.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/item/EnrichedItemDTOMapper.java
@@ -27,7 +27,6 @@ import org.openhab.core.items.GroupItem;
 import org.openhab.core.items.Item;
 import org.openhab.core.items.dto.ItemDTO;
 import org.openhab.core.items.dto.ItemDTOMapper;
-import org.openhab.core.library.items.NumberItem;
 import org.openhab.core.transform.TransformationException;
 import org.openhab.core.transform.TransformationHelper;
 import org.openhab.core.types.StateDescription;
@@ -94,10 +93,6 @@ public class EnrichedItemDTOMapper {
 
         EnrichedItemDTO enrichedItemDTO;
 
-        String unitSymbol = null;
-        if (item instanceof NumberItem numberItem) {
-            unitSymbol = numberItem.getUnitSymbol();
-        }
         if (item instanceof GroupItem groupItem) {
             EnrichedItemDTO[] memberDTOs;
             if (drillDown) {
@@ -116,10 +111,10 @@ public class EnrichedItemDTOMapper {
                 memberDTOs = new EnrichedItemDTO[0];
             }
             enrichedItemDTO = new EnrichedGroupItemDTO(itemDTO, memberDTOs, link, state, transformedState,
-                    stateDescription, unitSymbol);
+                    stateDescription);
         } else {
             enrichedItemDTO = new EnrichedItemDTO(itemDTO, link, state, transformedState, stateDescription,
-                    item.getCommandDescription(locale), unitSymbol);
+                    item.getCommandDescription(locale));
         }
 
         return enrichedItemDTO;
--- a/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java
+++ b/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java
@@ -500,7 +500,7 @@ public class ItemUIRegistryImpl implements ItemUIRegistry {
             if (item instanceof NumberItem numberItem && numberItem.getDimension() != null) {
                 Matcher m = FORMAT_PATTERN.matcher(pattern);
                 int matcherEnd = 0;
-                while (m.find() && m.group(1) == null) {
+                if (m.find() && m.group(1) == null) {
                     matcherEnd = m.end();
                 }
                 String unit = pattern.substring(matcherEnd).trim();
--- a/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java
+++ b/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java
@@ -371,7 +371,7 @@ public class ItemUIRegistryImpl implements ItemUIRegistry {
                 }
             }
         } catch (ItemNotFoundException e) {
-            logger.error("Cannot retrieve item '{}' for widget {}", itemName, w.eClass().getInstanceTypeName());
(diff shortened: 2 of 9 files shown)


lolodomo:
All my OH free time is currently focussed on implementing the mew semantic tag registry but I hope I can review that PR during the weekend.

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java
```diff
@@ -455,22 +455,76 @@ private Switch createPlayerButtons() {
                 }
 
                 label = label.trim();
-                label = label.substring(0, label.indexOf("[") + 1) + formatPattern + "]";
+                int index = label.indexOf("[");
+                if (index >= 0) {
+                    label = label.substring(0, label.indexOf("[") + 1) + formatPattern + "]";
```
lolodomo, line 460: Please use index

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java
```diff
@@ -455,22 +455,76 @@ private Switch createPlayerButtons() {
...
+                if (!postfix.isBlank()) {
+                    logger.warn(
+                            "Item '{}' with unit, nothing allowed after unit in label pattern '{}', dropping postfix",
+                            itemName, pattern);
+                }
+                pattern = pattern.substring(0, matcherEnd) + (!unit.isBlank() ? " " + unit : "");
+            }
+
+        } catch (ItemNotFoundException e) {
+            logger.error("Cannot retrieve item '{}' for widget {}", itemName, w.eClass().getInstanceTypeName());
```
lolodomo, line 522: Looks like a too much high log level. WARN level should be sufficient.

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/item/EnrichedItemDTOMapper.java
```diff
@@ -93,6 +94,10 @@ private static EnrichedItemDTO map(Item item, ItemDTO itemDTO, boolean drillDown
 
         EnrichedItemDTO enrichedItemDTO;
 
+        String unitSymbol = null;
+        if (item instanceof NumberItem numberItem) {
+            unitSymbol = numberItem.getUnitSymbol();
```
lolodomo, line 99: This is the item internal unit ?

lolodomo (changes requested):
Note that you are updating a very sensible (and complex) part.
I hope this will break nothing.

Note also that your change in `convertState` will break the slider widget in Basic UI as Basic UI does not expect a unit in that case. A fix in Basic UI should be prepared.

lolodomo:
@J-N-K or @kaikreuzer  : as this PR is updating a very sensible part of sitemap UIs, please do an attentive review. I have the feeling it is OK. Please also check the pattern strings, I am not very familiar with their syntax.

As already mentioned, the change at line 694 will break slider in Basic UI for numbers with dimension. So a fix must be prepared for Basic UI and merges should be consecutive.

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/item/EnrichedItemDTOMapper.java
J-N-K: yes

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java
mherwege: This wasn't consistent across the class and I just copied one with error level. I will make it warn everywhere.

mherwege:
> As already mentioned, the change at line 694 will break slider in Basic UI for numbers with dimension. So a fix must be prepared for Basic UI and merges should be consecutive.

I will look into this. The change in line 694 is based on @J-N-K suggestion here: https://github.com/openhab/openhab-core/issues/3624#issuecomment-1561817225. Keen to know if there are other consequences. This case is not covered by unit tests.

lolodomo:
In file smarthome.js (BasicUI), line 1698 has to be fixed and comment at line 1683 has to be removed. Line 1659 has to be checked too (I am not sure of the behaviour of parse). 
Setpoint widget is properly handled, so code can be looked at.

lolodomo:
@mherwege : can you prepare a small PR for Basic UI?

mherwege:
@lolodomo I create [the PR ](https://github.com/openhab/openhab-webui/pull/1922), but have one outstanding question on there.

spacemanspiff2007:
Just for my understanding:
When I asked to include the metadata in the item events it was always rejected because it might be not available when the event is issued (or some implementation using a custom metadata provider). 
How is this reflected with in this PR? Will the item unit then be wrongly not set?

I think it would be nice to have the item unit as a dedicated field in the normal item response, too ...

lolodomo:
> @lolodomo I create [the PR ](https://github.com/openhab/openhab-webui/pull/1922), but have one outstanding question on there.

Your BasicUI PR is fine and ready to be merged. 
So we are now waiting for core maintainers review of the current PR. 

I believe there is a good chance of fast merge so you can start adjusting your PR for nee input widget (BasicUI).

lolodomo:
> I believe there is a good chance of fast merge so you can start adjusting your PR for new input widget (BasicUI).

Maybe create a separate PR in case this PR is not merged fast. 
In any case, my intention is to merge your Basic UI stuff related to input widget whatever the solution, before OH4 is released.

lolodomo:
To avoid a new infinite debate about unit management in OH, I would suggest to cut this PR in two, one for REST item (to be debated) , and one for REST sitemap. Idea is to not delay the REST sitemap part which is waited.

mherwege:
> Maybe create a separate PR in case this PR is not merged fast.
> In any case, my intention is to merge your Basic UI stuff related to input widget whatever the solution, before OH4 is released.

PR is here https://github.com/openhab/openhab-webui/pull/1923.

mherwege:
> To avoid a new infinite debate about unit management in OH, I would suggest to cut this PR in two, one for REST item (to be debated) , and one for REST sitemap. Idea is to not delay the REST sitemap part which is waited.

I removed the changes to the item REST part and will do that in a separate PR. However, I am not sure this is solving the potential problem. The unit is retrieved in ItemUIRegistryImpl, and the changes there are necessary for supporting BasicUI. If it is available there, it will be in the REST api, or am I missing something? If it is not available, it will also not work in BasicUI.

lolodomo:
> However, I am not sure this is solving the potential problem. The unit is retrieved in ItemUIRegistryImpl, and the changes there are necessary for supporting BasicUI. If it is available there, it will be in the REST api, or am I missing something? If it is not available, it will also not work in BasicUI.

Basic UI uses code in `ItemUIRegistryImpl` + REST sitemap API but does not rely on REST item API.

mherwege:
> Basic UI uses code in `ItemUIRegistryImpl` + REST sitemap API but does not rely on REST item API.

It uses exactly the same code to get the unitSymbol in `ItemUIRegistryImpl` and the proposed change to the REST item API:

```
        if (item instanceof NumberItem numberItem) {
            unitSymbol = numberItem.getUnitSymbol();
        }
```

Do we have a stronger guarantee that `getUnitSymbol` returns the unit from the unit metadata when called from `ItemUIRegistryImpl`, and not the default unit if the metadata is not available yet? I don't know. If that's not the case, I don't see a difference. And it suggests that overriding the default unit should actually not be stored in the metadata, but should be a field in NumberItem to avoid this kind of situation. But that's far beyond the scope of this PR, and is independent of it.

lolodomo:
Basic UI is now ready for this change (PR is merged).

lolodomo:
@openhab/core-maintainers : please tell us if there is an intention or not to review & merge this PR before OH4 is released. I need this answer to review & merge the right Basic UI PR from @mherwege (input field).

J-N-K (commented):
Otherwise LGTM, even though iterating over the regex seems to be expensive.

@lolodomo Did you try if that works?

lolodomo:
> @lolodomo Did you try if that works?

No, it did not try it myself.
But I hope @mherwege did try it.

mherwege:
> But I hope @mherwege did try it.

Yes, I did try.

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java
```diff
@@ -455,22 +455,76 @@ private Switch createPlayerButtons() {
...
+                    unit = unit.substring(0, unitEnd);
+                }
+                if (!postfix.isBlank()) {
+                    logger.warn(
+                            "Item '{}' with unit, nothing allowed after unit in label pattern '{}', dropping postfix",
+                            itemName, pattern);
+                }
+                pattern = pattern.substring(0, matcherEnd) + (!unit.isBlank() ? " " + unit : "");
+            }
+
```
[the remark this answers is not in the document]
mherwege: Done.

mherwege:
@J-N-K I removed the regex iteration. As this check is only done for number items, there should only be one % format string, so I can just take the first one and declare invalid what comes after.

lolodomo:
@J-N-K : any chance to merge this PR before new milestone is generated ?

J-N-K (approved):
LGTM, thanks.

PR #3649, opened 2023-06-08
[upgradetool] Adjust metadata unit for system channel types

J-N-K:
This enhances the metadata creation to system-channel-types that provide a state description with a unit. In addition warnings are printed if the channel MIGHT set a state description but the tool was not able to determine that.


kaikreuzer (approved):
Great, thank you!

PR #3651, opened 2023-06-11
Fix AuthFilter service declaration

J-N-K:
Regression from #3622 

The `AuthFilter` no longer declared `ContainerRequestFilter` as service.


kaikreuzer (approved):
lgtm, thanks!

Issue #3650, opened 2023-06-09
Sitemap event subscription for whole sitemap instead of one page only


TAKeanice:
Motivation:

I am trying to implement a display based on sitemaps (improving upon the [m5panel project](https://github.com/TAKeanice/m5panel)).
As my display interprets the sitemap slightly different than the BasicUI, and in general, different displays might interpret sitemaps in different ways, the subscription mechanism does not fit: An implementation for TV-sized screens may want to display all of the pages of a sitemap at once, while a small mobile display like the m5panel does not want to issue a new request on each of the frequently occuring navigations, but rather show the page navigated to with the up-to-date widgets straight away.

Current situation:

It took me quite a while to figure out how the subscriptions work for the basic ui: The backend is involved in changing the page of the subscription on navigation to another page. This implies, that at all times only one page can be visible and kept up to date.
Subscriptions are currently limited to a single page per event stream, and an alternative display implementation would have to open one subscription for each page it wants to keep up to date.

Proposed change:

Allow subscriptions for all pages of a sitemap at once (i.e., send updates for all widgets in the whole sitemap instead of only the widgets on a single page)

Implementation option:

* The Rest API (specifically the SitemapResource) accepts requests that only specify the sitemap. Currently, it issues a BAD_REQUEST error response when no pageId parameter is supplied.
* The SitemapSubscriptionService changes its pageOfSubscription map to hold a set of pages for each subscription id. All places where the pageOfSubscription map is used adapt to that change. Transitively, the getPageId method changes to return a set of pageIds for the subscription, and the usages of that method (two in SitemapResource) are adapted accordingly.
* The method `public void setPageId(String subscriptionId, String sitemapName, String pageId)` reacts to a `null` value for pageId by subscribing to _all_ pages of the sitemap

https://github.com/openhab/openhab-core/pull/3635 already refactors some of the SitemapSubscriptionService and the SitemapResource. Maybe it´s possible for @J-N-K to increase the scope of the PR while he´s at it? Otherwise, I am open to contributing a change myself. Please let me know :)

TheTrueRandom:
@TAKeanice Maybe for your usecase it could even make sense to subscribe to the [websocket](https://www.openhab.org/docs/configuration/websocket.html) stream and filter the events based on topics you are interested in. From my experience it is really fast and easy to work with.

TAKeanice:
[...]

PR #3652, opened 2023-06-12
Improvement: Allow subscriptions for complete sitemaps (not limited to a single page)

TAKeanice:
The changes in this PR allow to leave out the pageId in a request for updates from a subscription. The motivation for this was discussed in https://github.com/openhab/openhab-core/issues/3650

Leaving out the pageId in a subscription request makes the event API send events for updates of _all_ widgets within a sitemap. This is accomplished by walking the sitemap tree and collecting all widgets in it when the subscription is created. Previously, only the widgets on the given page were collected at that occasion.

In the issue above, performance issues were discussed. I do not know how to test for them, and would happily accept help with that matter.


lolodomo:
I will have a look to your proposal. But as this is something we are not recommending at all, first can you confirm that this will not become the default behaviour for existing UIs ? And I would like something added in the javadoc explaining that this is a possible not recommended usage.
Any official openHAB should not use that. This could lead to a massive network traffic between server and browser (SSE events).

TAKeanice:
@lolodomo thank you for giving this PR some support. I split up the methods to create a sitemap-wide subscription and a normal subscription endpoint, and commented the ones that should not be used. There should be no possibility to unconsciously subscribe to the whole sitemap.

The things I did not change apparently contained some nullability issues with the pageId. I did not try to resolve them to not break legacy behaviour.

TAKeanice:
@lolodomo the compliance with your request to make sure not the whole sitemap is listened to by default is now also ensured by a regression test. I don´t see any reason why this should not be merged.

TAKeanice:
@lolodomo I know everyone is busy and stuff, but now that it´s spring and everything is getting back to life, would you consider reviving your interest in this PR, too?

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/SitemapSubscriptionService.java
```diff
@@ -281,19 +317,41 @@ private EList<Widget> collectWidgets(String sitemapName, String pageId) {
...
     }
 
-    private String getValue(String sitemapName, String pageId) {
+    private String getScopeIdentifier(String sitemapName) {
+        return sitemapName;
+    }
+
+    private String getScopeIdentifier(String sitemapName, String pageId) {
         return sitemapName + SITEMAP_PAGE_SEPARATOR + pageId;
     }
```
J-N-K, line 356:
```suggestion
    private String getScopeIdentifier(String sitemapName, @Nullable String pageId) {
        return pageId == null ? sitemapName : sitemapName + SITEMAP_PAGE_SEPARATOR + pageId;
    }
```

I think you can also collapse both versions of `addCallbackToListener` in one method, calling either `collectAllWidgetsInSitemap` or `collectWidgetsOnPage` (depending on `pageId == null`).

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/internal/SitemapResource.java
```diff
@@ -265,10 +267,50 @@ public Response getSitemapData(@Context HttpHeaders headers,
...
+     *
+     * Subscribe to a whole sitemap (all pages at once).
+     * This is not a recommended option as it incurs high SSE traffic
+     * and might pose a high load on the server, depending on the sitemap size.
+     * No built-in openhab UI should use this functionality.
+     *
+     */
     @GET
-    @Path("/{sitemapname: [a-zA-Z_0-9]+}/{pageid: [a-zA-Z_0-9]+}")
+    @Path("/{sitemapname: [a-zA-Z_0-9]+}/wholeSitemap")
```
J-N-K, line 279: I would prefer something like `*`, because `wholeSitemap` is also a valid page id (at least it matchs the regexp).

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/SitemapSubscriptionService.java
```diff
@@ -215,18 +217,39 @@ public boolean exists(String subscriptionId) {
...
 
+    private boolean isPageListener(String sitemapWithPageId) {
+        return sitemapWithPageId.contains(SITEMAP_PAGE_SEPARATOR);
+    }
+
     private String extractPageId(String sitemapWithPageId) {
         return sitemapWithPageId.split(SITEMAP_PAGE_SEPARATOR)[1];
     }
 
+    public void unsetPageId(String subscriptionId, String sitemapName) {
```
J-N-K, line 236: I don't think you need this method if you allow `pageId` to be `null` and act accordingly. This also simplifies code in the `SitemapResource`

J-N-K (changes requested):
I think you can simplify the code very much by allowig `pageId` to be null (and consider that as "whole sitemap"). You can remove nearly all added methods in this case.

TAKeanice:
@J-N-K I the request by @lolodomo was specifically not to do that, since it could accident is be used in some official implementation. My original implementation was like you suggested. If you want, I revert my changes to get that version back. Expecting your final decision :)

J-N-K:
I can't find that request in this PR, can you point me to the comment?

TAKeanice:
> the comment

https://github.com/openhab/openhab-core/pull/3652#issuecomment-1773755492_

Have I misinterpreted the request to make sure it does not get used accidentally?

J-N-K:
IMO yes, my understanding is that a comment should state "whole sitemap is not recommended" (maybe also in the API documentation, not only in JavaDoc) and that a subscription with page should never result in a subscription to the whole sitemap.

TAKeanice:
I will make some simplifications to the private methods, that should reduce code duplication.

lolodomo:
> IMO yes, my understanding is that a comment should state "whole sitemap is not recommended

Yes, that was my request.
Also as a consequence that it should not become the default behaviour.

TAKeanice:
The reason I split the methods is that the method `getSitemapEvents` https://github.com/openhab/openhab-core/blob/81f2bd9366d3f937da892ac704671207fce5e10c/bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/internal/SitemapResource.java#L346-L349 allows null as pageId. If I unsplit the whole implementation by allowing null pageIds throughout, omitting the pageId when calling that method would result in accidentally using sitemap-wide subscriptions. That was the nullability issue I referred to here: https://github.com/openhab/openhab-core/pull/3652#issuecomment-1807185491

J-N-K:
It does not allow `null` as `pageId`:

```
 if (sitemapname != null && pageId != null) {
            subscriptions.setPageId(subscriptionId, sitemapname, pageId);
        }
```

My proposal was to use `null` in Java code and `*` as `pageId` for "whole sitemap". `*` is not a valid `pageId`, so there can be no collisions. You can adjust the regexp so that it accepts either `*` or a valid `pageId`.

TAKeanice:
I improved the code as suggested

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/SitemapSubscriptionService.java
```diff
@@ -284,17 +307,17 @@ private EList<Widget> collectWidgets(String sitemapName, String pageId) {
...
                 oldListener.serviceRegistration().unregister();
                 pageChangeListeners.remove(sitemapPage);
             }
         }
     }
 
-    private String getValue(String sitemapName, String pageId) {
-        return sitemapName + SITEMAP_PAGE_SEPARATOR + pageId;
+    private String getScopeIdentifier(String sitemapName, @Nullable String pageId) {
+        return sitemapName + (pageId == null ? "" : SITEMAP_PAGE_SEPARATOR + pageId);
```
J-N-K, line 320:
```suggestion
        return pageId == null ? sitemapName : sitemapName + SITEMAP_PAGE_SEPARATOR + pageId;
```

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/internal/SitemapResource.java
```diff
@@ -265,10 +267,50 @@ public Response getSitemapData(@Context HttpHeaders headers,
...
+     *
+     */
+    @GET
+    @Path("/{sitemapname: [a-zA-Z_0-9]+}/*")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Operation(operationId = "pollDataForSitemap", summary = "Polls the data for a whole sitemap. Not recommended due to potentially high traffic.", responses = {
+            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = SitemapDTO.class))),
+            @ApiResponse(responseCode = "404", description = "Sitemap with requested name does not exist"),
+            @ApiResponse(responseCode = "400", description = "Invalid subscription id has been provided.") })
+    public Response getPageData(@Context HttpHeaders headers,
```
J-N-K, line 285:
```suggestion
    public Response getSitemapData(@Context HttpHeaders headers,
```

J-N-K (changes requested):
Otherwise LGTM, @lolodomo can you have a look or test this before merge?

J-N-K:
There seems to be an issue with spotless

TAKeanice:
> There seems to be an issue with spotless

yes, it does not like formatting in conditions :D Fixed that.

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/internal/SitemapResource.java
```diff
@@ -333,6 +375,28 @@ public Object createEventSubscription() {
...
+     * and might pose a high load on the server, depending on the sitemap size.
+     * No built-in openhab UI should use this functionality.
+     *
+     */
+    @GET
+    @Path(SEGMENT_EVENTS + "/{subscriptionid: [a-zA-Z_0-9-]+}/*")
+    @Produces(MediaType.SERVER_SENT_EVENTS)
+    @Operation(operationId = "getSitemapEvents", summary = "Get sitemap events for a whole sitemap. Not recommended due to potentially high traffic.", responses = {
+            @ApiResponse(responseCode = "200", description = "OK"),
+            @ApiResponse(responseCode = "400", description = "Page not linked to the subscription."),
```
lolodomo, line 391: Please update the description.
I have even the felling this response can not occur; if that is true, rather remove this line.

lolodomo (changes requested):

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/SitemapSubscriptionService.java
```diff
@@ -215,67 +217,88 @@ public boolean exists(String subscriptionId) {
...
+        return sitemapWithPageId.contains(SITEMAP_PAGE_SEPARATOR);
+    }
+
     private String extractPageId(String sitemapWithPageId) {
         return sitemapWithPageId.split(SITEMAP_PAGE_SEPARATOR)[1];
     }
 
     /**
-     * Updates the subscription to send events for the provided page id.
+     * Updates the subscription to send events for the provided page id (or whole sitemap if page is is null).
```
lolodomo, line 237: Typo: page **id** is null

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/SitemapSubscriptionService.java
```diff
@@ -72,6 +73,7 @@ public class SitemapSubscriptionService implements ModelRepositoryChangeListener
     private static final String SITEMAP_PAGE_SEPARATOR = "#";
     private static final String SITEMAP_SUFFIX = ".sitemap";
     private static final int DEFAULT_MAX_SUBSCRIPTIONS = 50;
+    public static final Duration WAIT_AFTER_CREATE_SECONDS = Duration.ofSeconds(30);
```
lolodomo, line 76: Minor but rather private

lolodomo (changes requested):

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/internal/SitemapResource.java
```diff
@@ -265,10 +267,50 @@ public Response getSitemapData(@Context HttpHeaders headers,
...
+                return JSONResponse.createErrorResponse(Response.Status.BAD_REQUEST, e.getMessage());
+            }
+        }
+
+        boolean timeout = false;
+        if (headers.getRequestHeader("X-Atmosphere-Transport") != null) {
+            timeout = blockUnlessChangeOccursAnywhere(sitemapname);
+        }
+        SitemapDTO responseObject = getSitemapBean(sitemapname, uriInfo.getBaseUriBuilder().build(), locale,
+                includeHiddenWidgets, timeout);
```
lolodomo, line 306: Why do you need timeout information in the response ?
It was not needed in the current implementation.

TAKeanice: Right, some of these changes might have crept in because I started development quite some time ago when the code still was like that. Same with the nullability problems. I will fix it this evening.

TAKeanice: I remember why the timeout field is necessary. It was meant analogous to the timeout in PageDTO, but I can just use the field timeout in the homepage of the sitemap.

TAKeanice: Timeout parameter removed from SitemapDTO, but still needed in getSitemapBean to set on homepage PageDTO. The parameter is necessary to tell the caller whether a change has occurred or just a timeout when long-polling for events

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/SitemapSubscriptionService.java
TAKeanice: Is now private

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/SitemapSubscriptionService.java
TAKeanice: wrote pageId to use the same spelling as the parameter

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/internal/SitemapResource.java
TAKeanice: The method `getSitemapEvents` returns a `BAD_REQUEST` when there is either no sitemap or no pageId (when trying to subscribe to a page only) or when either of them could not be set as location for the subscription

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/internal/SitemapResource.java
```diff
@@ -341,26 +405,31 @@ public Object createEventSubscription() {
     @Produces(MediaType.SERVER_SENT_EVENTS)
     @Operation(operationId = "getSitemapEvents", summary = "Get sitemap events.", responses = {
             @ApiResponse(responseCode = "200", description = "OK"),
-            @ApiResponse(responseCode = "400", description = "Page not linked to the subscription."),
+            @ApiResponse(responseCode = "400", description = "Missing sitemapname or page parameter, or page not linked successfully to the subscription."),
```
J-N-K, line 408:
```suggestion
            @ApiResponse(responseCode = "400", description = "Missing sitemap or page parameter, or page not linked successfully to the subscription."),
```

J-N-K (changes requested):
Otherwise LGTM, maybe @lolodomo can also approve here.

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/internal/SitemapResource.java
```diff
@@ -341,26 +405,31 @@ public Object createEventSubscription() {
...
         }
-        if (sitemapname != null && pageId != null) {
-            subscriptions.setPageId(subscriptionId, sitemapname, pageId);
+        if (sitemapname != null && (subscribeToWholeSitemap || pageId != null)) {
+            subscriptions.updateSubscriptionLocation(subscriptionId, sitemapname, pageId);
         }
-        if (subscriptions.getSitemapName(subscriptionId) == null || subscriptions.getPageId(subscriptionId) == null) {
-            logger.debug("Subscription id {} is not yet linked to a sitemap/page.", subscriptionId);
+        if (subscriptions.getSitemapName(subscriptionId) == null
+                || subscriptions.getPageId(subscriptionId) == null && !subscribeToWholeSitemap) {
```
lolodomo, line 431: I believe some parenthesis are missing around the second line. Please confirm.

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/SitemapSubscriptionService.java
```diff
@@ -215,67 +217,88 @@ public boolean exists(String subscriptionId) {
...
+        return sitemapWithPageId.contains(SITEMAP_PAGE_SEPARATOR);
+    }
+
     private String extractPageId(String sitemapWithPageId) {
         return sitemapWithPageId.split(SITEMAP_PAGE_SEPARATOR)[1];
     }
 
     /**
-     * Updates the subscription to send events for the provided page id.
+     * Updates the subscription to send events for the provided page id (or whole sitemap if pageId is is null).
```
lolodomo, line 237: Very minor but you still kept the doubled is => is is null

lolodomo (changes requested):

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/internal/SitemapResource.java
J-N-K: good catch

TAKeanice: No, because && has higher operator precedence.

TAKeanice: The condition says "we need to have successfully subscribed to a sitemap and, if we do not want to subscribe to a whole sitemap, we need to have successfully subscribed to a page"

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/internal/SitemapResource.java
```diff
@@ -714,7 +803,7 @@ private boolean blockUnlessChangeOccurs(String sitemapname, String pageId) {
      *            the widgets of the page to observe
      * @return true if the timeout is reached
      */
-    private boolean waitForChanges(EList<Widget> widgets) {
+    private boolean waitForChanges(List<Widget> widgets) {
         long startTime = (new Date()).getTime();
         boolean timeout = false;
         Set<String> items = getAllItems(widgets).stream().map(Item::getName).collect(Collectors.toSet());
```
lolodomo, line 809: This will be interesting that you add a DEBUG log just after this line showing how many widgets you have and how many items it resulted.

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/internal/SitemapResource.java
lolodomo: Even if the code is good due to precedence of such operator over another, I will prefer you add parenthesis as precedence between && and || is not something very intuitive.

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/internal/SitemapResource.java
lolodomo: In a common user situation (user defining only one sitemap), it should lead to have more or less all items.

TAKeanice: Good idea, I added that but in the method in SitemapSubscriptionService and deduplicated the code from this method.

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/internal/SitemapResource.java
TAKeanice: Matter of taste, I always prefer less boilerplate but will follow your preference here as you´re the maintainers :) To explain my preference, I think in complicated boolean statements, too many parenthesis in a row confuse new developers, as they have to count the parenthesis to know what they enclose. I think, although not coming from a Python background, that smart indentations help understanding much better. For that statement for example I would prefer the following, but I believe spotless prohibits that:

```Java
if (subscriptions.getSitemapName(subscriptionId) == null
     || subscriptions.getPageId(subscriptionId) == null 
        && !subscribeToWholeSitemap)
```

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/internal/SitemapResource.java
TAKeanice: Sorry, I got that wrong, added the log statement into waitForChanges as well.

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/SitemapSubscriptionService.java
```diff
@@ -215,86 +217,111 @@ public boolean exists(String subscriptionId) {
...
+            return widgets;
+        }
+
+        if (pageId != null && !pageId.equals(sitemap.getName())) {
+            // subscribing to subpage of sitemap --> get all widgets from that page
+            Widget pageWidget = itemUIRegistry.getWidget(sitemap, pageId);
+            if (pageWidget instanceof LinkableWidget widget) {
+                widgets = itemUIRegistry.getChildren(widget);
+                // We add the page widget. It will help any UI to update the page title.
+                widgets.add(pageWidget);
```
lolodomo, line 289: In the current implementation, `pageWidget`was not considered by `blockUnlessChangeOccurs`.

I am not sure if it was an error or it was not an error what could be the impact of considering it now.
WDYT ?

J-N-K:
There is an issue with the tests, can you fix that? I'll nmerge afterwards. Thanks for your patience.

lolodomo:
I have not checked but I will not be surprised that tests are failing due to the change I am commenting in my last message.

TAKeanice:
@J-N-K and @lolodomo I think the tests fail due to my code deduplication, I'll fix that today. It was a pleasure working on finalizing this the last couple of days, thank you for that!

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/SitemapSubscriptionService.java
TAKeanice: I think both implementations for collecting widgets are meant to do the same and omitting the page in one of them was the typical consequence of code duplication. I will check what that causes with the tests though.

TAKeanice:
The test fails because the subscriptionService is just a mock, I need to properly instantiate it and mock its itemUIRegistry in the test.

J-N-K:
You can also keep the mocked `subscriptionService` and return another mock, depending on the called method (using `when`).

TAKeanice:
> You can also keep the mocked `subscriptionService`

but that would defeat the purpose of the test, since an important part of the logic resides in collectWidgets

J-N-K:
I didn't look at the test, it was more a general comment.

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/SitemapSubscriptionService.java
lolodomo: I just hope it was not done deliberately.

TAKeanice: I will take a look at the history of the code before resolving this conversation

lolodomo: Is `blockUntilChangeOccurs` only used when long polling mode is used ? Sorry, I do not remember.

TAKeanice: Yes

lolodomo: In this case, this will not impact Basic UI when used with any modern browser (meaning supporting SSE).
But not sure about Android app and iOS app, even if I know that the Android app is also using a SSE connection.

By the way, as previously said, the change probably makes sense.

J-N-K (approved):
Thanks!

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/SitemapSubscriptionService.java
TAKeanice: Seems like both places have been there since the code was added to the repository in 2019 (a37cceab), so I can´t judge. But the effect will only be a little more frequent updates, that occur also when the parent page itself changes, so I agree that it will probably not cause any trouble.

TAKeanice:
@J-N-K the test issues have been resolved.

lolodomo:
By renaming methods, you broke UIs using them :(
Basic UI is no more compiling, I am going to fix that.

lolodomo:
I search in GitHub and apparently Basic UI was the unique app calling this method.

PR #3654, opened 2023-06-13
[rest] Add unit to item response

mherwege:
This extends the enrichedItemDTO with a unitSymbol field, representing the default unit for a number item. This would be the value configured in unit metadata or the default for the dimension.

This field is required to better support sitemap input widgets in apps. When no unit is provided in the label pattern or state description, it allows the input widget to show the default unit symbol. Combined with https://github.com/openhab/openhab-core/pull/3644 this would allow to make the functionality for the input widget equivalent to the proposed functionality for BasicUI.


J-N-K (approved):
LGTM

PR #3664, opened 2023-06-18
Fix missing query prefix in JarFileAddonService

J-N-K:
See title. This brings the query id in-line with the other add-on services. Before queries without add-on service id were not properly processed. As a result the add-on configuration page could not be accessed.

Also adds correct content type.


kaikreuzer (approved):
Thank you!

PR #3668, opened 2023-06-25
Replace `StringBuffer` usages with `StringBuilder`

wborn:
See: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/StringBuilder.html

> This class provides an API compatible with StringBuffer, but with no guarantee of synchronization.
> This class is designed for use as a drop-in replacement for StringBuffer in places where the string buffer was being used by a single thread (as is generally the case).
> Where possible, it is recommended that this class be used in preference to StringBuffer as it will be faster under most implementations.


J-N-K (approved):
LGTM, thanks!

PR #3674, opened 2023-06-28
Always use ThingUid as the unique identifier for thing status metrics

### LINKED ISSUE #3672: [metrics] Duplicate values in Influx for same thing
I have installed the metrics service via .cfg file (FYI using only GUI does not seem to do anything, only once I created metrics.cfg on the backend did it start inserting values) However while setting up a Grafana Dashboard to monitor when a thing goes OFFLINE, I found loads of examples where the last value in the UID was defined by itself, but shows OFFLINE.

It would also be handy if there was a config to use Thing Label instead of UUID (substituting _ for spaces as I assume influx would not enjoy that..)

## Expected Behavior
time                     metric_type openhab_core_metric thing                                               value
2023-02-24T23:56:40.993Z gauge       true                sonos:Amp:RINCON_347E5C0D150501400                  3

## Current Behavior
time                     metric_type openhab_core_metric thing                                               value
2023-02-24T23:56:40.993Z gauge       true                sonos:Amp:RINCON_347E5C0D150501400                  3
2023-02-24T23:56:40.992Z gauge       true                RINCON_347E5C0D150501400                            4

## Your Environment
* Version used: 3.4.0
* Influx 1.8.2

scotthraban:
At bind time, ThingId was being used, but then ThingUid was being used when changes occured, causing multiple meters to be created for the same Thing, with the status always being 0 for the meter created at bind time.

This change uses the full ThingUid, for both binding and event processing, since ThingUid is the globally unique value.

Fixes #3672

In case it is useful, the original PR to add this code can be found here: https://github.com/openhab/openhab-core/pull/2133

### DIFF
--- a/bundles/org.openhab.core.io.monitor/src/main/java/org/openhab/core/io/monitor/internal/metrics/ThingStateMetric.java
+++ b/bundles/org.openhab.core.io.monitor/src/main/java/org/openhab/core/io/monitor/internal/metrics/ThingStateMetric.java
@@ -40,16 +40,18 @@ import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.Tags;
 
 /**
- * The {@link ThingStateMetric} class implements a metric for the openHAB things states.
+ * The {@link ThingStateMetric} class implements a metric for the openHAB things
+ * states.
  *
  * @author Robert Bach - Initial contribution
+ * @author Scott Hraban - Create Meter using thingUid instead of thingId during
+ *         bind phase
  */
 @NonNullByDefault
 public class ThingStateMetric implements OpenhabCoreMeterBinder, EventSubscriber {
     private final Logger logger = LoggerFactory.getLogger(ThingStateMetric.class);
     public static final String METRIC_NAME = "openhab.thing.state";
     private static final String THING_TAG_NAME = "thing";
-    private static final String THINGSTATUS_TOPIC_PREFIX = "openhab/things/";
     private final ThingRegistry thingRegistry;
     private final Meter.Id commonMeterId;
     private final Map<Meter.Id, AtomicInteger> registeredMeters = new HashMap<>();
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.io.monitor/src/main/java/org/openhab/core/io/monitor/internal/metrics/ThingStateMetric.java
```diff
@@ -70,7 +72,7 @@ public void bindTo(@NonNullByDefault({}) MeterRegistry meterRegistry) {
         logger.debug("ThingStateMetric is being bound...");
         this.meterRegistry = meterRegistry;
         thingRegistry.getAll().forEach(
-                thing -> createOrUpdateMetricForBundleState(thing.getUID().getId(), thing.getStatus().ordinal()));
+                thing -> createOrUpdateMetricForBundleState(thing.getUID().getAsString(), thing.getStatus().ordinal()));
```
scotthraban, line 75: The core of the change, using thingUid instead of thingId.

bundles/org.openhab.core.io.monitor/src/main/java/org/openhab/core/io/monitor/internal/metrics/ThingStateMetric.java
```diff
@@ -111,10 +113,13 @@ public Set<String> getSubscribedEventTypes() {
...
     @Override
     public void receive(Event event) {
-        logger.trace("Received ThingStatusInfo(Changed)Event...");
-        String thingId = event.getTopic().substring(THINGSTATUS_TOPIC_PREFIX.length(),
-                event.getTopic().lastIndexOf('/'));
-        ThingStatus status = gson.fromJson(event.getPayload(), ThingStatusInfo.class).getStatus();
-        createOrUpdateMetricForBundleState(thingId, status.ordinal());
+        if (event instanceof ThingStatusInfoEvent) {
+            logger.trace("Received ThingStatusInfo(Changed)Event...");
+            String thingUid = ((ThingStatusInfoEvent) event).getThingUID().getAsString();
```
scotthraban, line 118: A more direct way of getting at the thingUid, rather than parsing it as a string out of the event topic. Since we are only subscribing for events of this type, we are safe to cast to the event type, after checking that something horrible wrong did not happen of course.

J-N-K, line 118:
```suggestion
        if (event instanceof ThingStatusInfoEvent thingEvent) {
            logger.trace("Received ThingStatusInfo(Changed)Event...");
            String thingUid = thingEvent.getThingUID().getAsString();
```

J-N-K (changes requested):
Thanks! Just one small comment

scotthraban:
Thanks, that is awesome, learn something new everyday!

J-N-K (approved):

PR #3675, opened 2023-06-30
[audio] Improve audio duration computation

dalgwen:
Add another way of computing sound duration (using information from the AudioFormat class)

Allows the use of a "Sizeable" interface (for AudioStream that we know the length of, but we cannot clone). We can then improve the success of the duration detection, for example for the pulseaudio sink (PR coming after). We can also give the length information to sink in more cases, even if we cannot clone the stream.

Add the support of the mark / reset methods to some common AudioStream. We then allow more stream analysis for sink requiring it (Stream analysis often requires to get back in time after consuming a few bytes)

### DIFF
--- a/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/FileAudioStream.java
+++ b/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/FileAudioStream.java
@@ -42,9 +42,11 @@ public class FileAudioStream extends FixedLengthAudioStream implements Disposabl
 
     private final File file;
     private final AudioFormat audioFormat;
-    private InputStream inputStream;
+    private FileInputStream inputStream;
     private final long length;
     private final boolean isTemporaryFile;
+    private int markedOffset = 0;
+    private int alreadyRead = 0;
 
     public FileAudioStream(File file) throws AudioException {
         this(file, getAudioFormat(file));
@@ -87,7 +89,7 @@ public class FileAudioStream extends FixedLengthAudioStream implements Disposabl
         }
     }
 
-    private static InputStream getInputStream(File file) throws AudioException {
+    private static FileInputStream getInputStream(File file) throws AudioException {
         try {
             return new FileInputStream(file);
         } catch (FileNotFoundException e) {
@@ -102,7 +104,9 @@ public class FileAudioStream extends FixedLengthAudioStream implements Disposabl
 
     @Override
     public int read() throws IOException {
-        return inputStream.read();
+        int read = inputStream.read();
+        alreadyRead++;
+        return read;
     }
 
     @Override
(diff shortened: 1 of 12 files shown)


bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/FileAudioStream.java
```diff
@@ -102,7 +104,9 @@ public AudioFormat getFormat() {
 
     @Override
     public int read() throws IOException {
-        return inputStream.read();
+        int read = inputStream.read();
+        alreadyRead++;
```
lolodomo, line 108: Isn't it rather `alreadyRead += read;` ?

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/utils/AudioSinkUtilsImpl.java
```diff
@@ -66,11 +66,24 @@ public class AudioSinkUtilsImpl implements AudioSinkUtils {
                             .longValue();
                     return startTime + computedDuration;
                 } catch (IOException | UnsupportedAudioFileException e) {
-                    logger.debug("Cannot compute the duration of input stream", e);
+                    logger.debug("Cannot compute the duration of input stream with method java stream sound analysis",
+                            e);
+                    Integer bitRate = audioFormat.getBitRate();
+                    Integer channel = audioFormat.getChannels();
```
lolodomo, line 72: channel is apparently not used, is it normal ?

lolodomo (changes requested):

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/FileAudioStream.java
dalgwen: The "alreadyRead" counter is the amount of bytes read.
The read() method reads only one byte and return the value of this byte.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/utils/AudioSinkUtilsImpl.java
dalgwen: Good catch. I used this on a previous try but it is not needed.
I will remove it.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/FileAudioStream.java
lolodomo: Ok

lolodomo:
Build failed in "Automation Integration Tests". Not sure it have any link with your PR.
@openhab/core-maintainers : can one of you trigger a new build please ?
Except that problem, code LGTM.

kaikreuzer (approved):

PR #3681, opened 2023-07-03
[persistence] Don't require `relative`, `inverted` and `unit` fields for filter configuration over REST

florian-h05:
While creating the persistence edit page, I noticed that PUT `/rest/persistence/{serviceId}` requests fail when the boolean properties `inverted` and `relative` of some persistence filters are not set. First, I ensured that they are set when the UI saves the persistence configuration, however IMO this should be in core.
Setting up a treshold filter with the UI, it did not work because the `unit` field was blank (I got an NPE from `PersistenceTresholdFilter`, and the `PersistenceIncludeFilter` would also throw a NPE.

This PR sets reasonable defaults for those fields that are required by core, but not neccassarily set by the user.

/cc @J-N-K


J-N-K:
I'm not sure this is the best way. IMO it would better to make the constructors accept a null value and set the default in the filter class. That makes it much clearer what the default is and also works in other places. WDYT?

splatch:
Isn't it better to make this particular DTO mapper non-static? It seems to be that `PersistenceFilters` have rather a dynamic nature and mapping of them in such tied way will always lead to troubles.

florian-h05:
> That makes it much clearer what the default is and also works in other places. WDYT?

I will try that.

florian-h05:
I am currently on holiday in France, so I won’t be able to work on this soon. Let’s postpone it for openHAB 4.1, since the feature freeze will start on Sunday.

PR #3683, opened 2023-07-03
Fix wrong ready marker in ScriptEngineFactoryBundleTracker

### LINKED ISSUE #3676: Rules refuse to load at startup due to jars in addons folder
Upgraded from 3485 to 3519.  I have 4 jars in my addons folder, and have for some time now.  Since upgrading, at start the rules never fire.  I let it sit for over 15 minutes at one point.  Thinking I had a bad jar (which was unlikely as they all worked on 3485) I moved them out of the addons folder and immediately saw:

2023-06-30 15:03:00.153 [INFO ] [nal.ScriptEngineFactoryBundleTracker] - All automation bundles ready.

After further testing, moving ANY jar out of the addons folder seems to clear out the issue.  It is irrelevant which jar is moved, it seems that the identification that a change happened is enough to break it free.  I can move the jar right back in and everything works as expected.

J-N-K:
Fixes #3676


kaikreuzer (approved):
lgtm, thanks!

PR #3688, opened 2023-07-05
Fix marketplace add-ons missing config description URI

### LINKED ISSUE #3685: [marketplace] Unable to access add-on configuration when installed from marketplace
Since the removal of the "Other Services" section in the UI, a couple of voice services that I had in the marketplace are imposible to configure through the main ui.

I think is a core problem because after going around it I detected that the field  "configDescriptionURI", is returned black from the "/rest/addons/marketplace:142036?serviceId=marketplace" call. 
When the addon is installed locally that field contains the configuration uri and I can correctly access the configuration page.

This problem is reproducible in the 4.0.0-m4 version using the [ActionTemplateInterpreter Next](https://community.openhab.org/t/actiontemplateinterpreter-next-3-4-0-4-0-0/142036) or [HABSpeaker](https://community.openhab.org/t/habspeaker/140655) add-ons from the marketplace (both appear in the in the voice section now, I had the posts incorrectly tagged as binding before).

EDIT: I think the problem exists in all kind of marketplace addons, not only voice.

J-N-K:
Fixes #3685

### DIFF
--- a/bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/AbstractRemoteAddonService.java
+++ b/bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/AbstractRemoteAddonService.java
@@ -20,6 +20,7 @@ import java.util.Dictionary;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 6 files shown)


kaikreuzer (approved):
lgtm, thanks!

PR #3698, opened 2023-07-10
[voice] Reduce collisions on exact match and use item synonyms

GiviMAD:
Hi openHAB maintainers, I would like to add this changes to the StandardInterpreter so it uses the item synonyms and to reduce collision between items labels in case of single exact match. I think the changes can be seen as a bug fix as the current item resolution method is unchanged an takes priority in case of succeed.

Note that I have added a test suite for the StandardInterpreter with tests cases for the changes included and also some debug logs to the AbstractRuleBasedInterpreter getMatchingItems method so users have a way to know what the collisions are.

Hope everything is in place.

Best regards

### DIFF
--- a/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/text/AbstractRuleBasedInterpreter.java
+++ b/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/text/AbstractRuleBasedInterpreter.java
@@ -12,16 +12,8 @@
  */
 package org.openhab.core.voice.text;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
+import java.util.*;
 import java.util.Map.Entry;
-import java.util.ResourceBundle;
-import java.util.Set;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
@@ -567,8 +559,8 @@ public abstract class AbstractRuleBasedInterpreter implements HumanLanguageInter
      */
     protected List<Item> getMatchingItems(ResourceBundle language, String[] labelFragments,
             @Nullable Class<?> commandType) {
-        List<Item> items = new ArrayList<>();
-        List<Item> exactMatchItems = new ArrayList<>();
+        Set<Item> items = new HashSet<>();
+        Set<Item> exactMatchItems = new HashSet<>();
         Map<Item, List<Set<String>>> map = getItemTokens(language.getLocale());
         for (Entry<Item, List<Set<String>>> entry : map.entrySet()) {
             Item item = entry.getKey();
@@ -601,16 +593,7 @@ public abstract class AbstractRuleBasedInterpreter implements HumanLanguageInter
                             }
                         }
                         if (insert) {
-                            for (int i = 0; i < items.size(); i++) {
-                                String itemName = items.get(i).getName();
-                                if (itemName.startsWith(name)) {
-                                    logger.debug(
-                                            "Discarding partial matched item '{}' because its name starts with '{}'",
-                                            itemName, name);
-                                    items.remove(i);
-                                    i--;
-                                }
-                            }
+                            items.removeIf((matchedItem) -> matchedItem.getName().startsWith(name));
                             items.add(item);
                             if (exactMatch) {
                                 exactMatchItems.add(item);
@@ -627,7 +610,7 @@ public abstract class AbstractRuleBasedInterpreter implements HumanLanguageInter
             logger.debug("Exact matched items against {}{}: {}", labelFragments, typeDetails,
                     exactMatchItems.stream().map(Item::getName).toArray(String[]::new));
         }
-        return items.size() != 1 && exactMatchItems.size() == 1 ? exactMatchItems : items;
+        return new ArrayList<>(items.size() != 1 && exactMatchItems.size() == 1 ? exactMatchItems : items);
     }
 
     /**
--- a/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/text/AbstractRuleBasedInterpreter.java
+++ b/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/text/AbstractRuleBasedInterpreter.java
@@ -621,7 +621,7 @@ public abstract class AbstractRuleBasedInterpreter implements HumanLanguageInter
             }
         }
         if (logger.isDebugEnabled()) {
-            String typeDetails = commandType != null ? " that accept " + commandType : "";
+            String typeDetails = commandType != null ? " that accept " + commandType.getSimpleName() : "";
             logger.debug("Partial matched items against {}{}: {}", labelFragments, typeDetails,
                     items.stream().map(Item::getName).toArray(String[]::new));
             logger.debug("Exact matched items against {}{}: {}", labelFragments, typeDetails,
--- a/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/text/AbstractRuleBasedInterpreter.java
+++ b/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/text/AbstractRuleBasedInterpreter.java
@@ -200,9 +200,10 @@ public abstract class AbstractRuleBasedInterpreter implements HumanLanguageInter
 
     /**
      * Retrieves the list of identifier token sets per item currently contained in the {@link ItemRegistry}.
-     * Each item entry in the resulting hash map will feature a list of different token sets. Each token set
-     * represents one possible way "through" a chain of parent groups, where each groups tokenized name is
-     * part of the set.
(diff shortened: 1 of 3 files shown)


GiviMAD:
I was testing the proposed changes at my home over the last snapshot and realize those didn't work well with groups, I have to introduce more changes to fix my problems.

I think with the latest changes name collisions are better handled. Between that and the use of the synonyms seems like I can fine tune how it works without so much trouble.

For example now I can have multiple groups items whose labels collides like "bedroom lamp" and "livingroom lamp" and by adding a synonym to one them control what is the target when I use the command "turn off lamp".

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/text/AbstractRuleBasedInterpreter.java
```diff
@@ -538,52 +589,73 @@ protected String executeSingle(ResourceBundle language, String[] labelFragments,
...
+        return new ArrayList<>(items.size() != 1 && exactMatchItems.size() == 1 ? exactMatchItems : items);
+    }
+
+    private static void insertDiscardingMembers(Set<Item> items, Item item) {
+        String name = item.getName();
+        boolean insert = true;
+        for (Item si : items) {
+            if (name.startsWith(si.getName())) {
+                insert = false;
+            }
```
lolodomo, line 649: You can add a break here.

lolodomo (commented):
Code looks good but is very difficult to understand due to such undocumented object `Map<Locale, Map<Item, List<List<List<String>>>>> itemTokens` !

GiviMAD:
> Code looks good but is very difficult to understand due to such undocumented object Map<Locale, Map<Item, List<List<List<String>>>>> itemTokens !

True.

Having a group item labeled as "Samsung TV" with synonym "Smart TV" which contains a switch item labeled as "Power" the object will look as:
```
{
	localeRef: {
		groupItemRef: [
			[ ["Samsung", "TV"] ],
			[ ["Smart", "TV"] ],
		],
		switchItemRef: [
			[ ["Power"] ],
			[ ["Samsung", "TV"], ["Power"] ],
			[ ["Smart", "TV"], ["Power"] ],
		]
	}
}
```

The first list level are the alternative paths to the item, the second level is a path to the item conformed by collections of tokens for the different parents and itself.

Not sure what is the best way of document this, maybe I can add that example to the getItemTokens comment (which is outdated after my changes, I didn't realize before).

GiviMAD:
I have updated the method comment with the current behavior.

bundles/org.openhab.core.voice/src/test/java/org/openhab/core/voice/internal/text/StandardInterpreterTest.java
```diff
@@ -0,0 +1,119 @@
...
+import org.openhab.core.library.types.OnOffType;
+import org.openhab.core.voice.text.InterpretationException;
+
+/**
+ * Test the standard interpreter
+ *
+ * @author Miguel Álvarez - Initial contribution
+ */
+@NonNullByDefault
+public class StandardInterpreterTest {
```
J-N-K, line 48:
```suggestion
@NonNullByDefault
@ExtendWith(MockitoExtensions.class)
public class StandardInterpreterTest {
```

bundles/org.openhab.core.voice/src/test/java/org/openhab/core/voice/internal/text/StandardInterpreterTest.java
```diff
@@ -0,0 +1,119 @@
...
+    private @Mock @NonNullByDefault({}) EventPublisher eventPublisherMock;
+
+    private @Mock @NonNullByDefault({}) ItemRegistry itemRegistryMock;
+    private @Mock @NonNullByDefault({}) MetadataRegistry metadataRegistryMock;
+    private @NonNullByDefault({}) StandardInterpreter standardInterpreter;
+    private static final String OK_RESPONSE = "Ok.";
+
+    @BeforeEach
+    public void setUp() {
+        MockitoAnnotations.openMocks(this);
```
J-N-K, line 59:
```suggestion
```

bundles/org.openhab.core.voice/src/test/java/org/openhab/core/voice/internal/text/StandardInterpreterTest.java
```diff
@@ -0,0 +1,119 @@
...
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.voice.internal.text;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.reset;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
```
J-N-K, line 18:
```suggestion
import static org.mockito.Mockito.verify;
import static org.mockito.Mockito.when;
```

bundles/org.openhab.core.voice/src/test/java/org/openhab/core/voice/internal/text/StandardInterpreterTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+    @Test
+    public void noNameCollisionOnSingleExactMatch() throws InterpretationException {
+        var computerItem = new SwitchItem("computer");
+        computerItem.setLabel("Computer");
+        var computerScreenItem = new SwitchItem("screen");
+        computerScreenItem.setLabel("Computer Screen");
+        List<Item> items = List.of(computerItem, computerScreenItem);
+        Mockito.when(itemRegistryMock.getAll()).thenReturn(items);
+        Mockito.when(itemRegistryMock.getItems()).thenReturn(items);
```
J-N-K, line 71:
```suggestion
        when(itemRegistryMock.getAll()).thenReturn(items);
        when(itemRegistryMock.getItems()).thenReturn(items);
```

also below

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/text/AbstractRuleBasedInterpreter.java
```diff
@@ -538,52 +589,73 @@ protected String executeSingle(ResourceBundle language, String[] labelFragments,
...
+                List<String> unmatchedFragments = new ArrayList<>(lowercaseLabelFragments);
+                for (List<String> segment : parts) {
+                    if (segment.equals(lowercaseLabelFragments)) {
+                        exactMatch = true;
+                        unmatchedFragments.clear();
                         break;
                     }
+                    if (!unmatchedFragments.isEmpty()) {
+                        segment.stream().filter(unmatchedFragments::contains).forEach(unmatchedFragments::remove);
+                    }
```
J-N-K, line 619:
```suggestion
                    unmatchedFragments.removeAll(segment);
```

Wouldn't this also work?

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/text/AbstractRuleBasedInterpreter.java
```diff
@@ -538,52 +589,73 @@ protected String executeSingle(ResourceBundle language, String[] labelFragments,
...
+            Item item = entry.getKey();
+            for (List<List<String>> parts : entry.getValue()) {
+                boolean exactMatch = false;
+                logger.trace("Checking tokens {} against the item tokens {}", labelFragments, parts);
+                List<String> lowercaseLabelFragments = Arrays.stream(labelFragments)
+                        .map(lf -> lf.toLowerCase(language.getLocale())).toList();
+                List<String> unmatchedFragments = new ArrayList<>(lowercaseLabelFragments);
+                for (List<String> segment : parts) {
+                    if (segment.equals(lowercaseLabelFragments)) {
+                        exactMatch = true;
```
J-N-K, line 613: Why is that important? And I believe this is not true in general:

- labelFragments `[a, b, c, d]`
- segments: `[[b], [c, d], [a]]`
- first segment, there is `[a, c, d]` left
- second segment, there is `[a]` left
- third segment matches everything that is left, so `exactMatch = true`

Is that intended?

J-N-K (changes requested):

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/text/AbstractRuleBasedInterpreter.java
GiviMAD: Thank you for the review, and sorry for taking so long to answer, it caught me in the middle of another task that took more than expected.

Maybe I don't understanding the question but I think the code ok.
Each segment is compared against lowercaseLabelFragments which is never modified, checking if some segment in the path is an exact match agains the label extracted from the voice command. Each iteration I'm trying to remove from unmatchedFragments but that is a different list initialized with same strings.

I have renamed the segment and parts variables trying to make that part easier to read.

J-N-K (approved):
Thanks.

GiviMAD:
Thank you!

PR #3699, opened 2023-07-12
[voice] Add length limit to TTS handled by cache

dalgwen:
During the TTS cache development, there was some discussion about adding a boolean to enable / disable cache per request. We didn't keep that because it makes the API complicated.
But time passed and new usages arise, and maybe it now can be discussed, on a new form ?

For example, there is now a ChatGPT binding. It is possible to make some  advanced chat/queries with a LLM. Most of the time, the chat response are lengthy. And so one "discussion" with a LLM can clog the LRU cache and removes older entries worth keeping.

Proposal :
We can use the TTS length as a simple way to differentiate TTS sentences that should be cached, automatically. 
TTS we should cache are those that can be repeated, and they are nearly always pretty short. examples :
"Ok, I'll do that"
"Hello, what can I do for you ?",
"You've got a message"
"Please close the garage door".

On the other side, one can safely assume that long TTS are generated for report of some kind (meteo, chatGPT ?), probably not meant to be repeated. They are also the ones taking the most space in the cache.

I put a 150 default character limit (configurable). To be discussed if the whole idea worth something.

### DIFF
--- a/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/cache/TTSLRUCacheImpl.java
+++ b/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/cache/TTSLRUCacheImpl.java
@@ -53,6 +53,7 @@ public class TTSLRUCacheImpl implements TTSCache {
 
     // a small default cache size for all the TTS services (in kB)
     private static final long DEFAULT_CACHE_SIZE_TTS = 10240;
+    private static final int DEFAULT_MAX_TEXT_LENGTH_CACHE_TTS = 150;
 
     static final String CONFIG_CACHE_SIZE_TTS = "cacheSizeTTS";
     static final String CONFIG_ENABLE_CACHE_TTS = "enableCacheTTS";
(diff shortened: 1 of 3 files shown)


lolodomo (approved):
LGTM

lolodomo:
Looks to me like a very good idea.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/cache/TTSLRUCacheImpl.java
```diff
@@ -66,6 +67,11 @@ public class TTSLRUCacheImpl implements TTSCache {
      * current request is not known and may or may not exceed the limit.
      */
     protected long cacheSizeTTS = DEFAULT_CACHE_SIZE_TTS * 1024;
+    /**
+     * Length text limit. If exceeded, the cache will passthrough the request without storing it.
+     * (One can safely assume that long TTS are generated for report, probably not meant to be repeated)
+     */
+    private long maxTextLengthCacheTTS = 150;
```
J-N-K, line 74:
```suggestion
    private long maxTextLengthCacheTTS = DEFAULT_MAX_TEXT_LENGTH_CACHE_TTS;
```

and use that also below

bundles/org.openhab.core.voice/src/main/resources/OH-INF/config/voice.xml
```diff
@@ -59,6 +59,12 @@
 			<description>The limit size of the TTS cache (in kB).</description>
 			<default>10240</default>
 		</parameter>
+		<parameter name="maxTextLengthCacheTTS" type="integer">
+			<label>TTS Cache Maximum Text Length</label>
+			<description>The limit length of a text handled by the TTS cache. If exceeded, will passthrough to the TTS without
+				storing it. 0 for no limit.</description>
```
J-N-K, line 65:
```suggestion
			<description>The maximum length of texts handled by the TTS cache. If exceeded, will pass the text to the TTS without
				storing it. 0 for no limit.</description>
```

J-N-K (changes requested):
In general LGTM. Is the limit in characters? Maybe add that to the description.

dalgwen:
Thank you both for your reviews !

J-N-K (approved):
LGTM, thanks

PR #3702, opened 2023-07-14
Increase large event queue warn limit

### LINKED ISSUE #3690: The queue for a subscriber of type 'class org.openhab.core.thing.internal.CommunicationManager' exceeds 1000 elements
2023-07-06 09:36:55.459 [WARN ] [ab.core.internal.events.EventHandler] - The queue for a subscriber of type 'class org.openhab.core.thing.internal.CommunicationManager' exceeds 1000 elements. System may be unstable.
...
2023-07-06 09:36:56.294 [WARN ] [ab.core.internal.events.EventHandler] - The queue for a subscriber of type 'class org.openhab.core.thing.internal.CommunicationManager' exceeds 1000 elements. System may be unstable.

Upgraded to snapshot 3528 to validate resolution of a different issue.  Now at startup, I get flooded with almost 1000 of the above message inside of a second (first and last shown) and then they calm down.  I've never seen this before, it was not present on snapshot 3519 that I came from.

EDIT: Unsure if it's related, but the CPU on my container is hovering now between 15% and 20% constantly which it hadn't done before.

J-N-K:
Closes #3690 

On larger installations with 1000+ items the limit of 1000 elements in the queue is exceeded easily. Since no events are lost (they just take longer to be processed), the limit that needs to be exceeded before a warning is issued should be increased.

See reports on the forum.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/events/EventHandler.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/events/EventHandler.java
@@ -44,6 +44,7 @@ import org.slf4j.LoggerFactory;
 @NonNullByDefault
 public class EventHandler implements AutoCloseable {
 
+    private static final int EVENT_QUEUE_WARN_LIMIT = 5000;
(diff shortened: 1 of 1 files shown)


kaikreuzer (approved):
Thanks!

PR #3707, opened 2023-07-17
Align system-channel-types for electricity

J-N-K:
These are only used by one binding at the moment and it's very inconsistent to have power, voltage and current with the "electric" prefix and energy with "electrical".


clinique (approved):

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/DefaultSystemChannelTypeProvider.java
```diff
@@ -391,7 +391,7 @@ public class DefaultSystemChannelTypeProvider implements ChannelTypeProvider {
      * Electrical-energy: system wide {@link ChannelType} which shows the electrical energy
      */
     public static final ChannelType SYSTEM_ELECTRICAL_ENERGY = ChannelTypeBuilder
-            .state(SYSTEM_CHANNEL_TYPE_UID_ELECTRICAL_ENERGY, "Electrical Energy", "Number:Energy")
+            .state(SYSTEM_CHANNEL_TYPE_UID_ELECTRIC_ENERGY, "Electrical Energy", "Number:Energy")
```
jlaur, line 394: Should the label also be updated accordingly?
```suggestion
            .state(SYSTEM_CHANNEL_TYPE_UID_ELECTRIC_ENERGY, "Electric Energy", "Number:Energy")
```

and description below:

Issue #3515, opened 2023-04-01
Spotless neither checks nor fixes files in `src/test/java`


J-N-K:
Found in #3510 

The configuration is correctly imported and changed files are properly identified, but then they are reported as "clean" even if they are not. This could be a configuration issue at our end or n external bug.

J-N-K:
It's even more confusing.

```
public class   InputStreamCacheWrapperTest {
```

get fixed to

```
public class   InputStreamCacheWrapperTest {
```

while in the same file 

```
@Test
    public void           cacheWrapperStreamReadTest() throws IOException {
        when(cacheEntry.read(0, 1)).thenReturn(new byte[] { 1 });
```

is accepted and not changed. So it looks like the file gets processed, but no all checks are applied. If I do the same to files under `/src/main/java`, the extra-spaces are correctly removed.

wborn:
It seems to check the files again when using 4.25 here:  https://github.com/openhab/openhab-core/blob/4851bfbcae59861ce5a1bb4a2e8d08a456a60cbc/pom.xml#L81

Doesn't seem to be fixed when using newer versions there.

J-N-K:
I think that 4.26 is needed to fix the other spotless issues with method calls in instanceof.

holgerfriedrich:
4.27 seems to be available, (together with spotless 2.37.0....) I cannot check if it fixes the issue above, as I am not using Eclipse.

Changelog is here: https://github.com/diffplug/spotless/blob/main/plugin-maven/CHANGES.md

PR #3710, opened 2023-07-17
Use Eclipse formatter 4.25 with Spotless

wborn:
This is the result after running:

`mvn spotless:apply -Dspotless.eclipse.version=4.25`

Perhaps we should just revert the version in the property?
I still don't know of any negative side effects and the PR will probably also build successfully while it is uses 4.26.

See also https://github.com/openhab/openhab-core/issues/3515#issuecomment-1493256202.


J-N-K:
The problem with 4.25 is something with `if (Element instanceof Foo bar) {` (I can't remember exactly what was formatted wrong, I believe a space was missing somewhere). This is fixed in 4.26

wborn:
I now also tested the newer Spotless and Eclipse formatter versions but they still don't fix the issue that it does not check/apply test files.

J-N-K (approved):

PR #3714, opened 2023-07-19
Fix CommunicationManager command handling

### LINKED ISSUE #3694: Profile JS Transformation Item ->  Thing for Datatype Datetime not triggered
JS-Transformation in a profile for direction item to thing is not triggered in case input value is of datatype datetime. 

Details  see also reference https://community.openhab.org/t/oh4-m4-problem-with-profile-js-transformation-of-datetime-type/147662

OH4 Milstone 4

J-N-K:
Fixes #3694

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/CommunicationManager.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/CommunicationManager.java
@@ -288,7 +288,7 @@ public class CommunicationManager implements EventSubscriber, RegistryChangeList
         if (item != null && thing != null) {
             Channel channel = thing.getChannel(link.getLinkedUID());
             if (channel != null) {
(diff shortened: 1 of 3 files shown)


kaikreuzer:
Lgtm in general, but I also cannot judge on possible side effects. It might be too risky to merge it one day before the RC.
Shall we better wait and only merge it for 4.1?

J-N-K:
The problem is that I'm not sure if this worked before (in 3.4) and is a regression from the `CommunicationManager` refactoring.

kaikreuzer:
It is a fairly rare use case, so I'd rather live with that regression than to potentially introduce other ones unintentionally now.

J-N-K:
Fine. We can also do a 4.0.1 if it turns out to be a bigger issue.

wborn:
> Shall we better wait and only merge it for 4.1?

Perhaps we can merge it now?

kaikreuzer (approved):
Good idea. :-)

PR #3721, opened 2023-07-21
Provide correct version information for XChart in NOTICE

Flole998:
@kaikreuzer This is basically the PR you requested. We want to make sure we provide accurate information about the XChart version used in our NOTICE file. For some reason it was outdated.


kaikreuzer (approved):
Ok, perfect, thank you!

PR #3724, opened 2023-07-22
[Auth Filter] Fix implicit user role for websocket connections

GiviMAD:
Hi openHAB maintainers.

I just discovered testing the version 4.0.0 RC1 that I left a bug on the check of the trusted networks for the WebSocket connections.

I just by mistake couple of times servletRequest instead of request which do not api the api call validations as in that case both point to the same ref, but causes a null pointer exception when validating a websocket connection.

This issue is reproducible when you have implicitRole disabled but trustedNetworks configured in the "API Security" options and connect over WebSocket to the server (using the functionality exposed by the bundle org.openhab.core.io.websocket).

### DIFF
--- a/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/AuthFilter.java
+++ b/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/AuthFilter.java
@@ -296,7 +296,7 @@ public class AuthFilter implements ContainerRequestFilter {
                         return authenticateBasicAuth(authValue);
                 }
             }
-        } else if (isImplicitUserRole(servletRequest)) {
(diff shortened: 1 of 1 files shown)


GiviMAD:
Pinning @openhab/core-maintainers in case you are not notified of the PR creation, just in case you want to merge it before the release, not really urgent as I don't think there is anything using the WebSocket feature.

J-N-K:
What is the result? An implicit user role is assumed when it shouldn't, or the request is rejected even if it should have been allowed?

GiviMAD:
The connection is aborted by the exception, it only happen when you try to access without accessToken.

kaikreuzer (approved):
This is a quite obvious bug and I see no risk of regressions here - I therefore think it is safe to merge.

GiviMAD:
Thank you!

PR #3725, opened 2023-07-22
Fix system started DSL rules

### LINKED ISSUE #3723: New warning "No pre-parsed script found for xxx"
With snapshot 3556 (equivalent to OH4 RC1), I encounter for the first time ever these warning logs "No pre-parsed script found for rule":
```
11:03:37.535 [INFO ] [del.core.internal.ModelRepositoryImpl] - Loading model 'maison3.rules'
11:04:16.776 [WARN ] [ntime.internal.engine.DSLScriptEngine] - No pre-parsed script found for maison3-28.
11:04:34.208 [INFO ] [org.openhab.core.model.script.Rule   ] - Démarrage début
11:04:34.237 [INFO ] [org.openhab.core.model.script.Rule   ] - Démarrage fin
11:04:34.269 [INFO ] [re.automation.internal.RuleEngineImpl] - Rule engine started.

...

11:05:50.825 [WARN ] [ntime.internal.engine.DSLScriptEngine] - No pre-parsed script found for maison3-28.
```

maison3-28 is my rule with trigger "System started", the one logging "Démarrage début" and "Démarrage fin".

What is surprising is that it occurs a second time during the startup process.

J-N-K:
Fixes #3723 

@lolodomo can you test if that resolves your issues? On my system it seems to work fine.

### DIFF
--- a/bundles/org.openhab.core.model.rule.runtime/src/org/openhab/core/model/rule/runtime/internal/DSLRuleProvider.java
+++ b/bundles/org.openhab.core.model.rule.runtime/src/org/openhab/core/model/rule/runtime/internal/DSLRuleProvider.java
@@ -173,7 +173,7 @@ public class DSLRuleProvider
                 default:
                     logger.debug("Unknown event type.");
             }
-
+            notifyProviderChangeListeners(modelRules);
         } else if ("script".equals(ruleModelType)) {
             List<ModelRulePair> modelRules = new ArrayList<>();
             switch (type) {
--- a/bundles/org.openhab.core.model.rule.runtime/src/org/openhab/core/model/rule/runtime/internal/DSLRuleProvider.java
+++ b/bundles/org.openhab.core.model.rule.runtime/src/org/openhab/core/model/rule/runtime/internal/DSLRuleProvider.java
@@ -296,7 +296,8 @@ public class DSLRuleProvider
             s += "\n\n";
         }
         String firstLine = s.lines().findFirst().orElse("");
-        String indentation = firstLine.substring(0, firstLine.length() - firstLine.stripLeading().length());
+        String indentation = firstLine == null ? ""
+                : firstLine.substring(0, firstLine.length() - firstLine.stripLeading().length());
         return s.lines().map(line -> (line.startsWith(indentation) ? line.substring(indentation.length()) : line))
                 .collect(Collectors.joining("\n"));
     }
@@ -304,7 +305,7 @@ public class DSLRuleProvider
     private @Nullable Trigger mapTrigger(EventTrigger t) {
         if (t instanceof SystemOnStartupTrigger) {
             Configuration cfg = new Configuration();
-            cfg.put("startlevel", 20);
+            cfg.put("startlevel", 40);
             return TriggerBuilder.create().withId(Integer.toString(triggerId++))
                     .withTypeUID("core.SystemStartlevelTrigger").withConfiguration(cfg).build();
         } else if (t instanceof SystemStartlevelTrigger slTrigger) {
@@ -446,10 +447,11 @@ public class DSLRuleProvider
 
     private void notifyProviderChangeListeners(List<ModelRulePair> modelRules) {
         modelRules.forEach(rulePair -> {
-            if (rulePair.oldRule() != null) {
-                rules.remove(rulePair.oldRule().getUID());
+            Rule oldRule = rulePair.oldRule();
+            if (oldRule != null) {
+                rules.remove(oldRule.getUID());
                 rules.put(rulePair.newRule().getUID(), rulePair.newRule());
-                listeners.forEach(listener -> listener.updated(this, rulePair.oldRule(), rulePair.newRule()));
+                listeners.forEach(listener -> listener.updated(this, oldRule, rulePair.newRule()));
             } else {
                 rules.put(rulePair.newRule().getUID(), rulePair.newRule());
                 listeners.forEach(listener -> listener.added(this, rulePair.newRule()));
--- a/bundles/org.openhab.core.model.rule.runtime/src/org/openhab/core/model/rule/runtime/internal/DSLRuleProvider.java
+++ b/bundles/org.openhab.core.model.rule.runtime/src/org/openhab/core/model/rule/runtime/internal/DSLRuleProvider.java
@@ -158,7 +158,8 @@ public class DSLRuleProvider
                         int index = 1;
                         for (org.openhab.core.model.rule.rules.Rule rule : ruleModel.getRules()) {
                             Rule newRule = toRule(ruleModelName, rule, index);
-                            Rule oldRule = rules.get(ruleModelName);
+                            Rule oldRule = rules.remove(ruleModelName);
+                            rules.put(newRule.getUID(), newRule);
                             xExpressions.put(ruleModelName + "-" + index, rule.getScript());
                             modelRules.add(new ModelRulePair(newRule, oldRule));
                             index++;
--- a/bundles/org.openhab.core.model.rule.runtime/src/org/openhab/core/model/rule/runtime/internal/DSLRuleProvider.java
+++ b/bundles/org.openhab.core.model.rule.runtime/src/org/openhab/core/model/rule/runtime/internal/DSLRuleProvider.java
@@ -172,44 +172,30 @@ public class DSLRuleProvider
                 default:
                     logger.debug("Unknown event type.");
             }
-            modelRules.forEach(rulePair -> {
-                if (rulePair.oldRule() != null) {
-                    rules.remove(rulePair.oldRule().getUID());
-                    rules.put(rulePair.newRule().getUID(), rulePair.newRule());
-                    listeners.forEach(listener -> listener.updated(this, rulePair.oldRule(), rulePair.newRule()));
-                } else {
-                    rules.put(rulePair.newRule().getUID(), rulePair.newRule());
-                    listeners.forEach(listener -> listener.added(this, rulePair.newRule()));
-                }
-            });
 
         } else if ("script".equals(ruleModelType)) {
+            List<ModelRulePair> modelRules = new ArrayList<>();
             switch (type) {
                 case MODIFIED:
-                    Rule oldRule = rules.remove(modelFileName);
-                    if (oldRule != null) {
-                        listeners.forEach(listener -> listener.removed(this, oldRule));
-
-                    }
                 case ADDED:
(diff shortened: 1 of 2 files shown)


J-N-K:
I don't have a good to to test this, since I neither use file-based rules nor scripts. It should be tested that

- rules/scripts can be added
- rules/scripts can be modified, only show up once and execute the new code
- rules/scripts can be removed

lolodomo:
After a bundle update and a OH server restart; the warning has disappeared but now the rule is run twice, apparently the second just after the first:
```
20:39:59.891 [INFO ] [org.openhab.core.model.script.Rule   ] - Démarrage début
20:39:59.907 [INFO ] [org.openhab.core.model.script.Rule   ] - Démarrage fin
20:40:00.002 [INFO ] [org.openhab.core.model.script.Rule   ] - Démarrage début
20:40:00.034 [INFO ] [org.openhab.core.model.script.Rule   ] - Démarrage fin
20:40:00.064 [INFO ] [re.automation.internal.RuleEngineImpl] - Rule engine started.
```

J-N-K:
I have no idea why that happens, maybe a check for the start-level is missing. Can you post a file that shows this behavior? In my tests the rues run exactly once. Unfortunately I can't work on that today.

kaikreuzer:
> Can you post a file that shows this behavior?

[This file](https://github.com/openhab/openhab-distro/blob/main/launch/app/runtime/conf/rules/demo.rules) will actually do (adding some logging to it).

I get this result at startup:
```
23:51:13.363 [INFO ] [org.openhab.core.Activator          ] - Starting openHAB 4.0.0 (build Unknown Build No.)
23:51:16.033 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model 'demo.items'
23:51:16.368 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model 'mapdb.persist'
23:51:16.549 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model 'demo.sitemap'
23:51:16.660 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model 'demo.things'
23:51:18.436 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model 'demo.rules'
23:51:23.795 [INFO ] [org.openhab.core.model.script.TEST  ] - Rule executed.
23:51:28.769 [INFO ] [org.openhab.core.model.script.TEST  ] - Rule executed.
23:51:28.772 [INFO ] [e.automation.internal.RuleEngineImpl] - Rule engine started.
23:51:28.773 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'DemoDateTime' changed from 2023-07-22T23:51:23.799362+0200 to 2023-07-22T23:51:28.771569+0200
```
So indeed the rule is executed twice and the first time happens quite a time before the rule engine is officially started.

lolodomo:
I tried again with the last change in the PR but same result, that is rule run twice.

bundles/org.openhab.core.model.rule.runtime/src/org/openhab/core/model/rule/runtime/internal/DSLRuleProvider.java
```diff
@@ -317,15 +296,16 @@ private String removeIndentation(String script) {
...
+                : firstLine.substring(0, firstLine.length() - firstLine.stripLeading().length());
         return s.lines().map(line -> (line.startsWith(indentation) ? line.substring(indentation.length()) : line))
                 .collect(Collectors.joining("\n"));
     }
 
     private @Nullable Trigger mapTrigger(EventTrigger t) {
         if (t instanceof SystemOnStartupTrigger) {
             Configuration cfg = new Configuration();
-            cfg.put("startlevel", 20);
+            cfg.put("startlevel", 40);
```
kaikreuzer, line 308: This change fixes the issue for me. I have no idea, why I have ever set it to 20 in the first place.
According to the documentation at https://www.openhab.org/docs/configuration/rules-dsl.html#system-based-triggers, "startlevels less than 40 are not available as triggers". Setting it to 40 will only execute them once during startup.

kaikreuzer:
@lolodomo I just pushed a commit that fixes the issue for me. Would you be able to test this as well, please?

lolodomo:
> @lolodomo I just pushed a commit that fixes the issue for me. Would you be able to test this as well, please?

I do it now...

kaikreuzer:
Thanks for your nightshift.

lolodomo:
Kai, it looks like the rule is run twice but now the second time occurs very late at startup when everything is loaded.
```
00:46:13.976 [INFO ] [org.openhab.core.model.script.Rule   ] - Démarrage début
00:46:14.005 [INFO ] [org.openhab.core.model.script.Rule   ] - Démarrage fin
00:46:14.036 [INFO ] [re.automation.internal.RuleEngineImpl] - Rule engine started.

...

00:47:34.478 [INFO ] [org.openhab.core.model.script.Rule   ] - Démarrage début
00:47:34.489 [INFO ] [org.openhab.core.model.script.Rule   ] - Démarrage fin
```

kaikreuzer:
I assume your rule has a simple "System started" trigger and nothing else?

lolodomo:
But I believe it was already happening in the past that this rule was run a second time very late at startup.
Maybe we are back to the previous behaviour.

lolodomo:
> I assume your rule has a simple "System started" trigger and nothing else?

Yes

```
rule "Démarrage"
when
	System started
then
	logInfo("Rule","Démarrage début")
	lastDesarmTime = now.minusDays(1)
	lastMinimoteTime = now
	patternVoice = Pattern::compile("(donner|donne|passer en hors gel|passer en mode confort|passer en mode réduit) (le |la |les |l'|du )(.*) (du |de la |de l'|dans le |dans la |dans l')(.*)")
	logInfo("Rule","Démarrage fin")
end
```

lolodomo:
I can try quickly with snapshot 3544 to confirm that the rule was already run a second time very late.

kaikreuzer:
> But I believe it was already happening in the past 

Which is still very weird - I never had startup rules being triggered twice.
Is there maybe some mechanism on your system that touches these rules, so that they are reloaded for some reason?

kaikreuzer:
> I can try quickly with snapshot 3544 to confirm that the rule was already run a second time very late.

Yeah, would be great.

lolodomo:
I can confirm that the behaviour with snapshot 3544 is the same as the current with this PR.
So there is probably an issue because this rule is run a second time very late but it is not new.

If your tests are positive, I believe we can go with that.

kaikreuzer (approved):
Ok, thanks for testing - then this at least is nothing newly introduced by this PR.
I hereby approved the changes done by @J-N-K and leave it to him to merge the PR if he approves my changes as well.

J-N-K (commented):
LGTM

Nadahar:
There is a bug that is at least related to this PR. I'm working on implementing format conversion for rules, and for DSL rules, the rule UID is decided by the file/model name and a 1-based index that increases for each rule in the file.

In `DSLRuleProvider` we have this field:

```java
private final Map<String, Rule> rules = new ConcurrentHashMap<>();
```

It contains a mapping between the "model name" and the `Rule` object. In contrast, this is done like this in `YamlRuleProvider`:

```java
    private final Map<String, Collection<Rule>> rulesMap = new ConcurrentHashMap<>();
```

Since one model can contain more than one rule, there is a "challenge" in having a one-to-one mapping between model name and rule, which is why the rule UID (with the model name + index) is used when inserting rules into the map.

The handling for this is broken when a rule is modified. This PR contributes, but it looks to me like it was already broken before, because when getting the "old rule", the index isn't used. As such, the "old rule" in the listener notification is always `null`. It doesn't matter that the "old rule" isn't actually removed, because since it's a `Map`, it's replaced when the "new rule" is inserted. But, the change notification is broken. I'm not sure what the consequences of that is.

On top of this, `removeRuleModel(ruleModelName)` is called as the very first thing in the "modified handing". I'm not sure what the rationale behind that is, but it means that even if the missing index is fixed, there will never be an "old rule" found. The reason for that is that `removeRuleModel()` is made to take the special index system into account, so it removes all models that *start with* the model name plus an index (handled by `belongsToModel()`).

This "special arrangement" also causes challenges for file format conversion, because it relies on temporary adding a model, process the rules, and then remove them. That doesn't work because of the index mismatching.

I'm not familiar with exactly what problem this PR addressed, nor with the consequence of the "modified notifications" missing the "old rule", but it is a bug nonetheless. I also need to find a way to deal with this strange system to do the format conversion.

@lolodomo I see that you have met some of these challenges in #5449, but it seems to me like the change notification is still broken.

I don't know how to proceed with what I'm doing without having some resolution to this, and I don't know what the time-frame for #5449 is.

Implementing user specifiable rule UIDs in DSL rules has further challenges (like how to handle the "mix" of auto-generated and manually specified UIDs), so I'm unsure if a separate fix for the model name issue should be created, or if I should wait for #5449.

I could make a fix for the issue from what I see here, but as always with DSL, it's hard to have the full overview of what implications any changes might have on other parts of the system.

lolodomo:
> I don't know how to proceed with what I'm doing without having some resolution to this, and I don't know what the time-frame for #5449 is.

I should finish #5449  tomorrow and I would like a merge very soon after.

I changed, like for things and items, to have a map per model name (and not per rule UID). You then have the list of rules for each model file.
When parsing for conversion, you will use a temporary model that does not impact the rule registry..

But you're right, there is no "modify" notification to the registry, except for script. All old rules from a model are first notified as removed and then rules from model are notified as added.
The difference with YAML is that there is no detection of what rules are really added, removed and updated. This is not a disaster and maybe things and items are working on the same principle (I don't remember).

So yes I would suggest you wait for the merge of #5449 .

Nadahar:
OK, I think we should fix the notification though, because as far as I can tell, the listener is the rule registry. I can't be good that it doesn't receive the notification it expects.

lolodomo:
And without the new UID, a rule could have its auto-generated UID changed when another rule was removed from the same model. It is what I try to avoid by adding a UID that can be set by the user.

lolodomo:
> It can't be good that it doesn't receive the notification it expects.

It is just not optimized but the registry gets the proper rules.

Nadahar:
> And without the new UID, a rule could have its auto-generated UID changed when another rule was removed from the same model. It is what I try to avoid by adding a UID that can be set by the user.

Yes, that's not good - but you must still "allow" for autogenerated UIDs, or all existing DSL rules will break. As far as I can tell, it's complicated to handle the combination of autogenerated and manually specified UIDs correctly.

PR #3729, opened 2023-07-24
[rest] Add caching for TagResource & De-duplicate code for caching

florian-h05:
This implements caching for GET `/rest/tags` and GET `/rest/tags/:tagname` as in #3335 and de-duplicates RegistryChangeListener code used by nearly all cached REST resources by introducing the `RegistryChangedRunnableListener` class.

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/thing/ThingResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/thing/ThingResource.java
@@ -54,7 +54,7 @@ import javax.ws.rs.core.UriInfo;
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.auth.Role;
-import org.openhab.core.common.registry.RegistryChangeListener;
+import org.openhab.core.common.registry.RegistryChangedRunnableListener;
 import org.openhab.core.config.core.ConfigDescription;
 import org.openhab.core.config.core.ConfigDescriptionRegistry;
 import org.openhab.core.config.core.ConfigUtil;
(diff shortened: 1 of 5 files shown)


spacemanspiff2007:
Would it be also possible to emit a change event or is this out of scope?
For HABApp the parameter is useless but an event would make caching possible.

florian-h05:
From my POV it's out of scope of this PR, in general it's up to the core maintainers to decide whether to merge such a PR, just ask them.

J-N-K:
Regarding this PR: Fine with me, however I would prefer to have the `RegistryChangedResetListener` more generic (i.e. a `ProviderChangeListener<T>` that accepts a `Runnable` in the constructor) so that we can simply call `new RegistryChangedResetListener(this::resetLastChanged)` instead of duplicating the same code again and again. Do you feel that you can do that? Otherwise I'll merge this now and put that on my list.

Regarding the event: I don't think there is a good argument against having an event for "tags changed". If someone provides a PR for that, I would be willing to review and merge. I guess the event type needs to be defined and the registry needs to be enhanced so that it gets a reference to the `EventPublisher`.

florian-h05:
> Do you feel that you can do that?

Finished! I created a `RegistryChangedRunnableListener` that executes a passed in Runnable on any type of registry changes and used this for Item, rule & Thing resources as well. The UI resource is using different code, I therefore didn't change it.

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/thing/ThingResource.java
```diff
@@ -171,7 +171,13 @@ public class ThingResource implements RESTResource {
...
     private final ThingStatusInfoI18nLocalizationService thingStatusInfoI18nLocalizationService;
     private final ThingTypeRegistry thingTypeRegistry;
-    private final ResetLastModifiedChangeListener resetLastModifiedChangeListener = new ResetLastModifiedChangeListener();
+
+    private void resetCacheableListLastModified() {
+        cacheableListLastModified = null;
+    }
+
+    private final RegistryChangedRunnableListener<Thing> resetLastModifiedChangeListener = new RegistryChangedRunnableListener<>(
+            this::resetCacheableListLastModified);
```
J-N-K, line 180:
```suggestion
    private final RegistryChangedRunnableListener<Thing> resetLastModifiedChangeListener = new RegistryChangedRunnableListener<>(
            () -> cacheableListLastModified = null);
```

This should also work.

Issue #3732, opened 2023-07-28
Use Quantity types for historic state functions.


mherwege:
This was first reported in the [openhab-js](https://github.com/openhab/openhab-js/issues/281) repository, but is a core issue.

I have following js rule code (generated by Blockly, but that is not relevant in the context):

```
updateTime = items.getItem('PVEtotalGS1').history.lastUpdate( 'influxdb').plusMinutes(1);
prevTotalGS1 = items.getItem('PVEtotalGS1').history.historicState(updateTime, 'influxdb')?.state;
deltaGS2 = String(items.getItem('PVEtotalGS2').history.deltaSince(updateTime, 'influxdb')) + ' kWh';
totalGS1 = (Quantity(prevTotalGS1).add(Quantity(deltaGS2).multiply(Quantity(cache.shared.get('GS_RATIO')))));
PVETotalGS1 and PVETotalGS2 are both Number:Energy items.
```

for history.historicState a state is returned, which can then be used for Quantity calculations in line 4. history.deltaSince returns a numeric (DecimalType, not a QuantityType), so I had to explicitely attach the unit. To be consistent, it would be more logical to return a QuantityType instead for this function as well.

The same argument would apply to a number of other history functions.

So, my ask is to return QuantityType instead of DecimalType for historic state functions where it makes sense. It probably means using State instead of DecimalType as a return type.

In my understanding, these functions are only used in rules. It needs checking how much change is required, but it may make working with QuantityType in rules easier. Right now, the rules need to reconstruct a QuantityType to be able to do Quantity calculations.

If this would be seen as useful, I could have a try implementing it in core, but would need help from the rule language library developers to adapt the rule language libraries. It would be a breaking change for rules.

PR #3736, opened 2023-07-31
Return units in persistence extension commands and support future persisted states

mherwege:
Closes https://github.com/openhab/openhab-core/issues/3732
Closes https://github.com/openhab/openhab-core/issues/3896

This PR implements unit support in all PersistenceExtension commands, returning the full value including unit for Number item types with a dimension. It therefore changes return types from DecimalType to a State type, whereby the state can include the unit.
This has been tested to work with rules DSL, and it should be possible to extend other scripting languages to support this, making calculations with historic states of items with dimensions more straightforward.
This may require adapting user rules and is therefore a breaking change.

Secondly, this PR makes sure persistence extensions work properly with persisted items in the future:
- all `Between` commands have been adapted to be able to properly work across the `now` boundary,
- a series of `Untill` commands have been added, equivalent to the `Since` commands.
- `nexUpdate` and `nexState` commands have been added.
- the `historicState` command has been deprecated (but still available for backward compatibility) and replaced with `persistedState` also accepting future dates.
- the `evolutionRate` command has been renamed to `evolutionRateSince`, `evolutionRateUntill` and `evolutionRateBetween`, bringing it in line with the other commands. This change is required to allow one date signatures to differentiate between `Since` and `Untill`. `evolutionRate` is deprecated and kept for backward compatibility, and interpreted as `evolutionRateSince`.
- added `removeAllStates` `Since`, `Untill` and `Between` commands.

The code now also has full null annotations and checks. In doing so, changes were required to avoid potential null issues. Many of these null issues where due to null being interpreted as now for end dates before, which is not possible when time can be in the future.

The test classes where extended and refactored to also test for all future commands.

Signed-off by: Mark Herwege mark.herwege@telenet.be

### DIFF
--- a/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
+++ b/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
@@ -444,7 +444,7 @@ public class PersistenceExtensions {
     /**
      * Returns the next state of a given <code>item</code>.
      *
-     * @param item the item to get the previous state value for
+     * @param item the item to get the next state value for
      * @param skipEqual if true, skips equal state values and searches the first state not equal the current state
      * @return the next state or <code>null</code> if no next state could be found, or if the default
      *         persistence service is not configured or does not refer to a {@link QueryablePersistenceService}
@@ -1439,9 +1439,8 @@ public class PersistenceExtensions {
      *
      * @param item the item for which we will sum its persisted state values since <code>timestamp</code>
      * @param timestamp the point in time from which to start the summation
-     * @return the sum of the state values since <code>timestamp</code>, or {@link DecimalType#ZERO} if no historic
-     *         states could be found or if the default persistence service does not refer to a
-     *         {@link QueryablePersistenceService}
+     * @return the sum of the state values since <code>timestamp</code>, or null if <code>timestamp</code> is in the
+     *         future or the default persistence service does not refer to a {@link QueryablePersistenceService}
      */
     public @Nullable static State sumSince(Item item, ZonedDateTime timestamp) {
         return internalSumBetween(item, timestamp, null);
@@ -1453,9 +1452,8 @@ public class PersistenceExtensions {
      *
      * @param item the item for which we will sum its persisted state values to <code>timestamp</code>
      * @param timestamp the point in time to which to start the summation
-     * @return the sum of the state values to <code>timestamp</code>, or {@link DecimalType#ZERO} if no historic
-     *         states could be found or if the default persistence service does not refer to a
-     *         {@link QueryablePersistenceService}
+     * @return the sum of the state values until <code>timestamp</code>, or null if <code>timestamp</code> is in the
+     *         past or the default persistence service does not refer to a {@link QueryablePersistenceService}
      */
     public @Nullable static State sumUntil(Item item, ZonedDateTime timestamp) {
         return internalSumBetween(item, null, timestamp);
@@ -1469,8 +1467,8 @@ public class PersistenceExtensions {
      *            <code>end</code>
      * @param begin the point in time from which to start the summation
      * @param end the point in time to which to start the summation
-     * @return the sum of the state values between the given points in time, or {@link DecimalType#ZERO} if no historic
-     *         states could be found or if the default persistence service does not refer to a
+     * @return the sum of the state values between the given points in time, or null if <code>begin</code> is after
+     *         <code>end</code> or if the default persistence service does not refer to a
      *         {@link QueryablePersistenceService}
      */
     public @Nullable static State sumBetween(Item item, ZonedDateTime begin, ZonedDateTime end) {
@@ -1484,9 +1482,8 @@ public class PersistenceExtensions {
      * @param item the item for which we will sum its persisted state values since <code>timestamp</code>
      * @param timestamp the point in time from which to start the summation
      * @param serviceId the name of the {@link PersistenceService} to use
-     * @return the sum of the state values since the given point in time, or {@link DecimalType#ZERO} if no historic
-     *         states could be found for the <code>item</code> or if <code>serviceId</code> does not refer to a
-     *         {@link QueryablePersistenceService}
+     * @return the sum of the state values since <code>timestamp</code>, or null if <code>timestamp</code> is in the
+     *         future or <code>serviceId</code> does not refer to a {@link QueryablePersistenceService}
      */
     public @Nullable static State sumSince(Item item, ZonedDateTime timestamp, String serviceId) {
         return internalSumBetween(item, timestamp, null, serviceId);
@@ -1499,9 +1496,8 @@ public class PersistenceExtensions {
      * @param item the item for which we will sum its persisted state values to <code>timestamp</code>
      * @param timestamp the point in time to which to start the summation
      * @param serviceId the name of the {@link PersistenceService} to use
-     * @return the sum of the state values to the given point in time, or {@link DecimalType#ZERO} if no historic
-     *         states could be found for the <code>item</code> or if <code>serviceId</code> does not refer to a
-     *         {@link QueryablePersistenceService}
+     * @return the sum of the state values until <code>timestamp</code>, or null if <code>timestamp</code> is in the
+     *         past or <code>serviceId</code> does not refer to a {@link QueryablePersistenceService}
      */
     public @Nullable static State sumUntil(Item item, ZonedDateTime timestamp, String serviceId) {
         return internalSumBetween(item, null, timestamp, serviceId);
@@ -1516,9 +1512,8 @@ public class PersistenceExtensions {
      * @param begin the point in time from which to start the summation
      * @param end the point in time to which to start the summation
      * @param serviceId the name of the {@link PersistenceService} to use
-     * @return the sum of the state values between the given points in time, or {@link DecimalType#ZERO} if no historic
-     *         states could be found for the <code>item</code> or if <code>serviceId</code> does not refer to a
-     *         {@link QueryablePersistenceService}
+     * @return the sum of the state values between the given points in time, or null if <code>begin</code> is after
+     *         <code>end</code> or <code>serviceId</code> does not refer to a {@link QueryablePersistenceService}
      */
     public @Nullable static State sumBetween(Item item, ZonedDateTime begin, ZonedDateTime end, String serviceId) {
         return internalSumBetween(item, begin, end, serviceId);
--- a/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
+++ b/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
@@ -1678,6 +1678,27 @@ public class PersistenceExtensions {
      * Gets the evolution rate of the state of a given {@link Item} since a certain point in time.
      * The default {@link PersistenceService} is used.
      *
+     * This method has been deprecated and {@link #evolutionRateSince(Item, ZonedDateTime)} should be used instead.
+     *
+     * @param item the item to get the evolution rate value for
+     * @param timestamp the point in time from which to compute the evolution rate
+     * @return the evolution rate in percent (positive and negative) between now and then, or <code>null</code> if
+     *         there is no default persistence service available, the default persistence service is not a
+     *         {@link QueryablePersistenceService}, or if there is no persisted state for the given <code>item</code> at
+     *         the given <code>timestamp</code>, or if there is a state but it is zero (which would cause a
+     *         divide-by-zero error)
+     */
+    @Deprecated
+    public static @Nullable DecimalType evolutionRate(Item item, ZonedDateTime timestamp) {
+        LoggerFactory.getLogger(PersistenceExtensions.class).info(
+                "The evolutionRate method has been deprecated and will be removed in a future version, use evolutionRateSince instead.");
+        return internalEvolutionRateBetween(item, timestamp, null);
+    }
+
+    /**
+     * Gets the evolution rate of the state of a given {@link Item} since a certain point in time.
+     * The default {@link PersistenceService} is used.
+     *
      * @param item the item to get the evolution rate value for
      * @param timestamp the point in time from which to compute the evolution rate
      * @return the evolution rate in percent (positive and negative) between now and then, or <code>null</code> if
@@ -1721,6 +1742,30 @@ public class PersistenceExtensions {
      */
     public static @Nullable DecimalType evolutionRateBetween(Item item, ZonedDateTime begin, ZonedDateTime end) {
         return internalEvolutionRateBetween(item, begin, end);
+    }
+
+    /**
+     * Gets the evolution rate of the state of a given {@link Item} since a certain point in time.
+     * The {@link PersistenceService} identified by the <code>serviceId</code> is used.
+     *
+     * This method has been deprecated and {@link #evolutionRateSince(Item, ZonedDateTime, String)} should be used
+     * instead.
+     *
+     * @param item the {@link Item} to get the evolution rate value for
+     * @param timestamp the point in time from which to compute the evolution rate
+     * @param serviceId the name of the {@link PersistenceService} to use
+     * @return the evolution rate in percent (positive and negative) between now and then, or <code>null</code> if
+     *         the persistence service given by <code>serviceId</code> is not available or is not a
+     *         {@link QueryablePersistenceService}, or if there is no persisted state for the given
+     *         <code>item</code> at the given <code>timestamp</code> using the persistence service given by
+     *         <code>serviceId</code>, or if there is a state but it is zero (which would cause a divide-by-zero
+     *         error)
+     */
+    @Deprecated
+    public static @Nullable DecimalType evolutionRate(Item item, ZonedDateTime timestamp, String serviceId) {
+        LoggerFactory.getLogger(PersistenceExtensions.class).info(
+                "The evolutionRate method has been deprecated and will be removed in a future version, use evolutionRateSince instead.");
+        return internalEvolutionRateBetween(item, timestamp, null, serviceId);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


J-N-K:
I understand what you want to achieve, but why did you chose `HistoricItem` as return value? The timestamp has no meaning for averages or similar, a simple `State` would be fine (and could be a `DecimalType` or a `QuantityType` (depending on the item's unit).

mherwege:
> I understand what you want to achieve, but why did you chose `HistoricItem` as return value? The timestamp has no meaning for averages or similar, a simple `State` would be fine (and could be a `DecimalType` or a `QuantityType` (depending on the item's unit).

HistoricItem was already used for minimum, maximum and previous, so wanted to stick to that. But I agree the date returned in the HistoricItem is arbitrary. I will adjust.

mherwege:
@J-N-K the methods now return State.

As this is API breaking anyway, I also aligned method names. EvolutionRate whas the only group of methods that did not have a distinction between Since and Between in the name. I believe this is a rarely used method anyway and I think it is cleaner to have them aligned. If you feel differently, I can revert this last commit.

florian-h05:
Any chance to get this merged for openHAB 4.1 or is that something for a new major release because it's breaking?

kaikreuzer:
@J-N-K As you had commented before, I hoped that you continue on this PR - any chance to do so?

> This may require adapting user rules and is therefore a breaking change.

Breaking user rules is never nice. I'm just trying to get my head around how much this might break. Do you have a typical example where user's will have to adapt their rules? Are these rather exceptional cases or do most rules break by this?
Is there any migration path that we could document as a "quick fix"?
I am not 100% sure whether we should merge this now so shortly before the release. Waiting for a next major release isn't really an option either, that will still take a while...

florian-h05:
If this gets merged for 4.1, we should try to implement a solution for #3896 for 4.1 as well to avoid having multiple sets of breaking changes.

J-N-K:
I have a rule that calculates the mean temperature over the last three days. It uses averageSince and then adds the unit before posting the update to the item.

I still think this is a good addition, but I have no idea how to make it "less breaking".

mherwege:
> I have a rule that calculates the mean temperature over the last three days. It uses averageSince and then adds the unit before posting the update to the item.

Similar for me, and that was the reason to create this PR. It forces you to add the unit again after retrieving the average. That means you need to 'know' the unit in the rule when retrieving to put the right unit back and it goes against the principle of UOM where you don't care.

florian-h05:
> If we all agree we want to do this, I could extend this PR to include a solution for #3896.

I agree, but I'm no core maintainer ;-)
Thank you for the offer! 

> Scripting languages may also be affected. Would any change be required in the javascript helper libraries?

Yes, for renamed methods there would be changes required, however those are minor as the library is only wrapping `PersistenceExtensions` inside a JS class to return JS datatypes.
I would also keep the old ones in the library and log a deprecation warning, so this is less breaking.

For the new functionalities, I have to think about how to integrate them in the library, but I already have an idea.

I can't speak for jRuby, but I guess HABApp is not affected because it is using the REST API to communicate with openHAB and not any Java calls.

kaikreuzer:
> If we all agree we want to do this

I agree that I want this change, I am just not sure when (before or after 4.1). I tend towards waiting after the release, so that we can get some feedback by snapshot and milestone users first and maybe come up with a good migration plan for 4.2.

florian-h05:
Yeah I've already asked myself when this should be implemented.

On one hand, it would be nice to provide the right "tools" together with the new time-series support, 
but on the other hand 4.0 had several breaking changes (and so it would be "nice" to have less/no breaking changes in 4.1) and 1.5 weeks till the 4.1 code-freeze is also a bit short to coordinate the changes across all repos and test this.

J-N-K:
> Similar for me, and that was the reason to create this PR. It forces you to add the unit again after retrieving the average. That means you need to 'know' the unit in the rule when retrieving to put the right unit back and it goes against the principle of UOM where you don't care.

Full ack. But nevertheless all existing rules using persistence extensions need to be changed.

spacemanspiff2007:
> Similar for me, and that was the reason to create this PR. It forces you to add the unit again after retrieving the average.

But why? I thought when the unit is properly set you can just post an empty value and the unit from the metadata will be assumed? Imho you should just be able to post the unitless value and everything should work as expected.

mherwege:
> But why? I thought when the unit is properly set you can just post a unit with an empty value and the unit from the metadata will be assumed? Imho you should just be able to post the unitless value and everything should work as expected.
> The whole point of explicitly providing the unit in the item definition is that there are no ambiguites and the unit can always properly be resolved.

That's true if you immediately post an update to an item. Sometimes, these are intermediate steps in calculations. When you retrieve something from persistence, you have to know what unit it is, and you need to consider that in each step in the calculation, convert the value if necessary yourself as it may be in the wrong unit, and then post an update to an item with the value in the unit the item expects. So the rule needs to know in each step what unit it is working in, but it is not explicit. I prefer to have the Quantities with units in the calculations all the way, to avoid all of this. That way the rule does not have to know the unit it retrieves from persistence and the unit of the item it writes to.

mherwege:
> I agree that I want this change, I am just not sure when (before or after 4.1). I tend towards waiting after the release, so that we can get some feedback by snapshot and milestone users first and maybe come up with a good migration plan for 4.2.

I will work on extending this with logic for future states with target 4.2 for the whole PR.

spacemanspiff2007:
> That's true if you immediately post an update to an item that has the same unit in metadata as the unit for the item you retrieve from persistence.

That's exactly the example @J-N-K made above and that's why I was confused because this should already work.

J-N-K:
My example was just a rule that needs adjustment, not an example to justify this change :-)

J-N-K:
What's the state here? Are the discussed extensions already included and we can review and merge? The sooner, the better: more time for testing.

mherwege:
> What's the state here? Are the discussed extensions already included and we can review and merge? The sooner, the better: more time for testing.

Sorry, not yet. And I won’t be able to work on it for the next few days.

mherwege:
@J-N-K, @florian-h05 I completed the work on persistence extensions, so this is now ready for review.
I messed up my local git history, so had to force push a new state with the full content of the file. I am sorry about that.

florian-h05 (commented):
API looks good, thanks!

I will try to test this soon.

jimtng:
Sorry to be so late to the party. I would suggest calling the methods `Until` instead of `Till`.

mherwege:
> Sorry to be so late to the party. I would suggest calling the methods `Until` instead of `Till`.

Fine for me, change done.

jimtng:
> Fine for me, change done.

Thanks so much! The OCD in me would love the "till" in the comments also updated to say "until" to be more consistent.

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
```diff
@@ -1281,68 +2122,259 @@ public static Iterable<HistoricItem> getAllStatesBetween(Item item, ZonedDateTim
...
+                }
+            }
+        }
+        return item.getStateAs(DecimalType.class);
     }
 
-    private static @Nullable HistoricItem historicItemOrCurrentState(Item item, HistoricItem historicItem,
-            DecimalType value) {
-        if (historicItem == null && value != null) {
+    private static @Nullable HistoricItem historicItemOrCurrentState(Item item, @Nullable HistoricItem historicItem) {
```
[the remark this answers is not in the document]
mherwege: This would be API breaking. HistoricItem a a State with a ZonedDateTime. It is returned when the ZonedDateTime makes sense. My original proposal was to use HistoricItem everywhere, but @J-N-K correctly reacted that it does not make sense for many of the methods. So I kept HistoricItem where it was already returned (and it makes sense) and use State where the ZonedDateTime does not make sense.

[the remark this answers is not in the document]
jimtng: No, I didn't mean to return HistoricItem everywhere. 

I mean that for methods that return a HistoricItem (such as maximumSince), or an Iterable<HistoricItem> (such as getAllStatesSince), make sure that the historicItem.getState() would return a QuantityType instead of DecimalType where applicable.

So from an API consumer's perspective:

```java
HistoricItem historyMax = PersistenceExtensions.maximumSince(myItem, ZonedDateTime.now().minusDays(1));
var max = historyMax.getState(); // <= this will be a QuantityType instead of DecimalType
```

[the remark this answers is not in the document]
mherwege: Unless I am missing something, I believe that is case already. There are also a series of tests testing these methods with QuantityTypes.

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
```diff
@@ -207,11 +427,77 @@ public static void persist(Item item) {
...
+     *         persistence service is not configured or does not refer to a {@link QueryablePersistenceService}
+     */
+    public static @Nullable HistoricItem nextState(Item item) {
+        return internalAdjacentState(item, false, true);
+    }
+
+    /**
+     * Returns the next state of a given <code>item</code>.
+     *
+     * @param item the item to get the previous state value for
```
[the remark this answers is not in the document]
mherwege: Done

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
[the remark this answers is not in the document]
jimtng: My apologies, I didn't look at the tests, you're correct. Where did the QuantityType come from? Did it come from the persistence service?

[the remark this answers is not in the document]
mherwege: Yes, indeed.

mherwege:
I am going through the javadocs again. Many of them need updates or clarifications. I am working on that and will come back with refinements on them.

mherwege:
> I am going through the javadocs again. Many of them need updates or clarifications. I am working on that and will come back with refinements on them.

Done, and removed some commented code and solved an issue.

jimtng:
> Scripting languages may also be affected. Would any change be required in the javascript helper libraries? What about jruby

If the old `evolutionRate()` were to remain, perhaps with a deprecation warning like `historicState`, then it wouldn't break any jruby scripts, and perhaps also other scripting languages?

mherwege:
> If the old `evolutionRate()` were to remain, perhaps with a deprecation warning like `historicState`, then it wouldn't break any jruby scripts, and perhaps also other scripting languages?

Done that now. A warning similar to `historicState` will be logged.

jimtng:
If this is not going to be API Breaking, the label should be removed.

J-N-K:
I this now ready for review?

mherwege:
> I this now ready for review?

It is from my perspective. I already addressed a number of review comments from other people.

florian-h05:
@openhab/core-maintainers Can you please review this PR? 
It should not be merged too late before the 4.2 release as we also need some time to adjust the helper libraries and the documentation, so the earlier the better.

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
```diff
@@ -67,52 +91,193 @@ public PersistenceExtensions(@Reference PersistenceServiceRegistry registry) {
      * @param serviceId the name of the {@link PersistenceService} to use
      */
     public static void persist(Item item, String serviceId) {
+        internalPersist(item, serviceId);
+    }
+
+    private static void internalPersist(Item item) {
+        String serviceId = getDefaultServiceId();
+        if (serviceId == null) {
```
J-N-K, line 99: Shouldn't this result in a warning similar to the one on l. 111? ("No default persistence service configured")

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
```diff
@@ -67,52 +91,193 @@ public PersistenceExtensions(@Reference PersistenceServiceRegistry registry) {
...
+     * @param state the state to be stored
+     * @param serviceId the name of the {@link PersistenceService} to use
+     */
+    public static void persist(Item item, ZonedDateTime timestamp, State state, String serviceId) {
+        internalPersist(item, timestamp, state, serviceId);
+    }
+
+    private static void internalPersist(Item item, ZonedDateTime timestamp, State state) {
+        String serviceId = getDefaultServiceId();
+        if (serviceId == null) {
```
J-N-K, line 142: see above

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
```diff
@@ -67,52 +91,193 @@ public PersistenceExtensions(@Reference PersistenceServiceRegistry registry) {
...
+    }
+
+    private static @Nullable HistoricItem internalPersistedState(Item item, @Nullable ZonedDateTime timestamp,
+            String serviceId) {
+        if (timestamp == null) {
+            return null;
+        }
         PersistenceService service = getService(serviceId);
-        if (service instanceof QueryablePersistenceService qService) {
+        if (service != null && service instanceof QueryablePersistenceService qService) {
```
J-N-K, line 280: Why is that needed? If `service` is an instance of `QueryablePersistenceService` is is not `null` (also applies to other methods).

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
```diff
@@ -124,52 +289,89 @@ public static void persist(Item item) {
             } else {
                 return null;
             }
-        } else {
-            LoggerFactory.getLogger(PersistenceExtensions.class)
-                    .warn("There is no queryable persistence service registered with the id '{}'", serviceId);
-            return null;
         }
```
J-N-K, line 292: I would prefer to keep the `else` branch, it is more clear (also applies to other methods).

J-N-K (changes requested):
In general looks very good to me. This is a tremendous work and makes the persistence extensions more consistent and usable.

I have left some minor comments and can you resolve the merge conflict? Thanks.

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
mherwege: `getDefaultServiceId()` already logs a warning.

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
mherwege: OK

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
mherwege: Indeed, not required.

mherwege:
@J-N-K I rebased and made some changes relative to the review comments.

J-N-K (approved):
Thanks! Can you als add a PR for the breaking-change notice in openhab-distro and adjust the documentation on okpenhab-docs?

florian-h05:
FYI openHAB JavaScript library PR to adjust to this PR is already created: https://github.com/openhab/openhab-js/pull/331
Blockly needs to be updated next ...

lolodomo:
@mherwege : does it require a note in the release notes to mention a breaking change ?

J-N-K:
Already done:https://github.com/openhab/openhab-distro/pull/1652

jimtng:
core build seems to be on a daily basis

jimtng:
openhab-core build is done now. Next, we need the distro build to kick in

J-N-K:
Distro-build should not be required to build webui

jimtng:
I've just noticed that the "between" variant of the old `evolutionRate()` was removed in this PR, instead of deprecated. Should we add it back for the sake of compatibility?

mherwege:
> I've just noticed that the "between" variant of the old `evolutionRate()` was removed in this PR, instead of deprecated. Should we add it back for the sake of compatibility?

Yes, makes sense. I missed that one.

jimtng:
Another change I've just noticed: Previously I could call any methods with a null serviceid. 

Now, due to `@NonNullByDefault` on the `PersistenceExtensions` class, and the absence of `@Nullable` in front of serviceId, this is no longer possible.

This change has caused some errors in existing scripts. Should we restore the previous behaviour?

jimtng:
@mherwege I edited/updated the [post above](https://github.com/openhab/openhab-core/pull/3736#issuecomment-2093140530), in case you've only read the original message.

mherwege:
> Another change I've just noticed: Previously I could call any methods with a null serviceid to indicate that the default persistence service is to be used.

It has never been documented it could be used that way. Also, quickly looking through the old code again, it doesn't look like it worked for all actions, and it even gave wrong result for some (part of the query working and another part returning null and defaulting to a wrong default value, so the combination being wrong).
While it is certainly doable to introduce this, I am wondering if it is worth doing this to fix wrong usage of the actions in the first place.

mherwege:
@jimtng PR is created to allow null serviceId.

jimtng:
> @jimtng PR is created to allow null serviceId.


Thanks! I was pondering about your previous response. I agree that it wasn't documented in the javadocs. I've started working on modifications in the jruby helper library to avoid using a null serviceId. However, I think it's a handy thing to support but either way is fine.

njordan77:
Hello, i changed to M3 and now a rule does throw an error based on this line. Says "unknown variable or line" pointint to the "minus"... Any idea what i have to use instead? Thanks Norbert
`
var AvailChargePower = (miFro_ACpower.averageSince(now.minusMinutes(2)) - 400)`

PR #3742, opened 2023-08-05
Move factory interface to public package

J-N-K:
The factory interface should be available for usage in bindings, so it needs to be in an exported package. This was overlooked when the `AbstractTransformingChannelHandler` was moved to an internal package.


J-N-K:
@kaikreuzer @wborn Can someone merge this simple fix? We need it in openhab-addons.

wborn (approved):
Thanks!

Issue #3718, opened 2023-07-21
Java 11.0.20 breaks OpenHAB 3.4.4 (Improved ZIP64 Extra Field Validation)


RolandRosenfeld:
After upgrade of zulu11 Java from 11.0.19 to 11.0.20 starting OpenHAB 3.4.4 failed starting with messages like
```
2023-07-21 14:55:10.438 [ERROR] [Events.Framework                    ] - FrameworkEvent ERROR
java.util.zip.ZipException: Exception in opening zip file: /var/lib/openhab/cache/org.eclipse.osgi/18/0/bundleFile
[...]
Caused by: java.util.zip.ZipException: Invalid CEN header (invalid zip64 extra data field size)

```

This seems to be triggered by the following change in Java 11.0.20 (same with 17.0.8):
https://www.oracle.com/java/technologies/javase/11all-relnotes.html 

> core-libs/java.util.jar
> [➜](https://www.oracle.com/java/technologies/javase/11all-relnotes.html#JDK-8302483) Improved ZIP64 Extra Field Validation (JDK-8302483 (not public))
> java.util.zip.ZipFile has been updated to provide additional validation of ZIP64 extra fields when opening a ZIP file. This validation may be disabled by setting the system property jdk.util.zip.disableZip64ExtraFieldValidation to true.

This topic is already discussed in the community forum: https://community.openhab.org/t/openhab-not-working-with-zulu11-0-20/147910

## Expected Behavior

OpenHAB should work with recent Java version.

## Current Behavior
OpenHAB 3.4.4 fails starting with suggested zulu11 in most recent 11.0.20 version.

## Possible Solution

As a workaround it helps adding
```
EXTRA_JAVA_OPTS="-Djdk.util.zip.disableZip64ExtraFieldValidation=true"
```
to /etc/default/openhab (for a Debian system, this may be different on other systems).

It may be better to either change the code to support additional ZIP64 extra fields or at least do some Java voodoo to automatically set the above option.

## Steps to Reproduce (for Bugs)

1. Start with a working system with OpenHAB 3.4.4 and ZULU11 11.0.19
2. Upgrade ZULU11 to 11.0.20
3. OpenHAB does not work any more.

## Your Environment

* Version used: OpenHAB 3.4.4-2 (Debian Package), zulu11 11.0.20-1 (Debian Package), both installed via apt repos.
* Debian 11 (bullseye) on amd64 architecture

rkoshak:
[...]

PR #3747, opened 2023-08-09
ZipException workarounds

wborn:
I ran into some more build issues due to this "Improved ZIP64 Extra Field Validation" (see https://github.com/openhab/openhab-core/issues/3718)  now that Temurin 17.0.8 upgrades are available in APT repos.

The org.openhab.core.karaf bundle fails to build as Java thinks there are some issues with org.apache.karaf.features.core-4.4.0.jar. However `zip -T ...` thinks it's OK. So perhaps this is a false positive. It doesn't complain about org.apache.karaf.features.core-4.4.3.jar so let's use that instead.

Also the itests generate lots of ZipException stacktraces due to the issue with geronimo-osgi-locator.jar.

### DIFF
--- a/itests/itest-include.bndrun
+++ b/itests/itest-include.bndrun
@@ -19,7 +19,9 @@ Test-Cases: ${classes;CONCRETE;PUBLIC;NAMED;*Test}
 
 # An unused random HTTP port is used during tests to prevent resource conflicts
 # This property is set by the build-helper-maven-plugin in the itests pom.xml
--runvm: -Dorg.osgi.service.http.port=${org.osgi.service.http.port}
+-runvm: \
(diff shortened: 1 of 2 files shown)


J-N-K (approved):
Thanks!

morph166955:
While not an issue persay, this seems to break the ability to create the hybrid 3.x 4.x jars.  Noticed yesterday when compiling androidtv.

wborn:
That is to be expected and it will get more complicated over time as differences between branches grow. It's not recommended and unsupported for various reasons.

morph166955:
Fully agreed and not trying to say we need to do anything to resolve it.  Just noting for anyone who is doing the hybrids where the break occurred for later on.

PR #3748, opened 2023-08-11
Fix StackOverflowError in SemanticsMetadataProvider

J-N-K:
In case of cyclic/recursive membership of groups (i.e. a group has itself or a child as parent) the recursive processing of the items leads to an `StackOverflowError` (reported in the forum and the german Facebook group). Also this is clearly a configuration error it should not result in the bundle not starting.

### DIFF
--- a/bundles/org.openhab.core.semantics/pom.xml
+++ b/bundles/org.openhab.core.semantics/pom.xml
@@ -21,6 +21,12 @@
       <version>${project.version}</version>
     </dependency>
     <dependency>
+      <groupId>org.openhab.core.bundles</groupId>
+      <artifactId>org.openhab.core.test</artifactId>
+      <version>${project.version}</version>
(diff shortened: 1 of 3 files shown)


wborn (commented):
The fix looks good to me. 👍 Would it also be possible to add a unit test for this? Did you also consider using the warn loglevel? It is quite severe but I don't know if it is error level severe because the system can handle it.

J-N-K:
I chose the same log level that is used here: https://github.com/openhab/openhab-core/blob/04105c1b1a35e6147c713faf8e7091bf1a3884fe/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/item/EnrichedItemDTOMapper.java#L107-L109

wborn (approved):
Thank you for also adding the tests and log level explanation. 👍

PR #3766, opened 2023-08-20
[webaudio] Allow stopping play

dalgwen:
By sending an empty string we can ask the webui to stop playing sound, as specification says.
See [here ](https://github.com/openhab/openhab-webui/pull/2031)for the corresponding webui modification.

Also : By removing the URLAudioStream special case, we force the webaudio servlet to act as a proxy for every audio. It has the benefit of eliminitating CORS issue.
But there is still an issue with infinite stream. webUI (and webaudio API as it is used) does not support unlimited stream. I'm not knowledgeable enough of the webaudio API for this kind of modification (and it seems pretty hard).

The two modifications comes from this community discussion :
https://community.openhab.org/t/webaudio-not-working-when-trying-to-stream-webradio/148610/18

### DIFF
--- a/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/webaudio/WebAudioAudioSink.java
+++ b/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/webaudio/WebAudioAudioSink.java
@@ -24,7 +24,6 @@ import org.openhab.core.audio.AudioSink;
 import org.openhab.core.audio.AudioSinkAsync;
 import org.openhab.core.audio.AudioStream;
 import org.openhab.core.audio.StreamServed;
-import org.openhab.core.audio.URLAudioStream;
 import org.openhab.core.audio.UnsupportedAudioFormatException;
 import org.openhab.core.audio.UnsupportedAudioStreamException;
 import org.openhab.core.events.EventPublisher;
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/webaudio/WebAudioAudioSink.java
```diff
@@ -67,29 +66,19 @@ public void processAsynchronously(@Nullable AudioStream audioStream)
         if (audioStream == null) {
             // in case the audioStream is null, this should be interpreted as a request to end any currently playing
             // stream.
-            logger.debug("Web Audio sink does not support stopping the currently playing stream.");
+            sendEvent("");
             return;
         }
         logger.debug("Received audio stream of format {}", audioStream.getFormat());
-        if (audioStream instanceof URLAudioStream urlAudioStream) {
```
J-N-K, line 74: Why can we remove this different handling here?

dalgwen: > Why can we remove this different handling here?

I should have explained it better, sorry.

The two cases, as it is now :
- if we recognize an URLAudioStream, we send the URL to the webui, and it will play it directly. It can be (and probably will ?) on another network host / internet. So potential CORS issue.
- If it is not an URLAudioStream, we don't have the choice : we set up the audio servlet to serve the sound. We then send the audioservlet URL to the webui, and the webui will call it to get the sound. From the webui point of view, it is local, so no CORS issue.

By removing the first case, we ensure that the webui will always use the openHAB audio servlet to get the sound.
The URLAudioStream will be used as any other kind of stream : the audioservlet will get the audio sound and act as a proxy.

> But there is still an issue with infinite stream. webUI (and webaudio API as it is used) does not support unlimited stream. I'm not knowledgeable enough of the webaudio API for this kind of modification (and it seems pretty hard).

I forgot to mention what happens. The webui will try to fully load the sound before playing it (no streaming support for the kind of play method we use here). Of course an infinite stream has no end, and therefore the sound will never start playing.  Incredibly, streaming audio is not standard with the webaudio API.

J-N-K (approved):
LGTM

PR #3767, opened 2023-08-20
Migrate to java.nio.file.createTempFile

holgerfriedrich:
Use function from nio package as it uses more restrictive file permissions.

I did not test this. We use nio.file.createTempFile in all other places in core.


J-N-K (approved):
Thanks!

PR #3769, opened 2023-08-21
Fix sitemap visibility rule with blanks

mherwege:
See https://github.com/openhab/openhab-webui/pull/1487#issuecomment-1673764126

Sitemap visibility and color rules, maintained in mainUI failed with blanks in the condition state. This happens because string quotes are stripped from the json REST format.

This fix modifies the regex identifying the condition states to allow any character, with the exception of = in color rules, as that will identify the target color.


J-N-K (approved):
Thanks!

PR #3775, opened 2023-08-22
Add TrustAllSslContextFactory for self-signed certificates

jimtng:
It suppresses warnings about trusting all certificates during instantiation.

This is added to core so that all add-ons can use it to address https://github.com/openhab/openhab-addons/issues/10446

Changes in the binding code to use `TrustAllSslContextFactory` instead of `SslContextFactory.Client(true)` will also be required.

### DIFF
--- a/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/TrustAllSslContextFactory.java
+++ b/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/TrustAllSslContextFactory.java
@@ -1,0 +1,41 @@
+/**
+ * Copyright (c) 2010-2023 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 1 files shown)


jimtng:
The way it could work is this:

- Each binding can add a configuration called `verifyCertificate` which is set to `true` (or false as they may deem reasonable) by default. Then when it's true, the binding code would use the standard SslContextFactory, which will fail the connection when it encounters a self-signed and/or mismatched CN (Certificate Name), amongst other things that it would verify. That is the proper and standard way to guard against MITM attacks.
- When `verifyCertificate` config is set to false, the binding code would use this TrustAllSslContextFactory. 

Right now bindings are already using some form of trustall mechanism against everything, using `new SslCertificateFactory.Client(true)`. So nothing is changed by this PR, only that it offers a way to suppress the useless warning **at start up** of the binding, not at connection time.

wborn:
> Do you have a scheme/mechanism in mind?

A Thing could for instance have a configuration option to trust only CA signed certificates (default) or also a self signed certificate. When a self signed certificate is trusted it will download it and add it to a Storage automatically and from then trust this self signed certificate for the Thing connection. This would be similar to how we also keep track of the OAuth2 tokens for Things in a Storage. Then you can add some console commands or add an API/UI to show and modify whats in this Storage.

PR #3779, opened 2023-08-26
Fix GenericThingProvider

J-N-K:
When things from multiple handler factories are provisioned from the same file the internal thing map was not properly updated.

### DIFF
--- a/bundles/org.openhab.core.model.thing/src/org/openhab/core/model/thing/internal/GenericThingProvider.xtend
+++ b/bundles/org.openhab.core.model.thing/src/org/openhab/core/model/thing/internal/GenericThingProvider.xtend
@@ -614,9 +614,9 @@ class GenericThingProvider extends AbstractProviderLazyNullness<Thing> implement
                 createThing(newThings, factory)
             ]
         }
-        thingsMap.put(modelName, newThings)
 
         newThings.forEach [ newThing |
+            thingsMap.get(modelName).add(newThing)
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.model.thing/src/org/openhab/core/model/thing/internal/GenericThingProvider.xtend
```diff
@@ -614,9 +614,9 @@ class GenericThingProvider extends AbstractProviderLazyNullness<Thing> implement
                 createThing(newThings, factory)
             ]
         }
-        thingsMap.put(modelName, newThings)
 
         newThings.forEach [ newThing |
+            thingsMap.get(modelName).add(newThing)
```
wborn, line 619: Is `thingsMap.get(modelName)` guaranteed to be non-null? I see the method is also called from `thingHandlerFactoryAdded` .

J-N-K: I think yes, because when called from `thingHandlerFactoryAdded` the model names are retrieved from the keys of the `thingsMap` and the `thingsMap` has non-null values.

wborn (approved):
Thanks for the fix! Let's give it a test.

andrewfg:
@J-N-K I am not 100% sure but I think I also observe on OH 4.0.3 the same issue with my .items file (which contains items for all of my things). Could that be a similar issue? Or am I dreaming?

PR #3785, opened 2023-08-28
[cache] Fix concurrent use of a cache entry

### LINKED ISSUE #3507: LRUMediaCacheEntryTest unstable
This test failed in a Jenkins build: https://ci.openhab.org/job/PR-openHAB-Core/5631

```
[ERROR] Tests run: 5, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1.524 s <<< FAILURE! - in org.openhab.core.cache.lru.LRUMediaCacheEntryTest
[ERROR] org.openhab.core.cache.lru.LRUMediaCacheEntryTest.loadTwoThreadsAtTheSameTimeFromTheSameSupplierTest  Time elapsed: 0.028 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: array lengths differ, expected: <102400> but was: <0>
	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
	at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
	at org.junit.jupiter.api.AssertArrayEquals.assertArraysHaveSameLength(AssertArrayEquals.java:428)
	at org.junit.jupiter.api.AssertArrayEquals.assertArrayEquals(AssertArrayEquals.java:205)
	at org.junit.jupiter.api.AssertArrayEquals.assertArrayEquals(AssertArrayEquals.java:63)
	at org.junit.jupiter.api.AssertArrayEquals.assertArrayEquals(AssertArrayEquals.java:59)
	at org.junit.jupiter.api.Assertions.assertArrayEquals(Assertions.java:1226)
	at org.openhab.core.cache.lru.LRUMediaCacheEntryTest.loadTwoThreadsAtTheSameTimeFromTheSameSupplierTest(LRUMediaCacheEntryTest.java:199)
```

It was recently added in #3057 by @dalgwen.

dalgwen:
Closes #3507
Fix some rare case where the concurrent reads of a cache entry can fail.

Explanation: 
Lines 286 and 288 are executed nearly at the same time. "Nearly" was sufficient to cause error when setting maxToRead and inducing the thread to think that the wanted data cannot be read.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/cache/lru/LRUMediaCacheEntry.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/cache/lru/LRUMediaCacheEntry.java
@@ -296,7 +296,7 @@ public class LRUMediaCacheEntry<V> {
             }
         }
         // the cache file is now filled, get bytes from it.
-        long maxToRead = Math.min(currentSize, sizeToRead);
+        long maxToRead = Math.min(fileChannelLocal.size(), sizeToRead);
         ByteBuffer byteBufferFromChannelFile = ByteBuffer.allocate((int) maxToRead);
         int byteReadNumber = fileChannelLocal.read(byteBufferFromChannelFile, Integer.valueOf(start).longValue());
         logger.trace("Read {} bytes from the filechannel", byteReadNumber);


wborn:
Thanks for having a look! 

Woudn't there also concurrency issues when multiple InputStreams call `read(int start, int sizeToRead)` with random `start` and `sizeToRead` values?

There are also issues when `currentSize` is being updated by one thread and another thread already completed reading and set `completed` to `true`:

```java
    protected Long getTotalSize() {
        if (completed) { // we already know the total size of the sound
            return currentSize;
```

wborn:
I now see that threads won't go into the while loop once `completed` is true. But I still don't understand your explanation:

> Lines 286 and 288 are executed nearly at the same time. "Nearly" was sufficient to cause error when setting maxToRead and inducing the thread to think that the wanted data cannot be read.

Why would this matter if the statements are executed sequentially? Is there something asynchronous going on in the FileChannel?

dalgwen:
OK, I was not very clear, sorry.

Imagine 2 threads that want the same cache entry. Neither of them had read any data, and its a cache miss, so the cache entry is empty.

- Thread 1 starts and checks line 271 if the FileChannel has the data it wants. The FileChannel is length 0 and so it enters the synchronized block to get data from the input stream. It gets data, goes to line 286 and writes to the FileChannel the data he just get. It is now line 287, trace logging a message. Now that the FileChannel has the data it wants, it will go on and read it. But we pause it here because enter thread 2.

- Thread 2 also needs data. Line 271, it checks if the data it wants is available. It uses the FileChannel size for this purpose. The FileChannel has just been updated by Thread 1. Yeah, it seems to have the data it needs. So it doesn't enter the synchronized block and goes directlly line 299. It computes its maxToRead value (which is the amount of data he will get now). Before the correction, it uses currentSize to compute this value. BUT thread 1 has not updated the currentSize value, because it is still logging line 287 and the update of currentSize is line 288. So thread 2 has a wrong currentSize (still 0) and allocate a size 0 ByteBuffer. So the code after that will fill it with... no byte at all.

By the way, for maxToRead, using a min between sizeToRead and FileChannel.size is needed because we can imagine a file smaller than the sizeToRead asked by the caller. Allocating the exact data size needed is mandatory to fill it and returns exactly the right amount of data read from the FileChannel, without reallocating another buffer inbetween.

So, using currentSize to compute the size of the buffer we will fill was wrong, in some rare instances there was some concurrency issues. Using FileChannel.size fixes that.

wborn (approved):
Thanks for the detailed explanation! I now understand why it happens and did not think about a thread being suspended while it is running in the synchronized block. :upside_down_face: It's certainly worth it to have no delays. In the end the unit test also helped with finding a real concurrency issue. :slightly_smiling_face:

PR #3796, opened 2023-09-08
[rest] Add uptime to /systeminfo endpoint

### LINKED ISSUE #3656: [REST] Add uptime counter to systeminfo endpoint
Handling startup of openHAB from a HABApp perspective is hard.
OpenHAB might be stuck at `startLevel` 70 and never reach `startLevel` 100 if thing configs are broken.
Even if openHAB reaches `startLevel` 100 it's possible that the item state restore event (e.g. from MapDB) is queued up so that the item changes to the restored state significantly after openHAB reached  `startLevel` 100 (I've seen up to 15s).

It would be very helpful if there was an `uptime` field in `systemInfo` that would contain the openHAB uptime e.g. in secs.
That way HABApp could easily differentiate e.g. between a openHAB restart and a HABApp restart.

J-N-K:
Closes #3656


wborn (approved):
Thank you!

Issue #3795, opened 2023-09-07
addon info missing in REST response


mherwege:
Work has been done to include 'connection' and 'country' information in the addon info data structures. This could be useful to filter in a UI.
However, when testing the REST call for an addon, it looks like the connection and country info is lost somewhere and never filled in the REST response (at least for the addons in the openhab repository).

Try:
```
http://192.168.0.10:8080/rest/addons/binding-dsmr
```

The response body is:
```
{
  "uid": "binding-dsmr",
  "id": "dsmr",
  "label": "DSMR Binding",
  "version": "4.0.2",
  "compatible": true,
  "contentType": "application/vnd.openhab.feature;type=karaf",
  "link": "https://www.openhab.org/addons/bindings/dsmr/",
  "author": "openHAB",
  "verifiedAuthor": true,
  "installed": false,
  "type": "binding",
  "configDescriptionURI": "",
  "keywords": "",
  "countries": [],
  "connection": "",
  "properties": {},
  "loggerPackages": [
    "org.openhab.binding.dsmr"
  ]
}
```

`countries` and `connection` is empty. The content of the `addon.xml` file for the binding contains values for these fields: https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.dsmr/src/main/resources/OH-INF/addon/addon.xml

wborn:
It's currently only possible to add this data for installed add-ons because without it the data is not available, see: https://github.com/openhab/openhab-core/issues/1701

mherwege:
Unfortunately, it also did not show for installed addons.

wborn:
I see countries but no connection:

```
{
  "uid": "binding-dsmr",
  "id": "dsmr",
  "label": "DSMR Binding",
  "version": "4.1.0.SNAPSHOT",
  "compatible": true,
  "contentType": "application/vnd.openhab.feature;type=karaf",
  "link": "https://www.openhab.org/addons/bindings/dsmr/",
  "author": "openHAB",
  "verifiedAuthor": true,
  "installed": true,
  "type": "binding",
  "description": "This binding integrates Dutch, Belgium, Luxembourg and Austrian Smart Meters",
  "configDescriptionURI": "",
  "keywords": "",
  "countries": [
    "nl",
    "be",
    "lu",
    "at"
  ],
  "connection": "",
  "properties": {},
  "loggerPackages": [
    "org.openhab.binding.dsmr"
  ]
}
```

mherwege:
[...]

PR #3797, opened 2023-09-09
Fix countries and connection AddonInfo issues

wborn:
Fixes the following:

* connection and countries details missing for some AddonServices
* missing connection getter on AddonInfo
* countries lists has empty String when when countries info is missing

Related to #3795


J-N-K (approved):
LGTM, thanks.

PR #3798, opened 2023-09-09
[voice] Add dialog group and location

GiviMAD:
I would like to merge this minor improvement into the voice bundle, it introduces two new configurations into the dialog context, the dialogGroup which defaults to "default" and the locationItem with defaults to null.

The dialogGroup is used to discard spot events raised at same time in different dialog executions, so you can have multiple near speakers and not "awake" them at the same time. Also fixes a race error that happened when a KS service sent two consecutive spot events.

The locationItem, is meant to optionally contain the name of a "Location" group item, it is propagated to the StandarInterpreter (to its base class to be correct) so it can reduce label collisions based on this location context.


kaikreuzer (approved):
lgtm, thanks!

GiviMAD:
thank you!

PR #3803, opened 2023-09-18
Inject SCRIPT_CONTEXT into the execution context

jimtng:
Currently `ctx` is only accessible as a top-level local variable in JRuby. Because the helper library is loaded in a separate `eval` invocation, it doesn't have access to `ctx`. By injecting an uppercase attribute name, it is treated as a global constant which is accessible by the helper library.


jimtng:
Using https://github.com/openhab/openhab-addons/pull/15618 instead

PR #3804, opened 2023-09-20
[sitemap] Provide information about widget label source to clients

maniac103:
The label can be populated from a label specified on the respective sitemap widget, or (if no label was specified) from the label of the backing item. Allow clients to differentiate between both cases. This allows clients (BasicUI, mobile apps) to e.g. show the label for charts if they were specified on the sitemap widget and to hide it when it just was populated from the item label.

Related to openhab/openhab-webui#2065

### DIFF
--- a/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java
+++ b/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java
@@ -112,6 +112,7 @@ import org.slf4j.LoggerFactory;
  * @author Mark Herwege - new method getFormatPattern(widget), clean pattern
  * @author Laurent Garnier - Support added for multiple AND conditions in labelcolor/valuecolor/visibility
  * @author Laurent Garnier - new icon parameter based on conditional rules
+ * @author Danny Baumann - widget label source support
  */
 @NonNullByDefault
 @Component(immediate = true, configurationPid = "org.openhab.sitemap", //
@@ -141,6 +142,16 @@ public class ItemUIRegistryImpl implements ItemUIRegistry {
     private final Map<Widget, Widget> defaultWidgets = Collections.synchronizedMap(new WeakHashMap<>());
 
     private String groupMembersSorting = DEFAULT_SORTING;
+
+    private static class WidgetLabelWithSource {
+        public final String label;
+        public final WidgetLabelSource source;
+
+        public WidgetLabelWithSource(String l, WidgetLabelSource s) {
+            label = l;
+            source = s;
+        }
+    }
 
     @Activate
     public ItemUIRegistryImpl(@Reference ItemRegistry itemRegistry) {
@@ -325,7 +336,7 @@ public class ItemUIRegistryImpl implements ItemUIRegistry {
 
     @Override
     public @Nullable String getLabel(Widget w) {
-        String label = getLabelFromWidget(w);
+        String label = getLabelFromWidget(w).label;
 
         String itemName = w.getItem();
         if (itemName == null || itemName.isBlank()) {
@@ -468,6 +479,11 @@ public class ItemUIRegistryImpl implements ItemUIRegistry {
         return transform(label, considerTransform, labelMappedOption);
     }
 
+    @Override
+    public WidgetLabelSource getLabelSource(Widget w) {
+        return getLabelFromWidget(w).source;
+    }
+
     private QuantityType<?> convertStateToWidgetUnit(QuantityType<?> quantityState, Widget w) {
         Unit<?> widgetUnit = UnitUtils.parseUnit(getFormatPattern(w));
         if (widgetUnit != null && !widgetUnit.equals(quantityState.getUnit())) {
@@ -479,7 +495,7 @@ public class ItemUIRegistryImpl implements ItemUIRegistry {
 
     @Override
     public @Nullable String getFormatPattern(Widget w) {
-        String label = getLabelFromWidget(w);
+        String label = getLabelFromWidget(w).label;
         String pattern = getFormatPattern(label);
         String itemName = w.getItem();
         try {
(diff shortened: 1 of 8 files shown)


maniac103:
Open question: How can I actually test this in a development environment, IOW, how can I run the locally built core classes? There's documentation on how to run locally built bindings, but I haven't found the same (or generally, any kind of development documentation) for -core. Did I miss something?

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/internal/WidgetLabelSource.java
```diff
@@ -0,0 +1,34 @@
...
+
+    public static WidgetLabelSource forWidget(Widget w, String label) {
+        if (label == null || label.isEmpty()) {
+            return None;
+        } else if (w.getLabel() != null) {
+            return SitemapDefinition;
+        } else {
+            return ItemLabel;
+        }
+    }
```
lolodomo, line 33: In case the label is set to "[%s]" for example, either on sitemap element or on item, your method will return SitemapDefinition or ItemLabel, but not None. Is it really what you expect ?

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/internal/WidgetLabelSource.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.openhab.core.model.sitemap.sitemap.Widget;
+
+/**
+ * @author Danny Baumann - Initial contribution
+ */
+public enum WidgetLabelSource {
+    SitemapDefinition,
+    ItemLabel,
+    None;
```
lolodomo, line 23: The convention is rather to use uppercase.

lolodomo (changes requested):

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/internal/WidgetLabelSource.java
maniac103: Fixed.

bundles/org.openhab.core.io.rest.sitemap/src/main/java/org/openhab/core/io/rest/sitemap/internal/WidgetLabelSource.java
maniac103: I think so. After all, there *is* a label to work with, even if it almost only consists of the item state. I introduced `NONE` only because if there's no label at all, neither `ITEM` nor `SITEMAP` do really fit.

lolodomo:
You probably forgot one origin: the linked channel.

maniac103:
> You probably forgot one origin: the linked channel.

As far as I can tell from [the code](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java#L543-L561) there's only the widget, the item label and the item name which the label can be taken from. I've forgotten the latter, so I completely revamped the approach to also cover that one.
Or do you by 'linked channel' refer to [this](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java#L216-L219) piece of the logic? If so, IMO it's fine to group it under 'item label' as well, since it doesn't really matter where exactly the item label is fetched from.

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/items/ItemUIRegistry.java
```diff
@@ -51,10 +68,10 @@ public interface ItemUIRegistry extends ItemRegistry, ItemUIProvider {
      * current value of the item and padded by a "<span>" element.
      *
      * @param w the widget to retrieve the label for
-     * @return the label to use for the widget
+     * @return the label to use for the widget, together with its source
      */
     @Nullable
-    String getLabel(Widget w);
+    WidgetLabelWithSource getLabel(Widget w);
```
lolodomo, line 74: Maybe you can avoid introducing this serious breaking change ?
If not, Basic UI will certainly no more compile.

maniac103: Well, I *can* avoid it by adding a second method `getLabelSource`, but the basic question is do we prefer backwards compatibility or API design here? If the latter, I can also provide a patch to adjust BasicUI to the changed API.

maniac103: I have splitted retrieving the label source to a separate method as mentioned. This means the 'retrieve label' logic now needs to run twice for each sitemap widget or SSE event (once for actual label, a second time for source), but this shouldn't be too much of an issue.

maniac103:
The force-push to `b3c7734` just rebases onto `main`; its patch set should be identical to `d93cebb`.

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java
```diff
@@ -543,24 +559,31 @@ private QuantityType<?> convertStateToWidgetUnit(QuantityType<?> quantityState,
...
+                } else {
                     label = itemName;
+                    source = WidgetLabelSource.ITEM_NAME;
                 }
             }
         }
         // use an empty string, if no label could be found
-        return label != null ? label : "";
+        String finalLabel = label != null ? label : "";
+        return new WidgetLabelWithSource(finalLabel, finalLabel.isEmpty() ? WidgetLabelSource.NONE : source);
```
lolodomo, line 586: I am just annoyed with that because NONE in that case is not only the undefined source but also another source defined but being empty. Why not but in that case we need a clear comment in front of the enum for the NONE value.

lolodomo (changes requested):
Ok for me. Just one comment to be added to clearly explain what means NONE source. Because NONE covers two different things in your implementation.

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java
lolodomo: Or why not letting each UI checking if the label is empty ?
In that case, NONE will only have one meaning, related to the source and not to the value itself.

lolodomo: I think the source should not carry information about the label value itself.

maniac103: So you basically just want me to return `source` here instead of the ternary? That's OK, I guess.

lolodomo: Yes

lolodomo (approved):
Looks good to me.

maniac103:
Last commit just added some documentation for the enum values.

lolodomo:
@openhab/core-maintainers : is it possible to review and merge this PR ? I have already done a full pre-review.
@maniac103 is waiting for it for the Android app.

kaikreuzer (approved):
lgtm, thanks!

Issue #3793, opened 2023-09-01
Textually defined things not available after restart


schup011:
## Expected Behavior
As in OH3.x, during the start up procedure things defined by textual .things files should appear after a certain time in the log as being INITIALIZING and then becoming ONLINE. This can also be seen in the Things section in MainUI.

## Current Behavior
After a restart of OH4.x, things are not appearing neither in the log nor in MainUI. Some items and rules are working, some others not. Probably, this depends on the binding. Also a sitemap defined by a .sitemap file is not working and not defined.
After touching the respective files (.things and .sitemaps), everything becomes ONLINE and is working as expected.
After a restart with cleaned cache (openhab-CLI clean-cache), usually things files are loaded without a problem. Another restart without cleaning the cache reproduces the problem.

## Steps to Reproduce (for Bugs)

1. Provide a .things file thing definition.
2. Start openHAB. If the things are available, restart openhab (in my platform sudo systemctl restart openhab.service).

## Context
Especially in the situation of having migrated, frequent restarts of openhab are sometimes unavoidable. I am also migrating rules from Jython to Javascript at the moment.

## Your Environment

Fresh openhabian installation on Raspberry Pi 4 with 2GB RAM. 
Fresh SanDisk Extreme SD Card.
Standard installation, including Mosquitto &  Influx+Grafana.

schup011:
Maybe one hint: Recently, I added some things in the MainUI. It could be that having at least one "MainUI-defined" thing could prevent this bug to happen. At least it did not happen in my system for the last two restarts (sorry, no good statistics).

morph166955:
Upgraded to snapshot 3615 this morning, same issue here.  None of my things files load.  items and rules are fine, but it never loads things.  If I go and manually touch a file (add a blank line for example) it loads at that time (only that file).

AnTrii:
I'm at OpenHAB 4.0.2 Release / Java 17.0.8 / Debian 11 / kernel 5.15.108-1, same issue. Have defined a MainUI Modbus TCP Thing for test. After restarting OpenHAB I see only that one MainUI Thing. After restart, there is no .things mentioning whatsoever in OpenHAB logs. By the way, reinstalling OpenHAB make Things to work again until restart.

morph166955:
For reasons unknown, this bug resolved itself for me some time yesterday.

morph166955:
...And now a few hours later, it's back.

morph166955:
[...]

Issue #3784, opened 2023-08-27
JDBC DB flooded with items after some time of running


int5749:
Running openHAB 4.0.2 but issue already exist since migration to openHAB 3.x
Since this occurs occasionally i wasn't able to trace if there is any trigger.

As usual I do have various persistent service installed, MapDB, rrd4j, jdbc, influx

All are running fine in general but just JDBC DB is flooded with all items and DB is getting overloaded afterwards.

Hope this is right place to report this and happy to assist with guided traces, etc.

Regards,
Joerg

wborn:
> JDBC DB is flooded with all items and DB is getting overloaded afterwards.

What do you mean with that? Does it persist items that are not in your [persistence configuration](https://www.openhab.org/docs/configuration/persistence.html#persistence-configuration)? What persistence configuration do you use?

int5749:
> > JDBC DB is flooded with all items and DB is getting overloaded afterwards.
> 
> What do you mean with that? Does it persist items that are not in your [persistence configuration](https://www.openhab.org/docs/configuration/persistence.html#persistence-configuration)? What persistence configuration do you use?

Yes, there are just few items ~20-25 in my jdbc.persist file and is working fine for some time. Than looks like all ~1000 items are entered to my jdbc DB which causes DB to get issues when retrieving single items.

wborn:
This issue may have the same root cause as #3793. If it does not load and apply the strategies in the .persist file it will use the default persistence strategy which usually is to persist everything.

PR #3821, opened 2023-10-02
FolderObserver improvements

wborn:
* Mark extensions as ready in addModelsToRepo only if parser was available
* Store valid extensions using Set instead of List
* Use java.nio where possible
* Add more debug logging
* Cleanup and simplify code

---

This should help with fixing/debugging #3784 and #3793


J-N-K (approved):
LGTM. Especially the extended logging might help.

PR #3826, opened 2023-10-04
Upgrade CXF to 3.6.2

wborn:
Upgrades CXF from 3.6.1 to 3.6.2.

For release notes, see:

https://issues.apache.org/jira/secure/ReleaseNote.jspa?projectId=12310511&version=12353311

It has a fix to address OOM issues when using HTTP clients:

https://issues.apache.org/jira/browse/CXF-8885


kaikreuzer (approved):
Sounds important, thanks!

PR #3827, opened 2023-10-04
Remove unused javax.inject annotations

wborn:
AFAIK these annotations are not used when using declarative services.

I stumbled upon these annotations when feature validation failed while creating #3817. Xtext now uses Guice 7 which is using jakarta.inject imports.

Only si.uom:si-units:2.1 still imports javax.inject but it will also switch to jakarta.inject when 2.2 gets released.


kaikreuzer (approved):
Thanks!

PR #3831, opened 2023-10-07
Simplify code using Stream.toList

wborn:
Stream.toList was introduced in Java 16 and creates an unmodifiable List so it can be used to simplify code whenever the List is not expected to be modified.


bundles/org.openhab.core/src/main/java/org/openhab/core/net/NetUtil.java
```diff
@@ -531,19 +530,20 @@ private void pollAndNotifyNetworkInterfaceAddress() {
...
-                .filter(lastKnownInterfaceAddr -> !newInterfaceAddresses.contains(lastKnownInterfaceAddr))
-                .collect(Collectors.toList());
+                .filter(lastKnownInterfaceAddr -> !newInterfaceAddresses.contains(lastKnownInterfaceAddr)).toList();
 
         lastKnownInterfaceAddresses = newInterfaceAddresses;
 
         if (!added.isEmpty() || !removed.isEmpty()) {
-            LOGGER.debug("added {} network interfaces: {}", added.size(), Arrays.deepToString(added.toArray()));
-            LOGGER.debug("removed {} network interfaces: {}", removed.size(), Arrays.deepToString(removed.toArray()));
+            if (LOGGER.isDebugEnabled()) {
```
J-N-K, line 542: This can be added to the statement above

J-N-K (changes requested):
Thanks for the cleanup

bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/net/NetworkConfigOptionProvider.java
```diff
@@ -52,13 +51,11 @@ public class NetworkConfigOptionProvider implements ConfigOptionProvider {
             case PARAM_PRIMARY_ADDRESS:
                 Stream<CidrAddress> ipv4Addresses = NetUtil.getAllInterfaceAddresses().stream()
                         .filter(a -> a.getAddress() instanceof Inet4Address);
-                return ipv4Addresses.map(a -> new ParameterOption(a.toString(), a.toString()))
-                        .collect(Collectors.toList());
+                return ipv4Addresses.map(a -> new ParameterOption(a.toString(), a.toString())).toList();
```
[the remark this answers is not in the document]
wborn: I see there are a few more after briefly grepping the code but for most the value and label are different.

bundles/org.openhab.core/src/main/java/org/openhab/core/net/NetUtil.java
wborn: Which statement exactly? It always has to call `notifyListeners(added, removed);` below regardless of the log level. I only wanted to prevent the relatively expensive debug logging string operations.

J-N-K: I didn't see that.

J-N-K (approved):
Thanks again

PR #3834, opened 2023-10-07
Make DecimalType, QuantityType, PercentType accept lowercased exponent notation

### LINKED ISSUE #3833: DecimalType: Parsing floats with a lowercase "e" separating mantissa and exponent fails
Hi,
using stable 3.3 I get following entry in my log file when my PV inverter sends really small wattage values at night:

`Incoming payload '5.605193857299268e-45' not supported by type 'NumberValue'`

Seems like NumberValue/DecimalType/Number does not like the lowercase "e", separating mantissa and exponent.

Using an uppercase "E" (-> 5.605193857299268E-45) works.

While I could [work around this using a transformation](https://community.openhab.org/t/error-parsing-floats-with-negative-exponent/138746), I think others could fall in this trap too, because this is the notation that is used by C++ (and probably other languages too), see https://cplusplus.com/reference/ios/scientific/

Could this be changed to accept both notations? Thanks!

## Expected Behavior
Float value 5.605193857299268e-45 should be parsed correctly to the "close to zero" value.
Using an uppercase E works as intended.

## Current Behavior
Parsing fails and item is not updated. Log entry is:
`2022-09-08 20:51:01.336 [WARN ] [ab.binding.mqtt.generic.ChannelState] - Incoming payload '5.605193857299268e-45' not supported by type 'NumberValue'`

## Possible Solution
Accept "e" and "E".

In my case: Send 5.605193857299268e-45 to an mqtt number channel.

## Your Environment
* Version used: openHAB 3.3

jimtng:
Resolve #3833 and https://github.com/openhab/openhab-addons/issues/13371

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityType.java
@@ -70,10 +70,10 @@ public class QuantityType<T extends Quantity<T>> extends Number
     public static final QuantityType<Dimensionless> ZERO = new QuantityType<>(0, AbstractUnit.ONE);
     public static final QuantityType<Dimensionless> ONE = new QuantityType<>(1, AbstractUnit.ONE);
 
-    // Regular expression to split unit from value
-    // split on any blank character, even none (\\s*) which occurs after a digit (?<=\\d) and before
-    // a "unit" character ?=[a-zA-Z°µ%'] which itself must not be preceded by plus/minus digit (?![\\+\\-]?\\d).
-    // The later would be an exponent from the scalar value.
+    // Regular expression to split unit from value. Split on any blank character, even none (\\s*)
+    // which occurs after a digit (?<=\\d) and before a "unit" character ?=[a-zA-Z°µ\u03BC\u00B5%']
+    // which itself must not be preceded by plus/minus digit (?![\\+\\-]?\\d).
+    // The latter would be an exponent from the scalar value.
     private static final String UNIT_PATTERN = "(?<=\\d)\\s*(?=[a-zA-Z°µ\u03BC\u00B5%'](?![\\+\\-]?\\d))";
 
     static {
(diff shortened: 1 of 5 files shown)


wborn:
If this notation is to be supported I would also expect it to be supported by all the others types such as PercentType and QuantityType. There don't seem to be any unit tests for those that indicate it is supported.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityType.java
```diff
@@ -74,7 +74,7 @@ public class QuantityType<T extends Quantity<T>> extends Number
     // split on any blank character, even none (\\s*) which occurs after a digit (?<=\\d) and before
     // a "unit" character ?=[a-zA-Z°µ%'] which itself must not be preceded by plus/minus digit (?![\\+\\-]?\\d).
     // The later would be an exponent from the scalar value.
-    private static final String UNIT_PATTERN = "(?<=\\d)\\s*(?=[a-zA-Z°µ%'](?![\\+\\-]?\\d))";
+    private static final String UNIT_PATTERN = "(?<=\\d)\\s*(?=[a-zA-Z°µ\u03BC\u00B5%'](?![\\+\\-]?\\d))";
```
wborn, line 77: Can you also update the comment and add tests for this change?

jimtng: Updated. There is a test in UnitTests.java that also check for `\u03BC` but not for `\u00B5`. It caught the fact that the regex failed to split up the unit. 

However, I feel that this regex pattern is too restrictive. Although I don't have a specific case where it would fail, the fact that it failed because it didn't account for the special unicode character for "micro" here makes me wonder if there would be some other esoteric but valid unicode characters it would also fail on.

I tried this other pattern that's more broad, and it seems to pass the current tests too: 
`"(?<=\\d)\\s*(?=\\D(?![\\+\\-]?\\d))"`
However, I hesitate to use it here pending further testing. Is it worth pursuing, perhaps in a separate PR? Nobody has ever complained about it before, so I'd probably rather just let it be.

For the purpose of this PR, the current pattern in the PR should suffice and it at least sticks to the way it had been working before.

wborn: > Is it worth pursuing, perhaps in a separate PR? Nobody has ever complained about it before, so I'd probably rather just let it be.

I would also only do that when neccessary. :wink: 

But isn't there duplication in the pattern now? The µ seems to be the same character as `\u00B5`:

I see `true` when I run:

```java
System.out.println("µ".equals("\u00B5")); 
```

wborn (approved):
Thank you for the fix!

J-N-K:
This breaks the KNX tests, see https://github.com/openhab/openhab-addons/actions/runs/6461008060/job/17539806529?pr=15723

- `1 /s` can't be parsed anymore
- `1 Ω·m` can't be parsed anymore

jimtng:
This is because the regex is too restrictive as I mentioned above. Perhaps we do need to change it to a broader pattern after all.

PR #3836, opened 2023-10-08
Simplify code that creates List, Map and Set objects

wborn:
Simplifies the code by using List.of, List.copyOf etc. where possible which results in less code and imports.

---

Follow up to #3831

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/template/RuleTemplate.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/template/RuleTemplate.java
@@ -14,7 +14,6 @@ package org.openhab.core.automation.template;
 
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.UUID;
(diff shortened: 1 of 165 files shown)


bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/template/RuleTemplate.java
```diff
@@ -117,13 +116,13 @@ public RuleTemplate(@Nullable String UID, @Nullable String label, @Nullable Stri
...
-        this.actions = actions == null ? Collections.emptyList() : Collections.unmodifiableList(actions);
-        this.configDescriptions = configDescriptions == null ? Collections.emptyList()
+        this.triggers = triggers == null ? List.of() : Collections.unmodifiableList(triggers);
+        this.conditions = conditions == null ? List.of() : Collections.unmodifiableList(conditions);
+        this.actions = actions == null ? List.of() : Collections.unmodifiableList(actions);
+        this.configDescriptions = configDescriptions == null ? List.of()
                 : Collections.unmodifiableList(configDescriptions);
         this.visibility = visibility == null ? Visibility.VISIBLE : visibility;
-        this.tags = tags == null ? new HashSet<>() : new HashSet<>(tags);
+        this.tags = tags == null ? Set.of() : Collections.unmodifiableSet(tags);
```
J-N-K, line 125: Why do you use `Collections.unmodifiableSet` instead of `Set.copyOf`?

wborn: I made it similar to the other code that uses `unmodifiableList`. Using `unmodifiableList` has some benefits like better performance because it doesn't allocate a new list and it is "read through". But that can also give issues when the code providing the input parameters updates the collections.

J-N-K (approved):
LGTM

PR #3837, opened 2023-10-08
Fixes source filter not working for `GenericEventTrigger`

florian-h05:
While reviewing https://github.com/openhab/openhab-js/pull/300, I noticed that filtering by source `org.openhab.core.expire` was not possible.

This was because the specified source was checked against the event topic, which makes no sense at all given that different schemes. Source filtering is done inside the `apply` method.

In case you need some code to reproduce the described issue, just let me know.


J-N-K (approved):
Thanks!

kaikreuzer (approved):

PR #3838, opened 2023-10-08
[rest] Add caching for UoM info

florian-h05:
Adds caching as in #3729 for the UoM info endpoint added in #3611.
This is required for making Main UI dynamically load the supported dimensions from the REST API.


J-N-K (approved):
LGTM

PR #3840, opened 2023-10-09
Use 'uid' instead of 'UID' as method parameters

wborn:
Using 'UID' is confusing as method parameter because it can be mistaken for some kind of constant while reading code.

Some were already updated in #3836.


J-N-K (approved):
Thanks for the cleanup. Very useful

PR #3841, opened 2023-10-09
Fix QuantityType UNIT_PATTERN regex

jimtng:
Fix the problem caused by #3834 

- [Reported problem](https://github.com/openhab/openhab-core/pull/3834#issuecomment-1753706608)
- https://github.com/openhab/openhab-addons/pull/15723 tests passed with this


wborn (approved):
LGTM and the unit tests are nice too!

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityType.java
```diff
@@ -70,11 +70,11 @@ public class QuantityType<T extends Quantity<T>> extends Number
...
 
-    // Regular expression to split unit from value. Split on any blank character, even none (\\s*)
-    // which occurs after a digit (?<=\\d) and before a "unit" character ?=[a-zA-Z°µ\u03BC%']
-    // which itself must not be preceded by plus/minus digit (?![\\+\\-]?\\d).
+    // Regular expression to split unit from value. Split on any blank character, or
+    // between a digit (?<=\\d) and a non-digit character (?=\\D)
+    // which must not be preceded by plus/minus digit (?![\\+\\-]?\\d).
     // The latter would be an exponent from the scalar value.
-    private static final String UNIT_PATTERN = "(?<=\\d)\\s*(?=[a-zA-Z°µ\u03BC%'](?![\\+\\-]?\\d))";
+    private static final String UNIT_PATTERN = "\\s+|(?<=\\d)(?=\\D(?![\\+\\-]?\\d))";
```
J-N-K, line 77:
```suggestion
    private static final Pattern UNIT_PATTERN = Pattern.compile("\\s+|(?<=\\d)(?=\\D(?![\\+\\-]?\\d))");
```

Compiling a regex pattern is very expensive. We should do this once here and then use `UNIT_PATTERN.split(...)` in l. 120 instead of compiling it very time we create a `QuantityType`.

J-N-K (changes requested):
LGTM. I have just one suggestion.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityType.java
J-N-K: Interesting. I couldn't see @wborn's review. I'll add the changes myself.

PR #3842, opened 2023-10-10
Improve code in QuantityType

J-N-K:
See comment in #3841


wborn (approved):
Thanks for speeding up the parsing! :+1:

PR #3847, opened 2023-10-16
[modbus] reduce log level when modbus slave returns DEVICE_BUSY exception

nagisa:
This exception is meant to indicate that the request should be retried shortly, essentially, and at least some of the devices I own seem to be busy bees. Thus my logs receive significant spam of this warning.

Since the exception is transient and retrying it is the expected course of action, I think it makes sense to reduce the log level here slightly and only output an error when the retries get exhausted.

### DIFF
--- a/bundles/org.openhab.core.io.transport.modbus/src/main/java/org/openhab/core/io/transport/modbus/internal/ModbusManagerImpl.java
+++ b/bundles/org.openhab.core.io.transport.modbus/src/main/java/org/openhab/core/io/transport/modbus/internal/ModbusManagerImpl.java
@@ -285,6 +285,12 @@ public class ModbusManagerImpl implements ModbusManager {
     private static final String MODBUS_POLLER_THREAD_POOL_NAME = "modbusManagerPollerThreadPool";
 
     /**
+     * The slave exception code indicating that the device is currently busy processing another
+     * command.
+     */
+    private static final int MODBUS_EXCEPTION_SLAVE_DEVICE_BUSY = 6;
+
+    /**
      * Log message with WARN level if the task queues exceed this limit.
      *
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.io.transport.modbus/src/main/java/org/openhab/core/io/transport/modbus/internal/ModbusManagerImpl.java
```diff
@@ -662,7 +668,11 @@ private <R, C extends ModbusResultCallback, F extends ModbusFailureCallback<R>,
                 } catch (ModbusSlaveException e) {
                     lastError.set(new ModbusSlaveErrorResponseExceptionImpl(e));
                     // Slave returned explicit error response, no reason to re-establish new connection
-                    if (willRetry) {
+                    if (willRetry && e.getType() == MODBUS_EXCEPTION_SLAVE_DEVICE_BUSY) {
+                        logger.info(
```
J-N-K, line 672: When re-trying is an expected operation, you can reduce that to DEBUG.

J-N-K (changes requested):

lolodomo:
@nagisa : did you see the review comment?

bundles/org.openhab.core.io.transport.modbus/src/main/java/org/openhab/core/io/transport/modbus/internal/ModbusManagerImpl.java
nagisa, line 672:
```suggestion
                        logger.debug(
```

lolodomo:
@J-N-K : you can probably merge this PR now.

J-N-K (approved):
Thanks! And sorry for letting you wait so long. I didn't see my comments were addressed.

PR #3849, opened 2023-10-17
[core] Add conversion for HSB to RGBW and back

genesis81:
For KNX DPT251.600 to use all 4 colors.

With the new feature, the HSBType can converted into RGBW, and also back to HSB.

Due to the conversion, some accuracy is lost, but the result is approximately correct.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
@@ -279,7 +279,7 @@ public class ColorUtil {
      * Transform <a href="https://en.wikipedia.org/wiki/SRGB">sRGB</a> color format to
      * <a href="https://en.wikipedia.org/wiki/HSL_and_HSV">HSV</a> based {@link HSBType}.
      *
-     * @param rgbw array of four int with the RGBW values in the range 0 to 255.
+     * @param rgbw array of three or four int with the RGB(W) values in the range 0 to 255.
      * @return the corresponding {@link HSBType}.
      * @throws IllegalArgumentException when input array has wrong size or exceeds allowed value range.
      */
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
@@ -44,11 +44,11 @@ public class ColorUtil {
     private static final BigDecimal BIG_DECIMAL_120 = BigDecimal.valueOf(120);
     private static final BigDecimal BIG_DECIMAL_100 = BigDecimal.valueOf(100);
     private static final BigDecimal BIG_DECIMAL_60 = BigDecimal.valueOf(60);
+    private static final BigDecimal BIG_DECIMAL_50 = BigDecimal.valueOf(50);
     private static final BigDecimal BIG_DECIMAL_5 = BigDecimal.valueOf(5);
     private static final BigDecimal BIG_DECIMAL_3 = BigDecimal.valueOf(3);
     private static final BigDecimal BIG_DECIMAL_2 = BigDecimal.valueOf(2);
     private static final BigDecimal BIG_DECIMAL_2_POINT_55 = new BigDecimal("2.55");
-    private static final BigDecimal BIG_DECIMAL_0 = BigDecimal.valueOf(0);
     private static final BigDecimal BIG_DECIMAL_127_POINT_5 = BigDecimal.valueOf(127.5);
 
     public static final Gamut DEFAULT_GAMUT = new Gamut(new double[] { 0.9961, 0.0001 }, new double[] { 0, 0.9961 },
@@ -158,10 +158,10 @@ public class ColorUtil {
      * @return array of four {@link PercentType} with the RGBW values in the range 0 to 100 percent.
      */
     public static PercentType[] hsbToRgbwPercent(HSBType hsb) {
-        int[] rgb = hsbToRgb(hsb);
-        final BigDecimal Ri = new BigDecimal(rgb[0]);
-        final BigDecimal Gi = new BigDecimal(rgb[1]);
-        final BigDecimal Bi = new BigDecimal(rgb[2]);
+        PercentType[] rgb = hsbToRgbPercent(hsb);
+        final BigDecimal Ri = rgb[0].toBigDecimal();
+        final BigDecimal Gi = rgb[1].toBigDecimal();
+        final BigDecimal Bi = rgb[2].toBigDecimal();
         // Get the maximum between R, G, and B
         final BigDecimal tM = Ri.max(Gi.max(Bi));
 
@@ -171,7 +171,7 @@ public class ColorUtil {
         }
 
         // This section serves to figure out what the color with 100% hue is
-        final BigDecimal multiplier = BIG_DECIMAL_255.divide(tM, 0, RoundingMode.DOWN);
+        final BigDecimal multiplier = BIG_DECIMAL_100.divide(tM, 0, RoundingMode.DOWN);
         final BigDecimal hR = Ri.multiply(multiplier);
         final BigDecimal hG = Gi.multiply(multiplier);
         final BigDecimal hB = Bi.multiply(multiplier);
@@ -179,23 +179,14 @@ public class ColorUtil {
         // This calculates the Whiteness (not strictly speaking Luminance) of the color
         final BigDecimal M = hR.max(hG.max(hB));
         final BigDecimal m = hR.min(hG.min(hB));
-        final BigDecimal Luminance = ((M.add(m).divide(BIG_DECIMAL_2).subtract(BIG_DECIMAL_127_POINT_5))
-                .multiply(BIG_DECIMAL_255.divide(BIG_DECIMAL_127_POINT_5))).divide(multiplier);
-
-        // Calculate the output values
-        BigDecimal Ro = (Ri.subtract(Luminance).multiply(BIG_DECIMAL_100).divide(BIG_DECIMAL_255, 0,
-                RoundingMode.DOWN));
-        BigDecimal Go = (Gi.subtract(Luminance).multiply(BIG_DECIMAL_100).divide(BIG_DECIMAL_255, 0,
-                RoundingMode.DOWN));
-        BigDecimal Bo = (Bi.subtract(Luminance).multiply(BIG_DECIMAL_100).divide(BIG_DECIMAL_255, 0,
-                RoundingMode.DOWN));
-        BigDecimal Wo = Luminance.multiply(BIG_DECIMAL_100).divide(BIG_DECIMAL_255, 0, RoundingMode.DOWN);
+        final BigDecimal Luminance = ((M.add(m).divide(BIG_DECIMAL_2).subtract(BIG_DECIMAL_50))
+                .multiply(BIG_DECIMAL_100.divide(BIG_DECIMAL_50))).divide(multiplier);
 
         // check range
-        Ro = BIG_DECIMAL_100.min(Ro.max(BIG_DECIMAL_0));
-        Go = BIG_DECIMAL_100.min(Go.max(BIG_DECIMAL_0));
-        Bo = BIG_DECIMAL_100.min(Bo.max(BIG_DECIMAL_0));
-        Wo = BIG_DECIMAL_100.min(Wo.max(BIG_DECIMAL_0));
+        final BigDecimal Ro = Ri.subtract(Luminance).max(BigDecimal.ZERO);
+        final BigDecimal Go = Gi.subtract(Luminance).max(BigDecimal.ZERO);
+        final BigDecimal Bo = Bi.subtract(Luminance).max(BigDecimal.ZERO);
+        final BigDecimal Wo = Luminance.max(BigDecimal.ZERO);
 
         return new PercentType[] { new PercentType(Ro), new PercentType(Go), new PercentType(Bo), new PercentType(Wo) };
     }
@@ -223,7 +214,7 @@ public class ColorUtil {
         BigDecimal Bi = BigDecimal.valueOf(b).add(Luminance);
 
         // Get the maximum between R, G, and B
-        final BigDecimal tM = BIG_DECIMAL_255.min(Ri.max(Gi.max(Bi)).max(BIG_DECIMAL_0));
+        final BigDecimal tM = BIG_DECIMAL_255.min(Ri.max(Gi.max(Bi)).max(BigDecimal.ZERO));
 
         // If the maximum value is 0, immediately return pure black.
         if (tM.floatValue() == 0) {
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -46,7 +46,10 @@ public class ColorUtil {
     private static final BigDecimal BIG_DECIMAL_60 = BigDecimal.valueOf(60);
     private static final BigDecimal BIG_DECIMAL_5 = BigDecimal.valueOf(5);
     private static final BigDecimal BIG_DECIMAL_3 = BigDecimal.valueOf(3);
+    private static final BigDecimal BIG_DECIMAL_2 = BigDecimal.valueOf(2);
     private static final BigDecimal BIG_DECIMAL_2_POINT_55 = new BigDecimal("2.55");
+    private static final BigDecimal BIG_DECIMAL_0 = BigDecimal.valueOf(0);
```
J-N-K, line 51: There is already `BigDecimal.ZERO`, is there a reason to create a duplicate?

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -140,6 +143,102 @@ public static PercentType[] hsbToRgbPercent(HSBType hsb) {
...
+     * This function does not round the components. For conversion to integer values in the range 0 to 255 use
+     * {@link #hsbToRgb(HSBType)}.
+     *
+     * See also: {@link #hsbToRgb(HSBType)}, {@link #hsbTosRgb(HSBType)}, {@link #hsbToRgbPercent(HSBType)}
+     *
+     * @param hsb an {@link HSBType} value.
+     * @return array of four {@link PercentType} with the RGBW values in the range 0 to 100 percent.
+     */
+    public static PercentType[] hsbToRgbwPercent(HSBType hsb) {
+        int[] rgb = hsbToRgb(hsb);
```
J-N-K, line 161: If you convert to `BigDecimal` afterwards (and do other calculations), why don't you use `hsbToRgbPercent`? You can use `.toBigDecimal()` which avoids creation of a new `BigDecimal`.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -140,6 +143,102 @@ public static PercentType[] hsbToRgbPercent(HSBType hsb) {
...
+                RoundingMode.DOWN));
+        BigDecimal Bo = (Bi.subtract(Luminance).multiply(BIG_DECIMAL_100).divide(BIG_DECIMAL_255, 0,
+                RoundingMode.DOWN));
+        BigDecimal Wo = Luminance.multiply(BIG_DECIMAL_100).divide(BIG_DECIMAL_255, 0, RoundingMode.DOWN);
+
+        // check range
+        Ro = BIG_DECIMAL_100.min(Ro.max(BIG_DECIMAL_0));
+        Go = BIG_DECIMAL_100.min(Go.max(BIG_DECIMAL_0));
+        Bo = BIG_DECIMAL_100.min(Bo.max(BIG_DECIMAL_0));
+        Wo = BIG_DECIMAL_100.min(Wo.max(BIG_DECIMAL_0));
```
J-N-K, line 198: If you start with the `BigDecimal`s from `PercentType`, you should be fine without that multiplication/checking.

J-N-K (changes requested):
Thanks for your contribution. Looks like a useful addition. I have left some comments.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
genesis81: I will replace this with `BigDecimal.ZERO`

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
genesis81: I need the value in byte from 0 to 255. When i convert `PercentType.toBigDecimal()` i have only 0-100.

genesis81: I will check the calculation think, is a good idea.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
genesis81: Done

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -140,6 +142,93 @@ public static PercentType[] hsbToRgbPercent(HSBType hsb) {
...
+     * This function does not round the components. For conversion to integer values in the range 0 to 255 use
+     * {@link #hsbToRgb(HSBType)}.
+     *
+     * See also: {@link #hsbToRgb(HSBType)}, {@link #hsbTosRgb(HSBType)}, {@link #hsbToRgbPercent(HSBType)}
+     *
+     * @param rgb array of three int with the RGB values in the range 0 to 255.
+     * @return hsb an {@link HSBType} value.
+     *
+     */
+    public static HSBType rgbwTohsb(double r, double g, double b, double w) {
```
holgerfriedrich, line 208: We have used arrays of int or PercentType until now as input parameters. Existing method ```rgbToHsb(int[3])``` vs. new method ```rgbwToHsb(r,g,b,w)```. Do we need the precision of double, or would a ```rgbwToHsb(int[4])``` be sufficient (then IMHO there should be only one public interface which can handle arrays of size 3 and 4, maybe implemented by 2 private methods).

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -140,6 +142,93 @@ public static PercentType[] hsbToRgbPercent(HSBType hsb) {
...
     }
 
+    /**
+     * Transform <a href="https://en.wikipedia.org/wiki/HSL_and_HSV">HSV</a> based {@link HSBType} to RGBW.
+     *
+     * See <a href=
+     * "https://stackoverflow.com/questions/40312216/converting-rgb-to-rgbw">Converting RGB to RGBW</a>.
+     *
+     * This function does not round the components. For conversion to integer values in the range 0 to 255 use
+     * {@link #hsbToRgb(HSBType)}.
```
holgerfriedrich, line 152: should we add a ```int[] hsbToRgbw(HSBType)``` returning 4 values as well?

holgerfriedrich (changes requested):
Hi @genesis81, it is good to introduce RGBW as well.

I have added my thoughts regarding style, but also regarding the interfaces an how this fits to the existing methods for RGB.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
genesis81: I'm not sure what we have for side effects when we do this.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -216,13 +266,59 @@ public static double[] hsbToXY(HSBType hsb, Gamut gamut) {
      * @throws IllegalArgumentException when input array has wrong size or exceeds allowed value range.
      */
     public static HSBType rgbToHsb(int[] rgb) throws IllegalArgumentException {
+        if (rgb.length == 4) {
```
holgerfriedrich, line 269: Javadoc of this function to be extended, parameter rgb can have 3 or 4 values, etc.
Maybe add docs from private method for rgbw as well, as private methods are not included in Javadoc.

holgerfriedrich (changes requested):
thanks, just a few last comments....

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -216,13 +266,59 @@ public static double[] hsbToXY(HSBType hsb, Gamut gamut) {
...
+
+        final BigDecimal multiplier = BIG_DECIMAL_255.divide(tM, 0, RoundingMode.DOWN);
+
+        BigDecimal Ro = Ri.divide(multiplier).min(BIG_DECIMAL_255).max(BigDecimal.ZERO);
+        BigDecimal Go = Gi.divide(multiplier).min(BIG_DECIMAL_255).max(BigDecimal.ZERO);
+        BigDecimal Bo = Bi.divide(multiplier).min(BIG_DECIMAL_255).max(BigDecimal.ZERO);
+
+        return HSBType.fromRGB(Ro.intValue(), Go.intValue(), Bo.intValue());
+    }
+
```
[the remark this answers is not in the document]
holgerfriedrich: handled in #3867

genesis81:
> thanks, just a few last comments....

No problem, i add the points. Thanks for review.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -261,19 +279,19 @@ public static double[] hsbToXY(HSBType hsb, Gamut gamut) {
      * Transform <a href="https://en.wikipedia.org/wiki/SRGB">sRGB</a> color format to
      * <a href="https://en.wikipedia.org/wiki/HSL_and_HSV">HSV</a> based {@link HSBType}.
      *
-     * @param rgb array of three int with the RGB values in the range 0 to 255.
+     * @param rgbw array of four int with the RGBW values in the range 0 to 255.
```
holgerfriedrich, line 282: array of 3 or 4 int values representing either RGB or RGBW values in range....

genesis81: Done, thanks.

holgerfriedrich (approved):
LGTM, thanks.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -140,6 +160,54 @@ public static PercentType[] hsbToRgbPercent(HSBType hsb) {
...
+    public static PercentType[] hsbToRgbwPercent(HSBType hsb) {
+        PercentType[] rgb = hsbToRgbPercent(hsb);
+        final BigDecimal Ri = rgb[0].toBigDecimal();
+        final BigDecimal Gi = rgb[1].toBigDecimal();
+        final BigDecimal Bi = rgb[2].toBigDecimal();
+        // Get the maximum between R, G, and B
+        final BigDecimal tM = Ri.max(Gi.max(Bi));
+
+        // If the maximum value is 0, immediately return pure black.
+        if (tM.floatValue() == 0) {
```
J-N-K, line 186:
```suggestion
        if (BigDecimal.ZERO.equals(tM)) {
```

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -140,6 +160,54 @@ public static PercentType[] hsbToRgbPercent(HSBType hsb) {
...
+        // This section serves to figure out what the color with 100% hue is
+        final BigDecimal multiplier = BIG_DECIMAL_100.divide(tM, 0, RoundingMode.DOWN);
+        final BigDecimal hR = Ri.multiply(multiplier);
+        final BigDecimal hG = Gi.multiply(multiplier);
+        final BigDecimal hB = Bi.multiply(multiplier);
+
+        // This calculates the Whiteness (not strictly speaking Luminance) of the color
+        final BigDecimal M = hR.max(hG.max(hB));
+        final BigDecimal m = hR.min(hG.min(hB));
+        final BigDecimal Luminance = ((M.add(m).divide(BIG_DECIMAL_2).subtract(BIG_DECIMAL_50))
```
J-N-K, line 199:
```suggestion
        final BigDecimal whitenessMax = hR.max(hG.max(hB));
        final BigDecimal whitenessMin = hR.min(hG.min(hB));
        final BigDecimal luminance = ((M.add(m).divide(BIG_DECIMAL_2).subtract(BIG_DECIMAL_50))
```

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -140,6 +160,54 @@ public static PercentType[] hsbToRgbPercent(HSBType hsb) {
...
+                .multiply(BIG_DECIMAL_100.divide(BIG_DECIMAL_50))).divide(multiplier);
+
+        // check range
+        final BigDecimal Ro = Ri.subtract(Luminance).max(BigDecimal.ZERO);
+        final BigDecimal Go = Gi.subtract(Luminance).max(BigDecimal.ZERO);
+        final BigDecimal Bo = Bi.subtract(Luminance).max(BigDecimal.ZERO);
+        final BigDecimal Wo = Luminance.max(BigDecimal.ZERO);
```
J-N-K, line 206:
```suggestion
        BigDecimal outRed = Ri.subtract(Luminance).max(BigDecimal.ZERO);
        BigDecimal outGreen = Gi.subtract(Luminance).max(BigDecimal.ZERO);
        BigDecimal outBlue = Bi.subtract(Luminance).max(BigDecimal.ZERO);
        BigDecimal outWhite = Luminance.max(BigDecimal.ZERO);
```

no need to add final to local variables,

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -140,6 +160,54 @@ public static PercentType[] hsbToRgbPercent(HSBType hsb) {
...
+     * @param hsb an {@link HSBType} value.
+     * @return array of four {@link PercentType} with the RGBW values in the range 0 to 100 percent.
+     */
+    public static PercentType[] hsbToRgbwPercent(HSBType hsb) {
+        PercentType[] rgb = hsbToRgbPercent(hsb);
+        final BigDecimal Ri = rgb[0].toBigDecimal();
+        final BigDecimal Gi = rgb[1].toBigDecimal();
+        final BigDecimal Bi = rgb[2].toBigDecimal();
+        // Get the maximum between R, G, and B
+        final BigDecimal tM = Ri.max(Gi.max(Bi));
```
J-N-K, line 183:
```suggestion
        final BigDecimal inRed = rgb[0].toBigDecimal();
        final BigDecimal inGreen = rgb[1].toBigDecimal();
        final BigDecimal inBlue = rgb[2].toBigDecimal();
        // Get the maximum between R, G, and B
        final BigDecimal maxColor = Ri.max(Gi.max(Bi));
```

J-N-K (changes requested):
I only left some comments regarding the variable naming. Please also adjust that in the other direction. Once that is done, we are good to go. Please make sure that all names start with a lower case letter.

J-N-K (commented):
`hsbToRgbwPercent` looks good now. As I said in my last review: please also apply that to `rgbwToHsb`.

J-N-K (approved):
Thanks!

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -211,23 +280,69 @@ public static double[] hsbToXY(HSBType hsb, Gamut gamut) {
...
             throw new IllegalArgumentException("RGB array only allows values between 0 and 255");
         }
-        return rgbToHsb(new PercentType[] { convertByteToColorPercent(rgb[0]), convertByteToColorPercent(rgb[1]),
-                convertByteToColorPercent(rgb[2]) });
+        return rgbToHsb(new PercentType[] { convertByteToColorPercent(rgbw[0]), convertByteToColorPercent(rgbw[1]),
+                convertByteToColorPercent(rgbw[2]) });
+    }
+
+    /**
+     * Transform <a href="https://en.wikipedia.org/wiki/HSL_and_HSV">HSV</a> based {@link HSBType} to RGBW.
```
jlaur, line 299: I just found this PR and was interested because I think this method could be used instead of doing two passes here:
https://github.com/openhab/openhab-addons/blob/a13fd80bfe5cf0b6c55e13756f42abc27b55796b/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewRepeaterHandler.java#L264-L265

Anyway, commenting because this description seems to be copied from the method for the other direction. This one is RGBW to `HSBType`.

holgerfriedrich: thanks, included in #3882

Issue #3846, opened 2023-10-16
Sitemaps defined in UI do not load anymore in 4.1M2


mherwege:
I updated my install to 4.1M2. I have all my sitemaps defined in the UI. They are still confirmed to be valid in the UI, but don't load anymore. Also saving them again does not solve the issue.
Putting them in text files makes them work again.

Below is the log.

@lolodomo I suspect it is caused by https://github.com/openhab/openhab-core/pull/3819, converting an single condition to a collection of conditions. Therefore the logic expects a different json format. The UI sitemap creation support will need to be enhanced with the new functionality, and sitemaps will have to be resaved. Sitemaps in the saved json db do not have the correct syntax anymore.

```
2023-10-16 18:01:03.915 [ERROR] [omponents.UIComponentSitemapProvider] - Cannot build sitemap thuis
java.lang.ClassCastException: class org.openhab.core.model.sitemap.sitemap.impl.ConditionImpl cannot be cast to class java.util.Collection (org.openhab.core.model.sitemap.sitemap.impl.ConditionImpl is in unnamed module of loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @c7f527; java.util.Collection is in module java.base of loader 'bootstrap')
	at org.openhab.core.model.sitemap.sitemap.impl.VisibilityRuleImpl.eSet(VisibilityRuleImpl.java:130) ~[?:?]
	at org.openhab.core.ui.internal.components.UIComponentSitemapProvider.addWidgetVisibility(UIComponentSitemapProvider.java:387) ~[?:?]
	at org.openhab.core.ui.internal.components.UIComponentSitemapProvider.buildWidget(UIComponentSitemapProvider.java:296) ~[?:?]
	at org.openhab.core.ui.internal.components.UIComponentSitemapProvider.buildWidget(UIComponentSitemapProvider.java:288) ~[?:?]
	at org.openhab.core.ui.internal.components.UIComponentSitemapProvider.buildSitemap(UIComponentSitemapProvider.java:165) ~[?:?]
	at org.openhab.core.ui.internal.components.UIComponentSitemapProvider.getSitemapNames(UIComponentSitemapProvider.java:124) ~[?:?]
	at org.openhab.core.io.rest.sitemap.internal.SitemapResource.getSitemapBeans(SitemapResource.java:429) ~[?:?]
	at org.openhab.core.io.rest.sitemap.internal.SitemapResource.getSitemaps(SitemapResource.java:244) ~[?:?]
[...]

PR #3850, opened 2023-10-18
Fix UI defined sitemaps

mherwege:
Close https://github.com/openhab/openhab-core/issues/3846.

Recent enhancements to the sitemap syntax completely broke UI defined sitemaps.

This PR makes sitemap definitions in the UI work again, and is a quick fix.

However, most of the new syntax will still not work in the UI. This requires a larger refactoring both in core and webui.

EDIT:
With the second commit, AND conditions in visibility, color and icon rules are now allowed.
The JSON config format has not been changed (except for the addition of iconrules) to keep full backward compatibility and avoid major refactoring on the UI side. Logic has been created to correctly interpret the condition text strings.

This logic has one inherent problem. Condition strings will be split on ' AND ' strings in their composing parts. If any of the comparisons contains an ' AND ' string (in capitals, including the blanks before and after), provider will give the wrong result.

I will follow up with a UI PR to introduce icon rules and allow AND conditions.

@lolodomo Could you please have a look.

### DIFF
--- a/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
+++ b/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
@@ -13,6 +13,7 @@
 package org.openhab.core.ui.internal.components;
 
 import java.math.BigDecimal;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
@@ -32,6 +33,7 @@ import org.openhab.core.model.core.ModelRepositoryChangeListener;
 import org.openhab.core.model.sitemap.SitemapProvider;
 import org.openhab.core.model.sitemap.sitemap.Button;
 import org.openhab.core.model.sitemap.sitemap.ColorArray;
+import org.openhab.core.model.sitemap.sitemap.IconRule;
 import org.openhab.core.model.sitemap.sitemap.LinkableWidget;
 import org.openhab.core.model.sitemap.sitemap.Mapping;
 import org.openhab.core.model.sitemap.sitemap.Sitemap;
@@ -48,6 +50,7 @@ import org.openhab.core.model.sitemap.sitemap.impl.ConditionImpl;
 import org.openhab.core.model.sitemap.sitemap.impl.DefaultImpl;
 import org.openhab.core.model.sitemap.sitemap.impl.FrameImpl;
 import org.openhab.core.model.sitemap.sitemap.impl.GroupImpl;
+import org.openhab.core.model.sitemap.sitemap.impl.IconRuleImpl;
 import org.openhab.core.model.sitemap.sitemap.impl.ImageImpl;
 import org.openhab.core.model.sitemap.sitemap.impl.InputImpl;
 import org.openhab.core.model.sitemap.sitemap.impl.MappingImpl;
@@ -94,10 +97,8 @@ public class UIComponentSitemapProvider implements SitemapProvider, RegistryChan
     private static final String SITEMAP_PREFIX = "uicomponents_";
     private static final String SITEMAP_SUFFIX = ".sitemap";
 
-    private static final Pattern VISIBILITY_PATTERN = Pattern
-            .compile("(?<item>[A-Za-z]\\w*)\\s*(?<condition>==|!=|<=|>=|<|>)\\s*(?<sign>\\+|-)?(?<state>.+)");
-    private static final Pattern COLOR_PATTERN = Pattern.compile(
-            "((?<item>[A-Za-z]\\w*)?\\s*((?<condition>==|!=|<=|>=|<|>)\\s*(?<sign>\\+|-)?(?<state>[^=]*[^= ]+))?\\s*=)?\\s*(?<arg>\\S+)");
+    private static final Pattern CONDITION_PATTERN = Pattern
+            .compile("(?<item>[A-Za-z]\\w*)?\\s*(?<condition>==|!=|<=|>=|<|>)?\\s*(?<sign>\\+|-)?(?<state>.+)");
 
     private Map<String, Sitemap> sitemaps = new HashMap<>();
     private @Nullable UIComponentRegistryFactory componentRegistryFactory;
@@ -160,7 +161,11 @@ public class UIComponentSitemapProvider implements SitemapProvider, RegistryChan
 
         SitemapImpl sitemap = (SitemapImpl) SitemapFactory.eINSTANCE.createSitemap();
         sitemap.setName(SITEMAP_PREFIX + rootComponent.getUID());
-        sitemap.setLabel(rootComponent.getConfig().get("label").toString());
+        Object label = rootComponent.getConfig().get("label");
+        if (label == null) {
+            return sitemap;
+        }
+        sitemap.setLabel(label.toString());
 
         if (rootComponent.getSlots() != null && rootComponent.getSlots().containsKey("widgets")) {
             for (UIComponent component : rootComponent.getSlot("widgets")) {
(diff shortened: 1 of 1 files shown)


lolodomo (commented):
Partial review

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
```diff
@@ -331,10 +335,28 @@ private void setWidgetPropertyFromComponentConfig(Widget widget, @Nullable UICom
         }
     }
 
+    private void setWidgetIconPropertyFromComponentConfig(Widget widget, @Nullable UIComponent component) {
+        if (component == null || component.getConfig() == null) {
+            return;
+        }
+        Object staticIcon = component.getConfig().get("staticIcon");
+        if (staticIcon != null && Boolean.valueOf(ConfigUtil.normalizeType(staticIcon).toString())) {
```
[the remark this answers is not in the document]
mherwege: It is not in the JSON db and JSON config. It is only in the grammar.

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
```diff
@@ -160,7 +161,11 @@ protected Sitemap buildSitemap(RootUIComponent rootComponent) {
 
         SitemapImpl sitemap = (SitemapImpl) SitemapFactory.eINSTANCE.createSitemap();
         sitemap.setName(SITEMAP_PREFIX + rootComponent.getUID());
-        sitemap.setLabel(rootComponent.getConfig().get("label").toString());
+        Object label = rootComponent.getConfig().get("label");
+        if (label == null) {
+            return sitemap;
```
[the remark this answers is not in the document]
mherwege: Label is a required parameter. It being null should not happen as it is verified in the UI when constructing the sitemap. But this avoids a null waring in Eclipse.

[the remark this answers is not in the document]
lolodomo: Label is absolutely not mandatory but rather optional.
https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/Sitemap.xtext#L10

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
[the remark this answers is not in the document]
lolodomo: Ok

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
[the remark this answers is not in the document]
mherwege: Sorry, my mistake. Previous code would have failed also if no label existed.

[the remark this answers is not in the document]
mherwege: Checking UI code, UI configured sitemaps always imposed a label. I will remove the requirement in the UI.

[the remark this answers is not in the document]
mherwege: Checking again, the UI heavily relies on the label being available. The name of the sitemap will always be something like 'uicomponents_...', so is not very good anyway. I propose to keep enforcing a label for UI provided sitemaps.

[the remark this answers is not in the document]
lolodomo: Ok, makes sense to me.

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
```diff
@@ -412,50 +427,78 @@ private void addIconColor(EList<ColorArray> iconColor, UIComponent component) {
...
+    private String getRuleArgument(String rule) {
+        int argIndex = rule.lastIndexOf("=") + 1;
+        return rule.substring(argIndex).trim();
+    }
+
+    private List<String> getRuleConditions(String rule, @Nullable String argument) {
+        String conditions = rule;
+        if (argument != null) {
+            conditions = rule.substring(0, rule.lastIndexOf(argument)).trim();
+            if (conditions.endsWith("=")) {
```
lolodomo, line 494: And if there is one or several spaces between = and the "argument" ?

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
```diff
@@ -412,50 +427,78 @@ private void addIconColor(EList<ColorArray> iconColor, UIComponent component) {
...
+                conditions = conditions.substring(0, conditions.length() - 1);
+            }
+        }
+        List<String> conditionsList = List.of(conditions.split(" AND "));
```
lolodomo, line 498: Does UI systematically adds space around the AND ?

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
```diff
@@ -412,50 +427,78 @@ private void addIconColor(EList<ColorArray> iconColor, UIComponent component) {
...
+                        String argument = getRuleArgument(sourceColor.toString());
+                        List<String> conditionsString = getRuleConditions(sourceColor.toString(), argument);
                         ColorArrayImpl colorArray = (ColorArrayImpl) SitemapFactory.eINSTANCE.createColorArray();
-                        ConditionImpl condition = (ConditionImpl) SitemapFactory.eINSTANCE.createCondition();
-                        condition.setItem(matcher.group("item"));
-                        condition.setCondition(matcher.group("condition"));
-                        condition.setSign(matcher.group("sign"));
-                        condition.setState(matcher.group("state"));
-                        colorArray.eSet(SitemapPackage.COLOR_ARRAY__CONDITIONS, condition);
+                        colorArray.eSet(SitemapPackage.COLOR_ARRAY__ARG, argument);
```
lolodomo, line 437: Using `colorArray.setArg` is probably proper.

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
```diff
@@ -412,50 +427,78 @@ private void addIconColor(EList<ColorArray> iconColor, UIComponent component) {
...
-                        condition.setItem(matcher.group("item"));
-                        condition.setCondition(matcher.group("condition"));
-                        condition.setSign(matcher.group("sign"));
-                        condition.setState(matcher.group("state"));
-                        iconRule.eSet(SitemapPackage.ICON_RULE__CONDITIONS, condition);
-                        iconRule.setArg(matcher.group("arg"));
-                        iconDef.add(iconRule);
-                    } else {
-                        logger.warn("Syntax error in icon rule '{}' for widget {}", sourceIcon, component.getType());
+                        iconRule.eSet(SitemapPackage.ICON_RULE__ARG, argument);
```
lolodomo, line 456: Using `colorArray.setArg` is probably proper.

lolodomo (changes requested):
End of review

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
mherwege: Yes, it does.

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
mherwege: The full rule will be something like:
```
Item>=5 = red
```
The passed in argument would be `red`.
If there is an argument (!= null), it is split off and conditions should end with = (after trim). If there is a second = before, it should actually be part of the state string (5 in the example). So I think this is right.

lolodomo: I did not take attention that the previous line ended with a call to trim(). So, no problem, you can ignore this comment.

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
mherwege: Done.

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
mherwege: Done.

lolodomo (approved):
LGTM, thank you a lot

mherwege:
@lolodomo Thanks for the quick review.

lolodomo:
@openhab/core-maintainers : a merge would be appreciated.

kaikreuzer (approved):
Thanks @mherwege!

PR #3853, opened 2023-10-19
[voice] Support item descriptions in built-in interpreter.

GiviMAD:
Hello OpenHAB's maintainers, I would like to merge the following changes in the voice bundle.

It enables support in the standard interpreter to use the item StateDescription/CommandDescription labels to update them in the current available languages.

As an example: 
If you have a dimmer item labeled as `Brightness` with command description `50=normal`, the command `set brightness to normal` will be handled and send a PercentageType command with value 50 to the item.

French and German command needs to be reviewed, I made them by coping the color command or using a translator.  

It introduces a minor behavior change, if the language (a locale instance) is provided only the rules for that language are loaded. I think it's the correct thing to do in order to avoid register rules you are not going to use, but I can revert that part if you think it can have any side effect.

Tested in the 4.1.0.M2, seems to work great.

### DIFF
--- a/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/text/StandardInterpreter.java
+++ b/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/text/StandardInterpreter.java
@@ -12,6 +12,7 @@
  */
 package org.openhab.core.voice.internal.text;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


GiviMAD:
I have added basic support for custom rules using the item metadata in the last commit.

Let me know if it's better to move that to another PR.

If it's ok to merge these changes I would like to remove the actiontemplateinterpreter from the addons repo afterwards.

Best regards.

kaikreuzer (approved):
Thanks, quite some code changes, which isn't easy to review, but scanning through it, it lgtm, so let's merge!

lolodomo:
> I would like to remove the actiontemplateinterpreter from the addons repo afterwards.

So we have to suppress actiontemplatehli now?

GiviMAD:
Oh, I was not notified about these messages, thank you for accepting the PR!

I been using it for some weeks and it worked great. The only thing I doubt about is whether was a good idea or not to register the rules for every item with state/command description or if it would be a better idea to require a flag provided through the introduced metadata key 'voice-system', because for me it works fast but maybe in a system with a great amount of items it can affect the performance. But it's something easy to change if needed.

I have another minor idea for allowing creating itemRules using an item as template, I need to give it a try, then I can try to document the full functionality somewhere if you want.

> So we have to suppress actiontemplatehli now?

@lolodomo, yes, I would like to do so, I can create the PR if you want me to do it.

lolodomo:
> yes, I would like to do so, I can create the PR if you want me to do it.

Yes please

PR #3858, opened 2023-10-29
Fix thing upgrades for bridges

J-N-K:
Bridges could not be updated because the updated bridge was a `ThingImpl` instead of a `BridgeImpl`. Also the copy-creator in the `BridgeBuilder` was missing. This should also be cherry-picked to 4.0.x.


wborn (approved):
Thanks for the fix! :+1:

jlaur:
@J-N-K - can you also add the "patch" label so that it will be included in the release notes? (or @kaikreuzer if you are about to generate them)

PR #3859, opened 2023-11-01
Add log web socket

J-N-K:
This adds a web socket connection for receiving logs. Similar to #3345 but for web sockets instead of SSE. 

@ghys FYI. It works fine with the `LogReaderService`and the `LogListener`.


spacemanspiff2007:
Would it be also possible to add a short example to the docs how to use this?

J-N-K:
Sure. Short description: open a web socket connection to `/ws/logs` and receive all logs. Sending an array of regex will enable a filter: only lines where the logging bundle matches at least one regex will be sent over the connection (e.g. `["org\.openhab\.core\..*"]` would only allow core messages to pass). An empty array again enables "full logging".

spacemanspiff2007:
Perfect. The regex matches the whole line? So that's what I would want to use to filter e.g. for the levels INFO, WARNING and ERROR. Or is there an option to set the log level?

spacemanspiff2007:
Just to clarify - I am looking for a way to subscribe to all messages of level INFO and above (e.g. WARNING, ERROR, CRITICAL and whatever other levels there may be).

kaikreuzer (approved):
Thanks!

spacemanspiff2007:
What if I want to filter by log level and not by regex.
How can I set the filter to get everything of log level including INFO and above?
Or is there an overview of the availalbe log levels? I'm guessing the log levels can be added dynamically - are there levels which are specific to openHAB?

spacemanspiff2007:
@J-N-K :
Is it possible to filter by log level? E.g. if I want only messages with level INFO and above?

J-N-K:
No, this passes on what goes to the logger. I also think it would be confusing, because if you add `TRACE` but no logger is configured for TRACE logging, you would not get any other information.

spacemanspiff2007:
> No, this passes on what goes to the logger. I also think it would be confusing, because if you add `TRACE` but no logger is configured for TRACE logging, you would not get any other information.

You misunderstand. If I configure `TRACE` I'd expect to get `TRACE` and above, e.g. all messages with`TRACE`, `DEBUG`, `INFO`, `WARNING` and `ERROR`.
If I configure `INFO` I'd expect to get `INFO`, `WARNING` and `ERROR`.

Or do you have another idea how I can get all `WARNING` and `ERROR` messages without client side filtering.

PR #3867, opened 2023-11-09
 Fix javadoc warnings in directory org.openhab.core

holgerfriedrich:
Other directories still to be handled in other PR.

PR #3870, opened 2023-11-10
Fix javadoc warnings

holgerfriedrich:
This is a follow-up to #3867.

After applying both patches there are still some warnings left.
Not sure how to deal with the following ones:
```
ThingTypeRegistry.java:137: warning: Tag @link: reference not found: ThingType#getChannelType(ChannelUID)
ThingTypeRegistry.java:153: warning: Tag @link: reference not found: ThingType#getChannelType(ChannelUID)
```
References to RuleStatus.DISABLED which were missing with the first commit now are replaced by RuleStatusDetail.DISABLED.
The references to XtextFormattingTokenSerializer which were missing with the first commit now are replaced by org.eclipse.xtext.xtext.XtextFormatter.
The ref to getChannelType should probably be fixed, but I am not sure which function to reference.

There are some other correct refs failing, likely due the way the docs are built in the subdirectories.
Probably we cannot do much for those and the external references:
```
ScriptBusEvent.java:31: warning: Tag @link: reference not found: org.openhab.core.model.script.actions.BusEvent
ScriptThingActions.java:22: warning: Tag @link: reference not found: org.openhab.core.model.script.internal.engine.action.ThingActionService
Configuration.java:29: warning: Tag @link: reference not found: org.openhab.core.thing.Thing
WebSocketAdapter.java:38: warning: Tag @link: reference not found: org.eclipse.jetty.websocket.api.annotations
```

Maybe we should enforce no javadoc warnings in the CI (but then we would need some special handling if we cannot resolve the ones above).

### DIFF
--- a/bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/AbstractRemoteAddonService.java
+++ b/bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/AbstractRemoteAddonService.java
@@ -136,7 +136,7 @@ public abstract class AbstractRemoteAddonService implements AddonService {
     }
 
     /**
-     * Add a {@link MarketplaceAddonHandler) to this service
+     * Add a {@link MarketplaceAddonHandler} to this service
      *
      * This needs to be implemented by the addon-services because the handlers are references to OSGi services and
      * the @Reference annotation is not inherited.
@@ -147,7 +147,7 @@ public abstract class AbstractRemoteAddonService implements AddonService {
     protected abstract void addAddonHandler(MarketplaceAddonHandler handler);
 
     /**
-     * Remove a {@link MarketplaceAddonHandler) from this service
+     * Remove a {@link MarketplaceAddonHandler} from this service
      *
      * This needs to be implemented by the addon-services because the handlers are references to OSGi services and
      * unbind methods can't be inherited.
--- a/bundles/org.openhab.core.addon/src/main/java/org/openhab/core/addon/AddonEvent.java
+++ b/bundles/org.openhab.core.addon/src/main/java/org/openhab/core/addon/AddonEvent.java
@@ -17,7 +17,8 @@ import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.events.AbstractEvent;
 
 /**
- * This is an {@link Event} that is sent on add-on operations, such as installing and uninstalling.
+ * This is an {@link org.openhab.core.events.Event} that is sent on add-on operations, such as installing and
+ * uninstalling.
  *
  * @author Kai Kreuzer - Initial contribution
  */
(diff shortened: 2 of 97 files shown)


wborn:
> The references to XtextFormattingTokenSerializer are also no longer working, even searching the web for this does not give any results.

I think it lives on as the `XtextFormatter` but I created https://github.com/eclipse/xtext/pull/2840 to get this confirmed. :slightly_smiling_face:

wborn:
> reference not found: RuleStatus#DISABLED

This seems to have been moved to `org.openhab.core.automation.RuleStatusDetail` in https://github.com/eclipse-archived/smarthome/pull/4468.

holgerfriedrich:
Thanks @wborn for you help on this.
Now it is down to this one from ThingTypeRegistry:
ThingType#getChannelType

```
    /**
     * Returns the channel type for a given channel.
     *
     * <p>
     * <strong>Attention:</strong> If you iterate over multiple channels to find the according channel types, please
     * fetch the thing type first using
     * {@link ThingTypeRegistry#getThingType(ThingTypeUID)} and use
     * {@link ThingType#getChannelType(ChannelUID)} afterwards.
     *
     * @param channel channel
     * @return channel type or null if no channel type was found
     */
    public @Nullable ChannelType getChannelType(Channel channel) {
```

wborn:
> Probably we cannot do much for those and the external references

Maybe just use `{@code ...}`  instead of `{@link ...}` ?

holgerfriedrich:
Now it should be free of warnings (together with #3867).

WDYT @wborn, should we start a discussion about adding javadoc to CI build and enforcing warning free commits? If yes, I would start this in a separate issue.

wborn (approved):
Thanks for the cleanup!

wborn:
> CI build and enforcing warning free commits

How will this impact build times? They are already take a lot of time.

holgerfriedrich:
You are right, let's keep the response time low for the normal CI builds.

Maybe we could add a new Jenkins job or add a javadoc step to the build job we use for the build status badge:
image
In such a job, all warnings/errors could be displayed like in the SAT statistics.
And it could provide an easy way to access the current javadocs.

wborn:
There's already a JavaDoc build which can be reconfigured and for which we can add a build badge:

`[JavaDoc Build Status](https://ci.openhab.org/view/Documentation/job/openHAB-JavaDoc/)`

[JavaDoc Build Status](https://ci.openhab.org/view/Documentation/job/openHAB-JavaDoc/)

holgerfriedrich:
It is always passing with status success, see the older builds.

We need to set the build status to unstable or failed. I do not know how the job is configured, for pipelines the exit status can be used.
We have to check for "warning:" in the log and set build status to unstable. Same for "error:" and status fail.

wborn:
With #3886 the build will fail on warnings/erros.

PR #3874, opened 2023-11-15
Use a scheduled thread pool in JsonStorage + Bugfixes

joerg1985:
# Description

The intention of this PR was to have less idling threads created by the Timer instances.
A `ScheduledExecutorService` is now used instead to have only some worker threads.

This PR will also fix two issues of this class:
1. The commitTimerTask field was never set to another value than null.
    See line 411 in the old version, it should be `this.commitTimerTask = new CommitTimerTask();` and not `commitTimerTask = new CommitTimerTask();`
2. The conversion between milliseconds and nanoseconds was incorrect it should have been `maxDeferredPeriod * 1000000L` and not `maxDeferredPeriod * 1000L`.

### DIFF
--- a/bundles/org.openhab.core.storage.json/src/main/java/org/openhab/core/storage/json/internal/JsonStorage.java
+++ b/bundles/org.openhab.core.storage.json/src/main/java/org/openhab/core/storage/json/internal/JsonStorage.java
@@ -23,13 +23,15 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.Timer;
-import java.util.TimerTask;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


kaikreuzer:
Thanks for this PR!
Just for my understanding: Why are there idling threads? I would have thought that the current implementation is meant to have only exactly one thread being active, doing the next scheduled flush. Through the bug that you have discovered, the cancelling and rescheduling does not work, so I assume that this is the reason for the multiple threads, isn't it?

But if this is the case, fixing the bug should be enough. Introducing a thread pool (with 10 threads) does not seem to be very useful here or am I missing something?

joerg1985:
There was one instance per file and one timer thread per instance, on my system 20 instances are a live, backed by 20 threads. Now there is a named pool of 5 threads (as per default configuration) shared by all 20 instances. 

`ThreadPoolManager.getScheduledPool` will allways return the same instance from the internal cache. I did not make the `scheduledExecutorService` field `static`, as other areas using the `ThreadPoolManager` store the pool in an not static field. So i did assume this is the code style used here.

kaikreuzer (approved):
Ah, that's what I was missing indeed - we have of course many instances of `JsonStorage`. Thanks for the explanation!
Yeah, in this case your change makes a lot of sense. Thanks for it and the fixes that come included!

PR #3877, opened 2023-11-20
[ui] create Text widgets for read only Switch items

ccutrer:
Prompted by https://github.com/openhab/openhab-addons/pull/15927


ccutrer:
I haven't been able to test this... when I try to install my locally compiled bundle into my running openHAB instance, it can't resolve all the dependencies. Any pointers on how to do so?

antroids:
> I haven't been able to test this... when I try to install my locally compiled bundle into my running openHAB instance, it can't resolve all the dependencies. Any pointers on how to do so?

You could try to checkout the same tag as the version of your instance, merge your changes and replace the bundle.

lolodomo:
I already imagine lots of users complaining that the switch rendering is no more there, that they prefer to see a switch even if read only, that they can use the text widget to get only a text.

My feeling is that it is rather a UI problem. The app should disable the control of the switch when only readable.

ccutrer:
That's a very good point. Will close and look into front ends disabling controls.

PR #3881, opened 2023-11-22
[oauth] Add support for passing extra fields to oauth token request

x1mc:
As is the oauth2 client does not support oauth2 endpoints that require more than the standard fields in the form data posted to the token URL.  This PR adds that support.

An example of where this is needed is the Velux API described here: https://github.com/nougad/velux-cli/blob/master/velux-protocol.md#initial-login

The `user_prefix` param is non standard and there was no way to include it with this client.

I tried to keep the existing constructor options in place so this shouldn't break anything outside of oauth2client that happens to be using the OAuthConnector class.

### DIFF
--- a/bundles/org.openhab.core.auth.oauth2client/src/main/java/org/openhab/core/auth/oauth2client/internal/OAuthClientServiceImpl.java
+++ b/bundles/org.openhab.core.auth.oauth2client/src/main/java/org/openhab/core/auth/oauth2client/internal/OAuthClientServiceImpl.java
@@ -25,6 +25,7 @@ import java.util.UUID;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.jetty.util.Fields;
 import org.eclipse.jetty.util.UrlEncoded;
 import org.openhab.core.auth.client.oauth2.AccessTokenRefreshListener;
 import org.openhab.core.auth.client.oauth2.AccessTokenResponse;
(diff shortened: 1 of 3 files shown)


lolodomo:
This would interesting to get @clinique feedback as I remember he has to handle something not standard for the netatmo binding.

x1mc:
> This would interesting to get @clinique feedback as I remember he has to handle something not standard for the netatmo binding.

Not surprising as the Velux Active API appears to just be an extension of the netatmo API. I've been playing around with creating a binding for Velux Active (different than the existing Velux binding) and I did have to pull in the oauth response deserializer from the netatmo binding for that (see [here](https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/ApiBridgeHandler.java#L153-L158), [here](https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/deserialization/AccessTokenResponseDeserializer.java) and [here](https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/dto/NetatmoAccessTokenResponse.java)), which perhaps is the non standard piece @clinique had to handle.

clinique:
You're correct. I had to use a special deserializer because Netatmo API does not provide the scope field as expected in the RFC.

kaikreuzer:
> The correction is still not done on [Netatmo side](https://forum.netatmo.com/viewtopic.php?t=19293).

... since almost 3 years. So it probably will never be done. 🙄

kaikreuzer (approved):
Changes look good to me, thanks for the enhancement!

jlaur:
@x1mc - do you need a tester for your Velux Active binding? 🙂 I was in contact with Velux a few years ago to know their standpoint if such an integration would appear, and they discouraged me from writing one. Nevertheless, I would still love one, especially if it would make it possible to switch between home and away from openHAB rather than having to press that stupid button. 😉

PR #3884, opened 2023-11-23
Use a single thread to watch all event executors

joerg1985:
# Description

This PR will use a single thread to watch all event executors. Having one watcher thread for each event thread is not needed, the only task is to print a log line with a warning about slow dispatching of a event.

On my system, this one thread will replace 15 watcher threads.


kaikreuzer (approved):
Great, again a few threads less! 👍

PR #3888, opened 2023-11-25
Fix AutoUpdatePolicy for channel

### LINKED ISSUE #3887: Auto update policy is not working on channel level
In this XSD:
https://www.openhab.org/schemas/thing-description-1.0.0.xsd

`autoUpdatePolicy` is defined at channel level:

```xml
<xs:complexType name="channel">
    <xs:sequence>
        <xs:element name="label" type="xs:string" minOccurs="0"/>
        <xs:element name="description" type="xs:string" minOccurs="0"/>
        <xs:element name="properties" type="thing-description:properties" minOccurs="0"/>
        <xs:element name="autoUpdatePolicy" type="thing-description:auto-update-policy" minOccurs="0"/>
    </xs:sequence>
    <xs:attribute name="id" type="config-description:idRestrictionPattern" use="required"/>
    <xs:attribute name="typeId" type="thing-description:namespaceIdRestrictionPattern" use="required"/>
</xs:complexType>
```

However, this has no effect. Example:

```xml
<channel id="brightness" typeId="system.brightness">
    <autoUpdatePolicy>veto</autoUpdatePolicy>
</channel>
```

This has also been reported in the foum:
https://community.openhab.org/t/autoupdatepolicy-seems-to-be-ignored-on-xml-channel-definitions/149271

## Expected Behavior

Defined auto update policy for the channel should be respected.

## Current Behavior

Defined auto update policy for the channel is ignored, and the auto update policy from the channel type will be used.

## Steps to Reproduce

1. Declare a channel with veto policy.
2. Send a command to the channel for changing state.
3. Observe that state is updated immediately.

## Context

I'm currently working on the Hue binding where many API v2 channels are using system channel types. I need to apply the **veto** policy to all of them, so currently they all need to be duplicated into the binding for this to work.

## Your Environment
* Version used: 4.1.0-SNAPSHOT - Build 3725

jlaur:
Fixes #3887

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/builder/ChannelBuilder.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/builder/ChannelBuilder.java
@@ -105,6 +105,10 @@ public class ChannelBuilder {
         if (description != null) {
             channelBuilder.withDescription(description);
         }
+        AutoUpdatePolicy autoUpdatePolicy = channel.getAutoUpdatePolicy();
+        if (autoUpdatePolicy != null) {
+            channelBuilder.withAutoUpdatePolicy(autoUpdatePolicy);
+        }
         return channelBuilder;
     }
 
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingFactoryHelper.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingFactoryHelper.java
@@ -26,6 +26,7 @@ import org.openhab.core.thing.ChannelUID;
 import org.openhab.core.thing.ThingUID;
 import org.openhab.core.thing.binding.ThingFactory;
 import org.openhab.core.thing.binding.builder.ChannelBuilder;
+import org.openhab.core.thing.type.AutoUpdatePolicy;
 import org.openhab.core.thing.type.ChannelDefinition;
 import org.openhab.core.thing.type.ChannelGroupDefinition;
 import org.openhab.core.thing.type.ChannelGroupType;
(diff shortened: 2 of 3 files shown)


bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/builder/ChannelBuilder.java
```diff
@@ -105,6 +105,10 @@ public static ChannelBuilder create(Channel channel) {
         if (description != null) {
             channelBuilder.withDescription(description);
         }
+        AutoUpdatePolicy autoUpdatePolicy = channel.getAutoUpdatePolicy();
```
J-N-K, line 108: Since `.withAutoUpdatePolicy` accepts `null` you can just use that in l. 99

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingFactoryHelper.java
```diff
@@ -184,12 +185,17 @@ public static ChannelBuilder createChannelBuilder(ChannelUID channelUID, Channel
...
+            autoUpdatePolicy = channelType.getAutoUpdatePolicy();
+        }
+
         final ChannelBuilder channelBuilder = ChannelBuilder.create(channelUID, channelType.getItemType()) //
                 .withType(channelType.getUID()) //
                 .withDefaultTags(channelType.getTags()) //
                 .withKind(channelType.getKind()) //
                 .withLabel(label) //
-                .withAutoUpdatePolicy(channelType.getAutoUpdatePolicy());
+                .withAutoUpdatePolicy(autoUpdatePolicy);
```
J-N-K, line 198: Can you add a test-case for this, too? The test should show that

- no policy is set if neither the channel-definition nor the channel-type-definition set a policy
- if either channel- or channel-type-definition contain a policy, it is used
- if both contain a policy, the channel-definition takes precedence

J-N-K (changes requested):
Thanks. That is really a nasty bug.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingFactoryHelper.java
jlaur: Just to be sure, you mean a new test class, right? I didn't find any existing coverage.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/builder/ChannelBuilder.java
jlaur: Right - done.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingFactoryHelper.java
jlaur: I'm afraid I need a bit of guidance here. I found some integration tests that seem related:
https://github.com/openhab/openhab-core/blob/3624682c1e2574d395b9fd0e81592404365722b8/itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/internal/ThingManagerOSGiJavaTest.java#L347-L382

Did you mean to add new tests here? In that case, how can I build and run those tests?

PR #3889, opened 2023-11-26
Add time-series support for websockets

J-N-K:
Follow-up to #3597 

This allows sending time-series events from web socket connections.


kaikreuzer (approved):
Thanks!

Issue #3868, opened 2023-11-09
Develop additional AddonSuggestionFinder components


andrewfg:
@andrewfg It is maybe worth adding some futures to your architecture document as well (maybe make that an issue here as well):
- MQTT broker discovery
- discovery of JSON storage provider addons
- discovery of marketplace addons
- serial discovery (could be useful to discover and suggest ZWave and Zigbee dongles and suggest the binding)
- bluetooth discovery (may need moving some of the core bluetooth functionality from addons to core similar to mDNS and uPnP)

Note that Homeassistant does serial and bluetooth discovery on top of mDNS and SSDP (uPnP).

I don't know how feasible each of these is, but that would build upon what you are doing here.

_Originally posted by @mherwege in https://github.com/openhab/openhab-core/issues/3806#issuecomment-1794586738_

holgerfriedrich:
Other ideas:
* Use ```ProcessHandle.allProcesses()``` to get a list of process running on the server. We could first try to cover the optional components installable via openHABian:
![image](https://github.com/openhab/openhab-core/assets/2121139/be7c999f-cf20-4a10-9b98-0ce75f7aba70)
Thinks like influxdb, knxd, etc should be easy to detect.
* Multicast/Broadcast discovery: construct a special IP frame and check if a device responds. Not sure how to generalize the construction of the special IP frame. Would need either binding-specific finders in core (ugly but effective) or a way to blueprint IP frames with patterns for IP address and ports etc. See https://github.com/openhab/openhab-core/pull/3806#issuecomment-1795804713

J-N-K:
For 3rd Party JSON add-on service the discovery information could be made available from the `addons.json`, but it would require the add-on service to be installed when the assistant runs. That would require the assistant to run constantly (and maybe showing suggestions in the add-on store instead of the suggestions we have now).

mherwege:
> For 3rd Party JSON add-on service the discovery information could be made available from the `addons.json`, but it would require the add-on service to be installed when the assistant runs. That would require the assistant to run constantly (and maybe showing suggestions in the add-on store instead of the suggestions we have now).

[...]

PR #3029, opened 2022-07-04
Upgrade to JmDNS 3.5.8

kaikreuzer:
For changes see https://github.com/jmdns/jmdns/milestone/10?closed=1

Signed-off-by: Kai Kreuzer <kai@openhab.org>

PR #3904, opened 2023-12-08
Service to suggest addons based on running processes

holgerfriedrich:
This is a first draft of a new finder to suggest addons based on the software running on the OH server.
It checks the process list and tries to detect known software.

It has been mentioned in
https://github.com/openhab/openhab-core/issues/3868#issuecomment-1806398840

It will require changes in the ```addon.xml``` of the binding as well, but no change to the xsd, as the given ```match-property``` is reused. Here an example for influxdb.

```
       <discovery-methods>
               <discovery-method>
                       <service-type>process</service-type>
                       <match-properties>
                               <match-property>
                                       <name>command</name>
                                       <regex>.*influxd.*</regex>
                               </match-property>
                       </match-properties>
               </discovery-method>
       </discovery-methods>
```

@andrewfg Maybe you can have a short look if I forgot something before I try to create a full distro package to test it.
Any recommendation on how to test in a simpler way would be very welcome....


bundles/org.openhab.core.config.discovery.addon.process/src/main/java/org/openhab/core/config/discovery/addon/process/ProcessAddonFinder.java
```diff
@@ -0,0 +1,119 @@
...
+            return "";
+        String[] argsArray = args.get();
+        if (argsArray.length < 1)
+            return "";
+        return argsArray[0];
+    }
+
+    @Override
+    public Set<AddonInfo> getSuggestedAddons() {
+        Set<String> processList = ProcessHandle.allProcesses().map(this::getProcessCommandProcess)
```
mherwege, line 74: Shouldn’t you catch SecurityException here?

kaikreuzer:
> We went out of our way to be able to switch off discovery for UPnP and mDNS. I don’t think it is required here as this has lower impact. But we should all agree on it.

I would prefer to have the same mechanism in place for all the discoveries. Having processes scanned in the background might not be something everybody wants.

mherwege:
> Having processes scanned in the background

The PR is it is does not cache the running processes anywhere, so there is no background scanning. It will just make a call to get the processes when called from the suggestions REST endpoint. In the UI that means: when running the setup wizard or when opening the addon store. We can of course allow to disable that call as well. The main reason to be able to enable/disable was to avoid having all UPnP/mDNS dependencies loaded and running in the background only for addon discovery. As there is no extra dependency here, and as long as there is no scheduled scan and result caching, I don't see the point.

kaikreuzer:
> But of course, see my concern about caching.

Yes, that's why I was assuming that the logic of the PR will be changed to do some background scanning in future.

andrewfg:
> Any recommendation on how to test in a simpler way would be very welcome....

None really. It crosses over from addons to core so the only way to test is a full distro.

mherwege:
> The easy way to test this is to put an xml file in the userdata/addons folder with the additional discovery content. Any xml file with valid content will be picked up from there. So there is no need to do a full build.

@andrewfg @holgerfriedrich I think this is the easy way to test without a full build. You can test in the IDE.

holgerfriedrich:
Thank you all for the comments.

I finally got it working for knxd.....
```
18:35:45.323 [TRACE] [very.addon.process.ProcessAddonFinder] - Candidate binding-knx, pattern ".*knxd"
18:35:45.324 [TRACE] [very.addon.process.ProcessAddonFinder] - Candidate binding-knx, pattern ".*knxd.exe$"
18:35:45.324 [DEBUG] [very.addon.process.ProcessAddonFinder] - Suggested add-on found: binding-knx
```

I will create a corresponding PR in addons repo. The finder allows to specify multiple patterns.
Windows and Linux can be handled using separate patterns if needed.

I did not manage yet to test without building a distro package. Still not sure how to do core development with Visual Studio Code.

For now, it is without caching and background scanning. ```getSuggestedAddons``` should finish quickly, so for now no need to extend this.

holgerfriedrich:
Both this PR and PR in addons repo ready for review.

Unfortunatly, this currently does only work for bindings, not for others like persistence. See https://github.com/openhab/openhab-core/issues/3868#issuecomment-1848741848.
Should I create an issue for that or do you think this is out of scope?

mherwege:
> The UI (add-on store) shows suggestions only for bindings.

PR https://github.com/openhab/openhab-webui/pull/2213 created to solve this.

bundles/org.openhab.core.config.discovery.addon/src/main/java/org/openhab/core/config/discovery/addon/AddonFinderConstants.java
```diff
@@ -33,14 +33,20 @@ public class AddonFinderConstants {
...
     public static final String CFG_FINDER_UPNP = "suggestionFinderUpnp";
     public static final String SERVICE_NAME_UPNP = SERVICE_TYPE_UPNP + ADDON_SUGGESTION_FINDER;
     public static final String FEATURE_UPNP = ADDON_SUGGESTION_FINDER_FEATURE + SERVICE_TYPE_UPNP;
 
-    public static final List<String> SUGGESTION_FINDERS = List.of(SERVICE_NAME_MDNS, SERVICE_NAME_UPNP);
+    public static final List<String> SUGGESTION_FINDERS = List.of(SERVICE_NAME_MDNS, SERVICE_NAME_PROCESS,
+            SERVICE_NAME_UPNP);
     public static final Map<String, String> SUGGESTION_FINDER_CONFIGS = Map.of(SERVICE_NAME_MDNS, CFG_FINDER_MDNS,
-            SERVICE_NAME_UPNP, CFG_FINDER_UPNP);
+            SERVICE_NAME_PROCESS, CFG_FINDER_PROCESS, SERVICE_NAME_UPNP, CFG_FINDER_UPNP);
```
mherwege, line 49: No need to add it here, as there is no configuration parameter defined.

features/karaf/openhab-core/src/main/feature/feature.xml
```diff
@@ -83,6 +83,12 @@
 		<feature dependency="true">openhab.tp-jmdns</feature>
 	</feature>
 
+	<feature name="openhab-core-config-discovery-addon-process" version="${project.version}">
```
mherwege, line 86: If you add this feature to `openhab-runtime-base` you will not need to add this finder to the constants in AddonFinderConstants.java. The feature will then automatically get installed and not through the service. The only reason to do it through the service is to allow enabling/disabling.

bundles/org.openhab.core.config.discovery.addon.process/src/main/java/org/openhab/core/config/discovery/addon/process/ProcessAddonFinder.java
```diff
@@ -0,0 +1,128 @@
...
+ *
+ * @author Holger Friedrich - Initial contribution
+ */
+@NonNullByDefault
+@Component(service = AddonFinder.class, name = ProcessAddonFinder.SERVICE_NAME)
+public class ProcessAddonFinder extends BaseAddonFinder {
+
+    private static final String COMMAND = "command";
+
+    public static final String SERVICE_TYPE = SERVICE_TYPE_PROCESS;
```
mherwege, line 49: You can eliminate these constants from AddonFinderConstants.java and define them here instead with the same logic.

bundles/org.openhab.core.config.discovery.addon.process/src/main/java/org/openhab/core/config/discovery/addon/process/ProcessAddonFinder.java
```diff
@@ -0,0 +1,128 @@
...
+        Set<String> processList = Collections.emptySet();
+        try {
+            processList = ProcessHandle.allProcesses().map(this::getProcessCommandProcess)
+                    .filter(Predicate.not(String::isEmpty)).collect(Collectors.toUnmodifiableSet());
+        } catch (SecurityException | UnsupportedOperationException unused) {
+            logger.info("Cannot obtain process list, suggesting add-ons not possible");
+            return result;
+        }
+        // logging task list is commented out by default due to privacy reasons
+        // logger.trace("Processes visible: {}", processList.toString());
```
mherwege, line 86: Agreed, so remove from code.

features/karaf/openhab-core/src/main/feature/feature.xml
andrewfg: I think someone did already ask to have enabling/disabling. ??

mherwege: For me, that's conditional. If retrieving the processes is too high a load, we should. Otherwise I don't see the point. Right now, the implementation does not cache anything and does a call to retrieve the processes any time the suggestions are retrieved. If that is quick, no need to enable/disable a little piece of code. There is nothing running in the background in separate threads either.
@kaikreuzer Did not react on my feedback on it. If he feels strongly, then, of course it needs to be there, but it should also be added to the configuration parameters.

andrewfg: @mherwege I am OK with that.

bundles/org.openhab.core.config.discovery.addon.process/src/main/java/org/openhab/core/config/discovery/addon/process/ProcessAddonFinder.java
holgerfriedrich: ...but I had to make ADDON_SUGGESTION_FINDER public...

mherwege (approved):

andrewfg (approved):
LGTM

holgerfriedrich:
@kaikreuzer maybe you could have a short look and merge this and corresponding openhab/opengab-addons#16031?
(already approved Mark and Andrew)
Thanks!

holgerfriedrich:
Needs to be adapted after #3029 is merged

andrewfg:
@holgerfriedrich just for the avoidance of doubt: di you adapt to the new addon.xml schema?
@kaikreuzer / @mherwege I would suggest to wait until after those new addon.xml schema PRs have been merged.

mherwege:
> @holgerfriedrich just for the avoidance of doubt: di you adapt to the new addon.xml schema? @kaikreuzer / @mherwege I would suggest to wait until after those new addon.xml schema PRs have been merged.

There is no impact of the new format for this PR. There are no discovery parameters.

andrewfg:
> There are no discovery parameters. 

@holgerfriedrich the whole point of the revised schema is because you mis-used the match-properties as if they would be discovery-parameters. Therefore you need to change your mis-used properties to parameters instead.

@mherwege for info..

mherwege:
@andrewfg I think you confuse this PR with this one: https://github.com/openhab/openhab-core/pull/3920

The other one is the one I made the change for, and is not yet ready.

andrewfg:
> The other one is the one I made the change for

Oops! Sorry.

bundles/org.openhab.core.config.discovery.addon.process/src/main/java/org/openhab/core/config/discovery/addon/process/ProcessAddonFinder.java
```diff
@@ -0,0 +1,126 @@
...
+    @Override
+    public Set<AddonInfo> getSuggestedAddons() {
+        logger.trace("ProcessAddonFinder::getSuggestedAddons");
+        Set<AddonInfo> result = new HashSet<>();
+        Set<String> processList = Collections.emptySet();
+        try {
+            processList = ProcessHandle.allProcesses().map(this::getProcessCommandProcess)
+                    .filter(Predicate.not(String::isEmpty)).collect(Collectors.toUnmodifiableSet());
+        } catch (SecurityException | UnsupportedOperationException unused) {
+            logger.info("Cannot obtain process list, suggesting add-ons not possible");
```
kaikreuzer, line 82:
```suggestion
            logger.info("Cannot obtain process list, suggesting add-ons based on running processes is not possible");
```

kaikreuzer (approved):
lgtm, thanks!

kaikreuzer:
@holgerfriedrich I just tried the latest distro snapshot (3787) and see this log during startup:
```
15:00:02.417 [WARN ] [ternal.xml.AddonInfoAddonsXmlProvider] - File 'openhab-addons.xml' has invalid content
```
Any idea why this happens?

kaikreuzer:
Hm, I have an idea: My version of the file is from Dec 10, i.e. it wasn't overwritten during the update. We will need to fix this, but it isn't relevant for the milestone release as the file didn't exist in the previous milestone.

holgerfriedrich:
> @holgerfriedrich I just tried the latest distro snapshot (3787) and see this log during startup:
> 
> ```
> 15:00:02.417 [WARN ] [ternal.xml.AddonInfoAddonsXmlProvider] - File 'openhab-addons.xml' has invalid content
> ```
> 
> Any idea why this happens?

@kaikreuzer this happened before when the schema was adapted. I think we have to improve logging in the addon info provider. 

> Hm, I have an idea: My version of the file is from Dec 10, i.e. it wasn't overwritten during the update. We will need to fix this, but it isn't relevant for the milestone release as the file didn't exist in the previous milestone.

This is good. In case of parser failure, all of the suggestions finders won't work.

May it be a problem for users of the daily snapshots?

andrewfg:
> Any idea why this happens?

@holgerfriedrich / @kaikreuzer / @mherwege - we changed the schema between two snapshots; so perhaps the new snapshot failed (correctly) to parse an addons.xml file from a prior snapshot based on the prior schema. Therefore I think that the final build this will probably have all reconciled again. => But please do check before the final final release!

andrewfg:
> I think we have to improve logging in the addon info provider.

A few things..
1. I did already improve the logging so that each entry has its regex validated when the individual binding entries are initialized.
2. Your problem above is that the whole file is read in one go before the individual entries are validated, and as you say any single bad entry will kill the whole load.
3. The good news is that -- _**if the new schema is on the server**_ -- the add-ons CI build of the will check each add-on individually and fail specifically for that add-on.

So I think that once everything has synchronized between the build processes of openhab-core and openhab-addons then the problem should have been resolved..

mherwege:
> Hm, I have an idea: My version of the file is from Dec 10, i.e. it wasn't overwritten during the update. We will need to fix this, but it isn't relevant for the milestone release as the file didn't exist in the previous milestone.

The file is in the userdata folder. Does it get deleted and recreated with an upgrade? If not we need to make sure this is done?

kaikreuzer:
As mentioned above: "it wasn't overwritten during the update." So yes, it is an issue for anyone using snapshot updates and it will be a problem in future for anyone else. So we need to fix this.

mherwege:
I don’t know how the installation and upgrades work exactly. @J-N-K can you help with this? Is it something to be done in the distro? Should the file go somewhere else to be removed and recreated in an upgrade? The code in core only takes care of reading the file that is there.

andrewfg:
For the avoidance of doubt..
- If it is a clean (over-write) install then surely any pre-existing openhab-addos.xml will be overwritten by the contents of the openhab-4.1.0-SNAPSHOT.zip
- If it is an upgrade (incremental) install via a script e.g. via openHabian then I guess the script can decide which files to overwrite and which not

What other scenarios are we dealing with than the above?

mherwege:
This file should always be overwritten, as it will contain the latest info when upgrading.

The issue is here: https://github.com/J-N-K/openhab-distro/blob/bf272a68a0f52a3474a069f4ed1e517b1b338ea6/features/distro/src/main/feature/feature.xml#L13

override should be true to force an overwrite.

I am not behind a computer now, so tough to create a PR, but it should be done to correct this.

J-N-K:
I guess there are several solutions:

1. Adapt the update script, so the file get's deleted on upgrade (add an additional line after l. 323 / l. 744
2. Change the file location to `userdata/etc` and add it to `userdata_sysfiles.lst`
3. Change the file location to `userdata/cache`

I would probably vote for the last one, then it SHOULD be recreated when you clear the cache (which especially for snapshots might be an issue otherwise).

J-N-K:
Or (even better) do it like @mherwege suggests.

andrewfg:
@holgerfriedrich I have a couple of (admittedly very late) questions concerning this finder resp. PR..

1. Does it work on all Operating Systems i.e. MacOS, Windows, Linux, resp. other Unix platforms?
2. Does it detect all types of process i.e. both regular executables and also services?

holgerfriedrich:
@andrewfg I use java to obtain a list of processes, see https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/ProcessHandle.html#allProcesses()
The output is different on Linux and Windows. ```getProcessCommandProcess()``` was introduced to harmonize the output on different OS. Windows will return empty entries for some processes. May be due to user permissions.

I tested with Linux and Windows. The trace output which dumps the process list got removed in one of the last commits.
```logger.trace("Processes visible: {}", processList.toString());``` - maybe you want to add it and see.

**Comment by andrewfg** [comment]
So to answer (some of) my own questions:

1. At least on Linux the finder DOES identify processes that were started as a service. See log listings below. HOWEVER in such cases the `command` parameter is 'null', and the `arguments[0]` parameter can be irrelevant (the important information may be in `arguments[n]` where 'n' may vary from case to case). => Therefore the current version of this finder service is NOT able to suggest addons based on running services.
2. At least on Linux the DOES identify processes for scripts that are being run by a script execution application (see the Python example in the list below). However the current version of this finder service is NOT able to suggest addons based on running script services.

```
command:null, arguments:[], commandLine:/lib/systemd/systemd-journald 
command:null, arguments:[-c, /etc/mosquitto/mosquitto.conf], commandLine:/usr/sbin/mosquitto -c /etc/mosquitto/mosquitto.conf 
command:null, arguments:[-n, -iNONE], commandLine:/usr/sbin/rsyslogd -n -iNONE 
command:null, arguments:[-u, /home/openhabian/grott/grott.py], commandLine:/usr/bin/python3 -u /home/openhabian/grott/grott.py 
command:null, arguments:[/dev/serial1, bcm43xx, 3000000, flow, -], commandLine:/usr/bin/hciattach /dev/serial1 bcm43xx 3000000 flow - 
command:null, arguments:[], commandLine:/usr/libexec/bluetooth/bluetoothd 
command:null, arguments:[-w, -q], commandLine:/usr/sbin/dhcpcd -w -q 
```

Therefore I will shortly create a new PR to improve this finder so that it can discover addons that depend on running services and/or scripts..

andrewfg:
I am thinking of keeping the existing property `command` as it is. And adding a second property `commandLine`. That should avoid breaking anything existing.

NEVERTHELESS I note that your existing code may take `arguments[0]` as a fallback if `command` is null. However from my tests on Linux, it seems to me that `arguments[-1]` (which doesn’t actually exist) would be the proper fallback for `command`. => So can you please tell me on what platform / test case did you find that `arguments[0]` would be the equivalent of `command`?

PR #3905, opened 2023-12-08
[addon] Fix localisation of suggested Add-ons

andrewfg:
Fixes [#2204](https://github.com/openhab/openhab-webui/issues/2204)

Since time is of the essence, I am opening this PR now for code review. I will create and test a full OH build tomorrow.

Signed-off-by: Andrew Fiddian-Green <software@whitebear.ch>

PR #3908, opened 2023-12-09
Fix addon name localisation before installation

mherwege:
Fixes https://github.com/openhab/openhab-webui/issues/2204
Replaces https://github.com/openhab/openhab-core/pull/3905

The issue is that no addon info is available for distribution addons before installation. The only thing available is the feature description and the addon endpoint gets it from the addon data, not the addonInfo data. So a merge of addonInfo is not even happening.
This change introduces a flag in AddonInfo indicating if it is the master info or not. When the info is coming from the addon repository scan, it is flagged to not be the master data and name and description will be ignored (because localization info is not available). Addon services will check if the addon info is master info, and only then update name and description.

### DIFF
--- a/bundles/org.openhab.core.addon.eclipse/src/main/java/org/openhab/core/addon/eclipse/internal/EclipseAddonService.java
+++ b/bundles/org.openhab.core.addon.eclipse/src/main/java/org/openhab/core/addon/eclipse/internal/EclipseAddonService.java
@@ -142,9 +142,12 @@ public class EclipseAddonService implements AddonService {
         AddonInfo addonInfo = addonInfoRegistry.getAddonInfo(uid, locale);
 
         if (addonInfo != null) {
-            // only enrich if this add-on is installed, otherwise wrong data might be added
-            addon = addon.withLabel(addonInfo.getName()).withDescription(addonInfo.getDescription())
-                    .withConnection(addonInfo.getConnection()).withCountries(addonInfo.getCountries())
+            if (addonInfo.isMasterAddonInfo()) {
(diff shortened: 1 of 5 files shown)


mherwege:
@andrewfg FYI
@kaikreuzer I think this should solve the isse.

andrewfg:
> I think this should solve the issue.

@mherwege / @kaikreuzer shall I do my usual full build clean install test before you merge it? i.e. tomorrow.

kaikreuzer (approved):
Thanks, let's have a try with this.

andrewfg:
> let's have a try with this.

@kaikreuzer / @mherwege all Ok; it works on my full build.

jimtng:
@mherwege should we just revert this instead of #4572? isMasterAddon was added for the sole purpose that #4572 is removing. The only difference is the `NA` that would also get restored by reverting this. 

Alternatively we could revert this, and then remove the NA stuff?

mherwege:
@jimtng Indeed, I think this should be reverted. I don't think there is any harm in having the NA treatment back, and that keeps us on the safe side. It might be needed when loading if all providers are not available at the same time.

jimtng:
> If we revert this, is there a need to update the developer documentation to point out the name and description are required and they should be provided in full in the addon.xml? Reviewer instructions?

This one needs to be updated:
https://www.openhab.org/docs/developer/utils/i18n.html#using-custom-keys

jimtng:
@mherwege I'll prepare a PR to revert this

jimtng:
#4573

Issue #3912, opened 2023-12-11
AddonSuggestionFinder component for KNX


holgerfriedrich:
KNXnet/IP devices can be discovered using multicast traffic to IP 224.0.23.12, port 3671.

A draft implementation of a finder has already been mentioned in https://github.com/openhab/openhab-core/pull/3806#issuecomment-1796852618.
It is quite simple, opening a socket, preparing a frame, send it out, and wait a while for responses.

The question is if we want an AddonSuggestionFinder which is very specific to a single binding.

A generalization would be possible, but more complicated. For a more generic "IP" finder, more information like target IP, content of search frame, and expected response could be stored in the match properties of addon.xml. However, it already starts getting complicated already when port or IP address needs to be placed dynamically into the frame.

Not sure if this is worth the effort, as I cannot judge for which addons this could be useful.

holgerfriedrich:
@mherwege @andrewfg @kaikreuzer Could you share your opinion in this?
@kaikreuzer What is the deadline for the implementation / merge to make it for 4.1 release?

mherwege:
I don't think introducing binding specific finders in core is the way to go. Therefore, if we ever want to do it this way (creating binding specific logic), I would argue we would need a special class of add-on in the add-on repository for a finder and install and run them all by default (otherwise it does not make sense, if you first have to manually install a finder to find something you can install the add-on just as well).
In my view, we should do it if it can be generalized. And I do think there may be other bindings that would benefit from it (sending an IP broadcast frame and expecting a response). I myself coded such logic in the nikohomecontrol binding. Version 1 of the system was only discoverable in this way. Version 2 can also be discovered using mDNS. The binding however uses the IP (UDP) broadcast for discovery and does some packet analysis on it.
I think the best way forward here is to go and look for all bindings that do this kind of discovery and analyse what is being sent and how the result is interpreted analysed. If there is commonality in at least some of them, we could define additional discovery criteria parameters in the xml and implement a general finder that uses these.
If you ask me, I think this will be after a 4.1 release. While I see value, I believe it needs more analysis to decide on what to do.
This, of course, is just my view and I want to hear the opinion of others.

kaikreuzer:
I fully agree to everything @mherwege said. 😎

andrewfg:
[...]

PR #3920, opened 2023-12-13
Service to suggest addons via generic IP scan

holgerfriedrich:
This is the implementation of a generic IP-based add-on suggestion finder.
It will try to discover add-ons by scanning the IP network with specific packages.

The description is taken from the metadata.

At startup, a single scan is done in a separate thread.

The finder can be disabled, as it actively sends packets to the network which might not be desired. 

Example configuration:
for KNX we want to send a multicast package to 224.0.23.12 port 3671.

The section in ```addon.xml``` looks like this:
```
		<discovery-method>
			<service-type>ip</service-type>
			<match-properties>
				<match-property>
					<name>type</name>
					<regex>ip_multicast</regex>
				</match-property>
				<match-property>
					<name>dest_ip</name>
					<regex>224.0.23.12</regex>
				</match-property>
				<match-property>
					<name>dest_port</name>
					<regex>3671</regex>
				</match-property>
				<match-property>
					<name>request</name>
					<regex>0x06 0x10 0x02 0x01 0x00 0x0e 0x08 0x01 $src_ip $src_port</regex>
				</match-property>
				<match-property>
					<name>timeout_ms</name>
					<regex>5000</regex>
				</match-property>
				<match-property>
					<name>response</name>
					<regex>any</regex>
				</match-property>
			</match-properties>
		</discovery-method>
```

Refs: #3912


mherwege:
> <discovery-method>
> 			<service-type>ip</service-type>
> 			<match-properties>
> 				<match-property>
> 					<name>type</name>
> 					<regex>ip_multicast</regex>
> 				</match-property>
> 				<match-property>
> 					<name>dest_ip</name>
> 					<regex>224.0.23.12</regex>
> 				</match-property>
> 				<match-property>
> 					<name>dest_port</name>
> 					<regex>3671</regex>
> 				</match-property>
> 				<match-property>
> 					<name>request</name>
> 					<regex>0x06 0x10 0x02 0x01 0x00 0x0e 0x08 0x01 $src_ip $src_port</regex>
> 				</match-property>
> 				<match-property>
> 					<name>timeout_ms</name>
> 					<regex>5000</regex>
> 				</match-property>
> 				<match-property>
> 					<name>response</name>
> 					<regex>any</regex>
> 				</match-property>
> 			</match-properties>
> 		</discovery-method>

It feels to me like we are trying to shoe-horn the parameters for this into the existing schema. You may have noticed that the mdns discovery has an extra parameter in the schema. I would suggest to extend the schema for this.
Also, I don't like using a field called `regex` and filling it with something else.

I think you should keep the check for the return value to be a pure regex (you can express hex numbers in a regex as well). All the others are not matching criteria, but input parameters.

mherwege:
> At startup, a single scan is done in a separate thread.

Shouldn't we run this on a schedule and repeat it at regular intervals? The interval could be another configuration.

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -0,0 +1,256 @@
...
+        logger.warn("IpAddonFinder::IpAddonFinder");
+    }
+
+    @Deactivate
+    public void deactivate() {
+        logger.warn("IpAddonFinder::deactivate");
+        stopScan();
+    }
+
+    void startScan() {
```
mherwege, line 90: I don't see this being called.

holgerfriedrich:
Thanks, Mark, for looking into this.

I wanted to see it working first, changing the schema is probably something to consider, as the regex for input seems wrong to me as well.

I am struggling a bit putting the scan from the getter function to task scheduled from the ctor... The list of candidates is now empty. I need to check later.

andrewfg:
@holgerfriedrich yup, the way that you mis- use the schema is definitely a horrible hack; but as a proof of concept this approach makes sense; however I still wonder if there are other add-ons other than KNX where such a network probe is useful. ??

andrewfg:
> That would be extendable without modifying the schema.

@mherwege we would still have to modify (extend) the schema to support the new parameter(s) elements. But I get what you are saying about the extensibility. But we should wait until after v4.1 for this. Let is see whether other finders require a yet more sophisticated model. Note my draft for the USB device related bindings #3922 would not need extra parameters.

holgerfriedrich:
Proof of concept is working,
image
image

holgerfriedrich:
> > At startup, a single scan is done in a separate thread.
> 
> Shouldn't we run this on a schedule and repeat it at regular intervals? The interval could be another configuration (next to switching this finder on/off completely).

KNX is not something you turn on and off, so I think it is fine to trigger it once at startup.
Likely this is desired for other bindings.
We need to analyze first which devices we want to detect using this approach.

holgerfriedrich:
> we should probably have defined the mdnsServiceType parameter in the schema differently to make it more generalizable, something like:
```
<discovery-method>
 	<service-type>mdns</service-type>
 	<parameters>
 	 	<parameter>
 	 	 	<name>mdnsServiceType</name>
 	 	 	<value>...</value>
```

This is probably a good idea.... Should we change it before this gets adopted in more places like the marketplace?

mherwege:
> > we should probably have defined the mdnsServiceType parameter in the schema differently to make it more generalizable, something like:
> 
> ```
> <discovery-method>
>  	<service-type>mdns</service-type>
>  	<parameters>
>  	 	<parameter>
>  	 	 	<name>mdnsServiceType</name>
>  	 	 	<value>...</value>
> ```
> 
> This is probably a good idea.... Should we change it before this gets adopted in more places like the marketplace?

That is also my thought. Better change it now before release than having to change it later and have a larger impact.
It still requires quite some changes though:
- Changes in core to AddonInfo and the mDNS finder
- Update to the schema and upload the new schema
- PR in add-ons repo to change all mDNS discovery methods

If we wait until after release, we could keep the existing mdnsServiceType for backward compatibility in the code but encourage the use of the parameter syntax.

@andrewfg @kaikreuzer What do you think? This is not an immediate need, but will avoid some hassle in the future? Is there still time to do this change?

J-N-K:
Please do those changes before the release. In general I think this should also increase the version, but @kaikreuzer can probably comment on why we haven't done so in the past.

mherwege:
@andrewfg I will create a PR to adapt core.

andrewfg:
@mherwege Ok I will check your core PR

I will also do a PR for the developer docs

mherwege:
Addon PR also created, changing all mDNS discovery definitions in the addon.xml files: https://github.com/openhab/openhab-addons/pull/16060

holgerfriedrich:
Adapted to openhab/openhab-core#3924, will not work without

kaikreuzer:
>  In general I think this should also increase the version, but @kaikreuzer can probably comment on why we haven't done so in the past.

Afair, we did backward compatible changes in the past and therefore kept the version, so that there wasn't any need to update the xmls in the add-ons (and therefore keep e.g. the ones in the marketplace compatible). With breaking changes, we should probably indeed increase the version. But for https://github.com/openhab/openhab-core/pull/3924, it is imho ok to stick to the version as there hasn't been a release yet.

holgerfriedrich:
@mherwege @andrewfg thanks a lot for your work on generalizing the schema!

I have rebased to current master which contains the new schema.

@mherwege @kaikreuzer 
This is now more or less at the level of a proof of concept. However, I would very much appreciate if I could get it into 4.1 to detect KNX installations. There has been more activity to make the whole chain working: discovery service to detect interfaces in the binding itself, and process suggestion finder which can already identify a running knxd deamon. The only piece missing is this one. 
I know that this somehow needs to be balanced with the approach of only putting mature and already generalized features into core.

Can you give advice what are the most important pieces missing right now? 
- How do I care for the activation/deactivation? I just extended addons.xml in org.openhab.core and added the i18n strings until now.
- Do you think it is important to clean the pattern matching for received packets? Right now it works only with the first received frame and the pattern ".*".

@kaikreuzer What it the time frame this needs to be done? Are we too late for 4.1 already?

kaikreuzer:
@holgerfriedrich I am not a big fan of IP scans and all sorts of side-effects might occur here. I'd therefore prefer to not merge this as a PoC in the very last minute, but rather test it thoroughly after the 4.1 release is done.

holgerfriedrich:
> @holgerfriedrich I am not a big fan of IP scans and all sorts of side-effects might occur here. I'd therefore prefer to not merge this as a PoC in the very last minute, but rather test it thoroughly after the 4.1 release is done.

I get you point.... Though, IP scan sounds a bit more aggressive than it looks to me in this case: A single UDP frame to a well defined multicast addess and port. One frame sent during startup. And we can even disable it.
We are going a big step towards better experience to new users - just wanted to be part of this with KNX as well.

kaikreuzer:
>  IP scan sounds a bit more aggressive than it looks to me in this case: A single UDP frame to a well defined multicast addess and port. 

Ok, understood. Then it is indeed less critical than I thought.

> What it the time frame this needs to be done? 

If @andrewfg and @mherwege approve this PR by tonight, I would be ok to still merge it; if we then notice that there are any issues with it, we could still deactivate it before the release of 4.1.

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -0,0 +1,265 @@
...
+    public static final String SERVICE_TYPE = SERVICE_TYPE_IP;
+    public static final String SERVICE_NAME = SERVICE_NAME_IP;
+
+    private final Logger logger = LoggerFactory.getLogger(IpAddonFinder.class);
+    private final ScheduledExecutorService scheduler = ThreadPoolManager.getScheduledPool(SERVICE_NAME);
+    private @Nullable Future<?> scanJob = null;
+    Set<AddonInfo> suggestions = new HashSet<>();
+
+    public IpAddonFinder() {
+        logger.warn("IpAddonFinder::IpAddonFinder");
```
andrewfg, line 80: Why a warning here?

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    @Deactivate
+    public void deactivate() {
+        logger.warn("IpAddonFinder::deactivate");
+        stopScan();
+    }
+
+    public void setAddonCandidates(List<AddonInfo> candidates) {
+        logger.warn("IpAddonFinder::setAddonCandidates({})", candidates.size());
```
andrewfg, line 90: Ditto

andrewfg (changes requested):
@holgerfriedrich I haven't had time yet to look at the actual logic. However here are a couple of obvious comments.

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -0,0 +1,265 @@
...
+    private final Logger logger = LoggerFactory.getLogger(IpAddonFinder.class);
+    private final ScheduledExecutorService scheduler = ThreadPoolManager.getScheduledPool(SERVICE_NAME);
+    private @Nullable Future<?> scanJob = null;
+    Set<AddonInfo> suggestions = new HashSet<>();
+
+    public IpAddonFinder() {
+        logger.warn("IpAddonFinder::IpAddonFinder");
+    }
+
+    @Deactivate
```
mherwege, line 83: I would expect an @Activate method as well and a regular (scheduled) poll. Remember this should be not just for KNX. The whole finder can be switched off. If it is not switched off, it should poll for all defined ip broadcast finders on a schedule.

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -0,0 +1,265 @@
...
+                Map<String, String> parameters = method.getParameters().stream()
+                        .collect(Collectors.toMap(property -> property.getName(), property -> property.getValue()));
+                Map<String, String> matchProperties = method.getMatchProperties().stream()
+                        .collect(Collectors.toMap(property -> property.getName(), property -> property.getRegex()));
+
+                // parse standard set op parameters:
+                String type = Objects.toString(parameters.get("type"), "");
+                String request = Objects.toString(parameters.get("request"), "");
+                String response = Objects.toString(matchProperties.get("response"), "");
+                int timeoutMs = 0;
```
mherwege, line 133: Maybe use a meaningful default for this if it is not provided, rather than not trying. Define it in a constant.

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -0,0 +1,265 @@
...
+                    logger.info("{}: discovery-parameter dest_port cannot be parsed", candidate.getUID());
+                    continue;
+                }
+
+                //
+                // handle known types
+                //
+                try {
+                    switch (Objects.toString(type)) {
+                        case "ip_multicast":
```
mherwege, line 161: Define in a constant. Also should the type be in camel case?

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -0,0 +1,265 @@
...
+        InetSocketAddress sock = (InetSocketAddress) channel.getLocalAddress();
+
+        ByteArrayOutputStream requestFrame = new ByteArrayOutputStream();
+        StringTokenizer parts = new StringTokenizer(request);
+
+        while (parts.hasMoreTokens()) {
+            String token = parts.nextToken();
+            if (token.startsWith("$")) {
+                switch (token) {
+                    case "$src_ip":
```
mherwege, line 234: Should be a constant, and I would prefer camel case.

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -0,0 +1,265 @@
...
+                        byte[] adr = sock.getAddress().getAddress();
+                        requestFrame.write(adr);
+                        break;
+                    case "$src_port":
```
mherwege, line 238: See above.

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -0,0 +1,265 @@
...
+
+                                    // listen to responses
+                                    Selector selector = Selector.open();
+                                    ByteBuffer buffer = ByteBuffer.wrap(new byte[50]);
+                                    channel.register(selector, SelectionKey.OP_READ);
+                                    selector.select(timeoutMs);
+                                    Iterator<SelectionKey> it = selector.selectedKeys().iterator();
+
+                                    switch (Objects.toString(response)) {
+                                        case ".*":
```
mherwege, line 190: I don't think this can be generalized. Can't you read the buffer, convert to String and do a regex on it?
We could leave this improvement for later when we make more add-ons discoverable through this finder. But it should be improved.

bundles/org.openhab.core/src/main/resources/OH-INF/config/addons.xml
```diff
@@ -29,6 +29,12 @@
 			<description>Use mDNS network scan to suggest add-ons. Enabling/disabling may take up to 1 minute.</description>
 			<default>true</default>
 		</parameter>
+		<parameter name="suggestionFinderIp" type="boolean">
+			<advanced>true</advanced>
+			<label>IP-based Suggestion Finder</label>
+			<description>Use IP network scan to suggest add-ons.</description>
```
mherwege, line 35: Maybe replace 'scan' by 'discovery broadcasts' to make the wording less aggressive.

bundles/org.openhab.core/src/main/resources/OH-INF/i18n/addons.properties
```diff
@@ -2,6 +2,8 @@ system.config.addons.includeIncompatible.label = Include (Potentially) Incompati
 system.config.addons.includeIncompatible.description = Some add-on services may provide add-ons where compatibility with the currently running system is not expected. Enabling this option will include these entries in the list of available add-ons.
 system.config.addons.remote.label = Access Remote Repository
 system.config.addons.remote.description = Defines whether openHAB should access the remote repository for add-on installation.
+system.config.addons.suggestionFinderIp.label = IP-based Suggestion Finder
+system.config.addons.suggestionFinderIp.description = Use IP network scan to suggest add-ons.
```
mherwege, line 6: See above.

mherwege (changes requested):

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
holgerfriedrich: no default needed, just to aviod null for latet parsing

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
holgerfriedrich: I looked at the metadata we use in addons.xml, but CamlCase is also fine if you like

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
holgerfriedrich: It is just the hint that at this point of the frame the current IP needs to be inserted...

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
holgerfriedrich: yes, in principle....

bundles/org.openhab.core/src/main/resources/OH-INF/config/addons.xml
holgerfriedrich: yes, I can change this later

holgerfriedrich:
I think I won't find time today to get this all improved.... We could try to bring it into the milestone to see if it causes problems, and slightly/carefully improve towards release 4.1 build.
@mherwege @andrewfg

holgerfriedrich:
Just a note, I was not able to deactivate any of the finders. I deactivated via UI, restarted, and everything was discovered and shown in the wizard (avm, etc ) - what did I wrong?

mherwege:
> Just a note, I was not able to deactivate any of the finders. I deactivated via UI, restarted, and everything was discovered and shown in the wizard (avm, etc ) - what did I wrong?

Did you wait long enough before restarting? It can take 1 minute for the change to be effective.

mherwege:
And how did you test? This will only have impact in a full install in Karaf, not when running in Eclipse.

holgerfriedrich:
> And how did you test? This will only have impact in a full install in Karaf, not when running in Eclipse.

Probably did not wait long enough, did it in a full distro karaf.

holgerfriedrich:
@mherwege just added string constants for the parameters, still using lowercase with _. I reviewed addons.xml again, the other identifiers are all in lower case. Maybe use "srcip" instead of "src_ip"?

holgerfriedrich:
> > Just a note, I was not able to deactivate any of the finders. I deactivated via UI, restarted, and everything was discovered and shown in the wizard (avm, etc ) - what did I wrong?
> 
> Did you wait long enough before restarting? It can take 1 minute for the change to be effective.

ok, confirmed, did the restart too early.... it works when I wait longer.

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
mherwege: I would prefer camel case for consistency between finders. I called the parameter value for mDNS mdnsServiceType. Field names in xml are snake case, but content of fields is not defined.

holgerfriedrich: ok, now everything is camelCase (with lower case first letter)

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
holgerfriedrich: String constants are used in the current code.
Unknown type is handled in the code below, defaulting to "" is fine. Empty request frame is a valid use case. 
I did not want to check for all possible parameters, as other types than ipMulticast might not use all of them.

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
holgerfriedrich: The long answer to this is.....
Yes of course we can hex dump the received frame and do a regex match. Maybe the regex needs to be dynamic (using $srcIp etc before doing the match).
I would need to change the receive loop, as just taking the first frame and return is no longer possible. I.e. a receive loop, which takes care for the overall timeout and then checking every frame within the time period.

Not sure if this is required for the first implementation, I would probably leave a shortcut for ".*" in anyway as it avoids the regex matching etc.

mherwege:
LGTM with refinements to be done for other IP finder scenarios. But this should not stop making this availalble already.

mherwege (approved):

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
mherwege: I am OK to improve this for 4.2. I hope you will do some work to look at all bindings that use this type of discovery and try to cover more of these. Start by defining criteria, and then see where this core matching needs to ba adapted.

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
mherwege: This should also be addressed in a 4.2 PR. I understand the code to only do a scan on start of the finder or added AddonInfo. I would expect a regular scan. I think this is fine now, but I would like to see it improved in the future.

holgerfriedrich:
> LGTM with refinements to be done for other IP finder scenarios. But this should not stop making this availalble already.

@kaikreuzer ping

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -0,0 +1,279 @@
...
+                // parse standard set op parameters:
+                String type = Objects.toString(parameters.get("type"), "");
+                String request = Objects.toString(parameters.get(PARAMETER_REQUEST), "");
+                String response = Objects.toString(matchProperties.get(MATCH_PROPERTY_RESPONSE), "");
+                int timeoutMs = 0;
+                try {
+                    timeoutMs = Integer.parseInt(Objects.toString(parameters.get(PARAMETER_TIMEOUT_MS)));
+                } catch (NumberFormatException e) {
+                    logger.info("{}: discovery-parameter " + PARAMETER_TIMEOUT_MS + "cannot be parsed",
+                            candidate.getUID());
```
kaikreuzer, line 148:
```suggestion
                    logger.warn("{}: discovery-parameter '{}' cannot be parsed",
                            candidate.getUID(), PARAMETER_TIMEOUT_MS);
```

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -0,0 +1,279 @@
...
+                    continue;
+                }
+                @Nullable
+                InetAddress destIp = null;
+                try {
+                    destIp = InetAddress.getByName(parameters.get(PARAMETER_DEST_IP));
+                } catch (UnknownHostException e) {
+                    logger.info("{}: discovery-parameter " + PARAMETER_DEST_IP + " cannot be parsed",
+                            candidate.getUID());
```
kaikreuzer, line 157:
```suggestion
                    logger.warn("{}: discovery-parameter '{}' cannot be parsed",
                            candidate.getUID(), PARAMETER_DEST_IP);
```

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -0,0 +1,279 @@
...
+                    continue;
+                }
+                int destPort = 0;
+                try {
+                    destPort = Integer.parseInt(Objects.toString(parameters.get(PARAMETER_DEST_PORT)));
+                } catch (NumberFormatException e) {
+                    logger.info("{}: discovery-parameter " + PARAMETER_DEST_PORT + " cannot be parsed",
+                            candidate.getUID());
```
kaikreuzer, line 165:
```suggestion
                    logger.warn("{}: discovery-parameter '{}' cannot be parsed",
                            candidate.getUID(), PARAMETER_DEST_PORT);
```

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -0,0 +1,279 @@
...
+                                                logger.debug("Received return frame from {}",
+                                                        ((InetSocketAddress) source).getAddress().getHostAddress());
+                                                suggestions.add(candidate);
+                                                logger.debug("Suggested add-on found: {}", candidate.getUID());
+                                            } else {
+                                                logger.trace("{}: no response", candidate.getUID());
+                                            }
+                                            break;
+                                        default:
+                                            logger.info("{}: match-property response \"{}\" is unknown",
```
kaikreuzer, line 216:
```suggestion
                                            logger.warn("{}: match-property response \"{}\" is unknown",
```

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -0,0 +1,279 @@
...
+                                    }
+
+                                } catch (IOException e) {
+                                    logger.debug("{}: network error", candidate.getUID(), e);
+                                }
+                            }
+                            break;
+
+                        default:
+                            logger.info("{}: discovery-parameter type \"{}\" is unknown", candidate.getUID(), type);
```
kaikreuzer, line 228:
```suggestion
                            logger.warn("{}: discovery-parameter type \"{}\" is unknown", candidate.getUID(), type);
```

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -0,0 +1,279 @@
...
+                        byte[] adr = sock.getAddress().getAddress();
+                        requestFrame.write(adr);
+                        break;
+                    case "$" + PARAMETER_SRC_PORT:
+                        int dPort = sock.getPort();
+                        requestFrame.write((byte) ((dPort >> 8) & 0xff));
+                        requestFrame.write((byte) (dPort & 0xff));
+                        break;
+                    default:
+                        logger.info("Unknown token in request frame \"{}\"", token);
```
kaikreuzer, line 258:
```suggestion
                        logger.warn("Unknown token in request frame \"{}\"", token);
```

kaikreuzer (changes requested):
Thanks - just some logging improvement suggestions, then we are good to merge.

holgerfriedrich:
> Thanks - just some logging improvement suggestions, then we are good to merge.

done, thanks

kaikreuzer (approved):
Thanks!

stefan-hoehn:
@mherwege I looked into this because currently the Govee binding isn't detected. After doing some research these are the ways to find out if a Govee device is on the network

-  by pinging "ping -c 1 239.255.255.250"
-  or by sending a "{"msg": {"cmd": "scan", "data": {"account_topic": "reserve"}}}" via Multicast ip 239.255.255.250:4001 and receiving back an answer on our port: 4002 (like the binding does).

Do you see any way based on this implementation to auto discover the binding ?

e.g.

``
Note that request is on 4001 but the answer would be on 4002..

<discovery-method>
			<service-type>ip</service-type>
			<discovery-parameters>
				<discovery-parameter>
					<name>type</name>
					<value>ipMulticast</value>
				</discovery-parameter>
				<discovery-parameter>
					<name>destIp</name>
					<value>239.255.255.250</value>
				</discovery-parameter>
				<discovery-parameter>
					<name>destPort</name>
					<value>4001</value>
				</discovery-parameter>
				<discovery-parameter>
					<name>request</name>
					<value>{"msg": {"cmd": "scan", "data": {"account_topic": "reserve"}}}</value>
				</discovery-parameter>
				<discovery-parameter>
					<name>timeoutMs</name>
					<value>5000</value>
				</discovery-parameter>
			</discovery-parameters>
			<match-properties>
				<match-property>
                                         
					<name>response</name>
					<regex>.*</regex>
				</match-property>
			</match-properties>
		</discovery-method>
	</discovery-methods>
``

mherwege:
Not at this point, but see https://github.com/openhab/openhab-core/issues/3936. We are looking to document different cases so we can make sure a more complete solution gets implementeren for 4.2.

stefan-hoehn:
Thanks, let me know when I can try out something for #3936 or support you there.

PR #2519, opened 2021-10-11
Add ser2net mDNS USB serial discovery

wborn:
* Add support for using multiple UsbSerialDiscovery services
* Add Ser2NetUsbSerialDiscovery that can use mDNS to discover ser2net RFC2217 serial ports
* Use discovered USB ports in SerialConfigOptionProvider

mDNS discovery is supported in ser2net 4.3.0 and newer.
E.g. you can install a ser2net version that provides this using APT in Ubuntu 21.04 and Debian 11.

Example ser2net YAML configuration that allows a serial port to be discovered using mDNS discovery:

```
%YAML 1.1
---
connection: &con01
  accepter: telnet(rfc2217),tcp,2222
  connector: serialdev,/dev/ttyUSB0
  options:
    mdns: true
    mdns-sysattrs: true
    mdns-name: devicename
```

Closes #1511

PR #3922, opened 2023-12-13
Add-on suggestion finder for USB devices

andrewfg:
This is a service to suggest USB device related add-ons.

For example Z-Wave, Zigbee, or EnOcean sticks.

---

Note: after merging this PR we also need to add the discovery methods to the respective bindings' `addon.xml` files. For example as follows..

```
// Zigbee
<discovery-methods>
	<discovery-method>
		<service-type>usb</service-type>
		<match-properties>
			<match-property>
				<name>product</name>
				<regex>(?i).*zigbee.*</regex>
			</match-property>
		</match-properties>
	</discovery-method>
</discovery-methods>

// Z-Wave
<discovery-methods>
	<discovery-method>
		<service-type>usb</service-type>
		<match-properties>
			<match-property>
				<name>product</name>
				<regex>(?i).*z-wave.*</regex>
			</match-property>
		</match-properties>
	</discovery-method>
</discovery-methods>

// EnOcean
<discovery-methods>
	<discovery-method>
		<service-type>usb</service-type>
		<match-properties>
			<match-property>
				<name>manufacturer</name>
				<regex>(?i)enocean.*</regex>
			</match-property>
		</match-properties>
	</discovery-method>
</discovery-methods>
```


andrewfg:
@mherwege this is a first draft of a finder for USB device related bindings. => And comments?

andrewfg:
Obviously `javax.usb` is not part of OH core, so karaf complains. I need to figure out how to import that module.

mherwege:
> Obviously `javax.usb` is not part of OH core, so karaf complains. I need to figure out how to import that module.

The knx binding has it as an (optional) dependency, probably transitive because I don’t see it appear in the code directly.
`javax.usb` is just the api, not the implementation. There are some concept implementations for it, but not covering all platforms (https://javax-usb.sourceforge.net/). I believe the most common implementation is http://usb4java.org/index.html (last updated in 2018). Another one I found is https://github.com/KeyBridge/lib-javax-usb3.
You can try if you get it working with any of these.

Be aware there may be plaform issues. E.g. Linux may need write permissions to the device, Windows would need a device driver.
To my knowledge none of the add-ons uses USB. They all use serial and USB sets up a gateway to a serial port. We need to validate that this can discover these devices and will not interfere with operations.

andrewfg:
@mherwege OK I understand. The serial port stuff is useless because serial ports a) have no data about the source, and b) would need to be  configured before they could be discovered. I tried to go the USB route because 1) who does serial by any other means than USB these days?, and 2) the stick can report its product, manufacturer, and serial number data (which could be used to check for matches before the serial is actually set up).

mherwege:
@andrewfg And your idea may very well work. But it will need a lot of testing. It would be useful to find a list of USB product an manufacurer characteristics to math on so not all sticks for zwave and zigbee that can be used with OH need to be tried individually to find the match criteria.

andrewfg:
> would be useful to find a list of USB product an manufacurer characteristics to math on so not all sticks for zwave and zigbee that can be used with OH need to be tried individually to find the match criteria

AFAICT the usb devices only support very limited properties — namely manufacturer, product id, and serial number; so probably in the first step we would just suggest on the manufacturer name. If we need to filter on more specific data, we could add that later.

@mherwege further to my above post, I realise that the core usbSerial discovery classes do have a strong basis for the discovery part as well as the serial part; so I think I can benefit from that after all.

More tomorrow..

mherwege:
> I realise that the core usbSerial discovery classes do have a strong basis for the discovery part as well as the serial part; so I think I can benefit from that after all.

I had a look at it as well. As far as I can understand, this will only work on linux. Embedding one of the libraries I mentioned could potentially make it work for multiple platforms. Apart from that, I think what is there in the code will also have the same constraints I cited above on linux (device must have write permission). This can be documented, en in openhabian it could be made part of the setup scripts.

andrewfg:
> As far as I can understand, this will only work on linux.

Indeed. So I am going back to plan A (`javax.usb`)

I found the following [post](https://community.openhab.org/t/3rd-party-maven-dependencies-in-openhab-core/113722/2) in the community, which may help me to import the java4usb module without causing Karaf issues. It seems that this technique has been used in `o.o.c.audio` and `o.o.c.io.transport.modbus` ..

mherwege:
Here is the original PR that introduced the functionality in core: https://github.com/eclipse-archived/smarthome/pull/5315

I wonder if it would make sense to work on the core functionality to improve platform support. I am not sure where the core functionality is currently used.

For the dependency: I would make this a feature (like mDNS and UPnP finders) that can be installed/uninstalled. In that case, you should be able to add this dependency to the feature. It would only be installed when the feature is enabled. Wouldn't that be enough?

mherwege:
Did a search, I can only find the Enocean binding as a consumer of the usb discovery service.

wborn:
The issue with most USB sticks is that they all use the same FTDI chips which results in discovery false positives. :-/

wborn:
It would also be better to reuse the existing USB discovery mechanism instad of creating a new one. That way you can also suggest add-ons based on ser2net discovery results (https://github.com/openhab/openhab-core/pull/2519).

andrewfg:
@mherwege today's latest commit is now able to find the USB devices (at least the six devices on my Windows system .. listed below).

```
vendorId:0bc2 (vendorName:null) / productId:2322 (productName:null)
vendorId:0a5c (vendorName:null) / productId:21e8 (productName:null)
vendorId:047d (vendorName:null) / productId:00f2 (productName:null)
vendorId:045e (vendorName:null) / productId:082a (productName:null)
vendorId:045e (vendorName:null) / productId:00db (productName:null)
vendorId:046d (vendorName:null) / productId:0826 (productName:null)
```
As you can see, the issue is that the device data returns vendor id codes and product id codes, rather than strings. The vendor id string is available via online lookup (and in the attached pdf). It seems that only certain usb devices include the strings within the device itself. 

[vendor_ids051920_0.pdf](https://github.com/openhab/openhab-core/files/13676516/vendor_ids051920_0.pdf)

> issue with most USB sticks is that they all use the same FTDI chips

Sorry but I don't understand.

> the Enocean binding as a consumer of the usb discovery service

I did look into that. .. ..

andrewfg:
> It would also be better to reuse the existing USB discovery mechanism instead of creating a new one

@wborn it would be nice to do that. But apparently it only works on Linux? And AFAICT it does not read the USB stick vendor name and product name. So it would not be much use in determining if the stick supports a certain functionality, or not. => But if you think that is not the case, please let me know.

wborn:
> Sorry but I don't understand.

https://github.com/openhab/openhab-addons/issues/4421

> But apparently it only works on Linux?

There can be multiple `UsbSerialDiscovery` implementations so you can add another one like linuxsysfs that works on Windows or other platforms.

> it does not read the USB stick vendor name and product name

Does it really read it from the device or does it have some internal database to return names for certain vendor/product IDs?

andrewfg:
> does it have some internal database to return names for certain vendor/product IDs?

I dunno. Still learning.

andrewfg:
> would also be better to reuse the existing USB discovery mechanism instead of creating a new one

> There can be multiple UsbSerialDiscovery implementations so you can add another one

Ok. I think I could mis-use the existing disco mechanism. And add a finder for windows too. However the existing disco is for creating things, so the mis-use would be to not create a thing, but that method could create a suggestion instead.

andrewfg:
@mherwege Aargh!! I really don't understand how Karaf works. The documentation is really rubbish! So I wonder if you can kindly help me to eliminate the current CI build error due to the following two dependencies that this finder requires?

```
    <dependency>
      <groupId>javax.usb</groupId>
      <artifactId>usb-api</artifactId>
      <version>1.0.2</version>
    </dependency>
    <dependency>
      <groupId>org.usb4java</groupId>
      <artifactId>usb4java-javax</artifactId>
      <version>1.3.0</version>
    </dependency>
```

mherwege:
I often struggle with this as well. I don't think the problem is with your pom file. The error is in the karaf feature verification:
```
[ERROR] Failed to execute goal org.apache.karaf.tooling:karaf-maven-plugin:4.4.4:verify (karaf-feature-verification) on project org.openhab.core.features.karaf.openhab-core: Feature resolution failed for [openhab-core-config-discovery-addon-usb/4.1.0.SNAPSHOT]
[ERROR] Message: Unable to resolve root: missing requirement [root] osgi.identity; osgi.identity=openhab-core-config-discovery-addon-usb; type=karaf.feature; version=4.1.0.SNAPSHOT; filter:="(&(osgi.identity=openhab-core-config-discovery-addon-usb)(type=karaf.feature)(version>=4.1.0.SNAPSHOT))" [caused by: Unable to resolve openhab-core-config-discovery-addon-usb/4.1.0.SNAPSHOT: missing requirement [openhab-core-config-discovery-addon-usb/4.1.0.SNAPSHOT] osgi.identity; osgi.identity=org.openhab.core.config.discovery.addon.usb; type=osgi.bundle; version="[4.1.0.202312141745,4.1.0.202312141745]"; resolution:=mandatory [caused by: Unable to resolve org.openhab.core.config.discovery.addon.usb/4.1.0.202312141745: missing requirement [org.openhab.core.config.discovery.addon.usb/4.1.0.202312141745] osgi.wiring.package; filter:="(osgi.wiring.package=org.usb4java)"]]
```
So it must be somewhere in the feature definition, where it cannot find `org.usb4java`. That's why it will work in Eclipse (where you manually include these bundles in your bndrun file) and not in the distro with karaf.

features/karaf/openhab-core/src/main/feature/feature.xml
```diff
@@ -91,6 +91,12 @@
 		<feature dependency="true">openhab.tp-jupnp</feature>
 	</feature>
 
+	<feature name="openhab-core-config-discovery-addon-usb" version="${project.version}">
+		<feature>openhab-core-base</feature>
+		<feature>openhab-core-config-discovery-addon</feature>
+		<bundle>mvn:org.openhab.core.bundles/org.openhab.core.config.discovery.addon.usb/${project.version}</bundle>
```
mherwege, line 97: Add a reference to the usb4java mvn bundle here. I think that will solve your feature verification issue.

I actually don't think you need all the special code in your pom, just add usb4java and javax.usb as dependencies.

mherwege:
> Does it really read it from the device or does it have some internal database to return names for certain vendor/product IDs?

I don't think it has an internal DB. The usb4java (http://usb4java.org/index.html) library uses libusb (https://libusb.info/) underneath. This is a native libary that talks to the USB devices, written in C. Multiple platform versions are packaged with usb4java. Looking at teh API for libusb, it has methods to retrieve that info from the device, and I assume that's also what is being used in usb4java.

mherwege:
I think trying to detect USB serial devices may be opening a can of worms here. I think some agreement needs to be reached on path forward. Here are a few observations:

### Current USB serial discovery

- As it stands, we have an API for thing discovery in core. This is installed by default. It only has an implementation that works for linux and ser2net (which itself will detect serial interfaces in the network based on mDNS, so a 'server' component needs to be installed on the machine with the actual serial interface).
- The API offers thing discovery methods in the context of thing discovery. It doesn't have methods that fit the add-on finders at the moment. This can of course be extended.
- The linux detection is based on sysfs, so using the linux file system to detect devices.
- There is no easy mechanism to detect USB serial devices on Windows, which I assume would be the other platform that should be supported as a minimum. A google search did not reveal much that could be used directly.
- I don't know if the current mechanism for linux would also work for Mac.
- The `linuxsysfs` discovery is available and running on any platform but will identify when it cannot perform a scan and not do anything.
- Only the enocean binding uses this for thing discovery.

### javax.usb

- `javax.usb` (https://javax-usb.sourceforge.net/) as an interface (and concept implementation for linux) for USB device management. It is a standard, so could be used, but the included implementations are old and not maintained.
- So using the interface is relevant, but I don't think we should use any of the provided implementations.

### usb4java

- `usb4java` (http://usb4java.org/index.html#) is a library wrapping a native library, libusb (https://libusb.info/), for USB device management and access. It implements its own interface (direct wrapper around the native library) and a `javax.usb` compliant implementation.
- It wraps the native libraries for linux, Windows and Mac on a number of architectures.
- It is available as a maven package.
- New USB serial discover implementations could be created using this library. However, it may lead to conflicts with the existing `linuxsysfs` discovery as both would discover the same device.
- We are not sure we can get relevant vendor and product info using this library yet.
- The libary does not look to be maintained much (last changes seem to be from 5 years ago). Documentation is good though. All source code is available.
- There already are forks, some with continuous development. But I haven't seen one with a maven package. Soo for example: https://github.com/KeyBridge/lib-javax-usb3
- The `libusb` native libraries do have a better maintainance track record. The latest versions of this are not included in `usb4java`. It may be a simple drop-in replacement, but it would mean we need to start maintaining a fork.
- `usb4java` uses Apache commons (and it seems to be for only 2 calls), which increases the package size a lot. We could probably eliminate it, but again, it would mean a fork to maintain ourselves.

I don't see a perfect solution here at the moment. But before going too far in this, I would like to see opinions and consensus being developed. Otherwise we risk spending a lot of time and not getting anywhere.

@andrewfg @wborn @kaikreuzer  What are your views on this?

mherwege:
> And AFAICT it does not read the USB stick vendor name and product name.

From the code, I think it does (or at least it tries to find it): https://github.com/openhab/openhab-core/blob/4b92db3775bf011d051963c56639c954d6d7e31f/bundles/org.openhab.core.config.discovery.usbserial.linuxsysfs/src/main/java/org/openhab/core/config/discovery/usbserial/linuxsysfs/internal/SysfsUsbSerialScanner.java#L247

andrewfg:
@mherwege thanks for your many lengthy posts. Just a few short answers..

1. The existing OH USB discovery comprises two sorts of classes. Namely several scanners which parse the USB device tree and read their data. And a discovery service that uses inputs from those scanners, and calls discovery participants with that data.
2. OH currently does not have a scanner for Windows.
3. It is possible that I could write such a scanner for Windows, that would plug into the discovery service above. But it might have fewer features than the other scanners. However it would provide some data anyway.
4. It is possible that I can write an addon finder service that uses the above mentioned existing and new scanners. But is different than the above mentioned disco service resp disco participants.
5. So in summary I would create two modules, one of which would enhance existing USB thing discovery on other platforms, and the other which would use the scanners to find potential addons.
6. I hope that today I will find a way to get my current PR to build in Karaf, and when that works I will push a new commit with the abovementioned two modules.

mherwege:
@andrewfg Your understanding is correct.

features/karaf/openhab-core/src/main/feature/feature.xml
andrewfg: Ok. I fixed it.

andrewfg:
> Your understanding is correct.

@mherwege based on 'my understanding' I split into two classes..
1. A new USB scanner `org.openhab.core.config.discovery.usbserial.otheros` for OS other than Linux
2. A new add-on suggestion finder `org.openhab.core.config.discovery.addon.usb`

I found out that in some cases the new `org.openhab.core.config.discovery.usbserial.otheros` scanner can find the USB devices, but not read the `manufacturer` and `product` description strings. There are various reasons for this. However it does always return the `vendorId` and `productId`. So for this reason I have added a local database to lookup on `vendorId:productId` pairs and provide a (backup) local `product` description.

andrewfg:
Please see the following..

- https://github.com/openhab/org.openhab.binding.zwave/issues/1906
- https://github.com/openhab/org.openhab.binding.zigbee/issues/818

andrewfg:
https://community.openhab.org/t/please-contribute-to-database-of-z-wave-zigbee-usb-sticks/152087

andrewfg:
image

bundles/org.openhab.core.config.discovery.usbserial.otheros/src/main/java/org/openhab/core/config/discovery/usbserial/otheros/internal/UsbProductDatabase.java
```diff
@@ -0,0 +1,76 @@
...
+ * Please do NOT include supported binding id in the case of very commonly used generic serial chips, since this may
+ * lead to false-positive discoveries.
+ *
+ * @author Andrew Fiddian-Green - Initial contribution
+ */
+@NonNullByDefault
+public class UsbProductDatabase {
+
+    // @formatter:off
+    public static final Map<String, String> PRODUCT_NAMES = Map.of(
```
cdjackson, line 44: Wouldn't it be better to move these definitions into a separate resource file rather than embedding them in Java?

cdjackson, line 44: I've not looked through the code to see how everything is used, but I wonder if using `String` here for both the definition of the USB VID:PID and also the name of the device, and I guess the binding name is encoded here somewhere is the right approach? Wouldn't it be better to define a class that breaks this information out and also allows a descriptive name (eg Silicon labs) and the binding name (eg ZigBee) to be defined, and used, independently?

andrewfg: IDK. A file is a file. Whether or not its name ends in .java or .json or .xml or .txt .. Or?? The advantage of a .java file is that a) it can be documented via javadoc, and b) requires no extra overhead to parse. Ultimately the data is consumed in the java finder anyway..

andrewfg: Good idea. I could change that. However before reengineering such fine details, perhaps we need to consider the broader question of whether you guys even agree to the concept at all?

cdjackson: Well, based on that argument, we can get rid of all the XML files that describe things, bindings, channels etc, and just code everything into the java source?

The advantage of separating it out (IMHO) is maintainability. Having a defined file that allows non-coders to add data without getting into Java is (again, IMHO) better and cleaner. Yes, it requires some work to parse a new file, but that's life - we could have saved loads of work on OH core by just coding all definitions directly in the source...

andrewfg: Ok.

andrewfg: But just also remember the other way round case too. The user has an OH system, he the buys a stick, plugs it in, and then looks for an addon to support it, and voila OH recommends the right binding for it.

andrewfg: But don't get me wrong. I won't waste any more time developing something that nobody wants. I started the eork because I was told that people do want it.

andrewfg: BTW I think there may be value in splitting this PR into two parts..
1. The USB device scanner for non Linux systems. Which may have limited functionality vs Linux. But more functionality than none at all. You may debate if that makes sense.
2. And add-on suggestion finder for addons that are related to specific USB devices (no matter how they may be discovered). You may debate if that adds value.

The advantage of that approach is that the two paths can be reviewed and progressed indepently. => WDYT?

cdjackson: > I won't waste any more time developing something that nobody wants.

Again - this is just my view - others may think differently... To me, on balance it doesn't seem like it adds a lot over just searching the addons for zwave (or whatever). OH requires a lot of configuration, and to me this doesn't appreciably reduce that burden - but maybe for the new user there's still a use case there... Hopefully others will chime in with more thoughts :)

holgerfriedrich:
@andrewfg I like the idea of checking connected USB devices! The big benefit in all this activities is to give a better experience to new users. Experienced users may find a suggestion supporting one of their devices they have not yet thought about.

As a developer of a binding, I would like to do a regex match in all lines of the output generated by ```lsusb```. This would allow matching IDs and/or descriptions.
If I get it correctly, the device description is not necessarily available.

However, building up a database is a lot of work and will require continuous efforts to adapt to new hardware.
Here are links to a few databases:
https://wiki.debian.org/DeviceDatabase/USB
https://github.com/systemd/systemd/blob/main/hwdb.d/20-usb-vendor-model.hwdb
http://www.linux-usb.org/usb.ids

bundles/org.openhab.core.config.discovery.usbserial.otheros/src/main/java/org/openhab/core/config/discovery/usbserial/otheros/internal/OtherOSUsbSerialDiscovery.java
```diff
@@ -0,0 +1,221 @@
...
+                short vendorId = d.idVendor();
+                short productId = d.idProduct();
+
+                String manufacturer = null;
+                String product = null;
+                String serialNumber = null;
+
+                /*
+                 * Note: the getString() calls below may fail depending on the Operating System:
+                 * - on Windows if no libusb device driver is installed for the device.
```
mherwege, line 174: I don't think this is correct. Without libusb you get nothing (even no vendorId or productId). But libusb is packaged as a native library with usb4java, so should be there on a number of platforms.

mherwege: Reading a bit further, there is a need for a driver for the device in Windows for libusb to be able to work with the device. And that may require installation if not installed already.

bundles/org.openhab.core.config.discovery.usbserial.otheros/src/main/java/org/openhab/core/config/discovery/usbserial/otheros/internal/OtherOSUsbSerialDiscovery.java
```diff
@@ -0,0 +1,221 @@
...
+                } catch (UnsupportedEncodingException | UsbDisconnectedException | UsbException e) {
+                    // ignore because this would be a 'normal' runtime failure
+                }
+
+                if (product == null || product.isBlank()) {
+                    product = UsbProductDatabase.getProduct(vendorId, productId);
+                }
+
+                /*
+                 * TODO are the following three data required (??)
```
mherwege, line 190: It is probably needed if used beyond the finder. My understanding is the serial port is at the interface level, not the device level. And there can be multiple interfaces in a configuration in a device.

mherwege:
@andrewfg I am not at all convinced you should try to build a database. If something can be discovered, you will always have vendorId and productId. So the regex in the finder criteria can match on that.
In developer documentation, we should refer to one of the online databases, indicate matching can be done on manufacturer and product as well, but many sticks don't provide it. As we can define multiple alternative match criteria, we can have matching with manufacturer and product if it is needed for differentiation. If it is not provided we cannot differentiate anyway. If the device is too generic, I wouldn't try matching.

bundles/org.openhab.core.config.discovery.usbserial.otheros/src/main/java/org/openhab/core/config/discovery/usbserial/otheros/internal/OtherOSUsbSerialDiscovery.java
andrewfg: It always finds vendorId and productId .. I tested it.

bundles/org.openhab.core.config.discovery.usbserial.otheros/src/main/java/org/openhab/core/config/discovery/usbserial/otheros/internal/OtherOSUsbSerialDiscovery.java
andrewfg: Yeah. But I think device level is sufficient for making addon suggestions.

andrewfg:
> So the regex in the finder criteria can match on that.

@mherwege this is true. But to write the regex you need to know the vendorId and productId of supported devices for each addon. IMHO that knowledge is (a kind of) database. We need the knowledge anyway, no matter what form it takes.

mherwege:
> IMHO that knowledge is (a kind of) database

True, but rather than developing the database ourselves, why not point to one of the online databases in the documentation for this feature? I would leave it to binding developers to do that.

mherwege:
> please let me know if this table has any errors.

For otheros on linux, my understanding is that it is not about read rights, but write rights. Or am I missing something?

andrewfg:
> not about read rights, but write rights

I was reading their javadocs, like you. But logically I think it is about reading information from a stick rather than writing to it .. or?

mherwege:
I would call the new UsbSerialDiscovery implementation somthing like 'javaxusb', rather than 'otheros'. It can discover for linux as well, and what does 'other' mean?

bundles/org.openhab.core.config.discovery.usbserial.otheros/src/main/java/org/openhab/core/config/discovery/usbserial/otheros/internal/OtherOSUsbSerialDiscovery.java
```diff
@@ -0,0 +1,221 @@
...
+    private final boolean isLinux;
+    private final @Nullable ScheduledExecutorService scheduler;
+
+    private @Nullable ScheduledFuture<?> scanTask;
+    private Set<UsbSerialDeviceInformation> lastScanResult = new HashSet<>();
+
+    @Activate
+    public OtherOSUsbSerialDiscovery() {
+        isLinux = System.getProperty("os.name", "unknown").toLowerCase().contains("linux");
+        scheduler = isLinux ? null
```
mherwege, line 73: Can we not create a mechanism in the service to merge results from different serial discovery services instead, and allow it to run on linux as well?
This may increase the chance of finding all info.

mherwege:
As it stands right now, the otheros discovery is started at system start. As I wrote before, it is quite a large library. And so far the use is only for the add-on finders.
The serial discovery itself is only used by the Enocean binding, and it looks like it is not very good (too often using a general UART chip).
I would think we should try to wrap this in a feature as well and start/stop the feature as part of the usb add-on discovery feature to avoid this overhead if someone does not need it.

mherwege:
> I was reading their javadocs, like you. But logically I think it is about reading information from a stick rather than writing to it .. or?

I have no idea. And I don't have any stick connected to my RPi, so cannot test.

bundles/org.openhab.core.config.discovery.usbserial.otheros/src/main/java/org/openhab/core/config/discovery/usbserial/otheros/internal/OtherOSUsbSerialDiscovery.java
mherwege: Agree, but if we make it available as a serial discovery, it needs all to be able to be used outside of the finder.

mherwege: This of course leads to the next question. The serial discovery classes focus on finding USB devices with serial interfaces. The finder, in theory could look broader than that (although I don't know if there is any current use). E.g. what about a camera or speaker directly connected through USB to the server. Most likely, discovery would be better done by detecting the process in that case, but in theory it could also be done by detecting the configurations and interfaces on the USB device. But then it wouldn't fit in this class anymore.

mherwege:
I am also wondering about ser2net. E.g. a stick with filter criteria defined in the ZWave binding could be discovered through ser2net. But to my knowledge, the ZWave binding does not currently support ser2net. How do we make sure to filter out these results?

andrewfg:
> call the new UsbSerialDiscovery implementation somthing like 'javaxusb

Ok.

bundles/org.openhab.core.config.discovery.usbserial.otheros/src/main/java/org/openhab/core/config/discovery/usbserial/otheros/internal/OtherOSUsbSerialDiscovery.java
andrewfg: > merge results from different serial discovery services instead

Yeah. I had a sleepless night last night and I was coding that in my head at 3am

bundles/org.openhab.core.config.discovery.usbserial.otheros/src/main/java/org/openhab/core/config/discovery/usbserial/otheros/internal/OtherOSUsbSerialDiscovery.java
andrewfg: Indeed.

andrewfg:
> How do we make sure to filter out these results?

I guess we could use one of the newly defined discovery method configuration parameters. Or something like that..

mherwege:
> I guess we could use one of the newly defined discovery method configuration parameters. Or something like that..

Sure, but at the moment you do not see if the device is local or remote in the serial device discovery result. So that info may have to be added (default local, with an override possible).

andrewfg:
@mherwege @holgerfriedrich @cdjackson the Javax-Usb UsbSerialDiscovery has been split into a separate PR #3930 -- this means that we can discuss the AddonFinder service in this PR and the UsbSerialDiscovery service in its own PR separately.

Note: this AddonFinder service can consume results from any or all UsbSerialDiscovery services -- so to be specific -- it is independent of the Javax-Usb UsbSerialDiscovery.

bundles/org.openhab.core.config.discovery.usbserial.otheros/src/main/java/org/openhab/core/config/discovery/usbserial/otheros/internal/OtherOSUsbSerialDiscovery.java
andrewfg: Done

bundles/org.openhab.core.config.discovery.usbserial.otheros/src/main/java/org/openhab/core/config/discovery/usbserial/otheros/internal/UsbProductDatabase.java
andrewfg: => continue discussion in #3930

andrewfg: > wonder if using String here for both the definition of the USB VID:PID and also the name of the device
> define a class that breaks this information out and also allows a descriptive name (eg Silicon labs) and the binding name (eg ZigBee) to be defined, and used, independently

=> continue discussion on these topics in #3930

andrewfg:
> I would think we should try to wrap this in a feature as well and start/stop the feature as part of the usb add-on discovery feature to avoid this overhead if someone does not need it.

@mherwege do you want to do that in this AddonFinder PR or in the UsbSerialDiscovery PR? Importnat is to remember that if an add-on IS suggested (and installed) then the UsbSerialDiscovery should not be disabled, since otherwise the bridge thing would not be discovered thereafter.

andrewfg:
building up a database is a lot of work and will require continuous efforts7

=> @holgerfriedrich continue [here](https://github.com/openhab/openhab-core/pull/3930#issuecomment-1860547865)

mherwege:
> @mherwege do you want to do that in this AddonFinder PR or in the UsbSerialDiscovery PR? Important is to remember that if an add-on IS suggested (and installed) then the UsbSerialDiscovery should not be disabled, since otherwise the bridge thing would not be discovered thereafter.

I think we should focus on add-on suggestions. For ZWave and Zigbee, there is no bridge thing discovery. So if it would get discovered, a user would still have to create the bridge manually. I believe there was discovery for at least one of them at some point in time, but it got removed, as not reliable enough. The suggestion finder does not have that issue. It does not have to be perfect. If it is good enough, the rest can be left to the user.
Enocean is the only binding using discovery for a serial bridge, but my understanding is it is not working reliably. So again, same conclusion.

Ultimately, if this feature is useful to be enabled by default, that can always be done later by adding it to the boot features. It would then simply not be de-installed anymore when the usb finder feature gets de-installed (because there would be 2 consumers for the feature).

andrewfg:
@mherwege this PR had been marked as 'wip' but in fact it is now ready for review. Note: it works with any USB scanner, such as the three the existing OH ones, and it is quite independent of #3930 or #3934

bundles/org.openhab.core.config.discovery.addon.usb/src/main/java/org/openhab/core/config/discovery/addon/usb/UsbAddonFinder.java
```diff
@@ -0,0 +1,194 @@
...
+    private final Map<Integer, UsbSerialDeviceInformation> usbDeviceInformations = new ConcurrentHashMap<>();
+
+    @Activate
+    public UsbAddonFinder() {
+    }
+
+    @Deactivate
+    public void deactivate() {
+    }
+
```
wborn, line 76: This seems redunant: 

```suggestion
```

bundles/org.openhab.core.config.discovery.addon.usb/src/main/java/org/openhab/core/config/discovery/addon/usb/UsbAddonFinder.java
```diff
@@ -0,0 +1,194 @@
...
+    @Override
+    public void setAddonCandidates(List<AddonInfo> candidates) {
+        super.setAddonCandidates(candidates);
+    }
+
+    @Override
+    public void unsetAddonCandidates() {
+        super.unsetAddonCandidates();
+    }
+
```
wborn, line 98: This too:

```suggestion
```

bundles/org.openhab.core.config.discovery.addon.usb/src/test/java/org/openhab/core/config/discovery/addon/usb/UsbAddonFinderTests.java
```diff
@@ -0,0 +1,101 @@
...
+import org.openhab.core.addon.AddonInfo;
+import org.openhab.core.addon.AddonMatchProperty;
+import org.openhab.core.config.discovery.usbserial.UsbSerialDeviceInformation;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import ch.qos.logback.classic.Level;
+
+/**
+ * This contains Junit tests for the {@link UsbAddonFinder} class.
```
wborn, line 37: It's stylized as "JUnit" but just remove it because it is obvious:

```suggestion
 * This contains tests for the {@link UsbAddonFinder} class.
```

bundles/org.openhab.core/src/main/resources/OH-INF/i18n/addons.properties
```diff
@@ -8,5 +8,7 @@ system.config.addons.suggestionFinderMdns.label = mDNS Suggestion Finder
 system.config.addons.suggestionFinderMdns.description = Use mDNS network scan to suggest add-ons. Enabling/disabling may take up to 1 minute.
 system.config.addons.suggestionFinderUpnp.label = UPnP Suggestion Finder
 system.config.addons.suggestionFinderUpnp.description = Use UPnP network scan to suggest add-ons. Enabling/disabling may take up to 1 minute.
+system.config.addons.suggestionFinderUsb.label = USB Suggestion Finder
+system.config.addons.suggestionFinderUsb.description = Scan connected USB devices to suggest add-ons. Enabling/disabling may take up to 1 minute.
```
wborn, line 12: This texts differs from the text in the XML because it is missing "the".

wborn (changes requested):
It looks good to me :+1: except for these few comments below :slightly_smiling_face: :

bundles/org.openhab.core/src/main/resources/OH-INF/config/addons.xml
```diff
@@ -17,6 +17,12 @@
 				expected. Enabling this option will include these entries in the list of available add-ons.</description>
 			<default>false</default>
 		</parameter>
+		<parameter name="suggestionFinderIp" type="boolean">
```
mherwege, line 20: This doesn’t belong in this PR as a change.

andrewfg: > You should rebase.

Yeah. This PR got horribly out of synch with main, due to the rapid other PRs prior to v4.1, so I am going to wait until it is finished, and then try to squash/flatten it. (If I can remember to git commands to do that).

mherwege:
I see two things that may need some further thougt:

1. The merge mechanism does not distinguish between the serial providers. How do we make sure we get manufacturer and product data from the device when one of the providers can read it, and not one from the internal database?
2. We need a mechanism to exclude ser2net as suggestions for specific bindings (zwave and zigbee bindings do not support it).

bundles/org.openhab.core/src/main/resources/OH-INF/config/addons.xml
mherwege: A rebase on main should eliminate this part. The issue is that you try merging in main.

andrewfg:
@mherwege I guess filtering on a particular finder could be a match property regex, (or possibly a parameter), but probably more likely the former.

mherwege:
@andrewfg The USBSerialDeviceInformation class does not contain info on the serial provider that found it. You should never merge info of a local serial device with info from a ser2net device, as it may be 2 different devices. Therefore, I would suggest introducing a boolean field ‘remote’ in the class (default false). Consider it when merging to not merge local and remote devices and make it part of the key. That also makes filtering easier and does not require to explicitly exclude one provider. It is only data based.

andrewfg:
I need to change from having vendorId and productId as separate properties, to having a combined chipId property..

mherwege:
> I need to change from having vendorId and productId as separate properties, to having a combined chipId property..

What was the reason for this? After all, you could do that already by providing both in the discovery parameters.

andrewfg:
> What was the reason for this?

A correct match is on an AND of vendorId and productId. If you look at the following Zigbee example you can see for your self how difficult it would be to code match properties for OR of multiple stick types each having an AND..

https://github.com/andrewfg/org.openhab.binding.zigbee/edit/suggestion-finder/org.openhab.binding.zigbee/src/main/resources/OH-INF/addon/addon.xml?pr=%2Fopenhab%2Forg.openhab.binding.zigbee%2Fpull%2F819

bundles/org.openhab.core.config.discovery.addon.usb/src/main/java/org/openhab/core/config/discovery/addon/usb/UsbAddonFinder.java
```diff
@@ -0,0 +1,186 @@
...
+                }
+
+                logger.trace("Checking candidate: {}", candidate.getUID());
+                for (UsbSerialDeviceInformation device : usbDeviceInformations.values()) {
+                    logger.trace("Checking device: {}", device);
+
+                    if (propertyMatches(matchProperties, PRODUCT, device.getProduct())
+                            && propertyMatches(matchProperties, MANUFACTURER, device.getManufacturer())
+                            && propertyMatches(matchProperties, CHIP_ID,
+                                    toHexString(device.getVendorId(), device.getProductId()))
```
wborn, line 110: Does this now also allows for supporting addon suggestions by other `UsbSerialDiscovery` implementations like ser2net (https://github.com/openhab/openhab-core/pull/3930#issuecomment-1866295251)?

bundles/org.openhab.core.config.discovery.addon.usb/src/main/java/org/openhab/core/config/discovery/addon/usb/UsbAddonFinder.java
```diff
@@ -0,0 +1,186 @@
...
+                        logger.debug("Suggested add-on found: {}", candidate.getUID());
+                        break;
+                    }
+                }
+            }
+        }
+        return result;
+    }
+
+    private String toHexString(int vendorId, int productId) {
```
wborn, line 122: It's no longer just a hex string.

```suggestion
    private String toChipId(int vendorId, int productId) {
```

bundles/org.openhab.core.config.discovery.addon.usb/src/main/java/org/openhab/core/config/discovery/addon/usb/UsbAddonFinder.java
```diff
@@ -0,0 +1,186 @@
...
+        return String.format("0x%04x:0x%04x", vendorId, productId);
```
wborn, line 123: It's more common to see these string in output of tools like lsusb without these `0x` prefixes.

```suggestion
        return String.format("%04x:%04x", vendorId, productId);
```

bundles/org.openhab.core.config.discovery.addon.usb/src/main/java/org/openhab/core/config/discovery/addon/usb/UsbAddonFinder.java
andrewfg: Indeed. This finder accepts discovered USB devices from all installed 'UsbSerialDiscovery' implementations. Currently that means the following..
- linuxsyfs (works on Linux only)
- ser2net (works for ip connected devices)
- windowsregistry [wip] (works on Windows only)
- javaxusb [wip] (works on all platforms, but with limited data i.e. only chipId)

Notes:
1. As @mherwege requested I added another 'remote' property to the UsbSerialDiscovery implementations, so that the suggestion finder can (say) reject a remotely (ser2net) device via the match-properties. See ZigBee example below..
2. This suggestion finder will process inputs from all UsbSerialDiscovery implementations; so on a Linux system (say) you will currently get duplicated USB device lists from linuxsyfs and javaxusb. However we could in principle suppress javaxusb on Linux resp. Windows where the other UsbSerialDiscovery implementation would provide more complete data.

Apropos https://github.com/openhab/openhab-core/pull/3930#issuecomment-1866295251 please look at the match-properties for (say) the ZigBee binding [here](https://github.com/openhab/org.openhab.binding.zigbee/blob/360000619924b4190b575fc25545220f19fbdb55/org.openhab.binding.zigbee/src/main/resources/OH-INF/addon/addon.xml) where you can see that we have two match-properties .. one for matching device description strings, and the other for matching chipIds.

bundles/org.openhab.core.config.discovery.addon.usb/src/main/java/org/openhab/core/config/discovery/addon/usb/UsbAddonFinder.java
andrewfg: Ok

bundles/org.openhab.core.config.discovery.addon.usb/src/main/java/org/openhab/core/config/discovery/addon/usb/UsbAddonFinder.java
andrewfg: Ok

bundles/org.openhab.core.config.discovery.addon.usb/src/main/java/org/openhab/core/config/discovery/addon/usb/UsbAddonFinder.java
wborn: > so that the suggestion finder can (say) reject a remote (ser2net) device via the match-properties. See ZigBee example below.

Are you sure it's still unsupported? There was https://github.com/openhab/org.openhab.binding.zigbee/pull/661.

andrewfg: > Are you sure it's still unsupported? 

I am not sure at all .. @mherwege said it ;) .. but I can change that of course.

andrewfg: @wborn do you know if the Z-Wave binding now also supports ip remote sticks?

andrewfg: Cool. I have removed the checks from the respective bindings' discovery-methods xml.

wborn: I also find it cool that OH will be able to suggest add-ons for a USB sticks connected to other computers exposed via ser2net. :sunglasses:

wborn (approved):
Thanks for the nice enhancement!

bundles/org.openhab.core.config.discovery.addon.usb/src/main/java/org/openhab/core/config/discovery/addon/usb/UsbAddonFinder.java
```diff
@@ -0,0 +1,186 @@
...
+     */
+    public static final String PRODUCT = "product";
+    public static final String MANUFACTURER = "manufacturer";
+    public static final String CHIP_ID = "chipId";
+    public static final String REMOTE = "remote";
+
+    public static final Set<String> SUPPORTED_PROPERTIES = Set.of(PRODUCT, MANUFACTURER, CHIP_ID, REMOTE);
+
+    private final Logger logger = LoggerFactory.getLogger(UsbAddonFinder.class);
+    private final Set<UsbSerialDiscovery> usbSerialDiscoveries = new CopyOnWriteArraySet<>();
```
Nadahar, line 70: @andrewfg Why it this set here? As far as I can tell, it's never used for anything..?

andrewfg:
@Nadahar I don't have a lot of time at the moment to dig into the details again. However the only purpose of this service (as far as I recall) is to discover suggested addons. Basically there are a few addons the depend on a USB stick hardware interface, (which  AFAIK may or may not be denominated as a serial port), and the service looks for specific Manufacturer Id and/or Chip Model Ids that are supported by those bindings. To find out more, you can examine the binding xml of the respective bindings. The Zwave, Zigbee and Enocean bindings come to mind, but there are probably others..

PR #3923, opened 2023-12-14
Extend OnOffType so it can be used as drop-in-replacement for Number types

Cybso:
I have added the methods of java class `Number` to `OnOffType`. This allows items of this class to be used as drop-on-replacement for numeric types in scripted rules:

```
    if (ir.getItem("someItem").getState().intValue() > 0) { .... }
```

Additionally, I have added a method `booleanValue()`, which allows this state to be used without importing OnOffType first.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/OnOffType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/OnOffType.java
@@ -60,6 +60,34 @@ public enum OnOffType implements PrimitiveType, State, Command {
         return super.toString();
     }
 
+    public boolean booleanValue() {
+        return this == ON;
+    }
+
+    public byte byteValue() {
+        return (byte) (this == ON ? 1 : 0);
+    }
+
+    public double doubleValue() {
+        return this == ON ? 1.0 : 0.0;
+    }
+
+    public float floatValue() {
+        return this == ON ? 1.0f : 0.0f;
+    }
+
+    public int intValue() {
+        return this == ON ? 1 : 0;
+    }
+
+    public long longValue() {
+        return this == ON ? 1L : 0L;
+    }
+
+    public short shortValue() {
+        return (short) (this == ON ? 1 : 0);
+    }
+
     @Override
     public <T extends State> @Nullable T as(@Nullable Class<T> target) {
         if (target == DecimalType.class) {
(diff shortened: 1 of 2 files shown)


wborn:
It's less code and easier to understand if you just use a static import for `OnOffType.ON`:

`if (ir.getItem("someItem").getState() == ON) {`

Cybso:
> It's less code and easier to understand if you just use a static import for `OnOffType.ON`:
> 
> `if (ir.getItem("someItem").getState() == ON) {`

Yes, but this is not what the patch intended. The patch allows to replace an Number item with a Boolean item without breaking existing rules.

wborn:
OnOffType is also used in add-on code and tests. Adding these options will probably result in contributors using the type in the wrong way and maintainers having to add review comments for this.

rkoshak:
> which allows this state to be used without importing OnOffType first.

Where do you need to import OnOffType in rules? It's imported by default in Rules DSL. jRuby I think provides a wrapper. JS Scripting always returns the state of Items as a String unless otherwise indicated. I don't know about Groovy. Is this actually solving a real problem or a perceived problem in rules. 

If it's a real problem, maybe the place to correct it is in the helper library for the language instead of in core where, as @wborn points out, it could have more knock on effects than the improvement is worth.

Wouldn't it make more sense to add a `getStateAs()` method than all of these new methods? That would be consistent with other Types like DimmerType, ColorType, etc. 

At a minimum I would expect to see just one `toNumber()` method. It's left to the caller to pull the primitive from the `Number`. Everywhere else a number appears in the OH API as seen from rules, when there's a number OH presents it as a `Number`, not a primitive. Consistency is important. 

Also, what about ContactType?

wborn:
Thanks for the contribution. I am going to close this PR because we do not think adding the Java `Number`-style methods to `OnOffType` is the right API direction.

`OnOffType` already supports explicit conversion to `DecimalType` through `as(DecimalType.class)`, which maps `ON` and `OFF` to numeric values while keeping the distinction between switch and numeric states clear. Adding methods such as `intValue()` and `doubleValue()` would not make `OnOffType` an actual `Number` in statically typed code, but could encourage add-on and rule code to treat unrelated state types interchangeably and conceal item-type changes.

The separate `booleanValue()` use case also does not currently have a sufficiently clear, language-independent motivation for adding it to the core type API. Should a particular scripting language still have an awkward conversion use case, that would probably be better addressed in its helper library based on a concrete example.

PR #3925, opened 2023-12-14
[Voice] Fix match question and allow force command

GiviMAD:
Hello, I would like to merge these fixes related to #3897.

It fixes an error introduced in that PR because the "?" should be removed when tokenizing unless when tokenizing the custom rule text, if not, it's imposible to match a question with the current implementation.

Also I have introduced the option to allow a custom rule to send a command no matter the item is already on that state, because if you are going to use the item to trigger a rule "on command received" instead of "on state change" you want the rule to behave that way and send the command always.


GiviMAD:
I'm going to add another commit here because I've just realize while adding the frontend metadata panel that the other namespace are named in camel case, so I think it should be voiceSystem instead of voice-system, sorry for the trouble.

GiviMAD:
The CI error seems unrelated to the changes.

kaikreuzer (approved):
Thanks!

GiviMAD:
Thanks for the review! 
I'm going to need some advice from you next year because I would like to move some features from HABSpeaker to the project. Happy new year in advance!

PR #3931, opened 2023-12-18
Don't log errors on NotFoundExceptions in Jetty

kaikreuzer:
(Partially) fixes https://github.com/openhab/openhab-distro/issues/1616

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/JSONResponseExceptionMapper.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/JSONResponseExceptionMapper.java
@@ -56,7 +56,7 @@ public class JSONResponseExceptionMapper implements ExceptionMapper<Exception> {
             // see https://github.com/openhab/openhab-distro/issues/1616
             logger.debug("Requested resource not (yet) found", e);
             // Returning null results in a Response.Status.NO_CONTENT response.
-            return null;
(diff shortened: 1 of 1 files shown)


florian-h05 (approved):
LGTM, thanks!

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/JSONResponseExceptionMapper.java
```diff
@@ -50,6 +51,12 @@ public class JSONResponseExceptionMapper implements ExceptionMapper<Exception> {
             logger.debug("Failed writing HTTP response, since other side closed the connection", e);
             // Returning null results in a Response.Status.NO_CONTENT response.
             return null;
+        } else if (e instanceof NotFoundException) {
+            // we catch this exception to avoid confusion errors in the log file, since this is not any error situation
+            // see https://github.com/openhab/openhab-distro/issues/1616
+            logger.debug("Requested resource not (yet) found", e);
+            // Returning null results in a Response.Status.NO_CONTENT response.
+            return null;
```
J-N-K, line 59: I don't think NO_CONTENT/204 is the correct response if the resource was not found. 204 indicates "success" and this is clearly not the case. 

```suggestion
            return Response.status(Response.Status.NOT_FOUND).build();
```

J-N-K (changes requested):

J-N-K (approved):
Thanks!

PR #3932, opened 2023-12-18
Reduce SAT warnings

holgerfriedrich:


wborn (approved):
Thanks for the reduction!

Issue #3936, opened 2023-12-19
Extend service to suggest addons via generic IP scan


holgerfriedrich:
PR #3920 introduced a (partly) generic addon finder, which allows to discover devices by sending single IP frames. This is useful for devices which do not actively announce and require a search request to a multicast address. Maybe the word scanning is a bit misleading, at the moment it is just a single search frame, and only once at start-up in the current implementation.

Note: if case you can use mdns or upnp, there are separate addon finders for this! If you want to detect a service running locally, there is one for processes as well.

The implementation is rather tuned to a single use right now - detecting KNX installations - but is intended to be extended towards release 4.2. **This issue is to collect requirements from other addons** which can be used when starting a generalization.

The configuration is via ```addon.xml``` and looks like this:
```
	<discovery-method>
		<discovery-method>
			<service-type>ip</service-type>
			<discovery-parameters>
				<discovery-parameter>
					<name>type</name>
					<value>ipMulticast</value>
				</discovery-parameter>
				<discovery-parameter>
					<name>destIp</name>
					<value>224.0.23.12</value>
				</discovery-parameter>
				<discovery-parameter>
					<name>destPort</name>
					<value>3671</value>
				</discovery-parameter>
				<discovery-parameter>
					<name>request</name>
					<value>0x06 0x10 0x02 0x01 0x00 0x0e 0x08 0x01 $srcIp $srcPort</value>
				</discovery-parameter>
				<discovery-parameter>
					<name>timeoutMs</name>
					<value>5000</value>
				</discovery-parameter>
			</discovery-parameters>
			<match-properties>
				<match-property>
					<name>response</name>
					<regex>.*</regex>
				</match-property>
			</match-properties>
		</discovery-method>
	</discovery-methods>

```

Frames to be sent are specified in the parameter section, and dynamic replacement of source IP and port is possible. Any return frame is matched, only the catch all .* is supported.

| addon |  detection method | scan interval | dynamic data | parsed response |
| --- | --- | --- | --- | --- |
| knx | mulitcast frame (hex data) | once | srcIp, srcPort | .* |
| ipcamera | multicase frame (xml) | tbd | randomNumber | tbd |

(table to be continued)

Skinah:
To summarize what is needed for the ipcamera addon as discussed in the forum link:

To discover ONVIF cameras it uses port 3702 and IP multicast address 239.255.255.250 using what is called WS-Discovery. It is just the following SOAP/XML packet sent.

```
[...]

PR #264, opened 2017-12-16
introduced eshRelease version

kaikreuzer:
As discussed at https://github.com/openhab/openhab-core/pull/259#discussion_r157341331

Signed-off-by: Kai Kreuzer <kai@openhab.org>

PR #3943, opened 2023-12-20
Minor extension to generic ip discovery

holgerfriedrich:
Add **new replacement $uuid** in request frame.
This is needed for discovery of ipcamera devices.

Add **new parameter listenPort** (required for govee).

Add **new parameter requestPlain** as alternative to request (enhance readability of request for text/xml/json based req., govee and ipcamera)

Improve resource handling:
* autocloseable
* ~~restrict to 5 addresses~~ (moved to next PR)
* ~~respect system.config.network.useOnlyOneAddress setting~~ (moved to next PR)

Refs: #3936

### DIFF
--- a/bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
+++ b/bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
@@ -102,9 +102,28 @@ import org.slf4j.LoggerFactory;
  * <td></td>
  * </tr>
  * <tr>
+ * <td>{@code listenPort}</td>
+ * <td>port to use for listening to responses (optional)</td>
+ * <td>privileged ports ({@code <1024}) not allowed</td>
+ * </tr>
+ * <tr>
  * <td>{@code request}</td>
  * <td>description of request frame as hex bytes separated by spaces (e.g. 0x01 0x02 ...)</td>
- * <td>dynamic replacement of variables $srcIp and $srcPort, no others implemented yet
+ * <td>dynamic replacement of variables $srcIp, $srcPort and $uuid, no others implemented yet
+ * </tr>
+ * <tr>
+ * <td>{@code requestPlain}</td>
+ * <td>description of request frame as plaintext string</td>
+ * <td>dynamic replacement of variables $srcIp, $srcPort and $uuid, no others implemented yet;
+ * there are five XML special characters which need to be escaped:
+ *
+ * <pre>{@code
+ * & - &amp;
+ * < - &lt;
+ * > - &gt;
+ * " - &quot;
+ * ' - &apos;
+ * }</pre>
  * </tr>
  * <tr>
  * <td>{@code timeoutMs}</td>
@@ -113,7 +132,25 @@ import org.slf4j.LoggerFactory;
  * </tr>
  * </table>
  * <p>
- * Packets are sent out on ever available network interface.
+ * <table border="1">
+ * <tr>
+ * <td><b>dynamic replacement</b> (in {@code request*})</td>
+ * <td><b>value</b></td>
+ * </tr>
+ * <tr>
+ * <td>{@code $srcIp}</td>
+ * <td>source IP address</td>
+ * </tr>
+ * <tr>
+ * <td>{@code $srcPort}</td>
+ * <td>source port</td>
+ * </tr>
+ * <td>{@code $uuid}</td>
+ * <td>String returned by {@code java.util.UUID.randomUUID()}</td>
+ * </tr>
+ * </table>
+ * <p>
+ * Packets are sent out on every available network interface.
  * <p>
  * There is currently only one match-property defined: {@code response}.
  * It allows a regex match, but currently only ".*" is supported.
--- a/bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
+++ b/bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
@@ -36,6 +36,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.StringTokenizer;
+import java.util.UUID;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
@@ -150,6 +151,7 @@ public class IpAddonFinder extends BaseAddonFinder {
     private static final String PARAMETER_SRC_IP = "srcIp";
     private static final String PARAMETER_SRC_PORT = "srcPort";
     private static final String PARAMETER_TIMEOUT_MS = "timeoutMs";
+    private static final String REPLACEMENT_UUID = "uuid";
 
     private final Logger logger = LoggerFactory.getLogger(IpAddonFinder.class);
     private final ScheduledExecutorService scheduler = ThreadPoolManager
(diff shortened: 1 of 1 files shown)


holgerfriedrich:
@Skinah FYI

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -164,6 +177,22 @@ private void scan() {
...
+                int listenPort = 0; // default, pick a non-privileged port
+                if (parameters.get(PARAMETER_LISTEN_PORT) != null) {
+                    try {
+                        listenPort = Integer.parseInt(Objects.toString(parameters.get(PARAMETER_LISTEN_PORT)));
+                    } catch (NumberFormatException e) {
+                        logger.warn("{}: discovery-parameter '{}' cannot be parsed", candidate.getUID(),
+                                PARAMETER_LISTEN_PORT);
+                        continue;
+                    }
+                    // not not allow privileged ports
```
Skinah, line 189:
```suggestion
                    // Do not allow privileged ports
```

Skinah:
Thank you for the changes.

andrewfg:
@holgerfriedrich the ZWay binding is probably a candidate to be suggested via your IP addon finder. => WDYT?

https://github.com/openhab/openhab-addons/tree/main/bundles/org.openhab.binding.zway#discovery

holgerfriedrich:
@andrewfg I can support / test if we go for zwave instead...

holgerfriedrich:
> @holgerfriedrich the ZWay binding is probably a candidate to be suggested via your IP addon finder. => WDYT?
> 
> https://github.com/openhab/openhab-addons/tree/main/bundles/org.openhab.binding.zway#discovery

Sorry, @andrewfg, I got you wrong in my first reply. Let's discuss this #3936. I will copy your comment to the other issue.

holgerfriedrich:
@mherwege I think this PR is ready for review. It contains extensions needed for govee and ipcamera bindings. @stefan-hoehn just confirmed it working for govee and approved my PR openhab/openhab-addons#16109.

andrewfg:
> support / test if we go for zwave instead...

Just for the avoidance of doubt:
- The Z-Wave binding is based on a Z-Wave USB stick attached to the PC which will be suggested by my USB finder.
- The ZWay binding is based on a Z-Wave bridge server somewhete on the LAN which would be suggested by your IP finder.

holgerfriedrich:
@mherwege Could you please take a look at this?
The PR contains minor extensions needed for govee and ipcamera.
Additionally, improves the resource behavior (autocloseable DatagramChannel, limit to 5 interfaces, respect system.config.network.useOnlyOneAddress).

holgerfriedrich:
@openhab/core-maintainers could you pls have a look? Auto-detect for 2 add-ons is depending on additional functions introduced in this PR:
- openhab/openhab-addons#16109 (govee)
- openhab/openhab-addons#16090 (ipcamera)

Thanks a lot!

mherwege:
In principle, this LGTM.

But I am wondering if we should not introduce use of https://github.com/openhab/openhab-core/pull/3981 at the same time. As we add requests being sent out to all interfaces, we should probably restrict the interfaces before getting too much into trouble.

holgerfriedrich:
@mherwege thanks for looking into this.

> But I am wondering if we should not introduce use of #3981 at the same time. As we add requests being sent out to all interfaces, we should probably restrict the interfaces before getting too much into trouble.

I had implemented a fallback to the main interface in case too many interfaces were found, but removed it for now to keep the PR cleaner.
I am in doubt that we would see an issue here because the requests are strictly sequential: we sent only one frame every few seconds (timeout is currently set to 5sec). 

What we will need to think about - but for me it is not yet scope of this PR - is either shortening the timeouts or doing requests more or less in parallel. Otherwise we will not finish the scan before the wizard suggests plugins during when OH is installed.

mherwege:
> I had implemented a fallback to the main interface in case too many interfaces were found, but removed it for now to keep the PR cleaner.

Agree to keep it out of here, but that’s where it should probably be in sync with the work on https://github.com/openhab/openhab-core/pull/3981, so for a follow-up PR.



> What we will need to think about - but for me it is not yet scope of this PR - is either shortening the timeouts or doing requests more or less in parallel. Otherwise we will not finish the scan before the wizard suggests plugins during when OH is installed.

Maybe define CompletableFuturs with timeout for each request and start them in parallel. Agree we will need more throughput on this. We can delay initial configuration in the UI a bit, showing a preparing screen, within reason.

holgerfriedrich:
@mherwege @andrewfg should we go forward merging this PR as it is?
Obviously, there is more work to do as described in #3936. This is just a first step.

andrewfg:
I would suggest to use CompletableFuture as @mherwege suggests. Then the scans can all be executed in parallel on separate threads.

mherwege:
> I would suggest to use CompletableFuture as @mherwege suggests. Then the scans can all be executed in parallel on separate threads.

I tend to agree here. It now becomes easy to add discovery info to add-ons. If a number of them are added in the addon repo, suddenly we have code running a sequence of scans potentially each timing out. We should avoid having to come back to this quickly when discovery info gets added to the addon.xml files.

andrewfg:
> suddenly we have code running a sequence of scans potentially each timing out

In particular in OH core, there should be no long running processes on main threads.

andrewfg:
I am thinking that the approach below would perhaps be appropriate..

https://github.com/openhab/openhab-addons/pull/16315

andrewfg (commented):
@holgerfriedrich apropos threading .. to be a bit more helpful I would suggest refactoring your code lines 265..395 into a separate method, and at line #264 call that method via something like List<Future> myFutures .. add(scheduler.submit(() -> thatMethod()))`.

And on shutdown you should also do `myFutures.forEach(f -> cancel())`;

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -322,6 +396,30 @@ private void scan() {
         logger.trace("IpAddonFinder::scan completed");
     }
 
+    // build from plaintext string
+    private byte[] buildRequestArrayPlain(DatagramChannel channel, String request)
+            throws java.io.IOException, ParseException {
+        InetSocketAddress sock = (InetSocketAddress) channel.getLocalAddress();
+
+        // replace first
+        StringBuffer req = new StringBuffer(request);
```
wborn, line 405: The `StringBuilder` is preferred when there is no need for synchronization as it has better performance:

```suggestion
        StringBuilder req = new StringBuilder(request);
```

holgerfriedrich:
I agree that better handling of multiple addresses and interfaces is desired for IP discovery. Which solution to implement is not yet clear, maybe there will be a result out of #4047.

Parallelization is required as well when more add-ons start using this. Thanks, @mherwege and @andrewfg for pointing at ```CompletableFuture``` and implementations. There are some technical questions not yet clarified, e.g. how to make sure that discovery of different add-ons at the same time does not cause problems. Just imagine two add-ons using the same server port, etc...

I still think that both is out of scope of this PR and tracked #3936.

The PR adds new config options and has already been validated with the corresponding add-on.

andrewfg:
@mherwege [HERE](
https://github.com/andrewfg/openhab-core/blob/ip-finder-async/bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java) is some code to do the scans asynchronously (not tested).

wborn (approved):
Thanks for the extension!

andrewfg:
@holgerfriedrich shall I convert my code to a new PR?

holgerfriedrich:
> @holgerfriedrich shall I convert my code to a new PR?

Thanks, @andrewfg. I have created #4094 based on your example above. Hope you are fine with that.

PR #3944, opened 2023-12-20
Move addons.xml to runtime folder

kaikreuzer:
See https://github.com/openhab/openhab-distro/issues/1616#issuecomment-1865154961

### DIFF
--- a/bundles/org.openhab.core.addon/src/main/java/org/openhab/core/addon/internal/xml/AddonInfoAddonsXmlProvider.java
+++ b/bundles/org.openhab.core.addon/src/main/java/org/openhab/core/addon/internal/xml/AddonInfoAddonsXmlProvider.java
@@ -83,7 +83,7 @@ public class AddonInfoAddonsXmlProvider implements AddonInfoProvider {
         File file = new File(fileName);
         try {
             if (!file.isFile()) {
-                logger.warn("File '{}' does not exist.", fileName);
+                logger.debug("File '{}' does not exist.", fileName);
                 return;
             }
         } catch (SecurityException e) {
--- a/bundles/org.openhab.core.addon/src/main/java/org/openhab/core/addon/internal/xml/AddonInfoAddonsXmlProvider.java
+++ b/bundles/org.openhab.core.addon/src/main/java/org/openhab/core/addon/internal/xml/AddonInfoAddonsXmlProvider.java
@@ -22,7 +22,6 @@ import java.util.Locale;
 import java.util.Set;
 import java.util.regex.PatternSyntaxException;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.core.addon/src/main/java/org/openhab/core/addon/internal/xml/AddonInfoAddonsXmlProvider.java
```diff
@@ -79,37 +80,34 @@ public Set<AddonInfo> getAddonInfos(@Nullable Locale locale) {
     }
 
     private void initialize() {
-        File fileFolder = new File(folder);
+        File file = new File(fileName);
         try {
-            if (!fileFolder.isDirectory()) {
-                logger.debug("Folder '{}' does not exist", folder);
+            if (!file.isFile()) {
+                logger.warn("File '{}' does not exist.", fileName);
```
J-N-K, line 86: IMO a missing file should not result in a warning. Distributions could decide to omit this file. A warning should be logged if it is a directory or the file cannot be read.

J-N-K (changes requested):
Otherwise LGTM

bundles/org.openhab.core.addon/src/main/java/org/openhab/core/addon/internal/xml/AddonInfoAddonsXmlProvider.java
kaikreuzer, line 86:
```suggestion
                logger.debug("File '{}' does not exist.", fileName);
```

kaikreuzer: Ok. It was useful while testing though. ;-)

J-N-K (approved):
LGTM. Can you please self-merge (also in distro) after the build succeeded? I have approved both change-Sets. I'm offline now.

kaikreuzer:
Will do, thanks!

bundles/org.openhab.core.addon/src/main/java/org/openhab/core/addon/internal/xml/AddonInfoAddonsXmlProvider.java
```diff
@@ -41,20 +40,22 @@
...
 /**
- * The {@link AddonInfoAddonsXmlProvider} reads all {@code userdata/addons/*.xml} files, each of which
- * should contain a list of {@code addon} elements, and convert their combined contents into a list
- * of {@link AddonInfo} objects can be accessed via the {@link AddonInfoProvider} interface.
+ * The {@link AddonInfoAddonsXmlProvider} reads the {@code runtime/etc/addons.xml} file, which
+ * should contain a list of {@code addon} elements, and convert its combined contents into a list
+ * of {@link AddonInfo} objects that can be accessed via the {@link AddonInfoProvider} interface.
  *
  * @author Andrew Fiddian-Green - Initial contribution
+ * @author Kai Kreuzer - Reduce it to support a single addons.xml file
```
andrewfg, line 48: @kaikreuzer I don't think we should reduce it to a single file. The architecture was intentionally created to allow multiple files to be provided e.g. from OH main distribution, from marketplace, or eventually other sources..

andrewfg: And @kaikreuzer .. correct me if I am wrong but the file is named ‘openhab-addons.xml’ and NOT simply ‘addon.xml’ .. I think you just made a breaking change here. Or??

andrewfg: @mherwege please comment.

andrewfg: See this https://github.com/J-N-K/openhab-distro/blob/bf272a68a0f52a3474a069f4ed1e517b1b338ea6/features/distro/src/main/feature/feature.xml#L12

andrewfg: … and I think that one of the recent PRs did indeed put some xml files in that folder because the distro scan was only picking up binding addons, and another xml file was needed to pick up other sorts of addons. I think you may have broken this now.

andrewfg: See this.. https://github.com/openhab/openhab-addons/pull/16044

mherwege: > … and I think that one of the recent PRs did indeed put some xml files in that folder because the distro scan was only picking up binding addons, and another xml file was needed to pick up other sorts of addons. I think you may have broken this now.

It was missing files in the aggregation, but when that was fixed, it all went in the same destination file. So this should have no impact.

mherwege: > @mherwege please comment.

I found the multiple files handy for testing (just drop an extra file in), but I don’t think it is required. Other add-on sources would normally be better at providing the metadata before installation. The 3rd part provider only needs an extension to the JSON add-on info content. The marketplace s another issue, but it will not be solved by an extra provider. More info is needed in the Discourse entry.

kaikreuzer: That was my thinking here as well. The addons.xml file is a system file and not meant to be tampered with by users - therefore it must not be in the userdata folder. And folders that are in `runtime` are not meant to be used by users - hence this would be the wrong place to allow users adding additional files. If we really need this in future, we can still add a conf folder where users can supply additional information.

And no worries, @andrewfg, I didn't break anything - the latest distro works nicely with the change.

andrewfg: Thanks @kaikreuzer and @mherwege for the clarification.

PR #3945, opened 2023-12-21
Javadoc: Introduce custom tags @ implNote and @ apiNote

holgerfriedrich:
In addons, we have introduced the custom tags quite a while ago, first on the binding level, and later at top level.

I just noticed, that javadoc build for core does not contain @ implNote sections (neither heading nor content).

With the patch it looks like this:
grafik

As of now, the part marked in yellow is missing.
https://ci.openhab.org/view/Documentation/job/openHAB-JavaDoc/javadoc/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.html

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -345,6 +345,18 @@ Import-Package: \\
                 </script>
               ]]>
             </bottom>
+            <tags>
+              <tag>
+                <name>apiNote</name>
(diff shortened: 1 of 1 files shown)


wborn (approved):
Makes sense to also support it in this repo. :+1:

PR #3951, opened 2023-12-21
Delay the IP scan to make sure that no further candidates are added

kaikreuzer:
…and do not use a dedicated thread pool

### DIFF
--- a/bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
+++ b/bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
@@ -104,6 +104,10 @@ public class IpAddonFinder extends BaseAddonFinder {
     }
 
     private void startScan() {
+        // The setAddonCandidates() method is called for each info provider.
+        // In order to do the scan only once, but on the full set of candidates, we have to delay the execution.
+        // At the same time we must make sure that a scheduled scan is rescheduled - or (after more than our delay) is
+        // executed once more.
         stopScan();
         scanJob = scheduler.schedule(this::scan, 20, TimeUnit.SECONDS);
     }
--- a/bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
+++ b/bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
@@ -12,8 +12,7 @@
  */
 package org.openhab.core.config.discovery.addon.ip;
 
-import static org.openhab.core.config.discovery.addon.AddonFinderConstants.SERVICE_NAME_IP;
-import static org.openhab.core.config.discovery.addon.AddonFinderConstants.SERVICE_TYPE_IP;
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -81,7 +80,8 @@ public class IpAddonFinder extends BaseAddonFinder {
     private static final String PARAMETER_TIMEOUT_MS = "timeoutMs";
 
     private final Logger logger = LoggerFactory.getLogger(IpAddonFinder.class);
-    private final ScheduledExecutorService scheduler = ThreadPoolManager.getScheduledPool(SERVICE_NAME);
```
kaikreuzer, line 84: This created a dedicated pool of 5 threads, although there is a single job to be run only a single time.

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -96,34 +96,30 @@ public void deactivate() {
...
+    @Override
     public void setAddonCandidates(List<AddonInfo> candidates) {
         logger.debug("IpAddonFinder::setAddonCandidates({})", candidates.size());
         super.setAddonCandidates(candidates);
         startScan();
     }
 
-    synchronized void startScan() {
-        if (scanJob == null) {
-            scanJob = scheduler.schedule(this::scan, 1, TimeUnit.SECONDS);
```
kaikreuzer, line 107: The `setAddonCandidates` method is called for each info provider.
In order to do the scan only once, but on the full set of candidates, we have to delay the execution.
At the same time we must make sure that a scheduled scan is rescheduled - or (after more than our delay) is executed once more.

holgerfriedrich: interesting....
I wonder if somebody could reach the setup wizard before the scan is done..... 20s might be too long.

mherwege (approved):
LGTM

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
kaikreuzer: > 20s might be too long

I think it is a good compromise. I doubt that people rush through the wizard. Creating an account and choosing a good password should take them already a while. And if you consider that startup times on a RPi are around 5 minutes, I think we should be on the safe side.

kaikreuzer: > btw: maybe we should add your explanation as comment to the code

Done.

holgerfriedrich (approved):
LGTM

kaikreuzer:
@J-N-K I'll self merge this now as it has 2 approvals and I want to get it into the release.

J-N-K (commented):
LGTM

PR #3954, opened 2023-12-23
Use `OnOffType.from` to reduce code

wborn:
You can create an `OnOffType` using a boolean nowadays which reduces the amount of code.


holgerfriedrich (approved):
LGTM

kaikreuzer (approved):
Thanks!

PR #3960, opened 2023-12-25
[Audio] Add piped audio stream and fix raw PCM streams format

GiviMAD:
Hello,

I want to add a new audio stream implementation to send infinite audio streams and multiplex its data to the different consumers. At the same time I discovered I had committed an error across all the source implementations which is create its format using the CONTAINER_WAV while those streams are raw pcm streams and should use the CONTAINER_NONE instead (I don't think it needs to be treated as a bug and back-ported because I was consistent with the error and it has no effect on the 4.1.x code).

I'm going to open a PR in the add-ons to adapt them to these changes.

The use cases for the new stream type are:

* Be consistent across the source implementations because both the built-in system source and the pulseaudio binding source are using the same mechanism for sharing data, and I'm planning on try to add another one and it requires same mechanism either.
* Been able to send infinite streams and handle them in a specific way; I want to use these to be able to connect a source with a sink and also to try to stream music from the server.


GiviMAD:
I have rebased the PR to remove a comment change that has been already fixed at main and was not related.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioFormat.java
```diff
@@ -436,29 +441,13 @@ public boolean isCompatible(@Nullable AudioFormat audioFormat) {
...
-                return false;
-            }
-            if (!(null == getFrequency() ? null == format.getFrequency()
-                    : getFrequency().equals(format.getFrequency()))) {
-                return false;
-            }
-            if (!(null == getChannels() ? null == format.getChannels() : getChannels().equals(format.getChannels()))) {
-                return false;
-            }
+            return Objects.equals(getCodec(), format.getCodec()) && //
```
kaikreuzer, line 444: If all conditions are met, this now returns `true` while before it returned `super.equals(obj)`. Is this intended?

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/PipedAudioStream.java
```diff
@@ -0,0 +1,275 @@
...
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This is an implementation of an {@link AudioStream} used to transmit raw audio data to a sink.
+ *
+ * It just pipes the audio through it, the default pipe size its equal to 1s of audio,
```
kaikreuzer, line 33:
```suggestion
 * It just pipes the audio through it, the default pipe size is equal to 1 second of audio,
```

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/PipedAudioStream.java
```diff
@@ -0,0 +1,275 @@
...
+        return pipedOutput;
+    }
+
+    /**
+     * Creates a new piped stream group used to open new streams and write data to them.
+     *
+     * Internal pipe size 0.5s.
+     *
+     * @param format the audio format of the group audio streams
+     * @return a group instance.
```
kaikreuzer, line 118:
```suggestion
     * @return a group instance
```

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/PipedAudioStream.java
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    protected PipedOutputStream getOutputStream() {
+        return pipedOutput;
+    }
+
+    /**
+     * Creates a new piped stream group used to open new streams and write data to them.
+     *
+     * Internal pipe size 0.5s.
```
kaikreuzer, line 115:
```suggestion
     * Internal pipe size is 0.5s.
```

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/PipedAudioStream.java
```diff
@@ -0,0 +1,275 @@
...
+                (float) Objects.requireNonNull(format.getChannels()) //
+        ) / 2f);
+        return new Group(format, pipeSize);
+    }
+
+    /**
+     * Creates a new piped stream group used to open new streams and write data to them.
+     *
+     * @param format the audio format of the group audio streams
+     * @param pipeSize the pipe size of the created streams.
```
kaikreuzer, line 133:
```suggestion
     * @param pipeSize the pipe size of the created streams
```

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/PipedAudioStream.java
```diff
@@ -0,0 +1,275 @@
...
+     * @return a piped stream group instance.
```
kaikreuzer, line 134:
```suggestion
     * @return a piped stream group instance
```

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/PipedAudioStream.java
```diff
@@ -0,0 +1,275 @@
...
+     */
+    public static Group newGroup(AudioFormat format, int pipeSize) {
+        return new Group(format, pipeSize);
+    }
+
+    /**
+     * The {@link PipedAudioStream.Group} its an {@link OutputStream} implementation that use can use to
```
kaikreuzer, line 141:
```suggestion
     * The {@link PipedAudioStream.Group} is an {@link OutputStream} implementation that can be use to
```

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/PipedAudioStream.java
```diff
@@ -0,0 +1,275 @@
...
+     * create one or more {@link PipedAudioStream} instances and write to them at once.
+     *
+     * The created {@link PipedAudioStream} instances are removed from the group when closed.
+     *
```
kaikreuzer, line 145:
```suggestion
```

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/PipedAudioStream.java
```diff
@@ -0,0 +1,275 @@
...
+        private final Logger logger = LoggerFactory.getLogger(Group.class);
+
+        protected Group(AudioFormat format, int pipeSize) {
+            this.pipeSize = pipeSize;
+            this.format = format;
+        }
+
+        /**
+         * Creates a new {@link PipedAudioStream} connected to the group.
+         * The stream unregister itself from the group on close.
```
kaikreuzer, line 160:
```suggestion
         * The stream unregisters itself from the group on close.
```

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/PipedAudioStream.java
```diff
@@ -0,0 +1,275 @@
...
+         * 
+         * @return a new {@link PipedAudioStream} to pipe data written to the group.
```
kaikreuzer, line 162:
```suggestion
         * @return a new {@link PipedAudioStream} to pipe data written to the group
```

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/PipedAudioStream.java
```diff
@@ -0,0 +1,275 @@
...
+         * @throws IOException when unable to create the stream.
```
kaikreuzer, line 163:
```suggestion
         * @throws IOException when unable to create the stream
```

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/PipedAudioStream.java
```diff
@@ -0,0 +1,275 @@
...
+            audioStream.onClose(() -> {
+                if (!openPipes.remove(audioStream)) {
+                    logger.warn("Trying to remove an unregistered stream, this is not expected");
+                }
+            });
+            return audioStream;
+        }
+
+        /**
+         * Returns true if this group has no streams connected
```
kaikreuzer, line 181:
```suggestion
         * Returns true if this group has no streams connected.
```

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/PipedAudioStream.java
```diff
@@ -0,0 +1,275 @@
...
+         *
+         * @return true if this group has no streams connected.
```
kaikreuzer, line 183:
```suggestion
         * @return true if this group has no streams connected
```

kaikreuzer (changes requested):
Thanks, looks good to me, I've added mainly some JavaDoc improvements.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioFormat.java
GiviMAD: Yes, I think that was a mistake, because if I understood it correctly that super.equals(obj) will execute the default implementation for object instances and return true only for the same instance, so the previous checks will always be true when true and having the methods overwrited makes no difference. I think the intention here was to consider equal formats with the same properties.

GiviMAD:
I have added a `size` method to check the number of streams in the group in case is needed.

kaikreuzer (approved):
lgtm, thanks!

Issue #3852, opened 2023-10-19
[marketplace] Wrong binding version installed


weymann:
**Reproduce**
1) Take latest snapshot or milestone of OH4.1
2) Go to Marketplace and install `MercedesMe `binding
3) Installation ends with excpetion

Observed in community [here](https://community.openhab.org/t/mercedes-me-binding/136852/161)
Reproduced by myself with OH 4.1 snapshot [here](https://community.openhab.org/t/mercedes-me-binding/136852/162)
After analysis 3.4 version was installed via Marketplace
![image](https://github.com/openhab/openhab-core/assets/20279271/991e9e79-2813-447f-8449-c2be9e4d9157)

Maybe something wrong with the [Marketplace page](https://community.openhab.org/t/mercedes-me/136866)?

rkoshak:
My understanding on how the marketplace works you can't combine versions in one post like that. The first link under "Resources" is going to be assumed to be the add-on jar, second one is assumed to be the source code, and everything after that is ignored. So it's treating your 3.4 .jar file as the add-on, the 4.0 jar file as the source code and ignoring the source code link entirely (not that anything is really done with it beyond verifying that a link is there).

So you either need one .jar file that can work for all versions of OH (see https://community.openhab.org/t/androidtv-binding-3-2-0-4-1-0/142203 for an example of that) or you need two marketplace postings and use the versioning in the title to indicate which OH the posting is for. MainUI will only show the posting that is compatible with the OH running

For example, my "Time Bases State Machine" rule template has two posts to the marketplace. The one that works with OH 3 uses `Time Based State Machine [3.2.0;3.4.9]` for the title and the one for OH 4 uses `Time Based State Machine [4.0.0.0;4.9.9.9]`. You can see many additional examples of this in the bundles category too (e.g. NAD AV Receiver among others).

I don't think this is a bug. The parser built into MainUI just isn't that smart.

J-N-K:
No, the parsing is done in core.

PR #3961, opened 2023-12-25
Fix duplicate UIDs in remote add-on services

J-N-K:
Missing UIDs or multiple versions of the same add-on resulted in duplicate entries for the community marketplace or JSON add-on service. With this fix add-ons with missing UID (i.e. from versions pre-4.0, before UIDs were introduced in the JSON) are now discarded and duplicate UIDs are filtered, with compatible add-ons and - if same compatibility - higher version being preferred. This is mostly a bug-fix, but also necessary for the upcoming multi-version community-marketplace add-ons I'm working on (see #3852). Since it affects users I suggest to also back-port it to `4.1.x`.


kaikreuzer (approved):
lgtm, thanks

kaikreuzer:
cherry-picked.

PR #3964, opened 2023-12-26
Fix month, week, day not supported

### LINKED ISSUE #3962: [QuantityType] Time does not accept y as unit
## Expected Behavior
If setting up an Item of type `Number:Time`, it should be possible to use `y` as unit.
```
Number:Time bsP1Age "Alter" (gScale) ["Duration"] {channel="mqtt:topic:mosquitto:bs444:age", unit="y"}
```
should result in an Item state with years and fractions of years, e.g. `50 y` or `50.83 y`

## Current Behavior
when defining the Item as above, this results in a warning:
```
[WARN ] [penhab.core.library.items.NumberItem] - Unit 'y' could not be parsed to a known unit. Keeping old unit 's' for item 'bsP1Age'.
```

## Environment
* Version used: openHAB4.1.0 (but also true for older versions)
* Operating System and version: debian bookworm (amd64 in proxmox lxc)

Interestingly, the corresponding channel is also of type number with unit `y` and openHAB does not complain about (but the value isn't calculated correctly from `y` to `s`) 
```
Type number : age "Alter" [ stateTopic="bs444/udo/", transformationPattern="JSONPATH:$.age", unit="y" ]
```
All other units (`s`, `min`, `d`, `h`, `week`) seem to work as intended.

J-N-K:
Fixes #3962

Due to the way month, week and day are defined in Indriya their symbol is not added as alias. This is a bug in indriya, but their release cycles are quite long and we should provide a fix for our users.

This should be backported to 4.1.x

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/Units.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/Units.java
@@ -191,6 +191,7 @@ public final class Units extends CustomUnits {
     public static final Unit<Time> HOUR = addUnit(tech.units.indriya.unit.Units.HOUR);
     public static final Unit<Time> DAY = addUnit(tech.units.indriya.unit.Units.DAY);
     public static final Unit<Time> WEEK = addUnit(tech.units.indriya.unit.Units.WEEK);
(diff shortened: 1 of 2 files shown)


kaikreuzer (approved):
lgtm, thanks.

kaikreuzer:
Cherry-picked.

PR #3966, opened 2023-12-27
Fix NPE in GroupStateTriggerHandler and GroupCommandTriggerHandler

J-N-K:
Reported on the community forum. The handler checks if the item's group names contains the configured value. A `null` group name results in the NPW during that check.


wborn (approved):
Thanks for the fix!
Maybe it is also easy to add unit tests for this?

J-N-K:
What should be covered by that test? The "real" problem is that we don't validate the trigger configurations before we create the modules.

wborn:
> What should be covered by that test?

Based on the changes the tests could check that a warning is logged whenever the `groupName` is missing and no NPE is thrown when the `receive` method is called.
Then if someone in the future removes the check or messes up the logic the tests would fail.

J-N-K:
Done

wborn (approved):
Thanks!

lolodomo:
@J-N-K : is it something to backport in branch 4.1.x ?

PR #3968, opened 2023-12-27
Remove unnecessary parenthesis from lambdas

wborn:
These parentheses do not change the meaning of the code and also do not improve the readability.


J-N-K (approved):
Thanks!

PR #3970, opened 2023-12-27
[rest] Add `no-cache` directive to cached REST responses

florian-h05:
Fixes https://github.com/openhab/openhab-webui/issues/2102, where Firefox users reported problems with outdated data in Main UI due to Firefox handling caching different than Chromium.
Also reported on the community, e.g. https://community.openhab.org/t/firefox-oh-4-0-1-issues/148633 and https://community.openhab.org/t/updating-widget-code-not-working/148265.

This forces the browser to revalidate the cache every time it is accessed to ensure the data is always fresh. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#directives.

This is approach is also suggested in the mdn web docs, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#up-to-date_contents_always.

This PR should be backported to 4.1.x.

### DIFF
--- a/bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/RuleResource.java
+++ b/bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/RuleResource.java
@@ -188,6 +188,7 @@ public class RuleResource implements RESTResource {
                     .map(rule -> EnrichedRuleDTOMapper.map(rule, ruleManager, managedRuleProvider));
 
             CacheControl cc = new CacheControl();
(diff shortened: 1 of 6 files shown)


wborn:
Do you know why the `CacheControl` is used in these Resources and not others like `InboxResource` or `AddonResource`?

florian-h05:
At the moment, only some of the resources have implemented caching, so obviously only the Resources with caching implement need CacheControl.

wborn (approved):
Thanks! It makes sense to only cache the most frequently used and biggest responses.

PR #3972, opened 2023-12-27
Replace lambdas with method references

wborn:
Method references are more readable because they refer to class names and usually result in less code.


J-N-K (approved):
Thanks for the cleanup

PR #3973, opened 2023-12-27
Use String.join instead of Collectors.joining

wborn:
String.join results in less code when joining an Iterable or Array.

### DIFF
--- a/bundles/org.openhab.core.model.thing/src/org/openhab/core/model/thing/valueconverter/UIDtoStringConverter.java
+++ b/bundles/org.openhab.core.model.thing/src/org/openhab/core/model/thing/valueconverter/UIDtoStringConverter.java
@@ -27,14 +27,14 @@ import org.openhab.core.thing.UID;
  */
 public class UIDtoStringConverter implements IValueConverter<String> {
 
-    private static final String SEPERATOR = ":";
+    private static final String SEPARATOR = ":";
 
     @Override
     public String toValue(final String string, INode node) throws ValueConverterException {
         if (string == null) {
             return null;
         }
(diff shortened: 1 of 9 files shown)


bundles/org.openhab.core.model.thing/src/org/openhab/core/model/thing/valueconverter/UIDtoStringConverter.java
```diff
@@ -48,7 +45,7 @@ public String toValue(final String string, INode node) throws ValueConverterExce
                 }
             }
         }
-        return Arrays.stream(ids).collect(Collectors.joining(SEPERATOR));
+        return String.join(SEPERATOR, ids);
```
J-N-K, line 48:
```suggestion
        return String.join(SEPARATOR, ids);
```

bundles/org.openhab.core.model.thing/src/org/openhab/core/model/thing/valueconverter/UIDtoStringConverter.java
```diff
@@ -65,7 +62,7 @@ public String toString(String value) throws ValueConverterException {
                 ids[i] = toEscapedString(id);
             }
         }
-        return Arrays.stream(ids).collect(Collectors.joining(SEPERATOR));
+        return String.join(SEPERATOR, ids);
```
J-N-K, line 65:
```suggestion
        return String.join(SEPARATOR, ids);
```

J-N-K (changes requested):
Can you fix the typo?

PR #3975, opened 2023-12-28
More code cleanups

wborn:
While cleaning up the code I found a some more code to cleanup:

* Remove unnecessary boxing
* Use `contains(..)` instead of `indexOf(..) != -1`
* Use `assertInstanceOf` in tests
* Make expensive trace logging conditional
* Remove redundant constructor
* Replace `collect(Collectors.toUnmodifiableList())` with `toList()`
* Replace `filter(..).count() == 0L` with `noneMatch(..)`
* Replace `filter(..).count() > 0` with `anyMatch(..)`


J-N-K (approved):
LGTM

PR #3981, opened 2023-12-29
Add network-interface context

wborn:
This allows for selecting network interface names in configuration parameters.

Just add `<context>network-interface</context>` to your config parameter similar to `<context>serial-port</context>` .

### DIFF
--- a/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/net/NetworkConfigOptionProvider.java
+++ b/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/net/NetworkConfigOptionProvider.java
@@ -13,9 +13,13 @@
 package org.openhab.core.config.core.internal.net;
 
 import java.net.Inet4Address;
+import java.net.InetAddress;
+import java.net.NetworkInterface;
+import java.net.SocketException;
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Enumeration;
 import java.util.List;
 import java.util.Locale;
 import java.util.stream.Stream;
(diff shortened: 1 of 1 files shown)


mherwege:
Could this also be used by the suggestion finders to limit the interfaces for scanning? Or by UPnP, mDNS. Or does use of the context limits it to use as binding parameters?
If it can be used elsewhere, where can it be configured?
Can it be configured in config files (other than for network binding things)?
I actually think this could be an overall parameter, just like the other network parameters.

wborn:
> Could this also be used by the suggestion finders to limit the interfaces for scanning? Or by UPnP, mDNS. Or does use of the context limits it to use as binding parameters?

You can use contexts with any `config-description` parameter, see [Supported Contexts](https://www.openhab.org/docs/developer/addons/config-xml.html#supported-contexts).

> If it can be used elsewhere, where can it be configured?

Just add `<context>network-interface</context>` to your text parameter in the XML, e.g. in:

* [network.xml](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core/src/main/resources/OH-INF/config/network.xml)
* [addons.xml](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core/src/main/resources/OH-INF/config/addons.xml)

> Can it be configured in config files (other than for network binding things)?

The values are just text so it should work. The context is used to define the available options. This can then be used by UIs to present configuration options and by the framework to validate values.

> I actually think this could be an overall parameter, just like the other network parameters.

Maybe as a default value instead of always using every available interface which can be many when using Docker. But you can still have the situation where you want to use one network interface most of the time but still want to have the option to only use some network interfaces other times e.g. when using arping to see if a container is up.

J-N-K (approved):
LGTM

kaikreuzer:
Cherry-picked to `4.1.x`.

PR #3989, opened 2023-12-31
Respond with 404 if add-on is missing in add-on service

### LINKED ISSUE #3987: API POST endpoint /rest/addons always returns 200 even with wrong parameters
POST-ing to /rest/addons to install a marketplace add-on will always return 200 (OK) with an empty body, no matter what parameters are passed. Also when installing an already installed add-on, it gives the same result.
I would expect a 404 if it gets passed a non-existing add-on.

This is a simple script (rule) I use to install marketplace addons during startup:

```ruby
rule "InstallMarketAddons" do
  description "Install marketplace add-ons at system start"
  on_start at_level: 80
  delay 20.seconds  # Let OH core do its magic first
  run do
    # Extract the marketplace add-ons line from `conf/services/addons.cfg`
    cfgLines = IO.readlines('{{ openhab_conf_dir }}/services/addons.cfg')
    if lineIndex = cfgLines.index{ |line| line =~ /marketplace/ } then
      logger.info "Install Marketplace addons defined in addons.cfg at system start"
      addons = cfgLines[lineIndex].split('=')[1].strip
      # Loop through each addon
      addons.split(",").each do |addon|
        addonName, addonId = addon.split(":")
        #TODO: Check for ill-formatted entries
        logger.info "Install addon #{addonName} (#{addonId}) from marketplace"
        # Create the HTTP object
        uri = URI.parse("{{ openhab_api_url }}/addons/marketplace:#{addonId}/install?serviceId=marketplace")
        header = {'Content-Type': 'text/plain', 'Accept': '*/*', 'X-OPENHAB-TOKEN': '{{ openhab_api_token }}'}
        http = Net::HTTP.new(uri.host, uri.port)
        request = Net::HTTP::Post.new(uri.request_uri, header)
        request.body = "id=#{addonId}"
        # Send the request
        response = http.request(request)
        # Note: The API always returns 200 with an empty body, so actually no use logging it
        logger.debug "Result #{response.code} (#{response.body}) - installing marketplace addon #{addonName} (#{addonId})"
      end # of loop through each entry
    end # if
  end
end
```

There are two Ansible variables in the script to specify the base URL of the API endpoint and the token.
The script works by reading a line starting with `marketplace=` and extracting the add-on id's to install.
By specifying a non-exsitings binding the reported behaviour can be reproduced.

## Your Environment
Debian 12.
OpenHAB 4.1.0 using the official OH container image.

J-N-K:
Fixes #3987 

We can easily check if the requested service supports the add-on and respond with "NOT_FOUND" when the service is unknown or the add-on is not available from that service.

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/addons/AddonResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/addons/AddonResource.java
@@ -253,7 +253,7 @@ public class AddonResource implements RESTResource {
     public Response installAddon(final @PathParam("addonId") @Parameter(description = "addon ID") String addonId,
             @QueryParam("serviceId") @Parameter(description = "service ID") @Nullable String serviceId) {
         AddonService addonService = (serviceId != null) ? getServiceById(serviceId) : getDefaultService();
-        if (addonService == null) {
+        if (addonService == null || addonService.getAddon(addonId, null) == null) {
             return Response.status(HttpStatus.NOT_FOUND_404).build();
         }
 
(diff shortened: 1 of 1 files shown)


kaikreuzer (approved):
lgtm, thanks!

kaikreuzer:
Cherry-picked to `4.1.x`.

PR #3990, opened 2024-01-01
Add the possibility to invert the mute system channel

kaikreuzer:
Very often, the mute feature is counter-intuitive: in the UI, the user rather wants to see whether a speaker/zone/amp is on (active) or off (deactivated).
This PR gives the possibility to invert the "mute" behavior to make the above scenario an option.


kaikreuzer:
Checking [how many bindings have a mute channel](https://github.com/search?q=repo%3Aopenhab%2Fopenhab-addons+typeId%3D%22system.mute%22&type=code&p=1), I am withdrawing this PR, since it would be too much effort to add support for this configuration to all of them and leaving it unsupported would be unexpected for users.

PR #3814, opened 2023-09-27
Upgrade to Karaf 4.4.4

wborn:
* Sync runtime dependencies with Karaf 4.4.4, most notably:
  * Jetty 9.4.52.v20230823
  * JNA 5.13.0
  * SLF4J 2.0.6
  * Pax Logging 2.2.3
  * Pax Web 8.0.22
* Resolve itest runbundles
* Use new Pax Web features to simplify dependency management
* Add specs features because Pax Web now depends on "asm"

PR #3991, opened 2024-01-01
Fix build number reporting at startup

kaikreuzer:
openHAB currently logs during startup something like
```
22:09:04.949 [INFO ] [org.openhab.core.Activator          ] - Starting openHAB 4.2.0 (build Unknown Build No.)
```
The word `build` seems unnecessary here.
- In case of a release build, the `version.properties` file contains `Release Build`
- In case of a snapshot build, the `version.properties` file contains `Build #3814`

In all cases, we hence do not need the word `build` before it.


wborn (approved):
Makes sense!

PR #3994, opened 2024-01-02
Replace or remove assert statements

wborn:
Java assertions are disabled by default so in this PR they are replaced/removed where applicable.


J-N-K (approved):
LGTM

PR #3995, opened 2024-01-02
Fix assertEquals order

wborn:
The first parameter should be the expected value and the second parameter the actual value.


J-N-K (approved):
Thanks!

PR #3996, opened 2024-01-02
Simplify assertions

wborn:
Using the appropriate assertion methods results in less and easier to read code as well as better error messages when assertions fail.

### DIFF
--- a/bundles/org.openhab.core/src/test/java/org/openhab/core/auth/client/oauth2/AccessTokenResponseTest.java
+++ b/bundles/org.openhab.core/src/test/java/org/openhab/core/auth/client/oauth2/AccessTokenResponseTest.java
@@ -76,6 +76,6 @@ public class AccessTokenResponseTest {
         // token has a life time of 60 seconds
         token.setExpiresIn(60);
 
-        assertTrue(!token.isExpired(Instant.now(), 10), "Token should have been expired due to buffer");
+        assertFalse(token.isExpired(Instant.now(), 10), "Token should have been expired due to buffer");
     }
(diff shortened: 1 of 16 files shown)


bundles/org.openhab.core/src/test/java/org/openhab/core/auth/client/oauth2/AccessTokenResponseTest.java
```diff
@@ -76,6 +76,6 @@ public void testTokenNotExpiredEvenWithBuffer() {
         // token has a life time of 60 seconds
         token.setExpiresIn(60);
 
-        assertTrue(!token.isExpired(Instant.now(), 10), "Token should have been expired due to buffer");
+        assertFalse(token.isExpired(Instant.now(), 10), "Token should have been expired due to buffer");
```
J-N-K, line 79: Isn't the test and the message a contradiction? We test `isExpired == false` and the failure message says the token should have expired?

J-N-K (changes requested):
Otherwise LGTM

bundles/org.openhab.core/src/test/java/org/openhab/core/auth/client/oauth2/AccessTokenResponseTest.java
wborn: Fixed the text. I think it was copied from `testTokenExpiredDueToBuffer` to `testTokenNotExpiredEvenWithBuffer`

J-N-K (approved):

PR #3997, opened 2024-01-02
Remove redundant array creation for calling varargs methods

wborn:
These array creations are unnecessary because arrays are created automatically for methods using varargs.


J-N-K (approved):
LGTM

PR #3999, opened 2024-01-02
Use isEmpty instead of 0 comparisons

wborn:
`isEmpty()` expresses the intent more clearly and is therefore preferred. Counting the number of elements can also be an expensive operation e.g. when using linked lists.


J-N-K (approved):
LGTM, thanks!

PR #4000, opened 2024-01-02
Remove redundant modifiers

wborn:
Removes redundant modifiers from the code.
These modifiers redeclare the default modifiers that apply to interfaces, enums etc.


J-N-K (approved):
Thanks for the cleanup

PR #4001, opened 2024-01-02
Use diamond operator

wborn:
Often the type can be inferred so the diamond operator can be used to simplify the code.


J-N-K (approved):
Thanks!

PR #4002, opened 2024-01-02
Use static inner classes

wborn:
A static inner class does not keep an implicit reference to its enclosing instance. This prevents a common cause of memory leaks and uses less memory per instance of the class.


J-N-K (approved):
Thanks for alla your cleanups

PR #4003, opened 2024-01-02
Iterate using Map entries

wborn:
Iteration using Map entries is preferred because it is more efficient and helps preventing NPEs.

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/provider/RuleResourceBundleImporter.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/provider/RuleResourceBundleImporter.java
@@ -151,8 +151,7 @@ public class RuleResourceBundleImporter extends AbstractResourceBundleProvider<R
         if (portfolio == null) {
             for (Entry<Vendor, List<String>> entry : providerPortfolio.entrySet()) {
                 if (entry.getKey().getVendorSymbolicName().equals(vendor.getVendorSymbolicName())) {
-                    List<String> vendorPortfolio = entry.getValue();
-                    return vendorPortfolio == null ? List.of() : vendorPortfolio;
+                    return entry.getValue();
                 }
             }
         }
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/provider/RuleResourceBundleImporter.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/provider/RuleResourceBundleImporter.java
@@ -17,6 +17,7 @@ import java.io.IOException;
 import java.net.URL;
 import java.util.Enumeration;
 import java.util.List;
+import java.util.Map.Entry;
 import java.util.Set;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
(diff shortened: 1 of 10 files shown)


bundles/org.openhab.core.io.transport.upnp/src/main/java/org/openhab/core/io/transport/upnp/internal/UpnpIOServiceImpl.java
```diff
@@ -154,11 +155,11 @@ protected void eventReceived(GENASubscription sub) {
                     deviceRoot.getIdentity().getUdn());
             for (UpnpIOParticipant participant : participants) {
                 if (Objects.equals(getDevice(participant), deviceRoot)) {
-                    for (String stateVariable : values.keySet()) {
-                        StateVariableValue value = values.get(stateVariable);
+                    for (Entry<String, StateVariableValue> entry : values.entrySet()) {
+                        StateVariableValue value = entry.getValue();
                         if (value.getValue() != null) {
```
J-N-K, line 160: I think this null-check can be omitted if we use `entry.getValue()` because the map has no null values.

bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/ConfigDescriptionValidatorImpl.java
```diff
@@ -101,8 +102,9 @@ public void validate(Map<String, Object> configurationParameters, URI configDesc
 
         Collection<ConfigValidationMessage> configDescriptionValidationMessages = new ArrayList<>();
 
-        for (String key : map.keySet()) {
-            ConfigDescriptionParameter configDescriptionParameter = map.get(key);
+        for (Entry<String, ConfigDescriptionParameter> entry : map.entrySet()) {
+            String key = entry.getKey();
+            ConfigDescriptionParameter configDescriptionParameter = entry.getValue();
             if (configDescriptionParameter != null) {
```
J-N-K, line 108: I think this null-check can be omitted if we use entry.getValue() because the map has no null values.

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/provider/RuleResourceBundleImporter.java
```diff
@@ -148,9 +149,9 @@ protected void addNewProvidedObjects(List<String> newPortfolio, List<String> pre
     protected List<String> getPreviousPortfolio(Vendor vendor) {
         List<String> portfolio = providerPortfolio.get(vendor);
         if (portfolio == null) {
-            for (Vendor v : providerPortfolio.keySet()) {
-                if (v.getVendorSymbolicName().equals(vendor.getVendorSymbolicName())) {
-                    List<String> vendorPortfolio = providerPortfolio.get(v);
+            for (Entry<Vendor, List<String>> entry : providerPortfolio.entrySet()) {
+                if (entry.getKey().getVendorSymbolicName().equals(vendor.getVendorSymbolicName())) {
+                    List<String> vendorPortfolio = entry.getValue();
                     return vendorPortfolio == null ? List.of() : vendorPortfolio;
```
J-N-K, line 155:
```suggestion
                    return entry.getValue();
```

The map does not contain null values

J-N-K (changes requested):

J-N-K (approved):
Thanks!

bundles/org.openhab.core.io.transport.upnp/src/main/java/org/openhab/core/io/transport/upnp/internal/UpnpIOServiceImpl.java
morph166955: I'm getting NPE's I believe due to the removal of the null check here.  I've opened #4037 to track that.

PR #4004, opened 2024-01-02
Fix JavaDoc issues

wborn:
Fixes various issues including wrong parameter names, references, links and dangling JavaDocs.

### DIFF
--- a/bundles/org.openhab.core.addon/src/main/java/org/openhab/core/addon/internal/xml/AddonDiscoveryMethodConverter.java
+++ b/bundles/org.openhab.core.addon/src/main/java/org/openhab/core/addon/internal/xml/AddonDiscoveryMethodConverter.java
@@ -22,6 +22,7 @@ import org.openhab.core.addon.AddonParameter;
 import org.openhab.core.config.core.xml.util.GenericUnmarshaller;
 import org.openhab.core.config.core.xml.util.NodeIterator;
 
+import com.thoughtworks.xstream.converters.Converter;
 import com.thoughtworks.xstream.converters.UnmarshallingContext;
 import com.thoughtworks.xstream.io.HierarchicalStreamReader;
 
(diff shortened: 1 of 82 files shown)


bundles/org.openhab.core.addon/src/main/java/org/openhab/core/addon/internal/xml/AddonDiscoveryMethodConverter.java
```diff
@@ -22,6 +22,7 @@
 import org.openhab.core.config.core.xml.util.GenericUnmarshaller;
 import org.openhab.core.config.core.xml.util.NodeIterator;
 
+import com.thoughtworks.xstream.converters.Converter;
```
J-N-K, line 25: Why do you use an import here instead of using the fully qualified class name  (like in the other classes)? IIRC IntelliJ complains about "unused imports" if you import classes that are only used in JavaDoc

wborn: I found it was fixed similarly in other classe like `BridgeTypeConverter` and `ChannelConverter`.
Do you see the warning there as well?
I don't see this in my IntelliJ, but maybe we use a different configuration. :slightly_smiling_face:

J-N-K (approved):

bundles/org.openhab.core.addon/src/main/java/org/openhab/core/addon/internal/xml/AddonDiscoveryMethodConverter.java
J-N-K: Looks good. I'm pretty sure there was an issue with that, either spotless removed the import (because it was "unused") or IntelliJ complained. In the first case it could be that this was solved by one of the latest spotless upgrades. Or it may depend on if other classes from the same package are imported, too.

PR #4007, opened 2024-01-03
Fix `SerialPort.getFlowControlMode()` infinite recursion

wborn:
An infinite recursion would occur when calling this method. I did not find any add-ons using this method but that could be due to this bug.


J-N-K (approved):
Thanks!

PR #4010, opened 2024-01-03
Use protected modifier with constructor of abstract classes

wborn:
Abstract classes should not have public constructors.
Constructors of abstract classes can only be called in constructors of their subclasses.
So there is no point in making them public.
The protected modifier should be enough.


J-N-K (approved):
Thanks

PR #4013, opened 2024-01-04
IpAddonFinder: Skip installed addons

### LINKED ISSUE #3967: Prevent addon suggestion finder to scan for installed plugins
Currently, the new addon suggestion finders will look for all addons with provide detection metods in ```addon(s).xml```.

Especially the IP scan which does active scanning, this should not be done if the corresponding binding is already installed.
It could cause side effects if an addon needs to listen to a specific port etc.

Currently, we detect everything, and inhibit the suggestion to be shown if the plugin is installed.

The question is if we want to solve this on a global level - providing less suggestions to the specific finder - or in the finders itself.

@mherwege @andrewfg WDYT?

holgerfriedrich:
Query AddonService to check if an addon is already installed. In case, skip scanning for the addon.

Fixes #3967

### DIFF
--- a/bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
+++ b/bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
@@ -36,6 +36,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.StringTokenizer;
+import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
@@ -53,10 +54,11 @@ import org.openhab.core.common.ThreadPoolManager;
 import org.openhab.core.config.discovery.addon.AddonFinder;
 import org.openhab.core.config.discovery.addon.BaseAddonFinder;
 import org.openhab.core.net.NetUtil;
-import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Deactivate;
 import org.osgi.service.component.annotations.Reference;
+import org.osgi.service.component.annotations.ReferenceCardinality;
+import org.osgi.service.component.annotations.ReferencePolicy;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -152,14 +154,12 @@ public class IpAddonFinder extends BaseAddonFinder {
     private final Logger logger = LoggerFactory.getLogger(IpAddonFinder.class);
     private final ScheduledExecutorService scheduler = ThreadPoolManager
             .getScheduledPool(ThreadPoolManager.THREAD_POOL_NAME_COMMON);
-    private final AddonService addonService;
+    private final Set<AddonService> addonServices = new CopyOnWriteArraySet<>();
     private @Nullable Future<?> scanJob = null;
     Set<AddonInfo> suggestions = new HashSet<>();
 
-    @Activate
-    public IpAddonFinder(@Reference AddonService as) {
+    public IpAddonFinder() {
         logger.trace("IpAddonFinder::IpAddonFinder");
-        this.addonService = as;
         // start of scan will be triggered by setAddonCandidates to ensure addonCandidates are available
     }
 
@@ -174,6 +174,15 @@ public class IpAddonFinder extends BaseAddonFinder {
         logger.debug("IpAddonFinder::setAddonCandidates({})", candidates.size());
         super.setAddonCandidates(candidates);
         startScan();
+    }
+
+    @Reference(cardinality = ReferenceCardinality.MULTIPLE, policy = ReferencePolicy.DYNAMIC)
+    protected void addAddonService(AddonService featureService) {
+        this.addonServices.add(featureService);
+    }
+
+    protected void removeAddonService(AddonService featureService) {
+        this.addonServices.remove(featureService);
     }
 
     private void startScan() {
@@ -205,8 +214,7 @@ public class IpAddonFinder extends BaseAddonFinder {
                 logger.trace("Checking candidate: {}", candidate.getUID());
 
                 // skip scanning if already installed
-                Addon addon = addonService.getAddon(candidate.getUID(), null);
-                if (addon != null && addon.isInstalled()) {
+                if (isAddonInstalled(candidate.getUID())) {
                     logger.trace("Skipping {}, already installed", candidate.getUID());
                     continue;
                 }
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -197,6 +204,16 @@ private void scan() {
 
                 logger.trace("Checking candidate: {}", candidate.getUID());
 
+                // skip scanning if already installed
+                if (addonService != null) {
+                    @Nullable
```
mherwege, line 209: Is this annotation required?

mherwege (approved):
Otherwise LGTM

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
holgerfriedrich: I removed it, SAT does not complain without.
To get rid of the last SAT warning I had to introduce a temporary copy of the AddonService :-/

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -148,11 +152,14 @@ public class IpAddonFinder extends BaseAddonFinder {
     private final Logger logger = LoggerFactory.getLogger(IpAddonFinder.class);
     private final ScheduledExecutorService scheduler = ThreadPoolManager
             .getScheduledPool(ThreadPoolManager.THREAD_POOL_NAME_COMMON);
+    private final @Nullable AddonService addonService;
```
wborn, line 155: It's always injected with a non-null argument in the constructor:

```suggestion
    private final AddonService addonService;
```

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -197,6 +204,16 @@ private void scan() {
 
                 logger.trace("Checking candidate: {}", candidate.getUID());
 
+                // skip scanning if already installed
+                AddonService tmpAddonService = addonService;
+                if (tmpAddonService != null) {
```
wborn, line 209: This can be removed when `addonService` is final and non-null.

```suggestion
```

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -205,13 +205,10 @@ private void scan() {
...
 
                 // skip scanning if already installed
-                AddonService tmpAddonService = addonService;
-                if (tmpAddonService != null) {
-                    Addon a = tmpAddonService.getAddon(candidate.getUID(), null);
-                    if (a != null && a.isInstalled()) {
-                        logger.trace("Skipping {}, already installed", candidate.getUID());
-                        continue;
-                    }
+                Addon addon = addonService.getAddon(candidate.getUID(), null);
```
wborn, line 208: There are several `AddonService` implementations so you may need to check them all like the `AddonResource` does.

holgerfriedrich: I'll have a look. Is it really that complicated to find out if an addon is installed?

wborn: Yes because there has not been anything else that needed all the logic that is in the `AddonResource` so far.

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -356,4 +364,14 @@ public Set<AddonInfo> getSuggestedAddons() {
     public String getServiceName() {
         return SERVICE_NAME;
     }
+
+    private boolean isAddonInstalled(String addonId) {
```
mherwege, line 368: Maybe we should move this to BaseAddonFinder. It would allow to use this in other add-on finders as well.

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
holgerfriedrich: @wborn Thanks for pointing me to the right place. I reworked my PR, could you have a look please?

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
holgerfriedrich: Sure, if it is needed, I can do so. I had the impression, that only IP finder will skip searching if an addon is already installed.

Should we move it, although it introduces some overhead to the other finders? OSGI would then call register all addonservices to BaseAddonFinder using
`@Reference(cardinality = ReferenceCardinality.MULTIPLE, policy = ReferencePolicy.DYNAMIC)
    protected void addAddonService(AddonService featureService) {`...

If we want to share it, maybe extending AddonResource would be an option.

@wborn Maybe you could comment on this as well....

wborn: > If we want to share it, maybe extending AddonResource would be an option.

The `AddonResource` is in the org.openhab.core.io.rest.core bundle which is not the type of bundle you want other bundles to depend on. That's because that bundle depends on JAX-RS, Swagger etc.

It might be better to add something reusable to the org.openhab.core.addon bundle. It already provides the `AddonService` interface so anything that needs to interact with `AddonService`s already depends on it. E.g. an `AddonServiceRegistry` could be added there that makes the logic of the  `AddonResource` more reusable. The `AddonResource` could then also use it.

holgerfriedrich: Ok, good to know. Actually, an ```AddonServiceRegistry``` was the first thing I was looking for - and wondered that it is not yet there.

However, hhis is clearly out of scope of this PR.
Can we move forward with this one as it is for now?

wborn (approved):
Thanks for the improvement!

PR #4016, opened 2024-01-05
Fix CurrencyUnit

J-N-K:
This fixes some issues that were recently detected regarding the new currency units:

- Currency codes could not be used for the base currency (but for additional currencies) when a symbol is defined. The reason is that in this case only the symbol and not symbol and label were added to the unit formatter.

- Units composed of a currency and an additional currency (like `EUR/kWh`) were not properly converted (only the non-currency part was converted). This was caused by two problems: `CurrencyUnit` did not inherit from `AbstractUnit` and therefore the converter was not properly found (always `IDENTITY`). After I fixed that, an additional problem occurred: the `CurrencyConverter` could not be used because the list of converters is a final, non-accessible list.

@jlaur FYI.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/library/unit/CurrencyConverter.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/library/unit/CurrencyConverter.java
@@ -12,8 +12,10 @@
  */
 package org.openhab.core.internal.library.unit;
 
+import java.lang.reflect.Field;
 import java.math.BigDecimal;
 import java.math.MathContext;
+import java.util.Map;
 import java.util.Objects;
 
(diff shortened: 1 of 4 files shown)


jlaur:
@J-N-K, @openhab/core-maintainers - perhaps something to be considered for 4.1.1 since it fixes an issue in a new 4.1 feature?

kaikreuzer:
@J-N-K Should this be backported to 4.1.x?

J-N-K:
Yes, especially because of the symbol/code issue. The conversion bug will probably not be an issue on 4.1 because there is no currency provider with more than one currency.

kaikreuzer:
Ok, cherry-picked to `4.1.x`.

PR #4019, opened 2024-01-05
Improve marketplace add-on handling

J-N-K:
It has been reported several times that add-ons were not properly installed / missing after an upgrade or the installation of incompatible add-ons resulted in broken installations.

After an upgrade (or clean cache) the `AddonHandler`s try to re-install the add-ons from the download cache (`<userdata>/marketplace`). This happens without checking compatibility. This was needed before OH4, because the cache was the only source providing information about installed add-ons. This is now different, since we store the add-on information in a JSON database, so the UIDs of the add-ons are known.

This PR changes improves the add-on services. It now

1. Reads the information about the installed add-ons from the database and sets the installation status based on information from the handlers.
2. Removes all add-ons that are not installed from the JSON database and remembers their UIDs.
3. Refreshes the remote add-on list (including check for compatibility if not disabled).
4. Tries installation of the add-ons remembered in step 2. Since incompatible add-ons are missing in the add-on list, their installation fails and a warning is logged.

This PR is has two corresponding PR in openhab-distro and openhab-linuxpkg to ensure that the upgrade script and `openhab-cli` also clear the marketplace cache.

### DIFF
--- a/bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/AbstractRemoteAddonService.java
+++ b/bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/AbstractRemoteAddonService.java
@@ -12,6 +12,8 @@
  */
 package org.openhab.core.addon.marketplace;
 
+import static org.openhab.core.common.ThreadPoolManager.THREAD_POOL_NAME_COMMON;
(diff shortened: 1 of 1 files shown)


wborn (approved):
Great! This certainly helps preventing incompatiblity issues resulting from outdated Marketplace add-ons. :+1:

PR #4020, opened 2024-01-05
Simplify Map operations using `computeIfAbsent`

wborn:
If the specified key is not already associated with a value (or is mapped to null), the given mapping function computes the value.

### DIFF
--- a/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/RuleSupportScriptExtension.java
+++ b/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/RuleSupportScriptExtension.java
@@ -18,6 +18,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
@@ -154,11 +155,8 @@ public class RuleSupportScriptExtension implements ScriptExtensionProvider {
             return obj;
         }
 
-        Map<String, Object> objects = objectCache.get(scriptIdentifier);
-        if (objects == null) {
-            objects = new HashMap<>();
-            objectCache.put(scriptIdentifier, objects);
-        }
+        Map<String, Object> objects = Objects
+                .requireNonNull(objectCache.computeIfAbsent(scriptIdentifier, k -> new HashMap<>()));
 
         obj = objects.get(type);
         if (obj != null) {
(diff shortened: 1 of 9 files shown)


bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/RuleSupportScriptExtension.java
```diff
@@ -154,11 +154,7 @@ public Collection<String> getTypes() {
             return obj;
         }
 
-        Map<String, Object> objects = objectCache.get(scriptIdentifier);
-        if (objects == null) {
-            objects = new HashMap<>();
-            objectCache.put(scriptIdentifier, objects);
-        }
+        Map<String, Object> objects = objectCache.computeIfAbsent(scriptIdentifier, k -> new HashMap<>());
```
J-N-K, line 157: The problem is that this always creates a warning in the compiler because the EEA can't properly model "return value is non-null if second parameter is non-null" and assumes the return value is nullable.

Unfortunately the ECJ guys don't see the problem or the need to improve that.

jlaur: We had the same situation recently here: https://github.com/openhab/openhab-addons/pull/15828#discussion_r1382012932. Maybe a similar work-around could be used here as well.

J-N-K (approved):
Thanks!

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleRegistryImpl.java
```diff
@@ -376,11 +377,8 @@ private Rule resolveRuleByTemplate(Rule rule) {
...
     private void updateRuleTemplateMapping(String templateUID, String ruleUID, boolean resolved) {
         synchronized (this) {
-            Set<String> ruleUIDs = mapTemplateToRules.get(templateUID);
-            if (ruleUIDs == null) {
-                ruleUIDs = new HashSet<>();
-                mapTemplateToRules.put(templateUID, ruleUIDs);
-            }
+            Set<String> ruleUIDs = Objects
+                    .requireNonNull(mapTemplateToRules.computeIfAbsent(templateUID, k -> new HashSet<>()));
             if (resolved) {
```
Nadahar, line 382: I'm sorry, but I completely fail to see how this is simplifying the code. I doubt there are any performance benefits, and the old code is certainly much clearer as to what actually happens. Especially, `Objects.requireNonNull()` obscures the whole thing, and although cheap, it still is a completely useless operation that does require pushing and popping the stack etc.

I think a better description would be "Trade readability for slightly shorter code".

PR #4021, opened 2024-01-05
Some more code cleanup

wborn:
This cleanup includes:

* Use enhanced for loops
* Use text blocks
* Use Objects.equals
* Fix some typos
* Remove redundant variable initialization
* Remove redundant null checks with instanceof
* Remove redundant thrown Exceptions
* Remove redundant empty String concatenation

### DIFF
--- a/bundles/org.openhab.core.ui.icon/src/test/java/org/openhab/core/ui/icon/internal/IconServletTest.java
+++ b/bundles/org.openhab.core.ui.icon/src/test/java/org/openhab/core/ui/icon/internal/IconServletTest.java
@@ -66,7 +66,7 @@ public class IconServletTest {
         }
 
         public String getOutput() {
-            return new String(outputStream.toByteArray());
+            return outputStream.toString();
         }
 
         public void reset() {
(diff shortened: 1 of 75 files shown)


bundles/org.openhab.core.ui.icon/src/test/java/org/openhab/core/ui/icon/internal/IconServletTest.java
```diff
@@ -66,7 +66,7 @@ public boolean isReady() {
         }
 
         public String getOutput() {
-            return new String(outputStream.toByteArray());
+            return outputStream.toString();
```
J-N-K, line 69: Is this guaranteed to be the same?

J-N-K (commented):
Otherwise LGTM

bundles/org.openhab.core.ui.icon/src/test/java/org/openhab/core/ui/icon/internal/IconServletTest.java
wborn: Yes it saves creating a copy of the byte array:

## ByteArrayOutputStream

```java
public synchronized byte[] toByteArray() {
    return Arrays.copyOf(buf, count);
}

public synchronized String toString() {
    return new String(buf, 0, count);
}
```

## String

```java
public String(byte[] bytes) {
    this(bytes, 0, bytes.length);
}
```

J-N-K (approved):
Thanks!

PR #4024, opened 2024-01-06
Improve YAML model repository

J-N-K:
Related to #3666 

This improves the existing YAML model parser and repository. In detail:

- `YamlFile` has been removed. This simplifies code for consumers, because they only need to implement `YamlModelListener` and the DTO as inheritor of `YamlDTO`.
- `getRootName` has been removed from `YamlModelListener`. The information is now provided by the `@YamlElementName` annotation on the respective DTO class.
- The `addedModel`, `updatedModel` and `removedModel` methods in `YamlModelListener` have been changed to accept `Collection<T>` instead of `Collection<? extends YamlDTO>` which simplifies code in the listener.
- The `YamlModelRepository` has been refactored to an interface with methods for adding/updating/removing elements from a model. The implementation is `YamlModelRepositoryImpl`. This interface is public and can be implemented to modify an existing model or create a new one. This allows to dynamically update the content.
- The `YamlModelRepositoryImpl` now parses the elements using `JsonNode` instead of de-serializing the file directly. `JsonNode`s are cached by model (file) name and type. This allows to mix different inheritors of `YamlDTO` in one file. Since all elements are cached as `JsonNode`, they can also be read before the respective `YamlModelListener` is registered. In addition, this removes the coupling between directory and model type.
- Tests for the `YamlModelRepository` have been added.
- A `readOnly` flag (defaults to `true`) has been added. Models without this flag (or `readOnly: true`) cannot be modified. This is to prevent changing user-defined models. On deserialization comments in the file are lost, so they would be missing in an updated model.

@lolodomo I would especially like to hear your comments since you did the original implementation.


J-N-K:
@wborn How do we proceed here?

bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/YamlElement.java
```diff
@@ -12,28 +12,54 @@
...
+ * <li>MUST have a default constructor to allow deserialization with Jackson</li>
+ * <li>MUST provide {@code equals(Object other)} and {@code hashcode()} methods</li>
+ * <li>MUST be annotated with {@link YamlElementName} containing the element name</li>
+ * <li>SHOULD implement a proper {@code toString()} method</li>
+ * </ul>
  *
  * @author Laurent Garnier - Initial contribution
+ * @author Jan N. Klug - Refactoring and improvments to JavaDoc
  */
-@NonNullByDefault
```
wborn, line 23: Why was this removed?

bundles/org.openhab.core.model.yaml/src/test/resources/modelFileUpdatePre.yaml
```diff
@@ -0,0 +1,11 @@
...
+
+version: 1
+
+firstType:
+  - uid: First
+    description: First original
+  - uid: Second
+    description: Second modified
+  - uid: Fourth
+    description: Fourth original
```
wborn, line 11: Can you add the missing new lines to those .yaml  files?

bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/YamlElement.java
J-N-K: This essentially a DTO. Missing elements in the data are de-serialized as `null`.

If we decide to make the DTO classes `@NonNullByDefault` we would need to either annotate all fields with `@Nullable` (which is the similar to not annotating the class at all) or provide default values (which doesn't make sense in most cases).

wborn (approved):
Thanks for the improvements!

lolodomo:
I did not take time to have a look. Does this PR break user YAML files containing custom tags ?

J-N-K:
No, I don't think so. The format is the same, just some more options added and code refactoring

PR #4025, opened 2024-01-07
S-Bus transport protocol. Derived from Modbus protocol

cipianpascu:
This is the first attempt to implement the S-Bus G4 transport protocol.
The code is derived from the Modbus transport protocol and has the following two additions:
1. TCP protocol is replaced with UDP protocol
2. All the protocol queries have a subnetId  byte in addition to the deviceId and functionId.
Please have a look to what I did, and let me know about the additional changes/fixes required for this new transport protocol/binding.


ssalonen:
I find this hard to review, the modifications to modbus are not in a separate commit. Can you point to the place where the protocol differs?

would it be possible to extend the modbus transport and binding to support also this case?

cipianpascu:
Hey Sami,

True, due to package renaming, it is not that easy to indentify the core changes. But I will try to highlight them.
First, I introduced the subnetId in j2sbus (forked from jamod) library:
https://github.com/cipianpascu/j2sbus/blob/feature/openhab/src/main/java/ro/ciprianpascu/sbus/msg/ModbusMessage.java#L75
https://github.com/cipianpascu/j2sbus/blob/feature/openhab/src/main/java/ro/ciprianpascu/sbus/ModbusCoupler.java#L145
paying attention to the length of the serialized message:
https://github.com/cipianpascu/j2sbus/blob/feature/openhab/src/main/java/ro/ciprianpascu/sbus/io/ModbusUDPTransport.java#L80

After that, I introduced the subnetId in the io.transport.sbus module (ModbusReadRequestBlueprint, ModbusWriteRegisterRequestBlueprint, ModbusWriteCoilRequestBlueprint) so that I can transfer the subnetId from the thing to the sbus requests in the ModbusLibraryWrapper:
https://github.com/openhab/openhab-core/pull/4025/files#diff-3a65c1afa0e4d747416822744bf25092732b40984a277583834f53a06b7d9545
https://github.com/openhab/openhab-core/pull/4025/files#diff-d88be59adf03f3211a19cf34f2cc0598e9d28b5ad974e919a44a1475e9296875
https://github.com/openhab/openhab-core/pull/4025/files#diff-c8270491534da242e8da50a1adc400087d11fc95fa36d0e41b5dc8896629f724

https://github.com/openhab/openhab-core/pull/4025/files#diff-ee24094074fe145ec56143be6e0806f7317702c193cc971cd502594071ed3445

Given the above changes, I don't see a way at this moment to appply the sbus changes to jamod library without breaking the existing functionality (and modbus specs). But it might be a posibility to add j2sbus library as a dependency to io.transport.modbus bundle and then ammend the ModbusLibraryWrapper.createRequest methods with a new if branch around the subnetId and calling. But this time, I find it a bit problematic to deal with the ro.ciprianpascu.sbus.msg.ModbusRequest interface.

Also, I created a [unifiedDiff.zip](https://github.com/openhab/openhab-core/files/14071693/unifiedDiff.zip), which ignores all the package and comments changes. This way, your review process should become easier.
Given the above, if you still believe that merging my changes into the modbus implementation is a good option, then I could give it a try.

Thank you,
Ciprian

cipianpascu:
I created a new binding from scratch targeting the SBUS protocol: https://github.com/openhab/openhab-addons/pull/18019
This PR is no longer needed.

PR #4026, opened 2024-01-07
[discovery] Refactor `java.util.Date` usages to `java.time.Instant`

jlaur:
This refactors some [java.util.Date](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Date.html) usages to the more modern [time API](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/time/package-summary.html).

The API is preserved as epoch milliseconds, so this is a non-breaking change.

There is two user-facing changes, though.

1. This console command has changed formatting:
```shell
openhab> openhab:inbox list
NEW [mydevice]: My Device [thingId=my:device:0, bridgeId=null, properties={ipAddress=192.168.0.1}, representationProperty=null, timestamp=2024-01-07T15:33:15.500Z, timeToLive=UNLIMITED]
```

Previously this would be displayed like this:
```shell
openhab> openhab:inbox list
NEW [mydevice]: My Device [thingId=my:device:0, bridgeId=null, properties={ipAddress=192.168.0.1}, representationProperty=null, timestamp=Sat Nov 11 16:33:15 CET 2023, timeToLive=UNLIMITED]
```

Since this is already quite technical, I didn't see a reason to invest in preserving the old formatting.

2. Changed debug logging

This is now UTC-formatted:
> Removed thing 'mydevice' from inbox because it was older than 2024-01-07T15:33:15.500Z

### DIFF
--- a/bundles/org.openhab.core.config.discovery/src/main/java/org/openhab/core/config/discovery/internal/DiscoveryResultImpl.java
+++ b/bundles/org.openhab.core.config.discovery/src/main/java/org/openhab/core/config/discovery/internal/DiscoveryResultImpl.java
@@ -12,8 +12,8 @@
  */
 package org.openhab.core.config.discovery.internal;
 
+import java.time.Instant;
 import java.util.Collections;
-import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
@@ -43,7 +43,7 @@ public class DiscoveryResultImpl implements DiscoveryResult {
     private @Nullable String representationProperty;
     private @NonNullByDefault({}) DiscoveryResultFlag flag;
     private @NonNullByDefault({}) String label;
-    private long timestamp;
+    private Instant timestamp = Instant.MIN;
     private long timeToLive = TTL_UNLIMITED;
 
     /**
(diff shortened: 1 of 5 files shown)


J-N-K (approved):
Thanks. IMO the new formatting is even an improvement compression to the one before.

bundles/org.openhab.core.config.discovery/src/main/java/org/openhab/core/config/discovery/internal/DiscoveryResultImpl.java
```diff
@@ -43,7 +43,7 @@ public class DiscoveryResultImpl implements DiscoveryResult {
     private @Nullable String representationProperty;
     private @NonNullByDefault({}) DiscoveryResultFlag flag;
     private @NonNullByDefault({}) String label;
-    private long timestamp;
+    private Instant timestamp = Instant.MIN;
```
jlaur, line 46: @J-N-K - unfortunately I ran into this nasty regression today:
```
2024-01-08 21:43:22.754 [ERROR] [re.storage.json.internal.JsonStorage] - Couldn't deserialize value 'org.openhab.core.storage.json.internal.StorageEntry@3df4a8fa'. Root cause is: java.lang.IllegalStateException: Expected BEGIN_OBJECT but was NUMBER at path $.timestamp
```

so now realize that this object is deserialized from the JSON file `userdata/org.openhab.core.config.discovery.DiscoveryResult.json`.

The easy solution is to revert this variable back to `long`. Alternately we can keep `Instant`, but implement custom deserialization that can handle both. WDYT? Right at this moment I'm not sure where this is deserialized.

J-N-K:
Write a converter for the JSON storage.

PR #4027, opened 2024-01-08
[discovery] Fix deserialization of DiscoveryResult

jlaur:
Partially revert #4026

See https://github.com/openhab/openhab-core/pull/4026/files#r1445323983

Even with this fix, there will be issues for discovery results created by the snapshot version today. A better solution might be to provide a custom deserializer for supporting both epoch and UTC, but I'm not sure how to do this right now for the storage service. I'm providing this PR as one option for fixing the regression.


jlaur:
This quickfix will hopefully be superceded by #4029.

PR #4032, opened 2024-01-10
[WIP] [audio/voice] Add pcm audio websocket with dialog support

GiviMAD:
These PR adds a WebSocket Adapter that allows transferring PCM audio to a sink and source that whose existence will be tied to the WebSocket connection (register/unregister the components on connection/disconnection) also allows to spawn a dialog processor instance connected to those (will also be tear down on disconnection).

The PR is incomplete and untested, I'll try to add a client to the UI and develop both PR in parallel. 

Also IDK if placing the code in the voice bundle is correct, I did it there because it already requires the audio bundle. For me it makes more sense to do it on the audio bundle but I'll have to require both the voice and websocket bundles there.

Best regards!


kaikreuzer:
> Also IDK if placing the code in the voice bundle is correct

As the websocket adapter is for audio streams in general, I would say that it should ideally be placed in the audio bundle.
The dependency on the voice manager is very small; maybe you could come up with some small hook, which could be used by the voice bundle to register the dialog processing logic into the websocket adapter?

GiviMAD:
> > Also IDK if placing the code in the voice bundle is correct
> 
> As the websocket adapter is for audio streams in general, I would say that it should ideally be placed in the audio bundle. The dependency on the voice manager is very small; maybe you could come up with some small hook, which could be used by the voice bundle to register the dialog processing logic into the websocket adapter?

Yes, having it on the voice bundle is quite unintuitive, I haven't though on that solution seems like the correct one.

Thank you for the feedback!

GiviMAD:
I have to refactor the code into the websocket.io bundle for it to build without modifying the BOM features, but I think it also makes sense to have it there with the other WebSocketAdapter implementations.

florian-h05:
@openhab/core-maintainers Would be great if this could be reviewed for openHAB 5, as it would allow us to have a voice assistant inside the UI.

florian-h05:
What would be nice is to have the same WebSocket API format as Home Assistant so we could use ESPHome voice satellites. I don’t know how feasible that is, I think I found HA‘s voice assistant WS endpoint implementation here: https://github.com/home-assistant/core/blob/dev/homeassistant/components/assist_pipeline/websocket_api.py

kaikreuzer:
> Would be great if this could be reviewed for openHAB 5, as it would allow us to have a voice assistant inside the UI.

I fully agree. Sorry @GiviMAD for having neglected that topic over the past months - I'll do my very best to drive this forward again for 5.0 (i.e. right after the 4.3 release)!

GiviMAD:
Hey @kaikreuzer and @florian-h05 sorry for the late reply. I have rebased the PR and done some changes to simplify its usage. Basically I have moved the audio conversion to the server. If you have any time to review it will be great having it available on the snapshots.

Also @dalgwen, I forgot to mention you here but if you can take a look in case you see something weird, it will be great.

Best regards!

florian-h05:
Thanks!

Did you have a look at HA‘s voice protocol implementation? API compatibility on the openHAB side would have the benefit that we could make use of HA‘s work in voice satellites, which would make things a lot easier for us as we would have voice satellites already available without the need to take care of developing them.

florian-h05:
Thanks for the explanation - I'm fine with that , just wanted to know what you think about it.

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/audiopcm/PCMWebSocketAudioSource.java
```diff
@@ -0,0 +1,200 @@
...
+import org.openhab.core.audio.AudioFormat;
+import org.openhab.core.audio.AudioSource;
+import org.openhab.core.audio.AudioStream;
+import org.openhab.core.audio.PipedAudioStream;
+import org.openhab.core.common.ThreadPoolManager;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This is an {@link AudioSource} implementation connected to the {@link PCMWebSocketConnection} that allow to
```
florian-h05, line 38: This is an {@link AudioSource} implementation connected to a {@link PCMWebSocketConnection} supporting a single PCM audio line through a WebSocket connection and shared across all active {@link AudioStream} instances.

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/audiopcm/PCMWebSocketAudioSource.java
```diff
@@ -0,0 +1,200 @@
...
+ * a single pcm audio line through the websocket which is shared across the active {@link AudioStream} instances.
+ *
+ * @author Miguel Álvarez Díez - Initial contribution
+ */
+@NonNullByDefault
+public class PCMWebSocketAudioSource implements AudioSource {
+    private final Logger logger = LoggerFactory.getLogger(PCMWebSocketAudioSource.class);
+    public static int supportedBitDepth = 16;
```
florian-h05, line 46: Please make the static fields final and use UPPERCASE naming.

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/audiopcm/PCMWebSocketAudioSource.java
```diff
@@ -0,0 +1,200 @@
...
+        }
+    }
+
+    public void close() throws Exception {
+        streamGroup.close();
+    }
+
+    public void writeToStreams(byte[] id, int sampleRate, int bitDepth, int channels, byte[] payload) {
+        if (streamGroup.isEmpty()) {
+            logger.debug("Source already disposed ignoring data");
```
florian-h05, line 108:
```suggestion
            logger.debug("Source already disposed, ignoring data");
```

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/audiopcm/PCMWebSocketAudioSource.java
```diff
@@ -0,0 +1,200 @@
...
+            return;
+        }
+        if (this.streamId == null) {
+            this.streamId = id;
+        } else if (!Arrays.equals(this.streamId, id)) {
+            // source is only ready to handle a single data line
+            logger.warn("Ignoring data from source stream {}", id);
```
florian-h05, line 115: Probably add the information that source can only handle a single data line to the log statement.

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/audiopcm/PCMWebSocketAudioUtil.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
+/**
+ * Audio utils.
+ *
+ * @author Miguel Álvarez Díez - Initial contribution
+ */
+@NonNullByDefault
+public class PCMWebSocketAudioUtil {
```
florian-h05, line 28: I am not sure what's the convention we have at openHAB, but normally I would make a util class final and make the default ctor private.

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/audiopcm/PCMWebSocketConnection.java
```diff
@@ -0,0 +1,371 @@
...
+import org.osgi.framework.ServiceRegistration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * The {@link PCMWebSocketConnection} represents a WebSocket connection used to transmit pcm audio
```
florian-h05, line 42:
```suggestion
 * The {@link PCMWebSocketConnection} represents a WebSocket connection used to transmit PCM audio.
```

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/audiopcm/PCMWebSocketConnection.java
```diff
@@ -0,0 +1,371 @@
...
+ * <p>
+ * The websocket uses the text protocol for send commands represented by {@link WebSocketCommand} and the binary
+ * protocol to transmit the audio data
+ * <p>
+ * The websocket supports only one line for the {@link PCMWebSocketAudioSource} (the audio is shared on the server),
+ * the data transmission is instructed by the server (START_LISTENING and STOP_LISTENING commands)
+ * <p>
+ * The websocket supports multiple lines for the {@link PCMWebSocketAudioSink} (to accomplis that, the outgoing data
```
florian-h05, line 50:
```suggestion
 * The websocket supports multiple lines for the {@link PCMWebSocketAudioSink} (to accomplish that, the outgoing data
```

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/audiopcm/PCMWebSocketConnection.java
```diff
@@ -0,0 +1,371 @@
...
+    public boolean isConnected() {
+        Session sess = this.session;
+        return sess != null && sess.isOpen();
+    }
+
+    private synchronized void registerSpeakerComponents(String id, ConnectionOptions clientOptions) throws IOException {
+        if (id.isBlank()) {
+            throw new IOException("Unable to register audio components");
+        }
+        String label = "UI (" + id + ")";
```
florian-h05, line 234: I think it might be better to leave out the UI from the label here, rather use PCM WebSocket (id), as not only the UI might connect.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/KSEdgeService.java
```diff
@@ -41,4 +43,14 @@ default KSServiceHandle spot(KSListener ksListener, AudioStream audioStream, Loc
             throws KSException {
         throw new KSException("An edge keyword spotter is not meant to process audio in the server");
     }
+
+    @Override
+    default Set<AudioFormat> getSupportedFormats() {
```
florian-h05, line 48: I would rather not have default implementations here, as normally I would expect that you need to specify the formats.

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/audiopcm/PCMWebSocketStreamIdUtil.java
```diff
@@ -0,0 +1,64 @@
...
+@NonNullByDefault
+public class PCMWebSocketStreamIdUtil {
+    /**
+     * Packet header length in bytes:
+     * 2 for id
+     * 4 for sample rate as int little-endian
+     * 1 for bitDepth
+     * 1 for channels
+     */
+    public static int packetHeaderByteLength = 2 + 4 + 1 + 1;
```
florian-h05, line 35: Please make this a final constant.
FYI you can also format the JavaDoc with HTML tags so it renders not just inline:

```java
    /**
     * Packet header length in bytes:
     * <ul>
     *   <li>2 for id</li>
     *   <li>4 for sample rate as int little-endian</li>
     *   <li>1 for bitDepth</li>
     *   <li>1 for channels</li>
     * </ul>
     */
```

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/audiopcm/PCMWebSocketStreamIdUtil.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public static AudioPacketData parseAudioPacket(byte[] bytes) {
+        assert (bytes.length >= packetHeaderByteLength);
```
florian-h05, line 38: Please replace this with a proper check and throw an exception.

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/audiopcm/PCMWebSocketStreamIdUtil.java
```diff
@@ -0,0 +1,64 @@
...
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
+/**
+ * Utils to read/write the audio packets send though the websocket binary protocol.
+ *
+ * @author Miguel Álvarez Díez - Initial contribution
+ */
+@NonNullByDefault
+public class PCMWebSocketStreamIdUtil {
```
florian-h05, line 27: As this is a util class, I personally would make it final and make the default ctor private.

florian-h05 (changes requested):
Please have a look at my comments.
Overall the code looks good, but I have to admit that I have only looked at the code and did not think much about the audio PCM details, I am not in to this topic.

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/audiopcm/PCMWebSocketAudioSink.java
```diff
@@ -0,0 +1,265 @@
...
+ * To identify the different audio lines the data chucks are prefixed by a header added by the
+ * {@link PCMWebSocketOutputStream} class.
+ *
+ * @author Miguel Álvarez Díez - Initial contribution
+ */
+@NonNullByDefault
+public class PCMWebSocketAudioSink implements AudioSink {
+    /**
+     * Byte send to the sink after last chunk to indicate that streaming has ended.
+     * Should try to be sent event on and error as the client should be aware that data transmission has ended.
```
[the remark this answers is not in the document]
GiviMAD: I have updated the comment because it was quite misleading.
 ```
     * Byte sent after the last chunk for a stream to indicate the stream has ended, so the client can dispose resources
     * associated with the stream.
     * It's sent on a finally clause that covers the audio transmission execution so it gets sent event if some
     * exception interrupts the audio transmission.
 ```

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/KSEdgeService.java
GiviMAD: I missed these when adding this interface on a previous PR.
To give it some context, the KSEdgeService is meant to represent a keyword spotting service which is running outside the server without using the source audio stream. I created it to allow a way of dialog where the server is not constantly streaming the audio from the source to feed the keyword spotting service.
Looking at it now I think I messed up a little, because it seems more correct to create a DialogTriggerService interface and change the KeywordSpotterService interface to implement that one, because at the end you don't what is happening in the remote to trigger the dialog. 
I can change that if you think is a better option, or create a PR to change it afterwards, for now it's not used anywhere else.

GiviMAD:
Thank you for the looking at it @florian-h05, I have responded in some comments.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/KSEdgeService.java
dalgwen: > Looking at it now I think I messed up a little, because it seems more correct to create a DialogTriggerService interface and change the KeywordSpotterService interface to implement that one, because at the end you don't what is happening in the remote to trigger the dialog.

Yes, I agree.
I am working on an addon for the Wyoming protocol, from the Home Assistant voice ecosystem, and a remote trigger for wakeword is a part of it that I definitely want to support. The architecture you are reffering to is definitely something I was planning to work on.
Also for example, I think that a switch should also be able to trigger a dialog, so a generic DialogTriggerService for those other cases would be a good step forward to support them 👍

bundles/org.openhab.core.io.websocket/pom.xml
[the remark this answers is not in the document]
GiviMAD: > As the websocket adapter is for audio streams in general, I would say that it should ideally be placed in the audio bundle.
The dependency on the voice manager is very small; maybe you could come up with some small hook, which could be used by the voice bundle to register the dialog processing logic into the websocket adapter?

I totally forgot this comment from @kaikreuzer where he suggested to place it on the audio bundle using a hook to avoid the circular dependency with the voice bundle. I will go with that one.

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/KSEdgeService.java
GiviMAD: I'll add a DialogTriggerService interface then and remove the KSEdgeService one.

bundles/org.openhab.core.io.websocket/pom.xml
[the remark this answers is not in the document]
GiviMAD: moved to audio bundle

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/DTServiceHandle.java
```diff
@@ -0,0 +1,28 @@
...
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.voice;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
+/**
+ * A handle to a {@link DTService}
```
florian-h05, line 18:
```suggestion
 * A handle to a {@link DTService}.
```

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/AudioDialogProviderImpl.java
```diff
@@ -0,0 +1,115 @@
...
+import org.openhab.core.voice.DTTriggeredEvent;
+import org.openhab.core.voice.VoiceManager;
+import org.osgi.service.component.annotations.Activate;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Reference;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Allows the audio bundle to register a dialog that can be triggered programmatically
```
florian-h05, line 35:
```suggestion
 * Allows the audio bundle to register a dialog that can be triggered programmatically.
```

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
```diff
@@ -152,33 +157,40 @@ private void initToneSynthesizer(@Nullable String listeningMelodyText) {
      * 
      * @throws IllegalStateException if keyword spot service is misconfigured
      */
-    public void start() throws IllegalStateException {
-        KSService ksService = dialogContext.ks();
+    public @Nullable DTServiceHandle start() throws IllegalStateException {
```
florian-h05, line 160: Please document the return type.

florian-h05 (approved):
A few minor comments, apart from this LGTM now. Many thanks!

florian-h05 (approved):
LGTM now, thanks!

@kaikreuzer Can you please review?

GiviMAD:
karaf features build is currently broken after moving the code to the audio bundle, because the audio bundle seems to be part of the base feature and the websocket bundle is only available in the runtime feature. Is it ok to move the code to a separate org.openhab.core.io.websocket.audio bundle and add it as part of the runtime feature?

GiviMAD:
> karaf features build is currently broken after moving the code to the audio bundle, because the audio bundle seems to be part of the base feature and the websocket bundle is only available in the runtime feature. Is it ok to move the code to a separate org.openhab.core.io.websocket.audio bundle and add it as part of the runtime feature?

I when ahead and refactored the code into the org.openhab.core.io.websocket.audio bundle (new), let me know if it's ok like this or if there is another option.

GiviMAD:
I was able to test the latest changes today using the UI PR and seems to be working properly.

florian-h05:
@openhab/core-maintainers Can you please review this for 5.0? It’s on the 5.0 feature roadmap.

bundles/org.openhab.core.io.websocket.audio/src/main/java/org/openhab/core/io/websocket/audio/internal/PCMWebSocketAdapter.java
```diff
@@ -0,0 +1,148 @@
...
+        logger.debug("stopping connection check");
+        pingTask.cancel(true);
+        disconnectAll();
+    }
+
+    private void pingHandlers() {
+        ArrayList<PCMWebSocketConnection> handlers = new ArrayList<>(webSocketConnections);
+        for (var handler : handlers) {
+            if (handler != null) {
+                boolean pinned = false;
```
kaikreuzer: Is this meant to be called `pinged`?

bundles/org.openhab.core.io.websocket.audio/src/main/java/org/openhab/core/io/websocket/audio/internal/PCMWebSocketAdapter.java
```diff
@@ -0,0 +1,148 @@
...
+                var remote = handler.getRemote();
+                if (remote != null) {
+                    try {
+                        remote.sendPing(wrap("oh".getBytes(StandardCharsets.UTF_8)));
+                        pinned = true;
+                    } catch (IOException ignored) {
+                    }
+                }
+                if (!pinned) {
+                    logger.warn("ping failed, disconnecting speaker {}", handler.getId());
```
kaikreuzer: This can happen whenever the speaker's websocket connection disappears, right?
So it is not a "bug situation" and hence we should only log it as debug, but not as warn.

bundles/org.openhab.core.io.websocket.audio/src/main/java/org/openhab/core/io/websocket/audio/internal/PCMWebSocketConnection.java
```diff
@@ -0,0 +1,373 @@
...
+
+    @Override
+    public void onWebSocketConnect(@Nullable Session sess) {
+        if (sess == null) {
+            // never
+            return;
+        }
+        this.session = sess;
+        this.remote = sess.getRemote();
+        logger.info("New client connected.");
```
kaikreuzer: If we log this as info, we might want to add more details to the message - e.g. something that let's the user differentiate between different clients that connect. Alternatively, reduce to debug.

bundles/org.openhab.core.io.websocket.audio/src/main/java/org/openhab/core/io/websocket/audio/internal/PCMWebSocketConnection.java
```diff
@@ -0,0 +1,373 @@
...
+                                scheduledDisconnection.cancel(true);
+                            }
+                            // update connection settings
+                            id = clientOptions.id;
+                            registerSpeakerComponents(id, clientOptions);
+                            sendClientCommand(new WebSocketCommand(WebSocketCommand.OutputCommands.INITIALIZED));
+                        }
+                        case ON_SPOT -> onRemoteSpot();
+                    }
+
```
kaikreuzer:
```suggestion
```

bundles/org.openhab.core.io.websocket.audio/src/main/java/org/openhab/core/io/websocket/audio/internal/PCMWebSocketConnection.java
```diff
@@ -0,0 +1,373 @@
...
+                } catch (IOException | IllegalStateException e) {
+                    logger.warn("Disconnecting client: {}", e.getMessage());
```
kaikreuzer: Is this unexpected? If so, better log it in a way that it clearly shows what went wrong.
Otherwise just do it as debug.

bundles/org.openhab.core.io.websocket.audio/src/main/java/org/openhab/core/io/websocket/audio/internal/PCMWebSocketConnection.java
```diff
@@ -0,0 +1,373 @@
...
+                    disconnect();
+                }
+            }
+        } catch (JsonProcessingException e) {
+            logger.warn("Exception parsing JSON message: ", e);
```
kaikreuzer:
```suggestion
            logger.warn("Exception parsing JSON message", e);
```

bundles/org.openhab.core.io.websocket.audio/src/main/java/org/openhab/core/io/websocket/audio/internal/PCMWebSocketConnection.java
```diff
@@ -0,0 +1,373 @@
...
+        }
+    }
+
+    @Override
+    public void onWebSocketError(@Nullable Throwable cause) {
+        logger.warn("WebSocket Error: ", cause);
```
kaikreuzer:
```suggestion
        logger.warn("WebSocket Error", cause);
```

kaikreuzer (changes requested):
Hey @GiviMAD,
I know that there is absolutely no excuse for why it has taken so long - I'll hence not even try to excuse it and merely beg your pardon for it. 😞 

The code looks good to me and I know that @florian-h05 is a big fan of this feature and had wished to see it being merged long ago already.

I merely left a few minor comments, mainly on logging. Besides this, it would be great if you could address the (afaics valid) copilot comments and I'll happily merge this PR!

miloit:
@openhab/core-maintainers might it be possible to get this pr merged already in the next milestone build?

kaikreuzer:
@miloit You have to ask @GiviMAD, not the maintainers. But considering that the review almost took 2 years, I will be the last one to put any time pressure on him. 😄

miloit:
@GiviMAD may I ask you to do the last changes would be great to have it finally merged

miloit:
@openhab/core-maintainers may it be possible to review the commits from @GiviMAD ....

Thanks @GiviMAD for the commits

bundles/org.openhab.core.io.websocket.audio/src/main/java/org/openhab/core/io/websocket/audio/internal/PCMWebSocketConnection.java
```diff
@@ -0,0 +1,373 @@
...
+ * <p>
+ * The websocket supports multiple lines for the {@link PCMWebSocketAudioSink} (to accomplish that, the outgoing data
+ * chucks are prefixed with a 6 byte header to transmit the identity and format specification, check
+ * {@link PCMWebSocketAudioSink.PCMWebSocketOutputStream})
+ *
+ * @author Miguel Álvarez Díez - Initial contribution
+ */
+@WebSocket
+@NonNullByDefault
+@SuppressWarnings("unused")
```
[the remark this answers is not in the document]
GiviMAD: I've ignored this comment as the annotation is used in other places

GiviMAD:
Hello, sorry for the delay, I have updated the PR and fixed the comments.

I'm going to update the UI PR to test it again but I think it is ok.

PR #4034, opened 2024-01-12
Fix wrong class name logged in LocaleServiceImpl

wborn:
The logging will be less confusing when the correct class name is logged.


kaikreuzer (approved):
Thanks!

PR #4052, opened 2024-01-20
[marketplace] Update CommunityUIWidgetAddonHandler.java to give a more helpful error

Skinah:
Installing from marketplace gives an Error that is not very helpful.

Couldn’t find the widget in the add-on entry

The cause was missing ```yaml code fence when a normal code fence was used. This used to work in the past so a number of widgets are done this way and will not install. Suggest a more useful message to help guide widget devs.

### DIFF
--- a/bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityUIWidgetAddonHandler.java
+++ b/bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityUIWidgetAddonHandler.java
@@ -86,7 +86,7 @@ public class CommunityUIWidgetAddonHandler implements MarketplaceAddonHandler {
             } else if (yamlContent != null) {
                 addWidgetAsYAML(addon.getUid(), yamlContent);
             } else {
-                throw new IllegalArgumentException("Couldn't find the widget in the add-on entry");
+                throw new IllegalArgumentException("Couldn't find the widget in the add-on entry, or the starting code fence is not marked as ```yaml");
             }
         } catch (IOException e) {
             logger.error("Widget from marketplace cannot be downloaded: {}", e.getMessage());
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityUIWidgetAddonHandler.java
```diff
@@ -86,7 +86,7 @@ public void install(Addon addon) throws MarketplaceHandlerException {
             } else if (yamlContent != null) {
                 addWidgetAsYAML(addon.getUid(), yamlContent);
             } else {
-                throw new IllegalArgumentException("Couldn't find the widget in the add-on entry");
+                throw new IllegalArgumentException("Couldn't find the widget in the add-on entry, or the starting code fence is not marked as ```yaml");
```
wborn, line 89:
```suggestion
                throw new IllegalArgumentException("Couldn't find the widget in the add-on entry. The starting code fence may not be marked as ```yaml");
```

It's not one or the other. It's a cause for not being able to find the widget.

Skinah: So what is the plan forwards? Reword it or ignore that there are. At least 4 widgets that can not be installed because of the missing yaml tag on the code fence? Wasted a few hours tracking the reason down because the message gave no clues as to the real cause.

wborn: I think it's clear my suggestion is to reword it? It's a good a addition to have a common cause in the exception message. 👍

Skinah:
Sorry @wborn missed your suggestion on my. Mobile phone screen. Accepted your suggestion Thanks.

wborn (changes requested):
Spotless also has an opinion about this: 

```diff
-················throw·new·IllegalArgumentException("Couldn't·find·the·widget·in·the·add-on·entry.·The·starting·code·fence·may·not·be·marked·as·```yaml");
+················throw·new·IllegalArgumentException(
+························"Couldn't·find·the·widget·in·the·add-on·entry.·The·starting·code·fence·may·not·be·marked·as·```yaml");
```

wborn (approved):
Thanks!

PR #4058, opened 2024-01-22
Script profile: Separate toHandlerScript for commands and states

florian-h05:
This allows much more fine-grained control for the script profile. 

E.g. it is now possible to mimic the behaviour of the `system:follow` profile, but apply a script transformation to the forwarded state. I discovered the need for this when trying to make a KNX channel follow an Item state, but at the same time needed to transform the Item state before sending it to KNX.

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/profile/ScriptProfile.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/profile/ScriptProfile.java
@@ -43,8 +43,8 @@ public class ScriptProfile implements StateProfile {
 
     public static final String CONFIG_TO_ITEM_SCRIPT = "toItemScript";
     public static final String CONFIG_TO_HANDLER_SCRIPT = "toHandlerScript";
-    public static final String CONFIG_TO_HANDLER_COMMAND_SCRIPT = "toHandlerCommandScript";
-    public static final String CONFIG_TO_HANDLER_STATE_SCRIPT = "toHandlerStateScript";
+    public static final String CONFIG_COMMAND_FROM_ITEM_SCRIPT = "commandFromItemScript";
+    public static final String CONFIG_STATE_FROM_ITEM_SCRIPT = "stateFromItemScript";
 
     private final Logger logger = LoggerFactory.getLogger(ScriptProfile.class);
 
@@ -56,8 +56,8 @@ public class ScriptProfile implements StateProfile {
     private final List<Class<? extends Command>> handlerAcceptedCommandTypes;
 
     private final String toItemScript;
-    private final String toHandlerCommandScript;
-    private final String toHandlerStateScript;
+    private final String commandFromItemScript;
+    private final String stateFromItemScript;
     private final ProfileTypeUID profileTypeUID;
 
     private final boolean isConfigured;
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.core.automation.module.script/src/main/resources/OH-INF/config/script-profile.xml
```diff
@@ -7,14 +7,27 @@
 	<config-description uri="profile:transform:SCRIPT">
 		<parameter name="toItemScript" type="text">
 			<label>Thing To Item Transformation</label>
-			<description>The Script for transforming state updates and commands from the Thing handler to the item. The script
+			<description>The Script for transforming state updates and commands from the Thing handler to the Item. The script
```
J-N-K, line 10: Why did you change from "item" to "Item"? IMO "item" is correct and "Script/Thing" should be changed to "script/thing" in that sentence.

florian-h05: I thought that openHAB terms like Item or Thing should be written uppercase

florian-h05: How to continue here?

florian-h05: I have written `item` lowercase and have not changed Script/Thing yet.
I would suggest to leave it as is and probably discuss a convention across all maintainers so no one is confused anymore whether Item has to be spelled upper- or lowercase.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/profile/ScriptProfile.java
```diff
@@ -92,18 +105,25 @@ public ProfileTypeUID getProfileTypeUID() {
...
+        }
+    }
+
+    private void internalFromItem(String script, Type type) {
+        String returnValue = executeScript(script, type);
+        if (returnValue != null) {
+            // try to parse the value
+            Command newCommand = TypeParser.parseCommand(handlerAcceptedCommandTypes, returnValue);
+            if (newCommand != null) {
+                callback.handleCommand(newCommand);
```
J-N-K, line 126: If we try to keep it as "open as possible", do we need to apply the same workaround here that we have in https://github.com/openhab/openhab-addons/pull/16263?

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/profile/ScriptProfile.java
```diff
@@ -92,18 +105,25 @@ public ProfileTypeUID getProfileTypeUID() {
 
     @Override
     public void onStateUpdateFromItem(State state) {
+        if (isConfigured && !toHandlerStateScript.isBlank()) {
```
J-N-K, line 108: Is it necessary to check `!toHandlerStateScript.isBlank()` here? `executeScript` checks `.isBlank()` anyway.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/profile/ScriptProfile.java
```diff
@@ -92,18 +105,25 @@ public ProfileTypeUID getProfileTypeUID() {
...
     public void onCommandFromItem(Command command) {
-        if (isConfigured) {
-            String returnValue = executeScript(toHandlerScript, command);
-            if (returnValue != null) {
-                // try to parse the value
-                Command newCommand = TypeParser.parseCommand(handlerAcceptedCommandTypes, returnValue);
-                if (newCommand != null) {
-                    callback.handleCommand(newCommand);
-                }
+        if (isConfigured && !toHandlerCommandScript.isBlank()) {
```
J-N-K, line 115: See above

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/profile/ScriptProfile.java
```diff
@@ -43,6 +43,8 @@ public class ScriptProfile implements StateProfile {
 
     public static final String CONFIG_TO_ITEM_SCRIPT = "toItemScript";
     public static final String CONFIG_TO_HANDLER_SCRIPT = "toHandlerScript";
+    public static final String CONFIG_TO_HANDLER_COMMAND_SCRIPT = "toHandlerCommandScript";
```
J-N-K, line 46: If we change the naming anyway: Wouldn't `stateFromItemScript` and `commandFromItemScript` be more clear? I'm not sure here, it just came to my mind.

J-N-K (changes requested):

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/profile/ScriptProfile.java
florian-h05: Agreed! Having `toItem` and `fromItem` IMO is better to understand for the user as `toHandler` and `toItem`, because I am not sure whether most users actually know what the/a (Thing) handler is.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/profile/ScriptProfile.java
florian-h05: I haven't noticed that, it is not neccassary to double check.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/profile/ScriptProfile.java
florian-h05: Probably yes, but I am not sure whether we want to "distribute" that work-around across the code or if it would be better to wait for a solution of the underlying issue/problem.

florian-h05: While having a look at `ProfileCallbackImpl`, the problem seems to be that `CommunicationManager::toAcceptedCommand` does not accept `OpenClosedType`, which IMO is correct as you normally do not command contact Items.

The question is, how to solve that.
I would think the cleanest solution is to add a method like `CommunicationManager::toAcceptedState` and to introduce a method called `handleState` or `handleUpdate` to the `ProfileCallback`. 
Then all profiles could call `handleUpdate` instead of `handleCommand` for state updates from the Item to the Thing.

florian-h05:
@J-N-K I don't want to be pushy, but it would be great to get this PR merged soon.

bundles/org.openhab.core.automation.module.script/src/main/resources/OH-INF/config/script-profile.xml
```diff
@@ -11,12 +11,25 @@
...
 				discard the commands and not pass them through.</description>
 			<limitToOptions>false</limitToOptions>
 		</parameter>
+		<parameter name="stateFromItemScript" type="text">
+			<label>Item State To Thing Transformation</label>
+			<description>The Script for transforming states from the item to the Thing handler. The script may return null to
+				discard the states and not pass them through.</description>
+			<limitToOptions>false</limitToOptions>
+		</parameter>
+		<parameter name="toHandlerScript" type="text">
```
J-N-K, line 26: Maybe show this one as "advanced"? That hides it for new users, while keeping it for this that already have it configured.

Side note: It would be great if Main UI would automatically show "advanced" parameters if something is configured there. This would also be helpful e.g. for HTTP binding things.

J-N-K (commented):
Otherwise LGTM

J-N-K (approved):
Thanks!

bundles/org.openhab.core.automation.module.script/src/main/resources/OH-INF/config/script-profile.xml
florian-h05: > Side note: It would be great if Main UI would automatically show "advanced" parameters if something is configured there. This would also be helpful e.g. for HTTP binding things.

There we go: https://github.com/openhab/openhab-webui/pull/2313

florian-h05:
@openhab/jruby-maintainers FYI in case you need to adjust documentation.

florian-h05:
Yeah, good idea.

PR #4061, opened 2024-01-23
AddonFinderProcess fixes

andrewfg:
See https://github.com/openhab/openhab-core/pull/3904#issuecomment-1904702212

### DIFF
--- a/bundles/org.openhab.core.config.discovery.addon.process/src/main/java/org/openhab/core/config/discovery/addon/process/ProcessAddonFinder.java
+++ b/bundles/org.openhab.core.config.discovery.addon.process/src/main/java/org/openhab/core/config/discovery/addon/process/ProcessAddonFinder.java
@@ -72,7 +72,10 @@ public class ProcessAddonFinder extends BaseAddonFinder {
                 String[] args = info.arguments().orElse(null);
                 if (args != null) {
                     for (int i = args.length - 1; i >= 0; i--) {
-                        cmd = cmd.substring(0, cmd.lastIndexOf(args[i]));
+                        int index = cmd.lastIndexOf(args[i]);
+                        if (index >= 0) {
+                            cmd = cmd.substring(0, index);
+                        }
                     }
                 }
                 cmd = cmd.stripTrailing();
--- a/bundles/org.openhab.core.config.discovery.addon.process/src/main/java/org/openhab/core/config/discovery/addon/process/ProcessAddonFinder.java
+++ b/bundles/org.openhab.core.config.discovery.addon.process/src/main/java/org/openhab/core/config/discovery/addon/process/ProcessAddonFinder.java
@@ -63,9 +63,6 @@ public class ProcessAddonFinder extends BaseAddonFinder {
         /**
          * Initializes the command and commandLine fields.
          * If the command field is not present, it parses the first token in the command line.
-         * <p>
-         * Note: the parser would produce a wrong result if argument separators are NOT just a single space.
-         * e.g. "C:\path\to file\program.exe arg0 ~t arg1 ~s arg2 arg3" -- ~t=tab, ~s=space, etc.
          */
         protected ProcessInfo(ProcessHandle.Info info) {
             commandLine = info.commandLine().orElse(null);
@@ -73,10 +70,9 @@ public class ProcessAddonFinder extends BaseAddonFinder {
             if ((cmd == null || cmd.isEmpty()) && commandLine != null) {
                 cmd = Objects.requireNonNull(commandLine);
                 String[] args = info.arguments().orElse(null);
-                if (args != null && args.length > 0) {
-                    int iArgs = cmd.lastIndexOf(String.join(" ", args));
-                    if (iArgs > 0) {
-                        cmd = cmd.substring(0, iArgs);
+                if (args != null) {
+                    for (int i = args.length - 1; i >= 0; i--) {
+                        cmd = cmd.substring(0, cmd.lastIndexOf(args[i]));
                     }
                 }
                 cmd = cmd.stripTrailing();
--- a/bundles/org.openhab.core.config.discovery.addon.process/src/main/java/org/openhab/core/config/discovery/addon/process/ProcessAddonFinder.java
+++ b/bundles/org.openhab.core.config.discovery.addon.process/src/main/java/org/openhab/core/config/discovery/addon/process/ProcessAddonFinder.java
@@ -31,6 +31,8 @@ import org.osgi.service.component.annotations.Component;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.sun.jna.Platform;
+
 /**
  * This is a {@link ProcessAddonFinder} for finding suggested add-ons by checking processes running
  * on the openHAB server.
@@ -49,6 +51,8 @@ public class ProcessAddonFinder extends BaseAddonFinder {
     private static final String COMMAND_LINE = "commandLine";
     private static final Set<String> SUPPORTED_PROPERTIES = Set.of(COMMAND, COMMAND_LINE);
 
+    private static final boolean NOT_WINDOWS = !Platform.isWindows();
+
     private final Logger logger = LoggerFactory.getLogger(ProcessAddonFinder.class);
 
     /**
@@ -61,9 +65,8 @@ public class ProcessAddonFinder extends BaseAddonFinder {
 
         protected ProcessInfo(ProcessHandle.Info info) {
             String cmd = info.command().orElse(null);
-            if (cmd == null) {
-                String[] args = info.arguments().orElse(new String[] {});
-                // TODO (TBC) I think that the following use of args[0] may be wrong ??
+            if ((cmd == null || cmd.isEmpty()) && NOT_WINDOWS) {
+                String[] args = info.commandLine().orElse("").split("\\s+");
                 cmd = args.length > 0 ? args[0] : null;
             }
             command = cmd;
@@ -92,9 +95,13 @@ public class ProcessAddonFinder extends BaseAddonFinder {
                 Map<String, Pattern> matchProperties = method.getMatchProperties().stream()
                         .collect(Collectors.toMap(AddonMatchProperty::getName, AddonMatchProperty::getPattern));
 
+                if (matchProperties.isEmpty()) {
+                    logger.warn("Add-on '{}' addon.xml file contains no 'match-property'", candidate.getUID());
+                    break;
+                }
+
                 Set<String> propertyNames = new HashSet<>(matchProperties.keySet());
                 propertyNames.removeAll(SUPPORTED_PROPERTIES);
-
                 if (!propertyNames.isEmpty()) {
                     logger.warn("Add-on '{}' addon.xml file contains unsupported 'match-property' [{}]",
                             candidate.getUID(), String.join(",", propertyNames));
--- a/bundles/org.openhab.core.config.discovery.addon.process/src/main/java/org/openhab/core/config/discovery/addon/process/ProcessAddonFinder.java
+++ b/bundles/org.openhab.core.config.discovery.addon.process/src/main/java/org/openhab/core/config/discovery/addon/process/ProcessAddonFinder.java
@@ -15,13 +15,13 @@ package org.openhab.core.config.discovery.addon.process;
 import static org.openhab.core.config.discovery.addon.AddonFinderConstants.ADDON_SUGGESTION_FINDER;
 
 import java.util.HashSet;
-import java.util.List;
-import java.util.Optional;
+import java.util.Map;
 import java.util.Set;
-import java.util.function.Predicate;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.addon.AddonDiscoveryMethod;
 import org.openhab.core.addon.AddonInfo;
 import org.openhab.core.addon.AddonMatchProperty;
@@ -46,35 +46,40 @@ public class ProcessAddonFinder extends BaseAddonFinder {
     public static final String SERVICE_NAME = SERVICE_TYPE + ADDON_SUGGESTION_FINDER;
 
     private static final String COMMAND = "command";
+    private static final String COMMAND_LINE = "commandLine";
+    private static final Set<String> SUPPORTED_PROPERTIES = Set.of(COMMAND, COMMAND_LINE);
 
     private final Logger logger = LoggerFactory.getLogger(ProcessAddonFinder.class);
 
-    // get list of running processes visible to openHAB,
-    // also tries to mitigate differences on different operating systems
-    String getProcessCommandProcess(ProcessHandle h) {
-        Optional<String> command = h.info().command();
-        if (command.isPresent()) {
-            return command.get();
+    /**
+     * Private class to extract match property parameters from a {@link ProcessHandle.Info} object.
+     * Tries to mitigate differences on different operating systems.
+     */
+    protected static class ProcessInfo {
+        protected final @Nullable String command;
+        protected final @Nullable String commandLine;
+
+        protected ProcessInfo(ProcessHandle.Info info) {
+            String cmd = info.command().orElse(null);
+            if (cmd == null) {
+                String[] args = info.arguments().orElse(new String[] {});
+                // TODO (TBC) I think that the following use of args[0] may be wrong ??
+                cmd = args.length > 0 ? args[0] : null;
+            }
+            command = cmd;
+            commandLine = info.commandLine().orElse(null);
         }
-        Optional<String[]> args = h.info().arguments();
-        if (args.isEmpty()) {
-            return "";
-        }
-        String[] argsArray = args.get();
-        if (argsArray.length < 1) {
-            return "";
-        }
-        return argsArray[0];
     }
 
     @Override
     public Set<AddonInfo> getSuggestedAddons() {
         logger.trace("ProcessAddonFinder::getSuggestedAddons");
         Set<AddonInfo> result = new HashSet<>();
-        Set<String> processList;
+        Set<ProcessInfo> processInfos;
+
         try {
-            processList = ProcessHandle.allProcesses().map(this::getProcessCommandProcess)
-                    .filter(Predicate.not(String::isEmpty)).collect(Collectors.toUnmodifiableSet());
+            processInfos = ProcessHandle.allProcesses().map(process -> new ProcessInfo(process.info()))
+                    .collect(Collectors.toUnmodifiableSet());
         } catch (SecurityException | UnsupportedOperationException unused) {
             logger.info("Cannot obtain process list, suggesting add-ons based on running processes is not possible");
             return result;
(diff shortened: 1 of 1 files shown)


andrewfg:
@holgerfriedrich / @mherwege for info..

andrewfg:
@holgerfriedrich as mentioned elsewhere I have some doubts about whether your usage of `args[0]` to get the executable name is actually correct. I think that the executable name would be in "`args[-1]`" (if such a thing would exist)..

Therefore I think the first code block quoted below should probably be replaced by the second code block below. However I did not make that change (yet) because you may have a good explanation about why `args[0]` does actually make sense. => Please advise.

**Doubtful Code**

 ```
        protected ProcessInfo(ProcessHandle.Info info) {
            String cmd = info.command().orElse(null);
            if (cmd == null) {
                String[] args = info.arguments().orElse(new String[] {});
                // TODO (TBC) I think that the following use of args[0] may be wrong ??
                cmd = args.length > 0 ? args[0] : null;
            }
            command = cmd;
            commandLine = info.commandLine().orElse(null);
        }
```

**Possible Replacement**

```
        protected ProcessInfo(ProcessHandle.Info info) {
            String cmd = info.command().orElse(null);
            if (cmd == null && !Platform.isWindows()) {
                String[] args = info.commandLine().orElse("").split("\\s+");
                cmd = args.length > 0 ? args[0] : null;
            }
            command = cmd;
            commandLine = info.commandLine().orElse(null);
        }
```

**Comment by holgerfriedrich** [comment]
@andrewfg I agree, the current implementation only works for Windows. Don't know what I tested, but obviously ```args[0]``` contains the first parameter and not the command :-/

I checked both of my test setups:
- Windows only gives ```command``` to me. ```arguments``` and ```commandLine``` are always empty. ```command``` is empty for a few tasks as well. 
- Linux: many entries with all 3 elements empty. ```command``` only populated for java task. ```arguments``` do not include the command. ```commandLine``` is available and contains the command as well. Spaces are not escaped, so splitting the commandLine may not work (edge case, spaces in directory names uncommon in Linux)

Regarding scripts: On Windows, we just get the Python executable, not the script which is called...  It seems we cannot get the arguments or commandLine using this API.

**Comment by andrewfg** [comment]
@holgerfriedrich -- I made some changes today. The `match-property` value support is as now follows. => Can you please confirm that this is Ok?

**Windows**
- `commandLine` property value not available (so no support for (e.g. python) scripts
- `command` supported directly via `ProcessHandle.info().command()`
 
**Other OS**
- `commandLine` supported directly via `ProcessHandle.info().commandLine()`
- `command` supported via parsing `ProcessHandle.info().commandLine().split()[0]` -- this is a change from your previously wrong `ProcessHandle.info().arguments()[0]`

**Comment by andrewfg** [comment]
@holgerfriedrich apropos Windows there is a thread on StackOverflow (below) .. it contains a post that suggests a hack to get the command line .. but I am not sure if we should go that far. => WDYT?

https://stackoverflow.com/questions/46767418/how-to-get-commandline-arguments-of-process-in-java-9

**Comment by holgerfriedrich** [comment]
> @holgerfriedrich apropos Windows there is a thread on StackOverflow (below) .. it contains a post that suggests a hack to get the command line .. but I am not sure if we should go that far. => WDYT?

I think this is too much of a hack. Too much platform specific complexity, unless we have a clear need for that.
With the additional comment about future removal: My recommendation is not to implement this and accept that we cannot detect scripts on Windows.

**Comment by andrewfg** [comment]
@holgerfriedrich just so you know: the change that I made from args[0] to commandLine[0] means that now on Linux systems those bindings that already have discovery info in their addons.xml will now be properly suggested (they were not suggested previously) e.g. MQTT ..

**Comment by holgerfriedrich** [comment]
Not sure if the implementation of ```ProcessHandle.allProcesses()``` is JDK specific, so when searching for the command on other OS, maybe check if ```command``` is populated and only if not start splitting ```commandLine```.

Linux paths can contain spaces as well, I just tried out and got a ```commandLine``` as follows: ```/home/openhabian/test dir/bash```. Space is in the directory name, and not escaped in the return value. But I think it is the same as above for Windows: edge case, no need to implement a solution for that. Directory names with spaces are very uncommon.

**Comment by holgerfriedrich** [comment]
> @holgerfriedrich just so you know: the change that I made from args[0] to commandLine[0] means that now on Linux systems those bindings that already have discovery info in their addons.xml will now be properly suggested (they were not suggested previously) e.g. MQTT ..

:+1: Maybe we can ask to get this cherry picked to the 4.1 branch.

I have not yet looked at the code, maybe I can do later today.

**Comment by andrewfg** [comment]
> check if command is populated and only if not start splitting

Yes. That is the case.

**Comment by andrewfg** [comment]
> I have not yet looked at the code, maybe I can do later today.

See my latest commit. Please read the javadoc..

File: bundles/org.openhab.core.config.discovery.addon.process/src/main/java/org/openhab/core/config/discovery/addon/process/ProcessAddonFinder.java [hunk 1]
```diff
@@ -84,28 +108,26 @@ public Set<AddonInfo> getSuggestedAddons() {
...
 
-                if (commands.isEmpty()) {
-                    logger.warn("Add-on '{}' addon.xml file does not specify match property \"{}\"", candidate.getUID(),
-                            COMMAND);
+                Set<String> propertyNames = new HashSet<>(matchProperties.keySet());
+                propertyNames.removeAll(SUPPORTED_PROPERTIES);
+                if (!propertyNames.isEmpty()) {
+                    logger.warn("Add-on '{}' addon.xml file contains unsupported 'match-property' [{}]",
+                            candidate.getUID(), String.join(",", propertyNames));
                     break;
```
holgerfriedrich, line 124: Do you think "break" should be removed here? It may cause backward compatibility issue if we want to add a new match property (for newer add-ons running on an old core version).

At least one match property is there, so we can safely run the check below.

mherwege: Agree, warning and ignoring should be sufficient.

mherwege:
Is the path in the command relevant? If not, you could split on file separator character first before separating on whitespace. That would eliminate the paths with whitespace problem.
The file separator character can easily be retrieved:
```
String fileSeparator = FileSystems.getDefault().getSeparator();
```

bundles/org.openhab.core.config.discovery.addon.process/src/main/java/org/openhab/core/config/discovery/addon/process/ProcessAddonFinder.java
andrewfg: > one match property is there, so we can safely run the check below.

If the only match property were an as yet unsupported one, the current code would result in a false positive. But I can fix that.

PS I also have an idea to optimise the command line parsing.

mherwege: At the moment, that is not an issue, as the finders only work on distribution add-ons. And there it is no issue as versions will be compatible. The marketplace add-ons will not be considered in the finders. So this whole discussion is in the end not relevant anyway.

andrewfg:
> the path in the command relevant?

Who knows? Depending on the addon, it may indeed be so. So I will try to keep it.

Also splitting on slash characters would IMHO make the parsing yet more complex.

mherwege:
> Also splitting on slash characters would IMHO make the parsing yet more complex.

I think you should be able to do it with one regex (not tested):

```
String fileSeparator = FileSystems.getDefault().getSeparator();
String regex = "^(([^\\" + fileSeparator + "]*\\" + fileSeparator + ")+)?[^\\" + fileSeparator + "\\s+]*"
Pattern cmdPattern = Pattern.compile(regex);
Matcher matcher = cmdPattern.matcher(commandLine)
String command = matcher.find() ? matcher.group() : null;
```

Of course the pattern can be compiled in advance.

andrewfg:
> I think you should be able to do it with one regex (not tested)

@mherwege I did test it .. and unfortunately it doesn't work.

Anyway, I just pushed another commit that I think resolves the issue in faster resp. better way.

bundles/org.openhab.core.config.discovery.addon.process/src/main/java/org/openhab/core/config/discovery/addon/process/ProcessAddonFinder.java
```diff
@@ -67,26 +64,27 @@ protected static class ProcessInfo {
...
-                    command = iArg0 > iWhite ? Objects.requireNonNull(commandLine).substring(0, iArg0).trim() : null;
-                } else {
-                    command = Objects.requireNonNull(commandLine).trim();
+                    if (args.length == 1) {
+                        cmd = cmd + " ";
+                    }
+                    int iArg0 = cmd.lastIndexOf(" " + args[0] + " ");
+                    if (iArg0 > 0) {
+                        cmd = cmd.substring(0, iArg0);
+                    }
```
mherwege, line 83: Did you try:
```
int iArg0 = cmd.lastIndexOf(String.join(" ", argsArray));
if (iArg0 > 0) {
    cmd = cmd.substring(0, iArg0);
}
```

andrewfg: Yes I certainly thought about that already. I was worried that the command line might have extra space characters between the args .. but let me make that change and we can all test it.

mherwege: > Since Windows and Linux are probably the vast majority of user platforms, perhaps the best approach is to use your "String.join(" ", argsArray)" approach, and fix it later if a user on another OS identifies an actual problem. => WDYT?

Yes, agreed.

I was thinking about a solution to overcome multiple whitespace characters between arguments, but that again is only valid if arguments are separated by whitespace characters. If one would find an OS that uses an other character it may not work. But that will probably create many more issues than this.

```
for (int i = argsArray.length - 1; i >= 0; i--) {
    cmd = cmd.substring(0, cmd.lastIndexOf(argsArray[i]);
}
```

andrewfg:
```
for (int i = argsArray.length - 1; i >= 0; i--) {
    cmd = cmd.substring(0, cmd.lastIndexOf(argsArray[i]);
}
```

^ I think this might also not work in the case that argsArray[m] would be a sub-string of argsArray[n] .. ??

mherwege: Potentially a bit slower indeed, but I doubt the impact will be important.

mherwege (approved):
LGTM

andrewfg:
@mherwege I think for safety sake I need to check that lastIndexOf(args[i]) does not return -1 .. it might be the case that args has a String that is somehow not in the commandLine..

mherwege:
> @mherwege I think for safety sake I need to check that lastIndexOf(args[i]) does not return -1 .. it might be the case that args has a String that is somehow not in the commandLine..

Would be unexpected, but indeed, defensive programming.

holgerfriedrich (approved):
LGTM

bundles/org.openhab.core.config.discovery.addon.process/src/main/java/org/openhab/core/config/discovery/addon/process/ProcessAddonFinder.java
```diff
@@ -46,35 +47,52 @@ public class ProcessAddonFinder extends BaseAddonFinder {
...
+                    for (int i = args.length - 1; i >= 0; i--) {
+                        int index = cmd.lastIndexOf(args[i]);
+                        if (index >= 0) {
+                            cmd = cmd.substring(0, index);
+                        }
+                    }
+                }
+                cmd = cmd.stripTrailing();
+            }
+            command = cmd;
```
holgerfriedrich, line 83:
```suggestion
    protected static record ProcessInfo(@Nullable String command, @Nullable String commandLine) {

        /**
         * Initializes the command and commandLine fields.
         * If the command field is not present, it parses the first token in the command line.
         */
        protected static ProcessInfo build(ProcessHandle.Info info) {
            @Nullable
            String commandLine = info.commandLine().orElse(null);
            String cmd = info.command().orElse(null);
            if ((cmd == null || cmd.isEmpty()) && commandLine != null) {
                cmd = Objects.requireNonNull(commandLine);
                String[] args = info.arguments().orElse(null);
                if (args != null) {
                    for (int i = args.length - 1; i >= 0; i--) {
                        int index = cmd.lastIndexOf(args[i]);
                        if (index >= 0) {
                            cmd = cmd.substring(0, index);
                        }
                    }
                }
                cmd = cmd.stripTrailing();
            }
            return new ProcessInfo(cmd, commandLine);
```

bundles/org.openhab.core.config.discovery.addon.process/src/main/java/org/openhab/core/config/discovery/addon/process/ProcessAddonFinder.java
```diff
@@ -46,35 +47,52 @@ public class ProcessAddonFinder extends BaseAddonFinder {
...
     public Set<AddonInfo> getSuggestedAddons() {
         logger.trace("ProcessAddonFinder::getSuggestedAddons");
         Set<AddonInfo> result = new HashSet<>();
-        Set<String> processList;
+        Set<ProcessInfo> processInfos;
+
         try {
-            processList = ProcessHandle.allProcesses().map(this::getProcessCommandProcess)
-                    .filter(Predicate.not(String::isEmpty)).collect(Collectors.toUnmodifiableSet());
+            processInfos = ProcessHandle.allProcesses().map(process -> new ProcessInfo(process.info()))
```
holgerfriedrich, line 94:
```suggestion
            processInfos = ProcessHandle.allProcesses().map(process -> ProcessInfo.build(process.info()))
```

holgerfriedrich (changes requested):
@andrewfg I tried parts of the code with Linux and Windows.
It seems that Set is no longer removing the duplicates.
Could you pls double check?

For me, Windows returned dozens of (null, null), and also other duplicate objects, in spite of the conversion ```.collect(Collectors.toUnmodifiableSet());```.

May be caused by our helper lass which does not properly implement comparison.
Using a record is one way to fix this.

bundles/org.openhab.core.config.discovery.addon.process/src/main/java/org/openhab/core/config/discovery/addon/process/ProcessAddonFinder.java
andrewfg: The set contains an entry for all processes. Even those with nulls. So we just need to filter the stream before creating the set.

bundles/org.openhab.core.config.discovery.addon.process/src/main/java/org/openhab/core/config/discovery/addon/process/ProcessAddonFinder.java
```diff
@@ -92,6 +92,7 @@ public Set<AddonInfo> getSuggestedAddons() {
 
         try {
             processInfos = ProcessHandle.allProcesses().map(process -> new ProcessInfo(process.info()))
+                    .filter(info -> (info.command != null) || (info.commandLine != null))
```
holgerfriedrich, line 95: This looks much simpler, but still leaves the problem of duplicates.
My Windows system gave me 75 entries with the implementation based on the record, and about 200 with this implementation. A lot of duplicate entries for firefox etc.
A ```Set``` will contain unique objects - if the compare/equals functions are implemented properly. This worked before on the set of strings. It seems not to work properly unless I implement ```equals``` and ```hashcode```  for ```ProcessInfo``` like this:
```
        @Override
        public boolean equals(@Nullable Object other) {
            if (other instanceof ProcessInfo pi) {
                return (Objects.equals(command, pi.command) && Objects.equals(commandLine, pi.commandLine));
            }
            return false;
        }

        @Override
        public int hashCode() {
            return Objects.hashCode(command) + Objects.hashCode(commandLine);
        }
```

andrewfg: @holgerfriedrich what you call 'duplicates' are not really duplicates. Each entry represents a distinct process. It may be that a single executable has started many processes (most likely each process having a different command line; which on Windows is nulled out). I can imagine this case as you say with a browser having many contexts, or a service runner starting many services, or a script runner running many scripts. Obviously implementing an 'equals' method would allow the set to be smaller, but I am not sure that it is really proper to consider different the ProcessInfo for such processes to be 'equal'. => @mherwege WDYT?

holgerfriedrich: From my perspective, any duplicate in the list - where ever it originates from - does not add any value to the matcher. Duplicates in this case mean that both ```command``` and ```commandLine``` of two entries are the same.
Why should the matcher do a regex matching for the same input twice? This is why I used ```Set``` in the original implementation, which implicitly removed duplicate entries.

andrewfg: Yes. OK. I will implement 'equals' when I get back home on Monday.

mherwege: The solution with records proposed by @holgerfriedlich already has the equals method defined by default if I am not mistaken. It is the same number of lines, so you could use that to avoid having to implement equals.

andrewfg: I will use @holgerfriedrich 's record (with its correct implicit equals) to replace the class (with its implicit equals needing to be overriden) combined with my code to filter out all null records

bundles/org.openhab.core.config.discovery.addon.process/src/main/java/org/openhab/core/config/discovery/addon/process/ProcessAddonFinder.java
```diff
@@ -46,35 +47,51 @@ public class ProcessAddonFinder extends BaseAddonFinder {
...
+
+        /**
+         * Initializes the command and commandLine fields.
+         * If the command field is not present, it parses the first token in the command line.
+         */
+        protected static ProcessInfo from(ProcessHandle.Info info) {
+            String commandLine = info.commandLine().orElse(null);
+            String cmd = info.command().orElse(null);
+            if ((cmd == null || cmd.isEmpty()) && commandLine != null) {
+                cmd = Objects.requireNonNull(commandLine);
```
mherwege, line 69: I don't think the `Objects.requireNonNull` is needed. `commandLine` cannot be null here.

bundles/org.openhab.core.config.discovery.addon.process/src/main/java/org/openhab/core/config/discovery/addon/process/ProcessAddonFinder.java
```diff
@@ -84,28 +101,30 @@ public Set<AddonInfo> getSuggestedAddons() {
...
-                    logger.warn("Add-on '{}' addon.xml file contains unsupported 'match-property'", candidate.getUID());
-                }
+                Map<String, Pattern> matchProperties = method.getMatchProperties().stream()
+                        .collect(Collectors.toMap(AddonMatchProperty::getName, AddonMatchProperty::getPattern));
 
-                if (commands.isEmpty()) {
-                    logger.warn("Add-on '{}' addon.xml file does not specify match property \"{}\"", candidate.getUID(),
-                            COMMAND);
+                if (matchProperties.isEmpty()) {
+                    logger.warn("Add-on '{}' addon.xml file contains no 'match-property'", candidate.getUID());
```
mherwege, line 108: Maybe remove explicit reference to addon.xml (formulate something like Add-on info ... contains no 'match-property').
The reason is we may have other sources of add-on info in the future when we would implement this for the marketplace or 3rd-party providers. These may not be relying on the addon.xml file for this info.

bundles/org.openhab.core.config.discovery.addon.process/src/main/java/org/openhab/core/config/discovery/addon/process/ProcessAddonFinder.java
```diff
@@ -84,28 +101,30 @@ public Set<AddonInfo> getSuggestedAddons() {
...
-                // now check if a process matches the pattern defined in addon.xml
-                logger.trace("Checking candidate: {}", candidate.getUID());
+                Set<String> propertyNames = new HashSet<>(matchProperties.keySet());
+                boolean noSupportedProperty = !propertyNames.removeAll(SUPPORTED_PROPERTIES);
 
-                for (AddonMatchProperty command : commands) {
-                    logger.trace("Candidate {}, pattern \"{}\"", candidate.getUID(), command.getRegex());
-                    boolean match = processList.stream().anyMatch(c -> command.getPattern().matcher(c).matches());
+                if (!propertyNames.isEmpty()) {
+                    logger.warn("Add-on '{}' addon.xml file contains unsupported 'match-property' [{}]",
```
mherwege, line 116: See above

mherwege (approved):
LGTM

andrewfg:
@mherwege / @holgerfriedrich please hold the merge; I was just checking again on my operative system and it seems to not be working at the moment..

holgerfriedrich (approved):
LGTM.
I just copied the record into my plugin for a quick test and tried it on Linux and Windows, works fine.

wborn (approved):
Thanks for the fixes!

PR #4063, opened 2024-01-24
[discovery] Add inbox console command completion

jlaur:
This adds command completion for the inbox console commands.


kaikreuzer (approved):
Cool, thanks!

PR #4067, opened 2024-01-26
Use the ChannelUID to retrieve the Channel from a Thing

joerg1985:
# Description

There is no need to recreate the ChannelUID to read the Channel from a Thing when it is already there.
This will avoid the instance creation and validation of the UID segments of the channel.


clinique (approved):

J-N-K (approved):
Nice improvement thanks!

PR #4069, opened 2024-01-27
[UoM] Add dimension for emission intensity

jlaur:
This adds a dimension that can be used to express emission rate of a given pollutant relative to the intensity of a specific activity, for example CO2 emissions in g/kWh (mass/energy). See https://en.wikipedia.org/wiki/Emission_intensity.

This is a preparation to support UoM for new datasets in the Energi Data Service binding:
image


J-N-K:
I'm not against this, but what is the use-case for this in the smart-home context? Is there any service that provides live emission data for the current power generation?

J-N-K (approved):
Thanks!

jlaur:
@openhab/core-maintainers - can you add the "4.2" milestone?

Issue #4072, opened 2024-01-28
Allow to run rules using scheduler to avoid "Failed to execute rule ‘{}' with status '{}'" error path


OlegAndreych:
I propose to use `org.openhab.core.automation.internal.TriggerHandlerCallbackImpl#getScheduler` when calling scripts to resolve an issue with simultaneous executions, as it is done for "normal" rules execution path.

That's needed for a code reuse using UI scripts.
Many rules are triggered with events with no known intervals and/or timing, so there's no sensible way to spread executions of these rules, so that they don't call scripts simultaneously.

An example of an issue can be found there: https://community.openhab.org/t/simultaneous-calling-of-script-problem/140087.

rkoshak:
I commented on the PR because I didn't see this issue at first. Here is a better place to have the discussion.

This *might* be a breaking change. Right now when one rule calls another rule, the call blocks until the called rule returns. 

For example:

```
console.info('Calling rule');
rules.runRule('longrunning', {}, true);
console.info('Rule returned');
```

with longrunning 

```
console.info('Long running rule called');
Java.type('java.lang.Thread').sleep(5000);
console.info('Sleep done');
```

produces the following output:

```
2024-01-29 11:06:35.142 [INFO ] [nhab.automation.script.ui.scratchpad] - Calling rule
2024-01-29 11:06:35.978 [INFO ] [hab.automation.script.ui.longrunning] - Long running rule called
2024-01-29 11:06:40.997 [INFO ] [hab.automation.script.ui.longrunning] - Sleep done
2024-01-29 11:06:40.997 [INFO ] [nhab.automation.script.ui.scratchpad] - Rule returned
```

If I understand this proposal that behavior would change and the call to `runRule` would immediately return. This could break some end user's rules if they depend on this blocking behavior.

OlegAndreych:
Hello)

It’s not a change, but an addition. 

`runNow` methods of the `org.openhab.core.automation.RuleManager` are still in place and work the same way as before.

It’s up to automation scripting addons maintainers/designers to decide how to expose new calling methods (instead or besides the existing) and whether to expose them at all.

My aim is to expose some API for running rules to the Groovy scripting without the need to resort to getting services from the OSGi and without errors on parallel execution.

I thought this PR would be a good first step.

OlegAndreych:
Talking about code snippets.


As I’ve said earlier, it’s up to maintainers/designers of an addon to decide how to expose new API.


[...]

PR #4074, opened 2024-01-28
Allow to run rules using scheduler to avoid "Failed to execute rule ‘{}' with status '{}'" error path

OlegAndreych:
I propose to use org.openhab.core.automation.internal.TriggerHandlerCallbackImpl#getScheduler when calling scripts to resolve an issue with simultaneous executions, as it is done for "normal" rules execution path.

That's needed for a code reuse using UI scripts.
Many rules are triggered with events with no known intervals and/or timing, so there's no sensible way to spread executions of these rules, so that they don't call scripts simultaneously.

An example of an issue can be found there: https://community.openhab.org/t/simultaneous-calling-of-script-problem/140087.

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/RuleManager.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/RuleManager.java
@@ -14,6 +14,7 @@ package org.openhab.core.automation;
 
 import java.time.ZonedDateTime;
 import java.util.Map;
+import java.util.concurrent.Future;
 import java.util.stream.Stream;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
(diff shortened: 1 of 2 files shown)


rkoshak:
This *might* amount to a breaking change, though it breaks in a way that may not impact anyone.

Right now when one rule calls another, the calling rule blocks until the called rule returns. If I understand this PR correctly, that won't be the case any longer. The call to run the other rule will immediately return instead of blocking, which could break end users's rules if they are depending on this blocking behavior.

OlegAndreych:
Note: we've discussed the comment from above in the issue https://github.com/openhab/openhab-core/issues/4072.

florian-h05:
@openhab/core-maintainers I think this has been done in #5069 so this PR can be closed.

Issue #3235, opened 2022-12-17
[RFC] Change loading mechanism for *.things and *.items files


spacemanspiff2007:
Currently when a file is edited all items from the file get removed and then the items from the file get re-added again.
This results in a log of items/things loosing the current state and runtime data.

A better solution would be in case of a file change to build a diff between the already existing items/things from the file.
Objects that were removed get deleted, objects that were change get changed and objects that were added can be added normally.

jimtng:
AFAIK, this is no longer the case in 4.x? i.e. it now leaves the unchanged items alone and their states remain unchanged?

spacemanspiff2007:
I think there was some work on the items loader so at least for *.items files this should work now. I'm unsure about the  *.things files - I think it's still an issue there.

jimtng:
> I'm unsure about the *.things files - I think it's still an issue there.

Can you try and post steps to reproduce the issue? I just did some quick tests, it seems that changes are picked up.

I posted a PR to fix detecting changes in .items channel configuration (e.g. for profiles), hopefully that'll get merged in the next few days.

PR #4075, opened 2024-01-31
Fix reloading channel config changes in .items file

jimtng:
Apply channel config changes in .items file

Changes in channel config weren't applied because ItemChannelLink.equals() include the link configurations in the comparison. This caused the new link not being found in the set lookup, which leads to erroneously calling notifyListenersAboutAddedElement, when it should've called notifyListenersAboutUpdatedElement instead.

The problem that this PR fixes:

Changing channel config in an .items file, e.g. only the `function` is being changed from:
```
String Test { channel="foo" [ profile="transform:REGEX", function="moo" ] }
```
to:
```
String Test { channel="foo" [ profile="transform:REGEX", function="cow" ] }
```

This wouldn't take effect unless a more severe measure was taken, e.g. commenting out the item and re-adding it, or in many cases, users would restart openHAB to reload such changes.

Reported here: 
https://community.openhab.org/t/problems-at-reducing-item-changes-with-regex/153400/6

Related:
https://github.com/openhab/openhab-core/issues/3235

This problem may have been started / caused by #1794

Note: I've set the initial capacity for the hashmap to 2, because most items only have 1 link, hopefully it saves a bit of memory by not pre-allocating for 16 slots by default.

### DIFF
--- a/bundles/org.openhab.core.model.thing/src/org/openhab/core/model/thing/internal/GenericItemChannelLinkProvider.java
+++ b/bundles/org.openhab.core.model.thing/src/org/openhab/core/model/thing/internal/GenericItemChannelLinkProvider.java
@@ -101,6 +101,10 @@ public class GenericItemChannelLinkProvider extends AbstractProvider<ItemChannel
 
         Map<ChannelUID, ItemChannelLink> links = itemChannelLinkMap.get(itemName);
         if (links == null) {
+            // Create a HashMap with an initial capacity of 2 (the default is 16) to save memory
+            // because most items have only one channel. A capacity of 2 is enough to avoid
+            // resizing the HashMap in most cases, whereas 1 would trigger a resize as soon as
+            // one element is added.
             itemChannelLinkMap.put(itemName, links = new HashMap<>(2));
         }
 
--- a/bundles/org.openhab.core.model.thing/src/org/openhab/core/model/thing/internal/GenericItemChannelLinkProvider.java
+++ b/bundles/org.openhab.core.model.thing/src/org/openhab/core/model/thing/internal/GenericItemChannelLinkProvider.java
@@ -136,9 +136,7 @@ public class GenericItemChannelLinkProvider extends AbstractProvider<ItemChannel
             // we remove all binding configurations that were not processed
             Map<ChannelUID, ItemChannelLink> links = itemChannelLinkMap.remove(itemName);
             if (links != null) {
-                for (ItemChannelLink removedItemChannelLink : links.values()) {
-                    notifyListenersAboutRemovedElement(removedItemChannelLink);
-                }
+                links.values().forEach(this::notifyListenersAboutRemovedElement);
             }
         }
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.model.thing/src/org/openhab/core/model/thing/internal/GenericItemChannelLinkProvider.java
```diff
@@ -98,15 +99,16 @@ private void createItemChannelLink(String context, String itemName, String chann
             previousItemNames.remove(itemName);
         }
 
-        Set<ItemChannelLink> links = itemChannelLinkMap.get(itemName);
+        Map<ChannelUID, ItemChannelLink> links = itemChannelLinkMap.get(itemName);
         if (links == null) {
-            itemChannelLinkMap.put(itemName, links = new HashSet<>());
+            itemChannelLinkMap.put(itemName, links = new HashMap<>(2));
```
J-N-K, line 104:
```suggestion
            itemChannelLinkMap.put(itemName, links = new HashMap<>(2));
```

Why `2`?

bundles/org.openhab.core.model.thing/src/org/openhab/core/model/thing/internal/GenericItemChannelLinkProvider.java
```diff
@@ -128,9 +130,9 @@ public void stopConfigurationUpdate(String context) {
         }
         for (String itemName : previousItemNames) {
             // we remove all binding configurations that were not processed
-            Set<ItemChannelLink> links = itemChannelLinkMap.remove(itemName);
+            Map<ChannelUID, ItemChannelLink> links = itemChannelLinkMap.remove(itemName);
             if (links != null) {
-                for (ItemChannelLink removedItemChannelLink : links) {
+                for (ItemChannelLink removedItemChannelLink : links.values()) {
                     notifyListenersAboutRemovedElement(removedItemChannelLink);
                 }
```
J-N-K, line 137:
```suggestion
                links.values().forEach(this::notifyListenersAboutRemovedElement);
```

J-N-K (changes requested):
Otherwise LGTM. Nice finding.

bundles/org.openhab.core.model.thing/src/org/openhab/core/model/thing/internal/GenericItemChannelLinkProvider.java
jimtng: The default is 16 with .75 load factor. That seems like a waste of memory because most items have just one link.

If I set it to one, as soon as one element is added, it would trigger a rehash and capacity expansion anyway, because it would have exceeded the load factor. So 2 is chosen here.

Tbh this is just a guess based on reading about the description of initial capacity and load factor from hashmap's doc. I haven't done actual memory analysis tests to verify.

jimtng: I've added a comment in the code for future reference

J-N-K (approved):
Thanks!

PR #4076, opened 2024-02-01
.things file: fix change detection

### LINKED ISSUE #3235: [RFC] Change loading mechanism for *.things and *.items files

jimtng:
Fixes #3235

General Problem Description:
When a .things file contains multiple Things and only one thing was changed, update notifications will be sent for _all_ the things inside the file.

It should only send an update notification for the one thing that changed, and not the other things.

This is caused by two issues:
- The old things weren't removed (causing memory leak?), resulting in accumulation of multiple versions and comparing the new one against the old one resulting in erroneous update
-  Numeric values (usually entered as integer) in a newly loaded Channel Configuration properties (in the newThing) are stored as BigDecimal with Scale 0, but subsequent normalization changed it to scale 1 (in the oldThing). This makes the equals(oldThing, newThing) to always return false even when nothing has changed, which leads to calling notifyListenersAboutUpdatedElement erroneously.

### DIFF
--- a/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/ConfigUtil.java
+++ b/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/ConfigUtil.java
@@ -177,10 +177,12 @@ public class ConfigUtil {
         if (configDescriptionParameter != null) {
             Normalizer normalizer = NormalizerFactory.getNormalizer(configDescriptionParameter);
             return normalizer.normalize(value);
-        } else if (value instanceof Boolean || value instanceof String || value instanceof BigDecimal) {
-            return value;
+        } else if (value instanceof Boolean) {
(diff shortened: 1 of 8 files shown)


lolodomo:
@jimtng : conflicts is mentioned

J-N-K (approved):
LGTM

J-N-K:
Can you rebase?

jimtng:
rebased

lolodomo:
Integration tests failed in org.openhab.core.automation

jimtng:
checks passed now

lolodomo:
Looks good.

J-N-K (approved):
Thanks!

Issue #4077, opened 2024-02-02
[UoM] Change default for Number:Dimensionless tio %


rkoshak:
See the related issue #3854 .

This request is to change the default unit for Dimensionless from ONE to %. By a very large amount, those who use Dimensionless are doing so for % (e.g. humidity). ppm and db are probably used more often than ONE, meaning that ONE is likely the *least* commonly used unit making it a really poor default.

This has been causing problems particularly for users of OH who upgrade from OH 3 to OH 4 as suddenly their percent Items are between 0 and 1 instead of 0 and 100 and their charts get messed up. This is somewhat mitigated for new Items by the fact that MainUI now allows the setting of the `unit` when the Item is created but it's catching many existing users.

It is a particularly confusing problem for many users because the binding may push a state description pattern so that the Item's state appears as a % in most places it's seen in the UI but the Item is actually carrying a state of unit ONE. "The overview tab shows 45%, why does my chart show .45?"

## Your Environment
OH 4.1.1 running in Docker but that's not really relevant to this.

J-N-K:
Please see my comment in the PR.

andrewfg:
Perhaps also it is worth noting that strictly speaking the unit for humidity is properly not "%" but rather "% r.H." ..

mherwege:
> Perhaps also it is worth noting that strictly speaking the unit for humidity is properly not "%" but rather "% r.H." ..

It is a ratio of densities expressed in %, so dimensionless: https://byjus.com/physics/unit-of-humidity/#:~:text=The%20absolute%20humidity%20unit%20is,per%20cubic%20metre%20of%20air. Writing r.H. behind it (relative Humidity) is clarification, but does not change the unit in my view.

andrewfg:
Yes, "% r.H." _extends_ "%" (to paraphrase in Java class syntax)

andrewfg:
I think the webui PR which has been merged will have resolved this @rkoshak .. so probably this issue can be closed. ??

Issue #3854, opened 2023-10-19
[RfC] Make it possible for bindings to provide unit hints


jlaur:
#3481 was quite a leap forward in terms of making UoM predictable and reliable.

Now that the smoke has cleared, I see a small opportunity to improve/fine-tune the user experience without changing anything in the UoM logic itself.

A few examples where user experience could improve:
- A binding integrating a washing machine or dishwasher provides water consumption for a `Number:Volume` channel. Default metric unit **m³** does not make much sense in this context. In most cases it should be either litres or gallons.
- A binding integrating a car provides distances for a `Number:Length` channel. Default metric unit **m** is not ideal. In most cases **km** would be preferred (similar for imperial).
- A binding provides relative humidity in percent for a `Number:Dimensionless` channel. Default unit **one** is not ideal, **%** would be expected.
- A binding provides loudness in decibel for a `Number:Dimensionless` channel. Default unit **one** is not ideal, **dB** would be expected.

Since the decision of which unit to use is (and should be) entirely up to the user, my only suggestion would be to add a possibility for bindings to give a **hint** about recommended units. The only usage of this hint would be for the UI to suggest reasonable default units when creating new items for channels.

In other words, this will not control or override anything, or make the inner UoM logic more complicated in any way. It would exist in parallel as something exposed by the relevant REST methods to be consumed by the UI in order to better assist the user. Additionally, it should be easy/light-weight for bindings to provide such hints, and the bindings should not have to deal with `LocaleProvider` etc. or duplicate any logic that could be implemented in core and webui.

The way I see this could be hinted by bindings would be to extend channel and channel-type definitions with an optional new tag. For example:
```xml
<channel-type id="humidity">
	<item-type>Number:Dimensionless</item-type>
	<label>Humidity</label>
	<description>Relative humidity</description>
	<category>Humidity</category>
	<state readOnly="true"/>
	<unit-hint>%</unit-hint>
</channel-type>
```

Or perhaps:
```xml
<channel-type id="humidity">
	<item-type unitHint="%">Number:Dimensionless</item-type>
	<label>Humidity</label>
	<description>Relative humidity</description>
	<category>Humidity</category>
	<state readOnly="true"/>
</channel-type>
```

And likewise for channel:
```xml
<channel id="humidity" typeId="system.atmospheric-humidity">
	<label>Humidity</label>
	<description>Relative humidity</description>
	<unit-hint>%</unit-hint>
</channel>
````

[...]

Issue #3552, opened 2023-04-14
Default imperial unit for Length changed from inches to feet


robnielsen:
The default imperial unit for Length is Inches (see https://www.openhab.org/docs/concepts/units-of-measurement.html), but now it's reporting units in feet.  It has been inches since openHAB 1.

```
2023-04-03 17:57:48.828 [INFO ] [openhab.event.ItemStateUpdatedEvent ] - Item 'nRainHour' updated to 0 ft
2023-04-03 17:57:48.835 [INFO ] [openhab.event.ItemStateUpdatedEvent ] - Item 'nRainToday' updated to 0 ft
2023-04-03 17:57:48.837 [INFO ] [openhab.event.ItemStateUpdatedEvent ] - Item 'nRainWeek' updated to 0 ft
2023-04-03 17:57:48.838 [INFO ] [openhab.event.ItemStateUpdatedEvent ] - Item 'nRainMonth' updated to 0.025590551181102362 ft
```

robnielsen:
It's raining here now.

`2023-04-15 08:31:04.950 [INFO ] [openhab.event.ItemStateUpdatedEvent ] - Item 'nRainHour' updated to 6.561679790026247E-4 ft`

J-N-K:
As I already said: You can set the state description to `%.2f in` and it's fine. 

I see your use-case (and I agree that it is very common to use `"` for inch and `'` for foot), but I also see that `"` is used as alias for seconds and `'` for minutes. This is a problem, because unit labels are global, they are not restricted to the dimension "length" or "time". So we can't infer the dimension from the unit. The question is whether we think that `"` for inch is far more common than for seconds, in that case we could probably implement it.

robnielsen:
It's definitely a regression from prior behavior (pre 4.0), especially when dealing with devices that report values in mm or less.  For example the Netatmo rain gauge has an accuracy of 0.1 mm which is 0.000328084 ft.

robnielsen:
Thinking more about this, I can't think of any use cases that would prefer feet over inches for home automation.

spacemanspiff2007:
@J-N-K 
This issue should be gone with the [new UoM concept](https://github.com/openhab/openhab-core/issues/3282) in 4.0, however this once again shows that implicitly picking a default unit and scale (in this case length) will always lead to confusion.
So I am once again begging you:

Please let's not do
```
Number:Length MyItem
```
which will currently default to 
```
Number:Length MyItem {unit="ft"}
```
[...]

PR #4078, opened 2024-02-02
[UoM] Change Dimensionless default unit to percent

mherwege:
Closes https://github.com/openhab/openhab-core/issues/4077

There has been a lot of debate and confusion about the default unit for QuantityType<Dimensionless>. There is some consensus that the most used case for Dimensionless is percent, and therefore it would make more sense to have this as the default unit.

This PR is breaking as anyone that is using the current default Unit.ONE in persistence will see there persisted data scale. It will require manually changing the unit for this item to Unit.ONE. The expectation is there are very few cases where Unit.ONE is the right unit.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/i18n/I18nProviderImpl.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/i18n/I18nProviderImpl.java
@@ -402,7 +402,7 @@ public class I18nProviderImpl
         addDefaultUnit(dimensionMap, DataAmount.class, Units.BYTE);
         addDefaultUnit(dimensionMap, DataTransferRate.class, Units.MEGABIT_PER_SECOND);
         addDefaultUnit(dimensionMap, Density.class, Units.KILOGRAM_PER_CUBICMETRE);
-        addDefaultUnit(dimensionMap, Dimensionless.class, Units.ONE);
+        addDefaultUnit(dimensionMap, Dimensionless.class, Units.PERCENT);
         addDefaultUnit(dimensionMap, ElectricCapacitance.class, Units.FARAD);
         addDefaultUnit(dimensionMap, ElectricCharge.class, Units.COULOMB);
         addDefaultUnit(dimensionMap, ElectricConductance.class, Units.SIEMENS);


J-N-K:
You may be able to convince my position is wrong, until then I put a veto on this one.

Changing the default unit of a dimension is a breaking change which may result in big issues in a lot of systems, especially when it comes to such a fundamental one like `Dimensionless`. Please see the discussion in https://github.com/openhab/openhab-core/issues/3552 for a change that only affected users of the imperial system and a far less used dimension (`Length`). 

I fail to the the big benefit to change the default unit which has been `ONE` since the introduction of UoM in https://github.com/eclipse-archived/smarthome/pull/4818 back in 2018.

rkoshak:
- https://community.openhab.org/t/unit-of-sensor-value-one-instead-of-or-ppm/143704
- https://community.openhab.org/t/oh-3-to-oh-4-most-percentage-values-are-now-divided-by-100/153460
- https://community.openhab.org/t/oh-4-units/148410
- https://community.openhab.org/t/uom-implementation-need-help/153479
- https://community.openhab.org/t/dynamic-icon-batterylevel-not-showing-what-i-expected/153386
- https://community.openhab.org/t/calculate-temperature-real-feel/152773/2
- https://community.openhab.org/t/networkupstools-gives-wrong-value-for-battery-charge-after-updating-to-openhab-4-0/152723
- https://community.openhab.org/t/humidity-and-sitemap-graph/152651/7
- https://community.openhab.org/t/migration-from-3-4-5-to-4-0-3-problem-with-battery-level-of-mi-home-devices/152374

I could go on but you get the point. This is the #1 most common problem I deal with on the forum since OH 4 was released. 

And while it's a breaking change, `ONE` is the *least* useful and used unit in Number:Dimensionless. It's a really bad default. I don't know how to properly structure a search but as far as I can tell, there are no bindings that publish anything in ONE. The vast majority are publishing % with a few ppm and db here and there based on what I could find with my limited search skills. 

And the problem is particularly confusing to end users because of the way bindings can pass state description with a unit for display. These users end up having a Number:Dimensionless that *appears* to be a % everywhere they look until they try to use it in a Chart or use the battery level icon or something. Then all of a sudden the value is two orders of magnitude off and they have no idea why. (Number:Dimensionless isn't the only unit that has this problem BTW).

It may have been the default since UoM was first introduced, but with the changes to UoM in OH 4 we introduced a new interaction between the state description and the Item state which is causing no end of problems and confusion to users. Previously, because the State Description was used as the `unit`. 

In #3854 binding developers are advocating for introducing a whole new mechanism to bindings to provide hints for what unit to use by default. But the main thing driving them to this is %. 

For the sake of end users I think we have three choices:

1. make Number:Dimensionless default to %, by far the most used unit
2. allow bindings to provide a unit hint so when the Item is linked a reasonable unit is chosen
3. no longer allow bindings to push state description patterns to the Item

If you veto 1, then I suggest 3 is the only reasonable option, even though it will result in a lot of disruption to binding developers and end users. 2 has already been vetoed and I'm not actually sure it's a viable option anyway, but it's the only way to address this problem in a non-breaking way.

The big thing that makes this distinct from #3552 is I would be surprised if there are more than a dozen users among *all* OH users who are using ONE on purpose.

J-N-K:
I like 2, as long as it is a suggestion. Users can decide to follow that (for new items) or not. It doesn't break anything, already existing installations are not affected. 

I think "automatically set the unit metadata when a link is created" as a core-feature is wrong, it breaks the separation of things and items.

spacemanspiff2007:
It would be very unexpected if `Number:Dimensionless` would default to percent so I'm against this, too.
I'll throw in a 4th option which would be a new type ``Number:Percent``. Even if it's no unit per se its so commonly used that it would justify an own type. This new type would default to `Number:Dimensionless` with metadata `unit=%` forced.

rkoshak:
> I like 2, as long as it is a suggestion. Users can decide to follow that (for new items) or not. It doesn't break anything, already existing installations are not affected.

What's important is this has to be what gets applied to the Item if the user takes no action. It becomes *the* default, which was expressly rejected previously. 

If it's just a suggestion that, if ignored, causes `ONE` to be the default, we haven't actually addressed the problem. We've just added more stuff that the user ends up right back where they are now with an Item they expected to be and has always been a % but is now a ONE.

However, it could be implemented at the MainUI level. MainUI already shows the default unit when creating the Item and offers a chance to change that unit prior to creating the Item. If that default becomes whatever the binding is pushing and that default results in `unit` metadata being created whether it was changed or not, from the end user's perspective that would work I think. *All* new Items would end up with `unit` metadata based on their choices or failure to make choices when creating the Item.

This didn't used to be a problem because the binding essentially got to set the unit. It's a problem now because we expressly forbid the binding from setting the unit, but still let them change how the Item's state appears. 

> I think "automatically set the unit metadata when a link is created" as a core-feature is wrong, it breaks the separation of things and items.

I agree. I've disliked this for a long time. But that's how State Description works now. Seems like that would be an even bigger breaking change than changing the default unit of Number:Dimensionless. It definitely would be more disruptive.

> It would be very unexpected if Number:Dimensionless would default to percent so I'm against this, too.

And it's equally unexpected for these users who have always had Number:Dimensionless Items which were "defaulting" to `%` because of the State Description Pattern pushed by the binding but now their Item's state is `ONE`.  Only it's even more than unexpected, it's inexplicable, since by all appearances the Item still is a `%` because of that State Description Pattern, it's just not behaving like one.

mherwege:
Implementing 2 provides a solution for new items defined from thing channels through mainUI. It does not help for users migrating from a previous version, or users using textual configuration.

jlaur:
> However, it could be implemented at the MainUI level. MainUI already shows the default unit when creating the Item and offers a chance to change that unit prior to creating the Item. If that default becomes whatever the binding is pushing and that default results in `unit` metadata being created whether it was changed or not, from the end user's perspective that would work I think. _All_ new Items would end up with `unit` metadata based on their choices or failure to make choices when creating the Item.

This is exactly what I proposed in #3854.

spacemanspiff2007:
> And it's equally unexpected for these users who have always had Number:Dimensionless Items which were "defaulting" to `%` because of the State Description Pattern pushed by the binding but now their Item's state is `ONE`. Only it's even more than unexpected, it's inexplicable, since by all appearances the Item still is a `%` because of that State Description Pattern, it's just not behaving like one.

But that's a migration issue, if we set the default to `%` it'll be a logical flaw because the default of dimensionless is not `%`.
I don't want to downplay the confusion and the issues users are facing but the solution can't be to introduce something that is logically flawed.

Imho the root cause is that the users don't have a unit set for a UoM item so that's what we should address.
Why not log a warning/info on startup/when *.items file is loaded that there is a UoM item without metadata `unit`?
There's a good chance that it might not behave like the user intended so it always makes sense to explicitly pin the unit.
HABApp already does this and I've got no complains from the user.
If you pick a UoM item, you have to pick a unit. That's a simple rule that everyone can understand.

The other issue is a usability issue.
Users have pick `Number:Dimensionless` and additionally set `unit=%`.
Percentage values are prevalent, so it makes sense to provide a shortcut.
I've suggested `Number:Percent` which will pin `unit=%` without further user interaction.

These are very small changes which should sufficiently address the issue.

Currently the UI already suggests a unit so I don't think we gain much if we introduce a way for the binding to suggest the unit.
The user already has to select the correct unit for the item type when creating the item so he can make a click more to select the appropriate unit (if he wants a different one).
I don't think it's worth the added complexity at all to make the bindings suggest something (which will probably work only in < 80% of the cases anyway and/or just for simple things).

andrewfg:
FWIW I was wondering what could be the scale of this issue, so I did some grepping on the add-ons source code..

- there are ca. 20 bindings containing the string `Units.ONE` in .java files
- there are ca. 111 bindings containing the string `Units.PERCENT` in .java files
- there are ca. 115 bindings containing the string `Number:Dimensionless` in thing .xml files

There are probably a few more add-ons (like [tado](https://github.com/openhab/openhab-addons/issues/16358)) which don't (yet) return any kind of QuantityType .. but which should probably do so..

andrewfg:
> Why not log a warning/info on startup/when *.items file is loaded that there is a UoM item without metadata unit?

+1

> If you pick a UoM item, you have to pick a unit. That's a simple rule that everyone can understand.

+1

> Percentage values are prevalent, so it makes sense to provide a shortcut.
> I've suggested Number:Percent which will pin unit=% without further user interaction.

+1

mherwege:
> But that's a migration issue, if we set the default to `%` it'll be a logical flaw because the default of dimensionless is not `%`.

That is arbitrary (just like in OH km/h is the default for speed, while the base unit in physics would be m/s).

I agree with the principal of always having to set a unit when creating a QuantityType item. I also agree with logging if it is not defined in textual configurations (+ documenting it should always be defined).
I don’t think creating artificial dimensions (percent) is the way to go though. I understand why you would want it. However, it makes conversion to Units.ONE impossible (express percent as ratio), and it means all bindings using Dimensionless to express percent should be changed (over 100). You could in theory leave it, but it creates another way to express the same thing. I see a major issue with backward compatibility also.
If not changing the default unit (and I understand why one would not want to do it), I do think the unit-hint concept could play a role in the UI to present the most meaningful unit, and bindings could provide that. I don’t see a way to do that without extending the channel or channel type configurations though.

spacemanspiff2007:
> I don’t think creating artificial dimensions (percent) is the way to go though. I understand why you would want it. However, it makes conversion to Units.ONE impossible (express percent as ratio), and it means all bindings using Dimensionless to express percent should be changed (over 100). You could in theory leave it, but it creates another way to express the same thing. I see a major issue with backward compatibility also.

Just in case you misunderstand my intention I'd like to clarify: 
The idea is that `Number:Percent` will create `Number:Dimensionless`with `unit` forced to `%`.
This would be a fully backwards compatible change, think of it as a shortcut for the user.

mherwege:
> The idea is that `Number:Percent` will create `Number:Dimensionless`with `unit` forced to `%`.

So, it would only live in mainUI? Textual configuration would always have to be `Number:Dimensionless`? And once created it will show as `Number:Dimensionless`. I could see that working, but expect questions like it doesn’t save the `Number:Percent` setting. And how do we make the UI propose `Number:Percent` in the first place when creating an item from a channel? I fear we are back to the unit-hint concept, or introducing dimension synonyms.

jlaur:
> Why not log a warning/info on startup/when *.items file is loaded that there is a UoM item without metadata `unit`?

What is the difference between missing unit metadata on a managed item and an item defined in an .items file, i.e. why would you only log this for file-based configuration?

I primarily use file-based configuration and have some UoM items without explicitly provided unit metadata. Since unit metadata is optional, I would find it quite annoying to have anything logged because of this. For example, I have items of type `Number:Time` which I don't persist, and where default unit is fine since I don't care about the internal representation. I already have state descriptions etc. in place where I want to control the appearance (e.g. time remaining for a media player in a sitemap).

I also have `Number:EnergyPrice` items without a unit provided. In 4.2 with a currency provider, providing a unit could cause conversions between currencies using an online service. In some cases this is desired, but I could also imagine a few scenarios where this might not be, and the lack of unit metadata could be some safe-guard to not silently ignore a currency mismatch in the configuration (i.e. between a Thing configuration and linked item).

Also, wouldn't the somehow degrade the regional settings/measurement system? Currently you can have unpersisted items without unit provided, and I believe the configured measurement system will then be taken into consideration when determining default unit? So you can change the unit at once for all these items by simply changing the measurement system. I know this is dangerous for persisted items without unit metadata, in which case it's of course preferable to explicitly define the unit. But at least the user has the freedom to choose.

I guess what I'm saying is: If we add such logging, isn't that the first step towards making unit metadata mandatory? I see this as an early warning/reminder before that happens. If we want to keep it optional, it shouldn't cause any warnings to be logged, as this would raise some concerns for perfectly valid configurations, and worse, it would be logged on each system startup, when the user really don't want to provide these units?

mherwege:
Would we then say we only need to force a unit when persisting? It is a totally different change of course, but we might look at forcing to set a unit in the persistence configuration for QuantityTypes. That would require extending textual and UI persistence configuration. The persistence unit would then be by item and persistence service.

J-N-K:
Just some thoughts:

- There is no difference between managed and un-managed items, they behave the same.
- I agree with @jlaur that a valid configuration should not log warnings. 
- There was a suggestion similar to mandatory unit-metadata (`Number:m/s`, `Number:%`, `Number:s`) when we discussed the UoM changes, but it didn't find much support. I have some doubts that mandatory metadata would be much more popular, especially when it its not important for most users.
- I don't think "artificial dimensions" should be introduced and adding `Number:Percent` as an alias to `Number:Dimensionless ... [ unit="%" ]` is also not the way to go. We would end up with a tons of such aliases and much confusion. E.g. what happens if I configure `Number:Percent ... [ unit="ONE" ]`?

lolodomo:
The idea of the log message is not bad at all but of course we should log something only for `Number:Dimensionless` as it is the only case where we all know that the default unit ONE will be inappropriate in 99.99% of cases.
But I would prefer an INFO message.

andrewfg:
> Would we then say we only need to force a unit when persisting? It is a totally different change of course, but we might look at forcing to set a unit in the persistence configuration for QuantityTypes. That would require extending textual and UI persistence configuration. The persistence unit would then be by item and persistence service.

Actually, the real, (and perhaps only), problem is that persistence is wrongly designed. Currently each slot in a persistence database is storing only the float part of the data. Whereas it should be storing the full QuantityType i.e. a float and a unit. If that were the case, a time series of persisted data would always be consistent, even across a change in units. @mherwege suggestion to bolt on a unit to the data series, via the persistence configuration, outside the actual data series, is also a hack. The proper solution is to rework persistence to store and retrieve QuantityTypes in addition to floats.

A further nice to have would be a migration tool to convert legacy persistence data streams to QuantityType data streams by manually providing a unit to the conversion tool.

jlaur:
> Actually, the real, (and perhaps only), problem is that persistence is wrongly designed.

This is an implementation issue in the concrete persistence services rather than a general design issue. At least I don't think anything prohibits the persistence services from storing unit information as well?

Personally I prefer a fixed unit in my MySQL tables so I can utilize the data outside openHAB without having to deal with conversions. But having e.g. a unit column shouldn't cause any problems except storage space (because it would usually be very redundant).

jlaur:
> Would we then say we only need to force a unit when persisting?

Just a short note here for completeness: The impact of unit issues is probably most severe for persistence. But it is not the only place.

For rules, a user could either:
- Define unit on item and use raw value knowing that unit.
- Leave default unit on item and convert quantity value to desired unit in the rule.
- Do both.
- Do neither and face issues.

spacemanspiff2007:
> What is the difference between missing unit metadata on a managed item and an item defined in an .items file, i.e. why would you only log this for file-based configuration?

My suggestions was to log it for all items on startup and when loading an *.items file it's only logged for the items defined in the file. Both items defined through gui and file should behave the same.


> So you can change the unit at once for all these items by simply changing the measurement system.

This is a non-feature because how often do you switch your system from e.g. metric to imperial. And even if you want to do it's still possible by editing all the items, it's just not that convenient. 
Also afaik when creating an item the GUI already suggests a `unit` so for these items a measurement system switch would already not work.


> * We would end up with a tons of such aliases and much confusion.

Do you already think of other aliases? If it's only one or two it'll be fine if there are many this should be rejected. 

> * E.g. what happens if I configure `Number:Percent ... [ unit="ONE" ]`?

That would obviously not allowed, that's why I wrote "force to %". Changing the `unit` for `Number:Percent` should result in an error and the change should be rejected.


> Currently each slot in a persistence database is storing only the float part of the data. Whereas it should be storing the full QuantityType i.e. a float and a unit.

Some persistence services don't support persisting a unit or strings so that would not work there.
Also the normalized value is not only used in persistence and rules but also when interacting with external systems e.g. for the HABApp rules. So that would effectively break the integration.

mherwege:
> Some persistence services don't support persisting a unit or strings so that would not work there.

Totally agree, and I am not saying to store the unit, but to set  a unit parameter in the persistence strategy for the item, so it gets converted to the unit before persisting. If the parameter is not set, use the default. The challenge is obviously that persistence is often defined or group (or all) level, so not the item level.
This is just a wild idea and is definitely more complex to pull off.



> Also the normalized value is not only used in persistence and rules but also when interacting with external systems e.g. for the HABApp rules. So that would effectively break the integration.

Why is that? An item state for a QuantityType item contains the unit, also in the REST response. Does HABApp drop it by default and assume the default? That negates the whole value of units in my view. All the info is available in the REST responses.

spacemanspiff2007:
> Why is that? An item state for a QuantityType item contains the unit, also in the REST response. Does HABApp drop it by default and assume the default? That negates the whole value of units in my view. All the info is available in the REST responses.

Almost - HABApp just drops the unit and treats the value as if it would be a value without unit.
That's why it's important to to set the `unit` metadata so the values don't change the scale unexpectedly and that's why HABApp logs a warning.
It's really a non issue because at some time you have decide on a unit and scale to do the calculations.
On the item level the benefit of UoM is questionable at best so this is not an issue.
Obviously if you want to bring UoM to HABApp PR welcome ;-)

andrewfg:
> My suggestions was to log it for all items on startup and when loading an *.items file it's only logged for the items defined in the file.

Perhaps use a console command instead of logging?

spacemanspiff2007:
> Perhaps use a console command instead of logging?

But if I have to run a console command I already know that there might be an issue.
Isn't the goal to reach the users who simply forget to supply the unit or don't know they should supply one at all?

The intersection of users who know how to run the console command and users who face this issue is probably very small, too.

andrewfg:
> if I have to run a console command I already know that there might be an issue

Ok. You are right.

rkoshak:
> I've suggested Number:Percent which will pin unit=% without further user interaction.

One concern I have is adding another source of confusion similar to what we have for Location. It's an Item and it's a semantic model category. I've had to help more than one user who created Location Items instead of Groups, with the semantic model Location tag and they can't figure out why they can't add Equipment to it.

I fully expect users to become confused between a Number:Percent and PercentType.

> providing a unit could cause conversions between currencies using an online service. In some cases this is desired, but I could also imagine a few scenarios where this might not be, and the lack of unit metadata could be some safe-guard to not silently ignore a currency mismatch in the configuration (i.e. between a Thing configuration and linked item).

To truly protect yourself from this scenario, don't use a Number:EnergyPrice Item in the first place and just use a Number Item. 

> Currently you can have unpersisted items without unit provided, and I believe the configured measurement system will then be taken into consideration when determining default unit?

The default unit is chosen based on what the default for that type of unit is for the selected measurement system selected in regional settings. A user who has a Number:Length without unit `metadata` with SI selected I think will get a default of `m` whereas one with Imperial selected will get `ft`. The user has the option to change this default by setting the `unit` metadata.

Whether or not the Item is persisted is irrelevant. 

> That would require extending textual and UI persistence configuration.

What would need to change for the UI?

> especially when it its not important for most users.

I think it is important for more users than you might think. 

> Also afaik when creating an item the GUI already suggests a unit so for these items a measurement system switch would already not work.

I believe the suggestion only results in `unit` metadata actually being added to the Item if the user changes it from the suggested default. So only where the user didn't change this will still work.

> Do you already think of other aliases? 

I can see developers of the car bindings wanting Number:LongDistance (or something like that) to measure length as km or miles instead of meters and feet by default. I'm not sure what the default units for volume are but could see something similar wanted there. 

Essentially, any time a binding developer hits a situation where they have a lot of Channels not using the default unit, they're going to want an alias to reduce the need of their users setting the `unit` metadata. 

> Totally agree, and I am not saying to store the unit, but to set a unit parameter in the persistence strategy for the item, so it gets converted to the unit before persisting. If the parameter is not set, use the default.

By default I assume you mean whatever unit the Item is configured to carry (i.e. `unit` metadata or system default if the metadata is not defined).

> Perhaps use a console command instead of logging?

@J-N-K posted a rule some time ago that will tell you all your Items lacking `unit` metadata. With permission I can convert that to a rule template and post it to the marketplace in a couple of hours.

boehan:
> > Also afaik when creating an item the GUI already suggests a unit so for these items a measurement system switch would already not work.
> 
> I believe the suggestion only results in `unit` metadata actually being added to the Item if the user changes it from the suggested default. So only where the user didn't change this will still work.

I may be wrong, but looking at https://github.com/openhab/openhab-webui/pull/2302, this only seems to be true for state description.

mherwege:
> I may be wrong, but looking at [openhab/openhab-webui#2302](https://github.com/openhab/openhab-webui/pull/2302), this only seems to be true for state description.

That's also my interpretation. Unit metadata is always set when creating an item from a thing channel.

PR #4088, opened 2024-02-12
Fix the recursive group membership check

jimtng:
Allow a group to be a member of its direct parent and also its parent's ancestors without raising an error.

So this scenario is now allowed:
```
Group G1
Group G2 (G1)
Group G3 (G1, G2)
```

Looping membership is still detected and prevented as before, thus Stack Overflow is still avoided.
So this scenario is still not allowed, as it was originally intended:
```
Group R1 (R2)
Group R2 (R1)
```

and also not allowed:
```
Group R1 (R3)
Group R2 (R1)
Group R3 (R2)
```

Hopefully the new error message is also easier to understand, although it will only be issued when a real membership loop occurred.

See https://community.openhab.org/t/recursive-group-membership-found/135192
https://github.com/openhab/openhab-core/pull/3748


mherwege:
Thanks, fixes a warning in my logs I have been ignoring all the time.

J-N-K (approved):
LGTM.

Issue #1639, opened 2020-09-12
Use consistent capitalization with Conditions, Triggers, Actions


wborn:
Because several capitalization strategies are used the UI for selecting Conditions, Triggers, Actions in rules is a bit messy:

## Conditions

![condition](https://user-images.githubusercontent.com/12213581/93004215-3c36e300-f545-11ea-824c-1edc832ad690.png)

## Triggers

![trigger](https://user-images.githubusercontent.com/12213581/93004211-36410200-f545-11ea-867d-5e11c5d39460.png)

## Actions

![action](https://user-images.githubusercontent.com/12213581/93004208-33461180-f545-11ea-9295-b4f2e0433b67.png)

kaikreuzer:
Correct capitalization is lower case as these strings are supposed to be read as a complete sentence when being combined.

wborn:
I already thought so because most are already lower case. But knowing the exact reason helps and makes sense!

wborn:
I had a look at more labels in add-ons. Many seem to add the add-on name too:

* Astro : Get Azimuth
* Volvo On Call : Honk-blink
* MPD : Send command
* Telegram answer
* publishXMPP

That may seem useful, but these actions also have a scope (e.g. `@ThingActionsScope(name = "network")`) which could be used by the framework for adding some context to thing actions so they can be more easily identified.

ghys:
> That may seem useful, but these actions also have a scope (e.g. @ThingActionsScope(name = "network")) which could be used by the framework for adding some context to thing actions so they can be more easily identified.

Interesting, would it be possible to have that in the API? These flat lists of module types (which probably should be sorted alphabetically) can get a little difficult to parse... I'm almost considering having a "simple mode" in the rules editor, at least during creation (like a wizard), where the most common types - those related to items states & commands, and perhaps scripts and time based triggers/cron - are made more prominent.

ghys:
Looking at the response to `/rest/module-types` I would tend to assume module types UIDs always contain the scope before the dot, is it correct?

`core.ItemStateChangeTrigger`
`ephemeris.WeekdayCondition`
`script.ScriptCondition`
`hue.fadingLightCommand`
`mqtt.publishMQTT`
`network.sendWakeOnLanPacket`

kaikreuzer:
Yes, I would say that this is (or at least should be!) the naming convention.

wborn:
If these scopes can be used in the UI we could remove the add-on names from the thing action labels.
[...]

PR #4090, opened 2024-02-13
Use sentence casing for ScriptActionType

jimtng:
image


J-N-K:
Wouldn't `the system has reached start level 100` instead of `When the system has reached start level 100` be the better fix? You can then read it as a sentence with the larger title `When` and `Then`.

jimtng:
This is specifically just for "execute an inline script"

jimtng:
The way I see it, there are two options:
- Change it here in core,
- Check for `script.ScriptAction` and capitalize the first letter on the fly from within webui.

I'd much rather go with option 1, but I found others that need the same treatment, e.g. in MediaActionTypeProvider.java and possibly others.

So it's _easier_ for me to just convert them all on the fly on the webui side. 

In any case the webui isn't going to add `Then ` in front of it, so it still needs to be sentence-cased _somewhere_.

wborn:
> Wouldn't `the system has reached start level 100` instead of `When the system has reached start level 100` be the better fix? 

That's how it should be, see also https://github.com/openhab/openhab-core/issues/1639

jimtng:
> That's how it should be, see also #1639

Thanks for that link! In that case, some clean up is needed in webui to make it consistent. Right now it's still a mixture of upper and lowercase.

PR #4091, opened 2024-02-13
Fix itests not running

J-N-K:
Why do unresolved tests result in "No error" instead of "Failed"?


kaikreuzer (approved):
Thanks for the fix.
Wrt the question: No idea. @wborn Would you know how to make them fail instead?

wborn:
I think this changed in bnd 7. It used to wait forever until it ran the test cases when the bundles weren't resolved . Now it seems to use a 5 sec timeout see https://github.com/bndtools/bnd/commit/4061cd3586d50707cc957a0c51a63f13360c4ce7.

Issue #3315, opened 2023-01-13
Upgrade to Jetty 10.x (or Jetty 11.x)


joakime:
Jetty 9.x is now at End of Community Support.

You should be using Jetty 10.x (or newer) at this point.

* https://github.com/eclipse/jetty.project/issues/7958

J-N-K:
We are currently upgrading Karaf to 4.4.3 with PaxWeb 8.0.15 (i.e. Jetty 9.4.50). There is some work on Karaf 4.5.x which will support PaxWeb 9 and Jetty 10 (https://lists.apache.org/thread/4lwbt9g0w707zomnq6cfx179n8w90d3h) but we'll have to check when we see a release there.

andrewfg:
@J-N-K 4yi, this relates to..

- https://github.com/openhab/openhab-addons/pull/13570
- https://github.com/eclipse/jetty.project/issues/9168

wborn:
The upgrade in Karaf is tracked by issue [KARAF-7544](https://issues.apache.org/jira/browse/KARAF-7544).

jlaur:
I was checking release notes in relation to #3770 and saw this: https://issues.apache.org/jira/browse/CXF-8764. So for this library we are now at Jetty 10? So we are now bundling both Jetty 9 and 10 because of this dependency? Is it possible in the same way to decouple the dependency for PaxWeb, so that we can use Jetty 10 at the same time as PaxWeb still uses Jetty 9?

J-N-K:
No, we currently only ship Jetty 9, as far as I can see we don't use the bundles that require jetty 10.

We could try if we can upgrade to PaxWeb 9 after Karaf 4.4 (because then we have SLF4J 2.0 available).

joakime:
BTW, Jetty 12 was released.   (with support for EE8 at `java.servlet`, EE9 at `jakarta.servlet`, and EE10 at `jakarta.servlet` at the same time)

lolodomo:
Any idea what will be required to change in bindings when moving to Jetty 10 ?

J-N-K:
From my understanding changes are only required for websockets, see https://eclipse.dev/jetty/documentation/jetty-10/programming-guide/index.html#pg-migration

lolodomo:
Ok, we could then search in code for all the classes that have to be renamed.

andrewfg:
[...]

PR #4092, opened 2024-02-13
[net] Expose Jetty EndPoint in http requests

GiviMAD:
Hello, I would like to merge these small workaround that allows to access the underling Jetty EndPoint when using an http client created using the WebClientFactory.

It mimics the way the EndPoint is exposed in recent Jetty versions so it will become useless in case of an update.

I need it in order to proxy the socket data for upgraded WebSocket connections through the OpenHAB Cloud Connector.

### DIFF
--- a/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/internal/WebClientFactoryImpl.java
+++ b/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/internal/WebClientFactoryImpl.java
@@ -424,7 +424,8 @@ public class WebClientFactoryImpl implements HttpClientFactory, WebSocketFactory
     }
 
     /**
-     * Extends the default {@link HttpClientTransportOverHTTP) but exposes the underling {@link EndPoint} of each request/response.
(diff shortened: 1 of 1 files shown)


GiviMAD:
Hello @J-N-K, would you able to give your opinion on whetter is it good to merge this change or if it will be better to upgrade the Jetty client to a more recent version where the EndPoint is already exposed?

Jetty 9 is quite old but still listed as stable here https://eclipse.dev/jetty/download.php#version-table.

I needed it for the cloud connector PR https://github.com/openhab/openhab-addons/pull/16403 to proxy the socket data between the cloud  and the connector though the Socket.IO connection. I chosed that solution to avoid the parsing of the websocket frames in the cloud, as the cloud is shared by many clients I think this way is the less resource intensive.

wborn:
>  whetter is it good to merge this change or if it will be better to upgrade the Jetty client to a more recent version where the EndPoint is already exposed?

The Jetty upgrade depends on Apache Karaf upgrading to a recent Pax Web version, see https://github.com/openhab/openhab-core/issues/3315. So until that becomes available the workaround makes sense.

wborn (approved):
Thanks for working on improving the WebSockets API compatibility!

GiviMAD:
Thank you for the review!

PR #4097, opened 2024-02-17
Call, Contact, Image items: Don't create command descr from state descr

florian-h05:
If a state description is set on a contact item, currently a command description is automatically created as well.
This is wrong as contact items do not accept commands (except refresh) and caused the UI to display a control for the default (list) widget for these contacts.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/items/GenericItem.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/items/GenericItem.java
@@ -468,11 +468,9 @@ public abstract class GenericItem implements ActiveItem {
 
     @Override
     public @Nullable CommandDescription getCommandDescription(@Nullable Locale locale) {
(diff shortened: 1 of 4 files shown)


J-N-K:
Doesn't that apply to `ImageItem` and `CallItem`, too?

florian-h05:
True, I can fix them too.

One thing that however needs to be thought of, is that it then won't be possible to set a command description anymore.
What if I would like to add a command description to send the REFRESH command?

We could create a protected method `GenericItem::getCommandOptions`  which onyl returns the command options, and does not fallback to state description. This method could then be used internally by `GenericItem::getCommandDescription` and by the overrides in `ContactItem`, `ImageItem` and `CallItem`. WDYT?

florian-h05:
It would be nice if we could squeeze that one into the milestone as well 👍

kaikreuzer (approved):
Let's go with that, thanks!

PR #4099, opened 2024-02-19
[rest] refactored the Stream2JSONInputStream to simplify the logic

joerg1985:
# Description

This PR will use the `SequenceInputStream` to simplify the logic for the `Stream2JSONInputStream`.
All interface methods can now be delegated to the `SequenceInputStream`, this makes it easy to support methods like:
- `public int read(byte @Nullable [] b, int off, int len) throws IOException`
- `public long transferTo(OutputStream target) throws IOException`

### DIFF
--- a/bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/Stream2JSONInputStream.java
+++ b/bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/Stream2JSONInputStream.java
@@ -15,11 +15,15 @@ package org.openhab.core.io.rest;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.SequenceInputStream;
 import java.nio.charset.StandardCharsets;
+import java.util.Enumeration;
 import java.util.Iterator;
 import java.util.stream.Stream;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.library.types.DateTimeType;
 
 import com.google.gson.Gson;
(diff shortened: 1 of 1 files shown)


joerg1985:
@J-N-K could you have a look at this one too? Thanks in advance

bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/Stream2JSONInputStream.java
```diff
@@ -32,74 +36,71 @@
...
      * Creates a new {@link Stream2JSONInputStream} backed by the given {@link Stream} source.
      *
      * @param source the {@link Stream} backing this input stream. Must not be null.
      */
     public Stream2JSONInputStream(Stream<?> source) {
-        iterator = source.map(e -> gson.toJson(e)).iterator();
-        jsonElementStream = new ByteArrayInputStream(new byte[0]);
-        firstIteratorElement = true;
-    }
+        Gson gson = new GsonBuilder().setDateFormat(DateTimeType.DATE_PATTERN_WITH_TZ_AND_MS).create();
```
J-N-K, line 52: Why do we need a new instance on every call? Can't we share it as class field?

bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/Stream2JSONInputStream.java
```diff
@@ -32,74 +36,71 @@
...
+        Iterator<String> iterator = source.map(e -> gson.toJson(e)).iterator();
 
-    @Override
-    public int read() throws IOException {
-        int result = jsonElementStream.read();
+        Enumeration<InputStream> enumeration = new Enumeration<>() {
+            private boolean eos = false;
```
J-N-K, line 56: Can you pick a more explicit name for this variable?

bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/Stream2JSONInputStream.java
```diff
@@ -32,74 +36,71 @@
...
-            fillBuffer(); // get the next element into a new jsonElementStream
-            result = jsonElementStream.read();
-        }
+            @Override
+            public InputStream nextElement() {
+                InputStream is;
+
+                if (next != null) {
+                    is = next;
+                    next = (!eos && (eos = !iterator.hasNext())) ? toStream("]") : null;
```
J-N-K, line 70: Using assignments in conditions works, but its hard to read and understand. Can you refactor that to a better readable code?

bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/Stream2JSONInputStream.java
```diff
@@ -32,74 +36,71 @@
...
+                    return is;
+                }
+
+                is = toStream(iterator.next());
+                next = (eos = !iterator.hasNext()) ? toStream("]") : toStream(",");
```
J-N-K, line 75: see above

J-N-K (changes requested):
In general a good improvement. I have left some comments.

bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/Stream2JSONInputStream.java
joerg1985: Yes, that is correct.

joerg1985:
@J-N-K i have updated the PR, thanks for the review!

J-N-K (approved):
Thanks! LGTM

PR #4106, opened 2024-02-22
unitSymbol in REST response for GroupItem

mherwege:
The item REST response does include a field `unitSymbol` for Number items, but not for Group items with Number base type. While the info can be retrieved by querying the metadata, it is easier in the UI to be able to use the same field.

This PR adds unitSymbol for Group items with base type Number in the REST response.

Required to avoid workaround for groups in https://github.com/openhab/openhab-webui/pull/2312


J-N-K (approved):
Thanks!

PR #4107, opened 2024-02-23
[rest] Add caching for add-on resource

florian-h05:
As there are no registry listeners to use, I decided to listen to `AddonEvent`s.

/cc @jimtng

### DIFF
--- a/bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/RESTConstants.java
+++ b/bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/RESTConstants.java
@@ -41,10 +41,11 @@ public class RESTConstants {
      */
     public static final String API_VERSION = "6";
 
-    public static final CacheControl CACHE_CONTROL = new CacheControl();
-    static {
-        CACHE_CONTROL.setNoCache(true);
-        CACHE_CONTROL.setMustRevalidate(true);
-        CACHE_CONTROL.setPrivate(true);
+    public static CacheControl cacheControl() {
+        CacheControl cc = new CacheControl();
+        cc.setNoCache(true);
+        cc.setMustRevalidate(true);
+        cc.setPrivate(true);
+        return cc;
     }
 }
--- a/bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/RESTConstants.java
+++ b/bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/RESTConstants.java
@@ -41,11 +41,10 @@ public class RESTConstants {
      */
     public static final String API_VERSION = "6";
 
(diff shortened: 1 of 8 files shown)


J-N-K (commented):
Are you sure this works the way you expect it to do? Server-side changes to remote add-on services (e.g. a new add-on post in the forum) do not produce an `AddonEvent` and will never be picked up until a system restart.

florian-h05:
No, I’m not sure. 

Do you know a better way to detect changes?

J-N-K:
The remote add-on services cache their content, too. The refresh time is 15 min, so I guess an automatic expiry of the cache with half of that would probably mitigate the issue.

florian-h05:
Thanks for the idea, I have just pushed that!

FYI I switched from having a `CACHE_CONTROL` constant on `RESTConstants` to having a default cache control getter on the `RESTConstant` class to migitate to risk that somewhere the const could be modified which would then apply to all places where the const is used.

florian-h05:
@J-N-K Can you please have a look at this PR?

bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/RESTConstants.java
```diff
@@ -38,4 +40,12 @@ public class RESTConstants {
...
      */
     public static final String API_VERSION = "6";
+
+    public static CacheControl cacheControl() {
+        CacheControl cc = new CacheControl();
+        cc.setNoCache(true);
+        cc.setMustRevalidate(true);
+        cc.setPrivate(true);
+        return cc;
+    }
```
J-N-K, line 50: I think haveing a static final `CACHE_CONTROL` constant would be better than creating a new object every time. I also feel that the danger these setting might be modified is very low.

florian-h05:
@J-N-K Done :+1: 

Sign-Off is missing for the revert commit, hope that's okay. Otherwise I need to force push ...

J-N-K (approved):

Nadahar:
I have just scratched my head a lot until I figured out that this was to blame. If you change your settings for whether to show unpublished or incompatible add-ons, nothing happens for a long time. It turns out that this is the reason.

I see that MainUI calls this quite a lot, which might have been the reason for doing this. Maybe it would be possible to make MainUI a bit more restrictive in repeating the call? In any way, the evaluation of preconditions must be made quite a lot smarter for this to not just come across as a bug.

For remote add-ons you already have a 15 minutes cache, which is bad enough to wait for if you try to edit something on the community marketplace and wants to verify that it works as intended. But, with these 7.5 minutes on top of that, that even applies to locally installed add-ons, make the whole thing very "unpredictable". If you install an add-on locally, the UI won't reflect that for up to 7.5 minutes.

To me, that is very close to a pure bug.

PR #4108, opened 2024-02-23
[rest] Skip ignored things in inbox by default

seime:
This PR adds a default filter to skip ignored inbox items. 
Most of the calls from MainUI uses inbox data only to show the number of _non-ignored inbox_ items. On any decent sized installation, the inbox contains hundreds of ignored items that currently is being transferred.

Only when an optional request parameter is passed, the ignored entires are returned.

This PR will be followed by a companion PR in the MainUI repo shortly.


J-N-K:
In general that sounds like a good idea, but wouldn't it make sense to have the default `true` to keep backward compatibility?

seime:
Maybe, but of the about 4 places in MainUI it was being  used, only 1 usage needed the ignored ones.

I'm not aware of any other places that utilize this API other than MainUI, but obviosly I cannot be sure. Are you aware of any?

J-N-K:
I'm not sure if 3rd party services (like HABApp) use these parts of the API. But in general we should avoid unnecessary breaking changes.

spacemanspiff2007:
Thanks for thinking about the HABApp users.
Currently we do not use the thing inbox, only the available things.

J-N-K (approved):
Thanks!

PR #4113, opened 2024-02-24
Replace deprecated `hamcrest-library` with `hamcrest` in test BOM

### LINKED ISSUE #4112: Replace deprecated `hamcrest-library` with `hamcrest` in test BOM
While developing unit tests, I noticed that the test dependencies contain `hamcrest-library` and `hamcrest-core`. These two libraries are obsolete in the mean time because everything was combined into `hamcrest`.

Both libraries contain text files and Javadocs saying:

```
All the classes in hamcrest-core.jar and hamcrest-library.jar has moved
into hamcrest.jar. Please update your dependencies.
```

I found that the dependency is defined in `org.openhab.core.bom.test/pom.xml`:

```xml
<dependency>
    <groupId>org.hamcrest</groupId>
    <artifactId>hamcrest-library</artifactId>
    <version>2.2</version>
</dependency>
```

It yields this dependency hierarchy:

```
hamcrest-library
+- hamcrest-core
     +- hamcrest
```

Since the first two libraries are deprecated/obsolete, we should directly depend on `hamcrest` instead of `hamcrest-library`. I will provide a PR soon.

david-pace:
Background: `hamcrest-library` and `hamcrest-core` are obsolete because they were combined into `hamcrest`.

Both deprecated JARs contain the following text file:

```
All the classes in hamcrest-core.jar and hamcrest-library.jar has moved
into hamcrest.jar. Please update your dependencies.
```

Resolves #4112

### DIFF
--- a/bom/test/pom.xml
+++ b/bom/test/pom.xml
@@ -34,7 +34,7 @@
     </dependency>
     <dependency>
       <groupId>org.hamcrest</groupId>
-      <artifactId>hamcrest-library</artifactId>
+      <artifactId>hamcrest</artifactId>
       <version>2.2</version>
     </dependency>
     <dependency>


wborn (approved):
Thanks!

PR #4116, opened 2024-02-28
Audio parsing : Fix byte / bit confusion

dalgwen:
Fix two errors for calculating bitrate and duration of sound.

In AudioSinkUtilsImpl : confusion between byte / bit
In AudioWaveUtils : confusion between byte / bit AND with the definition of framesize (framesize already includes number of channels)


wborn (approved):
Thanks for the fix!

jlaur:
@openhab/core-maintainers - can you add the "4.2" milestone?

PR #4117, opened 2024-02-28
Upgrade tool: Add upgrade task for script profile changes

florian-h05:
Refs #4058.

/cc @jimtng @J-N-K

### DIFF
--- a/tools/upgradetool/pom.xml
+++ b/tools/upgradetool/pom.xml
@@ -15,7 +15,7 @@
   <packaging>jar</packaging>
 
   <name>openHAB Core :: Tools :: Upgrade tool</name>
-  <description>A tool for upgrading openHAB from 3.4 to 4.0</description>
+  <description>A tool for upgrading openHAB from 3.4 to 4.x</description>
 
   <dependencies>
     <dependency>
--- a/tools/upgradetool/pom.xml
+++ b/tools/upgradetool/pom.xml
@@ -15,7 +15,7 @@
   <packaging>jar</packaging>
 
   <name>openHAB Core :: Tools :: Upgrade tool</name>
-  <description>A tool for upgrading openHAB from 3.4 to 4.x</description>
+  <description>A tool for upgrading openHAB</description>
 
   <dependencies>
     <dependency>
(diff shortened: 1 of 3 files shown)


tools/upgradetool/pom.xml
```diff
@@ -15,7 +15,7 @@
   <packaging>jar</packaging>
 
   <name>openHAB Core :: Tools :: Upgrade tool</name>
-  <description>A tool for upgrading openHAB from 3.4 to 4.0</description>
+  <description>A tool for upgrading openHAB from 3.4 to 4.x</description>
```
J-N-K, line 18:
```suggestion
  <description>A tool for upgrading openHAB</description>
```

tools/upgradetool/src/main/java/org/openhab/core/tools/internal/Upgrader.java
```diff
@@ -42,11 +43,15 @@
...
  *
  * @author Jan N. Klug - Initial contribution
+ * @author Florian Hotze - Add script profile upgrade
  */
 @NonNullByDefault
 public class Upgrader {
     public static final String ITEM_COPY_UNIT_TO_METADATA = "itemCopyUnitToMetadata";
     public static final String LINK_UPGRADE_JS_PROFILE = "linkUpgradeJsProfile";
+    public static final String LINK_UPGRADE_SCRIPT_PROFILE = "linkUpgradeScriptProfile";
+    private static final Set<String> SCRIPT_PROFILE_NAMES = Set.of("transform:NASHORNJS", "transform:JS",
```
J-N-K, line 53: Wouldn't it make more sense to use `transform:.*` and match a regex? Otherwise script profiles with JRule, JavaRule or other languages would not be transformed.

J-N-K (changes requested):

tools/upgradetool/src/main/java/org/openhab/core/tools/internal/Upgrader.java
florian-h05: Hmm, this should work as well :+1:

florian-h05: Done.

J-N-K (approved):
LGTM

PR #4118, opened 2024-02-28
[rest] Set editable for members of an item

florian-h05:
This sets the `editable` field of the `EnrichedItemDTO` for `members` of the `EnrichedGroupItemDTO`.

The `editable` field is needed by the UI, e.g. for https://github.com/openhab/openhab-webui/pull/2410.

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
@@ -281,7 +281,8 @@ public class ItemResource implements RESTResource {
         Stream<EnrichedItemDTO> itemStream = getItems(type, tags).stream() //
                 .map(item -> EnrichedItemDTOMapper.map(item, recursive, null, uriBuilder, locale)) //
                 .peek(dto -> addMetadata(dto, namespaces, null)) //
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
```diff
@@ -317,7 +323,7 @@ public Response getItemNamespaces(@PathParam("itemname") @Parameter(description
     @Operation(operationId = "getItemByName", summary = "Gets a single item.", responses = {
             @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = EnrichedItemDTO.class))),
             @ApiResponse(responseCode = "404", description = "Item not found") })
-    public Response getItemData(final @Context UriInfo uriInfo, final @Context HttpHeaders httpHeaders,
+    public Response getItemByName(final @Context UriInfo uriInfo, final @Context HttpHeaders httpHeaders,
```
florian-h05, line 326: FYI I decided to rename that method to better reflect what it is actually doing.

florian-h05:
@kaikreuzer This PR would also be nice to have for the snapshot, as it is required to fix an UI bug.

kaikreuzer (approved):
lgtm, thanks!

PR #4119, opened 2024-02-28
[rest] Enable filters for cacheable items list

florian-h05:
This allows the usage of filters when using `staticDataOnly=true`.


J-N-K (approved):
Thanks!

PR #4120, opened 2024-02-29
Fix hysteresis value mapping

ArXa1L:
Hello,
I found that hysteresis value mapping works differently than stated in its documentation:

[source:](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/profiles/SystemHysteresisStateProfile.java#L163)

```java
    private Type mapValue(double lower, double upper, double value) {
        if (value <= lower) {
            return low;
        } else if (value >= upper) {
            return high;
        }
        return previousType;
    }
```

[doc:](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.thing/src/main/resources/OH-INF/config/hyteresisProfile.xml)

```xml
<parameter name="lower" type="text" required="true">
  <label>Lower Bound</label>
  <description>Maps to OFF if value is below lower bound (plain number or number with unit).</description>
</parameter>
<parameter name="upper" type="text">
  <label>Upper Bound</label>
  <description>Maps to ON if value is above upper bound (plain number or number with unit).</description>
</parameter>
```

There are two ways to solve this the difference:
- update description 
- change the code

I decided that changing the value mapping would be more correct, because... when configuring an item, OpenHAB users most likely look at the documentation rather than the code


J-N-K:
Looking at the failed tests, it seems that the code behavior is correct (or at least intended). Is there any special reason why you would like to exclude the boundaries?

ArXa1L:
Hi, thank you for your answer!

My motivation was that when setting the hysteresis, I specified the boundaries based on the documentation. My switch didn't work well - it turned on and off noticeably earlier than expected. 

Now, given the failed tests, I think it's better to improve the documentation rather than change the boundaries' behavior. What do you think about such changes?

J-N-K:
Yes, improving the docs would be preferred.

Issue #4114, opened 2024-02-26
ColorUtil bugs and improvements


andrewfg:
There are a number of bugs in, and potential improvements to, the ColorUtil module as follows.

### 1. hsbToXy() method produces invalid xyY results

The hsbToXy() method produces invalid results. Test case as follows..

```
// test case => produces invalid color gamut::blue
HSBType hsb = new HSBType(String.format("%d,%d,%d", 0, 50, 0));
double[] xyG = ColorUtil.hsbToXY(hsb);
```

Reason is that when B part of HSB is zero, the ColorUtil.hsbToXY(hsb) method follows a special code path that produces the xyY value xyY[0,0,0] which is an invalid color which causes the gamut correction to convert that to gamut:blue.

### 2. xyToHsb() method produces invalid results

The xyToHsb() method produces invalid HSB results when the xy co-ordinate is outside the color gamut.

```
// test case 2a) => produces invalid color gamut::blue
double[] xyG = new double[] {0,0,0};
HSBType hsb = ColorUtil2.xyToHsb(xyG);
```

Test case 2a) reason is the analogous inverse case of case 1. above.

```
// test case 2b) => produces invalid HSB B part
double[] xyG = new double[] {0.05,0.65,1.0};
HSBType hsb = ColorUtil2.xyToHsb(xyG);
```
Test case 2b) reason is that if the incoming xy coordinate is outside the color gamut, the gamut correction converts the coordinate to the closest coordinate within the gamut, however it does not adjust the Y element match.

### 3. Round trips hsbToXy() => xyToHsb() are not always consistent

In some cases the round trip hsbToXy() => xyToHsb() produces an HS(B) output that is widely different than the HS(B) input.

### 4. hsbToRgbw() method throws exceptions

For certain HSB values the hsbToRgbw() method throws a 'Non-terminating decimal expansion; no exact representable decimal result'. Test case as follows..

```
HSBType hsb = new HSBType(String.format("%d,%d,%d", 0, 10, 15));
int[] i = ColorUtil.hsbToRgbw(hsb);
```

### 5. hsbToRgbw() produces wrong results

For certain HSB values the hsbToRgbw() method produces wrong RGBW results. Test case as follows.

```
HSBType hsb = new HSBType(String.format("%d,%d,%d", 0, 0, 35));
int[] i = ColorUtil.hsbToRgbw(hsb);
```

[...]

Issue #4102, opened 2024-02-19
ColorUtils.hsbToXy() has errors when B part is small 


andrewfg:
### Issue

The ColorUtils.hsbToxy() method seems to have very high errors when the HSB B part is very small. This is probably due to rounding errors. The reason is that hsbToXy internally uses hsbToRgb() and that latter method scales its output depending on the B value.

However since the output of hsbToXy() is simply a point on the CIE color space, in reality only the HS part of the HSB is necessary to determine that point in xy space. 

### Proposed solution

The ColorUtils.hsbToxy() method should normalise the B part to 100% so that the HS part calculates the respective XY point in the CIE color space with maximum precision. 

This means that hsbToXy will consistently produce the same xy color space point for the full 0..100% range of the B  part.

### To Do

Also check if the counterpart ColorUtils.xyToHsb method suffers the same problem..

### Example

Below is a listing of the output of hsbToXy() for various B values between 0 and 100. My expectation is that the `x` and `y` values should NOT vary depending on the B value..

```
HSB[39.000, 40.000, 0.000] => xyY[0.000, 0.000, 0.000]
HSB[39.000, 40.000, 0.100] => xyY[36.720, 36.430, 0.000]
HSB[39.000, 40.000, 0.200] => xyY[36.720, 36.430, 0.010]
HSB[39.000, 40.000, 0.300] => xyY[36.720, 36.430, 0.020]
HSB[39.000, 40.000, 0.400] => xyY[36.720, 36.430, 0.020]
HSB[39.000, 40.000, 0.500] => xyY[36.720, 36.430, 0.030]
HSB[39.000, 40.000, 0.600] => xyY[36.720, 36.430, 0.040]
HSB[39.000, 40.000, 0.700] => xyY[36.720, 36.430, 0.040]
HSB[39.000, 40.000, 0.800] => xyY[36.720, 36.430, 0.050]
HSB[39.000, 40.000, 0.900] => xyY[36.720, 36.430, 0.060]
HSB[39.000, 40.000, 1.000] => xyY[36.720, 36.430, 0.060]
HSB[39.000, 40.000, 1.100] => xyY[36.720, 36.430, 0.070]
HSB[39.000, 40.000, 1.200] => xyY[36.720, 36.430, 0.080]
HSB[39.000, 40.000, 1.300] => xyY[36.720, 36.430, 0.080]
HSB[39.000, 40.000, 1.400] => xyY[36.720, 36.430, 0.090]
[...]

PR #4124, opened 2024-03-02
ColorUtil bug fixes and improvements

andrewfg:
Solves #4114
Solves #4102

This PR addresses the issue mentioned in #4114 .. in particular it does the following..

1. Resolves the seven specific bullet point issues mentioned in the issue.
2. It resolves at core level an issue that is solved at binding level in [here](https://github.com/openhab/openhab-addons/pull/16436).
3. Provides additional JUnit tests to specifically confirm that the seven specific bullet point issues are resolved. (The existing JUnit test methods were not specific enough to have discovered the seven issues).

Note: I was originally thinking to rewrite ColorUtil based on 'java.awt.color' -- however the OH core requires the 'wide D65 color space' whereas 'java.awt.color' only supports a restricted 'sRGB color space' so it lacks the precision needed by OH applications.

---

**NOTA BENE**: merging this PR will cause two of the JUnit tests of `org.openhab.buinding.hue` to FAIL since the tests in that binding written by @jlaur make a wrong assumption about the return XY coordinates from a ColorUtil method. => Therefore requires https://github.com/openhab/openhab-addons/pull/16559


andrewfg:
@holgerfriedrich @genesis81 I am working on this PR to fix the issues mentioned here #4114. I discovered that the current RGB / RGBW conversion throws errors, so I tried to fix them too. In my opinion the existing RGBW <=> RGB algorithm is faulty (and anyway it looks like the code is not really doing what you might expect it to, due to BigDecimal rounding issues). So I wonder if you can look at that part of the code in this PR, and please give me your feedback?

genesis81:
> @holgerfriedrich @genesis81 I am working on this PR to fix the issues mentioned here #4114. I discovered that the current RGB / RGBW conversion throws errors, so I tried to fix them too. In my opinion the existing RGBW <=> RGB algorithm is faulty (and anyway it looks like the code is not really doing what you might expect it to, due to BigDecimal rounding issues). So I wonder if you can look at that part of the code in this PR, and please give me your feedback?

Hi @andrewfg 
i have a look on your code, and looks good for me, but i not tested at the moment. 
The problem with the conversion from HSB to RGBW is that we have not seperate white in HSB. We lose some information when we convert the 4 values from KNX to HSB and also backward. 
I think you conversion is mutch better then the old one. 
Should i check on my hardware?

Greets Marco

andrewfg:
> We lose some information when we convert the 4 values from KNX to HSB and also backward.

@genesis81 many thanks for your comment. It made me think deeper about the conversion algorithm. And now I realise that my code is too naive (it does not properly encode if HSB brightness > 50%). => So I will look at this further.

andrewfg:
> I will look at this further.

@genesis81 Hmm. I just made a few fine improvements to my prior code. And I think this will work better than the original RGBW code. (Perhaps you can test it?)

The issue that I am struggling with -- but cannot resolve -- is the following:

- An RGBW LED strip comprises 4 LED chains. For the sake of argument let us say that each chain can take 100mA.
- A maximum WHITE RGB input converted to RGBW could either drive 3 x 100mA to the RGB LEDs (case a) or 1 x 100mA to the W LED (case b).
- Assuming that the RGBW LED are engineered in a logical manner, (at least I would design them so), case (a) (3 x 100mA on RGB) must produce **_the exact same same luminous output_** as case (b) (1 x 100mA on W).
- And assuming such logical engineering design, the current RGB <=> RGBW conversion algorithms do indeed drive the respective RGBW LEDs in a proper and balanced manner across all possible RGB input values.
- And the current RGB <=> RGBW algorithms are fully linear and reversible. Both the original ColorUtil code and my revision thereof.

That is all good so far .. HOWEVER ..

- In theory it is fully possible to drive an RGBW LED strip with 3 x 100mA on RGB plus 1 x 100mA on W. Let us call this the 'over-drive' case.
- Such 'over-drive' cases cannot produce pure saturated colors (e.g. RED = 255, 0, 0, [0]) .. but they can produce brighter but NON-saturated colors (e.g. 'red' = 255, 0, 0, [255]). Or to put it in other words: over-driving the strip has an inherent trade-off between a gain in brightness and a loss in color saturation. And mathematically speaking, such a trade-off is a "one way function" -- **_an inherent loss of information_**.
- Ideally it would be nice-to-have an `RGB => RGBW` transform that could over-drive the LED strip. And this is indeed mathematically possible.
- However, in the trade-off cases, it is mathematically impossible to devise an `RGBW =>RGB` transform that can survive the round trip `RGB => RGBW => RGB` .. or in other words `RGB => RGBW !=> 'rgb'`

The conclusion is that within the current OH HSB philosophy it is not possible to over-drive such LED strips. This is simply because RGB has three dimensions whereas a reversible over-drive concept requires four. The current version of ColorUtil does not do over-drive, and (therefore) neither does my revised code .. (so at least I have not made anything worse).

Proposed way forward..

1. For the time being I suggest to keep the current OH approach (no over-drive).
2. One idea is that the CIE xyY color co-ordinate system MIGHT have sufficient information to provide the extra dimension. So instead of the round-trip `HSB => RGB => RGBW !=> 'rgb' => 'hsb'` it **_might_** be possible to devise a round trip `HSB => xyY => RGBW => xyY => HSB`. The work is not trivial; it would imply the lamps supporting multiple color gamuts depending on brightness (a fixed wide gamut when not over-driven merging into a variable gamut that narrows depending on the degree of over-drive). But, it is not yet proven if I could actually do this..
3. Another idea is that existing bindings could implement an extra channel e.g. `WhiteOverDrivePercent` which would provide the extra dimension.

=> WDYT?

holgerfriedrich:
@andrewfg thanks for looking into this! Unfortunately, conversions between different color spaces are always tricky and tend to lack handling of some special cases like the one you discovered.

I won't be much help on this, as I am quite busy with other topics. Just my two cents:

I think the approach 1) is fine. You cannot keep the overall brightness constant in the "over-drive" model when the color changes. This seems wrong for the standard conversion. 3) might be already possible, e.g. with KNX and my MDT dimmers you can configure W as a separate channel - thought it depends on the dimmer and the LED stripes in use.

Adding tests - both checking the conversion over wide ranges and checking specific known values from literature - is helpful. It seems we did not cover enough of the special cases, yet. I just checked the test coverage with jacoco tool from mybmw binding, looks good, though it does not tell you if we missed special cases. I'm still concerned about numerical stability for very small values. And finall, looking tat the code: ```if (hue.compareTo(BIG_DECIMAL_360) > 0``` in line 412 might require a ```>=```, as HSBTyoe cannot be constructed with hue of 360.

andrewfg:
@holgerfriedrich many thanks for the feedback. I will address your code suggestion in a next commit.

andrewfg:
> @holgerfriedrich many thanks for the feedback. I will address your code suggestion in a next commit.

Just to confirm that this has been done.

J-N-K:
Please ping when ready for review. Thanks.

andrewfg:
> Please ping when ready for review. Thanks.

@J-N-K just to confirm that this is indeed ready.

andrewfg:
@openhab/core-maintainers can we please try to move this forwards?

As mentioned in https://github.com/openhab/openhab-core/issues/4114 this PR solves seven open issues resp. bugs. It also extends the coverage of the JUnit tests so that other potential future issues become less likely. 

I have tested it live on the Hue binding, and it works perfectly. It does require a tweak to the Hue JUnit tests https://github.com/openhab/openhab-addons/pull/16559 so that when this PR is merged, the build will again pass its JUnit tests taking into account the improved accuracy of color conversions.

There are ca. 15 other bindings that depend on this module. From examining their code I think this PR will not break those bindings. Instead it will improve the accuracy and stability of their color conversions. I cannot do live tests those addons. But I can confirm that they do all build properly and pass their respective JUnit tests as-is. However notwithstanding this, I think it would probably be super prudent, just to be abundantly safe, if we could merge this PR as soon as possible into an early OH SNAPSHOT -- not immediately before a MILESTONE release -- so that others have a chance run it with those other 15 addons and fully confirm that this PR does not introduce regressions.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -421,82 +437,110 @@ public static HSBType xyToHsb(double[] xy) throws IllegalArgumentException {
...
-     * @return the processed value
+     * Get an array of int from an array of PercentType.
      */
-    private static double inverseCompand(double value) {
-        return value > 0.04045 ? Math.pow((value + 0.055) / (1.0 + 0.055), 2.4) : value / 12.92;
+    private static int[] getIntArray(PercentType[] percents) {
+        int[] ints = new int[percents.length];
+        for (int i = 0; i < percents.length; i++) {
+            ints[i] = percents[i].toBigDecimal().multiply(BIG_DECIMAL_255)
+                    .divide(BIG_DECIMAL_100, 0, RoundingMode.HALF_UP).intValue();
```
J-N-K, line 535: Use a constant `BIG_DECIMAL_2_POINT_55` here, that save one operation per conversion.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -421,82 +437,110 @@ public static HSBType xyToHsb(double[] xy) throws IllegalArgumentException {
...
+        }
+        return ints;
     }
 
     /**
-     * Inverse Gamma correction (sRGB companding)
-     *
-     * @param value the value to process
-     * @return the processed value
+     * Private class for points in the CIE xy color space
```
J-N-K, line 541: It's no longer a `private` class if you change the modifier to `public`

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -349,17 +339,43 @@ private static HSBType rgbwToHsb(PercentType[] rgbw) {
...
         }
 
-        BigDecimal r = rgb[0].toBigDecimal();
-        BigDecimal g = rgb[1].toBigDecimal();
-        BigDecimal b = rgb[2].toBigDecimal();
+        BigDecimal r;
+        BigDecimal g;
+        BigDecimal b;
+
+        if (rgbw.length < 4) {
```
J-N-K, line 351: If you compare `== 3` it's a bit more clear when this branch is selected (length == 0, 1, 2 are not possible due to the check before)

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -250,29 +245,60 @@ public static double[] hsbToXY(HSBType hsb) {
...
      */
     public static double[] hsbToXY(HSBType hsb, Gamut gamut) {
-        PercentType[] rgb = hsbToRgbPercent(hsb);
-        double r = inverseCompand(rgb[0].doubleValue() / PercentType.HUNDRED.doubleValue());
-        double g = inverseCompand(rgb[1].doubleValue() / PercentType.HUNDRED.doubleValue());
-        double b = inverseCompand(rgb[2].doubleValue() / PercentType.HUNDRED.doubleValue());
+        PercentType[] rgbPercents = hsbToRgbPercent(hsb);
+
+        // convert rgbPercents to doubles
+        double[] rgb = new double[rgbPercents.length];
```
J-N-K, line 255: Why do you use an array here? I think that explicitely using `r`, `g`, `b` and `w` (like it was before) is much clearer.

bundles/org.openhab.core/src/test/java/org/openhab/core/util/ColorUtilTest.java
```diff
@@ -313,4 +384,241 @@ private void assertRgbEquals(final int[] expected, final int[] actual) {
             assertEquals(expectedS, actualS);
         }
     }
+
+    /**
+     * Test round trips HSB => xyY => HSB
+     */
+    @Test
```
J-N-K, line 391: These tests should be `@ParametrizedTest`s. You can create the arguments from a method, see https://www.baeldung.com/parameterized-tests-junit-5#6-method

bundles/org.openhab.core/src/test/java/org/openhab/core/util/ColorUtilTest.java
```diff
@@ -313,4 +384,241 @@ private void assertRgbEquals(final int[] expected, final int[] actual) {
...
+
+                        // assert that H values are within 0.05 degrees
+                        double h1 = hsb1.getHue().doubleValue();
+                        h1 = h1 >= 180.0 ? 360.0 - h1 : h1;
+                        double h2 = hsb2.getHue().doubleValue();
+                        h2 = h2 >= 180.0 ? 360.0 - h2 : h2;
+                        assertEquals(h1, h2, 0.05);
+
+                        assertionExecuted = true;
+                    } catch (AssertionError e) {
```
J-N-K, line 429: Why do you catch it here instead of failing immediately?

bundles/org.openhab.core/src/test/java/org/openhab/core/util/ColorUtilTest.java
```diff
@@ -313,4 +384,241 @@ private void assertRgbEquals(final int[] expected, final int[] actual) {
...
+                }
+            }
+        }
+        assertTrue(assertionExecuted);
+    }
+
+    /**
+     * Test round trips xyY => HSB => xyY
+     */
+    @Test
```
J-N-K, line 446: see abobe

bundles/org.openhab.core/src/test/java/org/openhab/core/util/ColorUtilTest.java
```diff
@@ -313,4 +384,241 @@ private void assertRgbEquals(final int[] expected, final int[] actual) {
...
+                }
+            }
+        }
+        assertTrue(assertionExecuted);
+    }
+
+    /**
+     * Test round trips HSB => RGB => HSB
+     */
+    @Test
```
J-N-K, line 487: see above

J-N-K (changes requested):
Otherwise LGTM.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
andrewfg: Done.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
andrewfg: Ok. Done.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
andrewfg: Ok. Done.

bundles/org.openhab.core/src/test/java/org/openhab/core/util/ColorUtilTest.java
andrewfg: Ok. I can change this. My gut suggests there must be a better way of creating the argument stream rather than using nested for loops (as follows). But I cannot think of it. => WDYT?

```
    static class ArgsProviderRGBW implements ArgumentsProvider {
        @Override
        public Stream<? extends Arguments> provideArguments(@Nullable ExtensionContext context) throws Exception {
            List<Arguments> result = new ArrayList<>();
            final double step = 5.0;
            for (double r = 0; r <= 100; r = r + step) {
                for (double g = 0; g <= 100; g = g + step) {
                    for (double b = 0; b <= 100; b = b + step) {
                        for (double w = 0; w <= 100; w = w + step) {
                            result.add(Arguments.of(new double[] { r, g, b, w }));
                        }
                    }
                }
            }
            return result.stream();
        }
    }
```

bundles/org.openhab.core/src/test/java/org/openhab/core/util/ColorUtilTest.java
andrewfg: The purpose is for the error log to display a nicely formatted message with the values that failed. It makes it much easier to figure out which particular arguments version of the case actually caused the error.

```
String.format("HSB1:[%.6f,%.6f,%.6f] - xyY:[%.6f,%.6f,%.6f] - HSB2:[%.6f,%.6f,%.6f] - %s",
```

bundles/org.openhab.core/src/test/java/org/openhab/core/util/ColorUtilTest.java
andrewfg: Ok.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
andrewfg: Ok.

andrewfg:
@J-N-K I made the changes that you recommended, but the proposed change in `multiply(BIG_DECIMAL_255).divide(BIG_DECIMAL_100, 0, RoundingMode.HALF_UP).intValue();` causes the tests to fail -- probably due to rounding half up vs. otherwise. => I need to think about this some more.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
andrewfg: @J-N-K I am reverting this change as it breaks the JUnit tests. Not only do the tests fail, but the actual conversions are self evidently wrong too.

holgerfriedrich (approved):
@andrewfg also looks good to me.
Tests coverage is very good,
image
Gaps are basically only the trace outputs and failures.
```public static int hsbTosRgb(HSBType hsb)``` is untested.

Overall, I support merging this PR. 👍

@J-N-K do you want to comment to the open points above?

J-N-K (approved):
Thanks!

PR #4125, opened 2024-03-03
Compare reference using equals

holgerfriedrich:
Fixes error reported by new SAT checker

PMD 7.0.0-rc4 marks this as an error which breaks the build.

ref: openhab/static-code-analysis#462


holgerfriedrich:
@wborn could you pls. double check?

holgerfriedrich:
included in #4161

PR #4126, opened 2024-03-03
Relax test to accept files with CRLF line endings

holgerfriedrich:
I ran into trouble with ```YamlModelRepositoryImplTest``` several times in an Windows environment (where the sample files are checked out with CRLF endings).
I think line endings should not matter for this test, so I replaces the CRLF with LF.

Not sure if one should also care about may (CR).


kaikreuzer (approved):
Thanks!

PR #4129, opened 2024-03-04
[auth] Set doctype for auth page

florian-h05:
Fixes https://github.com/openhab/openhab-webui/issues/2443.

See https://developer.mozilla.org/en-US/docs/Web/HTML/Quirks_Mode_and_Standards_Mode.

### DIFF
--- a/bundles/org.openhab.core.io.http.auth/pages/authorize.html
+++ b/bundles/org.openhab.core.io.http.auth/pages/authorize.html
@@ -1,4 +1,4 @@
-<html>
+<!DOCTYPE html>
 <head>
 <title>openHAB</title>
 <meta charset="utf-8">


bundles/org.openhab.core.io.http.auth/pages/authorize.html
```diff
@@ -1,4 +1,4 @@
-<html>
+<!DOCTYPE html>
 <head>
 <title>openHAB</title>
 <meta charset="utf-8">
```
dilyanpalauzov, line 4: How about removing the line `<meta charset="utf-8">`?

florian-h05: No, I don‘t see a reason to do so. 
There is no harm in providing this information inside the HTML in addition to the header providing it.

dilyanpalauzov: As mentioned at https://github.com/openhab/openhab-webui/issues/2443 the webpage is sent with HTTP header
> Content-Type: text/html;charset=utf-8

I do not think that an additional `<meta charset="utf-8">` in the webpage helps anyhow. So it is unneeded code.

florian-h05: And I have read that this head is is sent.
As I said above: There is no harm to have the meta tag with the charset, and it is common practice to do so.

I would not consider it „unneeded code“, it is part of the HTML standard stuff you put on a HTML page.

Therefore I refuse to remove the charset meta tag, and there will be no more discussion about this.

J-N-K (approved):

PR #4137, opened 2024-03-07
REST response default persistence strategy when no persistence configuration

mherwege:
The persistence REST response, when no persistence configuration is available, is a 404 at the moment.

This PR changes that and will serve a persistence configuration matching the default persistence strategy for the persistence service if one is available.

This allows the UI to show this default strategy if none is configured and the user to change this default in the UI.

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/persistence/PersistenceResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/persistence/PersistenceResource.java
@@ -54,11 +54,13 @@ import org.openhab.core.persistence.FilterCriteria;
 import org.openhab.core.persistence.FilterCriteria.Ordering;
 import org.openhab.core.persistence.HistoricItem;
 import org.openhab.core.persistence.ModifiablePersistenceService;
+import org.openhab.core.persistence.PersistenceItemConfiguration;
 import org.openhab.core.persistence.PersistenceItemInfo;
 import org.openhab.core.persistence.PersistenceManager;
 import org.openhab.core.persistence.PersistenceService;
 import org.openhab.core.persistence.PersistenceServiceRegistry;
 import org.openhab.core.persistence.QueryablePersistenceService;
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/persistence/PersistenceResource.java
```diff
@@ -170,11 +172,24 @@ public Response httpGetPersistenceServices(@Context HttpHeaders headers,
     public Response httpGetPersistenceServiceConfiguration(@Context HttpHeaders headers,
             @Parameter(description = "Id of the persistence service.") @PathParam("serviceId") String serviceId) {
         PersistenceServiceConfiguration configuration = persistenceServiceConfigurationRegistry.get(serviceId);
+        boolean editable = managedPersistenceServiceConfigurationProvider.get(serviceId) != null;
+
+        if (configuration == null) {
+            PersistenceService service = persistenceServiceRegistry.get(serviceId);
+            if (service != null) {
+                List<PersistenceStrategy> defaults = service.getDefaultStrategies();
+                List<PersistenceStrategy> strategies = defaults.stream()
```
J-N-K, line 181: Why do you filter here? The `PersistenceManagerImpl` is applying a different configuration: 

https://github.com/openhab/openhab-core/blob/977fa0329a7d5473439845db3061e260328a6696/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/internal/PersistenceManagerImpl.java#L451-L457

mherwege: My proposal was not adding the full (default configuration) to the REST response, only the default strategies and adding the list of cron strategies to the list of strategies (otherwise a cron default strategy would not work).

It is indeed properly better to respond with the full default configuration and further improve the UI (the UI currently does not allow selecting all items in the configuration, it will show it though, and it can be removed).

J-N-K (approved):
LGTM, thanks.

PR #4141, opened 2024-03-10
[audio] Update bitrate calc

GiviMAD:
<s>I think bitrate was miscalculated after #4116 </s>

nBitsPerSecond = nSamplesPerChannelPerSecond * nBitsPerSample * nChannels;

PR seems to make no difference but seems a little clear to me. Feel free to close it if irrelevant.

### DIFF
--- a/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/utils/AudioWaveUtils.java
+++ b/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/utils/AudioWaveUtils.java
@@ -64,7 +64,7 @@ public class AudioWaveUtils {
             } else {
                 codecPCMSignedOrUnsigned = null;
             }
-            Integer bitRate = Math.round(format.getFrameRate() * format.getSampleSizeInBits() * format.getChannels());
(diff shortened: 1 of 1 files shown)


GiviMAD:
Hello @lolodomo could you test if these solves #4140?

wborn (approved):

lolodomo:
I will try and let you know, thank you for quick investigation.

lolodomo:
I just tested but unfortunately it does not solve the issue.

GiviMAD:
It seems like methods getSampleRate and getFrameRate return same value for formats without compression so the PR makes no effect.

J-N-K:
@GiviMAD Are you working on an improvement?

GiviMAD:
Nop, I just saw it while reviewing a recent PR and thought at the beginning that it could be wrong, but seems to be right for the case. 
I think the API is intended to either do ```format.getFrameRate() * format.getFrameSize()``` to get the compressed byte rate or ```format.getSampleRate() * format.getSampleSizeInBits()``` to get the uncompressed one. But because of the wav format support not compression ```format.getFrameRate()``` and ```format.getSampleRate()``` will return same value.

J-N-K (approved):

PR #4148, opened 2024-03-13
[automation] Allow empty script for script action & script condition

florian-h05:
Required to fix https://github.com/openhab/openhab-webui/issues/2461.

Currently when adding a new script action or script condition to a rule in the UI, the first thing the user sees is a red error message, because the script is not defined yet.
As this is confusing and might worry especially new users, this issue should be fixed.

My proposal is to add support for empty script to script action and script condition and update the UI to create the script field as empty when creating a new script action or condition.

Script action and condition both short-evaluate to a value they don't affect the rule's execution when the script is empty.


kaikreuzer (approved):
lgtm and makes sense, thanks!

PR #4149, opened 2024-03-14
Improve description of hysteresis bounds

ArXa1L:
Hello,
I found that hysteresis value mapping works differently than stated in its documentation:

[source:](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/profiles/SystemHysteresisStateProfile.java#L163)

```java
    private Type mapValue(double lower, double upper, double value) {
        if (value <= lower) {
            return low;
        } else if (value >= upper) {
            return high;
        }
        return previousType;
    }
```

[doc:](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.thing/src/main/resources/OH-INF/config/hyteresisProfile.xml)

```xml
<parameter name="lower" type="text" required="true">
  <label>Lower Bound</label>
  <description>Maps to OFF if value is below lower bound (plain number or number with unit).</description>
</parameter>
<parameter name="upper" type="text">
  <label>Upper Bound</label>
  <description>Maps to ON if value is above upper bound (plain number or number with unit).</description>
</parameter>
```

I thought that changing the value mapping would be more correct and opened PR #4120, but during the discussion we decided that it would be preferable to improve the documentation


seime (commented):
In `SystemProfiles_XX.properties`, these texts are translated to 15-20 languages. These are not adjusted accordingly, but it might be hard for the contributor to do so.

bundles/org.openhab.core.thing/src/main/resources/OH-INF/i18n/SystemProfiles_ru.properties
```diff
@@ -5,9 +5,9 @@ profile.config.system.offset.offset.label = Смещение
 profile.config.system.offset.offset.description = Смещение (просто число или с ед. измерения), применяемое к состоянию по отношению к элементу. Отрицательное смещение будет применено в обратном направлении.
 profile-type.system.hysteresis.label = Гистерезис
 profile.config.system.hysteresis.lower.label = Нижняя граница
-profile.config.system.hysteresis.lower.description = Приводит к состоянию OFF, если значение меньше нижней границы (просто число или с ед. измерения).
+profile.config.system.hysteresis.lower.description = Приводит к состоянию OFF, если значение меньше или равно нижней границе (просто число или с ед. измерения).
```
J-N-K, line 8: Please remove these translations (only keep the XML changes and the English) and add it later using crowdin. Otherwise it'll not show up there.

J-N-K (changes requested):

J-N-K (approved):
LGTM, thanks!

PR #4150, opened 2024-03-14
[oauth] Fix exception wrapping

jlaur:
In this method, exceptions are thrown, caught and wrapped into `OAuthException`. In some cases, this results in `OAuthException` being wrapped into another `OAuthException`. However, in other cases, `OAuthResponseException` is wrapped into `OAuthException`, effectively preventing the intended [OAuthResponseException](https://www.openhab.org/javadoc/latest/org/openhab/core/auth/client/oauth2/oauthresponseexception) from being thrown  from the method (since 4.0).

See https://github.com/openhab/openhab-addons/issues/16518#issuecomment-1998324602

Regression of #3271


J-N-K (approved):
LGTM

Issue #4155, opened 2024-03-17
[rfc] UoM shall I create a new unit(s) for humidity??


andrewfg:
Many OH bindings relate to interior HVAC and/or outside climate and thus have humidity channels. Humidity is measured in several ways..

1. current amount of water in the air as a proportion of the maximum amount of water that the air can hold (at the current temperature) -- it is called relative humidity -- and formally displayed as 57 % r.H.
2. grams of water per kilogram of dry air it is called absolute humidity -- and formally displayed as 57 g/Kg
3. grams of water per cubic meter of dry air it is also called absolute humidity -- and formally displayed as 57 g/m3

Measurement methods 1. and 2. are both dimensionless measures, whereas 3. is a density measure.

Method 1. **_can_** be represented in OH via the existing UoM unit `Units.PERCENT` but this is **_not strictly correct_** since the formal display "57 %" differs from the formal display "57 % r.H."

So my questions (resp. proposals) are whether I shall create two new units for OH as follows..
- A new `Units.RELATIVE_HUMIDITY` which is a subclass of `Units.PERCENT` where the formal display is overridden as "57 % r.H."
- A new `Units.ABSOLUTE_HUMIDITY` which is a subclass of `Units.ONE` with a scale of 1/1000 and a formal display of "57 g/Kg"

Note that Unit conversion between Units.RELATIVE_HUMIDITY and Units.ABSOLUTE_HUMIDITY is not strictly 100% possible. This is because air has a higher capacity to hold water when it is warmer. (Which BTW explains why you get condensation and eventually frost when air is chilled). However there is a 'quasi/normalized' Unit conversion possible between Units.RELATIVE_HUMIDITY and Units.ABSOLUTE_HUMIDITY which can be made based on air at a normalized temperature of 20 Celsius.

=> WDYT?

rkoshak:
I think there was a discussion recently that covered this idea and there were arguments against having two different dimensions that represent the same thing. Whether it's percent humidity or percent of something else, it's still a percent. 

You can add `"r.H"` to the state description when/if you want to be that specific in how the state is displayed. But in all other respects, it behaves as a percent so I'm not entirely certain making the distinction add anything. 

And IIRC a unit for absolute humidity has been added sometime in the past year or so already but I can't find that PR at the moment for some reason. 

[...]

PR #4157, opened 2024-03-20
UoM add units for humidity

andrewfg:
Resolves https://github.com/openhab/openhab-core/issues/4155

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/Units.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/Units.java
@@ -103,6 +103,8 @@ public final class Units extends CustomUnits {
             tech.units.indriya.unit.Units.KILOGRAM.divide(tech.units.indriya.unit.Units.CUBIC_METRE)));
     public static final Unit<Density> MICROGRAM_PER_CUBICMETRE = addUnit(new TransformedUnit<>(KILOGRAM_PER_CUBICMETRE,
             MultiplyConverter.ofRational(BigInteger.ONE, BigInteger.valueOf(1000000000))));
+    public static final Unit<Density> HUMIDITY_ABSOLUTE_DENSITY = addUnit(
+            new ProductUnit<>(tech.units.indriya.unit.Units.GRAM.divide(tech.units.indriya.unit.Units.CUBIC_METRE)));
     public static final Unit<Dimensionless> ONE = addUnit(AbstractUnit.ONE);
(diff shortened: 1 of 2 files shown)


mherwege:
I think the g/kg unit makes sense. I am more reluctant about % r.H. To me this is still %, nothing else.
If you want this to mean relative humidity, it should be called that way as channel name and item name, not by tweaking the unit.
another channel would be called absolute humidity and have a unit hint of g/kg. (I hope that PR gets merged eventually)
I would not have the UOM system convert between them, as it is only an approximative conversion. If you want such a conversion, it could be done in the binding, between the channels, with the clear disclaimer that it is only valid at 20°C. Another binding may be linked to a device that gives a more accurate value for both absolute and relative humidity, and keeping it in two distinct channels therefore feels more natural to me.

andrewfg:
> I think the g/kg unit makes sense. I am more reluctant about % r.H.

Honestly if we choose not to do relative humidity then absolute humidity "by mass ratio" should probably also be cancelled. There are in fact three humidity measures -- namely relative humidity (%rH), absolute humidity "by mass ratio" (g/Kg), and absolute humidity "by volume" (g/m³). You don't like the former. The latter is already a standard unit of density. So if we drop those two, we may as well drop them all.

> I would not have the UOM system convert between them, as it is only an approximative conversion.

Yeah. I only really did that because I could not find a way to PREVENT conversion (i.e. make it throw an exception). Given that I could not prevent conversion I thought it best to at least return something that more-or-less correct -- rather than absolutely wrong..

mherwege:
@andrewfg I don’t agree with this.
First, relative humidity is a fraction, typically expressed in %. Whether you put .r.H. behind it does not make a difference. It is a percentage. It would be nice to be able to put it behind the text, but it does not make a difference in meaning.
Second, converting between incompatible measurements, even in the same unit, is an error. For absolute humidity, while g/kg is a customary (dimensionless) unit, expressing it as kg/kg or % is valid as well. 1% of absolute humidity would be would be 10 g/kg. I agree adding g/kg as a customary unit because it otherwise is not supported for Number:Dimensionless, and it is a valid dimensionless unit.
If you think about it, also dB is a dimensionless unit. Everyone would agree it does not make sense to convert certain dimensionless measurements to dB, just like you would not convert number of apples to number of pears because it is the same dimensionless unit of ONE.

So creating an almost arbitrary conversion between incompatible measurements is a total no go to me.

see also:
https://en.wikipedia.org/wiki/Humidity?wprov=sfti1

andrewfg:
Closing this.

PR #4160, opened 2024-03-25
[rest] TokenResource: Set SameSite attribute for session id cookie

### LINKED ISSUE #4159: Wrong usage of `SameSite` cookie param in `TokenResource`
When logging in to Main UI , Firefox is printing this warning to the log:

```
Cookie “X-OPENHAB-SESSIONID” does not have a proper “SameSite” attribute value.
Soon, cookies without the “SameSite” attribute or with an invalid value will be treated as “Lax”.
This means that the cookie will no longer be sent in third-party contexts.
If your application depends on this cookie being available in such contexts, please add the “SameSite=None“ attribute to it.
To know more about the “SameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite
```

Also see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#samesitesamesite-value.

I would like to set `SameSite` to `strict`, which is supported by the `javax.ws.rs.core.NewCookie` (see https://github.com/jakartaee/rest/pull/868) since JakartaEE 3.1, however is the change from the linked PR not available.

How do I found out the version of Jakarta EE in use by core?

@J-N-K May I ask for your help?

florian-h05:
Fixes #4159.

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#samesitesamesite-value.

Because our current implementation of `javax.ws.rs.core.NewCookie` (which is `jakarta.ws.rs/jakarta.ws.rs-api/2.1.6` provided by `org.apache.aries.spec/org.apache.aries.javax.jax.rs-api/1.0.4`) does not allow to set the SameSite attribute, we have to workaround that limitation by setting the cookie directly through a response header. 
Once our implementation is `jakarta.ws.rs/jakarta.ws.rs-api/3.1.0` or newer, we can use `NewCookie` again.
Implementing the required changes to NewCookie on our own does not work, because the servlet also needs adustment.

### DIFF
--- a/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/TokenResource.java
+++ b/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/TokenResource.java
@@ -246,8 +246,7 @@ public class TokenResource implements RESTResource {
             try {
                 URI domainUri = new URI(session.get().getRedirectUri());
                 // workaround to set the SameSite cookie attribute until we upgrade to
-                // jakarta.ws.rs/jakarta.ws.rs-api/3.1.0 or newer, which is provided by
-                // org.apache.aries.spec/org.apache.aries.javax.jax.rs-api
+                // jakarta.ws.rs/jakarta.ws.rs-api/3.1.0 or newer
                 response.header("Set-Cookie",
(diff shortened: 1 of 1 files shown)


florian-h05:
Done 👍

J-N-K (approved):
Thanks!

PR #4161, opened 2024-03-25
Support Java 21

holgerfriedrich:
openhab-core can now be compiled an run both on Java 17 and Java 21.

- Upgrade SAT to 0.16.0
- Add profiles j17 and j21, automatically activated based on JDK version unless specified.
- Pin Java version for xtend to 17. As xtend does not yet support Java 21, compile model with Java 17.
- Includes #4125.

Refs: openhab/openhab-distro#1590

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,17 +64,14 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-    <oh.java.version>17</oh.java.version>
-    <maven.compiler.source>${oh.java.version}</maven.compiler.source>
-    <maven.compiler.target>${oh.java.version}</maven.compiler.target>
-    <maven.compiler.compilerVersion>${oh.java.version}</maven.compiler.compilerVersion>
+    <maven.compiler.release>17</maven.compiler.release>
 
     <bnd.version>7.0.0</bnd.version>
     <commons.net.version>3.7.2</commons.net.version>
     <eea.version>2.3.0</eea.version>
     <karaf.compile.version>4.4.5</karaf.compile.version>
     <karaf.tooling.version>4.4.5</karaf.tooling.version>
-    <sat.version>0.15.0</sat.version>
+    <sat.version>0.16.0</sat.version>
     <slf4j.version>2.0.7</slf4j.version>
     <xtext.version>2.34.0</xtext.version>
     <spotless.version>2.38.0</spotless.version>
@@ -670,7 +667,7 @@ Import-Package: \\
             <configuration>
               <rules>
                 <requireJavaVersion>
-                  <version>[17.0,18.0)</version>
+                  <version>[17.0,18.0),[21.0,22.0)</version>
                 </requireJavaVersion>
               </rules>
             </configuration>
@@ -844,6 +841,12 @@ Import-Package: \\
           </plugins>
         </pluginManagement>
       </build>
+    </profile>
+    <profile>
+      <id>j21</id>
+      <properties>
+        <maven.compiler.release>21</maven.compiler.release>
+      </properties>
     </profile>
   </profiles>
 </project>
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/scheduler/CronSchedulerImpl.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/scheduler/CronSchedulerImpl.java
@@ -107,7 +107,7 @@ public class CronSchedulerImpl implements CronScheduler {
             for (Iterator<Cron> cron = crons.iterator(); cron.hasNext();) {
                 final Cron c = cron.next();
 
-                if (c.target == s) {
+                if (c.target.equals(s)) {
                     cron.remove();
                     c.schedule.cancel(true);
                 }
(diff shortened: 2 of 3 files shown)


calle2010:
Have you considered to use maven.compiler.release instead of source/target? 
This prevents accidental usage of Java 21 API when compiling for Java 17 on a JDK 21. See also https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#release and https://openjdk.org/jeps/247.

holgerfriedrich:
> Have you considered to use maven.compiler.release instead of source/target? This prevents accidental usage of Java 21 API when compiling for Java 17 on a JDK 21. See also https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#release and https://openjdk.org/jeps/247.

@calle2010 Thanks. This seems reasonable. For core it allows to drop the definition of ```oh.java.version``` as well, as it is not used anywhere else.

pom.xml
```diff
@@ -845,5 +841,23 @@ Import-Package: \\
...
+        <maven.compiler.release>17</maven.compiler.release>
+      </properties>
+    </profile>
+    <profile>
+      <id>j21</id>
+      <activation>
+        <jdk>[21.0,22.0)</jdk>
+      </activation>
+      <properties>
+        <maven.compiler.release>21</maven.compiler.release>
```
wborn, line 859: It's probably better to keep the artifacts compatible with Java 17 regardless of the installed JDK so the artifacts remain compatible with the official release. That way you can use Java 21 on other projects without having to switch JDKs when working on openHAB. It will also help with preventing contributors using newer Java constructs that aren't available on Java 17. They would then only get this feedback when CI fails.

holgerfriedrich: @wborn Thanks for looking into this. Keeping release to 17 basically means that the JDK could be 21, but still behaving like J17. It addresses the point that users have a JDK21 installed. OH modules would then still be J17.

I personally had a different use case in mind, "compiling and running" on J21.
The build anyway results in a mixture of different class file versions. I would try to upgrade my test server to JDK21, thus J21 class files should be fine. I did not have the use case above in mind, this is why I proposed the auto-detection:
https://github.com/openhab/openhab-distro/issues/1590#issuecomment-2019099635
You are right, only CI will fail when users start to use new J21 features in this case. I do not have a solution for this.

Do you think it is a better approach to use release set to 17 unless a profile for J21 is explicitly activated?

The question is more or less about the approach we are taking towards an upcoming OH5.x which might require J21. How do we gain knowledge about potential problems when compiling and running in J21.....

wborn: We can add a GHA PR build for J21 like we did in the past. In such a build it can activate the profile to use J21 release.

OH5 will probably still take some. We just started using J17 with OH4. Maybe an OH5 will happen after J25 is released in 2025. 🔮 😉

The issue with requiring newer Java versions is that users may need to manually upgrade their JDK, OS or even hardware like we saw after requiring J17.

Running OH4 using J21 probably already works very well with all the recent updates.

holgerfriedrich: I changed the POM file, now default it profile ```j17``` for both JDK 17 and JDK 21.
Profile ```j21``` can be activated explicitly to build native Java 21 file.

bundles/org.openhab.core.model.item.ide/pom.xml
```diff
@@ -14,6 +14,11 @@
 
   <name>openHAB Core :: Bundles :: Model Item IDE</name>
 
+  <properties>
+    <maven.compiler.source>17</maven.compiler.source>
+    <maven.compiler.target>17</maven.compiler.target>
+  </properties>
```
wborn, line 20: Does it also work if you use `<maven.compiler.release>17</maven.compiler.release>` instead?

Is it also possible to add the property to the [model-specific-profile](https://github.com/openhab/openhab-core/blob/be9cbf26491f976272c361e73292336146026adb/bundles/pom.xml#L272-L356) instead?

pom.xml
```diff
@@ -845,5 +841,20 @@ Import-Package: \\
...
     </profile>
+    <profile>
+      <id>j17</id>
+      <activation>
+        <jdk>[17.0,18.0);[21.0,22.0)</jdk>
+      </activation>
+      <properties>
+        <maven.compiler.release>17</maven.compiler.release>
+      </properties>
+    </profile>
```
wborn, line 852: Is this profile necessary when `<maven.compiler.release>17</maven.compiler.release>` would be as default value? Maybe acttivating the j21 profile would then simply override it?

bundles/org.openhab.core.model.item.ide/pom.xml
holgerfriedrich: I will try ```<maven.compiler.release>17</maven.compiler.release>```...

```model-specific-profile``` would be way more elegant, though it covers more directories.

> org.openhab.core.model.codegen
> org.openhab.core.model.core
> org.openhab.core.model.item.runtime
> org.openhab.core.model.lsp
> org.openhab.core.model.persistence.runtime
> org.openhab.core.model.rule.runtime
> org.openhab.core.model.script.runtime
> org.openhab.core.model.sitemap.runtime
> org.openhab.core.model.thing.runtime

Will check.

holgerfriedrich:
@wborn Changes are now implemented as proposed. Now the PR is down to a few files. Thanks for directing me towards ```model-specific-profile```.

wborn (approved):
Thanks! :+1:

PR #4164, opened 2024-04-01
[util] Add abbreviate method

lsiepel:
While trying to remove the apache libs from the addons i found one more string method that should be moved to this core utility. 

It is used in the Max binding, i'll update the related PR accordingly: https://github.com/openhab/openhab-addons/pull/14413

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/util/StringUtils.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/util/StringUtils.java
@@ -28,6 +28,25 @@ import org.eclipse.jdt.annotation.Nullable;
 public class StringUtils {
 
     /**
+     * Input string is shortened to the maxwidth, the last 3 chars are replaced by ...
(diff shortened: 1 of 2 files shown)


andrewfg (approved):
LGTM

J-N-K:
Is there any other use of this method? It seems a very rare case and should probably be added to the binding code if it is not used elsewhere.

lsiepel:
> Is there any other use of this method? It seems a very rare case and should probably be added to the binding code if it is not used elsewhere.

I cannot oversee all binding code, but i'm not aware at any other usage at this moment in time. As abbreviating strings is a common scenario i would not classify it as very rare. I expect other bindings might benefit in the future too. Besides it is a small method.

andrewfg:
It's surely a chicken and egg situation. If the utility is  / would be  present. Then it will  / would have been, used.

J-N-K (approved):
LGTM

Issue #4166, opened 2024-04-03
QuantityType<Dimensionless> with unit one wrong update to item with different unit set


mherwege:
An `Number<Dimensionless>` item with a unit set in metadata to something different from the default of `one` will not update correctly in a `postUpdate` with a `QuantityType<Dimensionless>` and unit `one`.

This is because the unit (`one`) gets dropped [here](https://github.com/openhab/openhab-core/blob/dcfadaede6ebf3f50ce049857f028c36ddb29668/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityType.java#L432) . As there is no unit, `postUpdate` will assume it is a `Number` item and update with the configured unit.

To test this I created an Item Test with unit % and pattern %,.2f %%.

I tested with the javascript below (result is similar with DSL). Notice I tried updating the result of the division 4 times, once straight, twice after conversion to a compatible unit (I even used dB), and once where I explicitly appended one to the result. The timeouts are there to wait for the update to happen before logging.
The postUpdate in test 1 is the one producing the wrong result.

Issue was first discussed on the community: https://community.openhab.org/t/inconsistent-percentage-behaviour/154595/20?u=mherwege

```
var logger = log('Test');

var bkw    = Quantity('5 W')
var zwl    = Quantity('5 W')
var result = bkw.divide(bkw.add(zwl))   // I want to preserve the QuantityType for logging

logger.info('Result of calculation is: {}', result.toString())
logger.info('Unit is: {}', result.unit) // I would actually expect this to show `ONE`
logger.info('Dimension is: {}', result.dimension)
logger.info('Result %: {}', result.toUnit('%').toString())

logger.info('Test 1, no unit applied: {}', result.toString())
items.Test.postUpdate(result)
setTimeout(() => {
  logger.info('State: {}', items.Test.state)  
  logger.info('Test 2, % unit applied: {}', result.toUnit('%').toString())
  items.Test.postUpdate(result.toUnit('%'))
  setTimeout(() => {
    logger.info('State: {}', items.Test.state)
    logger.info('Test 3, dB unit applied: {}', result.toUnit('dB').toString())
    items.Test.postUpdate(result.toUnit('dB'))
    setTimeout(() => {
      logger.info('State: {}', items.Test.state)
      logger.info('Test 4, one appended: {}', result.toString() + ' one')
      items.Test.postUpdate(result.toString() + ' one')
      setTimeout(() => {
        logger.info('State: {}', items.Test.state)  
[...]

Issue #4163, opened 2024-03-27
QuantityType<Time> incorrectly formats durations


jpg0:
Currently `QuantityType<Time>` is hardcoded to format the held `Quantity<Time>` as a UTC `ZonedDateTime` when not provided a format string.

## Expected Behavior
Format as a `Duration` when the scale is `Scale.RELATIVE`, and a `ZonedDateTime` when the scale is `Scale.ABSOLUTE`.

## Current Behavior
Always formats to a UTC ZDT. This fails for me when I use `QuantityType<Time>` to represent a duration. For example, I am using it to represent the period of time that my irrigation needs to run, which is then sent to the controller over MQTT. It was failing to send the duration required, but instead sending a formatted ZonedDateTime measured at the duration (in seconds) past the epoch. Currently my outgoing format logic is required to parse the ZDT then determine how many seconds past the epoch it is, convert this back into a `QuantityType<Time>`, then reformat with a format string. Note that other quantities appear to have no such formatting logic hardcoded into the `QuantityType` class.

## Possible Solution
A small solution, localised to my specific problem, would be to allow specification of format strings to the MQTT binding. It seems that this is currently not supported and always passes a null format string. However, this would not fix the underlying problem, that `Quantity<Time>` representing a duration (i.e. with a relative scale) is incorrectly formatted by default. Indeed, looking at `QuantityType` it doesn't really appear to handle the scale parameter at all, and just hardcodes it to `Scale.RELATIVE`.

rkoshak:
I think more details need to be provided because unless you've set a state description pattern on the Item *and* you are sending the displayState I don't see how you would ever get a ZDT.

A `Quantity<Time>` should always represent a duration. By default, unless the `unit` metadata is set on the Item, that will get rendered as `N s` for N seconds. 

In the state description, it's possible to break this number of seconds up into DD:HH:MM:SS (for example) by using date time string formatting. I suspect you are somehow accessing this latter capability. But the default value should be `<number> <unit>` just like every other `QuantityType`. 

But as far as I can tell, it shouldn't even be possible to access this ZDT formatted String on the `QuantityType<Time>` from the MQTT binding. That's what's confusing me. 

To test things out I configured a Number Channel on an MQTT Thing leaving pretty much everything as the default. 

```
UID: mqtt:topic:broker:TimeTest
label: TimeTest
thingTypeUID: mqtt:topic
configuration: {}
bridgeUID: mqtt:broker:broker
channels:
  - id: NumberChan
[...]

PR #4169, opened 2024-04-04
Fix QuantityType dimensionless one and time formatting

mherwege:
Resolves https://github.com/openhab/openhab-core/issues/4166
Resolves https://github.com/openhab/openhab-core/issues/4163

This PR tries to resolve two issues with QuantityType:

1. Updating a Number:Dimensionless Item with a value of unit one, when the unit is set in metadata to another unit, will not update correctly. The proposed solution is to remove the special handling of unit one and make the unit part of the string again. It can be removed from visualisation using a state description.
2. Formatting a QuantityType<Time> tries to use the DateTime String formatting methods. When it fails it may output a UTC date. Durations will also not be correctly formatted when days/hours/minutes/seconds are above the month (31)/day (24)/hour (60)/minute (60) limits. In this PR, the formatting strings are preserved, but interpreted specifically for time with a relative scale (durations), rather than an absolute time.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityType.java
@@ -18,16 +18,22 @@ import java.math.BigDecimal;
 import java.text.DecimalFormat;
 import java.text.NumberFormat;
 import java.text.ParsePosition;
-import java.time.Instant;
-import java.time.ZoneOffset;
-import java.time.ZonedDateTime;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.HashMap;
 import java.util.IllegalFormatConversionException;
+import java.util.List;
 import java.util.Locale;
+import java.util.Map;
+import java.util.MissingFormatArgumentException;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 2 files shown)


andrewfg (approved):
LGTM thankyou.

andrewfg:
@mherwege many thanks. I like the extra bonus for the duration formatting.

kaikreuzer (approved):
Thanks!

kaikreuzer:
We have a few failing tests in the add-ons now, see https://ci.openhab.org/job/openHAB-Addons/lastCompletedBuild/testReport/.

Might these be related to this change, @mherwege?

mherwege:
It probably is. I am investigating. It looks like the mqtt tests assume a formatting of a QuantityType with format pattern %s just drops the unit. It may have done that in the past because in core, first a conversion to BigDecimal (which dropped the unit) was attempted before passing the value verbatim (with unit). I don't think the unit should be dropped like that when formatting a Number with Dimension.

mherwege:
@kaikreuzer I have made a change to the MQTT binding that resolves the issue caused by the change in this PR.

The fundamental problem is that the `format` method on a type is used both for representation purposes in UI's and internal processing. Some logic depends on the formatting to be done in a certain way, which blocks improvements in the UI, and the other way around. I hope this change resolves the issue. At leasts the failing tests work again, so it should keep backward compatibility.

kaikreuzer:
@mherwege Many thanks for taking care of it so quickly!

PR #4171, opened 2024-04-05
[rest] Fix Java 21 build

holgerfriedrich:
Rename internal class.
Old implementation was breaking the Java 21 build when output was set to Java 17 class files.
Compiler got confused by similar class names ```Entry``` from local class and Map.

Refs:
https://github.com/openhab/openhab-distro/issues/1590
https://github.com/openhab/openhab-distro/pull/1640


kaikreuzer (approved):
Thanks!

PR #4177, opened 2024-04-08
Do not create new `BigDecimals` if a given object is already `BigDecimal`

magx2:
Hey, i just wanted to add small optimization to not create a new `BigDecimal` if given object is already this class


wborn (approved):
Thanks!

PR #4178, opened 2024-04-09
Upgrade Jollyday to 0.27.0

wborn:
Upgrades Jollyday from 0.23.2 to 0.27.0.

For release notes, see:

https://github.com/focus-shift/jollyday/releases

### DIFF
--- a/itests/org.openhab.core.model.rule.tests/itest.bndrun
+++ b/itests/org.openhab.core.model.rule.tests/itest.bndrun
@@ -88,8 +88,6 @@ Fragment-Host: org.openhab.core.model.rule.runtime
 	com.fasterxml.jackson.core.jackson-databind;version='[2.16.0,2.16.1)',\
 	com.fasterxml.jackson.dataformat.jackson-dataformat-xml;version='[2.16.0,2.16.1)',\
 	stax2-api;version='[4.2.1,4.2.2)',\
-	de.focus_shift.jollyday-core;version='[0.23.2,0.23.3)',\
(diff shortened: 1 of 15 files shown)


itests/org.openhab.core.model.rule.tests/itest.bndrun
```diff
@@ -126,10 +124,11 @@ Fragment-Host: org.openhab.core.model.rule.runtime
 	org.objectweb.asm.tree;version='[9.6.0,9.6.1)',\
 	org.objectweb.asm.tree.analysis;version='[9.6.0,9.6.1)',\
 	org.objectweb.asm.util;version='[9.6.0,9.6.1)',\
-	org.openhab.core.model.persistence.runtime;version='[4.2.0,4.2.1)',\
```
kaikreuzer, line 129: This removal seems unrelated. Any idea why it is there and if it is ok to be removed?

wborn: I just let the resolver resolve the bundles and it came up with these bundles. The tests still run fine without it. There is probably some implementation detail that makes it resolve different bundles depending on the local environment. Maybe it's due to the OS, Java version or the order of files on disk. To workaround it we could blacklist those bundles that continuously get added/removed by the resolver.

kaikreuzer: Ok, thanks!

kaikreuzer (approved):

PR #4182, opened 2024-04-13
Fix marketplace showing incompatible add-ons by default

### LINKED ISSUE #4180: [marketplace] Supported Bundle Versions No Longer Processed
## Expected Behavior
Marketplace bundles should only be listed under the add-on store based on their supported version range included in the forum title.

## Current Behavior
It seems that starting OH 4.1.2, all marketplace bundles regardless of their supported version are listed in the add-on store. The odd part is that I couldn't find any changes in core to the code related to the supported version range matching. Previous OH versions also seem to be affected as well.

Open add-on store, search for "insteon"

Two beta versions will come up with one of them supporting 3.x and the other one 4.x.

image

## Your Environment
* Version used: OH 4.1.2

J-N-K:
Fixes #4180 

I really wonder why this is showing up now, since the bug was there since quite some time. "Show incompatible" can be configured in the add-on management section and should default to "false", instead it was "true".

Should be backported to 4.1.x


wborn:
Anyone can cherry pick bugfixes once approved. If you merge PRs on a phone cherry-picking is less easy. :wink:

Issue #4174, opened 2024-04-07
Java 21 deserialization of DateTime


holgerfriedrich:
I just gave running openHAB on Java 21 another test but there seems to be some issue with date formats when deserializing timestamps on Java 21.

These errors show up when restarting openHAB:

```
2024-04-06 09:05:10.272 [ERROR] [re.storage.json.internal.JsonStorage] - Couldn't deserialize value 'org.openhab.core.storage.json.internal.StorageEntry@4a2246a9'. Root cause is: Failed parsing 'Apr 6, 2024, 8:53:04 AM' as Date; at path $.sessions[0].createdTime
```

```
09:01:06.164 [ERROR] [rnal.common.AbstractInvocationHandler] - An error occurred while calling method 'QueryablePersistenceService.query()' on 'org.openhab.persistence.mapdb.internal.MapDbPersistenceService@39aca3e6': Failed parsing 'Apr 6, 2024, 8:58:09 AM' as Date; at path $.timestamp
com.google.gson.JsonSyntaxException: Failed parsing 'Apr 6, 2024, 8:58:09 AM' as Date; at path $.timestamp
	at com.google.gson.internal.bind.DateTypeAdapter.deserializeToDate(DateTypeAdapter.java:90) ~[bundleFile:?]
	at com.google.gson.internal.bind.DateTypeAdapter.read(DateTypeAdapter.java:75) ~[bundleFile:?]
	at com.google.gson.internal.bind.DateTypeAdapter.read(DateTypeAdapter.java:46) ~[bundleFile:?]
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.readIntoField(ReflectiveTypeAdapterFactory.java:212) ~[bundleFile:?]
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$FieldReflectionAdapter.readField(ReflectiveTypeAdapterFactory.java:433) ~[bundleFile:?]
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:393) ~[bundleFile:?]
	at com.google.gson.Gson.fromJson(Gson.java:1227) ~[bundleFile:?]
	at com.google.gson.Gson.fromJson(Gson.java:1137) ~[bundleFile:?]
	at com.google.gson.Gson.fromJson(Gson.java:1047) ~[bundleFile:?]
[...]

PR #4185, opened 2024-04-14
GsonBuilder: Explicitly set date format

holgerfriedrich:
Between Java 17 and Java 21, serialization of Date has changed due to CLDR 42 which uses a narrow non-breaking space.
To ease switching JDK versions, the serialization format is explicitly set to the Java 17 format.

Refs: #4174, openhab/openhab-addons#16657

This helps only for the configuration settings etc.
```GsonBuilder()``` is used in many places. Maybe we should use a custom implementation which sets the date format in the ctor.


J-N-K (commented):
This is 100% compatible, correct? Should we add a TODO issue to remove the workaround and rewrite configuration when we start to require Java 21?

holgerfriedrich:
It should not have any impact for Java 17. That's the intention.
Next snapshot after merge will show ;-)

I don't think we should switch to the new date format, even after switching to Java 21. UTF8 characters that look like a space but behave differently are likely to cause problems.

J-N-K (approved):
Thanks!

Issue #3983, opened 2023-12-30
4.2: bundles automatically restored to their original version


lolodomo:
Using 4.2 snapshot (3808), I encounter a strange behaviour: if I update one bundle (bundle:update) then after a short time (max few minutes), my bundle is automatically restored to its original version ! I encountered that behaviour yesterday after updating either o.o.core.io.transport.mdns or o.o.webui.ui.basic bundles.
What could be the trigger of that behaviour ?

Just to explain the conditions: I just installed the snapshot 3808 form scratch in a new directory. No kar file. Remote add-ons is enabled. I update bundles using bundle:update command.

Earlier I played many hours with release 4.1 and applying many times updates of o.o.webui.ui.basic without encountering that behaviour.
It seems something new in 4.2...

wborn:
I also just noticed this.
Enabling Karaf debug logging with `log:set debug org.apache.karaf` shows me:

```
18:30:17.923 [DEBUG] [af.features.internal.service.Deployer] - New snapshot available for mvn:org.openhab.addons.bundles/org.openhab.binding.shelly/4.2.0-SNAPSHOT
18:30:17.940 [INFO ] [.internal.service.FeaturesServiceImpl] - Changes to perform:
18:30:17.940 [INFO ] [.internal.service.FeaturesServiceImpl] -   Region: root
18:30:17.941 [INFO ] [.internal.service.FeaturesServiceImpl] -     Bundles to uninstall:
18:30:17.941 [INFO ] [.internal.service.FeaturesServiceImpl] -       org.openhab.core.config.discovery.addon.mdns/4.2.0.202312301510
18:30:17.941 [INFO ] [.internal.service.FeaturesServiceImpl] -     Bundles to update:
18:30:17.941 [INFO ] [.internal.service.FeaturesServiceImpl] -       org.openhab.binding.shelly/4.2.0.202312301537 with mvn:org.openhab.addons.bundles/org.openhab.binding.shelly/4.2.0-SNAPSHOT
18:30:17.941 [INFO ] [.internal.service.FeaturesServiceImpl] -     Bundles to install:
18:30:17.941 [INFO ] [.internal.service.FeaturesServiceImpl] -       mvn:net.java.dev.jna/jna-platform/5.13.0
18:30:17.941 [INFO ] [.internal.service.FeaturesServiceImpl] -       mvn:com.neuronrobotics/nrjavaserial/5.2.1.OH1
[...]

PR #4188, opened 2024-04-17
Use FeatureInstaller for suggestion finder installation

mherwege:
See https://github.com/openhab/openhab-core/issues/3983

This PR changes the installation/deinstallation of suggestion finders, not directly calling the FeatureService, but through FeatureInstaller. This should reduce the amount of bundle refreshes happening, but will not completely avoid it (the root cause is not with the finders, but the refersh on install/deinstall of any bundle).
At a minimum, it moves the installation/deinstallation into the same code as installing/deinstalling addons, so improvements in that code will also benefit the installation/deinstallation of the addon finders.

@J-N-K @lolodomo let me know if this change is worth doing, or the benefit is too limited.


J-N-K:
LGTM, I would like to hear what @wborn thinks about it.

J-N-K:
Let's give it a try.

J-N-K (approved):

PR #4190, opened 2024-04-19
Use flatMap to simplify map operations

holgerfriedrich:
minor code cleanup, basically just replacing
```.filter(Optional::isPresent).map(Optional::get)``` by ```.flatMap(Optional::stream)```


J-N-K (approved):
LGTM, thanks!

PR #4191, opened 2024-04-19
More instanceof pattern matching

holgerfriedrich:
Some clean-up.

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleRegistryImpl.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleRegistryImpl.java
@@ -589,7 +589,7 @@ public class RuleRegistryImpl extends AbstractRegistry<Rule, String, RuleProvide
                 return configValue instanceof Boolean;
             case INTEGER:
                 return configValue instanceof BigDecimal || configValue instanceof Integer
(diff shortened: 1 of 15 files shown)


bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleRegistryImpl.java
```diff
@@ -589,7 +589,7 @@ private boolean checkType(Type type, Object configValue) {
                 return configValue instanceof Boolean;
             case INTEGER:
                 return configValue instanceof BigDecimal || configValue instanceof Integer
-                        || configValue instanceof Double && ((Double) configValue).intValue() == (Double) configValue;
+                        || configValue instanceof Double doubleValue && doubleValue.intValue() == (Double) configValue;
```
J-N-K, line 592:
```suggestion
                        || configValue instanceof Double doubleValue && doubleValue.intValue() == doubleValue;
```

J-N-K (changes requested):
Otherwise LGTM

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleRegistryImpl.java
holgerfriedrich: done

J-N-K (approved):
Thanks!

PR #4194, opened 2024-04-21
[persistence] improve building the ItemHistoryDTO

joerg1985:
This PR will improve throughput and memory consumption while building the ItemHistoryDTO by:
- removing not needed auto-boxing of a field
- apply the diagonal lines workaround only when needed
- ensure the internal cache of `BigDecimal.toString` is hit and the `BigDecimal` is not rebuilded (relevant since PR #16379 reuses the State)


J-N-K (approved):
Thanks, LGTM!

PR #4198, opened 2024-04-26
Reuse existing BigDecimals in rules while converting types

joerg1985:
# Description

This PR will speed up calculations in rules by handling translation of common number types to `BigDecimal`.
These common number types should match most cases when calculations are done in rules.

### DIFF
--- a/bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/lib/NumberExtensions.java
+++ b/bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/lib/NumberExtensions.java
@@ -380,6 +380,20 @@ public class NumberExtensions {
             }
             return null;
         }
+        if (number instanceof DecimalType dtype) {
+            return dtype.toBigDecimal();
+        }
+        if (number instanceof BigDecimal decimal) {
+            return decimal;
+        }
+        if (number instanceof Long value) {
+            // use valueOf to hit the internal cache
+            return BigDecimal.valueOf(value);
+        }
+        if (number instanceof Integer value) {
+            // use valueOf to hit the internal cache
+            return BigDecimal.valueOf(value);
+        }
         if (number != null) {
             return new BigDecimal(number.toString());
         } else {
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DecimalType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DecimalType.java
@@ -53,7 +53,7 @@ public class DecimalType extends Number implements PrimitiveType, State, Command
     public DecimalType(Number value) {
         if (value instanceof QuantityType type) {
             this.value = type.toBigDecimal();
-        } else if (value instanceof HSBType type) {
+        } else if (value instanceof DecimalType type) {
             this.value = type.toBigDecimal();
         } else if (value instanceof BigDecimal decimal) {
             this.value = decimal;
--- a/bundles/org.openhab.core.model.script/src.moved/test/java/org/openhab/core/model/script/lib/NumberExtensionsTest.java
+++ b/bundles/org.openhab.core.model.script/src.moved/test/java/org/openhab/core/model/script/lib/NumberExtensionsTest.java
@@ -24,7 +24,11 @@ import javax.measure.quantity.Length;
 import javax.measure.quantity.Temperature;
 
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 import org.openhab.core.library.types.DecimalType;
+import org.openhab.core.library.types.HSBType;
+import org.openhab.core.library.types.PercentType;
 import org.openhab.core.library.types.QuantityType;
 import org.openhab.core.library.unit.Units;
 import org.openhab.core.types.Type;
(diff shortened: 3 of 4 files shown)


bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/lib/NumberExtensions.java
```diff
@@ -380,6 +380,24 @@ public static BigDecimal numberToBigDecimal(Number number) {
...
+            return dtype.toBigDecimal();
+        }
+        if (number instanceof BigDecimal decimal) {
+            return decimal;
+        }
+        if (number instanceof Double value) {
+            return new BigDecimal(value);
+        }
+        if (number instanceof Float value) {
+            return new BigDecimal(value);
```
holgerfriedrich, line 393: same as for double

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/lib/NumberExtensions.java
```diff
@@ -380,6 +380,24 @@ public static BigDecimal numberToBigDecimal(Number number) {
...
+        }
+        if (number instanceof Long value) {
+            return new BigDecimal(value);
```
holgerfriedrich, line 396: For type long maybe better use valueOf(long), which should allow for internal caching mechanisms.
See API note here https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/math/BigDecimal.html#valueOf(long)

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DecimalType.java
```diff
@@ -53,7 +53,7 @@ public DecimalType() {
     public DecimalType(Number value) {
         if (value instanceof QuantityType type) {
             this.value = type.toBigDecimal();
-        } else if (value instanceof HSBType type) {
+        } else if (value instanceof DecimalType type) {
```
holgerfriedrich, line 56: now uses internal representation, should be faster... As ```DecimalType``` is base class of ```HSBType``` and ```PercentType```, this will apply to more types.

holgerfriedrich (commented):
Thank you for the contribution!

I added a first review.
For me, ```BigDecimal``` is always tricky, especially when it comes to preserving the accuracy. Using the Ctors might behave differently compared to using the static factory methods ```valueOf()```. Using ```toString``` on the floating points before construction the BigDecimal might (in this case intended) limit the number of decimal places.

Did you already check if this PR changes the currently implemented behavior?

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DecimalType.java
joerg1985: Yes, this is only to hit more types

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/lib/NumberExtensions.java
joerg1985: Thanks for the hint, will change this.

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/lib/NumberExtensions.java
```diff
@@ -380,6 +380,24 @@ public static BigDecimal numberToBigDecimal(Number number) {
...
+        }
+        if (number instanceof Integer value) {
+            return new BigDecimal(value);
```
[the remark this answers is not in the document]
joerg1985: Thanks for the hint, will change this.

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/lib/NumberExtensions.java
joerg1985: see comment above

bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/lib/NumberExtensions.java
```diff
@@ -380,6 +380,24 @@ public static BigDecimal numberToBigDecimal(Number number) {
...
             return null;
         }
+        if (number instanceof DecimalType dtype) {
+            return dtype.toBigDecimal();
+        }
+        if (number instanceof BigDecimal decimal) {
+            return decimal;
+        }
+        if (number instanceof Double value) {
+            return new BigDecimal(value);
```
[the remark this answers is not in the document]
joerg1985: You are right it is better to run into the default case here, so i will remove the Double and Float case here.
I think the impact is not that big, this will only hit when a hardcoded value is applied and this should be without a crazy long fraction part.

joerg1985:
No i did not check the PR does change the behavior, i did expect the impact to be minimal.

I think there might be a change as the `DecimalType.toBigDecimal()` might return a more accurate representation than `new BigDecimal(DecimalType.toString())`, see javadoc of `BigDecimal.toPlainString`.

holgerfriedrich:
@joerg1985, I added a test case to see if this changed the semantics, see my PR in your repo.
In deed, there is one change, as ```NumberExtensions.numberToBigDecimal()``` could not handle ```HSBType``` before (NumberFormatException) and now uses ```HSBType.toBigDecimal``` which converts the "V" part properly.

For me, this behaviour is fine.

joerg1985:
@holgerfriedrich i have merged it into the PR branch, thanks

J-N-K (approved):
LGTM, thanks!

PR #4200, opened 2024-04-28
[persistence] TestPersistenceService w/ paging

ssalonen:
Previously it was not supporting paging. This validated correct behaviour with historicState & lastUpdate & previousState that have the pageSize defined.


J-N-K (commented):
LGTM, but do we have tests that rely on this behavior?

ssalonen:
No, the tests work either way. Now with this PR we have a bit more regression testing for those methods that utilize page size

J-N-K (approved):

PR #4204, opened 2024-04-30
Support press and release buttons in sitemap generator

mherwege:
Depends on: https://github.com/openhab/openhab-core/pull/4183

This PR adds support for press and release buttons in the sitemap generator.

It will required further testing after merging Depends on: https://github.com/openhab/openhab-core/pull/4183

The sitemap generator in the UI so far was stripping quotes from arguments in mappings and conditions sent to core.

As more functionality is added to the sitemap syntax, this is leading to problems to correctly interpret the arguments containing complex syntax:
- more characters become not allowed (splitting on =, :, ...)
- blanks are not always treated properly

To get over this limit, quote stripping should be removed on the UI side, but that means it needs to be done in core when creating the widget from the UIComponent.

Related to: https://github.com/openhab/openhab-webui/pull/2553


lolodomo:
Looks good but you will have to reinject the code for splitting press and release commands as I removed it from #4183

mherwege:
@lolodomo I don't know what happened, but the push of my commit seems to have been incomplete. I was sure I did the code for it. Should be OK now.

lolodomo (approved):
LGTM

J-N-K (approved):
LGTM

PR #4206, opened 2024-05-02
Fix initial suggestion finder installation

mherwege:
https://github.com/openhab/openhab-core/pull/4188 caused finders not to be installed anymore by default.
This PR should fix the issue.

Issue #4158, opened 2024-03-22
Resolver error introduced with Karaf upgrade to 4.4.5


holgerfriedrich:
> @florian-h05: I'm getting these error messages on my openHAB dev machine (latest snapshot) since this PR:

```
 org.apache.felix.resolver.reason.ReasonException: Unable to resolve root: missing requirement [root] osgi.identity; osgi.identity=kar; type=karaf.feature; version="[4.4.5,4.4.5]"; filter:="(&(osgi.identity=kar)(type=karaf.feature)(version>=4.4.5)(version<=4.4.5))" [caused by: Unable to resolve kar/4.4.5: missing requirement [kar/4.4.5] osgi.identity; osgi.identity=org.apache.karaf.kar.core; type=osgi.bundle; version="[4.4.5,4.4.5]"; resolution:=mandatory [caused by: Unable to resolve org.apache.karaf.kar.core/4.4.5: missing requirement [org.apache.karaf.kar.core/4.4.5] osgi.wiring.package; filter:="(&(osgi.wiring.package=org.osgi.framework)(version>=1.0.0)(!(version>=3.0.0)))" [caused by: Unable to resolve org.eclipse.osgi/3.18.0.v20220516-2155: missing requirement [org.eclipse.osgi/3.18.0.v20220516-2155] osgi.ee; filter:="(| (&(osgi.ee=JavaSE)(version=1.8)) (&(osgi.ee=JavaSE/compact1)(version=1.8)) )"]]]
        at org.apache.felix.resolver.Candidates$MissingRequirementError.toException(Candidates.java:1341)
        at org.apache.felix.resolver.ResolverImpl.doResolve(ResolverImpl.java:433)
        at org.apache.felix.resolver.ResolverImpl.resolve(ResolverImpl.java:420)
        at org.apache.felix.resolver.ResolverImpl.resolve(ResolverImpl.java:374)
        at org.apache.karaf.features.internal.region.SubsystemResolver.resolve(SubsystemResolver.java:256)
        at org.apache.karaf.features.internal.service.Deployer.deploy(Deployer.java:399)
        at org.apache.karaf.features.internal.service.FeaturesServiceImpl.doProvision(FeaturesServiceImpl.java:1069)
        at org.apache.karaf.features.internal.service.FeaturesServiceImpl.lambda$doProvisionInThread$13(FeaturesServiceImpl.java:1004)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
[...]

PR #4209, opened 2024-05-03
Fix suggestion finder installation

mherwege:
It looks like https://github.com/openhab/openhab-core/pull/4206 didn't fix the issue completely.

Here is a complement, that so far works well for me. The issue was in the variable sequence of starting the AddonSuggestionService and KarafAddonFinderService.

@jlaur Can you have a look if this works for you as well?

@J-N-K Sorry for these iterations. Hopefully, it works out well this time.

### DIFF
--- a/bundles/org.openhab.core.config.discovery.addon/src/main/java/org/openhab/core/config/discovery/addon/AddonSuggestionService.java
+++ b/bundles/org.openhab.core.config.discovery.addon/src/main/java/org/openhab/core/config/discovery/addon/AddonSuggestionService.java
@@ -80,7 +80,7 @@ public class AddonSuggestionService implements AutoCloseable {
         this.configurationAdmin = configurationAdmin;
         this.localeProvider = localeProvider;
 
-        SUGGESTION_FINDERS.forEach(f -> baseFinderConfig.put(f, true));
+        SUGGESTION_FINDERS.forEach(f -> baseFinderConfig.put(f, false));
         modified(config);
 
         // Changes to the configuration are expected to call the {@link modified} method. This works well when running
@@ -115,13 +115,13 @@ public class AddonSuggestionService implements AutoCloseable {
             String cfgParam = SUGGESTION_FINDER_CONFIGS.get(finder);
             if (cfgParam != null) {
                 boolean enabled = (config != null)
-                        ? ConfigParser.valueAsOrElse(config.get(cfgParam), Boolean.class, cfg)
-                        : cfg;
+                        ? ConfigParser.valueAsOrElse(config.get(cfgParam), Boolean.class, true)
+                        : true;
                 if (cfg != enabled) {
-                    baseFinderConfig.put(finder, enabled);
                     String type = SUGGESTION_FINDER_TYPES.get(finder);
                     AddonFinderService finderService = addonFinderService;
                     if (type != null && finderService != null) {
+                        baseFinderConfig.put(finder, enabled);
                         if (enabled) {
                             finderService.install(type);
                         } else {
--- a/bundles/org.openhab.core.config.discovery.addon/src/main/java/org/openhab/core/config/discovery/addon/AddonSuggestionService.java
+++ b/bundles/org.openhab.core.config.discovery.addon/src/main/java/org/openhab/core/config/discovery/addon/AddonSuggestionService.java
@@ -116,7 +116,7 @@ public class AddonSuggestionService implements AutoCloseable {
             if (cfgParam != null) {
                 boolean enabled = (config != null)
                         ? ConfigParser.valueAsOrElse(config.get(cfgParam), Boolean.class, true)
-                        : true;
+                        : cfg;
                 if (cfg != enabled) {
                     String type = SUGGESTION_FINDER_TYPES.get(finder);
                     AddonFinderService finderService = addonFinderService;


bundles/org.openhab.core.config.discovery.addon/src/main/java/org/openhab/core/config/discovery/addon/AddonSuggestionService.java
```diff
@@ -80,7 +80,7 @@ public AddonSuggestionService(final @Reference ConfigurationAdmin configurationA
         this.configurationAdmin = configurationAdmin;
         this.localeProvider = localeProvider;
 
-        SUGGESTION_FINDERS.forEach(f -> baseFinderConfig.put(f, true));
+        SUGGESTION_FINDERS.forEach(f -> baseFinderConfig.put(f, false));
         modified(config);
 
         // Changes to the configuration are expected to call the {@link modified} method. This works well when running
```
jlaur, line 86: Side question: Is this job still needed after #4188?

jlaur:
> Can you have a look if this works for you as well?

I did a quick test run:
- Latest snapshot (4064) installed.
- The code from this PR compiled.
- `org.openhab.core.config.discovery.addon-4.2.0-SNAPSHOT.jar` replaced.
- `cache` and `tmp` removed.

and it still doesn't seem to work for me:
```shell
openhab> bundle:list -s | grep addon
158 │ Active │  80 │ 4.2.0.202405030306    │ org.openhab.core.addon
159 │ Active │  80 │ 4.2.0.202405030313    │ org.openhab.core.addon.marketplace
160 │ Active │  80 │ 4.2.0.202405030314    │ org.openhab.core.addon.marketplace.karaf
171 │ Active │  80 │ 4.2.0.202405031242    │ org.openhab.core.config.discovery.addon
172 │ Active │  80 │ 4.2.0.202405030309    │ org.openhab.core.config.discovery.addon.process
```

bundles/org.openhab.core.config.discovery.addon/src/main/java/org/openhab/core/config/discovery/addon/AddonSuggestionService.java
mherwege: It is. The reason is that it refers to a configuration from a different component (to avoid proliferation of component configurations all separated out in the UI) and in Karaf that does not get synchronized reliably.

mherwege:
> I did a quick test run:
> 
> * Latest snapshot (4064) installed.
> * The code from this PR compiled.
> * `org.openhab.core.config.discovery.addon-4.2.0-SNAPSHOT.jar` replaced.
> * `cache` and `tmp` removed.

I did exactly that, same snapshot, compiled version (in Eclipse), and it is working for me. I am testing on Windows.

```
openhab> bundle:list -s | grep addon
158 │ Active │  80 │ 4.2.0.202405030306    │ org.openhab.core.addon
159 │ Active │  80 │ 4.2.0.202405030313    │ org.openhab.core.addon.marketplace
160 │ Active │  80 │ 4.2.0.202405030314    │ org.openhab.core.addon.marketplace.karaf
171 │ Active │  80 │ 4.2.0.202405030931    │ org.openhab.core.config.discovery.addon
172 │ Active │  80 │ 4.2.0.202405030309    │ org.openhab.core.config.discovery.addon.process
252 │ Active │  80 │ 4.2.0.202405030309    │ org.openhab.core.config.discovery.addon.ip
253 │ Active │  80 │ 4.2.0.202405030314    │ org.openhab.core.config.discovery.addon.mdns
254 │ Active │  80 │ 4.2.0.202405030310    │ org.openhab.core.config.discovery.addon.upnp
255 │ Active │  80 │ 4.2.0.202405030313    │ org.openhab.core.config.discovery.addon.usb
```

Notice the last 4, that only become available after a little wait.
It would be useful to monitor the trace log in your installation. Could it be that some of the packages are not downloaded yet, and it makes the feature installer fail? Look for an error on that.

holgerfriedrich:
@mherwege I did a full core build on Windows and prepared a distro zip w/o add-ons.
Though it seems to improve the loading of finders, on my machine the resolver gets confused which prevents OH from installing the add-ons.

I have checked with a lot of different versions starting even before #4188, all builds lead to system which runs successfully (including wizard and installation of plugins). 
Rebasing this PR on op of upstream/main leads to the same problem and thus fails.

This might be a coincidence. Did you see those problems as well?

mherwege:
I didn’t do a full build, but when testing I saw a few of the errors like in https://github.com/openhab/openhab-core/issues/4158. As these are generated in the feature installer, using the feature installer for installing the finders may have aggregated the problem.

holgerfriedrich:
I did another test rolling back 4188 and 4206. Better, but not without resolver errors. Could install a few add-ons in the wizard, then broke down. So you might be right.

J-N-K (approved):

holgerfriedrich:
I can confirm that it is working on S4065 on my RPI. All finders are back! 🎉 

Details:
Upgrade to S4065 via apt worked.
Stopping OH and restart after cleaning the cache worked as well.
A few add-ons are installed there for testing, e.g. knx and mapdb. Installed new add-on zwave.
No resolver issues seen on this installation.

PR #4213, opened 2024-05-03
Persistence extensions: allow null serviceId

mherwege:
See https://github.com/openhab/openhab-core/pull/3736#issuecomment-2093140530

### DIFF
--- a/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
+++ b/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
@@ -80,7 +80,7 @@ public class PersistenceExtensions {
      * @param item the item to store
      */
     public static void persist(Item item) {
-        internalPersist(item, (String) null);
+        internalPersist(item, null);
     }
 
     /**
@@ -95,17 +95,17 @@ public class PersistenceExtensions {
     }
 
     private static void internalPersist(Item item, @Nullable String serviceId) {
-        String sId = serviceId == null ? getDefaultServiceId() : serviceId;
-        if (sId == null) {
+        String effectiveServiceId = serviceId == null ? getDefaultServiceId() : serviceId;
+        if (effectiveServiceId == null) {
             return;
         }
-        PersistenceService service = getService(sId);
+        PersistenceService service = getService(effectiveServiceId);
         if (service != null) {
             service.store(item);
             return;
         }
         LoggerFactory.getLogger(PersistenceExtensions.class)
-                .warn("There is no persistence service registered with the id '{}'", sId);
+                .warn("There is no persistence service registered with the id '{}'", effectiveServiceId);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
```diff
@@ -80,7 +80,7 @@ public PersistenceExtensions(@Reference PersistenceServiceRegistry registry,
      * @param item the item to store
      */
     public static void persist(Item item) {
-        internalPersist(item);
+        internalPersist(item, (String) null);
```
J-N-K, line 83: Why do you need a cast to `String` here?

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
```diff
@@ -90,26 +90,22 @@ public static void persist(Item item) {
...
-    public static void persist(Item item, String serviceId) {
+    public static void persist(Item item, @Nullable String serviceId) {
         internalPersist(item, serviceId);
     }
 
-    private static void internalPersist(Item item) {
-        String serviceId = getDefaultServiceId();
-        if (serviceId == null) {
+    private static void internalPersist(Item item, @Nullable String serviceId) {
+        String sId = serviceId == null ? getDefaultServiceId() : serviceId;
```
J-N-K, line 98:
```suggestion
        String effectiveServiceId = Objects.requireNonNullElse(serviceId, getDefaultServiceId());
```

Also below

J-N-K (changes requested):
Otherwise a nice addition.

jimtng:
Perhaps label this as regression?

jimtng:
Also to ensure that this is honored in the future, could we please add some tests to check?

J-N-K:
I don't think it's a regression. It didn't work before, right?

jimtng:
> I don't think it's a regression. It didn't work before, right?

It worked like this before #3736:
When given null as the serviceId, it would use the default serviceId, the same as when the serviceId is omitted completely.

#3736 removed this and instead, it actually looks up `null` in the PersistenceServiceRegistry and naturally not find it, and as a result, fails the call by returning null. So yes this is very much a regression, although as noted, this behaviour wasn't documented in the javadoc.

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
mherwege: Was needed when I only half changed the file (version with signature `internalPersist(Item, TimeSeries)`), but no longer needed, so removed.

mherwege:
> When given null as the serviceId, it would use the default serviceId, the same as when the serviceId is omitted completely.

True, it would, but relying on it was dangerous. For some actions it did not work, and for others it gave wrong result as it was never meant to be used like that. For some, you got lucky and it actually did end up giving the right result.

J-N-K (approved):
Thanks!

PR #4214, opened 2024-05-04
Prevent potential incorrect cache lookup due to hash collisions

joerg1985:
This PR does replace the hash as key for the cache lookup with a `record` and removes the `ExpiringCacheMap`.

1. The lookup inside the cache could return wrong results when identical hashs are computed. This might be unlikely but possible and should be avoided. 
**Note:** Using `Objects.hash` does create an array for the varargs, so there should be no performance penalty when creating a new `record` instead.

2. The `ExpiringCacheMap` does not actively remove entries, it will only prevent entries older than the timeout to be read. There is no need to rebuild the cache after some time, i guess the idea was not to leak memory here. But this is not the case, therefore the regular `ConcurrentHashMap` can be used.
**Note:** This was the only use of `ExpiringCacheMap` i could see, so this class might get deprecated?

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/CommunicationManager.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/CommunicationManager.java
@@ -97,7 +97,6 @@ import org.slf4j.LoggerFactory;
 public class CommunicationManager implements EventSubscriber, RegistryChangeListener<ItemChannelLink> {
 
     private record CacheKey(String type, Profile profile, Thing thing) {
-
     }
 
     private static final Profile NO_OP_PROFILE = new Profile() {
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/CommunicationManager.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/CommunicationManager.java
@@ -12,7 +12,6 @@
  */
 package org.openhab.core.thing.internal;
 
-import java.time.Duration;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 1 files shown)


lolodomo:
> 2\. **Note:** This was the only use of `ExpiringCacheMap` i could see, so this class might get deprecated?

Note that it is used by several bindings.

J-N-K:
@kaikreuzer Do you remember why you used the `ExpiringCacheMap` when creatig #1880? I tend to agree that at least with the current implementation a `ConcurrentHashMap` is also fine.

joerg1985:
> > 2. **Note:** This was the only use of `ExpiringCacheMap` i could see, so this class might get deprecated?
> 
> Note that it is used by several bindings.

@lolodomo i missed this, thanks for the hint

kaikreuzer:
> @kaikreuzer Do you remember why you used the ExpiringCacheMap

I assume that the idea was indeed to avoid memory leaks, or rather that the number of cache entries can be reduced again if entries are no longer used. I probably didn't notice at the time that entries are not removed but merely renewed by the `ExpiringCacheMap`. So I guess that `ConcurrentHashMap` is a suitable replacement here indeed, while it also has the issue that instances are never released, if no longer used. Might we want to wrap the instances into a `WeakReference` to make this behavior possible? This could be a separate PR, though.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/CommunicationManager.java
```diff
@@ -98,6 +96,10 @@
 @Component(service = { EventSubscriber.class, CommunicationManager.class }, immediate = true)
 public class CommunicationManager implements EventSubscriber, RegistryChangeListener<ItemChannelLink> {
 
+    private record CacheKey(String type, Profile profile, Thing thing) {
+
+    }
```
J-N-K, line 101:
```suggestion
    private record CacheKey(String type, Profile profile, Thing thing) {
    }
```

J-N-K (approved):
Thanks!

PR #4216, opened 2024-05-05
Add support for marketplace transformations

### LINKED ISSUE #4046: [Marketplace] Add support for transformations to be installed from the marketplace
#The problem
I find myself reposting some of the same transformation scripts to the forum over and over (e.g. split a number into DD:HH:MM:SS, filter out changes that are too small, unit conversions not supported by UoM, etc.).

#Your suggestion
Add a new category to the marketplace to install transformations similar to the way that rule templates can be installed. 

I don't think these need to support properties like a rule template because arguments can be passed to them when they are applied. But if it's a no brainer to support that I'm sure we'll find a use for it. :-D

Issue for MainUI is https://github.com/openhab/openhab-webui/issues/2239#issuecomment-1889950735

J-N-K:
Closes #4046

### DIFF
--- a/bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityMarketplaceAddonService.java
+++ b/bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityMarketplaceAddonService.java
@@ -74,8 +74,9 @@ import org.slf4j.LoggerFactory;
 @ConfigurableService(category = "system", label = CommunityMarketplaceAddonService.SERVICE_NAME, description_uri = CommunityMarketplaceAddonService.CONFIG_URI)
 @NonNullByDefault
 public class CommunityMarketplaceAddonService extends AbstractRemoteAddonService {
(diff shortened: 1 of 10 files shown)


bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/community/CommunityMarketplaceAddonService.java
```diff
@@ -103,6 +101,8 @@ public class CommunityMarketplaceAddonService extends AbstractRemoteAddonService
     private static final Integer RULETEMPLATES_CATEGORY = 74;
     private static final Integer UIWIDGETS_CATEGORY = 75;
     private static final Integer BLOCKLIBRARIES_CATEGORY = 76;
+    // TODO: confirm category
+    private static final Integer TRANSFORMATIONS_CATEGORY = 77;
```
kaikreuzer, line 105:
```suggestion
    private static final Integer TRANSFORMATIONS_CATEGORY = 80;
```
See https://community.openhab.org/c/marketplace/transformations/

kaikreuzer (approved):
lgtm, thanks!

PR #4217, opened 2024-05-05
Improve thing description XSD

### LINKED ISSUE #4156: Dimension in `thing-description-1.0.0.xsd`
I randomly noticed that there is a list of dimensions in [thing-description-1.0.0.xsd](https://www.openhab.org/schemas/thing-description-1.0.0.xsd):

https://github.com/openhab/openhab-core/blob/c9ee9920478c783ea4f40dacab79ea41569ef505/bundles/org.openhab.core.thing/schema/thing/thing-description-1.0.0.xsd#L201-L250

This list hasn't been updated for a while, for example the following are missing:
- **DataAmount** introduced in #681
- **EmissionIntensity** introduced in #4069
- **EnergyPrice** introduced in #3503
- **RadiationSpecificActivity** introduced in #3608

and probably others as well.

But then I validated a thing type definition containing one of those, and it passed. Apparently `dimension` is not used at all, since `item-type` is defined as `xs:string`:
https://github.com/openhab/openhab-core/blob/c9ee9920478c783ea4f40dacab79ea41569ef505/bundles/org.openhab.core.thing/schema/thing/thing-description-1.0.0.xsd#L66

So I'm wondering: Should it even be in the file? Is it a left-over, or deliberately kept for some sort of legacy backwards compatibility?

## Expected Behavior

I would expect `dimension` to be either maintained and used, or not in the XSD at all.

## Current Behavior

The XSD contains a definition for `dimension`, but apparently it is not used.

## Possible Solution

Remove it from the XSD, if this is feasible without breaking compatibility and/or requiring a new version, which might complicate things.

J-N-K:
Fixes #4156 

Enables checking if item-type in channel defintion of things

### DIFF
--- a/bundles/org.openhab.core.thing/schema/thing/thing-description-1.0.0.xsd
+++ b/bundles/org.openhab.core.thing/schema/thing/thing-description-1.0.0.xsd
@@ -63,7 +63,7 @@
 
     <xs:complexType name="channelType">
         <xs:sequence>
-            <xs:element name="item-type" type="thing-description:itemType" minOccurs="0"/>
+             <xs:element name="item-type" type="thing-description:itemType" minOccurs="0"/>
             <xs:element name="kind" type="xs:string" minOccurs="0"/>
             <xs:element name="label" type="xs:string"/>
             <xs:element name="description" type="xs:string" minOccurs="0"/>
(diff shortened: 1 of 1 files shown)


andrewfg:
Is this related to https://github.com/openhab/openhab-core/issues/4156 ??

holgerfriedrich:
If I move the things a bit around, this is the diff:
image
image

kaikreuzer (approved):
lgtm, thanks!

kaikreuzer:
FTR: I've also updated the version on the website with https://github.com/openhab/website/pull/461

holgerfriedrich:
Add-on build is broken since this PR is merged.

```Number:Density``` is used by a few add-ons, but the XSD only allows ```Number:VolumetricDensity```.

Interestingly, SAT only fails for onewire, but not for airq.
Manual schema validation shows both xml files invalid, unless ```Number:Density``` is replaced by  ```Number:VolumetricDensity```.

J-N-K:
I'll align `I18nProviderImpl` and the XSD in a follow-up PR.

PR #4228, opened 2024-05-14
Harmonize library versions

holgerfriedrich:
* Use jdt.annotation 2.2.100 (i18n tool)
* Use gson 2.10.1 (i18n tool, upgrade tool)
* Use logback 1.3.14 (test)

This PR targets test and tools. Other upgrades will be done together with the next Karaf upgrade.
Upgrading logback should fix a dependabot warning, 
https://github.com/openhab/openhab-core/security/dependabot/77

Hope this does not break anything, as we don't have many tests esp. for the upgradetool.


wborn (approved):
Thanks!

PR #4229, opened 2024-05-14
Update i18n property files

holgerfriedrich:
Looks like a few strings are missing.

There are still some translations for binding.* instead of addon.* - maybe those should be fixed as well.


J-N-K (approved):
LGTM

PR #4232, opened 2024-05-17
String translations for IP add-on finder

holgerfriedrich:
Translate backslash sequences in requestPlain.
See [String.translateEscapes().](https://docs.oracle.com/javase/specs/jls/se15/html/jls-3.html#jls-3.10.7)

See discussion in openhab/openhab-addons#16768

@jlaur Would you like to test if this works out?


mlobstein:
Works as expected: 
image

wborn (approved):
Thanks!

PR #4233, opened 2024-05-18
Align I18nProviderImpl and thing description XSD

J-N-K:
See https://github.com/openhab/openhab-core/pull/4217#issuecomment-2118651155


holgerfriedrich:
can you please double check the following types which do not occur in the whole core and add-ons repo:
```
Number:AmountOfSubstance
Number:CatalyticActivity
Number:Currency
Number:ElectricCapacitance
Number:ElectricConductance
Number:ElectricConductivity
Number:ElectricInductance
Number:Force
Number:LuminousFlux
Number:LuminousIntensity
Number:MagneticFlux
Number:MagneticFluxDensity
Number:RadioactiveActivity
Number:SolidAngle
```

The following types have an implementation in core, but are not found in add-ons:
```
Number:RadiationDoseAbsorbed
Number:RadiationDoseEffective
```

J-N-K:
I only aligned the item-types to the supported units (i.e. those that have a default dimension). These were added in #3143, I can't remember exactly why I choose all those, but they are consistent with the list of supported dimensions [here](https://www.openhab.org/docs/concepts/units-of-measurement.html), so I guess we should not remove them.

holgerfriedrich (approved):
LGTM, thanks

kaikreuzer:
I've updated the website accordingly with https://github.com/openhab/website/pull/462

PR #4224, opened 2024-05-11
Fix resolver errors during bundle add-on installation

J-N-K:
Fixes #4222

@lolodomo @holgerfriedrich @jimtng Can you confirm this also fixes the issues on your systems? My Win11/Parallels now works fine.

PR #4235, opened 2024-05-20
Fix exception in WindowsUsbSerialDiscovery

mherwege:
Fixes https://github.com/openhab/openhab-core/pull/4224#issuecomment-2119849776


J-N-K (approved):
Thanks

lolodomo:
Thank you @mherwege 
Did you check we don't have the same issue in the other UsbSerialDiscovery ?

mherwege:
> Thank you @mherwege 
> Did you check we don't have the same issue in the other UsbSerialDiscovery ?

No, I didn't.

lolodomo:
I found `Ser2NetUsbSerialDiscovery`: no bug because the deactivate method is missing. Should probably be added.

PR #4237, opened 2024-05-21
New SDDP service for addon discovery and thing discovery

### LINKED ISSUE #4234: Add SDDP service to core
If regex matching is implemented then the epsonprojector and sonyprojector addons could be detected as many newer models respond to Control4 SDDP discovery probes as described here:

https://github.com/sammck/sddp-discovery-protocol/blob/main/README.md

SDDP is similar to UPNP/SSDP in that a multicast search frame is sent and any supported devices respond back to the multicast address with a packet containing device information. By using a regex these packets would be filtered to determine the device type and manufacturer.

_Originally posted by @mlobstein in https://github.com/openhab/openhab-core/issues/3936#issuecomment-2118832389_

---

Note: I myself (via my then employer) was one of the founder members at the original launch of UPNP hosted by Microsoft in Redmond, WA., so I am very familiar with coding for SSDP. => I will volunteer myself to write an equivalent SDDP module for OH core. The goal would be to provide an SDDP service with twofold application..

1. Use the new SDDP service to implement in OH Core an Addon Finder for Things that are discoverable via SDDP
2. Expose the new SDDP service to OH Addon authors for them to use it in discovering Things and respective properties.

andrewfg:
Resolves #4234

This PR provides an SDDP service for discovering equipment on the LAN. It has twofold application:

1. Bindings can implement an `SddpDiscoveryParticipant` that allows discovering Things. It operates similarly to the MDNS and UPNP discovery services.
2. The core uses the same service and implements an Addon Finder. It operates similarly to the MDNS and UPNP addon finders.


andrewfg:
Pinging @mherwege and @mlobstein for info..

bundles/org.openhab.core.config.discovery.sddp/src/test/java/org/openhab/core/config/discovery/sddp/test/SddpDiscoveryTests.java
```diff
@@ -0,0 +1,129 @@
...
+            "From: \"192.168.4.237:1902\"\r\n" +
+            "Host: \"JVC_PROJECTOR-E0DADC152802\"\r\n" +
+            "Max-Age: 1800\r\n" +
+            "Type: \"JVCKENWOOD:Projector\"\r\n" +
+            "Primary-Proxy: \"projector\"\r\n" +
+            "Proxies: \"projector\"\r\n" +
+            "Manufacturer: \"JVCKENWOOD\"\r\n" +
+            "Model: \"DLA-RS3100_NZ8\"\r\n" +
+            "Driver: \"projector_JVCKENWOOD_DLA-RS3100_NZ8.c4i\"\r\n";
+    // @formatter:on
```
wborn, line 46: Maybe you can use Java text blocks in this file? 

https://www.baeldung.com/java-text-blocks

andrewfg: @wborn that's cool!

andrewfg: Done

wborn: Nice :slightly_smiling_face:

bundles/org.openhab.core.config.discovery.addon.sddp/src/test/java/org/openhab/core/config/discovery/addon/sddp/test/SddpAddonFinderTests.java
```diff
@@ -0,0 +1,90 @@
...
+                "From", "\"192.168.4.237:1902\"",
+                "Host", "\"JVC_PROJECTOR-E0DADC152802\"",
+                "Max-Age", "1800",
+                "Type", "\"JVCKENWOOD:Projector\"",
+                "Primary-Proxy", "\"projector\"",
+                "Proxies", "\"projector\"",
+                "Manufacturer", "\"JVCKENWOOD\"",
+                "Model", "\"DLA-RS3100_NZ8\"",
+                "Driver", "\"projector_JVCKENWOOD_DLA-RS3100_NZ8.c4i\"");
+        // @formatter:on
```
wborn, line 53: Maybe this can also become a Java text block?

wborn: Hmm probably not, because it is a map with key/values.

andrewfg: Indeed.

andrewfg:
Note to Reviewers: SDDP has a very low LAN footprint. In background (passive scan) mode it simply subscribes to, and listens on, the designated SDDP multicast address/port for UDP `NOTIFY ALIVE SDDP` datagrams. And (unlike UPNP) there is no further LAN traffic once a device is observed. And if an active scan is started, if makes just one single UDP `SEARCH * SDDP` ping on the LAN. Therefore I don't think there is any risk of LAN storms. So I think it is NOT necessary for this service to be un-installable. => WDYT?

bundles/org.openhab.core.config.discovery.sddp/src/main/java/org/openhab/core/config/discovery/sddp/SddpDevice.java
```diff
@@ -0,0 +1,84 @@
...
+    public final Instant expireInstant;
+
+    /**
+     * Constructor
+     * 
+     * @param headers a map of parameter name / value pairs.
+     */
+    public SddpDevice(Map<String, String> headers) {
+        from = headers.getOrDefault("From", "").replaceAll("^\"|\"$", "");
+        host = headers.getOrDefault("Host", "").replaceAll("^\"|\"$", "");
```
mlobstein, line 48: Could the host field be further split into a UUID field (the MAC address)? The conventions of this field seems to _Manufacturer-PrimaryProxy-MAC Address_. This will be useful for the creating the ThingUID in a SddpDiscoveryParticipant class.

[SddpDiscoveryParticipant](https://github.com/sammck/sddp-discovery-protocol/blob/main/README.md#servers)

On the one device that I have that spits out SDDP packets, the `Host` field looks like this:
**Host: "Panasonic-dvd-80c755axxxxx"** (mac redacted)

bundles/org.openhab.core.config.discovery.sddp/src/main/java/org/openhab/core/config/discovery/sddp/SddpDevice.java
```diff
@@ -0,0 +1,84 @@
...
+    public final String driver;
+    public final Instant expireInstant;
+
+    /**
+     * Constructor
+     * 
+     * @param headers a map of parameter name / value pairs.
+     */
+    public SddpDevice(Map<String, String> headers) {
+        from = headers.getOrDefault("From", "").replaceAll("^\"|\"$", "");
```
mlobstein, line 47: It would also be helpful to split the **from:** field on the colon and provide the ip address as its own field. The port number here is not particularly useful since it is just the multicast port.

Sample **From:**
From: "192.168.4.237:1902"

bundles/org.openhab.core.config.discovery.sddp/src/main/java/org/openhab/core/config/discovery/sddp/SddpDevice.java
andrewfg: Yes. Ok.

bundles/org.openhab.core.config.discovery.sddp/src/main/java/org/openhab/core/config/discovery/sddp/SddpDevice.java
andrewfg: Yes. Ok.

bundles/org.openhab.core.config.discovery.sddp/src/main/java/org/openhab/core/config/discovery/sddp/SddpDevice.java
andrewfg: @mlobstein 4yi `host.split("-")` would work on your Panasonic but the example that you cite above has **Host: "JVC_PROJECTOR-E0DADC152802"** so we probably need to try `host.split("-|_")` .. or??

bundles/org.openhab.core.config.discovery.sddp/src/main/java/org/openhab/core/config/discovery/sddp/SddpDevice.java
andrewfg: PS even though all examples have port 1902, I believe it would theoretically be possible for another value to be present, so I also added a port property just-in-cas.

andrewfg:
@mherwege / @holgerfriedrich can you please kindly tell me what I need to do to ensure that the SddpAddonFinder will be installed by OH on startup?

bundles/org.openhab.core.config.discovery.sddp/src/main/java/org/openhab/core/config/discovery/sddp/SddpDevice.java
mlobstein: see below

bundles/org.openhab.core.config.discovery.sddp/src/main/java/org/openhab/core/config/discovery/sddp/SddpDevice.java
```diff
@@ -0,0 +1,95 @@
...
+        manufacturer = headers.getOrDefault("Manufacturer", "").replaceAll("^\"|\"$", "");
+        model = headers.getOrDefault("Model", "").replaceAll("^\"|\"$", "");
+        driver = headers.getOrDefault("Driver", "").replaceAll("^\"|\"$", "");
+
+        String[] fromParts = from.split(":");
+        ipAddress = fromParts[0];
+        port = fromParts.length > 1 ? fromParts[1] : "";
+
+        String[] hostParts = host.split("-|_");
+        macAddress = hostParts.length > 2 ? hostParts[2].replaceAll("(..)(?!$)", "$1:") : "";
```
mlobstein, line 65: I think that will work. The only thing the might be better is this to always get the last split and assume that the split of the field before mac address might not happen:
```suggestion
        macAddress = hostParts.length > 1 ? hostParts[hostParts.length - 1].replaceAll("(..)(?!$)", "$1:") : "";
```

andrewfg: Ok.

andrewfg:
**Status Update**

1. @mherwege / @holgerfriedrich as usual I am struggling on how to get the features / components loaded by OH. There are basically two components -- an `SsdpDiscoveryService` (which works similar to `MdnsDiscoveryService` to discover devices on the network), and an `SsdpAddonFinder` (which works similar to `MdnsAddonFinder`). The latter component has an OSGI dependency of the former. I have got it working so far as the `SsdpDiscoveryService` gets loaded, but unfortunately I cannot get the `SsdpAddonFinder` to load. => So can you please help me with this?

2. @mlobstein via the logging, I can see that the discovery service is passively finding stuff that sends SDDP NOTIFY on the LAN.  However due to the above-mentioned lack of `SsdpAddonFinder` it can not yet do a) active SEARCH * SSDP requests, nor b) actually match the actual found devices against the candidate bindings addon.xml. Currently I have two 'test' SDDP devices -- namely the HDPowerView Gen2 hub (which I think has a faulty implementation of SDDP), and a 'fake' device that I wrote myself for testing purposes; where the problem is that I wrote it myself as the mirror my own code for the discovery service, so it is more or less guaranteed to 'work'. => So I will definitely need your testing @mlobstein to confirm that in particular the active SEARCH * SSDP process, and suggestion matching process do both work.

holgerfriedrich:
@andrewfg I tried to get it running, but I end up in the same situation you described above: discovery service is running, add-on part does not get installed. Once I drop the addon it into addons folder, it gets installed and is running. I suspect the feature definition could be wrong. @mherwege did this for the other finders, so maybe he can help out here.

andrewfg:
> I suspect the feature definition could be wrong.

That is my suspicion too. There is zero documentation on the web about the syntax of feature.xml, so I have been trying different combinations all day by trial and error (actually so far only error...). Hoping @mherwege can help..

bundles/org.openhab.core.config.discovery.sddp/src/main/java/org/openhab/core/config/discovery/sddp/SddpDiscoveryService.java
```diff
@@ -0,0 +1,437 @@
...
+    private static final Duration SEARCH_LISTEN_DURATION = Duration.ofSeconds(5);
+    private static final Duration CACHE_PURGE_INTERVAL = Duration.ofSeconds(300);
+
+    private static final String SEARCH_RESPONSE_HEADER = "SDDP/1.0 200 OK";
+    private static final String ALIVE_NOTIFICATION_HEADER = "NOTIFY ALIVE SDDP/1.0";
+
+    private static final String SEARCH_REQUEST_BODY_FORMAT = """
+            SEARCH * SDDP/1.0
+            Host: "%s:%d"
+            """;
```
mlobstein, line 90: I don't using Text Blocks works properly to include the necessary line breaks in the search packet. I got the active scan to work by changing this to have the \r\n line breaks:
```suggestion
    private static final String SEARCH_REQUEST_BODY_FORMAT = "SEARCH * SDDP/1.0\r\n\"Host: %s:%d\"\r\n";
            """;
```

bundles/org.openhab.core.config.discovery.sddp/src/main/java/org/openhab/core/config/discovery/sddp/SddpDiscoveryService.java
```diff
@@ -0,0 +1,437 @@
...
+            packet = null;
+            logger.debug("Packet sent to '{}:{}' content:\r\n{}", SDDP_IP_ADDRESS, SDDP_PORT, search);
+
+            final Instant listenDoneTime = Instant.now().plus(SEARCH_LISTEN_DURATION);
+
+            // loop listen for responses
+            while (Instant.now().isBefore(listenDoneTime) && !Thread.currentThread().isInterrupted()) {
+                try {
+                    if (packet == null) {
+                        packet = new DatagramPacket(buffer, buffer.length);
```
mlobstein, line 309: Re-using the send buffer was causing the received packet content to be truncated:
```suggestion
            final Instant listenDoneTime = Instant.now().plus(SEARCH_LISTEN_DURATION);
            byte[] readBuffer = new byte[READ_BUFFER_SIZE];

            // loop listen for responses
            while (Instant.now().isBefore(listenDoneTime) && !Thread.currentThread().isInterrupted()) {
                try {
                    if (packet == null) {
                        packet = new DatagramPacket(readBuffer, readBuffer.length);
```

bundles/org.openhab.core.config.discovery.sddp/src/main/java/org/openhab/core/config/discovery/sddp/SddpDevice.java
```diff
@@ -0,0 +1,95 @@
...
+        manufacturer = headers.getOrDefault("Manufacturer", "").replaceAll("^\"|\"$", "");
+        model = headers.getOrDefault("Model", "").replaceAll("^\"|\"$", "");
+        driver = headers.getOrDefault("Driver", "").replaceAll("^\"|\"$", "");
+
+        String[] fromParts = from.split(":");
+        ipAddress = fromParts[0];
+        port = fromParts.length > 1 ? fromParts[1] : "";
+
+        String[] hostParts = host.split("-|_");
+        macAddress = hostParts.length > 1 ? hostParts[hostParts.length - 1].replaceAll("(..)(?!$)", "$1:") : "";
```
mlobstein, line 65: Is it necessary to add the colon delimiters? They will just have to be stripped out if this field is to be passed as the unique id to ThingUID();

mlobstein (changes requested):
I got the active scan to work after a little debugging! Please see the review changes in SddpDiscoveryService.

bundles/org.openhab.core.config.discovery.sddp/src/main/java/org/openhab/core/config/discovery/sddp/SddpDiscoveryService.java
andrewfg: Aha. Yes I think Java text blocks produce \n only, whereas SDDP requires \r\n ..

bundles/org.openhab.core.config.discovery.sddp/src/main/java/org/openhab/core/config/discovery/sddp/SddpDiscoveryService.java
andrewfg: Ok

bundles/org.openhab.core.config.discovery.sddp/src/main/java/org/openhab/core/config/discovery/sddp/SddpDevice.java
andrewfg: Not really. I could either remove them entirely, or use dashes (or underscores) instead. I am just thinking that some kind of delimiter makes it easier for the user.

features/karaf/openhab-core/src/main/feature/feature.xml
```diff
@@ -90,6 +91,18 @@
...
 
+	<feature name="openhab-core-config-discovery-sddp" version="${project.version}">
+		<feature>openhab-core-base</feature>
+		<bundle>mvn:org.openhab.core.bundles/org.openhab.core.config.discovery.sddp/${project.version}</bundle>
+	</feature>
+
+	<feature name="openhab-core-config-discovery-addon-sddp" version="${project.version}">
+		<feature>openhab-core-base</feature>
+		<feature>openhab-core-config-discovery-addon</feature>
+		<feature dependency="true">openhab-core-config-discovery-sddp</feature>
```
mherwege, line 102: You should then be able to remove this line.

mherwege:
> IMHO the osgi loader should be clever enough to figure out the sequential loading of two dependent features rather than needing to package stuff into three features like UPNP and MDNS. So before restructuring the architecture it would be nice to have expert feedback. I may be barking up the wrong tree..

I indeed don't think this should be necessary. Just try with the comments I gave above if that solves anything.

features/karaf/openhab-core/src/main/feature/feature.xml
```diff
@@ -90,6 +91,18 @@
 		<bundle>mvn:org.openhab.core.bundles/org.openhab.core.config.discovery.addon.ip/${project.version}</bundle>
 	</feature>
 
+	<feature name="openhab-core-config-discovery-sddp" version="${project.version}">
```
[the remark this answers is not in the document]
andrewfg: > this feature has a dependency on openhab-core-base and the other way around.

Yes, I think the issue was due to a circular dependency.

features/karaf/openhab-core/src/main/feature/feature.xml
andrewfg: See above..

mherwege: This is actually telling you that openhab-core-config-discovery-sddp does not need to be installed when starting this feature, but it will already be there. Depending on the sequence of starting things, this may not be true I think.
See https://stackoverflow.com/questions/51481442/apache-karaf-feature-dependency

andrewfg: > telling you that openhab-core-config-discovery-sddp does not need to be installed when starting this feature

Yes, I got that. However the SddpAddonFinder @activate constructor has a @reference for SddpDiscoveryService (which was in 'openhab-core-config-discovery-sddp' now moved to 'openhab-core-base') so when SddpAddonFinder is instantiated it should create a SddpDiscoveryService (if not already automatically present).

I can see in the logs that SddpDiscoveryService is indeed automatically loaded. And then when I force openhab-core-config-discovery--addon-sddp (by putting the Jar file in the addons folder, or by 'feature:install') I see the error "Not all dependencies satisfied, cannot activate". In the log below, you can see that SddpDiscoveryService is loaded and activated (automaticall) .. yet when tries to load SddpDiscoveryService says there is an unsatisfied reference to SddpDiscoveryService. Aargh @!&*%$

> Depending on the sequence of starting things.

There still seems to be a problem even when (as shown in the log time stamps) the components are started in the correct order.

```
2024-05-27 11:37:54.095 [DEBUG] [.discovery.sddp.SddpDiscoveryService] - bundle org.openhab.core.config.discovery.sddp:4.2.0.202405271007 (173)[org.openhab.core.config.discovery.sddp.SddpDiscoveryService] : BundleComponentActivator : ComponentHolder created.
..
2024-05-27 11:37:54.461 [DEBUG] [.discovery.sddp.SddpDiscoveryService] - bundle org.openhab.core.config.discovery.sddp:4.2.0.202405271007 (173)[org.openhab.core.config.discovery.sddp.SddpDiscoveryService(132)] : Changed state from satisfied to active

2024-05-27 11:38:24.519 [DEBUG] [discovery.addon.sddp.SddpAddonFinder] - bundle org.openhab.core.config.discovery.addon.sddp:4.2.0.202405271008 (269)[org.openhab.core.config.discovery.addon.sddp.SddpAddonFinder] : BundleComponentActivator : ComponentHolder created.
2024-05-27 11:38:24.525 [DEBUG] [discovery.addon.sddp.SddpAddonFinder] - bundle org.openhab.core.config.discovery.addon.sddp:4.2.0.202405271008 (269)[org.openhab.core.config.discovery.addon.sddp.SddpAddonFinder(365)] : New service tracker for $000, initial active: false, previous references: {}, classFilter: (objectClass=org.openhab.core.config.discovery.sddp.SddpDiscoveryService), initialReferenceFilter (objectClass=org.openhab.core.config.discovery.sddp.SddpDiscoveryService)
..
2024-05-27 11:38:24.532 [DEBUG] [discovery.addon.sddp.SddpAddonFinder] - bundle org.openhab.core.config.discovery.addon.sddp:4.2.0.202405271008 (269)[org.openhab.core.config.discovery.addon.sddp.SddpAddonFinder(365)] : Dependency not satisfied: $000
2024-05-27 11:38:24.532 [DEBUG] [discovery.addon.sddp.SddpAddonFinder] - bundle org.openhab.core.config.discovery.addon.sddp:4.2.0.202405271008 (269)[org.openhab.core.config.discovery.addon.sddp.SddpAddonFinder(365)] : Not all dependencies satisfied, cannot activate
```

andrewfg: => perhaps @wborn has some suggestion?

features/karaf/openhab-core/src/main/feature/feature.xml
```diff
@@ -90,6 +91,11 @@
 		<bundle>mvn:org.openhab.core.bundles/org.openhab.core.config.discovery.addon.ip/${project.version}</bundle>
 	</feature>
 
+	<feature name="openhab-core-config-discovery-addon-sddp" version="${project.version}">
+		<feature>openhab-core-base</feature>
+		<bundle>mvn:org.openhab.core.bundles/org.openhab.core.config.discovery.sddp/${project.version}</bundle>
```
mherwege, line 96: I think you removed the wrong line. You should still have:
```
<feature>openhab-core-config-discovery-addon</feature>
```

andrewfg: ... but I will make another build and try again .. :)

bundles/org.openhab.core.config.discovery.sddp/src/main/java/org/openhab/core/config/discovery/sddp/SddpDiscoveryService.java
andrewfg: PS I can confirm that changing \n to \r\n did also fix the problem with my, (what I thought was broken), HD Powerview Hub, so it does now properly respond to the Search request.

https://github.com/openhab/openhab-addons/pull/16790#issuecomment-2133372064

andrewfg: PS @mlobstein please look at the splitters in the HD Powerview Hub 'Host' header. It confirms that the last split is the Mac address, but the previous splits do not comply with your hypothesis for them.

features/karaf/openhab-core/src/main/feature/feature.xml
andrewfg: I tried it like you suggested. There are two issues..

1. In `G:\OH4_TEST\runtime\system\org\openhab\core\bundles` the `org.openhab.core.config.discovery.sddp` module IS installed but `org.openhab.core.config.discovery.addon.sddp` is NOT.

image

2. Adding the `org.openhab.core.config.discovery.addon.sddp` manually to `G:\OH4_TEST\addons` does cause the module to load; but it still reports 'Not all dependencies satisfied, cannot activate'. I am guessing because the jar is in the `addons` folder, it can find the path to the `runtime\system\org\openhab\core\bundles` folder. ??

bundles/org.openhab.core.config.discovery.sddp/src/main/java/org/openhab/core/config/discovery/sddp/SddpDiscoveryService.java
jlaur: You can still use a Text Block, and just provide the `\r` at the end of each line. See https://www.baeldung.com/java-text-blocks section 4.2.

jlaur: I see you did, please disregard. 🙂

features/karaf/openhab-core/src/main/feature/feature.xml
mherwege: > EDIT: Do I need to do something in `openhab-webui` to force the finder to be installed on start-up? I did already follow the coding in OH core to support installation of the finder (as for UPNP and MDNS etc.) .. but I wonder if you had hard-wired something in openhab-webui ??

No. When the configuration property is available, it is used by the AddonSuggestionService to trigger the install of the specific suggestion finder feature. There is nothing in the UI that does that.

andrewfg: >  When the configuration property is available, it is used by the AddonSuggestionService to trigger the install

Just out of interest, do I have to do something specific to make that property available? I seem to recall that for new AddonFinder extension classes the default behaviour (i.e. when there is no entry in the config file) is enabled. But perhaps the sate is disabled by default? In either case, changing the parameter setting via the UI seems not to load/unload the finder either..

mherwege: No, there should not be a need to set it. And it should have an impact when changing it from the UI.
Note that there is a bug for the IP add-on finder, where the default is not set to true, although it does install by default. You see that it has the wrong default value in the UI.

mherwege:
Trying to figure this out. I now compiled with your PR. From the console, I get:
```
openhab> feature:info openhab-core-config-discovery-addon-ip
Feature openhab-core-config-discovery-addon-ip 4.2.0.SNAPSHOT
Feature has no configuration
Feature has no configuration files
Feature depends on:
  openhab-core-base 0.0.0
  openhab-core-config-discovery-addon 0.0.0
Feature contains followed bundles:
  mvn:org.openhab.core.bundles/org.openhab.core.config.discovery.addon.ip/4.2.0-SNAPSHOT
Feature has no conditionals.
openhab> feature:info openhab-core-config-discovery-addon-sddp
Feature openhab-core-config-discovery-addon-sddp 4.2.0.SNAPSHOT
Feature has no configuration
Feature has no configuration files
Feature depends on:
  openhab-core-base 0.0.0
  openhab-core-config-discovery-addon 0.0.0
Feature contains followed bundles:
  mvn:org.openhab.core.bundles/org.openhab.core.config.discovery.sddp/4.2.0-SNAPSHOT
Feature has no conditionals.
```
I would think the feature definition requirements for the IP add-on finder should be fairly similar (also no dependency on underlying transport). Notice the difference in followed bundles.

features/karaf/openhab-core/src/main/feature/feature.xml
```diff
@@ -90,6 +91,12 @@
 		<bundle>mvn:org.openhab.core.bundles/org.openhab.core.config.discovery.addon.ip/${project.version}</bundle>
 	</feature>
 
+	<feature name="openhab-core-config-discovery-addon-sddp" version="${project.version}">
+		<feature>openhab-core-base</feature>
+		<feature>openhab-core-config-discovery-addon</feature>
+		<bundle>mvn:org.openhab.core.bundles/org.openhab.core.config.discovery.sddp/${project.version}</bundle>
```
mherwege, line 97: This needs to be:
`<bundle>mvn:org.openhab.core.bundles/org.openhab.core.config.discovery.addon.sddp/${project.version}</bundle>`

features/karaf/openhab-core/src/main/feature/feature.xml
andrewfg: > Note that there is a bug for the IP add-on finder, where the default is not set to true,

1. Where is the default (not) set?
 
2. I have been logging the AddonSuggestionService and I can see that it is trying to load all of the finders (including this one) at startup. It does this before the SsdpDiscoveryService tries to load, and again after it has loaded. But in neither  case does SddpAddonFinder seem actually to have been loaded. => I wonder if there is an issue with AddonSuggestionService when it is trying to load a finder that has its own @reference dependency on another module in core? (The other finders do also have @reference dependencies, but these, as mentioned, are linked to the UPNP and MDNS communication modules in openhab-tp.

features/karaf/openhab-core/src/main/feature/feature.xml
mherwege: And it probably would not hurt to keep
`<feature>openhab-core-config-discovery-sddp</feature>`
in this feature as well. It would avoid problems if the would be removed from the base feature at some point, and shouldn't hurt. I think the resolver will cope with that.

bundles/org.openhab.core.config.discovery.sddp/src/main/java/org/openhab/core/config/discovery/sddp/SddpDiscoveryService.java
mlobstein: I have some additional observations of SDDP packets from the DB-UB9000 that are not seen in the linked README:

When the device turns off, it sends an offline packet:

> NOTIFY OFFLINE SDDP/1.0
From: "192.168.1.2:1902"
Host: "Panasonic-dvd-80c755xxxxxx"
Type: "Panasonic:dvd"

In the network menu, there is an option to 'Send Identify', when chosen the following packet is sent:

> NOTIFY IDENTIFY SDDP/1.0
From: "192.168.1.2:1902"
Host: "Panasonic-dvd-80c755xxxxxx"
Type: "Panasonic:dvd"
Primary-Proxy: "dvd"
Proxies: "dvd"
Manufacturer: "Panasonic"
Model: "DP-UB9000"
Driver: "panasonic_dvd_dp-ub9000.c4z"

features/karaf/openhab-core/src/main/feature/feature.xml
mherwege: > 1. Where is the default (not) set?

In `bundles/org.openhab.core/src/main/resources/OH-INF/config/addons.xml`

mherwege: > 2\. I have been logging the AddonSuggestionService and I can see that it is trying to load all of the finders (including this one) at startup. It does this before the SsdpDiscoveryService tries to load, and again after it has loaded. But in neither  case does SddpAddonFinder seem actually to have been loaded. => I wonder if there is an issue with AddonSuggestionService when it is trying to load a finder that has its own @reference dependency on another module in core? (The other finders do also have @reference dependencies, but these, as mentioned, are linked to the UPNP and MDNS communication modules in openhab-tp.

See my comments below. You are referencing to the wrong bundle in the feature.

features/karaf/openhab-core/src/main/feature/feature.xml
```diff
@@ -68,6 +68,7 @@
 		<bundle>mvn:org.openhab.core.bundles/org.openhab.core.io.rest.core/${project.version}</bundle>
 		<bundle>mvn:org.openhab.core.bundles/org.openhab.core.io.rest.sse/${project.version}</bundle>
 		<feature>openhab-core-config-discovery-addon</feature>
+		<bundle>mvn:org.openhab.core.bundles/org.openhab.core.config.discovery.sddp/${project.version}</bundle>
```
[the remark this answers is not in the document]
andrewfg: Ok.

features/karaf/openhab-core/src/main/feature/feature.xml
andrewfg: @mherwege I made the changes that you suggest. However that causes the build to fail. I have tried building it locally with about 20 different variations. I have no idea what I am doing. And I have no idea what I am doing wrong. I give up!!

andrewfg:
@mherwege I decided to push the latest commit with the feature build error so that you (and/or others) can see the actual error messages, and suggest how to fix it. Apart from this, the code is complete. As soon as anyone can fix the feature build error, then IMHO it is ready to merge.

features/karaf/openhab-core/src/main/feature/feature.xml
```diff
@@ -105,6 +105,19 @@
...
 
+	<feature name="openhab-core-config-discovery-sddp" version="${project.version}">
+		<feature>openhab-core-base</feature>
+		<bundle>mvn:org.openhab.core.bundles/org.openhab.core.config.discovery.sddp/${project.version}</bundle>
+	</feature>
+
+	<feature name="openhab-core-config-discovery-addon-sddp" version="${project.version}">
+		<feature>openhab-core-base</feature>
+		<feature>openhab-core-config-discovery-addon</feature>
+		<feature>openhab-runtime-base</feature>
```
mherwege, line 116: This line should be removed.

andrewfg: It doesn't make any difference.

andrewfg:
@mherwege I had originally coded the SddpAddonFinder to require a @reference to the SddpDiscoveryService. The purpose being that when the former is loaded, it would force the latter to be loaded too. I am now changing it so that both the SddpAddonFinder and SddpDiscoveryService are bundled into the same feature. I am hoping therefore that when that feature is loaded, both of those services will be loaded without the one forcing the load of the other.

mherwege:
> @mherwege I had originally coded the SddpAddonFinder to require a @reference to the SddpDiscoveryService. The purpose being that when the former is loaded, it would force the latter to be loaded too. I am now changing it so that both the SddpAddonFinder and SddpDiscoveryService are bundled into the same feature. I am hoping therefore that when that feature is loaded, both of those services will be loaded without the one forcing the load of the other.

I don't think that is the proper way. The sddp-addon feature can be enabled/disabled through the UI. That will break discovery in the bindings that use sddp discovery as well. These bindings should just have a feature dependency to the sddp discovery feature, not the discovery addon feature. So I think this should remain in 2 separate features.
I am still trying to figure this out myself, without success so far.

andrewfg:
> I don't think that is the proper way. 

I can confirm that this does not work either..

andrewfg:
I'm wondering if the problem could be due to the module name hierarchy? We have two services in `org.openhab.core.config.discovery.sddp` and `org.openhab.core.config.discovery.addon.sddp` that are trying (and failing) to cross reference each other. So I wonder if the cross reference needs to be via tag interfaces at a common higher level in the name hierarchy -- for example in `org.openhab.core.config.discovery`?

mherwege:
I am not sure you actually need to import and set this SddpDiscoveryService.class at all. It only needs to be part of your feature and started with the sddp addon finder feature. I am trying to remove it from the fields and constructor of SddpAddonFinder and making the org.openhab.core.config.discovery.sddp bundle part of the sddp addon finder feature as well. If you install the feature, it should then also be loaded. The relation between the 2 is through the participant interface.

EDIT: I got it to correctly pass feature verification when doing this. I am now doing a distro build to check if all works.

features/karaf/openhab-core/src/main/feature/feature.xml
```diff
@@ -438,6 +450,7 @@
 		<feature>openhab-core-automation-module-script</feature>
 		<feature>openhab-core-automation-module-script-rulesupport</feature>
 		<feature>openhab-core-automation-module-media</feature>
+		<feature>openhab-core-config-discovery-sddp</feature>
```
mherwege, line 453: I am pretty sure you can remove this. You just need to include the `openhab-core-config-discovery-sddp` feature in every binding that uses sddp discovery (as you now have in the `openhab-core-config-discovery-addon-sddp` feature).

bundles/org.openhab.core.config.discovery.addon.sddp/src/main/java/org/openhab/core/config/discovery/addon/sddp/SddpAddonFinder.java
```diff
@@ -0,0 +1,172 @@
...
+    private static final Set<String> SUPPORTED_PROPERTIES = Set.of(DRIVER, HOST, IP_ADDRESS, MAC_ADDRESS, MANUFACTURER,
+            MODEL, PORT, PRIMARY_PROXY, PROXIES, TYPE);
+
+    private final Logger logger = LoggerFactory.getLogger(SddpAddonFinder.class);
+    private final Set<SddpDevice> foundDevices = new HashSet<>();
+
+    private @Nullable SddpDiscoveryService sddpDiscoveryService = null;
+
+    @Activate
+    public SddpAddonFinder(
```
mherwege, line 86: I doubt this is going to work reliably when there is more than on DiscoveryService. The first one being loaded will get passed in as the reference, and if it is not an SddpDiscoveryService, it will just activate but without a reference. You probably need a separate bind method to set it.

mherwege:
My changes did indeed make it pass feature verification, but it will not work. You do indeed need to set the SddpDiscoveryService and register the participant, which was still missing in my version.
Very good you are getting much closer now!

features/karaf/openhab-core/src/main/feature/feature.xml
andrewfg: Makes sense.

bundles/org.openhab.core.config.discovery.addon.sddp/src/main/java/org/openhab/core/config/discovery/addon/sddp/SddpAddonFinder.java
andrewfg: That's the problem with the lack of documentation. My feeling is that all conditions must be met before the service can activate. I had MDNS operating at the time I made the test published above, and it did work with the SDDP service being loaded. But as you say, it may be random. I will run some more tests with the logger.warn() to see if I can figure it out.

andrewfg: It looks like the @reference is only satisfied when **_both_** the service class name and the service name are satisfied. In other words the service won't be activated until a specific instance of SddpDiscoveryService has been created and loaded.

andrewfg:
@mlobstein I am currently running a full build if this PR together with your https://github.com/openhab/openhab-addons/pull/16802 and https://github.com/openhab/openhab-addons/pull/16794 to see if it all works as expected. I assume that you are doing the same thing? If so, then please provide feedback. Note that I don't own a physical Epson Projector so in my tests I am using a test app to simulate the projector. Therefore your own tests would ultimately be more definitive than mine.

mlobstein:
> @mlobstein I am currently running a full build if this PR together with your [openhab/openhab-addons#16802](https://github.com/openhab/openhab-addons/pull/16802) and [openhab/openhab-addons#16794](https://github.com/openhab/openhab-addons/pull/16794) to see if it all works as expected. I assume that you are doing the same thing? If so, then please provide feedback. Note that I don't own a physical Epson Projector so in my tests I am using a test app to simulate the projector. Therefore your own tests would ultimately be more definitive than mine.

I just pushed a small fix to my branch. The SDDP discovery is working as expected. I also do not have a projector that supports SDDP. I am testing with my Pana UB-9000 by changing the strings in the DiscoveryParticipant. The device is discovered automatically and from an active scan. The discovered thing can be added successfully.
image

andrewfg:
@mlobstein the finder does also discover my 'fake' Epson device -- see below

image

bundles/org.openhab.core.config.discovery.addon.sddp/src/main/java/org/openhab/core/config/discovery/addon/sddp/SddpAddonFinder.java
andrewfg: > I doubt this is going to work reliably when there is more than on DiscoveryService. 

@mherwege you were right. The finder @Reference was being satisfied by _any_ DiscoveryService, so sometimes if the first loaded DiscoveryService was an SddpDiscoveryService it worked, and otherwise it did not. I changed it so that the finder @Reference is only satisfied for specific DiscoveryService instances with the `target` attribute matching a @component that has a `property` value of `"protocol=sddp"` .. and it now seems to work predictably..

andrewfg:
@mherwege I would like to give you particular thanks for your help in general concerning OSGI component instantiation and referencing, and in specific concerning the respective karaf features construction. :)

I also want to thank you for your work on OH WebUI concerning the addon finders (enablement UI, display UI), as well as the primary IP selector page. The workflow on new installs is now very smooth and intuitive. Many thanks.

andrewfg:
> Can you fix them @andrewfg?

@wborn Umm. I would be happy to do that. But I am not sure actually what I have to do.

wborn:
How about changing the `)` in a `}` ?

andrewfg:
@wborn https://github.com/openhab/openhab-core/pull/4265

Nadahar:
I'll just add that it's very hard to find much information about SDDP, and because of the similar spelling and functionality of SSDP, I've always just "read it as SSDP". Assuming that I've actually found the "right one", and that it is Control4's Simple Device Discovery Protocol, about the closest thing to a description I can find is [here](https://pypi.org/project/sddp-discovery-protocol/):

> Simple Device Discovery Protocol (SDDP) is a simple multicast discovery protocol implemented by many "smart home" devices to allow a controlling agent to easily discover and connect to devices on a local subnet.
> 
> SDDP was created by [Control4](https://www.control4.com/), a provider of whole-home automation systems. While it is quite similar to UPnP's standard Simple Service Discovery Protocol ([SSDP](https://en.wikipedia.org/wiki/Simple_Service_Discovery_Protocol)--note the similar but different spelling), and it serves a virtually identical purpose, SDDP is not a standard protocol and it is not publicly documented. The protocol as implemented by this package was inferred through observation of network traffic.

The only reason I looked twice at it now, was because I was wondering why there was both a UPnP and a (what I thought it was) SSDP discovery service, since that's basically the same thing.

My suggestion is that some context/description is added somewhere that makes it clear what it is. I doubt there are many people that actually know about SDDP or what it's used for.

andrewfg:
I don't think it matters who knows about it so long as it finds some addons for normal users. Which it does..

Nadahar:
> I don't think it matters who knows about it so long as it finds some addons for normal users. Which it does..

I think it does. Users have the option to enable and disable it in the configuration, how are they supposed to do that when they have no idea what it is or does? Just using Simple Device Discovery Protocol instead of just SDDP in the configuration description, also mentioning that it is supported by devices with Control4 integrations, would help even more. With that simple additional information, it would be easy enough for people to figure out what it is. 

As it is now, Google will "reroute" a search to SSDP, giving you a lot of wrong hits.

Likewise, I think it would be helpful that the same information was mentioned in the class JavaDoc header, just to avoid wasting people's time.

andrewfg:
If you feel strongly about it you could write it yourself.

Nadahar:
> If you feel strongly about it you could write it yourself.

#5031

PR #4245, opened 2024-05-24
Change icon for channel type `system.mute`

jlaur:
There is a classic icon `SoundVolume_Mute` available:
https://www.openhab.org/docs/configuration/iconsets/classic/

Icon is here:
https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui.iconset.classic/src/main/resources/icons/soundvolume_mute.svg

This is not used for channel type `system.mute` which currently uses the same icon as `system.volume`.


jlaur:
@openhab/core-maintainers, @lolodomo - this PR is also a question: Is there a good reason for using `SystemVolume` for the `system.mute` channel type? AFAIK there is no standardized correlation between `system.volume` and `system.mute` like there is for `system.color`, `system.brightness` and `system.switch` in the lighting domain.

lolodomo:
Change makes sense in my opinion.
I believe we should also add 2 new icons SoundVolume_Mute-on and SoundVolume_Mute-off. SoundVolume_Mute-on will be identical to SoundVolume_Mute and SoundVolume_Mute-off will be identical to SoundVolume.

lolodomo (approved):
LGTM

J-N-K (approved):
LGTM

PR #4247, opened 2024-05-25
Fixed a deadlock in the PoolBasedSequentialScheduledExecutorService

joerg1985:
I did some testing with the new `PoolBasedSequentialScheduledExecutorService` and ended in a deadlock.
The `ScheduledThreadPoolExecutor` used in the background does not spawn new threads as expected.

This PR will call `setCorePoolSize` to spawn new threads or to stop idle threads. 
Additionally this PR does allow a task to be directly executed in case the current threads is from the pool. 

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/common/PoolBasedSequentialScheduledExecutorService.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/common/PoolBasedSequentialScheduledExecutorService.java
@@ -18,6 +18,7 @@ import java.util.Collections;
 import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Set;
+import java.util.WeakHashMap;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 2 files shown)


joerg1985:
@kaikreuzer it would be great if this would be included in 4.2.M4

bundles/org.openhab.core/src/main/java/org/openhab/core/common/PoolBasedSequentialScheduledExecutorService.java
```diff
@@ -100,22 +120,24 @@ public PoolBasedSequentialScheduledExecutorService(ScheduledThreadPoolExecutor p
                     tail = empty;
                 }
             }
-        }, 2, 4, TimeUnit.SECONDS);
+        },
+                // avoid cleaners of promptly created instances to run at the same time
```
kaikreuzer, line 124: the indentation does not look right here.

kaikreuzer (commented):
The logic is getting pretty complex, but I hope it resolves the deadlock as expected.
Just a small formatting issue to be fixed before merging.

bundles/org.openhab.core/src/main/java/org/openhab/core/common/PoolBasedSequentialScheduledExecutorService.java
kaikreuzer: Weird - I guess this is some bug in the spotless formatter. Well, let's leave it as it is then!

PR #4255, opened 2024-05-28
[serial] Use a fast path to avoid pattern matching

joerg1985:
I was wondering why a fresh installed openhab instance without addons was creating ~ 1000 `int[]` instances every 15s.
This fallout is visible when monitoring the free memory of the instance:
image

This PR will add a fast path to avoid this fallout from matching a regex against dozens of not matching directories.

### DIFF
--- a/bundles/org.openhab.core.config.discovery.usbserial.linuxsysfs/src/main/java/org/openhab/core/config/discovery/usbserial/linuxsysfs/internal/SysfsUsbSerialScanner.java
+++ b/bundles/org.openhab.core.config.discovery.usbserial.linuxsysfs/src/main/java/org/openhab/core/config/discovery/usbserial/linuxsysfs/internal/SysfsUsbSerialScanner.java
@@ -222,7 +222,10 @@ public class SysfsUsbSerialScanner implements UsbSerialScanner {
      * of a USB device.
      */
     private @Nullable Path getUsbInterfaceParentPath(Path sysfsPath) {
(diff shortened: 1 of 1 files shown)


holgerfriedrich (commented):
Good catch, you patch avoids calling the more expensive pattern matching (at least as long as the pattern contains the ```-```.
If I get it correctly, you search for ```-``` many times: on the full path and all the parent path strings the way up, as ```getUsbInterfaceParentPath``` is called recursively. If you want to avoid that, maybe add a second helper function for the recursive calls, check for the shortcut only once, and then call the helper. WDYT?

joerg1985:
Yes there is room for improvments, but also the danger of over-engineering.
Usually the matching part is at the end and it does not need to go more than 3 parents up the tree, therefore i think it is okay to check every time.

holgerfriedrich (approved):

joerg1985:
Ps: in the negative case there is no `-` at all, so it will check only the full path once and return.

holgerfriedrich:
Agreed. Thank you.

wborn:
This probably also significantly reduces the CPU consumption by the serial scanner thread? It is the most CPU consuming thread on my system and I still wanted to look into how to reduce this. :slightly_smiling_face:

joerg1985:
@wborn the GC threads are taking a lot of time on my installation, so i am locking where the fallout is coming from and try to fix this. This seems to happen also in some libraries too, e.g. https://github.com/jmdns/jmdns/pull/275

Issue #4254, opened 2024-05-28
Persisted group items with base type Number:Dimension not restored correctly in persistence extensions


mherwege:
I have a rule retrieving data from persistence and doing calculations in it. I was able to isolate it and here is a js rule (with a working and failing example), and the result in the log:

```
var lastUpdate = time.toZDT('2024-01-01 00:00:00')
console.info('last update: ' + lastUpdate);
console.info('EVMeterstand state: ' + items.getItem('EVMeterstand')?.state);
console.info('EVMeterstand numeric state: ' + items.getItem('EVMeterstand')?.numericState);
console.info('EVMeterstand historic state: ' + items.getItem('EVMeterstand').persistence.persistedState(lastUpdate, 'influxdb')?.state);
console.info('EVMeterstand numeric historic state: ' + items.getItem('EVMeterstand').persistence.persistedState(lastUpdate, 'influxdb')?.numericState);
console.info('EVMeterstand delta calculated: ' + (Quantity(items.getItem('EVMeterstand')?.state).subtract(Quantity(items.getItem('EVMeterstand').persistence.persistedState(lastUpdate, 'influxdb')?.state))));
console.info('EVMeterstand numeric delta calculated: ' + (items.getItem('EVMeterstand')?.numericState - items.getItem('EVMeterstand').persistence.persistedState(lastUpdate, 'influxdb')?.numericState));
console.info('EVMeterstand delta: ' + items.getItem('EVMeterstand').persistence.deltaSince(lastUpdate, 'influxdb')?.state);
console.info('EVMeterstand numeric delta: ' + items.getItem('EVMeterstand').persistence.deltaSince(lastUpdate, 'influxdb')?.numericState);
console.info('PVEmeter state: ' + items.getItem('PVEmeter')?.state);
console.info('PVEmeter numeric state: ' + items.getItem('PVEmeter')?.numericState);
console.info('PVEmeter historic state: ' + items.getItem('PVEmeter').persistence.persistedState(lastUpdate, 'influxdb')?.state);
console.info('PVEmeter numeric historic state: ' + items.getItem('PVEmeter').persistence.persistedState(lastUpdate, 'influxdb')?.numericState);
console.info('PVEmeter delta calculated: ' + (Quantity(items.getItem('PVEmeter')?.state).subtract(Quantity(items.getItem('PVEmeter').persistence.persistedState(lastUpdate, 'influxdb')?.state))));
[...]

PR #4257, opened 2024-05-29
Fix persistence extensions for group items and number dimension items with different units

mherwege:
Closes https://github.com/openhab/openhab-core/issues/4254

Persistence extensions retrieving persisted data for GroupItems with a QuantityType as base type returned null or a value without the unit. This PR fixes that issue.
This was not an issue before https://github.com/openhab/openhab-core/pull/3736 as a DecimalType would be returned, ignoring the unit.

Second issue this resolves:
In the (rare) case a persistence service would have values stored with different units than the default (or configured) unit, the calculation would be off. This PR also corrects that. Most persistence services today assume the unit is the default (or configured) unit and don't store it explicitly. But it would be possible for some services.

### DIFF
--- a/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
+++ b/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
@@ -62,6 +62,7 @@ import org.slf4j.LoggerFactory;
  * @author Mark Herwege - Changed return types to State for some interval methods to also return unit
  * @author Mark Herwege - Extended for future dates
  * @author Mark Herwege - lastChange and nextChange methods
+ * @author mark Herwege - handle persisted GroupItem with QuantityType
  */
 @Component(immediate = true)
 @NonNullByDefault
@@ -981,10 +982,21 @@ public class PersistenceExtensions {
         Iterator<HistoricItem> it = result.iterator();
         HistoricItem maximumHistoricItem = null;
 
+        Item baseItem = item;
+        if (baseItem instanceof GroupItem groupItem) {
+            baseItem = groupItem.getBaseItem();
+        }
+        Unit<?> unit = (baseItem instanceof NumberItem numberItem) ? numberItem.getUnit() : null;
+
         DecimalType maximum = null;
         while (it.hasNext()) {
             HistoricItem historicItem = it.next();
-            DecimalType value = historicItem.getState().as(DecimalType.class);
+            State state = historicItem.getState();
+            if ((unit != null) && state instanceof QuantityType<?> qtState) {
+                qtState = qtState.toUnit(unit);
+                state = qtState != null ? qtState : state;
+            }
+            DecimalType value = state.as(DecimalType.class);
             if (value != null) {
                 if (maximum == null || value.compareTo(maximum) > 0) {
                     maximum = value;
@@ -1106,10 +1118,21 @@ public class PersistenceExtensions {
         Iterator<HistoricItem> it = result.iterator();
         HistoricItem minimumHistoricItem = null;
 
+        Item baseItem = item;
+        if (baseItem instanceof GroupItem groupItem) {
+            baseItem = groupItem.getBaseItem();
+        }
+        Unit<?> unit = (baseItem instanceof NumberItem numberItem) ? numberItem.getUnit() : null;
+
         DecimalType minimum = null;
         while (it.hasNext()) {
             HistoricItem historicItem = it.next();
-            DecimalType value = historicItem.getState().as(DecimalType.class);
+            State state = historicItem.getState();
+            if ((unit != null) && state instanceof QuantityType<?> qtState) {
+                qtState = qtState.toUnit(unit);
+                state = qtState != null ? qtState : state;
+            }
+            DecimalType value = state.as(DecimalType.class);
             if (value != null) {
                 if (minimum == null || value.compareTo(minimum) < 0) {
                     minimum = value;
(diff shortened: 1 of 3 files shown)


mherwege:
@J-N-K Could you have a look at this? It fixes an issue I discovered when deploying the enhanced peristence extensions to my production environment.

J-N-K (commented):
LGTM. What about a test for that?

mherwege:
I added a series of tests for a Group Item with Number:Dimension base Item.

mherwege:
Adding these tests and running through them made me realise I made a mistake with the treatment of QuantityType for HistoricItem. The individual persistence services don't return a QuantityType State, but a DecimalType State for Number Items (with or without dimension). So, when the Item has a dimension, the Unit needs to be added to the returned HistoricState in all cases, not just when doing calculations with the return values.
I previously made a wrong assumption also in the test service, where a QuantityType was constructed in the test service. That should not be done, as it is not what the persistence services return.

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
```diff
@@ -2542,9 +2543,16 @@ public State getState() {
...
+                        qtState = qtState.toUnit(unit);
+                        if (qtState != null) {
+                            state = qtState;
+                        }
+                    } else {
+                        DecimalType decimalState = state.as(DecimalType.class);
+                        if (decimalState != null) {
+                            state = new QuantityType<>(decimalState, unit);
+                        }
                     }
```
J-N-K: I'm really not sure about this one. The persistence service should return the correct type, i.e. a `QuantityType` if the state was originally a `QuantityType`. Which persistence service is not behaving like that?

mherwege: As far as I could tell from the code: JPA and MongoDB (for GroupItems).

J-N-K: IMO this is a bug and should be fixed in the add-ons.

mherwege: Even when the service returns QuantityType, in theory, the unit could be different for different datapoints. If one wants to calculate with it, it needs to be converted to a common unit. It probably does not happen at the moment (most don’t even store the unit and just get it from the item). But it could happen with e.g. inmemory persistence.

mherwege:
@J-N-K So I removed the automatic appliction of units when I get a DecimalType form the persistence service again. I agree the persistence service should be fixed.
However, it did bring me to adjusting the calculation logic. Most persistence services use the unit configured for the item (or default unit) when returning a QuantityType State. The calculations so far assumed that was the case. But this is not guaranteed. Actually, every persisted datapoint could have a different unit (but should be in the same dimension). The calculations now convert these to the configured (or default) unit before calculating with the decimal value.
This is a corner case, but a flaw in the logic that could be important in some corner cases.

mherwege:
I don't see the Java 21 build failure being related.

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
```diff
@@ -1105,10 +1118,21 @@ private static void internalPersist(Item item, TimeSeries timeSeries, @Nullable
...
         DecimalType minimum = null;
         while (it.hasNext()) {
             HistoricItem historicItem = it.next();
-            DecimalType value = historicItem.getState().as(DecimalType.class);
+            State state = historicItem.getState();
+            if ((unit != null) && state instanceof QuantityType<?> qtState) {
+                qtState = qtState.toUnit(unit);
+                state = qtState != null ? qtState : state;
+            }
+            DecimalType value = state.as(DecimalType.class);
```
J-N-K, line 1135: My understanding is that you get the unit from the item (or in case of a group from the base item). You then try to apply this unit to the state returned from the persistence service to ensure that all values have the correct unit. That is pretty logical and IMO correct. 

My problem is the next line (l. 1133): `.toUnit` returns `null` when a conversion is not possible. For example the persistence service returned a value of `5 m` but the item's unit is `s`. If that is the case, you fall back to the original (not converted value) and later extract the numeric only part of that state. IMO this is wrong, a value that can't be converted should be discarded instead. Maybe this is also a case where a warning should be logged, because that is for sure a severe configuration issue.

mherwege: @J-N-K You are of course right. I considered it but didn't think the chance of it happening was high (especially not with the current persistence services) and just avoided the null warning, but it is definitely better to skip and log a warning in the event that it would happen. I just inserted code for that.

mherwege:
I extracted the logic to check the retrieved persisted value, do the checks and the unit conversion in a separate method that can be reused, reducing code duplication.
There now also is a check if the retrieved value is a QuantityType when the item has a dimension, if not it logs a warning. That should make it easier to identify issues with persistence services and QuantityTypes.

mherwege:
@J-N-K I hope this can still make it in 4.2 to fix the two issues this solves:
- Persistence Actions do not work with persisted Group Items and QuantityTypes
- Errors in calculations when the persistence service returns QuantityTypes with units different from the Item unit (less likely to be hit at the moment, but still a bug)
I won't have access to my PC after today for a while, so will not be able to do further work on this until after the release. I hope it is in a good enough state to be accepted.

J-N-K:
Sure. This is a bugfix. I'll try to have a look this evening.

J-N-K (approved):
LGTM. We need to fix JPA and MongoDB before the release, I'll have a look.

mherwege:
Thanks.

PR #4262, opened 2024-05-31
Fix ip suggestion finder default parameter

mherwege:
The IP addon suggestion finder is on by default (like all other suggestion finders).

Because the default parameter was missing, it was wrongly shown in the UI.


wborn (approved):
Thanks!

PR #4267, opened 2024-06-05
Fix persist with date method

mherwege:
Evaluating https://github.com/openhab/openhab-js/issues/335, one bug was identified in the persistence extension when persisting with a date.

It is not sure it solves the whole issue, but it is a bug that should be fixed.


jlaur:
Thanks for the fix, @mherwege! I'm pretty sure this is causing the issue I'm having with JDBC persistence, since alias is not supported, thus state and timestamp not being considered in these overloads.

J-N-K (approved):
Thanks!

PR #4276, opened 2024-06-18
Relative functions for DateTimeType

clinique:
As from one scripting language to the other, dealing with Dates is not always so easy, I propose to add these small functions that I use very often while scripting. 

Here are the methods added to `DateTimeType` :

- `isBefore` / `isAfter` another `DateTimeType`
- `sameDay` compared to another `DateTimeType` or a `ZonedDateTime`
- `isToday` / `isTomorrow` / `isYesterday`
- `toToday`, `toTomorrow`, `toYesterday` changes day/month/year accordingly, preserving time part of the `DateTimeType`
- `isBeforeDate` / `isAfterDate` compared to another `DateTimeType` or a `ZonedDateTime` ignoring the time component
- `isBeforeTime` / `isAfterTime` compared to another `DateTimeType` or a `ZonedDateTime` ignoring the date component

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
@@ -20,6 +20,7 @@ import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
+import java.time.temporal.ChronoUnit;
 import java.time.zone.ZoneRulesException;
 import java.util.Locale;
 
@@ -37,6 +38,7 @@ import org.openhab.core.types.State;
  * @author Wouter Born - increase parsing and formatting precision
  * @author Laurent Garnier - added methods toLocaleZone and toZone
  * @author Gaël L'hopital - added ability to use second and milliseconds unix time
+ * @author Gaël L'hopital - added isToday, isTomorrow, isYesterday, sameDay
  */
 @NonNullByDefault
 public class DateTimeType implements PrimitiveType, State, Command {
@@ -247,5 +249,26 @@ public class DateTimeType implements PrimitiveType, State, Command {
         }
 
         return date;
+    }
+
+    public boolean isToday() {
+        return sameDay(ZonedDateTime.now());
+    }
+
+    public boolean isTomorrow() {
+        return sameDay(ZonedDateTime.now().plusDays(1));
+    }
+
+    public boolean isYesterday() {
+        return sameDay(ZonedDateTime.now().minusDays(1));
+    }
+
+    public boolean sameDay(DateTimeType other) {
+        return sameDay(other.zonedDateTime);
+    }
+
+    public boolean sameDay(ZonedDateTime other) {
+        return zonedDateTime.truncatedTo(ChronoUnit.DAYS)
+                .isEqual(other.withZoneSameLocal(zonedDateTime.getZone()).truncatedTo(ChronoUnit.DAYS));
     }
 }
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
@@ -20,7 +20,6 @@ import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
-import java.time.temporal.ChronoUnit;
 import java.time.zone.ZoneRulesException;
 import java.util.Locale;
 
@@ -38,7 +37,6 @@ import org.openhab.core.types.State;
  * @author Wouter Born - increase parsing and formatting precision
  * @author Laurent Garnier - added methods toLocaleZone and toZone
  * @author Gaël L'hopital - added ability to use second and milliseconds unix time
- * @author Gaël L'hopital - added isToday, isTomorrow, isYesterday, sameDay
  */
 @NonNullByDefault
 public class DateTimeType implements PrimitiveType, State, Command {
@@ -249,26 +247,5 @@ public class DateTimeType implements PrimitiveType, State, Command {
         }
 
         return date;
-    }
-
-    public boolean isToday() {
-        return sameDay(ZonedDateTime.now());
-    }
-
-    public boolean isTomorrow() {
-        return sameDay(ZonedDateTime.now().plusDays(1));
-    }
-
-    public boolean isYesterday() {
-        return sameDay(ZonedDateTime.now().minusDays(1));
-    }
-
-    public boolean sameDay(DateTimeType other) {
-        return sameDay(other.zonedDateTime);
-    }
-
-    public boolean sameDay(ZonedDateTime other) {
-        return zonedDateTime.truncatedTo(ChronoUnit.DAYS)
-                .isEqual(other.withZoneSameLocal(zonedDateTime.getZone()).truncatedTo(ChronoUnit.DAYS));
     }
 }
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
@@ -331,5 +331,4 @@ public class DateTimeType implements PrimitiveType, State, Command {
     public boolean isAfterDate(ZonedDateTime other) {
         return zonedDateTime.truncatedTo(ChronoUnit.DAYS).isAfter(other.truncatedTo(ChronoUnit.DAYS));
     }
-
 }
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
@@ -269,6 +269,67 @@ public class DateTimeType implements PrimitiveType, State, Command {
 
     public boolean sameDay(ZonedDateTime other) {
         return zonedDateTime.truncatedTo(ChronoUnit.DAYS)
-                .isEqual(other.withZoneSameLocal(zonedDateTime.getZone()).truncatedTo(ChronoUnit.DAYS));
+                .isEqual(other.withZoneSameInstant(zonedDateTime.getZone()).truncatedTo(ChronoUnit.DAYS));
     }
+
+    public DateTimeType toToday() {
+        return shiftDaysFromToday(0);
+    }
+
+    public DateTimeType toTomorrow() {
+        return shiftDaysFromToday(1);
+    }
+
+    public DateTimeType toYesterday() {
+        return shiftDaysFromToday(-1);
+    }
+
+    private DateTimeType shiftDaysFromToday(int days) {
+        ZonedDateTime now = ZonedDateTime.now().plusDays(days);
+        return new DateTimeType(zonedDateTime.withYear(now.getYear()).withMonth(now.getMonthValue())
+                .withDayOfMonth(now.getDayOfMonth()));
+    }
+
+    public boolean isBefore(DateTimeType other) {
+        return zonedDateTime.isBefore(other.zonedDateTime);
+    }
+
+    public boolean isAfter(DateTimeType other) {
+        return zonedDateTime.isAfter(other.zonedDateTime);
+    }
+
+    public boolean isBeforeDate(DateTimeType other) {
+        return isBeforeDate(other.zonedDateTime);
+    }
+
+    public boolean isBeforeDate(ZonedDateTime other) {
+        return zonedDateTime.truncatedTo(ChronoUnit.DAYS).isBefore(other.truncatedTo(ChronoUnit.DAYS));
+    }
+
+    public boolean isBeforeTime(DateTimeType other) {
+        return isBeforeTime(other.zonedDateTime);
+    }
+
+    public boolean isBeforeTime(ZonedDateTime other) {
+        return zonedDateTime.withYear(other.getYear()).withMonth(other.getMonthValue())
+                .withDayOfMonth(other.getDayOfMonth()).isBefore(other);
+    }
+
+    public boolean isAfterTime(DateTimeType other) {
+        return isAfterTime(other.zonedDateTime);
+    }
+
+    public boolean isAfterTime(ZonedDateTime other) {
+        return zonedDateTime.withYear(other.getYear()).withMonth(other.getMonthValue())
+                .withDayOfMonth(other.getDayOfMonth()).isAfter(other);
+    }
+
+    public boolean isAfterDate(DateTimeType other) {
+        return isAfterDate(other.zonedDateTime);
+    }
+
+    public boolean isAfterDate(ZonedDateTime other) {
+        return zonedDateTime.truncatedTo(ChronoUnit.DAYS).isAfter(other.truncatedTo(ChronoUnit.DAYS));
+    }
+
 }
--- a/bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/DateTimeTypeTest.java
+++ b/bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/DateTimeTypeTest.java
@@ -43,7 +43,6 @@ import org.junit.jupiter.params.provider.MethodSource;
  * @author Erdoan Hadzhiyusein - Added ZonedDateTime tests
  * @author Laurent Garnier - Enhanced tests
  * @author Gaël L'hopital - added ability to use second and milliseconds unix time
- * @author Gaël L'hopital - added isToday, isTomorrow, isYesterday tests
  */
 @NonNullByDefault
 public class DateTimeTypeTest {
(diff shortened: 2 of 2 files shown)


rkoshak:
Given that some of the languages do not expose and have you work with the raw Java DateTimeType Objects adding these will not make them available to all rules languages.

Some of the stuff like this though has been added to JS Scripting already so I think this PR could be more ambitious if youi wanted to be. There you will find:

- `time.toZDT()`: will convert just about anything to a ZonedDateTime that can be converted to a ZonedDateTime including DateTime Items, Number Items (state is treated as the number of milliseconds from now), Number:Time Items (state is treated as the amount of time from now), ISO8601 Strings, Java DateTime Strings, ISO8601 duration Strings (e.g. PT5M30S, treated as the amount of time from now), a String representing a time of day in 12hr or 24hr format, etc.

- `myZDT.toToday()`: returns the ZDT with today's date, adjusting for DST changeovers as necessary

- `isBeforeTime()`, `isBeforeDate()`, `isBeforeDateTime()`, the `isAfter` equivalents, and `isBetweenTimes()` `isBetweenDates()`, and `isBetweenDateTimes()` which lets you compare ZDTs to see where they fall compared to other ZDTs.

- `isClose()` which is a fuzzy comparison to see if a ZDT is within a certain amount of time of another.

So you can do stuff like this:

```
// Is it currently between 8AM and 4PM
time.toZDT().isBetweenTimes('8:00AM', `4:00PM`);

// Is now between the times in two Items?
time.toZDT().isBetweenTimes(items.StartTime, items.EndTime); // dates are ignored

// is now between the start time in the Item and three hours after the start time?
time.toZDT().isBetweenTimes(items.StartTime, time.toZDT(items.StartTime).plusHours(3));

// Is the Item's state today? (there are others)
time.toZDT(items.MyDateTime).isBetweenDateTimes('00:00', '23:59:59');

// Move the Item's date time to today
time.toZDT(items.MyDateTime).toToday();

// Is the Item's state tomorrow?
time.toZDT(items.MyDateTime).toLocalDate().equals(time.toZDT('P1DT').toLocalDate());

// Move an Item's time to today
time.toZDT(items.MyDateTime).toToday();
```

Note that `time.toZDT()` returns `now`.

The real magic (IMO) with the JS Scripting approach is that in many of these conversion and comparison functions you can use anything that can be converted by `time.toZDT()`. It really gets interesting when stuff that usually takes a DateTime can accept anything that can be converted by `toZDT()`. For example, creating a Timer to go off in five minutes:

    actions.ScriptExecution.createTimer(Quantity('5 M'), () => { do something }); // note this doesn't currently work because createTimer doesn't support this sort of conversion

In addition to `toToday()`, my OHRT also implements a `toTomorrow()` and `toYesterday()` method. 

In JS you can monkeypatch these into the ZonedDateTime class but I'm pretty sure that's not an option on the Java side of things. But a series of utility functions on the DateTimeType to not only compare but also convert "stuff" to ZonedDateTimes. 

There is also a lot of convenience stuff built into jRuby which might prove some inspiration as well. 

Once I implemented a lot of the above (originally as part of OHRT and then I was able to get most of them accepted into the openhab-js library) all the complexities of working with DateTimes kind of just disappeared. `time.toZDT()` normalizes everything and once normalized comparisons and manipulation becomes standardized and simple. And because `time.toZDT()` treats anything that can be interpreted as a time duration and how much time to add to `now`, and the main comparison functions can accept one of these durations instead of a ZDT itself, I pretty much never have to do a string of `plusSeconds().plusMinutes().plusHours()` type stuff nor `withHour().withMinute().withSecond()` type stuff.  The resultant code becomes *much* clearer and easier to read. 

It's just an idea and may be more than you're willing to take on, but I can say this has greatly simplified JS Scripting rules.

bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/DateTimeTypeTest.java
```diff
@@ -290,6 +291,21 @@ public void epochTest() {
         assertThat(epochStandard, is(zdtStandard));
     }
 
+    @Test
+    public void relativeTest() {
+        DateTimeType dt1 = new DateTimeType("2019-06-12T17:30:00Z");
+        DateTimeType dt2 = new DateTimeType("2019-06-12T00:00:00+0000");
```
jimtng, line 297: To ensure that this checks for the same instant:

dt1 = "2019-06-13T01:00:00+02"
dt2 = "2019-06-12T23:00:00Z"

jimtng (commented):
I think this is a handy addition, especially for rulesdsl. There is currently not a direct equivalent in jruby for what this adds, and when this is added, it will be automatically usable from within jruby without any modifications to the jruby helper library. However it's currently already quite easy to do this in jruby

```ruby
# dt1 and dt2 are DateTimeType objects, e.g. A_DateTime_Item.state

# Checking for same day
dt1.to_date == dt2.to_date # to_date creates a Date object, stripping off the time data

# to check for tomorrow
dt1.to_date == Date.today + 1 

# to check for yesterday
dt1.to_date == Date.today - 1 
```

After this PR, JRuby will be able to call:
```ruby
dt1.today? #jruby automatically translates isToday to ruby method naming convention `today?`

dt1.yesterday?
dt1.tomorrow?
dt1.same_day(dt2) # The Ruby way would be dt1 === dt2
```

However, in jruby we can add these to ZonedDateTime directly, thus making it universally usable in many more places, including DateTimeType, because in jruby a DateTimeType delegates its methods to its ZonedDateTime value.

@ccutrer wdyt?

bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/DateTimeTypeTest.java
```diff
@@ -290,6 +291,21 @@ public void epochTest() {
...
+
+        assertTrue(dt1.sameDay(dt2));
+        assertTrue(new DateTimeType().isToday());
+
+        DateTimeType now = new DateTimeType();
+        DateTimeType tomorrow = new DateTimeType(now.getZonedDateTime().plusHours(24));
```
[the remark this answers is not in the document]
clinique: I wanted to avoid using the same method than what's used in the object itself.

bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/DateTimeTypeTest.java
```diff
@@ -290,6 +291,21 @@ public void epochTest() {
...
+        DateTimeType yesterday = new DateTimeType(now.getZonedDateTime().minusHours(24));
```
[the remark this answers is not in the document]
clinique: Changed

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
```diff
@@ -248,4 +250,25 @@ private ZonedDateTime parse(String value) throws DateTimeParseException {
...
+        return sameDay(ZonedDateTime.now().minusDays(1));
+    }
+
+    public boolean sameDay(DateTimeType other) {
+        return sameDay(other.zonedDateTime);
+    }
+
+    public boolean sameDay(ZonedDateTime other) {
+        return zonedDateTime.truncatedTo(ChronoUnit.DAYS)
+                .isEqual(other.withZoneSameLocal(zonedDateTime.getZone()).truncatedTo(ChronoUnit.DAYS));
```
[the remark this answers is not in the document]
clinique: Changed

bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/DateTimeTypeTest.java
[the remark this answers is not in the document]
clinique: Changed

bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/DateTimeTypeTest.java
clinique: Changed

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
```diff
@@ -248,4 +250,85 @@ private ZonedDateTime parse(String value) throws DateTimeParseException {
...
+        return sameDay(ZonedDateTime.now().minusDays(1));
+    }
+
+    public boolean sameDay(DateTimeType other) {
+        return sameDay(other.zonedDateTime);
+    }
+
+    public boolean sameDay(ZonedDateTime other) {
+        return zonedDateTime.truncatedTo(ChronoUnit.DAYS)
+                .isEqual(other.withZoneSameInstant(zonedDateTime.getZone()).truncatedTo(ChronoUnit.DAYS));
```
jimtng, line 272:
```suggestion
        return zonedDateTime.truncatedTo(ChronoUnit.DAYS).isEqual(other.truncatedTo(ChronoUnit.DAYS));
```

isEqual already compares the instant so it's not necessary to equalise the timezone prior to comparison.

```
jshell> var d1 = java.time.ZonedDateTime.parse("2024-01-01T00:00:00Z")
d1 ==> 2024-01-01T00:00Z

jshell> var d2 = java.time.ZonedDateTime.parse("2024-01-01T10:00:00+10:00")
d2 ==> 2024-01-01T10:00+10:00

jshell> d1.isEqual(d2)
$17 ==> true

jshell> d1.getZone()
$18 ==> Z

jshell> d2.getZone()
$19 ==> +10:00
```

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
```diff
@@ -248,4 +250,85 @@ private ZonedDateTime parse(String value) throws DateTimeParseException {
...
+    }
+
+    public boolean isBefore(DateTimeType other) {
+        return zonedDateTime.isBefore(other.zonedDateTime);
+    }
+
+    public boolean isAfter(DateTimeType other) {
+        return zonedDateTime.isAfter(other.zonedDateTime);
+    }
+
```
jimtng, line 300: We might also need the versions of isBefore and isAfter that accept a ZonedDateTime

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
```diff
@@ -248,4 +250,85 @@ private ZonedDateTime parse(String value) throws DateTimeParseException {
...
+    public boolean isBeforeDate(DateTimeType other) {
+        return isBeforeDate(other.zonedDateTime);
+    }
+
+    public boolean isBeforeDate(ZonedDateTime other) {
+        return zonedDateTime.truncatedTo(ChronoUnit.DAYS).isBefore(other.truncatedTo(ChronoUnit.DAYS));
+    }
+
```
jimtng, line 308: I am against the introduction of isBeforeDate, isBeforeTime, etc. because their meaning is ambiguous depending on the time zones of the two objects involved.

Take two zdt objects containing the same date but within two different time zones. They would fail the given comparison.

```
jshell> var d1 = java.time.ZonedDateTime.parse("2024-01-01T00:00:00-11:00")
d1 ==> 2024-01-01T00:00-11:00

jshell> var d2 = java.time.ZonedDateTime.parse("2024-01-01T10:00:00+11:00")
d2 ==> 2024-01-01T10:00+11:00

jshell> d2.withZoneSameInstant(d1.getZone())
$22 ==> 2023-12-31T12:00-11:00

jshell> d1.truncatedTo(java.time.temporal.ChronoUnit.DAYS).isBefore(d2.truncatedTo(java.time.temporal.ChronoUnit.DAYS))
$23 ==> false
```

jimtng (changes requested):

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
```diff
@@ -248,4 +250,85 @@ private ZonedDateTime parse(String value) throws DateTimeParseException {
...
+    public DateTimeType toYesterday() {
+        return shiftDaysFromToday(-1);
+    }
+
+    private DateTimeType shiftDaysFromToday(int days) {
+        ZonedDateTime now = ZonedDateTime.now().plusDays(days);
+        return new DateTimeType(zonedDateTime.withYear(now.getYear()).withMonth(now.getMonthValue())
+                .withDayOfMonth(now.getDayOfMonth()));
+    }
+
```
jimtng, line 292: This also has the same ambiguity issue related to differences in time zones between what's in `zonedDateTime` vs the system time zone.

```
jshell> var d1 = java.time.ZonedDateTime.parse("2024-01-01T00:00:00+11:00")
d1 ==> 2024-01-01T00:00+11:00

jshell> var now = java.time.ZonedDateTime.parse("2025-05-01T00:00:00-11:00")
now ==> 2025-05-01T00:00-11:00

jshell> d1.withYear(now.getYear()).withMonth(now.getMonthValue())
   ...>                 .withDayOfMonth(now.getDayOfMonth()).withZoneSameInstant(now.getZone())
$38 ==> 2025-04-30T02:00-11:00
```

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
clinique: These already [exists](https://docs.oracle.com/javase/8/docs/api/java/time/chrono/ChronoZonedDateTime.html#isBefore-java.time.chrono.ChronoZonedDateTime-), reason why I thought it wasn't needed.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
clinique: So do you recommend I should align Timezones before comparison, rename the methods or drop them ?

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
jimtng: Basically it's for convenience, so you can do DateTimeItem.state.isBefore(zdt) instead of DateTimeItem.state.getZonedDateTime().isBefore(zdt), and also for uniformity since you are adding the ability to do 
DateTimeItem.state.isBefore(anotherDateTimeItem), but, I'll leave the decision whether to have it to the maintainers.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
jimtng: I would prefer to drop them, but that's just an opinion. Again, I defer to the maintainers to decide.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
clinique: Changed

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
clinique: Added

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
clinique: Made it relative to `zonedDateTime` zone's.

jlaur:
This PR will be obsoleted by #3583 if accepted.

jlaur:
@clinique - I don't think this is meaningful anymore since #3583 was merged (sorry).

The reason is that `DateTimeType` doesn't contain a time-zone anymore, so it doesn't have a concept of a "day", midnight, etc. Of course `ZoneId` could be added to all those methods, so that `LocalDate`/`LocalDateTime` could be derived and everything could still work. But I'm not sure `DateTimeType` is really the right place anymore, since the operations are in reality performed on `LocalDateTime`, and `DateTimeType` is `Instant`. In other words, the operations would be not closely related, so it seems a bit like bloat IMHO.

clinique:
> @clinique - I don't think this is meaningful anymore since #3583 was merged (sorry).
> 
> The reason is that `DateTimeType` doesn't contain a time-zone anymore, so it doesn't have a concept of a "day", midnight, etc. Of course `ZoneId` could be added to all those methods, so that `LocalDate`/`LocalDateTime` could be derived and everything could still work. But I'm not sure `DateTimeType` is really the right place anymore, since the operations are in reality performed on `LocalDateTime`, and `DateTimeType` is `Instant`. In other words, the operations would be not closely related, so it seems a bit like bloat IMHO.

This means that only `isBefore` and `isAfter` are still relevant. Is it worth changing the PR or close it as is ?

jlaur:
> This means that only `isBefore` and `isAfter` are still relevant. Is it worth changing the PR or close it as is ?

Ah, right. If you think it's worth to add those, I won't stop you. 🙂 Perhaps you can provide an example usage, because I can't seem to find those in the PR description. If it's for rule languages other than DSL, e.g. JavaScript, I would imagine you'd like to do something like this:
```javascript
if (items.MyDateTime1.isAfter(items.MyDateTime2)) {
    // ...
}
```

rather than:

```javascript
if (time.toZDT(items.MyDateTime1).isAfter(time.toZDT(items.MyDateTime2)) {
    // ...
}
```

?

But when having only one `DateTimeType` and a `ZonedDateTime`, you would still have to use this form:
```javascript
if (time.toZDT(items.MyDateTime1).isAfter(time.toZDT()) {
    // ...
}
```

or construct a new `DateTimeType`, but that would probably be more cumbersome?

rkoshak:
JS uses the joda-js library for time, not the java.time classes. And it wraps all all the Item classes with a JS class. this means that this PR won't really change how JS Scripting works.

If you are looking at useful methods for comparisons, JS Scripting adds a few to ZDT monkey patching, notably:
- `isBeforeTime()`/`isAfterTime` : only compares the time parts
- `isBetweenTimes()`: checks if this ZDT is between the two passed in values
- the above but for date only comparisons
- `isClose()`: true if this ZDT is within the passed in Duration of the passed in ZDT, great for fuzzy comparisons

These monkey patched methods can accept anything supported by `time.toZDT()` so you can do something like

    if(time.toZDT().isClose("08:00",time.Duration.ofMinutes(5))) // true if now is within 5 minutes of 8 am.

But if you are working with a DateTime Item, you must first convert that Item, or it's state, to a joda-js ZDT using `time.toZDT()`. For that to chang it would have to be changed in the openhab-js repo, not here. Changes here should only impact those languages that use the raw Java classes like Rules DSL, Groovy, etc.

ccutrer:
Ditto JRuby - we have extensive wrappers and monkey patching of similar convenience methods. If some of these get added to core, we'll just delegate to them instead of implementing ourselves, but it won't change the end user interface.

clinique:
Then I leave this as is. Thanks all.

PR #4278, opened 2024-06-22
[voice] Avoid text concatenation on speech recognition errors

GiviMAD:
<s>I would like to add this fix to the dialog processor, which avoid playing the content of the SpeechRecognitionErrorEvent unless it's explicitly indicated on it.

I think is more friendly to just fail and don't continue the dialog than to "speak" a text that can be a forwarded exception message, so I changed the log for the SpeechRecognitionErrorEvent and added the "notify" field on the SpeechRecognitionErrorEvent class which is false by default.</s>

I would like to add this fix to the dialog processor, which avoids the text concatenation on the content of the SpeechRecognitionErrorEvent and also allows to emit silent errors by passing an empty message.

### DIFF
--- a/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
+++ b/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
@@ -362,12 +362,14 @@ public class DialogProcessor implements KSListener, STTListener {
             logger.debug("RecognitionStopEvent event received");
             toggleProcessing(false);
         } else if (sttEvent instanceof SpeechRecognitionErrorEvent sre) {
-            logger.debug("SpeechRecognitionErrorEvent event received");
+            String message = sre.getMessage();
+            logger.debug("SpeechRecognitionErrorEvent event received: {}", message);
             if (!isSTTServerAborting) {
                 abortSTT();
                 toggleProcessing(false);
-                String text = i18nProvider.getText(bundle, "error.stt-error", null, dialogContext.locale());
-                say(text == null ? sre.getMessage() : text.replace("{0}", sre.getMessage()));
+                if (!message.isEmpty()) {
+                    say(message);
+                }
             }
         }
     }
--- a/bundles/org.openhab.core.voice/src/main/resources/OH-INF/i18n/voice.properties
+++ b/bundles/org.openhab.core.voice/src/main/resources/OH-INF/i18n/voice.properties
@@ -25,6 +25,5 @@ system.config.voice.maxTextLengthCacheTTS.label = TTS Cache Maximum Text Length
 system.config.voice.maxTextLengthCacheTTS.description = The maximum length of texts handled by the TTS cache (in character). If exceeded, will pass the text to the TTS without storing it. 0 for no limit.
 
 error.ks-error = Encountered error while spotting keywords, {0}
-error.stt-error = Encountered error while recognizing text, {0}
 error.stt-exception = Error during recognition, {0}
 service.system.voice.label = Voice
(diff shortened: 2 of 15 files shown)


GiviMAD:
@lolodomo, if you can give me your opinion on these changes, It's related to the review comment https://github.com/openhab/openhab-addons/pull/15166#discussion_r1645487632

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/SpeechRecognitionErrorEvent.java
```diff
@@ -43,4 +59,13 @@ public SpeechRecognitionErrorEvent(String message) {
     public String getMessage() {
         return this.message;
     }
+
+    /**
+     * Should notify the user
+     *
+     * @return True if user should be notified
+     */
+    public Boolean getNotify() {
```
lolodomo, line 68: boolean rather than Boolean

lolodomo (changes requested):
With these changes, you remove the "saying" of the error message with all the existing SST services. Is it really a good idea to say nothing ?

GiviMAD:
> With these changes, you remove the "saying" of the error message with all the existing SST services. Is it really a good idea to say nothing ?

I'm not sure about it, I remember to have an issue with an exception message coming through the speaker, and I think that is something that shouldn't happen, but maybe it's more correct to fix that on the specific binding instead of changing the behavior.

But I think that in some cases like for an empty transcription, you don't want to say anything, because it's probably the result of an incorrect spot event and better to not introduce extra noise telling something and disturb the user, other systems seem to work like that.

I can set notify to true by default if you think is a better idea.

lolodomo:
As I suggested in my last comment in the review of the whisper STT service, why not letting the user decide if he wants or not a spoken feedback ? If the message is blank, nothing is said. WDYT ?

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
```diff
@@ -362,12 +362,14 @@ public synchronized void sttEventReceived(STTEvent sttEvent) {
...
-            logger.debug("SpeechRecognitionErrorEvent event received");
+            logger.warn("SpeechRecognitionErrorEvent event received: {}", sre.getMessage());
             if (!isSTTServerAborting) {
                 abortSTT();
                 toggleProcessing(false);
-                String text = i18nProvider.getText(bundle, "error.stt-error", null, dialogContext.locale());
-                say(text == null ? sre.getMessage() : text.replace("{0}", sre.getMessage()));
+                if (sre.getNotify()) {
+                    String text = i18nProvider.getText(bundle, "error.stt-error", null, dialogContext.locale());
+                    say(text == null ? sre.getMessage() : text.replace("{0}", sre.getMessage()));
```
lolodomo, line 371: Maybe your problem was the injection of the message coming from `SpeechRecognitionErrorEvent`  inside a more global message defined by the entry "error.stt-error" ?
Maybe you should say simply the message coming from `SpeechRecognitionErrorEvent` ?

bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/DialogProcessor.java
```diff
@@ -362,12 +362,14 @@ public synchronized void sttEventReceived(STTEvent sttEvent) {
...
+            String message = sre.getMessage();
+            logger.debug("SpeechRecognitionErrorEvent event received: {}", message);
             if (!isSTTServerAborting) {
                 abortSTT();
                 toggleProcessing(false);
-                String text = i18nProvider.getText(bundle, "error.stt-error", null, dialogContext.locale());
-                say(text == null ? sre.getMessage() : text.replace("{0}", sre.getMessage()));
+                if (!message.isEmpty()) {
+                    say(message);
+                }
```
lolodomo, line 372: In that case, the message "error.stt-error" is never used and can be removed.

Please also check that all existing STT services are appropriate with this change.

lolodomo (approved):
LGTM

kaikreuzer (approved):
Thanks for the review, @lolodomo.

PR #4279, opened 2024-06-23
Extend Stringutils with padRight

lsiepel:
Earlier some string utils where added to core in order to remove the dependency on `org.apache.commons.lang3.StringUtils`.
While doing so i create a `padLeft`, but forgot a `padRight` method. This PR corrects that.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/util/StringUtils.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/util/StringUtils.java
@@ -15,6 +15,7 @@ package org.openhab.core.util;
 import java.security.SecureRandom;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
@@ -188,7 +189,7 @@ public class StringUtils {
      * @return the padded String
      */
     public static String padLeft(@Nullable String str, int minSize, String padString) {
-        String paddedString = str == null ? "" : str;
+        String paddedString = Objects.requireNonNullElse(str, "");
         return String.format("%" + minSize + "s", paddedString).replace(" ", padString);
     }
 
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core/src/main/java/org/openhab/core/util/StringUtils.java
```diff
@@ -192,6 +192,25 @@ public static String padLeft(@Nullable String str, int minSize, String padString
...
+      * padRight("openHAB", 4, "*")  => "openHAB"
+     * </pre>
+     *
+     * @param str the String to pad, may be null
+     * @param minSize the minimum String size to return
+     * @param padString the String to add when padding
+     * @return the padded String
+     */
+    public static String padRight(@Nullable String str, int minSize, String padString) {
+        String paddedString = str == null ? "" : str;
```
J-N-K, line 210:
```suggestion
        String paddedString = Objects.requireNonNullElse(str, "");
```

J-N-K (changes requested):

holgerfriedrich (commented):
I hope I do not get this wrong, but I there are two observations (for edge cases):
- It seems you replace all occurrences of ```" "``` in the original String as well.
- You replace a single character ```" "``` possibly by a multi-character String, which might result in the String to be longer than minSize.

Both might be unexpected for the user.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/StringUtils.java
lsiepel: Fixed

lsiepel:
> I hope I do not get this wrong, but I think there are two observations (for edge cases):
> 
> * It seems you replace all occurrences of `" "` in the original String as well.
> * You replace a single character `" "` possibly by a multi-character String, which might result in the String to be longer than minSize.
> 
> Both might be unexpected for the user.

Agree that might be unexpected, i just changed the implementation a bit, support for strings with whitespaces is added now.

holgerfriedrich (approved):
LGTM, thanks

J-N-K (approved):
Thanks!

PR #4281, opened 2024-06-23
[voice] Add hlis argument to interpret command

GiviMAD:
Allow define interpreters to use on the console command `voice interpret`.

Added param --hlis which allows a comma separated list of interpreter ids.


kaikreuzer (approved):
lgtm, thanks!

PR #4288, opened 2024-06-30
Ensure the PoolBasedSequentialScheduledExecutorService does keep a minimum size

joerg1985:
This PR will ensure the `PoolBasedSequentialScheduledExecutorService` does keep the configured minimum size.
The unit test does now only need to spawn 10 new threads instead of up to 20, this should fix the latency issue.

@holgerfriedrich could you please run the CI multiple times for this PR to see if it does fix the issue?

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/common/PoolBasedSequentialScheduledExecutorService.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/common/PoolBasedSequentialScheduledExecutorService.java
@@ -18,7 +18,6 @@ import java.util.Collections;
 import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


joerg1985:
@holgerfriedrich The issue is still present, now i have updated the timeout to see if this helps.

holgerfriedrich:
The CI seems to work better with the change - so far I could tell based on the few rebuilds I triggered. Jenkins fails seem to be unrelated.
@kaikreuzer Could you please have a look at this, AFAIR you reviewed and merged #4247. Thanks!

joerg1985:
@holgerfriedrich i could split this into the timeout change of the unit test and the other change. This would take the pressure out of the review. What do you think?

kaikreuzer (approved):
No need to split the PR, let's merge it together - thanks!

PR #4317, opened 2024-07-14
Add unitHint to AbstractStorageBasedTypeProvider.ChannelTypeEntity

### LINKED ISSUE #4315: AbstractStorageBasedTypeProvider doesn't store unitHint
## Expected Behavior

When using the AbstractStorageBasedTypeProvider to store channel types, all fields should be stored.

## Current Behavior

unitHint is missing from the AbstractStorageBasedTypeProvider.ChannelTypeEntity subclass, so it gets lost when using this class to store channel types. Root cause is that when #4079 was merged, this class wasn't updated to stay in sync.

## Possible Solution

Simple solution: Add unitHint field to AbstractStorageBasedTypeProvider.ChannelTypeEntity and make sure it's populated in `mapToEntity(ChannelType channelType)` and read in `mapFromEntity(ChannelTypeEntity entity)`.

Better solution (?): It seems unfortunate that the AbstractStorageBasedTypeProvider class uses it's own mapping classes. Could it be possible to use the standard DTO classes instead? That would reduce the risk of missing updating this class if adding/changing parameter on types in the future. Not sure how feasible this solution would be, perhaps there's a reason for implementing it the way it has been done that I don't see.

1. Use AbstractStorageBasedTypeProvider as base class for a ChannelTypeProvider
2. Add a ChannelType that includes a unitHint.
3. Read the ChannelType back and try to get the unitHint. It will be `null`.

## Your Environment

* Version used: 4.2.0
* Environment name and version: Zulu java 17
* Operating System and version: Windows 11

pacive:
#4079 added a unitHint property to ChannelTypes, but the AbstractStorageBasedTypeProvider wasn't updated to reflect the change. This adds this property so it can be persisted when adding ChannelTypes dynamically.

Fixes #4315

### DIFF
--- a/bundles/org.openhab.core.thing/src/test/java/org/openhab/core/thing/binding/AbstractStorageBasedTypeProviderTest.java
+++ b/bundles/org.openhab.core.thing/src/test/java/org/openhab/core/thing/binding/AbstractStorageBasedTypeProviderTest.java
@@ -76,6 +76,37 @@ public class AbstractStorageBasedTypeProviderTest {
     }
 
     @Test
+    public void testQuantityChannelTypeProperlyMappedToEntityAndBack() {
+        ChannelTypeUID channelTypeUID = new ChannelTypeUID("TestBinding:testQuantityChannelType");
+
+        ChannelType expected = ChannelTypeBuilder.state(channelTypeUID, "testLabel", "Number:Length")
+                .withDescription("testDescription").withCategory("testCategory")
+                .withConfigDescriptionURI(URI.create("testBinding:testConfig"))
+                .withAutoUpdatePolicy(AutoUpdatePolicy.VETO).isAdvanced(true).withTag("testTag")
+                .withCommandDescription(CommandDescriptionBuilder.create().build())
+                .withStateDescriptionFragment(StateDescriptionFragmentBuilder.create().build()).withUnitHint("km")
+                .build();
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.thing/src/test/java/org/openhab/core/thing/binding/AbstractStorageBasedTypeProviderTest.java
```diff
@@ -75,6 +75,37 @@ public void testStateChannelTypeProperlyMappedToEntityAndBack() {
         assertThat(actual.getTags(), hasItems(expected.getTags().toArray(String[]::new)));
     }
 
+    @Test
```
J-N-K, line 78: I don't think we need an extra test here. You can adjust the "old" test to a `Number:Length` with a unit of `km`. Otherwise LGTM.

J-N-K (changes requested):

bundles/org.openhab.core.thing/src/test/java/org/openhab/core/thing/binding/AbstractStorageBasedTypeProviderTest.java
pacive: Fixed

J-N-K (approved):
LGTM, Thanks!

PR #4325, opened 2024-07-22
Allow ProfileTypeBuilder to pass its three filters list to both implementations of ProfileType

### LINKED ISSUE #4293: ProfileTypeBuilder.withSupportedChannelTypeUIDs is ignored by StateProfileTypeImpl
When writing an addon that provides a transformation profile one will most likely use `ProfileTypeBuilder` to get an instance of `ProfileType` to be returned by the `ProfileTypeProvider.getProfileTypes` implementation.

This is can be done with code like this:

```java
            return ProfileTypeBuilder
                    .newState(uid, "Label")
                    .build();
```

With the code above, the transformation profile is usable for any item and any channel, which may not be desirable and so one may want to call `withSupportedChannelTypeUIDs` to limit the transformation profile to a given channel type.

## Expected Behavior
When `withSupportedChannelTypeUIDs` has been called on `ProfileTypeBuilder`, the created transformation profile should only appear on item links that are between any item and a channel only of the given type UID.

## Current Behavior
Sadly, the profile above is shown for links on any channel, there are no filtering done.
This is because `ProfileTypeBuilder.newState` sets up a callback that ignores `leSupportedChannelTypeUIDs` at this [line](https://github.com/openhab/openhab-core/blob/37e09f5ab8760d9a3a57348d2bb558efc238f535/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/profiles/ProfileTypeBuilder.java#L66)

## Possible Solution
I would suggest changing `StateProfileTypeImpl` to accept `supportedChannelTypeUIDs` in its constructor and override `getSupportedChannelTypeUIDs` just like `TriggerProfileTypeImpl` does.
Note that this would only fix the current issue, but I believe both `StateProfileTypeImpl` and `TriggerProfileTypeImpl should override the three possible filtering sources, not two each like now.

1. Create an addon that implements a `ProfileTypeProvider` service
2. Use `ProfileTypeBuilder.newState` alongside `withSupportedChannelTypeUIDs`
3. See that the transformation profile is not filtered out and available on all channel links

## Context
I'm currently trying to offer transformation profiles for Air Quality Indicator channels in my OpenMeteo binding and would like them to be limited to channels they apply to.

## Your Environment
* Version used: openHAB 4.2 snapshot
* Environment name and version : Java 17
* Operating System and version : Windows 10-11

obones:
This pull request moves getSupportedChannelTypeUIDs and getSupportedItemTypesOfChannel to the ProfileType interface so that StateProfileType and TriggerProfileType can both take the three filter lists given by ProfileTypeBuilder

It fixes #4293 and will allow bundle writers to limit their transformation profiles to just the channels/items that are related to their profiles.


J-N-K (changes requested):
Pleas also check why there are test failures in `ProfileTypeResourceTest`.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/profiles/ProfileType.java
```diff
@@ -31,6 +32,19 @@ public interface ProfileType extends Identifiable<ProfileTypeUID> {
...
 
+    /**
+     * Get a collection of ItemType names that a Channel needs to support in order to able to use this ProfileType
+     *
+     * @return a collection of supported ItemType names (an empty list means ALL types are supported)
+     */
+    Collection<String> getSupportedItemTypesOfChannel();
+
+    /**
+     *
```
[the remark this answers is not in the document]
obones: I'll see what I can do, but it's just a cut/paste from previously existing code.

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/profile/ProfileTypeResource.java
```diff
@@ -128,13 +126,7 @@ private Predicate<ProfileType> matchesChannelUID(@Nullable String channelTypeUID
...
             return t -> false;
         }
-        switch (channelType.getKind()) {
-            case STATE:
-                return t -> stateProfileMatchesProfileType(t, channelType);
-            case TRIGGER:
-                return t -> triggerProfileMatchesProfileType(t, channelType);
-        }
-        return t -> false;
+        return t -> profileTypeMatchesChannelType(t, channelType);
```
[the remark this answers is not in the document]
obones: Well, to me, now that the methods have been moved to the `ProfileType` ancestor interface, there is no need to test because any future profile type will have to implement them.
Did I misunderstand something here?

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/link/ItemChannelLinkConfigDescriptionProvider.java
```diff
@@ -123,8 +130,25 @@ private boolean isSupportedItemType(ProfileType profileType, Item item) {
...
+    private boolean isSupportedChannelType(ProfileType profileType, Channel channel, @Nullable Locale locale) {
+        if (profileType.getSupportedChannelTypeUIDs().isEmpty()
+                && profileType.getSupportedItemTypesOfChannel().isEmpty()) {
+            return true;
+        }
+
+        ChannelType channelType = channelTypeRegistry.getChannelType(channel.getChannelTypeUID(), locale);
+        if (channelType == null) {
+            // requested to filter against an unknown channel type
+            return false;
```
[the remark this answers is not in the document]
obones: The type returned by `getChannelType` is nullable so the compiler complains about that.
I agree that logically it shouldn't happen. Note that this code is a simple copy/paste from somewhere else, but I'll add a log message here

obones:
I don't know if it's the same for everyone, but it takes a really long time to build the entire `openhab-core` project, and I can't even get it to work properly in its entirety with VSCode under Windows.
Hence the reason why I pushed commits to see what GitHub actions would say and try to figure out which sub project is complaining.
I'll make sure to squash those commits into a more readable one.

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/profile/ProfileTypeResource.java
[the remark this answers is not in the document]
J-N-K: There is a difference between TRIGGER and STATE channels, they serve different purposes. So for a TRIGGER channel type only TRIGGER profiles should be returned.

J-N-K:
No need for squashing, I can do that during merge. In fact, if you squash and force-push usually the comment history is messed up, so better just push additional commits.

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/profile/ProfileTypeResource.java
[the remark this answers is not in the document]
obones: Ah, yes, I see what you mean, it makes sense indeed.

obones:
I have just pushed some changes that I believe are addressing your fine remarks.
Do not hesitate to tell me if I missed other things.

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/link/ItemChannelLinkResource.java
```diff
@@ -236,9 +236,10 @@ public Response link(@PathParam("itemName") @Parameter(description = "itemName")
             }
             if (!(profileType.getSupportedItemTypes().isEmpty()
                     || profileType.getSupportedItemTypes().contains(itemType))
-                    || !(((TriggerProfileType) profileType).getSupportedChannelTypeUIDs().isEmpty()
-                            || ((TriggerProfileType) profileType).getSupportedChannelTypeUIDs()
-                                    .contains(channel.getChannelTypeUID()))) {
+                    || !(profileType.getSupportedChannelTypeUIDs().isEmpty()
+                            || profileType.getSupportedChannelTypeUIDs().contains(channel.getChannelTypeUID()))
+                    || !(profileType.getSupportedItemTypesOfChannel().isEmpty()
+                            || profileType.getSupportedItemTypesOfChannel().contains(itemType))) {
```
J-N-K, line 242: That looks really horrible. It would probably be better to 

```
   Collection<String> supportedItemTypes = profileType.getSupportedItemTypes();
   ...
   if ((!supportedItemType.isEmpty() && !supportedItemTypes.contains(itemType)) || ...) {

```

Why do you check `profileType.getSupportedItemTypesOfChannel`? A profile can change the item-type:

- The timestamp update profile works with all channel-types (e.g. a humidity channel with supported item-type `Number:Dimensionless`) but can only be linked to `DateTime` items. So a check would fail (in fact, currently the profile does not provide information about accepted channel types, so this will not cause an issue, but in principle it could).
- A profile could map numeric values to strings, so the channel would support `Number`, but the profile `String`. In this case you could never link them.

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/profile/ProfileTypeResource.java
```diff
@@ -151,31 +144,26 @@ private boolean profileTypeMatchesItemType(ProfileType pt, String itemType) {
                 || supportedItemTypesOnProfileType.contains(itemType);
     }
 
-    private boolean triggerProfileMatchesProfileType(ProfileType profileType, ChannelType channelType) {
-        if (profileType instanceof TriggerProfileType triggerProfileType) {
-            if (triggerProfileType.getSupportedChannelTypeUIDs().isEmpty()) {
-                return true;
-            }
+    private boolean profileTypeMatchesChannelType(ProfileType profileType, ChannelType channelType) {
+        @Nullable
```
J-N-K, line 148: You don't need `@Nullable` annotation for local variables.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/link/ItemChannelLinkConfigDescriptionProvider.java
```diff
@@ -123,8 +125,35 @@ private boolean isSupportedItemType(ProfileType profileType, Item item) {
...
     }
 
-    private boolean isSupportedChannelType(TriggerProfileType profileType, Channel channel) {
-        return profileType.getSupportedChannelTypeUIDs().isEmpty()
-                || profileType.getSupportedChannelTypeUIDs().contains(channel.getChannelTypeUID());
+    private boolean isSupportedChannelType(ProfileType profileType, Channel channel, @Nullable Locale locale) {
+        @Nullable
+        ChannelKind supportedChannelKind = profileType.getSupportedChannelKind();
+        if (supportedChannelKind != null && supportedChannelKind != channel.getKind())
+            return false;
```
J-N-K, line 132:
```suggestion
        if ((channel.getKind() == STATE && !(profileType instanceof StateProfileTypeImpl)) || (channel.getKind() == TRIGGER && !(profileType instanceof TriggerProfileTypeImpl) {
            return false;
        }
        @Nullable
        ChannelKind supportedChannelKind = profileType.getSupportedChannelKind();
        if (supportedChannelKind != null && supportedChannelKind != channel.getKind())
            return false;
```

You don't need the new method in `ProfileType` then.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/profiles/StateProfileTypeImpl.java
```diff
@@ -31,25 +34,38 @@ public class StateProfileTypeImpl implements StateProfileType {
...
     public Collection<String> getSupportedItemTypes() {
         return supportedItemTypes;
     }
 
+    @Override
+    @Nullable
+    public ChannelKind getSupportedChannelKind() {
+        return ChannelKind.STATE;
+    }
+
```
J-N-K, line 68: see above, I don't think you need that

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/profiles/TriggerProfileTypeImpl.java
```diff
@@ -60,4 +64,15 @@ public String getLabel() {
...
+    @Override
+    public Collection<String> getSupportedItemTypesOfChannel() {
+        return supportedItemTypesOfChannel;
+    }
+
+    @Override
+    @Nullable
+    public ChannelKind getSupportedChannelKind() {
+        return ChannelKind.TRIGGER;
+    }
```
J-N-K, line 77: see above, I don't think you need that

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/profiles/ProfileType.java
```diff
@@ -26,11 +29,34 @@
...
+    Collection<ChannelTypeUID> getSupportedChannelTypeUIDs();
+
+    /**
+     * Get the ChannelKind this profile type supports.
+     *
+     * @return The supported ChannelKind for this profile type. If null then all channel kinds are supported
+     */
+    @Nullable
+    ChannelKind getSupportedChannelKind();
+
```
J-N-K, line 59: see above: I don't think you need that

J-N-K (changes requested):
Thanks again for you work and sorry for being a PITA here.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/link/ItemChannelLinkConfigDescriptionProvider.java
obones: I did this because I personally find it horrible to resort to `instanceof` to figure out what to do. I mean, the whole point of an interface is to abstract out the implementing class, or even derived interfaces.
If in the future there is another profile type, there would be another interface but there would not be a need to come back to this code because the `getSupportedChannelKind` already provides the answer for this.
Also, if there is a new member in the `ChannelKind` enumeration, once again the code above does not need to be changed, it's the same method that I introduced that would be called.

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/profile/ProfileTypeResource.java
obones: My IDE (VSCode) disagrees and outputs a warning saying that the value cannot be null. That's what I've been doing in my plugin bundles, but I can remove it if you prefer.

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/link/ItemChannelLinkResource.java
obones: I agree that it looks horrible, but it was jut to "fit in" with the already existing code.

As to why I do the check, well it's because I did a global search for the interface methods and made sure both methods are called in all the places where one was called following an `instanceof` call.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/link/ItemChannelLinkConfigDescriptionProvider.java
J-N-K: IMO very unlikely, but ok. Let's keep the methods.

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/profile/ProfileTypeResource.java
J-N-K: IMO it's wrong. The default locations for `@NonNullByDefault` are `PARAMETER, RETURN_TYPE, FIELD, TYPE_BOUND, TYPE_ARGUMENT`, so it does not apply to local variables.

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/link/ItemChannelLinkResource.java
J-N-K: I agree, the style was not good before either and if you want to keep it: ok. But the `getSupportedItemTypesOfChannel` is excluding valid combinations.

obones: I'm definitely in favor of changing the code style here, but as always when I'm submitting a pull request, I try to limit my changes to the code logic, adapting to the local style I encounter.
I'll try to come up with a more sensible way of writing this up.

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/profile/ProfileTypeResource.java
obones: Fair enough, I have removed them

obones:
The more I look at how things are going, the more I feel it would be beneficial to have a `ProfileTypeUtil` abstract class that provides two methods receiving (among other things) a `Channel` for the first and a `ChannelType` for the second and those methods would return a boolean indicating if a given `ProfileType` applies to the channnel(type)
This would avoid very similar code to be placed in the three different places it is currently placed.
Does that make any sense to you?

obones:
@J-N-K , sorry to ping you directly, but do you have an opinion on my last suggestion?

PR #4329, opened 2024-07-24
[rules] Don't attempt to pre-compile disabled rules

florian-h05:
Fixes an issue, where an error that compilation failed for disabled rules.

Reported on the community: https://community.openhab.org/t/oh-4-2-snapshot-disabled-rules-failed-to-compile-error-in-opehab-log/157402.
Follow-up for #4289.


rkoshak:
What happens when the rule that was disabled becomes enabled? Does the pre-compile occur or does the rule need to wait until it's triggered to compile?

florian-h05:
The rule is pre-compiled as part of the activation process: https://github.com/openhab/openhab-core/blob/e6c85191945a844942d9f7896a5a579aae2eb59a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleEngineImpl.java#L852

J-N-K (approved):
Thanks!

PR #4336, opened 2024-08-04
Icon servlet: Set Cache-Control header to enable icon caching

florian-h05:
I have removed the Modified-Since handling as it did not have an effect due to the missing Cache-Control header, and instead added Cache-Control with a max-age set. 
This enables "forever" caching of icons, which should be fine since they are static assets - they now have the same caching policy as files in $OPENHAB_CONF/the html folder.

Should be a nice change especially on mobile, where I can often see that icons need some time to load with slow cellular network.


J-N-K (approved):
Thanks!

PR #4344, opened 2024-08-11
Add MEDIAN group function

### LINKED ISSUE #4341: Median as group function.
A usefull additional function would be  the caclulation of the median of a group of items, in addition to AVG, MAX, MIN and SUM.

jimtng:
Resolve #4341


bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunction.java
```diff
@@ -111,6 +114,37 @@ public State calculate(@Nullable Set<Item> items) {
...
+            super(dimension);
+        }
+
+        @Override
+        @SuppressWarnings({ "unchecked", "rawtypes" })
+        public State calculate(@Nullable Set<Item> items) {
+            if (items != null) {
+                List<QuantityType> states = items.stream().filter(item -> isSameDimension(item))
+                        .map(item -> item.getStateAs(QuantityType.class)).filter(Objects::nonNull)
+                        .sorted((s1, s2) -> s1.compareTo(s2)).collect(Collectors.toList());
```
J-N-K, line 132:
```suggestion
                        .sorted((s1, s2) -> s1.compareTo(s2)).toList();
```

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/ArithmeticGroupFunction.java
```diff
@@ -253,6 +256,47 @@ public State[] getParameters() {
...
+    class Median implements GroupFunction {
+
+        public Median() {
+        }
+
+        @Override
+        public State calculate(@Nullable Set<Item> items) {
+            if (items != null) {
+                List<DecimalType> states = items.stream().map(item -> item.getStateAs(DecimalType.class))
+                        .filter(Objects::nonNull).sorted((s1, s2) -> s1.compareTo(s2)).collect(Collectors.toList());
```
J-N-K, line 271:
```suggestion
                        .filter(Objects::nonNull).sorted((s1, s2) -> s1.compareTo(s2)).toList();
```

J-N-K (changes requested):
Otherwise LGTM. Thanks!

mherwege:
@jimtng Great you took this one. I am wondering if it makes sense to use a more performant algorithm using heaps? See https://www.baeldung.com/java-stream-integers-median-using-heap. I expect it will for the other case with persistence when looking at a longer period with many data points. Here the impact could be important because this recalculation (and sorting) will ve done for every state change of an item in the group.

mherwege:
And here is a text on another, at first side even better, algorithm: https://rcoh.me/posts/linear-time-median-finding/

jimtng:
@mherwege great idea! I'm going to do some tests comparing the two/three implementations. 

I'm also wondering about keeping an in-memory sorted list and update it whenever the members changed. That would involve a sorting operation at the beginning, and only deletion + insertion of (changed|added|removed) item(s), which should probably be even faster than any quickselect / heap median algorithm. However, the way it's currently done, we got given a list of items with no knowledge of the groupitem (only the baseitem). So to do that, a lot of things would need to be changed.

mherwege:
@jimtng I am not sure the extra complexity of handling item state changes + item addition + item removal in a group are worth it for this use case. Using an algorithm that scales linearly (or at least most of the time) with number of items I suspect does.
Looking at the algorithms, the advantage of the heap methods is when you want to have a stream of medians with every item you add. It avoids sorting (or quickselect) every time. That is not the case here. You only need to calculate once for a change every time, so quickselect looks like the most promising. I of course applaud you intending to test it.

mherwege:
> deletion + insertion of (changed|added|removed) item(s)

That’s what the heap approach could optimise.

jimtng:
Note: waiting for #4345 to be merged, and refactor to use the `Statistics.median` method.

mherwege:
@jimtng https://github.com/openhab/openhab-core/pull/4345 has been merged.

jimtng:
Thanks, updated!

J-N-K (approved):
LGTM

PR #4352, opened 2024-08-18
Support channel transformations with parentheses

jimtng:
Support both `JSONPATH:$.xx` and `JSONPATH($.xx)` syntaxes.

The aim is to make transformation pattern uniform between bindings and core, e.g. in item / sitemap labels, so we can just use the parentheses everywhere. Perhaps colon syntax can be phased out eventually but for now they're both supported to maintain compatibility.

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/generic/ChannelTransformation.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/generic/ChannelTransformation.java
@@ -15,6 +15,8 @@ package org.openhab.core.thing.binding.generic;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
+import java.util.regex.Matcher;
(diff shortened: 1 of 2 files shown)


J-N-K (changes requested):
In general looks good to me. Can you also add a test for the mixed case (i.e. both patterns in one transformation chain)?

jimtng:
tests for mixed syntax added

J-N-K (approved):
LGTM, thanks for the improvement

lolodomo:
I guess something has to be changed in core.ui to support syntax with ":"?
https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java#L125

jimtng:
At the moment this class is used just for addons. I am looking through the core code to see if we can apply a common code to handle all transformation related tasks, so that they will all support chaining and perhaps also support `:`.

PR #4355, opened 2024-08-19
Add isTransform(), isEmpty() and isPresent() to ChannelTransformation

jimtng:
This is needed by the user of this class, e.g. mqtt binding, to know whether to convert the command to `StringType` when transformations were involved, or continue using the native type.

Example use cases

https://github.com/openhab/openhab-addons/blob/0d6feb72aa23dcfce946e1367ea0ce8b8c03b330/bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/ChannelState.java#L136

https://github.com/openhab/openhab-addons/blob/0d6feb72aa23dcfce946e1367ea0ce8b8c03b330/bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/ChannelState.java#L340

`isTransform()` is needed by the [modbus addon](https://github.com/openhab/openhab-addons/pull/17306)

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/generic/ChannelTransformation.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/generic/ChannelTransformation.java
@@ -71,15 +71,28 @@ public class ChannelTransformation {
         return Arrays.stream(transformationString.split("∩")).filter(s -> !s.isBlank());
     }
 
+    public boolean isEmpty() {
+        return transformationSteps.isEmpty();
+    }
+
+    public boolean isPresent() {
+        return !isEmpty();
+    }
+
     public Optional<String> apply(String value) {
         Optional<String> valueOptional = Optional.of(value);
 
         // process all transformations
         for (TransformationStep transformationStep : transformationSteps) {
-            valueOptional = valueOptional.flatMap(transformationStep::apply);
+            valueOptional = valueOptional.flatMap(v -> {
+                Optional<String> result = transformationStep.apply(v);
+                if (logger.isTraceEnabled()) {
+                    logger.trace("Transformed '{}' to '{}' using '{}'", v, result.orElse(null), transformationStep);
+                }
+                return result;
+            });
         }
 
-        logger.trace("Transformed '{}' to '{}' using '{}'", value, valueOptional, transformationSteps);
         return valueOptional;
     }
 
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/generic/ChannelTransformation.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/generic/ChannelTransformation.java
@@ -72,32 +72,30 @@ public class ChannelTransformation {
     }
 
     /**
-     * Checks whether the given string contains valid transformations.
+     * Checks whether this object contains no transformation steps.
      * 
-     * Valid single and chained transformations will return true.
-     * 
-     * @param value the transformation string to check.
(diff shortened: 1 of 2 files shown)


jimtng:
Modbus refactoring also requires this PR.

J-N-K:
I don't understand the need for `.isEmpty` and `.isPresent`. In case it is empty, the original value is returned, so what will break if we call `.apply` in every case?

jimtng:
See https://github.com/openhab/openhab-addons/blob/0d6feb72aa23dcfce946e1367ea0ce8b8c03b330/bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/ChannelState.java#L340

We need the ability to check if a transformation is at all defined. If it's defined, the command to be published will be converted to StringType, and subsequently formatBeforePublish will treat the data as a String data.

If it's not defined, then the original command type will be preserved (not converted to StringType) and the formatBeforePublish can apply type-specific formatting, e.g. special date format.

So that's how we need to be able to know whether transformation is defined.

I can instead check if `config.transformationPatternOut` is null or blank, and set the outgoingTransformation to null, and check it that way if you don't think we should add isEmpty/isPresent here.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/generic/ChannelTransformation.java
```diff
@@ -71,15 +71,47 @@ private static Stream<String> splitTransformationString(String transformationStr
...
 
+    /**
+     * Checks whether the given string contains valid transformations.
+     * 
+     * Valid single and chained transformations will return true.
+     * 
+     * @param value the transformation string to check.
+     * @return <code>true</code> if the string contains valid transformations, <code>false</code> otherwise.
+     */
+    public static boolean isTransform(@Nullable String value) {
```
J-N-K, line 82: Please move this down and place it before the `TransformationStep` class. I think `isValidTransformation` would be a better naming.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/generic/ChannelTransformation.java
```diff
@@ -71,15 +71,47 @@ private static Stream<String> splitTransformationString(String transformationStr
...
+            return false;
+        }
+        try {
+            return splitTransformationString(value).map(TransformationStep::new).count() > 0;
+        } catch (IllegalArgumentException e) {
+            return false;
+        }
+    }
+
+    public boolean isEmpty() {
```
J-N-K, line 93: Add JavaDoc

J-N-K (changes requested):
IMO it would be fine to do this where needed, but I'm ok with adding it here. Please see my comments.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/generic/ChannelTransformation.java
jimtng: I was following the naming from the other classes. Renamed as suggested.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/generic/ChannelTransformation.java
jimtng: added

J-N-K (approved):
LGTM

J-N-K:
I have triggered a new core snapshot, should be available in about 35 minutes.

PR #4357, opened 2024-08-19
Filter out blank lines and comments in a multi-line transformations

jimtng:
This allows users to insert comments inbetween transformations in a multi-line / chained (or even a single transformation).

e.g.

```
# Extract value only if it's two digits of more
REGEX(\d{2,})
```


rkoshak:
Given that JSON doesn't support comments do these comments survive the conversion between JSON to YAML and back to 
JSON for managed configs?

jimtng:
Yes, the comments will survive the json/yaml conversions, because they are part of the actual configuration data.

holgerfriedrich (commented):
LGTM.
Did you consider extending the tests in ```ChannelTransformationTest.java```?

holgerfriedrich (approved):
Thank you

PR #4361, opened 2024-08-23
Rename parameter of ChannelHandler

jimtng:
@J-N-K did this get missed when moving from the http binding to core?


J-N-K (approved):
LGTM, thanks. Seems a leftover, indeed.

Issue #4359, opened 2024-08-21
PersistenceExtensions ignores item persistence aliases


jpg0:
## Expected Behavior

`PersistenceExtensions.lastUpdate(item)` should work for any supplied item, even if aliased

## Current Behavior

As code directly reaches into persistence services using a query filter, it bypasses any alias configuration, so if the item has an alias then the filter misses it as it uses the item name directly.

## Possible Solution
Apply alias to the filter prior to executing it.

mherwege:
What persistence service are you using?

Does it work for any of the persistence extensions, or any of the mechanisms reading from the database (e.g. restoreOnStartup)?

As far as I understand, the alias is there to allow defining a different name for an item to be used in the database of the persistence service. That can be e.g. for adjusting names to comply with DB constraints. As such, it should be transparant to the user. Persistence extensions are a user mechanims used in rules and should be agnostic of the alias. But I don't see a consistent mechanism in the code to map an item name back to an alias. I think many persistence services actually don't do anything with the alias and just ignore it.

To make this work consistently, there may be need for some architectural discussion:
- Where should the alias mapping be kept? At the moment it is in the item persistence configuration.
- Having the mapping in the persistence configuration is fine for automatic writing through a persistence strategy, but what for reading back? Should we make it the responsability of the specific persistence service to map back, so it can keep track of aliases used in the past (if you would change the persistence configuration). If so, can it store a mapping table, real name to list of aliases?
- You can use the persistence extension `persist` to persist an item state. This can also be done for an item that has no persistence configuration at all, so no alias defined. What should happen? Again, I don't think it would be wise to define the alias in the call, making it the responsability of the user to know the underlying alias.

Honestly, I think the whole alias concept, while good in principle, is not working at all at the moment for the various reasons listed above. It needs more than a quick fix in the persistence extensions to solve that. As nobody has kept all persistence code to properly support this `alias` I also don't know how much it is used, and where it is critical. I am just wondering if it worth investing the time to fix it.

jpg0:
I've just dug a little deeper and I think you're right. The alias is used when saving state, however not when reading it. Restoring is done in `PersistenceManagerImpl$PersistenceServiceContainer#restoreItemStateIfPossible` and tbh it would be fairly straightforward to add the mapping there (pass it in, as the caller knows it). Right now I can't see how it's anything other than completely broken without this.

As for larger architectural concerns, I'm not sure that I'm qualified to answer, but here goes:

[...]

PR #4363, opened 2024-08-27
Persistence aliases

mherwege:
See discussion https://github.com/openhab/openhab-core/issues/4359

At its current state, defining and using aliases for persistence is largely broken.

1. At the point of definining aliases, one can define an alias in the syntax for an item, item group or all items. There are several issues with this way of defining:
- The last two options obviously don't make sense.
- One has to define with all parameters (strategy, filters) for each item you want to define an alias for (otherwise the default will be used). It becomes impossible to use a group definition for the strategy and filters.
- It is not possible to define an alias for an item without a strategy (once you define it, the default strategy will be applied). This can be useful if an item is persisted manually through a rule, and not through a strategy.
2. There are methods to store items with aliases. However, on retrieval (`query`), it is assumed the `itemName` in the `FilterCriteria` is the alias. Many persistence services use this filter `itemName` to look up the item in the item registry (e.g. to find the unit). This does not work if an alias is passed in the filter criteria.

This PR does 2 things:
1. Redefine syntax to have a separate section to define aliases. This is a breaking change, but allows defining aliases in a proper and consistent way, and is easy enough to apply.
2. Add interface methods (with default implemention) `QueryablePersistenceService.query(FilterCriteria, String)` and `ModifiablePersistenceService.remove(FilterCriteria, String)`. The second argument defaults to `null` in its default implementation. Persistence services implementing support for aliases and using item registry lookups, should implement these new methods to make querying with aliases work. However, as there is a default implementation, if the persistence service is not changed, it will behave as it does now (failing if the persistence service tries to do a lookup in the item registry). All core calls to the `query` and `remove` methods are also changed to add the alias if one exists. I expect some classes in the ui repo (and maybe some scripting addons) will need limited changes as well to properly support aliases. (I have not done this, but the methods without alias could be annotated `@Deprecated` to make it clear to the consumer they should use the methods with alias arguments).

For 2, the better approach would probably be to make the individual persistence services completely independent of the item registry. The interface should then have `store`, `query` and `remove` methods that are only concerned about the storage key and values, whereby the storage key is the item name or alias. The core persistence layer would be responsible for using item name or alias when calling these methods, and then also for all actions currently done in the individual persistence services where the lookup in the item registry is required (e.g. unit lookup and conversion). Also ChartServlets would be impacted by this. As this is obviously a major API change, I have refrained from doing that. With the current proposal, the interaction with persistence services should keep on working as before, and persistence services that support aliases (that is a limited number) can easily be enhanced to properly support them (instead of the broken support in the current implementation without applying required changes).

Another option would be to completely remove support for aliases from openHAB core, and leave it to individual persistence services if they want to implement something, marking the `store` methods with `alias` parameters as deprecated. I also think that would be a step back and is not needed. But it may simplify code and, as it stands, I don't have the impression aliases are used a lot (if they where, I would expect to see many more issues being created as they don't work).

This PR is offered as a starting point and open for discussion. My design premises have been to try to fix it while not breaking the interface with persistence addons. I would very much appreciate feedback if this approach makes sense, and would be accepted before I invest further time in it.

@jpg0 FYI and evaluation.

EDIT: changed the default query method implementations to do an alias lookup, so the returned results are correct as long as the implementing persistence service does not do an item registry lookup. This again limits impact on persistence service implementations.

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/persistence/PersistenceResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/persistence/PersistenceResource.java
@@ -16,9 +16,13 @@ import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
+import java.util.Date;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.annotation.security.RolesAllowed;
 import javax.ws.rs.Consumes;
@@ -104,6 +108,7 @@ import io.swagger.v3.oas.annotations.tags.Tag;
  * @author Lyubomir Papazov - Change java.util.Date references to be of type java.time.ZonedDateTime
  * @author Markus Rathgeb - Migrated to JAX-RS Whiteboard Specification
  * @author Wouter Born - Migrated to OpenAPI annotations
+ * @author Mark Herwege - Implement aliases
  */
 @Component
 @JaxrsResource
@@ -181,9 +186,10 @@ public class PersistenceResource implements RESTResource {
             PersistenceService service = persistenceServiceRegistry.get(serviceId);
             if (service != null) {
                 List<PersistenceStrategy> strategies = service.getDefaultStrategies();
-                List<PersistenceItemConfiguration> configs = List.of(
-                        new PersistenceItemConfiguration(List.of(new PersistenceAllConfig()), null, strategies, null));
-                configuration = new PersistenceServiceConfiguration(serviceId, configs, strategies, strategies,
+                List<PersistenceItemConfiguration> configs = List
+                        .of(new PersistenceItemConfiguration(List.of(new PersistenceAllConfig()), strategies, null));
+                Map<String, String> aliases = Map.of();
+                configuration = new PersistenceServiceConfiguration(serviceId, configs, aliases, strategies, strategies,
                         List.of());
                 editable = true;
             }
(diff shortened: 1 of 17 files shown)


jpg0:
Thanks for the PR. I think it addresses the current problems and introduces the least possible change. The code looks fine from a quick look.

mherwege:
This is now tested, and if the concept is accepted, it is ready from my perspective. This is of course open for further discussion.

kaikreuzer:
Thanks @mherwege for bringing this up and suggesting a solution. I'd be fine with this change, but could also consider to wipe aliases completely in openHAB 5 - as you say, they do not seem to be used much anyhow. But your suggestion is clearly less intrusive.

When we merge this, we would need a "breaking change note" in the distro and an update of the documentation, right?
I'd hence refrain from merging this just before creating milestone 1.

mherwege:
> I'd be fine with this change, but could also consider to wipe aliases completely in openHAB 5 - as you say, they do not seem to be used much anyhow.

I see value in the functionality when changing item names. So I would want to keep it and improve it.

> When we merge this, we would need a "breaking change note" in the distro and an update of the documentation, right?

Absolutely. I will prepare these.

I have rebased and resolved some conflicts with already merged persistence enhancements.

mherwege:
Breaking change alert and documentation update have been created.

kaikreuzer (approved):
Excellent, many thanks for your work!

holgerfriedrich:
@mherwege did you create a PR to addons as well?
This seems to break `MapDbPersistenceService.java` in line 214, and `MapDbPersistenceServiceOSGiTest.java`.

mherwege:
@holgerfriedrich I am preparing one where I get aliases to work for all persistence services.

I have the fix already for mapdb and will split it out in a separate PR. The issue is actually caused by parallel work on https://github.com/openhab/openhab-addons/pull/17820 and this PR, requiring a further update to mapdb as the mapdb PR got merged earlier.

holgerfriedrich:
@mherwege no problem, good to know that you are working on it  👍 

The addon build was red for a few days now and I had hoped that my last PR on core had resolved all issues....
So once you are ready, the CI will hopefully turn green again.

mherwege:
PR created: https://github.com/openhab/openhab-addons/pull/18285

PR #4370, opened 2024-09-05
Support space-separated date/time format for DateTimeType

jimtng:
Support `yyyy-MM-dd HH:mm`  syntax in addition to `yyyy-MM-ddTHH:mm`.

The syntax with space often appears in incoming data, e.g. mqtt. By supporting this, we won't need to run the input data through a transformation.


holgerfriedrich (approved):
Seems to be a useful addition.
I thought if it would be a better way to add the check for the missing "T" to the method ```parse()```, but I am fine with the implementation.

Issue #947, opened 2019-07-29
Add TOGGLE command to OnOffType devices


giig1967g:
Hi,
I would like to suggest to add a new command to the Switch devices (OnOffType): TOGGLE.
This command will invert the status of the switch, ie:
if the switch in ON, the TOGGLE command will turn it OFF.

kaikreuzer:
This was extensively discussed in https://github.com/eclipse/smarthome/pull/4638 and the conclusion was that this would be a huge breaking change that we are not willing to accept.

splatch:
@kaikreuzer since 2.5.0 is out and 3.0 is a new horizon, does it make sense to consider `Toggle` as a first class citizen in roadmap for next release? Some automation devices already support "toggle" functionality thus adding `Toggle` command will allow having 1:1 mapping between hardware and software layer.

PR #4381, opened 2024-09-15
Toggle functionality for "boolean" types

magx2:
Hey,

Writing DSL functions I found out that it would be really nice to be able to toggle types to opposite, i.e. You got command `ON` from item x and you want to send command `OFF` (opposite command) to item y:

```
val oppositeState = if(x.state==ON) OFF else ON
y.sendCommand(oppositeState )
```
vs
```
y.sendCommand(x.state.toggle)
```

WDYT?

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/OnOffType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/OnOffType.java
@@ -60,6 +60,10 @@ public enum OnOffType implements PrimitiveType, State, Command {
         return super.toString();
     }
 
+    public OnOffType toggle() {
+        return this == ON ? OFF : ON;
+    }
+
     @Override
     public <T extends State> @Nullable T as(@Nullable Class<T> target) {
         if (target == DecimalType.class) {
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/OpenClosedType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/OpenClosedType.java
@@ -42,6 +42,10 @@ public enum OpenClosedType implements PrimitiveType, State, Command {
         return super.toString();
     }
 
+    public OpenClosedType toggle() {
+        return this == OPEN ? CLOSED : OPEN;
+    }
+
     @Override
     public <T extends State> @Nullable T as(@Nullable Class<T> target) {
         if (target == DecimalType.class) {
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/PlayPauseType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/PlayPauseType.java
@@ -42,4 +42,8 @@ public enum PlayPauseType implements PrimitiveType, State, Command {
     public String toFullString() {
         return super.toString();
     }
+
+    public PlayPauseType toggle() {
+        return this == PLAY ? PAUSE : PLAY;
+    }
 }
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/UpDownType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/UpDownType.java
@@ -44,6 +44,10 @@ public enum UpDownType implements PrimitiveType, State, Command {
         return super.toString();
     }
 
+    public UpDownType toggle() {
+        return this == UP ? DOWN : UP;
+    }
+
     @Override
     public <T extends State> @Nullable T as(@Nullable Class<T> target) {
         if (target == DecimalType.class) {


spacemanspiff2007:
How about a `TOGGLE` command which can be sent to the corresponding item type?

magx2:
Is there a toggle command?

spacemanspiff2007:
Currently not - I am suggesting it might make sense to implement one.

magx2:
> Currently not - I am suggesting it might make sense to implement one.

No way. I already saw long discussions about it. If you want you can do it

https://github.com/openhab/openhab-core/issues/947
https://github.com/eclipse-archived/smarthome/pull/4638
https://community.openhab.org/t/controlling-a-dimmer-item-by-physical-buttons-various-questons-to-achieve-clean-solution/131128/15

andrewfg:
Umm. I have an issue with your word 'toggle' .. its dictionary definition is _"to switch a feature on a computer on **and** off by pressing the same button or key"_ .. in other words the 'toggle' operation executes TWO operations e.g. ON=>OFF=>ON.

So I some other word like 'invert' or 'reverse' or 'flip' ..

magx2:
@andrewfg 

I think you misunderstand the [Cambridge dictonary](https://dictionary.cambridge.org/dictionary/english/toggle) (or I did).

Look at examples of usage:
- _Use this key to toggle between the two typefaces_ -> _Use `toggle` method to toggle between states_
- _By toggling this key, you can switch the italics on and off._ -> _By toggling this method, you can switch the state on and off_

mherwege:
I believe `toggle` is a verb in this context. It implies an action to take. This is not the case in what you coded. There is no state change involved in the original item. It shouldn't be a verb. What about something like `inversed`?
To be perfectly honest, I don't see the value. While it may simplify the rules code a little bit, it can also confuse, as is also apparent from the discussion on the terminology. It is not an attribute of the the original state, but rather something you want to apply to another item's (or the same item's) state. So I actually think we should not do this and keep it explicit in the rule. But that is my opinion on it.

spacemanspiff2007:
Toggling on NULL and UNDEF is something I would not expect.
If a lamp is on, I start openHAB and then call toggle a ON command will be sent because the lamp state has not yet updated - the toggle command does the wrong thing.

The toggle scope is limited to ON -> OFF and OFF -> ON. 
All other states should do nothing and/or throw an error.
That's how HABApp implements it.

jimtng:
> Toggling on NULL and UNDEF is something I would not expect.
> If a lamp is on, I start openHAB and then call toggle a ON command will be sent because the lamp state has not yet updated and is NULL -> the toggle command does the wrong thing.

What's a practical example of when a script would issue a `toggle` command?

It is usually called by a momentary button press: press it to turn it on, and press it again to turn it off, and press it again to turn it back on. In this instance, physical light is already on, openhab just loaded, the item state is still NULL, so you press the button, it would send ON command, nothing happens (because it's already on). So user presses the button again, but this time because the item state would've already been initialised (either by item autoupdate, or by sending the command, the item would poll the current state), so the subsequent press turns it off.

Now if you instead throw an error, the item state will remain NULL, and the button is not "working" and you'll have to wait until the item state gets updated by the binding.

spacemanspiff2007:
I get why you would choose the pragmatic approach.
However you make a lot of assumptions and it's not guaranteed that these assumptions are always true.
But if I call a ``toggle`` method I expect it to toggle every time, not most of the time.
Contrary to commands like ON and OFF toggle is not stateless, so it by definition it requires that the current state is known.

magx2:
> What's a practical example of when a script would issue a `toggle` command?

I have a physical button that only clicks - it does not stay in on state, just send and one time event with on and then goes off.

jimtng:
> But if I call a `toggle` method I expect it to toggle every time, not most of the time.

"most of the time" implies that the command is unreliable. Sometimes calling toggle would work, sometimes calling it wouldn't work. 

The implementation I described would only "not toggle the actual device" during the initial load and only if the current state is unknown. If the state is known prior to the call, it would behave the same way, with the same reliability of toggling _every time_ too.

The only difference is that it handles an unexpected state in a pragmatic way, so that user need not code additional  workarounds in most cases.

Should they insist on not toggling when the state is unknown, simply add a check for that
```
if MySwitch.state? # is the state known?
  MySwitch.toggle 
else
  logger.info "Sorry you cannot toggle MySwitch yet because its state is unknown."
end
```

As a result, when you press the button, nothing happens, the light stays ON (which is the same behaviour anyway). And even if you pressed the button 10 more times, as long as the device state remains unknown, no "toggling" would happen. That's far from "I expect it to toggle every time". So which is worse?

Either implementation still allows the user to achieve either behaviour, with or without some additional code. At the end of the day it's a design decision.

Getting back to this PR though: it would introduce another issue

`MySwitch.command(MySwitch.state.inverse())` would fail if MySwitch.state is NULL / UNDEF because you haven't defined UnDefType.inverse(). PS I cal it inverse rather than toggle here.

magx2:
I just found that JS rules also has toggle option: https://github.com/openhab/openhab-js/blob/main/src/items/items.js#L341

spacemanspiff2007:
> To be fair in this use case `ON` and `OFF` are also not stateless

The `ON` and `OFF` command is stateless since it will always turn on / turn off. 
Toggle is not stateless because the new state depends on the current state.


> "most of the time" implies that the command is unreliable: sometimes it works, sometimes it doesn't.

Control question: 
If the device is on and the item state is NULL and I call the toggle method does the device output toggle?
Typically most of the time the state is not NULL so the toggle method works most of the time.

---

You always assume that the target state should be on. 
What if I have e.g. a pump? Maybe it would make sense to turn it off?
Maybe I run software pwm for underfloor heading in rule - depending on the logic of NULL handling the output percentage will be inverted (e.g. suddenly 70% instead of 30%).


> it's a design decision.

Correct

magx2:
> > To be fair in this use case `ON` and `OFF` are also not stateless
> 
> The `ON` and `OFF` command is stateless since it will always turn on / turn off. Toggle is not stateless because the new state depends on the current state.

You need to know current state to send `ON`/`OFF`, so (in this usecase) it's not stateless. You have quoted only half of my message, the other half explained what I've meant.

PR #4382, opened 2024-09-16
[thing] Add toString overrides to ThingImpl & BridgeImpl

florian-h05:
This is nice to have when working with Things in scripts.

I have chosen to not include the config because the config might include sensitive data like passwords.


holgerfriedrich (commented):
Thanks, this seems useful.

I'd like to see the comment "Configuration is deliberately excluded because it might include sensitive data like passwords." in Javadoc - unfortunately, it will not be visible, as the Impl classes are not shown. Adding ```toString()``` to the interface ```Thing``` would probably clutter the interface too much.... So maybe just as a code comment in ```ThingImpl.java```?

holgerfriedrich (approved):

holgerfriedrich:
LGTM, thanks

PR #4384, opened 2024-09-18
Extend HistoricItem to work with Instant instead of ZonedDateTime

joerg1985:
This PR will add a new `default` method to the HistoricItem interface, to get the timestamp as `Instant`.
The most areas using the `HistoricItem` must transform the internal `Instant` to a `ZonedDateTime` (e.g. [all the persistence implementations](https://github.com/joerg1985/openhab-addons/commit/e400f7352cc8151db25304e33fce256c025b6e9f)) or need to transform the returned `ZonedDateTime` to an `Instant` (e.g. see the below).

So i my mind it does make sense to use a `Instant` here and not transform the data.
The link to the persistence implementations above does show the changes needed to the addons.


rkoshak:
Will this be a breaking change for rules which currently expect a ZonedDateTime?

joerg1985:
@rkoshak no this is a none breaking change, the new method has a default implementation and all old methods are still in place.

J-N-K (approved):
LGTM, thanks!

PR #4386, opened 2024-09-19
Add XOR ArithmeticGroupFunction

### LINKED ISSUE #4385: Add XOR ArithmeticGroupFunction
It would be nice to be able to XOR (not just AND, OR, NAND, NOR) states.

fabianvo:
Resolve #4385 

PR provides a simple implementation for XOR. It uses "1 of n" for n > 2 (n = item count in group).


bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/ArithmeticGroupFunction.java
```diff
@@ -35,6 +35,7 @@
  * @author Kai Kreuzer - Initial contribution
  * @author Thomas Eichstädt-Engelen - Added "N" functions
  * @author Gaël L'hopital - Added count function
+ * @author Fabian Vollmann - Added ONE function
```
kaikreuzer, line 38: Why ONE? Don't you mean XOR?

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/ArithmeticGroupFunction.java
```diff
@@ -212,6 +213,82 @@ public State calculate(@Nullable Set<Item> items) {
         }
     }
 
+    /**
+     * This does a logical 'xor' operation. Only if items are of opposing 'activeState' and
+     * 'passiveState', 'activeState' is returned. For more than 2 states, this is '1 of n'.
```
kaikreuzer, line 218: > Only if items are of opposing 'activeState' and 'passiveState', 'activeState' is returned

I do not fully get that statement. Could you try to reformulate it to make it clearer?

kaikreuzer (commented):
Thanks, lgtm apart from the two minor comments above.
May I also ask you to prepare a PR for the documentation update at https://github.com/openhab/openhab-docs/blob/main/configuration/items.md#derive-group-state-from-member-items?

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/ArithmeticGroupFunction.java
fabianvo: changed to XOR

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/ArithmeticGroupFunction.java
fabianvo: I hope comment is easier to understand now

kaikreuzer (approved):
Thank you!

PR #4394, opened 2024-09-30
[rest] Persistence: Optionally add current Item state to response

florian-h05:
This adds a new query param `itemState` to the GET `/rest/persistence/items/{itemname}` endpoint.

This new parameter allows to add the current Item state to the data returned from persistence, which is a huge improvement for charts in the UI:

- "normal" charts range until the end of the requested time range
- "aggregate" charts such as a monthly chart are able to display the values of the current month even if the monthly value is only persisted at the end of the month

There are two special cases where the Item state is not added with the current time to the response:

- If the `boundary` query param is also set to true and a state was already added to the requested end time, there is no need to include the current Item state as there already is a state at the end of the time range.
- If the current time is after the requested end time and no end state was added, add the current state as the end state.

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/persistence/PersistenceResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/persistence/PersistenceResource.java
@@ -12,6 +12,7 @@
  */
 package org.openhab.core.io.rest.core.internal.persistence;
 
+import java.time.Instant;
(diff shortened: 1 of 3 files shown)


stefan-hoehn:
I can totally understand why this param makes sense. These are the situations where you want to have all the information presented and where the data to be shown has not been persisted yet (for a reason), so you start developing you own specific implementation and work around. Therefore, your extension is very welcome! Very good idea.

kaikreuzer (approved):
If it helps for nicer charts, let's add it. :-)

PR #4400, opened 2024-10-02
[persistence] Allow negative filter boundaries

### LINKED ISSUE #4396: Persistence Include/Exclude Filters don't accept negative numbers in textual configuration
## Expected Behavior

I want to configure an [Include Filter](https://www.openhab.org/docs/configuration/persistence.html#include-filter-exclude-filter) in my textual persistence configuration with a range of -8000 to 32000. I therefore added this filter definition in my `influx.persist` file:

```
Filters {
    stromLeistungRange : [] -8000 32000
}
```

## Current Behavior
When I add the above mentioned Include Filter to the textual persistence configuration, openHAB reports this file as invalid:

```
2024-10-01 20:52:27.197 [WARN ] [el.core.internal.ModelRepositoryImpl] - Configuration model 'influx.persist' has errors, therefore ignoring it: [6,29]: extraneous input '-' expecting RULE_INT
``` 

## Possible Solution

Allow negative numbers in the textual persistence configuration.

Put a `*.persist` file into the `persistence` directory of the openHAB configuration and save it with this content:

```
Strategies {
    default = everyChange
}

Filters {
    stromLeistungRange : [] -8000 32000
}
```

## Context

My solar inverter sometimes reports invalid values at the min or max value of the integer range. To prevent this to be stored in my persistence backend (InfluxDB) I want to only accept the range of -8000 to 32000 Wh.

As I run my openHAB instance as docker container and store the configuration in a Git repository, I don't want to use the Main UI to configure openHAB. Therefore I rely on the `*.persist` file to define such a filter.

## Your Environment

* Version used: 4.2.1 - Release Build

holgerfriedrich:
Modify xtext file for textual configuration of persistence.
Fixes #4396.

I have tested that OH now accepts the negative values when parsing the configuration.
I have not checked the filtering itself.

### DIFF
--- a/bundles/org.openhab.core.model.persistence/src/org/openhab/core/model/persistence/Persistence.xtext
+++ b/bundles/org.openhab.core.model.persistence/src/org/openhab/core/model/persistence/Persistence.xtext
@@ -76,5 +76,5 @@ GroupConfig:
 ;
 
 DECIMAL returns ecore::EBigDecimal :
-	INT ('.' INT)?
+	'-'? INT ('.' INT)?
 ;


wborn (approved):
Thanks for the fix!

PR #4401, opened 2024-10-03
Add ColorUtil.xyToDuv

ccutrer:
This adds a new helper method to calculate the Duv (Delta u, v) of a color. The code is taken almost literally from the mqtt.espmilighthub binding, but changed slightly to fit in with other functions in ColorUtil. I intend to use this from espmilighthub, but also the HomeKit add-on. HomeKit doesn't allow linking both HSB _and_ Color Temperature characteristics to a single accessory, even though the iOS Home App will present a Color Temperature picker for any HSB light. So I want to allow linking devices that support both RGB+CCT to HomeKit, and if the Duv is within threshold, send a command to the Color Temperature channel instead.

I also think it might be useful to add a profile to apply to a Color channel link that does the same thing - if a color is close enough to "white", send a command to the color temperature item instead. This would cover all other use cases, such as BasicUI and MainUI.

This can also be useful for user rules dealing with RGB+CCT light.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
@@ -539,24 +539,24 @@ public class ColorUtil {
      * Duv describes the distance of a color point from the black body curve. It's useful for calculating
      * if a color is near to "white", at any color temperature.
      * 
-     * @param xy array of double with CIE 1931 x,y[,Y] in the range 0.0000 to 1.0000 <code>Y</code> value is optional.
+     * @param xy array of double with CIE 1931 x,y in the range 0.0000 to 1.0000
      * @return the calculated Duv metric
      * @throws IllegalArgumentException when input array has wrong size or exceeds allowed value range.
      */
-    public static double xyToDuv(double[] xyY) throws IllegalArgumentException {
-        if (xyY.length < 2 || xyY.length > 4) {
-            throw new IllegalArgumentException("xyToHsb() requires 2, 3 or 4 arguments");
+    public static double xyToDuv(double[] xy) throws IllegalArgumentException {
+        if (xy.length != 2) {
+            throw new IllegalArgumentException("xyToDuv() requires 2 arguments");
         }
 
-        for (int i = 0; i < xyY.length; i++) {
-            if (xyY[i] < 0 || xyY[i] > 1) {
+        for (int i = 0; i < xy.length; i++) {
+            if (xy[i] < 0 || xy[i] > 1) {
                 throw new IllegalArgumentException(
-                        String.format("xyToHsb() argument %d value '%f' out of range [0..1.0]", i, xyY[i]));
+                        String.format("xyToDuv() argument %d value '%f' out of range [0..1.0]", i, xy[i]));
             }
         }
 
-        var x = BigDecimal.valueOf(xyY[0]);
-        var y = BigDecimal.valueOf(xyY[1]);
+        var x = BigDecimal.valueOf(xy[0]);
+        var y = BigDecimal.valueOf(xy[1]);
         var u = BIG_DECIMAL_4.multiply(x).divide(
                 BIG_DECIMAL_2.multiply(x).negate().add(BIG_DECIMAL_12.multiply(y).add(BIG_DECIMAL_3)),
                 MathContext.DECIMAL128);
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
@@ -1225,11 +1225,20 @@ public class ColorUtil {
         return (437 * Math.pow(n, 3)) + (3601 * Math.pow(n, 2)) + (6861 * n) + 5517;
     }
 
+    /**
+     * Calculates a polynomial regression.
+     *
+     * This calculates the equation K[4]*x^0 + K[3]*x^1 + K[2]*x^2 + K[1]*x^3 + K[0]*x^4
+     *
+     * @param x The independent variable distributed through each term of the polynomial
+     * @param coefficients The coefficients of the polynomial. Note that the terms are in
+     *            order from largest exponent to smallest exponent, which is the reverse order
+     *            of the usual way of writing it in academic papers
+     * @return the result of substituting x into the regression polynomial
+     */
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -523,6 +532,45 @@ public static HSBType xyToHsb(double[] xyY, Gamut gamut) throws IllegalArgumentE
...
+     * Duv describes the distance of a color point from the black body curve. It's useful for calculating
+     * if a color is near to "white", at any color temperature.
+     * 
+     * @param xy array of double with CIE 1931 x,y[,Y] in the range 0.0000 to 1.0000 <code>Y</code> value is optional.
+     * @return the calculated Duv metric
+     * @throws IllegalArgumentException when input array has wrong size or exceeds allowed value range.
+     */
+    public static double xyToDuv(double[] xyY) throws IllegalArgumentException {
+        if (xyY.length < 2 || xyY.length > 4) {
+            throw new IllegalArgumentException("xyToHsb() requires 2, 3 or 4 arguments");
```
jimtng, line 548: xyToDuv?

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -523,6 +532,45 @@ public static HSBType xyToHsb(double[] xyY, Gamut gamut) throws IllegalArgumentE
...
+        }
+
+        for (int i = 0; i < xyY.length; i++) {
+            if (xyY[i] < 0 || xyY[i] > 1) {
+                throw new IllegalArgumentException(
+                        String.format("xyToHsb() argument %d value '%f' out of range [0..1.0]", i, xyY[i]));
```
jimtng, line 554: Ditto here

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -1176,4 +1224,16 @@ public static double xyToKelvin(double[] xy) {
         double n = (xy[0] - 0.3320) / (0.1858 - xy[1]);
         return (437 * Math.pow(n, 3)) + (3601 * Math.pow(n, 2)) + (6861 * n) + 5517;
     }
+
+    private static BigDecimal polynomialFit(BigDecimal x, BigDecimal[] coefficients) {
```
holgerfriedrich, line 1228: Could you please add Javadoc?  We did the same for other private methods as well. The comment below should also be mentioned here.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -1176,4 +1224,25 @@ public static double xyToKelvin(double[] xy) {
...
+     *            order from largest exponent to smallest exponent, which is the reverse order
+     *            of the usual way of writing it in academic papers
+     * @return the result of substituting x into the regression polynomial
+     */
+    private static BigDecimal polynomialFit(BigDecimal x, BigDecimal[] coefficients) {
+        var result = BigDecimal.ZERO;
+        var xAccumulator = BigDecimal.ONE;
+        for (int i = coefficients.length - 1; i >= 0; i--) {
+            result = result.add(coefficients[i].multiply(xAccumulator));
+            xAccumulator = xAccumulator.multiply(x);
```
holgerfriedrich, line 1244:
```suggestion
    private static double polynomialFit(double x, double[] coefficients) {
        double result = 0.0;
        double xAccumulator = 1.0;
        for (int i = coefficients.length - 1; i >= 0; i--) {
            result = result + coefficients[i] * xAccumulator;
            xAccumulator = xAccumulator * x;

```

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -46,10 +47,18 @@ public class ColorUtil {
...
     private static final BigDecimal BIG_DECIMAL_5 = BigDecimal.valueOf(5);
+    private static final BigDecimal BIG_DECIMAL_4 = BigDecimal.valueOf(4);
     private static final BigDecimal BIG_DECIMAL_3 = BigDecimal.valueOf(3);
     private static final BigDecimal BIG_DECIMAL_2 = BigDecimal.valueOf(2);
     private static final BigDecimal BIG_DECIMAL_2_POINT_55 = new BigDecimal("2.55");
+    private static final BigDecimal BIG_DECIMAL_0_POINT_292 = new BigDecimal("0.292");
+    private static final BigDecimal BIG_DECIMAL_0_POINT_24 = new BigDecimal("0.24");
+    private static final BigDecimal[] CORM_COEFFICIENTS = { new BigDecimal("-0.00616793"), new BigDecimal("0.0893944"),
+            new BigDecimal("-0.5179722"), new BigDecimal("1.5317403"), new BigDecimal("-2.4243787"),
+            new BigDecimal("1.925865"), new BigDecimal("-0.471106") };
```
holgerfriedrich, line 61:
```suggestion
    private static final double[] CORM_COEFFICIENTS = { -0.00616793, 0.0893944, -0.5179722,
1.5317403, -2.4243787,
            1.925865, -0.471106 };
```

holgerfriedrich (commented):
Looking again at this PR, I am asking myself why it is necessary to use BigDecimal for the computations - especially using MathContext.DECIMAL128 which ensures that at least 34 digits are used.
I did a quick rewrite to double, and the tests still passed.

So is there a reason to use the computationally much more demanding BigDecimals? Are you aware of any instability caused by special input values, etc.?

I added my implementation as comments, please feel free to resolve those conversations without change in case you have a good reason to go for the BigDecimals.

ccutrer:
Nope, no good reason. I suspect the original code started out as float (single precision), and there _were_ accuracy issues, and so it jumped all the way to BigDecimal when double would have been plenty sufficient.

holgerfriedrich:
> Nope, no good reason. I suspect the original code started out as float (single precision), and there _were_ accuracy issues, and so it jumped all the way to BigDecimal when double would have been plenty sufficient.

Then I would prefer the implementation based on double. It has way better performance and is easier to read.

Maybe you could remove all unused BigDecimal constants as well.
Thanks!

holgerfriedrich (approved):
LGTM, thanks!

andrewfg:
@ccutrer for the esp light bulb you may want to look at my recently merged PR in the Zigbee binding.

PR #4367, opened 2024-08-31
[ColorUtil] New methods kelvinToXY and xyToKelvin

andrewfg:
Resolves #4369

This PR provides two new methods in the `ColorUtils` class as follows:
1. New method `kelvinToXY()` to convert a colour temperature in Kelvin to a point on the colour temperature locus in the CIE XY colour space, as described [here](https://www.waveformlighting.com/tech/calculate-cie-1931-xy-coordinates-from-cct)
2. New method `xyToKelvin()` to convert a point on the colour temperature locus in the CIE XY colour space to a colour temperature in Kelvin, as described [here](https://www.waveformlighting.com/tech/calculate-color-temperature-cct-from-cie-1931-xy-coordinates)

image

Note: the PR is created in ColorUtils in OH Core in case these conversions might be needed in other bindings.

Signed-off-by: AndrewFG <software@whitebear.ch>

PR #4410, opened 2024-10-07
[DSL] Expose ColorUtil methods to DSL rules

holgerfriedrich:
* Expose xyToKelvin and kelvinToXY
* Add missing throws declaration
* Expose xyToDuv

Unfortunately, we forgot about the throws declaration in #4367. This is API breaking, you might need to add a catch statement in binding code. We did not find any issues in the add-on repo so far, it still compiles fine.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
@@ -1651,9 +1651,9 @@ public class ColorUtil {
      * 
      * @param xy an array with the CIE colour XY values to be converted
      * @return the colour temperature in K
-     * @throws IndexOutOfBoundsException if the wrong number of arguments is provided
+     * @throws IllegalArgumentException if the wrong number of arguments is provided
      */
-    public static double xyToKelvin(double[] xy) {
+    public static double xyToKelvin(double[] xy) throws IllegalArgumentException {
         if (xy.length != 2) {
             throw new IllegalArgumentException("xyToKelvin() requires 2 arguments");
         }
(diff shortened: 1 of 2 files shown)


andrewfg (approved):
LGTM many thanks.

holgerfriedrich:
Lets wait for #4401, then we can add ```xyToDuv``` as well.

holgerfriedrich:
Done.
I will create an additional PR in docs.
@florian-h05 do you want to create a change in JS scripting as well (follow-up to openhab/openhab-js#319)?

florian-h05:
These methods are just added to the CoreUtil class, correct?
In that case only a doc update in openhab-js is needed.

andrewfg:
@holgerfriedrich apropos the thrown exception potentially breaking the API — two things..
1. In https://github.com/openhab/org.openhab.binding.zigbee/pull/855 I did not call the respective method, so that one will not be broken.
2. Did you check the ColorUtil JUnit tests (here in core) since they do call the respective method?

holgerfriedrich:
@andrewfg thanks for checking, the build seems still fine. I think in tests it is not mandatory to catch all declared exceptions....

andrewfg (approved):
LGTM

holgerfriedrich:
Rebased to let CI run on ubuntu-24.04

lolodomo (approved):

J-N-K (approved):
Thanks!

bundles/org.openhab.core/src/main/java/org/openhab/core/util/ColorUtil.java
```diff
@@ -1651,9 +1651,9 @@ public static double[] kelvinToXY(double kelvin) throws IndexOutOfBoundsExceptio
      * 
      * @param xy an array with the CIE colour XY values to be converted
      * @return the colour temperature in K
-     * @throws IndexOutOfBoundsException if the wrong number of arguments is provided
+     * @throws IllegalArgumentException if the wrong number of arguments is provided
      */
-    public static double xyToKelvin(double[] xy) {
+    public static double xyToKelvin(double[] xy) throws IllegalArgumentException {
```
J-N-K, line 1656: Since an IAE is a runtime exception this should not break code.

PR #4415, opened 2024-10-16
Improve marketplace remote handling

J-N-K:
This improves the handling of remote add-ons. In some corner cases the list of remote add-ons was not properly refreshed after an add-on has been removed.


wborn (approved):
Thanks for the fix!

PR #4417, opened 2024-10-20
Parse BigInteger directly in DecimalType

magx2:
Instead of running expensive `BigInteger.toString()` and then parsing `BigDecimal` from it, run direclty constructor `BigDecimal(BigInteger)`


J-N-K (approved):
Thanks!

J-N-K:
Can you fix spotless?

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DecimalType.java
```diff
@@ -57,6 +58,8 @@ public DecimalType(Number value) {
             this.value = type.toBigDecimal();
         } else if (value instanceof BigDecimal decimal) {
             this.value = decimal;
+        }else if (value instanceof BigInteger integer) {
```
J-N-K, line 61:
```suggestion
        } else if (value instanceof BigInteger integer) {
```

PR #4419, opened 2024-10-21
ActionInputsHelper: Add unit symbol to Quantity parameter description

florian-h05:
Useful for the UI.


florian-h05:
@J-N-K Can you please take a look at this minor change since you reviewed #4392?

lolodomo:
Why do you want to mix-up description and unit ?
You have already the unit in a separate field.

florian-h05:
Since QuantityType is currently described as decimal parameter, the UI does not need to handle it in a special way and therefore does not use the unit field.
However, the user should know which unit he enters, and this IMO is more efficient handled in core.

J-N-K:
I tend to agree with @lolodomo. Concatenating description and unit is as easy on client side as on server side. Depending on the UI (which may not only be MainUI), it would be better to have the description and unit separated.

florian-h05:
I will create a UI PR then 👍

PR #4427, opened 2024-10-26
[automation] Allow returning null from supplier function in ValueCache

florian-h05:
In some cases, it is useful to be able to return null from the supplier function using null as an "unspecified" value.

This is a proposal, please let me know what you think @J-N-K.

### DIFF
--- a/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/CacheScriptExtension.java
+++ b/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/internal/CacheScriptExtension.java
@@ -172,8 +172,8 @@ public class CacheScriptExtension implements ScriptExtensionProvider {
         }
 
         @Override
-        public Object get(String key, Supplier<Object> supplier) {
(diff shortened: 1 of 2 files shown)


J-N-K:
What would be the use-case for that? My intention when writing this code was to make it work like `Optional.orElseGet` (https://docs.oracle.com/javase/10/docs/api/java/util/Optional.html#orElseGet(java.util.function.Supplier)), but I'm open for suggestions here.

florian-h05:
I currently cannot really think of the exact use case, but I regularly return null in my JS rules from functions from the JS perspective it would be nice to have.
FYI I have mainly opened this PR here because of https://github.com/openhab/openhab-js/issues/386.

jimtng:
According to: https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Map.html#computeIfAbsent(K,java.util.function.Function)

> If the mapping function returns null, no mapping is recorded.

So if we were to allow get(key, supplier) to return null, does it mean a new mapping is stored having `null` as its value, or that it didn't store the mapping?

I think the documentation should be updated to reflect this instead of changing the behaviour. Throwing an exception is probably the correct behaviour.

#4704 relies on the fact that if the supplier function returns null, it means a request to remove the entry from the map. Which means we expect that the map doesn't store `null` values.

PR #4428, opened 2024-10-26
ConfigDescriptionParameter: Change default format for datetime & Update context docs

florian-h05:
This changes the default format for the datetime context to the ISO standard. This context is not used by add-ons and supported by the UI, so it should be possible to change it (again after #4392.)

Also update the context docs from https://next.openhab.org/docs/developer/addons/config-xml.html#supported-contexts and the UI code.


florian-h05:
@lolodomo Let's change the format for the `datetime` context to use the ISO default format as this does not require to use a custom formatter anywhere that context is used.

J-N-K (approved):
LGTM

J-N-K:
I agree, ISO date strings are more likely to be expected and used in this context.

Issue #3891, opened 2023-11-26
[Sitemap] Widget for color temperature


mueller-ma:
### Is your feature request related to a problem? Please describe.

For some time I created two sliders for lightbulbs that can change their color temperature. One for brightness, one for color temperature.
Recently I acquired a Nanoleaf Elements light (LEDs with brightness and color temperature settings). The binding allows to use a color picker to control both brightness and color temperature settings. All colors expect in the range cold white to warm white are ignored by the binding, but I can control both with just one widget.

### Describe the solution you'd like

Add a new option to the `Color` widget (e.g. `temperatureOnly`) that limits the color selection from cold white to warm white.

Example from the Ikea Tradfri app:
![tradfri](https://github.com/openhab/openhab-core/assets/22525368/ca066cb6-1936-4817-a89a-7e2d770b83c1)

### Coordination between maintainers

Notify maintainers of other UIs:
@openhab/webui-maintainers 
@openhab/android-maintainers 
@openhab/ios-maintainers 

Checklist for implementation:
- [x] Core
- [x] Basic UI
- [x] Main UI
- [x] Android app
- [ ] iOS app
- [x] Docs

lolodomo:
For a color item, we have already the colorpicker widget to select the color, the slider widget to select its brightness and the switch widget to turn it on or off.

For color temperature, we could imagine either a new widget or an option to the existing colorpicker widget. In both cases, we need to implement a new UI in all apps. It could be a simple slider ? As a min and max values are probably required, a new widget similar to slider widget would be better I believe. We could name it colortemperature.

lolodomo:
To be thought how an UI can then build the command to be sent to the server.

maniac103:
> The state is transmitted as HSB and the binding then has to parse brightness and color temperature from that single state.

Color temperature channels are either Dimmers or Numbers though, not Colors, so I don't think that idea will fly.

For controlling color temperature channels, the client would need to differentiate between dimmer (relative) or absolute channels (by item type):
- Dimmer items have a slider range of 0..100; we'd need to show some generic gradient for those. Commands sent are ordinary percentages.
- Number items should have min and max in their state description, so we could calculate the precise colors for the gradient. Commands sent equal those of normal number sliders

mueller-ma:
Ok, so then it's required to add two items to one widget, like `ColorTemperature brightnessItem=foo temperatureItem=bar`. This is probably the first widget with more than one item. Is this an argument against it?

[...]

PR #4429, opened 2024-10-26
[ColorUtil] Extend Colour Temperature range 1000 K .. 10000 K

andrewfg:
Extend the supported colour temperature range of the `kelvinToXy()` method from 2000 .. 6500 K to 1000 .. 10000 K

See https://github.com/openhab/openhab-core/issues/3891#issuecomment-2428965622

Please note that the inverse `xyToKelvin()` method (which is based on McCamy's approximation) is only accurate above 2000 K. This means that round trip conversions `xyToKelvin(kelvinToXy())` are very inaccurate for colour temperatures below 2000 K. This is not perfect but it seems that most real bindings in OH are using a range of 2000 K to 6500 K .. where the accuracy is better than 1%..

Signed-off-by: AndrewFG <software@whitebear.ch>


andrewfg:
=> @lolodomo for info

holgerfriedrich (commented):
Hi @andrewfg, 
could we consider restricting the number of tests runs to a reasonable number? 270000 tests take quite a while to complete.
Mybe increasing ```allHSB::step``` to 20 and ```allRGBW::step``` to 25 could be a compromise.... WDYT?

andrewfg:
Yeah. :)

bundles/org.openhab.core/src/test/java/org/openhab/core/util/ColorUtilTest.java
```diff
@@ -435,7 +435,7 @@ private static Stream<Arguments> allXY() {
      */
     private static Stream<Arguments> allRGBW() {
         List<Arguments> result = new ArrayList<>();
-        final double step = 5.0;
+        final double step = 17.0;
```
holgerfriedrich, line 438: Do you want to hit the upper boundary 100? Then a factor of 100 would be a better choice.
With step size 17, the loop will end at 85.

bundles/org.openhab.core/src/test/java/org/openhab/core/util/ColorUtilTest.java
```diff
@@ -389,7 +389,7 @@ public Stream<? extends Arguments> provideArguments(@Nullable ExtensionContext c
      */
     private static Stream<Arguments> allHSB() {
         List<Arguments> result = new ArrayList<>();
-        final double step = 5.0;
+        final double step = 17.0;
```
holgerfriedrich, line 392: Do you want to hit the upper boundaries 100 and 360? Then a factor of 100 and 360 would be a better choice.
With step size 17, the loop will end at 85 and 357, respectively. 10 or 20 might be a better choice.

bundles/org.openhab.core/src/test/java/org/openhab/core/util/ColorUtilTest.java
andrewfg: I know. I was trying to set a step that tests intersticial values between the obvious round numbers. So I was thinking of using primes. But perhaps a factor of 100 would indeed be better..

bundles/org.openhab.core/src/test/java/org/openhab/core/util/ColorUtilTest.java
andrewfg: See above.

holgerfriedrich (approved):
LGTM, thanks!

PR #4430, opened 2024-10-27
Thing actions: Support `@ActionOutput` for single return values

florian-h05:
See discussion in https://github.com/openhab/openhab-addons/issues/17504#issuecomment-2439906483.

This adds processing of the ActionOutput annotation for Thing actions with a single return value, which allows providing a label for use in the UI. The output name for those actions is "result", which is now the default value in the @ActionOutput annotation. If a binding overrides the default name, a warning is logged.

If a Thing action returns a Map<String, Object> but does not provide the @ActionOutputs annotation, a warning is logged.


J-N-K (approved):
LGTM. Thanks!

PR #4433, opened 2024-10-29
[Units] Add MK⁻¹ unit alias

andrewfg:
See discussion https://github.com/openhab/openhab-addons/issues/17638#issuecomment-2444023328

Signed-off-by: AndrewFG <software@whitebear.ch>

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/Units.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/Units.java
@@ -293,7 +293,7 @@ public final class Units extends CustomUnits {
         SimpleUnitFormat.getInstance().label(MILLIAMPERE_HOUR, "mAh");
         SimpleUnitFormat.getInstance().label(MILLIBAR, "mbar");
         SimpleUnitFormat.getInstance().label(MILLIMETRE_OF_MERCURY, MILLIMETRE_OF_MERCURY.getSymbol());
-        SimpleUnitFormat.getInstance().label(MIRED, "mired");
+        SimpleUnitFormat.getInstance().label(MIRED, "MK⁻¹");
         SimpleUnitFormat.getInstance().label(PARTS_PER_BILLION, "ppb");
         SimpleUnitFormat.getInstance().label(PARTS_PER_MILLION, "ppm");
         SimpleUnitFormat.getInstance().label(PETABYTE, "PB");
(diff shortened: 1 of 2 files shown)


andrewfg:
@maniac103 for info..

ccutrer:
Should we also be adding `mirek` as an alias? And perhaps changing the proper name of the unit to `MK⁻¹ `, with `mired` as the alias?

andrewfg:
^
I would support that.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/Units.java
```diff
@@ -319,6 +319,10 @@ public final class Units extends CustomUnits {
         SimpleUnitFormat.getInstance().alias(MONTH, "mo");
         SimpleUnitFormat.getInstance().alias(WEEK, "wk");
         SimpleUnitFormat.getInstance().alias(YEAR, "y");
+
+        // other aliases
+        SimpleUnitFormat.getInstance().alias(MIRED, "mired");
+        SimpleUnitFormat.getInstance().alias(MIRED, "mirek");
```
jlaur, line 325: These could probably be moved up right after line 296? All the others have their alias right after the label, except for the few works-arounds just above.

andrewfg: Ok. Done.

lolodomo:
I followed the discussion and it looks good to me too.

lolodomo:
If I have a String `myUnit` containing a unit, what is the code to use to check that is a unit in "mired" whatever the accepted symbol ?
If I use `Units.MIRED.toString().equals(myUnit)`, this will probably work only with "mired" with the current code and only "MK⁻¹" when this PR is merged.

andrewfg:
> If I have a String myUnit containing a unit, what is the code to use to check that is a unit in "mired" whatever the accepted symbol ?

Umm. From memory I believe it is the following. .. But perhaps I should add a Junit test to confirm that works?

```
Unit unit = QuantityType.valueOf(yourString).getUnit();
```

andrewfg:
@lolodomo I just added the Junit test (it passes).

ccutrer:
Or `UnitUtils.parseUnit` if you only have the unit, without a scalar for a full quantity

lolodomo:
> Or `UnitUtils.parseUnit` if you only have the unit, without a scalar for a full quantity

Oh  yes, looks better, thank you.

PR #4434, opened 2024-10-30
[rest] VoiceResource: Return answer from /interpreters endpoint

florian-h05:
Also add annotations for the response.

### DIFF
--- a/bundles/org.openhab.core.io.rest.voice/src/main/java/org/openhab/core/io/rest/voice/internal/VoiceResource.java
+++ b/bundles/org.openhab.core.io.rest.voice/src/main/java/org/openhab/core/io/rest/voice/internal/VoiceResource.java
@@ -139,8 +139,9 @@ public class VoiceResource implements RESTResource {
     @POST
     @Path("/interpreters/{ids: [a-zA-Z_0-9,]+}")
     @Consumes(MediaType.TEXT_PLAIN)
+    @Produces(MediaType.TEXT_PLAIN)
     @Operation(operationId = "interpretText", summary = "Sends a text to a given human language interpreter(s).", responses = {
-            @ApiResponse(responseCode = "200", description = "OK"),
+            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = String.class))),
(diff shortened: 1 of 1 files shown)


florian-h05:
As the current answer was undocumented, WDYT about changing the format to JSON and putting the current answer string into a answer key?
My plans for the UI need extended core APIs ...

wborn:
> WDYT about changing the format to JSON and putting the current answer string into a answer key?

I wouldn't change it now unless there is a good reason and it is actually used for that. Seems like this API has been around  since OH 2.x so changing it will most likely may break some user integrations.

bundles/org.openhab.core.io.rest.voice/src/main/java/org/openhab/core/io/rest/voice/internal/VoiceResource.java
```diff
@@ -189,8 +191,8 @@ public Response interpret(
         }
 
         try {
-            hli.interpret(locale, text);
-            return Response.ok(null, MediaType.TEXT_PLAIN).build();
+            String answer = hli.interpret(locale, text);
+            return Response.ok(answer, MediaType.TEXT_PLAIN).build();
```
wborn, line 195: Did it always return null?

florian-h05:
My problem is, that I not only need to return an answer from the HLI but also some „action“ data to instruct the UI to show a widget for some Item, but that would also need changes to the HLI interface. Not sure how to handle that.

bundles/org.openhab.core.io.rest.voice/src/main/java/org/openhab/core/io/rest/voice/internal/VoiceResource.java
florian-h05: What to you mean?

wborn:
It might be possible to support both if you add another method with the `@Consumes(MediaType.APPLICATION_JSON)` annotation and keep the method having `@Consumes(MediaType.TEXT_PLAIN)`. :wink:

florian-h05:
Thanks for the hint — I will try that out then, but let’s get this merged first.

wborn (approved):
Thanks!

PR #4440, opened 2024-11-06
[automation] ActionInputsHelper: Enhance ZonedDateTime & Instant support

florian-h05:


florian-h05:
@lolodomo Do you agree?

lolodomo:
> @lolodomo Do you agree?

This makes the filling of date & time easier in UI but the user cannot set a specific time zone in that case.

Ok

kaikreuzer (approved):
lgtm

PR #4444, opened 2024-11-10
Improve FolderObserver

### LINKED ISSUE #3823: Thing manager logic causes things to stuck in NOT_YET_READY
I'm in process of updating bindings to OH 4 and spotted a bug, which I believe is related to new logic which improves reliability and predictability of thing startup process.

Bindings I have are rather basic in their form, one which I test with use no dynamic (generated) channels or thing types. Its entirely provided through XML descriptors. However, OH 4.0 makes it stuck in "NOT_YET_READY" state, despite of all descriptors being available and valid.

## Expected Behavior
Valid thing types should cause handlers to be initialized.

## Current Behavior
Despite of thing descriptors being properly parsed and available thing is marked as not ready, because its handler is refused.
I've traced it to a logic which populated "missing prerequisites". It turns that this collection is being feed in few places, but released only in one. This means that depending on startup order (and performance of machine), some handlers may never be used, because thing manager will miss opportunity to remove missing prerequisites, even if these conditions are met.
Please see below screenshot where missing prerequisites contains `co7io-amsads:network`, but thing thing type is already registered in appropriate registry:

image

So far I was not able to get a reproducer, as it is strictly speaking, dependent to startup order. If thing is added before thing descriptor is parsed - there is high probability that `ThingManager` will never update solved prerequisites. 

## Your Environment
* openHAB 4.0.3
* JVM Temurin 17
* Linux 6.5

J-N-K:
Party fixes #3823

It has been reported that missing model directories cause issued during startup. With this PR a directory is created when it is missing.

### DIFF
--- a/bundles/org.openhab.core.model.core/src/main/java/org/openhab/core/model/core/internal/folder/FolderObserver.java
+++ b/bundles/org.openhab.core.model.core/src/main/java/org/openhab/core/model/core/internal/folder/FolderObserver.java
@@ -96,6 +96,14 @@ public class FolderObserver implements WatchService.WatchEventListener {
         logger.debug("Adding parser for '{}' extension", extension);
         parsers.add(extension);
 
+        // check if the desired directory exists and create it if it's missing
+        try {
+            Path extensionPath = watchService.getWatchPath().resolve(extension);
+            Files.createDirectories(extensionPath);
+        } catch (IOException e) {
+            logger.error("Model path for extension '{}' is missing and creation failed: {}", extension, e.getMessage());
+        }
+
         if (activated) {
             processIgnoredPaths(extension);
             readyService.markReady(new ReadyMarker(READYMARKER_TYPE, extension));


lolodomo:
Looks good

J-N-K:
No, Not complete

lolodomo:
Is it something we can include in 4.3 or is it still not yet ready?

J-N-K:
Unfortunately I can't remember what is missing. IMO this is an improvement, even if it does not fully solve the issue. IIRC there are some corner-cases where it still does not work.

holgerfriedrich:
@J-N-K I remember that I had seen the discussion in #3823 if this should be solved by creating the missing folders or learn more about the root cause. This is why I did not merge it at that time.

So shall we proceed merging this one to improve the situation?
If yes, maybe we should add a log message in case the directory is created, just as hint that something unexpected happened.
WDYT?

kaikreuzer:
I have a general question here: The FolderObserver watches folders in the `conf` folder, right?
This folder is to be treated as read-only by openHAB (in contrast to the `userdata` folder), so creating directories in here does not look like the right approach to me.

lolodomo:
Very good remark.
Maybe we should just log a warning to inform users that the folder is missing.
PS: it concerns only users using docker.

bundles/org.openhab.core.model.core/src/main/java/org/openhab/core/model/core/internal/folder/FolderObserver.java
```diff
@@ -96,6 +96,14 @@ protected void addModelParser(ModelParser modelParser) {
         logger.debug("Adding parser for '{}' extension", extension);
         parsers.add(extension);
 
+        // check if the desired directory exists and create it if it's missing
+        try {
+            Path extensionPath = watchService.getWatchPath().resolve(extension);
+            Files.createDirectories(extensionPath);
+        } catch (IOException e) {
+            logger.error("Model path for extension '{}' is missing and creation failed: {}", extension, e.getMessage());
+        }
```
J-N-K, line 105: So the suggestion is 

```
Path extensionPath = watchService.getWatchPath().resolve(extension);
if (!Files.isDirectory(extensionPath)) {
  logger.error("Model path for extension '{}' is missing or not accessible. Please create it and restart openHAB.", extension);
}
```

wborn:
> Maybe we should just log a warning to inform users that the folder is missing.
> PS: it concerns only users using docker.

Yes some users just copy a few of their directories and think everything else will then magically work.
It's an issue for any distro so perhaps we should add something to the OH startup shell/batch files to detect this.

Nadahar:
I don't fully understand what causes this problem, but as one user explains in #3823:

> So, for openhab to launch correctly, the openhab/conf/items folder must be present. If it is not there, then the problems will be described above. My system does not have this folder, items are stored in the file openhab/userdata/jsondb/org.openhab.core.items.Item.json. After an empty folder has been created. All things changed the status from NOT_YET_READY to Online the next time openhab was launched.

If no Items files are in use, why must this folder exist? It doesn't seem logical to me, whether it's created or logged, it seems like the real problem is why the lack of this folder causes the system initialization to fail.

lolodomo:
I believe we should add the error log for OH 5 release.
@J-N-K seems to be no more active so maybe we can create another PR and close this one? I can do it tomorrow.

PR #4445, opened 2024-11-10
[rest] Fix ThingActionsResource does not accept new Thing action UIDs with hash

florian-h05:
Regression from #4441.

### DIFF
--- a/bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/ThingActionsResource.java
+++ b/bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/ThingActionsResource.java
@@ -217,7 +217,9 @@ public class ThingActionsResource implements RESTResource {
 
     @POST
     @RolesAllowed({ Role.USER, Role.ADMIN })
-    @Path("/{thingUID}/{actionUid: [a-zA-Z0-9]+\\.[a-zA-Z0-9]+}")
+    // accept actionUid in the form of "scope.actionTypeUid" or "scope.actionTypeUid#signatureHash"
+    // # is URL encoded as %23
+    @Path("/{thingUID}/{actionUid: [a-zA-Z0-9]+\\.[a-zA-Z0-9]+(%23[a-zA-Z0-9]+)?}")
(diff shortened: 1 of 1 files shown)


lolodomo:
Probably correct

bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/ThingActionsResource.java
```diff
@@ -217,7 +217,9 @@ public Response getActions(@PathParam("thingUID") @Parameter(description = "thin
 
     @POST
     @RolesAllowed({ Role.USER, Role.ADMIN })
-    @Path("/{thingUID}/{actionUid: [a-zA-Z0-9]+\\.[a-zA-Z0-9]+}")
+    // accept actionUid in the form of "scope.actionTypeUid" or "scope.actionTypeUid#signatureHash"
+    // # is URL encoded as %23
+    @Path("/{thingUID}/{actionUid: [a-zA-Z0-9]+\\.[a-zA-Z0-9]+(%23[a-zA-Z0-9]+)?}")
```
J-N-K, line 222: The hash is a hex string, correct? In that case `A-Fa-f0-9` would be sufficient. Since we do a lookup on the hash and that is case-sensitive it Coups be a good idea to also restrict that.

J-N-K (changes requested):

bundles/org.openhab.core.automation.rest/src/main/java/org/openhab/core/automation/rest/internal/ThingActionsResource.java
florian-h05: Makes sense, just implemented and pushed that.
Local test looks good.

J-N-K (approved):

florian-h05:
Can someone please trigger a new snapshot build?

kaikreuzer:
It's already done: https://ci.openhab.org/job/openHAB-Core/58/changes

kaikreuzer:
Now there's also a new distro for testing: https://ci.openhab.org/job/openHAB-Distribution/4374/

Issue #4431, opened 2024-10-28
Sitemap icon rules / icon colors doesnt work when code is generated by MainUI


pergolafabio:
Hi, as discussed on this community topic here:
https://community.openhab.org/t/cant-get-iconcolor-to-work/159856/36

Seems icon rules and icon colors are not working when generated with MainUI:

example: icon color:
this works using code
`Switch icon=if:mdi:lightbulb item=licht_nachthal label="Nachthal" mappings=[ON="Aan", OFF="Uit"] iconcolor=[ON="yellow"]
`
But it doesnt work when using mainUI, i need to add the additional item name:
`Switch icon=if:mdi:lightbulb item=licht_nachthal label="Nachthal" mappings=[ON="Aan", OFF="Uit"] iconcolor=[licht_nachthal==ON="yellow"]`

example: icon rules
this works using code:
`Switch item=cover_velux_links mappings=[UP="Omhoog ▲", DOWN="Omlaag ▼"] iconcolor=[100="red", 0="green"] icon=[100="if:cbi:tilt-window-closed"] label="Velux Links"`

when the code is generated in MainUI:

this doesnt work:
`Switch item=cover_velux_links mappings=[UP="Omhoog ▲", DOWN="Omlaag ▼"] iconcolor=[100="red", 0="green"] icon=[100="if:cbi:tilt-window-closed"] label="Velux Links"`

Also adding the itemname here doesnt work....
like:
`Switch item=cover_velux_links mappings=[UP="Omhoog ▲", DOWN="Omlaag ▼"] iconcolor=[100="red", 0="green"] icon=[cover_velux_links==100="if:cbi:tilt-window-closed"] label="Velux Links"`

## Your Environment

* Version used: 4.2.2
* Environment: Docker

lolodomo:
Please rename the title "BasicUI icon rules..." into "Sitemap icon rules...".

PR #4449, opened 2024-11-13
Sitemap editor: Fix color and icon rules

mherwege:
Closes https://github.com/openhab/openhab-core/issues/4431

This closes an issue with color and icon rules when no item and comparator was provided in the rule (should default to item of the widget and == comparator) and the rule condition started with an alpha character.


lolodomo:
I don't understand the case that was wrong. Can you please provide an example of string that was not matched ?

mherwege:
> I don't understand the case that was wrong. Can you please provide an example of sitemap element that was not matched ?

@lolodomo Look at the linked issue.

`Switch icon=if:mdi:lightbulb item=licht_nachthal label="Nachthal" mappings=[ON="Aan", OFF="Uit"] iconcolor=[ON="yellow"]`

The ON in the iconcolor was split by the regex in O and N, the O being interpreted as the item.

lolodomo:
Ok. If you rested it, it is fine for me.
I see that it is a change only for sitemap generator.

holgerfriedrich (approved):
LGTM

PR #4453, opened 2024-11-16
Refactor repeated string literals by defining constants

Jaslene39:
**Addressed Issue:**
The inconsistent use of string literals throughout the codebase increases the risk of errors, makes the code harder to maintain, and complicates refactoring. Repeated string literals contribute to "code smell," as any change requires updating all instances, which is error-prone and time-consuming.

**Reengineered Solution:**
I refactor the codebase by replacing all repeated string literals with centrally defined constants. (means identifying all instances where the same string value appears multiple times in the code and replacing them with a constant that represents that value.)

**Reengineering Strategy:
Rework:**
Refactor and modularize string literals into constants to improve code maintainability and streamline updates.

**Impact of Changes:**

Improved Maintainability: Centralized constants make it easier to update string values in one place without searching through the codebase.
Enhanced Readability: Constants with descriptive names make the code easier to understand and navigate.
Streamlined Refactoring: Changes to string values are straightforward, reducing time and effort during future updates.
Scalability: The codebase becomes more modular and easier to expand without introducing additional "code smells."

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/util/ActionInputsHelper.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/util/ActionInputsHelper.java
@@ -65,6 +65,34 @@ public class ActionInputsHelper {
     private final TimeZoneProvider timeZoneProvider;
     private final UnitProvider unitProvider;
 
+    // Primitive types
+    private static final String BOOLEAN = "boolean";
+    private static final String BYTE = "byte";
+    private static final String SHORT = "short";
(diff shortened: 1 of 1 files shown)


holgerfriedrich (changes requested):
Thank you for your contribution to OH!

I would typically agree to the arguments given, why string constants should be used.
In this specific case, however, I think the readability of the code decreases.
In addition, I see a risk that the constants named after standard types may be accidentally used.

Overall, my personal feeling is that we should not proceed with this specific PR.
Let's see what other maintainers think about this.

Issue #4432, opened 2024-10-29
Ranges (min,max,step) should be provided with a unit


lsiepel:
A common scenario for a thermostat's is that they have a limited operation range. To support these ranges the xml thing type structure has min, max and step xml attributes. After the introduction of UoM, these attributes where not adapted. Binding's now typically have a state pattern of `%unit%` and a `unitHint` set. The min/max/step is hardcoded to a value and it is not known to what unit it relates. 

To workaround this, bindings often choose to support the widest range. With that i mean, the min is set to the value corrosponding to the Celsius and max to the Fahrenheit representation of the operating range. It is obvious that this gives issues. as the actual operating range is more limited.

I would propose to add a xml attribute next to min, max, step to specify the unit for that range: "rangeUnit" It's value would be the actual unit just like `unitHint`. This attribute could be used to determine the min/max/step. This can happen at the same place, just with a little UoM logic added.
If the attribute is missing it would fallback to the current behaviour. So an opt-in, full backwards comptable solution that handles _all_ cases. 

Alternatives:
Use `unitHint` to determine the used unit. Rules out here: https://github.com/openhab/openhab-addons/issues/17638#issuecomment-2442371837 (in essence: this is a UI hint and should not be misused for something else)
Use `state` pattern. This would not cover all bindings as many have it set to `%unit%`

Related to: https://github.com/openhab/openhab-core/pull/3132 
Related to: https://github.com/openhab/openhab-addons/issues/17638 (solving this for just mireds/mireks)

boehan:
Just my thoughts on this:
Another alternative would be to allow units within the definition of `min`, `max`, `step`. Full state description definition could then, e.g., look like this:
`<state min="12 °C" max="30 °C" step="0.5 K" pattern="%.1f %unit%" readOnly="false"></state>`

Potential advantages:
- the state description would keep its clean structure (in consistency with `pattern`) and doesn't need an additional parameter
- it would allow to define different units for `min`, `max` and `step` (may not be needed and probably doesn't make much sense in above example, even with °F instead of °C, but could be useful for wide ranges of `max - min` compared to `step` (e.g., `min`/`max` on `m` vs. `step` in `cm`)

IMO, changes would be minimal:
- support UoM in parsing of `min`, `max` and `step` instead of decimal only
- fall back to current behavior (or state pattern as proposed in [#3132](https://github.com/openhab/openhab-core/pull/3132)) 
  - if no unit is given
  - and/or log an error if given units are not compatible
[...]

PR #4460, opened 2024-11-21
First part

lsiepel:
A common scenario for a thermostat's is that they have a limited operation range. To support these ranges the xml thing type structure has min, max and step xml attributes. After the introduction of UoM, these attributes where not adapted. Binding's now typically have a state pattern of %unit% and a unitHint set. The min/max/step lack a unit so it is not known to what unit it relates.

[x] Adapt XML
[x] Adapt tests
[ ] Validation

Fixes: https://github.com/openhab/openhab-core/issues/4432

### DIFF
--- a/bundles/org.openhab.core.thing/schema/thing/thing-description-1.0.0.xsd
+++ b/bundles/org.openhab.core.thing/schema/thing/thing-description-1.0.0.xsd
@@ -226,6 +226,7 @@
 		<xs:attribute name="min" type="xs:decimal" use="optional"/>
 		<xs:attribute name="max" type="xs:decimal" use="optional"/>
 		<xs:attribute name="step" type="xs:decimal" use="optional"/>
+		<xs:attribute name="rangeUnit" type="xs:string" use="optional"/>
 		<xs:attribute name="pattern" type="xs:string" use="optional"/>
 		<xs:attribute name="readOnly" type="xs:boolean" default="false" use="optional"/>
 	</xs:complexType>
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/xml/internal/StateDescriptionConverter.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/xml/internal/StateDescriptionConverter.java
@@ -51,8 +51,9 @@ public class StateDescriptionConverter extends GenericUnmarshaller<StateDescript
     public StateDescriptionConverter() {
         super(StateDescription.class);
 
-        this.attributeMapValidator = new ConverterAttributeMapValidator(new String[][] { { "min", "false" },
-                { "max", "false" }, { "step", "false" }, { "pattern", "false" }, { "readOnly", "false" } });
+        this.attributeMapValidator = new ConverterAttributeMapValidator(
+                new String[][] { { "min", "false" }, { "max", "false" }, { "step", "false" }, { "rangeUnit", "false" },
+                        { "pattern", "false" }, { "readOnly", "false" } });
     }
 
     private @Nullable BigDecimal toBigDecimal(Map<String, String> attributes, String attribute,
@@ -121,6 +122,11 @@ public class StateDescriptionConverter extends GenericUnmarshaller<StateDescript
         BigDecimal step = toBigDecimal(attributes, "step", null);
         if (step != null) {
             builder.withStep(step);
+        }
+
+        String rangeUnit = attributes.get("pattern");
+        if (rangeUnit != null) {
+            builder.withRangeUnit(rangeUnit);
         }
 
         String pattern = attributes.get("pattern");
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/MetadataStateDescriptionFragmentProvider.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/MetadataStateDescriptionFragmentProvider.java
@@ -96,6 +96,11 @@ public class MetadataStateDescriptionFragmentProvider implements StateDescriptio
                     builder.withStep(getBigDecimal(step));
                 }
 
+                Object rangeUnit = metadata.getConfiguration().get("rangeUnit");
+                if (rangeUnit != null) {
+                    builder.withRangeUnit((String) rangeUnit);
+                }
+
                 Object readOnly = metadata.getConfiguration().get("readOnly");
                 if (readOnly != null) {
                     builder.withReadOnly(getBoolean(readOnly));
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/types/StateDescriptionFragmentImpl.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/types/StateDescriptionFragmentImpl.java
@@ -34,14 +34,16 @@ public class StateDescriptionFragmentImpl implements StateDescriptionFragment {
 
     private static class StateDescriptionImpl extends StateDescription {
         StateDescriptionImpl(@Nullable BigDecimal minimum, @Nullable BigDecimal maximum, @Nullable BigDecimal step,
-                @Nullable String pattern, boolean readOnly, @Nullable List<StateOption> options) {
-            super(minimum, maximum, step, pattern, readOnly, options);
+                @Nullable String rangeUnit, @Nullable String pattern, boolean readOnly,
+                @Nullable List<StateOption> options) {
+            super(minimum, maximum, step, rangeUnit, pattern, readOnly, options);
         }
     }
 
     private @Nullable BigDecimal minimum;
     private @Nullable BigDecimal maximum;
     private @Nullable BigDecimal step;
+    private @Nullable String rangeUnit;
     private @Nullable String pattern;
     private @Nullable Boolean readOnly;
     private @Nullable List<StateOption> options;
@@ -59,6 +61,7 @@ public class StateDescriptionFragmentImpl implements StateDescriptionFragment {
      * @param minimum minimum value of the state
      * @param maximum maximum value of the state
      * @param step step size
+     * @param rangeUnit unit that applies to the min, max and step value
      * @param pattern pattern to render the state
      * @param readOnly if the state can be changed by the system
      * @param options predefined list of options
@@ -66,11 +69,12 @@ public class StateDescriptionFragmentImpl implements StateDescriptionFragment {
      */
     @Deprecated
     public StateDescriptionFragmentImpl(@Nullable BigDecimal minimum, @Nullable BigDecimal maximum,
-            @Nullable BigDecimal step, @Nullable String pattern, @Nullable Boolean readOnly,
+            @Nullable BigDecimal step, @Nullable String rangeUnit, @Nullable String pattern, @Nullable Boolean readOnly,
             @Nullable List<StateOption> options) {
         this.minimum = minimum;
         this.maximum = maximum;
         this.step = step;
+        this.rangeUnit = rangeUnit;
         this.pattern = pattern;
         this.readOnly = readOnly;
         this.options = options == null || options.isEmpty() ? List.of() : Collections.unmodifiableList(options);
@@ -88,6 +92,7 @@ public class StateDescriptionFragmentImpl implements StateDescriptionFragment {
         this.minimum = legacy.getMinimum();
         this.maximum = legacy.getMaximum();
         this.step = legacy.getStep();
+        this.rangeUnit = legacy.getRangeUnit();
         this.pattern = legacy.getPattern();
         this.readOnly = Boolean.valueOf(legacy.isReadOnly());
         if (!legacy.getOptions().isEmpty()) {
@@ -104,6 +109,7 @@ public class StateDescriptionFragmentImpl implements StateDescriptionFragment {
         this.minimum = source.getMinimum();
         this.maximum = source.getMaximum();
         this.step = source.getStep();
+        this.rangeUnit = source.getRangeUnit();
         this.pattern = source.getPattern();
         this.readOnly = source.isReadOnly();
         this.options = source.getOptions();
@@ -137,6 +143,15 @@ public class StateDescriptionFragmentImpl implements StateDescriptionFragment {
     }
 
     @Override
+    public @Nullable String getRangeUnit() {
+        return rangeUnit;
+    }
+
+    public void setRangeUnit(String rangeUnit) {
+        this.rangeUnit = rangeUnit;
+    }
+
+    @Override
     public @Nullable String getPattern() {
         return pattern;
     }
@@ -165,12 +180,13 @@ public class StateDescriptionFragmentImpl implements StateDescriptionFragment {
 
     @Override
     public @Nullable StateDescription toStateDescription() {
-        if (minimum == null && maximum == null && step == null && readOnly == null && pattern == null
-                && options == null) {
+        if (minimum == null && maximum == null && step == null && rangeUnit == null && readOnly == null
+                && pattern == null && options == null) {
             return null;
         }
         final Boolean ro = readOnly;
-        return new StateDescriptionImpl(minimum, maximum, step, pattern, ro != null && ro.booleanValue(), options);
+        return new StateDescriptionImpl(minimum, maximum, step, rangeUnit, pattern, ro != null && ro.booleanValue(),
+                options);
     }
 
     /**
@@ -189,6 +205,9 @@ public class StateDescriptionFragmentImpl implements StateDescriptionFragment {
         }
         if (step == null) {
             step = fragment.getStep();
+        }
+        if (rangeUnit == null) {
+            rangeUnit = fragment.getRangeUnit();
         }
         if (pattern == null) {
             pattern = fragment.getPattern();
@@ -209,6 +228,7 @@ public class StateDescriptionFragmentImpl implements StateDescriptionFragment {
         result = prime * result + (minimum != null ? minimum.hashCode() : 0);
         result = prime * result + (maximum != null ? maximum.hashCode() : 0);
         result = prime * result + (step != null ? step.hashCode() : 0);
+        result = prime * result + (rangeUnit != null ? rangeUnit.hashCode() : 0);
         result = prime * result + (pattern != null ? pattern.hashCode() : 0);
         result = prime * result + (readOnly ? 1231 : 1237);
         result = prime * result + (options != null ? options.hashCode() : 0);
@@ -228,13 +248,15 @@ public class StateDescriptionFragmentImpl implements StateDescriptionFragment {
         }
         StateDescriptionFragmentImpl other = (StateDescriptionFragmentImpl) obj;
         return Objects.equals(minimum, other.minimum) && Objects.equals(maximum, other.maximum)
-                && Objects.equals(step, other.step) && Objects.equals(pattern, other.pattern)
-                && Objects.equals(readOnly, other.readOnly) && Objects.equals(options, other.options);
+                && Objects.equals(step, other.step) && Objects.equals(rangeUnit, other.rangeUnit)
+                && Objects.equals(pattern, other.pattern) && Objects.equals(readOnly, other.readOnly)
+                && Objects.equals(options, other.options);
     }
 
     @Override
     public String toString() {
-        return "StateDescription [minimum=" + minimum + ", maximum=" + maximum + ", step=" + step + ", pattern="
-                + pattern + ", readOnly=" + readOnly + ", channelStateOptions=" + options + "]";
+        return "StateDescription [minimum=" + minimum + ", maximum=" + maximum + ", step=" + step + ", rangeUnit="
+                + rangeUnit + ", pattern=" + pattern + ", readOnly=" + readOnly + ", channelStateOptions=" + options
+                + "]";
     }
 }
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/AbstractStorageBasedTypeProvider.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/AbstractStorageBasedTypeProvider.java
@@ -271,6 +271,7 @@ public abstract class AbstractStorageBasedTypeProvider
         entity.maximum = fragment.getMaximum();
         entity.minimum = fragment.getMinimum();
         entity.step = fragment.getStep();
+        entity.rangeUnit = fragment.getRangeUnit();
         entity.options = fragment.getOptions();
         entity.pattern = fragment.getPattern();
         entity.isReadOnly = fragment.isReadOnly();
@@ -371,6 +372,9 @@ public abstract class AbstractStorageBasedTypeProvider
         if (entity.step != null) {
             builder.withStep(Objects.requireNonNull(entity.step));
         }
+        if (entity.rangeUnit != null) {
+            builder.withRangeUnit(Objects.requireNonNull(entity.rangeUnit));
+        }
         if (entity.options != null) {
             builder.withOptions(Objects.requireNonNull(entity.options));
         }
@@ -444,6 +448,7 @@ public abstract class AbstractStorageBasedTypeProvider
         public @Nullable BigDecimal maximum;
         public @Nullable BigDecimal minimum;
         public @Nullable BigDecimal step;
+        public @Nullable String rangeUnit;
         public @Nullable List<StateOption> options;
         public @Nullable String pattern;
         public boolean isReadOnly = false;
--- a/bundles/org.openhab.core.thing/src/test/java/org/openhab/core/thing/events/ThingEventFactoryTest.java
+++ b/bundles/org.openhab.core.thing/src/test/java/org/openhab/core/thing/events/ThingEventFactoryTest.java
@@ -107,6 +107,7 @@ public class ThingEventFactoryTest {
                     .withMinimum(BigDecimal.ZERO) //
                     .withMaximum(new BigDecimal(1000)) //
                     .withStep(new BigDecimal(100)) //
+                    .withRangeUnit("K") //
                     .withPattern("%.0f K") //
                     .build());
     private static final String CHANNEL_DESCRIPTION_OLD_STATE_DESCRIPTION_PAYLOAD = JSONCONVERTER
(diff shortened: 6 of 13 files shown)


lolodomo:
@openhab/core-maintainers : having this enhancement in 4.3 would be really great. I would like to use it in Basic UI for a perfect working of the Colortemperaturepicker widget.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/xml/internal/StateDescriptionConverter.java
```diff
@@ -123,6 +124,11 @@ private StateOption toChannelStateOption(NodeValue nodeValue) throws ConversionE
             builder.withStep(step);
         }
 
+        String rangeUnit = attributes.get("pattern");
```
mherwege, line 127: Shouldn’t this be “rangeUnit” ?

mherwege:
Would it make sense to populate rangeUnit from pattern when it is not explicitly provided? That will make sure at least the pattern unit from the default is used.

lsiepel:
> Would it make sense to populate rangeUnit from pattern when it is not explicitly provided? That will make sure at least the pattern unit from the default is used.

1. Most UoM channels the unit is not explicitly set in the state pattern (usually %unit%). So there is not much benefit doing this.
2. It introduces breaking changes / uncontrolled behavior. For instance: due to the current behavior of no unit, many bindings have the min/max set to the outer range of all Units. 
Maybe an example would make clear what i mean: 
Take a UoM Temperature channel. Commonly the users can switch between Fahenheit and Celsius. To limit the range, but also support both Units, the author sets the lower limit to Celsius and the upper limit to Fahrenheit.

mherwege:
Fair enough. I only see @lolodomo and others going through addons and changing the default pattern to have a real unit to the benefit of the color temperature picker, and running into issues because the user configured pattern unit is used in the end, and not the pattern from the channel definition. This would avoid an extra round on this. If %unit# is set, it of course will not work and will still provide a null value without changing the channel definition.
While defining unitRange is really what should be done, it may avoid some issues in a few cases.
And I also think it still doesn’t hurt in your second example. It will not be worse than before.

I full agree this should ultimately be a required field when defining min/max/step on a number channel with a dimension. One, we cannot do that overnight as it will break all bindings at once, and two, it is difficult to enforce in code. An attribute is either mandatory, or it is not. But in this case it should only be mandatory for QuantityType channels.

lsiepel:
Ok, what else do tou ( @florian-h05 and @lolodomo) need from core to get this working in the ui’s?  

I think we should leave the validation as is (none) and leave it up to the UI. @openhab/core-maintainers wdyt?

lolodomo:
I see no change in REST bundle so something is very probably missing in this PR.

There is also a notion of merge for state descriptions, the case of the new rangeUnit is not obvious to merge and could also have an impact on min/max. To be checked in your code.

lolodomo:
> There is also a notion of merge for state descriptions, the case of the new rangeUnit is not obvious to merge and could also have an impact on min/max. To be checked in your code.

Imagine the developer sets for the channel a range 153-400 in range unit mirek. Now the user sets on his item a state description with range unit in Kelvin and min to 3000 and no max. What should be the final values for range unit, min and max ?

lolodomo:
With your current code, I believe the result would be range Kelvin (K) min 3000 max 400 and that is wrong.

florian-h05:
> Ok, what else do you need from core to get this working in the ui’s?

For Main UI (and Basic UI I guess as well) we need the min/max/step values converted to the unit of the display state and provided through the REST API as we cannot do these unit conversions in the UI. This way the UI just needs to enforce min/max/step when controlling the Item (which is done or should be done if not already the case in the display state unit) without caring about the actual unit.

mherwege:
> For Main UI (and Basic UI I guess as well) we need the min/max/step values converted to the unit of the display state and provided through the REST API as we cannot do these unit conversions in the UI. This way the UI just needs to enforce min/max/step when controlling the Item (which is done or should be done if not already the case in the display state unit) without caring about the actual unit.

And wouldn't we need anything when defining State Description metadata? I would assume we need to have the `rangeUnit` field there as well.

mherwege:
> There is also a notion of merge for state descriptions, the case of the new rangeUnit is not obvious to merge and could also have an impact on min/max. To be checked in your code.

In my view, the `min`, `max`, `step` and `rangeUnit` fields should be treated as a whole. Don't mix these from 2 providers. I think `pattern` should still be set from any provider. I am undecided on `options`.

lolodomo:
> For Main UI (and Basic UI I guess as well) we need the min/max/step values converted to the unit of the display state and provided through the REST API as we cannot do these unit conversions in the UI.

Is it what's done currently in the item REST API?
For sitemap REST API, I will study.

lolodomo:
@openhab/android-maintainers : I recently discovered that on BasicUI min/max/step from item state description was not used (until the colortemperaturepicker element was added). Are you using them in the Android app?

maniac103:
> Are we using them in the Android app?

Yes. It's used as fallback if no widget level min/max/step is set.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/xml/internal/StateDescriptionConverter.java
lsiepel: Fixed

lsiepel:
> I see no change in REST bundle so something is very probably missing in this PR.
> 
> There is also a notion of merge for state descriptions, the case of the new rangeUnit is not obvious to merge and could also have an impact on min/max. To be checked in your code.
> 
> PS: Main UI should then be enhanced to also propose the range unit for state description.

Could you point me towars the REST change you expect? I fail to find it. I searched `Step` (as i would assume that is also exposed by the API)in the whole repository and couldn't find it anywhere related to REST

bundles/org.openhab.core.thing/schema/thing/thing-description-1.0.0.xsd
```diff
@@ -226,6 +226,7 @@
 		<xs:attribute name="min" type="xs:decimal" use="optional"/>
 		<xs:attribute name="max" type="xs:decimal" use="optional"/>
 		<xs:attribute name="step" type="xs:decimal" use="optional"/>
+		<xs:attribute name="rangeUnit" type="xs:string" use="optional"/>
```
jlaur, line 233: @lsiepel - did you consider just adding units directly, like:
```xml
<state min="-5 °C" max="5 °C" step="0.1 °C" pattern="%.1f %unit%" readOnly="false"/>
```

i.e., just to understand the reasoning, I assume `rangeUnit` is introduced to avoid a new version of the schema, because min, max and step are currently of type `xs:decimal`? Or is it to enforce the same unit for all three?

boehan: I was suggesting the same in https://github.com/openhab/openhab-core/issues/4432#issuecomment-2446698566. This would also allow to use different units, e.g. 
```xml
<state min="1 m" max="5 m" step="1 dm" pattern="%.1f %unit%" readOnly="false"/>
```

lsiepel: It has been a while, so i might need to dig into my memmory.

The suggestion is more flexible. It comes of the cost of more complexity; multiple units at play, more places to validate and existing functions need to be checked. Also the mentioend xml versioning is at play. But still possible!

If this is the way forward, i'll adapt the PR.

andrewfg: I think it should be possible to do it. There are two parts of the problem -- namely 1) the XML and XSD and 2) the Java DTO and instance classes. Probably in the Java classes you have anyway two fields for a decimal value and its unit. (??) So the trick (only) is to write an xml parser that can read the three variants (`min=123` / `min=123 W` / `min=123` && `unit="W"`) and populate the DTO accordingly. Once you have the data in a Java DTO it is easy after that.

mherwege:
What consequence would it have for the UI if min, max and step are expressed in different units? Would the UI still use the item configured unit, so the min, max and step always get recalculated into the item unit? Or will it have different units there as well? How should that be represented (e.g. a stepper control with min, max and step).

What about metric vs imperial? Do you want the ability to define different min/max/step for metric and imperial? We have the ability to have different unit hints depending on the user having a metric or imperial setup. The step is mainly useful in UI controls.  If you e.g. a min, max and step in °C (e.g. 0.1 °C), it will lead to funny behaviour when the user has °F as the unit. You would expect a step size that is more compatible to °F then.

I tend to think that keeping it simple is the way to go here. I wouldn't allow different units for min, max and step to avoid funny behaviour like above, but would consider an ability to have different values depending on the measurement system used.

I very much like the enhancement and it is urgently needed. We just need to think through all consequences.

jlaur:
> What consequence would it have for the UI if min, max and step are expressed in different units?

I think it shouldn't have any.

> Would the UI still use the item configured unit, so the min, max and step always get recalculated into the item unit? Or will it have different units there as well? How should that be represented (e.g. a stepper control with min, max and step).

My understanding is that it should be recalculated into the item unit, otherwise it's hard to make sense of. Take for example this example: openhab/openhab-addons#20139:
```xml
<state min="-5" max="5" step="0.1" pattern="%.1f %unit%" readOnly="false"/>
```

The thermostat device accepts relative temperatures values in the range [-5 °C : 5 °C] with one decimal. For Fahrenheit the range would be [-9 °F : 9 °F], because of the range the devices actually supports, and there will be some rounding if steps are 0.1 °F. No matter if we write:

```xml
<state min="-5 °C" max="5 °C" step="0.1 °C" pattern="%.1f %unit%" readOnly="false"/>
```

or:

```xml
<state min="-5" max="5" step="0.1" rangeUnit="°C" pattern="%.1f %unit%" readOnly="false"/>
```

we might have some awkwardness regarding the `step` value, because in °F it would become 0.18 °F. That will probably look strange in a UI? The `min` and `max` in °C are real boundaries, so they have to be enforced, but in a UI, step could be represented differently than 0.1, as long as the final value in °C is rounded to one decimal. For this example, it could have been 0.2 °F steps. I don't know how to "solve" that, maybe I'm overthinking it? In other API's, there might be something like values going from 0-255 representing a position relative to `min`/`max`, and in such cases the UI would probably deviate more from the raw API spec.

I don't know if other use cases exist where we would like different ranges depending on the unit?

> What about metric vs imperial? Do you want the ability to define different min/max/step for metric and imperial?

OK, I guess that's the same question. Right now I cannot think of such a use case, but perhaps someone else can.

> I tend to think that keeping it simple is the way to go here. I wouldn't allow different units for min, max and step to avoid funny behaviour like above, but would consider an ability to have different values depending on the measurement system used.

I think we need some more use cases to go through some real examples and see how they would play out. I imagined the unit to be used only for recalculations, because that's what we are missing now. I understand this is tricky from a UI perspective, because we are going from something static (but wrong) to something dynamic. With:

```xml
<state min="-5" max="5" step="0.1" pattern="%.1f %unit%" readOnly="false"/>
```

it's easy to imagine a slider going from -5 to 5 with 100 different positions, given the step size. However, when interpreted as:

```xml
<state min="-9 °F" max="9 °C" step="0.18 °F" pattern="%.1f %unit%" readOnly="false"/>
```

it just seems a bit more odd. Which it is. However, if we actually respect the 0.18 step size in the UI, we would still have 100 different positions, at least.

lsiepel:
To be honest, i don't see the advantage of providing different units for the min/max/step. These preset values should match the device's boundaries. Never seen any that has differnt unit for min or max, precision is the only real advantage here. And as you are free to use any unit, you can stick to the precision the device supports. So unless there is a specific use-case one can come-up with i'd like to take this as a base rule for now.

The `rangeUnit` is converted to the item's unit. This enables a consistent min/max/step and value range, but could lead to strange values, pointed out by @jlaur.
On the other hand this is already a problem and we are greatly reducing it here. Currently a 0.5 step has a different precission/meaning depending on the unit used. With the suggested improvement it is 'just' a visualisation issue.

boehan:
> What consequence would it have for the UI

The representation of the state is defined by the pattern, the unit of step and range shouldn't change anything about it. stepsize and range are only user-facing in stateDescription definition (thinking about it, there is probably additional stepsize and limit settings in widget/sitemap definitions, which may be different from stateDescription - what about units for those?)
The problem of potentially different units of state and range/step is independent of the discussed implementations - otherwise we wouldn't have to define an additional specific unit anyway.

> To be honest, i don't see the advantage of providing different units for the min/max/step.

I have to admit that the practical advantage is probably rather limited. Ultimately, having any clear definition of a unit for range and step is a big improvement. For me it is probably just a UX preference to have the unit directly presented with the value rather than in a separate field (and in the end this could also be handled in the UI, if really wanted).

mherwege:
> The representation of the state is defined by the pattern, the unit of step and range shouldn't change anything about it.

It is used in the UI with controls (number, stepper, slider). So, yes, it does have an impact that should be considered. Right now, that relation is one to one so the steps are directly used. Doing a unit conversion on the step will impact that and may make these behave funny, or at least more so then today. And that’s especially true if you switch from metric to imperial as the scales become different and natural steps for the UI are different. That’s the reason why I suggest considering double definitions. The use of min, max and step is to give meaningful constraints to the UI.

lsiepel:
>  That’s the reason why I suggest considering double definitions.

To fully understand what you suggest, could you explain this more in detail?

mherwege:
The unitHint parameter allows something like unitHint="l,gal", whereby the second unit is the hint when your system is using imperial.

I expect the UI min, max and step to align with logical units so the controls use that.

E.g. if I have rangeUnit="°C" min=1 and max=30 step=1, the UI will step on 1, 2, 3, 4... 30
Convert this to °F with the same definition. Now the UI will step on 33.8, 35.6, 37.4... 86
The display may even be more erratic as the number of decimals in the state description plays a role.
What you probably want is, when the unit is °F, to go from 34 to 86 with steps of 1 or 2.

While the issue can arise in many places if you use a different unit than the one you defined the range in, it is especially troublesome when switching between measurement systems (metric or imperial).

So in my view, we should have something like:
- allow rangeUnit to have 2 values, comma separated, where the second one, if provided, is for an imperial measurement system
- have minAlt, maxAlt and stepAlt fields (or something like that) that will be used if imperial and defined
- the API should always serve min, max and step, but consider the measurement system you are in when an alternative unit is given in rangeUnit and there are alt values

From a documentation perspective, we should recommend the rangeUnit to be the same as the unitHint to limit the potential of UI strangeness

andrewfg:
If you use comma separated pairs for the unit, then use comma separated pairs for min, max, and step. (Dont use minAlt etc.)

mherwege:
> If you use comma separated pairs for the unit, then use comma separated pairs for min, max, and step. (Dont use minAlt etc.)

If possible, yes. But these are now defined as numeric. It may break things.

andrewfg:
> these are now defined as numeric. It may break things

It probably needs a custom Json type adapter that can parse both into the DTO..

mherwege:
> > these are now defined as numeric. It may break things
> 
> It probably needs a custom Json type adapter that can parse both into the DTO..

I am more worried about the xml schema.

lsiepel:
> > > these are now defined as numeric. It may break things
> > 
> > 
> > It probably needs a custom Json type adapter that can parse both into the DTO..
> 
> I am more worried about the xml schema.

Meaning you want to stick with the current suggested add of the optional field `rangeUnit` ?

lsiepel:
Iterating on the suggestion from @mherwege, I would like to point out a few corner cases.

If we introduce this rather complex mechanism, it seems that the problem is not really solved but merely moved elsewhere. Since unit conversion and quantization inevitably introduce some loss of precision, we need to clarify how we handle the resulting side effects:

- What happens if the command derived from the UI (after conversion and rounding) falls outside the technical limits of the device?
- Do we accept that, depending on the chosen display unit and step size, parts of the device’s full operating range may effectively become unreachable?

This leads to a more fundamental question:
What is the exact definition and intended meaning of the XML elements min, max, and step?

1. Do they represent the technical boundaries and resolution of the device?
1. Or are they strictly UI metadata intended to guide presentation and interaction?

Without a clear and explicit definition, it is difficult to reason about correct behavior, especially when conversions between units are involved.

mherwege:
> If we introduce this rather complex mechanism, it seems that the problem is not really solved but merely moved elsewhere. 

The problem is there already I am afraid. Right now, the UI will use the numeric min, max and step applied to the configured unit. So that means if you express min, max and step in meter and have a configured unit of feet, the range and step are totally different anyway. And that is what this PR tries to solve. Therefore I don't understand how the problem is moved somewhere else just by my proposal. If anything, it makes the problem less severe. I can't claim the problem goes away totally though, because even with conversions inside a measurement system (metric or imperial), you could end up with unnatural min, max and step values. But it would be much less common, and it least in the metric system the conversions are all in powers of 10.

> * What happens if the command derived from the UI (after conversion and rounding) falls outside the technical limits of the device?

I would think the binding remains  responsible to make sure what is send to the device is acceptable for the device. I don't see any difference from the current situation. We can already send a value that is not directly acceptable. Depending on the binding, it may get refused all together, or the value may be rounded to the next acceptable value. I believe it is fairly common in binding to have to round e.g. temperatures to the next 0.1 °C multiple for it to be accepted. And I would expect that is mostly already the case today.

> * Do we accept that, depending on the chosen display unit and step size, parts of the device’s full operating range may effectively become unreachable?

That's a good point, but isn't it up to the binding developers to define the range in such a way that it respects the constraints? So if you set the range slightly larger, the rounding I pointed out above should be used by the binding to round to min or max if required.

> 
> This leads to a more fundamental question: What is the exact definition and intended meaning of the XML elements min, max, and step?
> 
> 1. Do they represent the technical boundaries and resolution of the device?
> 2. Or are they strictly UI metadata intended to guide presentation and interaction?
> 
> Without a clear and explicit definition, it is difficult to reason about correct behavior, especially when conversions between units are involved.

My reading is that the original intention was to be both 1 and 2, but 1 as a means to inform the UI to guide presentation and interaction there. That's easy enough if there are no units involved. Rounding may actually always have been somewhat of an issue. But it breaks down in the current context. And if I then look at how min, max and step are used, it is only for representation in the UI, not in any way to drive logic in the binding when sending back values to the device.

I have also never seen a UI for e.g. a thermostat, where when you switch between units (and that is very typical switching from °C to °F) it suddenly starts making more funny jumps. It has steps that are logical for the unit in the UI. What happens internally is hidden.

In my proposal with a dual system, the use of the second set of paramters is optional. If only one set of values is specified, the automatic unit conversion will still respect the boundaries and steps as defined in the first set of parameters and may look less natural in the UI. Apart from rounding, that will remain very close to the base, and if that is a technical constraint, we should be very close (some rounding may always be required). But if the system has somewhat more flexibility, defining boundaries and steps by measurement system makes sense to me. And I believe allowing that and setting the parameters for it should be done in the binding, not by guessing in the UI. It is up to the binding to define what is possible, not up to the UI to guess.

My proposal may look complex, but I do believe it leaves the control with the binding developer, is backward compatible (except for a potential impact on the xml schema) and does not require immediate changes to the UI. The code to work with 2 measurements systems and pick the right one when the REST API calls for it is slightly more complex, but I am not convinced it has that big of an impact.
And as we are already allowing something similar with the unitHint parameter, it to me looks like something that can be understood in the same way.

lsiepel:
I think part of the discussion comes from mixing two different concerns: the device’s operating range and the UI interaction range.

From a core perspective, we should primarily describe the canonical operating range of the device. That means:

- min
- max
- step
- rangeUnit

together define the true constraint model of the channel/item in a single unit. For example, a thermostat may have a native operating range of `[-5 °C, 5 °C]` with `0.1 °C` increments. That information is useful for many purposes in the system, not only for UI widgets.

Once that canonical range is known, it can be converted to any other unit where needed. If a user prefers Fahrenheit, the UI (or REST layer) can convert the bounds to °F before presenting them.

The key point for me is that the core should define the constraint model, while the UI is responsible for presentation and interaction.

This also helps clarify the discussion about step. A converted  `step` (e.g. `0.1 °C` → `0.18 °F`) may not always be the most natural increment for a UI control. But that does not change the fact that the device’s native increment is `0.1 °C`. The UI can still choose a more convenient interaction step (e.g. `0.5 °F`) as long as it respects the canonical range and snaps values appropriately.

Because of that separation, I would prefer keeping `rangeUnit` as a single canonical unit for the range definition. That gives us a clear and consistent model in core. Introducing multiple units into the range definition would mix the device constraint model with UI interaction concerns.

So my suggestion would be:

- Use `rangeUnit` to define the native unit of `min/max/step`
- Treat those values as the canonical operating range. 
- Let UIs convert and present them in the user’s preferred unit.

This keeps the core model simple and still allows UIs to implement interaction models that feel natural in different unit systems.

bundles/org.openhab.core.thing/schema/thing/thing-description-1.0.0.xsd
lsiepel: There are two key reasons described in the last post.
having the optional `rangeUnit` makes it backwards compatible. It also reduces complexity by using one unit for min/max/step. Mixing them would make it unnecesairy complex, devices never have different units for these parts of the operating range.

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/MetadataStateDescriptionFragmentProvider.java
```diff
@@ -96,6 +96,11 @@ public MetadataStateDescriptionFragmentProvider(final @Reference MetadataRegistr
                     builder.withStep(getBigDecimal(step));
                 }
 
+                Object rangeUnit = metadata.getConfiguration().get("rangeUnit");
+                if (rangeUnit != null) {
+                    builder.withRangeUnit((String) rangeUnit);
```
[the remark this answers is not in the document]
lsiepel: This can be ignored. It has the same pattern as the state field.

bundles/org.openhab.core/src/main/java/org/openhab/core/internal/types/StateDescriptionFragmentImpl.java
```diff
@@ -136,6 +142,15 @@ public void setStep(BigDecimal step) {
         this.step = step;
     }
 
+    @Override
+    public @Nullable String getRangeUnit() {
+        return rangeUnit;
+    }
+
+    public void setRangeUnit(String rangeUnit) {
```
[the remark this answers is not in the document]
lsiepel: It may seem to make sense, i did not impolement this to make it consistent with the other fields.

lsiepel:
@lolodomo what do you need for the widget you planned before: https://github.com/openhab/openhab-core/pull/4460#issuecomment-2509659186

What are your thoughts @andrewfg on the xml schema part? Can we proceed as is, or do think it would be better to move it to the 1.1.0 schema ?

mherwege:
> I think part of the discussion comes from mixing two different concerns: the device’s operating range and the UI interaction range.
> 
> From a core perspective, we should primarily describe the canonical operating range of the device.

I totally agree on the first point. But I fundamentally disagree with you conclusion. So far, nothing has ever made the logic in the bindings or handling of channel changes or updates dependent on the min, max and step in the state description. It has always been the bindings responsibility to make values it sends to the device be acceptable for the device. Min, max and step have always been there for the UI, just like other elements in the state description.

I think we miss an opportunity to be more thorough on this and I am afraid we are left with too many issues.

mherwege:
> The UI can still choose a more convenient interaction step (e.g. `0.5 °F`) as long as it respects the canonical range and snaps values appropriately.

As the user can override these in metadata, this snapping to device acceptable values cannot be the responsibility of the UI.

lsiepel:
I think we are aligned on responsibilities, so let me clarify this PR.

I am not trying to enforce constraints in core. It will remain the binding’s responsibility to ensure values sent to the device are valid, and it is up to the UI to decide how to present and interact with a widget.

The goal here is simply that core provides the canonical range metadata (`min/max/step` with `rangeUnit`) so it can be interpreted consistently. If there is demand, this PR could be extended to provide convenience or utility functions or data to enable the binding and the UI to validate or whatever they need.

From there:

- bindings ensure device validity
- UIs decide how to present it (including unit conversion if needed)

Also a small note: this PR started quite a while ago. When re-reading my earlier comments I noticed my thinking has evolved slightly, which is also why the [ ] validation box in the original description was left open at the time.

> ...this snapping to device acceptable values cannot be the responsibility of the UI..

If you mean the UI should be responsible for ensuring that the value sent to the device is actually acceptable, I totally agree that this should not be the UI’s responsibility.

However, I do think the UI should be responsible for snapping to reasonable interaction values. The UI knows the presentation context (available size, touch vs pointer interaction, and the form of the control such as slider, list, buttons or stepper), so it is in the best position to choose sensible interaction steps.

The binding should still round, truncate or otherwise adjust the value as needed before sending it to the device. Bindings with UoM typically already implement this behaviour.

mherwege:
> * UIs decide how to present it (including unit conversion if needed)

It is not the UI that will do unit conversion. The UI does not have unit conversion logic. The REST API will return min, max and step already converted to the item unit or widget unit.

lsiepel:
> > * UIs decide how to present it (including unit conversion if needed)
> 
> It is not the UI that will do unit conversion. The UI does not have unit conversion logic. The REST API will return min, max and step already converted to the item unit or widget unit.

Yes, maybe conversion is not the right word, it will need to use the REST api to obtain information (that might be converted or not).

How do you suggest to proceed ?

mherwege:
I did propose a, what I believe to be a more complete, solution. This gives some power to the binding developer to create better values in different measurement systems, reducing the chance a user would want to override the provided defaults. In your implementation, most users will want to create metadata again for °F when the step is expressed in °C if they don’t want the erratic UI behavior. Then we gained little.
However, I still very much prefer what is in your solution already above no solution at all. So if there is no agreement I would like to see your solution merged as quickly as possible.

lsiepel:
Would be nice if @openhab/core-maintainers can share their thoughts. Not meant as an invitation to choose either solution as is, but to improve this PR and the way forward.

lolodomo:
> @lolodomo what do you need for the widget you planned before: [#4460 (comment)](https://github.com/openhab/openhab-core/pull/4460#issuecomment-2509659186)

IICR my need was to know what is the unit for provided min and max values, either Kelvin or mirek.

Issue #4439, opened 2024-11-04
Provide Riemann Sum implementation for items


dandjo:
## Requirement

I want to calculate the energy consumption of a sensor that only provides me with power data at irregular intervals. This could be, for example, the output of a heat pump, a heat output meter or other time-discrete values. For this, we need to implement an integral function.

## Suggestion

Provide a Riemann sum implementation with different alignments (left, right, mid, trapezoidal). Center/mid alignment is the best compromise between accuracy and implementation complexity.

[Riemann Sums](https://www.statisticshowto.com/calculus-problem-solving/riemann-sums/)

## Implementation

I provided a Riemann sum mid implementation in this discussion:

[Arithmetic mean vs average linear/step interpolation](https://community.openhab.org/t/arithmetic-mean-vs-average-linear-step-interpolation/159892/15)

## Your Environment

```
runtimeInfo:
  version: 4.3.0
  buildString: "Build #4362"
locale: en-AT
systemInfo:
  configFolder: /etc/openhab
  userdataFolder: /var/lib/openhab
  logFolder: /var/log/openhab
  javaVersion: 17.0.12
  javaVendor: Ubuntu
  osName: Linux
  osVersion: 6.8.0-1013-raspi
  osArchitecture: aarch64
  availableProcessors: 4
  freeMemory: 83704192
  totalMemory: 605028352
  uptime: 29022
  startLevel: 100
addons:
  - automation-jsscripting
  - binding-enigma2
  - binding-http
  - binding-mielecloud
  - binding-modbus
  - binding-mqtt
  - binding-netatmo
  - binding-tado
  - misc-openhabcloud
  - persistence-influxdb
  - persistence-mapdb
  - transformation-jsonpath
  - transformation-regex
  - ui-basic
clientInfo:
  device:
    ios: false
    android: false
    androidChrome: false
    desktop: true
    iphone: false
    ipod: false
    ipad: false
    edge: false
    ie: false
    firefox: false
    macos: false
    windows: false
    cordova: false
    phonegap: false
    electron: false
    nwjs: false
    webView: false
    webview: false
    standalone: false
    pixelRatio: 1
    prefersColorScheme: dark
  isSecureContext: false
  locationbarVisible: true
  menubarVisible: true
  navigator:
    cookieEnabled: true
    deviceMemory: N/A
    hardwareConcurrency: 12
    language: en-AT
    languages:
      - en-AT
      - en-GB
      - en-US
      - en
      - de
    onLine: true
[...]

PR #4461, opened 2024-11-22
Persistence Extensions: Riemann Sums

mherwege:
Closes https://github.com/openhab/openhab-core/issues/4439

At the moment, it is not possible to easily calculate the integral value of the curve represented by persisted values. This is useful for calculating e.g. total energy consumed (in kWh) from instantaneous power meter readings (in W).

The current `sum` persistence actions calculate a sum without considering the time dimension, and therefore are only useful when we have a guaranteed constant interval between persisted values for this.

The `average` actions do use a time weighting in their calculation. Multiplying the average with the total duration considered would give an approximation. However, the `average` calculations assume a constant value over the bucket and take the value at the start of the bucket. Using this average to calculate an integration would also mean an extra calculation (already divided in the action code and multiplied again in the rule).
An alternative approach to get an integral value is by setting up an integrated item and a rule that adds value * duration since previous change/upate, triggered on base item change. This works and is flexible in its calculation (value can be approximated based on previous and new item state). It suffers from things like unexpected shutdowns, where the base item and its aggretation may run out of sync.

This PR proposes to create a group of new actions, `riemannSum` , that will calculate the RiemannSum as an approximation for the integral value. Analoguous to the other actions, there are variants taking (or not) `startDate`, `endDate` and `serviceId` as input.
There is one extra key parameter, the Riemann type, representing the type of approximation used. Valid values are:

- `RiemannType.left`: takes the persisted value at the start of the bucket to represent the value for the whole bucket. This is most useful when there is a `persistOnChange` strategy and the values represent a step function. An example would be dynamic electricity rates, as they will  effectively be constant inside the bucket.
- `RiemannType.right`: takes the persisted value at the end of the bucket.
- `RiemannType.trapezoidal`: takes the average of the persisted value at the start end the end of the bucket, effectively making a linear interpolation to fit the curve. This type is most useful when the real values change continuously. It can be used for any persistence strategy and any interval.
- `RiemannType.midpoint`: uses 3 persisted values and uses the middle of the values as an approximation for the value halfway in the interval between the middle of point 1 and 2 and the middle of point 2 and 3. This is the best approximation when the real values change continuously, the persistence intervals are short and the bucket sizes between persisted values are relatively constant.

The existing `average`, `variance` and `deviation` actions have been enhanced to use the results of the Riemann sum calculations. They now have variants with an extra parameter where this type can be selected.
For backward compatibility and consistency between defaults of the different actions, the action versions for `riemannSum`, `average`, `variance` and `deviation` without a Riemann type parameter all use `RiemannType.left` as a default.

Tests have been enhanced end confirmed the existing actions with the existing parameters still behave as before.

If this PR is accepted the scripting libraries should be enhanced to cover these new actions and the new parameters in the existing actions.

For testing, a compiled jar can be downloaded from [here](https://www.dropbox.com/scl/fi/xmks0sp5g2yobdfxplw0i/org.openhab.core.persistence-4.3.0-SNAPSHOT.jar?rlkey=lmg3a0s3vy6kqqr2w8px0jtua&st=056uj3mx&dl=0).

### DIFF
--- a/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
+++ b/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
@@ -34,6 +34,7 @@ import org.openhab.core.items.Item;
 import org.openhab.core.library.items.NumberItem;
 import org.openhab.core.library.types.DecimalType;
 import org.openhab.core.library.types.QuantityType;
+import org.openhab.core.library.unit.Units;
 import org.openhab.core.persistence.FilterCriteria;
 import org.openhab.core.persistence.FilterCriteria.Ordering;
 import org.openhab.core.persistence.HistoricItem;
@@ -69,6 +70,7 @@ import org.slf4j.LoggerFactory;
  * @author Mark Herwege - handle persisted GroupItem with QuantityType
  * @author Mark Herwege - add median methods
  * @author Mark Herwege - use item lastChange and lastUpdate methods if not in peristence
+ * @author Mark Herwege - add Riemann sum methods
  */
 @Component(immediate = true)
 @NonNullByDefault
@@ -77,6 +79,13 @@ public class PersistenceExtensions {
     private static @Nullable PersistenceServiceRegistry registry;
     private static @Nullable PersistenceServiceConfigurationRegistry configRegistry;
     private static @Nullable TimeZoneProvider timeZoneProvider;
+
+    public static enum RiemannType {
+        LEFT,
+        MIDPOINT,
+        RIGHT,
+        TRAPEZOIDAL
+    }
 
     @Activate
     public PersistenceExtensions(@Reference PersistenceServiceRegistry registry,
@@ -1153,6 +1162,7 @@ public class PersistenceExtensions {
 
     /**
      * Gets the variance of the state of the given {@link Item} since a certain point in time.
+     * A left approximation type is used for the Riemann sum.
      * The default {@link PersistenceService} is used.
      *
      * @param item the {@link Item} to get the variance for
@@ -1162,11 +1172,27 @@ public class PersistenceExtensions {
      *         if there is no persisted state for the given <code>item</code> at the given <code>timestamp</code>
      */
     public static @Nullable State varianceSince(Item item, ZonedDateTime timestamp) {
-        return internalVarianceBetween(item, timestamp, null, null);
+        return internalVarianceBetween(item, timestamp, null, null, null);
+    }
+
+    /**
+     * Gets the variance of the state of the given {@link Item} since a certain point in time.
+     * The default {@link PersistenceService} is used.
+     *
+     * @param item the {@link Item} to get the variance for
+     * @param timestamp the point in time from which to compute the variance
+     * @param type LEFT, RIGHT, MIDPOINT or TRAPEZOIDAL representing approximation types for Riemann sums
+     * @return the variance between then and now, or <code>null</code> if <code>timestamp</code> is in the future, if
+     *         there is no default persistence service available, or it is not a {@link QueryablePersistenceService}, or
+     *         if there is no persisted state for the given <code>item</code> at the given <code>timestamp</code>
+     */
+    public static @Nullable State varianceSince(Item item, ZonedDateTime timestamp, @Nullable RiemannType type) {
+        return internalVarianceBetween(item, timestamp, null, type, null);
     }
 
     /**
      * Gets the variance of the state of the given {@link Item} until a certain point in time.
+     * A left approximation type is used for the Riemann sum.
      * The default {@link PersistenceService} is used.
      *
      * @param item the {@link Item} to get the variance for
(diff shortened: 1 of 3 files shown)


mherwege:
Tests run fine locally. I am trying to figure out why they do not work in the PR.

mherwege:
I put in some extra test logging to try to analyze where the difference is between my succeeding local tests and the failing online tests.

mherwege:
All test are passing now and extra logging has been removed. This is ready for review.

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
```diff
@@ -74,6 +76,13 @@ public class PersistenceExtensions {
     private static @Nullable PersistenceServiceRegistry registry;
     private static @Nullable TimeZoneProvider timeZoneProvider;
 
+    public static enum RiemannType {
```
andrewfg, line 79: Shouldn't these enums be in ALL_CAPS ??

andrewfg (changes requested):

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
mherwege: Yes, they should be. I will change it.

mherwege: > Shouldn't these enums be in ALL_CAPS ??

Changed

jimtng:
This class is getting very cumbersome with all the parameter permutations!

jimtng:
Does the old `sumXxx` return the same result as `riemannSumXxx` with `RiemannType.LEFT`?

mherwege:
> This class is getting very cumbersome with all the parameter permutations!

Totally agree. Do you have any suggestions? As it is all about static methods in one class, it is not easy to split without introducing a lot of breaking changes. The private methods could move to one or more utility classes, but I am not sure this will make maintaining it any easier.

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
```diff
@@ -1217,20 +1334,21 @@ private static void internalPersist(Item item, TimeSeries timeSeries, @Nullable
...
-    public static @Nullable State varianceBetween(Item item, ZonedDateTime begin, ZonedDateTime end,
+    public static @Nullable State varianceBetween(Item item, ZonedDateTime begin, ZonedDateTime end, RiemannType type,
             @Nullable String serviceId) {
-        return internalVarianceBetween(item, begin, end, serviceId);
+        return internalVarianceBetween(item, begin, end, type, serviceId);
     }
 
     private static @Nullable State internalVarianceBetween(Item item, @Nullable ZonedDateTime begin,
-            @Nullable ZonedDateTime end, @Nullable String serviceId) {
+            @Nullable ZonedDateTime end, RiemannType type, @Nullable String serviceId) {
```
jimtng, line 1351: This comment applies to all the other similar methods.

Whilst implementing this into the jruby helper library, I came across one particular issue.

Could you please make the `type` parameter `@Nullable`, and provide the default (i.e. `RiemannType.LEFT`) inside the `internalXxxxXxxx` methods, and pass `null` by the callers? This would:
- DRY and collect all the defaults into one place (in the internalXxxx method) rather than in the callers
- Makes it easy for the helper libraries to also call it with `null` to indicate "default", and therefore letting core decides what the default is.

As it is now, I'd either have to also hard-code this default in the helper library, thus maintaining a duplicate default, or conditionally call one of the overloaded methods that doesn't take the `type` argument in order to use the "default" type.

mherwege: Yes, I will do that. Makes sense.

mherwege: Done

mherwege:
@florian-h05 I was looking into making these new methods available in JS. I don’t know how to properly handle the RiemannType enum arguments. Does a Java enum work properly in GraalVM JS scripting? Or should I be using string arguments in this PR?That may make it harder as well as the serviceId argument is also a string and will make it impossible to distinguish in varargs.

florian-h05:
I would expose the RiemannType enum as a static member of the ItemPersistence class, I think that should work (even though enums do not natively exist in JS).

holgerfriedrich (commented):
@mherwege sorry for the long delay....
I have just seen that I already wrote in prep to the OH4.3 release "The most interesting pending feature is probably #4461, as it is useful for computations regarding energy consumption." Though, I did not manage to review this until now. 🙈 

Overall, LGTM!

I left a few comments.
There is a minor conflict, could you please rebase?

Finally, I think we need some user doc for this (this could be based on the text you already added to the PR description, and might also include pictures from #4439).

bundles/org.openhab.core.persistence/src/test/java/org/openhab/core/persistence/extensions/TestPersistenceService.java
```diff
@@ -234,17 +240,168 @@ static OnOffType switchValue(int hour) {
...
         } else if (year <= FUTURE_END) {
-            return new DecimalType(year);
+            return new DecimalType(year + (kelvinOffset ? KELVIN_OFFSET : 0));
         } else {
-            return new DecimalType(FUTURE_END);
+            return new DecimalType(FUTURE_END + (kelvinOffset ? KELVIN_OFFSET : 0));
+        }
+    }
+
+    static double riemannSum(@Nullable Integer beginYear, @Nullable Integer endYear, RiemannType type) {
```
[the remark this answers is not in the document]
mherwege: done

bundles/org.openhab.core.persistence/src/test/java/org/openhab/core/persistence/extensions/PersistenceExtensionsTest.java
```diff
@@ -1485,6 +1487,435 @@ public void testDeviationBetweenGroupQuantityType() {
...
+        int futureHours = 0;
+
+        createTestCachedValuesPersistenceService(now, historicHours, futureHours);
+
+        State sum = PersistenceExtensions.riemannSumSince(numberItem, now.minusHours(historicHours), type,
+                TestCachedValuesPersistenceService.ID);
+        assertNotNull(sum);
+        DecimalType dt = sum.as(DecimalType.class);
+        assertNotNull(dt);
+        assertThat(dt.doubleValue(), is(closeTo(100.0 * 3600 + 50.0 * 3600, 0.01)));
```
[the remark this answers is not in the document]
mherwege: I put in some extra inline comments.

mherwege:
> Finally, I think we need some user doc for this (this could be based on the text you already added to the PR description, and might also include pictures from #4439).

I totally agree. I intended to work on documentation once there was progress on this and the related PRs for javascript (https://github.com/openhab/openhab-js/pull/401) and Blockly (https://github.com/openhab/openhab-webui/pull/2893).

mherwege:
@holgerfriedrich The CI build fails on a test that was not touched in the last commit. It builds fine locally with no test failures. I don't see what the issue can be.

dandjo:
Thanks everyone for this!

PR #4465, opened 2024-11-30
ConfigDescriptionParameter & Input: Add null annotations & improve JavaDoc

florian-h05:


bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/ConfigDescriptionParameter.java
```diff
@@ -362,35 +371,35 @@ public String getContext() {
...
      */
     public boolean isRequired() {
-        return this.required;
+        return required;
     }
 
     /**
      * Returns the default value of the configuration parameter.
      *
      * @return the default value of the configuration parameter (could be null)
```
wborn, line 380: The annotations now provide ths info so please remove this redundant info from the impacted JavaDocs:

```suggestion
     * @return the default value of the configuration parameter
```

florian-h05: @wborn Done :+1:

PR #4472, opened 2024-12-04
A session of compile warning hunt

clinique:
Decreased the number of warnings from 277 to 196 when compiling org.openhab.core


bundles/org.openhab.core/src/main/java/org/openhab/core/events/system/SystemEventFactory.java
```diff
@@ -19,7 +19,6 @@
 import org.openhab.core.events.AbstractEventFactory;
 import org.openhab.core.events.Event;
 import org.openhab.core.events.EventFactory;
-import org.openhab.core.types.Type;
```
jlaur, line 22: When removing it here, you probably need to fully qualify it in the Javadoc (e.g. line 40).

clinique: Done

holgerfriedrich (approved):
LGTM, thanks a lot!

PR #4481, opened 2024-12-08
IP addon finder mac address format tweak

andrewfg:
Tweaks to #4477


andrewfg:
@holgerfriedrich ping!!

holgerfriedrich:
@andrewfg Sorry, I did not manage to review this before Kai pressed the button to build M5. Shouldn't be a problem to delay this, as the merged implementation works for the common delimiters.

andrewfg:
Yes. It really only affects developers who might write faulty code. And probably such developers will be using snapshots anyway.

andrewfg:
@holgerfriedrich this is ready to merge.

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -614,18 +613,44 @@ private byte[] macBytesFrom(InetSocketAddress inetSocketAddress) throws SocketEx
 
     /**
      * Use the given format specifier to format an array of mac address bytes
-     * 
+     *
      * @param format a standard format specifier; optionally ends with a delimiter e.g. "%02x:" or "%02X"
      * @param bytes the mac address as an array of bytes
-     * 
+     *
```
holgerfriedrich, line 619: This empty line seems to cause a new SAT warning, 
```
checkstyle | 3 | 619 | style | JavadocMethodStyleCheck | There should be no empty lines between tags in a Javadoc comment of a method or constructor.
```

```suggestion
```

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
```diff
@@ -614,18 +613,44 @@ private byte[] macBytesFrom(InetSocketAddress inetSocketAddress) throws SocketEx
...
+            }
+            index++;
+        }
+        switch (last - index) {
+            case 0:
+                return true;
+            case 1:
+                return !Character.isLetterOrDigit(format.charAt(last));
+        }
+        return false;
```
holgerfriedrich, line 654: Could you pls add a comment what this part of the method is trying to achieve?

Actually, it took me some time to get what you are trying to do (hopefully I got it correctly): 
* When there is no letter in the format string -> index will be last +1, switch will not hit a case -> false
* When there exactly one letter as last char of the format string -> the loop breaks at index=last -> switch case 0, true
* else if the letter is the 2nd last character -> break, switch will hit case "1" -> true if last char is a special character (!=letter/digit)
* else false....

seems to be the right thing... a regex-like description could be "[number]*[letter][!letterdigit]?" (this is not a proposal to use regex, comment would be sufficient)

holgerfriedrich (changes requested):
@andrewfg I am sorry for keeping you waiting so long.
Could you please fix the SAT warning? I will check again soon, promised.

bundles/org.openhab.core.config.discovery.addon.ip/src/main/java/org/openhab/core/config/discovery/addon/ip/IpAddonFinder.java
andrewfg: Removed

holgerfriedrich (approved):
LGTM, thanks!

PR #4491, opened 2024-12-10
[ephemeris] Add option for disabling the cache

clinique:
Adds possibility to disable user defined holiday description files caching.
This in interesting especially when defining/debugging these files.
A new option can be set  in the ephemeris.cfg file, `cache=true/false`.
Default value is true, so no breaking change if the option is absent.

### DIFF
--- a/bundles/org.openhab.core.ephemeris/src/main/java/org/openhab/core/ephemeris/internal/EphemerisManagerImpl.java
+++ b/bundles/org.openhab.core.ephemeris/src/main/java/org/openhab/core/ephemeris/internal/EphemerisManagerImpl.java
@@ -81,6 +81,7 @@ public class EphemerisManagerImpl implements EphemerisManager, ConfigOptionProvi
     public static final String CONFIG_COUNTRY = "country";
     public static final String CONFIG_REGION = "region";
     public static final String CONFIG_CITY = "city";
+    public static final String CONFIG_CACHE_FILES = "cache";
 
     private static final String RESOURCES_ROOT = "jollyday/";
     private static final String JOLLYDAY_COUNTRY_DESCRIPTIONS = RESOURCES_ROOT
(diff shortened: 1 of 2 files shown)


clinique:
The caching behaviour has been identified as a source of problems many times in the forum.
[Current workaround](https://community.openhab.org/t/updated-custom-ephemeris-file-required-openhab-restart-to-be-effectivein/160535) being to rename the file for every modification.

wborn:
It would be nicer if it would monitor changes to the file instead (e.g. using the `WatchService`) and then update the cached configuration.

clinique:
That was my first thought but I though this could be a first easy step. I blocked myself because ephemeris does not have a dedicated folder for user defined files, so they can be anywhere but I'll have a look at WatchService also.

kaikreuzer:
@clinique Could you please have a look at the failing tests?

kaikreuzer:
@clinique Any idea why the tests keep failing? It seems unrelated, but it is still failing, even after retriggering.

wborn:
Following up on the earlier `WatchService` discussion: there are actually two cache layers involved here.

The current PR avoids adding a custom-file `HolidayManager` to openHAB's `holidayManagers` cache when caching is disabled, but `HolidayManager.getInstance()` has its own internal Jollyday cache. That cache is not invalidated by this PR, so requesting the same URL can still return the previously cached manager and changes to the XML file may not be picked up.

There is also a similar issue in the openHAB cache when changing this setting at runtime. If a manager was already cached while caching was enabled, changing `cache` to `false` does not remove it because `getHolidayManager()` returns an existing entry before checking `cacheFiles`.

The cleanest approach is still to keep caching enabled and use openHAB's existing `WatchService` to detect changes to custom Jollyday configuration files. When a file changes, the corresponding entry should be removed from openHAB's `holidayManagers` cache and the Jollyday cache should be invalidated as well. The next access can then recreate the manager and reload the file.

On current `main`, files below `$OH_CONF` can use the existing configuration watcher. For files elsewhere on the filesystem, `WatchServiceFactory` supports creating a watcher for another base directory, although the lifecycle of those dynamically created watcher services would need to be managed.

Jollyday supports clearing its complete manager cache with `HolidayManager.clearManagerCache()`, but it does not provide targeted invalidation for a specific `ManagerParameter`/cache key. This is still the case in the most recent Jollyday release: upgrading Jollyday alone would therefore not solve the cache invalidation problem. Its internal cache supports retrieving entries and clearing the entire cache, but not removing an individual entry. Since URL-based managers already have a stable cache key, adding targeted invalidation upstream in Jollyday could be a useful improvement.

Clearing the complete Jollyday cache on a file-change event could still be an acceptable initial solution since these changes should be infrequent. A targeted upstream API would mainly make the invalidation more precise and avoid unnecessarily clearing unrelated Jollyday managers.

This would make the new `cache=false` option unnecessary for the stated use case: normal accesses stay cached, while modifications to custom files automatically cause them to be reloaded. A test that loads a custom holiday file, modifies it, and verifies that the next Ephemeris access sees the updated contents would also cover the expected behavior.

The current changes also appear to explain at least one of the failing tests. `getDaysUntil()` now calls `Optional#get()`, which throws `NoSuchElementException` when the requested holiday does not exist. The existing `testNextBankHoliday()` deliberately calls `getDaysUntil(..., "NONEXISTING")` and expects the previous behavior of returning `-1`.

PR #4497, opened 2024-12-16
Warning chase in org.openhab.core.audio

clinique:
From 24 to 2. Remains deprecated usage of FixedLengthAudioStream

### DIFF
--- a/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioFormat.java
+++ b/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioFormat.java
@@ -457,13 +457,13 @@ public class AudioFormat {
     public int hashCode() {
         final int prime = 31;
         int result = 1;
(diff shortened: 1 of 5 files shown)


bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioFormat.java
```diff
@@ -457,13 +457,13 @@ public boolean equals(@Nullable Object obj) {
     public int hashCode() {
         final int prime = 31;
         int result = 1;
-        result = prime * result + ((bigEndian == null) ? 0 : bigEndian.hashCode());
```
holgerfriedrich, line 460: Maybe you could use ```Object```, it should already include ```hashCode``` (and is might be overwritten in the derived objects, still works as expected).
```Object``` would still work if someone changes the datatypes above. Just imagine someone to change Integer to String - then we would fail silently (drop to default value "0").

holgerfriedrich (commented):
see comment, otherwise LGTM

holgerfriedrich (approved):
LGTM, thanks!

PR #4483, opened 2024-12-08
Remove Java 17 build, switch to 21

holgerfriedrich:
To be delayed until 4.3 is released.

PR #4499, opened 2024-12-16
Revert LinkedTransferQueue high CPU usage workaround

wborn:
Reverts the workaround (https://github.com/openhab/openhab-core/pull/3756) that was introduced for [JDK-8301341](https://bugs.openjdk.org/browse/JDK-8301341) because it is fixed in OpenJDK 21 (https://github.com/openjdk/jdk/pull/14317).

PR #4500, opened 2024-12-16
Run itests with Java 21

holgerfriedrich:
Missed that in #4483


wborn:
It's already part of https://github.com/openhab/openhab-core/pull/4499.

PR #4501, opened 2024-12-17
Add README for Binding Archetype Template

hendrickhere:
This PR adds a README.md to document the OpenHAB binding archetype template, explaining its structure and usage.


wborn (changes requested):
Thanks for working on the archetype documentation.

I do not think this PR is ready in its current form. The change has grown substantially beyond the stated goal of adding a README: it also modifies the Thing-model module, adds an impact analysis and diagram, and changes generated Java sources.

The Java template changes currently make generated bindings fail to compile, while the documented Maven invocation does not use the actual archetype coordinates

wborn (commented):
Adding the inline comments that were omitted from my previous request-changes review.

tools/archetype/binding/README.md
```diff
@@ -0,0 +1,35 @@
+# OpenHAB Binding Archetype
```
wborn, line 1: Please use the official project spelling, `openHAB`, consistently throughout the documentation.

There are several occurrences of `OpenHAB` in the files added by this PR that should be updated.

tools/archetype/binding/README.md
```diff
@@ -0,0 +1,35 @@
...
+---
+
+## How to Use This Archetype
+
+To create a new OpenHAB binding using this template:
+
+1. **Generate a New Binding Project**  
+   Use Maven to generate a project based on this archetype:
+   ```bash
+   mvn archetype:generate \
```
wborn, line 30: This command does not reference the archetype defined by this module. Its actual coordinates are:

```text
org.openhab.core.tools.archetypes:org.openhab.core.tools.archetypes.binding
```

Generation also needs an archetype version and required properties such as `bindingId`, `bindingIdCamelCase`, `author`, and `githubUser`.

Could this section use, or point developers to, `openhab-addons/bundles/create_openhab_binding_skeleton.sh` instead? That script already contains the supported invocation and would prevent this documentation from drifting out of sync.

tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/internal/__bindingIdCamelCase__Configuration.java
```diff
@@ -32,3 +32,12 @@ public class ${bindingIdCamelCase}Configuration {
     public String password = "";
     public int refreshInterval = 600;
 }
+
+@Test
```
wborn, line 36: The configuration class has already been closed above this line, so this creates an illegal top-level Java method. In addition, `@Test` and `assertEquals` are neither imported nor available to this production-source template.

Please remove this change. If the archetype should generate tests, they need to be placed in a separate `src/test/java` template and included explicitly in `archetype-metadata.xml`.

tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/internal/__bindingIdCamelCase__Handler.java
```diff
@@ -104,4 +104,9 @@ public void initialize() {
         // updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR,
         // "Can not access device as username and/or password are invalid");
     }
+
+    long startTime = System.currentTimeMillis();
+    initialize();
```
wborn, line 109: These statements are in the class body, after the `initialize()` method has already closed. Calls such as `initialize();` and `logger.debug(...)` are not legal class-member declarations, so every handler generated from this archetype will fail to compile.

Moving this call into `initialize()` would instead cause infinite recursion. Please remove this unrelated timing change entirely.

tools/archetype/binding/src/main/resources/archetype-resources/impact_analysis.md
```diff
@@ -0,0 +1,47 @@
+# Impact Analysis for Binding Module
```
wborn, line 1: This generic impact analysis and `call-graph.png` should not be part of the binding archetype. We do not want these files to end up in every add-on created using the binding skeleton, so they should be removed rather than added to `archetype-metadata.xml`.

If some of the architectural information is useful, it would be better contributed separately to the binding developer documentation in `openhab/openhab-docs`, where binding architecture and lifecycle are already documented and the information will be more discoverable.

bundles/org.openhab.core.model.thing/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Key Components
+
+### Formatting (`formatting/`)
+
+- `ThingFormatter.xtend`: Handles the formatting of Thing definitions in the OpenHAB configuration syntax.
+
+### Generator (`generator/`)
+
+- `ThingGenerator.xtend`: Generates the necessary code for Thing implementations based on model definitions.
```
wborn, line 15: This description is no longer applicable. In the revision on which this PR is based, `ThingGenerator.doGenerate()` was empty apart from commented example code.

`ThingGenerator.xtend` has since been deliberately removed by openHAB Core PR #5508 because these generated classes were unused and did not generate anything:

https://github.com/openhab/openhab-core/pull/5508

More generally, this README and the accompanying `GenericThingProvider` documentation changes concern the Thing-model module and are unrelated to the binding-archetype README described by this PR.

Could all Thing-model documentation enhancements be moved to a separate PR and rebased onto the current `main` branch? That would keep this PR focused and ensure the documentation reflects the module’s current structure and implementation.

PR #4505, opened 2024-12-19
Adding null annotations to config.core.internal.normalization 

clinique:
Null annotations added
Code evolution to Java 17

### DIFF
--- a/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/normalization/AbstractNormalizer.java
+++ b/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/normalization/AbstractNormalizer.java
@@ -12,6 +12,8 @@
  */
 package org.openhab.core.config.core.internal.normalization;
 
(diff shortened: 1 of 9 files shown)


bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/normalization/AbstractNormalizer.java
```diff
@@ -21,19 +23,16 @@
...
-            return null;
-        }
-        if (value instanceof String && "".equals(value)) {
-            return "";
-        }
-        return doNormalize(value);
+    public final @Nullable Object normalize(@Nullable Object value) {
+        return value == null ? null : //
+                "".equals(value) ? "" : //
+                        doNormalize(value);
```
J-N-K, line 35: IMO this makes it less readable.

clinique: @J-N-K : reversed

kaikreuzer (approved):
lgtm, thanks!

PR #4508, opened 2024-12-20
Let's also use records for configuration files.

clinique:
This PR enables the usage of record for configuration DTOs.
Comes along with a bit of revamp of the ConfigParser class.


kaikreuzer (approved):
lgtm, thanks!

PR #4513, opened 2024-12-22
[automation] Add provider script extension

florian-h05:


florian-h05:
@openhab/core-maintainers Can you please review this for 5.0? This is part of a bigger configuration as code effort for JS Scripting, but I wanted to have as much of its code in core so it can be used across all languages.

J-N-K:
What would be the use-case for temporary items oder things?

florian-h05:
The use case is to provide Items and Things programmatically from scripts without persisting them in the JSONDB. Conceptually much like the old DSL .items and .things files, but instead of having a DSL for definition you have fully fledged programming language available and can e.g. provide 10 Items through a loop.

kaikreuzer (approved):
lgtm, thanks!

holgerfriedrich:
@florian-h05 This PR seems to break the itests, currently causing our core build to fail (there has been another issue with antlr causing the build to fail, but this was fixed in the meantime).

Nadahar:
This might be nitpicking, but I find it confusing that this is placed in `org.openhab.core.automation.module.script.rulesupport`. Wouldn't e.g. `org.openhab.core.model.thing.scriptsupport`/`org.openhab.core.model.item.scriptsupport` be more suitable? It's hard to move things when they get incorporated into other code, so if they should be moved, the sooner the better.

Personally, I'm confused as to exactly what `model` means in OH, so I can't say if the package names should include `model` or not. I originally though `model` indicated that it was related to xtext/xtend, but I've seen it used so many places where this isn't the case lately that I no longer have any idea exactly what it means. Sometimes one can also believe that it is related to "semantic model", but that doesn't make sense for most of the `model` designated code.

florian-h05:
Hmm, I would rather place them into a new bundle `org.openhab.core.automation.module.script.modelsupport` as the functionality of this PR allows to provide the model (of Items and Things) from scripts.
However I am unsure as well. WDYT @kaikreuzer? Is the current location okay or should we move this new functionality?

Nadahar:
> Hmm, I would rather place them into a new bundle `org.openhab.core.automation.module.script.modelsupport` as the functionality of this PR allows to provide the model (of Items and Things) from scripts.

That depends on what "model" actually means, which I'm not sure of. But, I don't understand why it's under `automation` at all. To me, "automation" means "rules with accessories", where "accessories" are modules (actions, triggers and conditions) and rule templates. I see that "scripting" has also sort of ended up under automation, but I assume that's because it has been seen as tightly coupled with rules.

To me, Things and Items aren't "automation", automation is the system you build to automate things and items using triggers, conditions and actions.

"Automation modules" is at least clearly defined: [Developing automation modules](https://www.openhab.org/docs/developer/module-types/#developing-automation-modules), and it should not be under that as far as I can understand.

kaikreuzer:
@Nadahar You are correct, `model` packages used to refer to anything that was generated through Xtext, i.e. DSL stuff. It does NOT refer to the semantic model of items. Do you have examples of where stuff ended up in a `model` package that is not related to DSLs?

The `automation` package was introduced with the "new" rule engine that was implemented and which was independent of Xtext/DSLs, but introduced the modular approach through automation modules. The `script.rulesupport` package then again was introduced for any JSR223 scripts on top of the rule engine.

I understood this PR to allow scripts (scripts being considered as "action modules" as part of the automation space) to create items&things from within rules/scripts - I therefore had no problem with the chosen namespace.

Nadahar:
> Do you have examples of where stuff ended up in a `model` package that is not related to DSLs?

I think I've seen some other things as well, but what I have fresh in mind is the "new YAML format" in [org.openhab.core.model.yaml](https://github.com/openhab/openhab-core/tree/main/bundles/org.openhab.core.model.yaml)

> I understood this PR to allow scripts (scripts being considered as "action modules" as part of the automation space) to create items&things from within rules/scripts - I therefore had no problem with the chosen namespace.

Ok, I've understood it slightly differently, which is why I thought the namespace was strange. What I thought this was, is a generic way to create Things and Items using JSR223, as an alternative to managed or file-based (DSL or YAML) providers. As such, they aren't really related to automation. So, I guess the answer to the question of which namespace they fit into is the precise nature of the code in this PR. I haven't studied it that deep, I just looked through it and found the namespace strange based on my perception of the content.

kaikreuzer:
> I think I've seen some other things as well, but what I have fresh in mind is the "new YAML format" in [org.openhab.core.model.yaml](https://github.com/openhab/openhab-core/tree/main/bundles/org.openhab.core.model.yaml)

Ah, ok, I see. You're right, that might have maybe better gone to a package like `org.openhab.core.config.yaml` or so...

PR #4514, opened 2024-12-22
Update Eclipse files

wborn:
* Updates the JDK version
* Adds missing files
* Updates files so there are no changes when updating Maven projects

### DIFF
--- a/bundles/org.openhab.core.model.core/.classpath
+++ b/bundles/org.openhab.core.model.core/.classpath
@@ -9,6 +9,7 @@
 	<classpathentry excluding="**" kind="src" output="target/classes" path="src">
 		<attributes>
 			<attribute name="maven.pomderived" value="true"/>
+			<attribute name="optional" value="true"/>
 		</attributes>
 	</classpathentry>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-17">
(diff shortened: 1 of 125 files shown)


bundles/org.openhab.core.model.core/.classpath
```diff
@@ -9,6 +9,7 @@
 	<classpathentry excluding="**" kind="src" output="target/classes" path="src">
 		<attributes>
 			<attribute name="maven.pomderived" value="true"/>
+			<attribute name="optional" value="true"/>
 		</attributes>
 	</classpathentry>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-17">
```
holgerfriedrich, line 15: Is it intentional to leave JavaSE-17 here? There are a few more occurrences in other files as well.....

wborn: I also noticed that but it is expected because of this Xtext workaround:

https://github.com/openhab/openhab-core/blob/12f2314a2d1cbd3157bead4b6de9159bea37275d/bundles/pom.xml#L281-L282

holgerfriedrich: I have built a distro with Karaf 4.4.7 SNAPSHOT - purely on Java 21 - today. No problems so far, tested a few addons.

I have not compared the generated code, though, AFAIR we had compilation errors in generated Java code when I introduced this workaround. In the meantime xtext/xtend has been updated.
As we are in an early phase of 5.0, I think we could take the risk and remove this workaround.

holgerfriedrich (approved):
Let's merge this separately, LGTM.

PR #4515, opened 2024-12-22
[websocket] Support token authentication through header

### LINKED ISSUE #4490: Allow passing the access token in Sec-WebSocket-Protocol header to authorize WebSocket connections
Currently when establishing a WebSocket connection the access token can only be practically passed via the accessToken URI query parameter.
This is generally frowned upon since they might end up in web server/proxies logs. It's not that critical for access tokens because they're short-lived (1h) but it's still not ideal.
Browsers cannot add Authorization headers to WebSocket requests but can specify something to add to the `Sec-WebSocket-Protocol` header; and some (like [Kubernetes](https://github.com/kubernetes/kubernetes/commit/714f97d7baf4975ad3aa47735a868a81a984d1f0)) misuse this to send the token.

It might be nice to have that option as well.

More info:
https://stackoverflow.com/a/77060459 (option 5, note that option 6 suggests... switching to SSE)
https://ably.com/blog/websocket-authentication

## Your Environment
openHAB 4.3.0.M5

florian-h05:
Closes #4490.

Browser clients do not have the ability to set an `Authorization` header programatically on websocket requests. All they have control over is the URL and the websocket subprotocols sent (see https://developer.mozilla.org/en-US/docs/Web/API/WebSocket)

This PR adds support for specifying a bearer token via a websocket subprotocol, with the format `org.openhab.ws.accessToken.base64.<encoded-token>`.

The browser must specify at least one other subprotocol since the server must echo a selected subprotocol back.
For this, `org.openhab.ws.protocol.default` is used.

`<encoded-token>` must be base64-encoded without padding, since bearer tokens can contain characters a websocket protocol may not (`/` and `=`).

### DIFF
--- a/bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/CommonWebSocketServlet.java
+++ b/bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/CommonWebSocketServlet.java
@@ -121,7 +121,13 @@ public class CommonWebSocketServlet extends WebSocketServlet {
                 Matcher matcher = WEBSOCKET_ACCESS_TOKEN_PATTERN.matcher(secWebSocketProtocolHeader);
                 if (matcher.find() && matcher.group("base64") != null) {
                     String base64 = matcher.group("base64");
-                    accessToken = new String(Base64.getDecoder().decode(base64));
+                    try {
+                        accessToken = new String(Base64.getDecoder().decode(base64));
+                    } catch (IllegalArgumentException e) {
+                        logger.warn("Invalid base64 encoded access token in Sec-WebSocket-Protocol header from {}.",
+                                servletUpgradeRequest.getRemoteAddress());
+                        return null;
+                    }
                 } else {
                     logger.warn("Invalid use of Sec-WebSocket-Protocol header from {}.",
                             servletUpgradeRequest.getRemoteAddress());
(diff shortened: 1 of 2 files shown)


florian-h05:
@ghys This was easier to implement than I expected :+1:

ghys:
@florian-h05 you're underestimating yourself ;)

bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/CommonWebSocketServlet.java
```diff
@@ -94,7 +111,25 @@ private class CommonWebSocketCreator implements WebSocketCreator {
...
+            String accessToken = null;
+            String secWebSocketProtocolHeader = servletUpgradeRequest.getHeader(SEC_WEBSOCKET_PROTOCOL_HEADER);
+            if (secWebSocketProtocolHeader != null) { // if the client sends the Sec-WebSocket-Protocol header
+                // respond with the default protocol
+                servletUpgradeResponse.setHeader(SEC_WEBSOCKET_PROTOCOL_HEADER, WEBSOCKET_PROTOCOL_DEFAULT);
+                // extract the base64 encoded access token from the requested protocols
+                Matcher matcher = WEBSOCKET_ACCESS_TOKEN_PATTERN.matcher(secWebSocketProtocolHeader);
+                if (matcher.find() && matcher.group("base64") != null) {
+                    String base64 = matcher.group("base64");
+                    accessToken = new String(Base64.getDecoder().decode(base64));
```
holgerfriedrich, line 124: This may throw `IllegalArgumentException` if given token has invalid coding. Should this be handled?

florian-h05: Yes, I will fix that.

florian-h05: Done

holgerfriedrich (approved):
LGTM, thanks!

PR #4516, opened 2024-12-22
Enhance `DateTimeType` to parse `ZonedDateTime` with time-zone ID

jlaur:
This DSL rule:
```java
rule "Test"
when
    Item Test changed to ON
then
    var zoned = ZonedDateTime.parse("2024-12-22T16:29+01:00")
    logInfo("zoned to item", zoned.toString)
    TestDateTime.postUpdate(zoned.toString)

    zoned = (TestDateTime.state as DateTimeType).getZonedDateTime()
    logInfo("zoned from item", zoned.toString)

    TestDateTime.postUpdate(zoned.toString)
end
```

generates the following output in 4.3:
```
2024-12-23 00:27:00.168 [INFO ] [nhab.core.model.script.zoned to item] - 2024-12-22T16:29+01:00
2024-12-23 00:27:00.169 [INFO ] [ab.core.model.script.zoned from item] - 2024-12-22T16:29+01:00[Europe/Copenhagen]
2024-12-23 00:27:00.170 [WARN ] [b.core.model.script.actions.BusEvent] - Cannot convert '2024-12-22T16:29+01:00[Europe/Copenhagen]' to a state type which item 'TestDateTime' accepts: [DateTimeType, UnDefType].
```

Until 4.2 this rule worked, because `getZonedDateTime()` returned the same `ZonedDateTime` as provided when creating the `DateTimeType`. Since it didn't necessarily contain time-zone ID but only offset (e.g. "2024-12-22T16:29+01:00"), `DateTimeType.parse` would then be able to parse it in some scenarios. Now, it contains the time-zone ID (e.g. "2024-12-22T16:29+01:00[Europe/Copenhagen]") always, which `DateTimeType.parse` is not able to parse (and never was).

To make this rule compatible I would propose to add support for parsing a full `ZonedDateTime` string that includes time-zone ID. If this is accepted, I think it should also be cherry-picked into 4.3.x for immediate value. i.e. to fix rules that might have been broken by 4.3. By the time we reach 5.0, users would otherwise have been forced to refactor their rules already.

See https://community.openhab.org/t/datetime-conversion-after-update-to-4-3-0/161053/1

Regression of #3583

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java
@@ -302,8 +302,12 @@ public class DateTimeType implements PrimitiveType, State, Command, Comparable<D
                 try {
                     date = ZonedDateTime.parse(value, PARSER_TZ);
                 } catch (DateTimeParseException tzException) {
-                    LocalDateTime localDateTime = LocalDateTime.parse(value, PARSER);
-                    date = ZonedDateTime.of(localDateTime, ZoneId.systemDefault());
+                    try {
+                        date = ZonedDateTime.parse(value);
+                    } catch (DateTimeParseException e) {
+                        LocalDateTime localDateTime = LocalDateTime.parse(value, PARSER);
+                        date = ZonedDateTime.of(localDateTime, ZoneId.systemDefault());
+                    }
                 }
             }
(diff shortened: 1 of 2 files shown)


kaikreuzer (approved):
lgtm, thanks!

kaikreuzer:
Backported to `4.3.x`.

PR #4519, opened 2024-12-24
Remove workaround for Xtend/Java21

holgerfriedrich:
Model generation was pinned to Java 17, as Xtend generator caused issues at the time we introduced Java 21.
After the last updates, this seems no longer be the case.

References:
- https://github.com/openhab/openhab-core/pull/4514#discussion_r1895586941
- #4161


wborn (approved):
It's always nice when workarounds can be removed. 👍

PR #4524, opened 2024-12-28
Clean up arguments after executing Script Transformation

### LINKED ISSUE #4414: Parameters injected to JS transformations are not resetted across further invocations of the transformation script
## Background

(I'm on Openhab 4.3.0 M2 on a Raspberry Pi 4. This is the first issue I filed for Openhab, pls bear with me for any mistakes. My Openhabian system is manually upgraded to bullseye, although this is not supported... But I thought this issue still might be worth to get investigated since it is easily reproducible...)

Problem: When passing adding additional parameters to toItemScript as with `?fig=2` in the following snippet...

```
Number:Temperature Refoss_P11d_Temp "Temperature" { channel="mqtt:topic:openhab:refossp11d:temperature" [profile="transform:JS",toItemScript="significant.js?fig=2"], expire="30m" }
Number:ElectricCurrent Refoss_P11d_Current "Current [%.1f %unit%]"  { channel="mqtt:topic:openhab:refossp11d:current" [profile="transform:JS",toItemScript="significant.js"], expire="30m"}

```
this parameter is *shared* across all invocations of significant.js, i.e. fig is always 2 when significant.js is called, even when it is not explicitly given as in my second item above.
This is rather surprising behavior, and I was wondering whether it's a bug or a feature in M2 (or before?)

PS: Update: It seems that one has to manually set the injected variable to *undefined* at the end of the transformation script in order to have reset before the next invocation - IMHO this is not an easily expected behavior that at least should be well documented:

`fig = undefined; // reset fig to undefined for the next invocation
`

As [suggested in the forum](https://community.openhab.org/t/openhab-4-3-milestone-discussion/158139/47) by Florian H. I'm also filing an issue here:

## Expected Behavior
I'd expect any parameters that are NOT passed specifically (e.g. "?param1=val1&param2=val2") always be set to undefined....

## Current Behavior
When passing adding additional parameters (e.g. "significant.js?param1=val&param2=val") to a JS transformation script via toItemScript() - as with `?fig=2` above - these parameters are *shared* across all invocations of significant.js, i.e. fig is always 2 when significant.js is called, even when it is not explicitly given as in my second item above

## Possible Solution
Injected parameters should be implicitly reset to undefined before further invocations of the transformation script

Here's is a JS transformation to reproduce:
```
// return the input value unchanged, logging arg1 and arg2 if defined
(function(i) {
    // console.log("donothing.js: started");
    var strarg1 = "" ;
    var strarg2 = "" ;
    if (typeof arg1 !== 'undefined') { 
        strarg1 = "    arg1=" + arg1.toString() + " (" + typeof arg1 + ")";
    }
    if (typeof arg2 !== 'undefined') { 
        strarg2 = "    arg2=" + arg2.toString() + " (" + typeof arg2 + ")";
    }
    console.log("donothing.js: i=" + i.toString() + strarg1  + strarg2);
    arg1 = "UNCHANGED";
    arg2 = "UNCHANGED";
    return i;
})(input)
```

Use this as sample transformation as in
Number:Energy Refoss_TotalYesterday "Total Yesterday [%d %unit%]" { channel="mqtt:topic:openhab:refossp11b:yesterday"  [profile="transform:JS",toItemScript="donothing.js?arg1=ONE"], expire="90m"  }
Number:Energy Refoss_TotalToday "Total Today [%d %unit%]"  { channel="mqtt:topic:openhab:refossp11b:today" [profile="transform:JS",toItemScript="donothing.js?arg2=TWO"], expire="90m" }

will give output like this, showing the behavior:
```
2024-10-13 14:14:11.963 [INFO ] [org.openhab.automation.script       ] - donothing.js: i=0.0    arg1=ONE (string)    arg2=UNCHANGED (string)
2024-10-13 14:14:11.968 [INFO ] [org.openhab.automation.script       ] - donothing.js: i=0.0    arg1=UNCHANGED (string)    arg2=a (string)
2024-10-13 14:31:04.997 [INFO ] [org.openhab.automation.script       ] - donothing.js: i=0.0    arg1=ONE (string)    arg2=UNCHANGED (string)
2024-10-13 14:31:05.003 [INFO ] [org.openhab.automation.script       ] - donothing.js: i=0.0    arg1=UNCHANGED (string)    arg2=a (string)
2024-10-13 14:41:09.819 [INFO ] [org.openhab.automation.script       ] - donothing.js: i=0.0    arg1=ONE (string)    arg2=UNCHANGED (string)
2024-10-13 14:41:09.826 [INFO ] [org.openhab.automation.script       ] - donothing.js: i=0.0    arg1=UNCHANGED (string)    arg2=a (string)
2024-10-13 15:17:07.687 [INFO ] [org.openhab.automation.script       ] - donothing.js: i=0.0    arg1=ONE (string)    arg2=UNCHANGED (string)
2024-10-13 15:17:07.693 [INFO ] [org.openhab.automation.script       ] - donothing.js: i=0.0    arg1=UNCHANGED (string)    arg2=a (string)
```

## Context/Rationale
Background: The significant.js script mentioned initially is my transformation script for rounding incoming values from sensors to a more sensible, reduced number of significant figures/digits depending on the unit of the measurement. [ADDED: This would cause fewer updates for the values actually fed into the receiving item.] I want to use fig= to override my unit specific default, e.g. 2 for temperatures.)

## Your Environment
* OpenHab 4.3 M2 (and maybe before)
* Addons.cfg:
```
[automation = jsscripting
binding = astro,network,mqtt,tr064,mail,avmfritz,dwdunwetter,exec,amazonechocontrol,amazondashbutton,bluetooth,homeconnect,hpprinter,samsungtv,openweathermap,dwdpollenflug,weathercompany,telegram,mybmw,lgwebos
transformation = jinja,jsonpath,map,regex,basicprofiles,scale,exec
](url)
```

* Version on "Raspbian GNU/Linux 12 (bookworm)" (manually upgraded from 11, although not officially supported!)":
```
$ java --version
openjdk 17.0.12 2024-07-16
OpenJDK Runtime Environment (build 17.0.12+7-Raspbian-2deb12u1rpt1)
OpenJDK Client VM (build 17.0.12+7-Raspbian-2deb12u1rpt1, mixed mode, emulated-client)

```

jimtng:
Fix #4414

Note that the problem didn't occur with JRuby scripting, but it does occur with JS Scripting.

Perhaps this should be backported to 4.3.x?

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptTransformationService.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptTransformationService.java
@@ -17,6 +17,7 @@ import static org.openhab.core.automation.module.script.profile.ScriptProfileFac
 import java.net.URI;
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Locale;
@@ -185,8 +186,10 @@ public class ScriptTransformationService implements TransformationService, Confi
                         : scriptEngineContainer.getScriptEngine();
                 ScriptContext executionContext = engine.getContext();
                 executionContext.setAttribute("input", source, ScriptContext.ENGINE_SCOPE);
+                ArrayList<String> injectedParams = null;
 
                 if (params != null) {
+                    injectedParams = new ArrayList<>();
                     for (String param : params.split("&")) {
                         String[] splitString = param.split("=");
                         if (splitString.length != 2) {
@@ -197,6 +200,7 @@ public class ScriptTransformationService implements TransformationService, Confi
                             param = URLDecoder.decode(splitString[0], StandardCharsets.UTF_8);
                             String value = URLDecoder.decode(splitString[1], StandardCharsets.UTF_8);
                             executionContext.setAttribute(param, value, ScriptContext.ENGINE_SCOPE);
+                            injectedParams.add(param);
                         }
                     }
                 }
@@ -210,8 +214,16 @@ public class ScriptTransformationService implements TransformationService, Confi
                     scriptRecord.compiledScript = compiledScript;
                 }
 
-                Object result = compiledScript != null ? compiledScript.eval() : engine.eval(scriptRecord.script);
-                return result == null ? null : result.toString();
+                try {
+                    Object result = compiledScript != null ? compiledScript.eval() : engine.eval(scriptRecord.script);
+                    return result == null ? null : result.toString();
+                } finally {
+                    if (injectedParams != null) {
+                        for (String param : injectedParams) {
+                            executionContext.removeAttribute(param, ScriptContext.ENGINE_SCOPE);
+                        }
+                    }
+                }
             } catch (ScriptException e) {
                 throw new TransformationException("Failed to execute script.", e);
             } catch (IllegalStateException e) {
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptTransformationService.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptTransformationService.java
@@ -219,9 +219,8 @@ public class ScriptTransformationService implements TransformationService, Confi
                     return result == null ? null : result.toString();
                 } finally {
                     if (injectedParams != null) {
-                        for (String param : injectedParams) {
-                            executionContext.removeAttribute(param, ScriptContext.ENGINE_SCOPE);
-                        }
+                        injectedParams
+                                .forEach(param -> executionContext.removeAttribute(param, ScriptContext.ENGINE_SCOPE));
                     }
                 }
             } catch (ScriptException e) {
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/ScriptTransformationService.java
```diff
@@ -210,8 +214,16 @@ public void deactivate() {
...
                 }
 
-                Object result = compiledScript != null ? compiledScript.eval() : engine.eval(scriptRecord.script);
-                return result == null ? null : result.toString();
+                try {
+                    Object result = compiledScript != null ? compiledScript.eval() : engine.eval(scriptRecord.script);
+                    return result == null ? null : result.toString();
+                } finally {
+                    if (injectedParams != null) {
+                        for (String param : injectedParams) {
```
holgerfriedrich, line 222: I would prefer `forEach`, but this is just a suggestion.

```suggestion
                                                injectedParams
                                .forEach(param -> executionContext.removeAttribute(param, ScriptContext.ENGINE_SCOPE));
```

jimtng: good idea, that looks nicer indeed

holgerfriedrich (approved):
LGTM, thanks!

PR #4525, opened 2024-12-28
Add sharedCache and privateCache to file-based RulesDSL

### LINKED ISSUE #4447: sharedCache broken for rules DSL
The shared cache is no longer available in rules DSL?

[The test rule mentioned here](https://community.openhab.org/t/unable-to-use-private-cache-in-rules-dsl/143552/5)
results in

2024-11-12 11:04:00.271 [ERROR] [ore.automation.module.script.internal.handler.ScriptActionHandler] - Script execution of rule with UID 'x-1' failed: The name 'sharedCache' cannot be resolved to an item or type; line 5, column 27, length 11 in x

There's hints on the forum it worked in 4.1 and broke somewhere between 4.1 and 4.2.

Related: #3187

jimtng:
Fix #4447

### DIFF
--- a/bundles/org.openhab.core.model.rule/bnd.bnd
+++ b/bundles/org.openhab.core.model.rule/bnd.bnd
@@ -31,6 +31,7 @@ Import-Package: \
  org.openhab.core.model.persistence.extensions,\
  org.openhab.core.model.script,\
  org.openhab.core.model.script.engine.action,\
+ org.openhab.core.automation.module.script.rulesupport.shared,\
  com.google.common.base;version="14",\
  javax.measure,\
  javax.measure.quantity,\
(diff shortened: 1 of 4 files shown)


holgerfriedrich (approved):
LGTM, thanks!

This now matches the docs https://www.openhab.org/docs/configuration/jsr223.html#dsl, but still we have some contradicting docs here: https://next.openhab.org/docs/tutorial/rules_advanced.html (there just ```cache``` is used, which is at least wrong for DSL).

jimtng:
Should this be backported?

jimtng:
> we have some contradicting docs here: https://next.openhab.org/docs/tutorial/rules_advanced.html (there just `cache` is used, which is at least wrong for DSL).

This is for JavaScript, so it's indeed `cache`, not `sharedCache`.

I will submit some minor fixes to the docs, but they are not related to this PR. I only noticed that while reading the linked docs.

holgerfriedrich:
as it is working on 5.x, backported to 4.3.x branch

Issue #4466, opened 2024-12-01
Enable http/2 for https & Set Keep-Alive header for http


florian-h05:
When having multiple tabs of Main UI opened, the UI often fails to send HTTP requests to the server because the browser does not execute the requests.
This is because we are running into the max parallel HTTP connections limit.

To fix this issue, we need to keep alive HTTP connections for http and/or enable http/2 for https.
Enabling http/2 for https also brings other benefits, and http/2 is supported by all major browser for several years now (see https://caniuse.com/http2).
This fix seems to work as I have never experienced the above problems with my production instance running behing nginx where I use http/2 and https.

See https://community.openhab.org/t/openhab-4-3-milestone-discussion/158139/75?u=florian-h05, https://www.geeksforgeeks.org/what-are-max-parallel-http-connections-in-a-browser/ and https://www.baeldung.com/jetty-http-2 for more details.

Unfortunately I do not really know how Jetty dependencies are added to openHAB and I am not really confident in configuring Jetty through the XML in openhab-distro, so I need some help.

@J-N-K Since you already added WebSocket and Gzip to Jetty, may I ask for your help here?

wborn:
I did some testing with openHAB and HTTP/2 awhile ago but it will only work with TLS so you also need a valid (self) signed certificate. I also got h2c (http/2 without TLS) working, however it's not supported by browsers :upside_down_face: .

florian-h05:
I am aware of the fact that HTTP/2 needs TLS, openHAB already ships with a valid self-signed certificate. Browsers are putting more and more restrictions on unencrypted connections, take microphone access on the SIP client widget as another example.

The described issue can be easily fixed for HTTPS by enabling HTTP/2, for HTTP only it should be fixable or at least improveable by setting the Keep-Alive header.
As you already had HTTP/2 running, can you please make it running again?

ghys:
I've often wondered if it could be a service offered by the OH Foundation through myopenhab.org like this:

You would have a way to configure a DNS name that would resolve to your private instance IP like:

`<user>-<hardtoguesssuffix>.private.myopenhab.org A <localip>`

and also have a way to retrieve a certificate and private key for this name and set it for your instance instead of the self-signed certificate
(or the openHAB Cloud add-on could perhaps even do it for you)

Then you could use that DNS name instead of your local IP to access your instance over HTTPS and it would work - a secure origin for a local instance.

[...]

PR #4526, opened 2024-12-28
Start SPI Fly (OSGi service loader) early

florian-h05:
Related to https://github.com/openhab/openhab-core/issues/4466.

### DIFF
--- a/features/karaf/openhab-tp/src/main/feature/feature.xml
+++ b/features/karaf/openhab-tp/src/main/feature/feature.xml
@@ -15,6 +15,12 @@
 -->
 <features name="${project.artifactId}-${project.version}" xmlns="http://karaf.apache.org/xmlns/features/v1.4.0">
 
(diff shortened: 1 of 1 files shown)


bom/runtime/pom.xml
```diff
@@ -775,6 +775,18 @@
       <version>${jetty.version}</version>
       <scope>compile</scope>
     </dependency>
+    <dependency>
```
florian-h05, line 778: FYI I am not sure whether we need those dependencies, but I would think so.

J-N-K:
Please resolve the itests and convert to a PR.

florian-h05:
My distro build works fine without adding `jetty-alpn-server` and `jetty-alpn-java-server` compile dependencies.
Do we need them or not? I cannot really find any information only, but it seems we do not need them as JDK 21 provides ALPN APIs.

J-N-K:
I don't think we need them as `compile` dependencies, it's enough that they are available at runtime.

J-N-K (approved):

PR #4528, opened 2024-12-30
Fix config normalization in addon configuration

J-N-K:
Reported on the forum. When installing add-ons from the marketplace the config-description could not be found because a wrong uid was used when retrieving it. As a result the configuration was not normalized and all numeric values were converted to a `Double`instead of the correct type.


digitaldan:
Thanks @J-N-K !    

I would vote to  back port this to 4.3.x

holgerfriedrich (approved):
LGTM

PR #4535, opened 2025-01-03
Java 21 language features

holgerfriedrich:
* use getFirst and getLast methods
* replace new Locale by Locale.of
* replace Paths.get by Path.of
* use ThreadLocalRandom.current().nextDouble()
* add @ Serial annotations


andrewfg:
> use getFirst and getLast methods

And maybe also replace `list.add(index, ..)` with `list.addFirst(..)` and `list.addLast(..)` for those special cases `index == 0` resp. `index == list.size() - 1`. Although probably the latter is quite rare in OH since any sane developer would have simply used `list.add(..)`. => WDYT?

holgerfriedrich:
> > use getFirst and getLast methods
> 
> And maybe also replace `list.add(index, ..)` with `list.addFirst(..)` and `list.addLast(..)` for those special cases `index == 0` resp. `index == list.size() - 1`. Although probably the latter is quite rare in OH since any sane developer would have simply used `list.add(..)`. => WDYT?

This should already be included. I have seen one occurrence in the code, `add(0, ...)` in PersistenceExtensions.java.

andrewfg (approved):
It looks very clean. In particular the `List.getFirst()` calls. => LGTM

I assume you are working on a similar clean up in addons? But probably wait until the current round of fixes backported into 4.3 will have been completed.

holgerfriedrich:
@andrewfg I have not started work on the add-ons repo. Last time when I introduced the Java 17 features (mainly instanceof patterns), there were a lot of complaints by add-on authors, as they intend to build and release for the previous 4.x release. I had deliberately waited an additional release cycle (so core used Java17 features one release before). We need to discuss how we want to approach this.

kaikreuzer (approved):
lgtm, many thanks!

kaikreuzer:
Wrt add-on repo: Yes, I assume there would be quite some push-back. We should probably wait until 5.0 is released, so that it is easy for maintainers to build 4.x versions of it. It will also help for backporting code to the `4.3.x` branch.

PR #4540, opened 2025-01-04
[websocket] Change EventWebSocketAdapter adapter id to `events`

florian-h05:
This is aligns the event WS name with the SSE events endpoint.
This is a breaking change for clients that have explicitly used `/ws/event-subscriber` instead of the default `/ws`.


kaikreuzer (approved):
lgtm

PR #4536, opened 2025-01-04
Upgrade plugins

holgerfriedrich:
* sortpom-maven-plugin to 4.0.0 (breaking change) https://github.com/Ekryd/sortpom?tab=readme-ov-file#news
* maven-clean-plugin to 3.4.0 https://github.com/apache/maven-clean-plugin/releases/tag/maven-clean-plugin-3.4.0
* maven-deploy-plugin to 3.1.3 https://github.com/apache/maven-deploy-plugin/releases/tag/maven-deploy-plugin-3.1.3
* maven-enforcer-plugin to 3.5.0 https://github.com/apache/maven-enforcer/releases
* maven-install-plugin to 3.1.3 https://github.com/apache/maven-install-plugin/releases
* maven-plugin-plugin to 3.15.1
* license-maven-plugin to 4.6 https://github.com/mathieucarbou/license-maven-plugin/releases
* maven-jar-plugin to 3.4.2 https://github.com/apache/maven-jar-plugin/releases/tag/maven-jar-plugin-3.4.2
* maven-archetype-plugin to 3.3.1
* maven-dependency-plugin to 3.8.1
* maven-javadoc-plugin to 3.11.2
* maven-release-plugin to 3.1.1
* maven-shade-plugin to 3.6.0
* maven-site-plugin to 3.21.0
* maven-surefire-plugin to 3.5.2
* build-helper-maven-plugin https://github.com/mojohaus/build-helper-maven-plugin/releases/tag/3.6.0
* i18n-maven-plugin to latest OH release 4.3.1

PR #4541, opened 2025-01-06
Fix JavaDoc output directory

holgerfriedrich:
Regression, fix issue introduced by #4536.
JavaDoc output dir changed from target/site to target/reports.
This will get it back to the known location, also fixing the JavaDoc CI build.


wborn (approved):
Thanks!

PR #4542, opened 2025-01-07
Remove thing status description for HANDLER_MISSING_ERROR

jimtng:
Let webui handle the description.
See: https://github.com/openhab/openhab-webui/pull/2986

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
@@ -933,8 +933,8 @@ public class ThingManagerImpl implements ReadyTracker, ThingManager, ThingTracke
                             thing.getThingTypeUID());
                 }
             } else {
-                setThingStatus(thing, buildStatusInfo(ThingStatus.UNINITIALIZED,
-                        ThingStatusDetail.HANDLER_MISSING_ERROR, "Handler factory not found"));
+                setThingStatus(thing,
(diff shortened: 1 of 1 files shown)


jlaur:
If I understand this correctly from the community post, this is only to make it translateable through the webui repository?

I think this can already be achieved here as well without introducing any fragmentation. See:

https://github.com/openhab/openhab-core/blob/f00c7700cb13b4ab6dbc6a8e493f226d8099689c/itests/org.openhab.core.thing.tests/src/main/resources/test-bundle-pool/thingStatusInfoI18nTest.bundle/OH-INF/i18n/test-translations.properties#L1

https://github.com/openhab/openhab-core/blob/f00c7700cb13b4ab6dbc6a8e493f226d8099689c/itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/i18n/ThingStatusInfoI18nLocalizationServiceOSGiTest.java#L249-L254

jimtng:
- There are several instances of HANDLER_MISSING_ERRORS and only one provided a description in core. This PR removes that message here so that all of them can be dealt with in the same way, in webui.
- The description for Things should (IMO) say "Binding" because "Handler" is yet another unfamiliar term that user may not be familiar with. This is done in the linked webui PR.
- The messages, afaik, are only applicable in webui, and changes in webui are more "agile" so it's easier to centralise such messages and their translations in webui rather than core. Webui also can tweak the exact wordings to better match the situations in which the messages are displayed.

jimtng:
Perhaps more descriptions can be added in core rather than augmented through Main UI, so that they're visible in the logs too. For that reason, I'm closing this PR.

PR #4546, opened 2025-01-08
HttpUtil: fix invalid URI

mherwege:
Providing an invalid URI input was throwing an IllegalArgumentException from the underlying jetty library. This change properly catches it and throws an IOException instead, aligned with the error treatment in the rest of the method.

See also https://github.com/openhab/openhab-addons/pull/18066

### DIFF
--- a/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/HttpUtil.java
+++ b/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/HttpUtil.java
@@ -152,6 +152,10 @@ public class HttpUtil {
     public static String executeUrl(String httpMethod, String url, Properties httpHeaders, InputStream content,
             String contentType, int timeout, String proxyHost, Integer proxyPort, String proxyUser,
             String proxyPassword, String nonProxyHosts) throws IOException {
+        if (url == null) {
+            LOGGER.debug("Execute url failed: no url provided");
+            return null;
+        }
         ContentResponse response = executeUrlAndGetReponse(httpMethod, url, httpHeaders, content, contentType, timeout,
                 proxyHost, proxyPort, proxyUser, proxyPassword, nonProxyHosts);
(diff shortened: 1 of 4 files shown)


holgerfriedrich:
@mherwege not a review yet, just a remark:
Since we are using Java 21 now, `URL(string)` ctors are anyway deprecated and need to be replaced by `URI.create(...).toUrl()`.
I have not managed to do this yet, but already got into trouble by the different handling of exceptions. URL throws MalformedURLException, URI.create may throw NullPointerException or IllegalArgumentException.
So this sounds familiar.

Can url be null? Then method `shouldUseProxy` could be problematic as well, as it may call `.matches` on a null string if exclusions are given.

mherwege:
> Since we are using Java 21 now, `URL(string)` ctors are anyway deprecated and need to be replaced by `URI.create(...).toUrl()`.

I don't think that is entirely what the documentation says. Ity says to replace it with `URI.toUrl()`. But to create the URI, `new URI(String str)` is still OK and the `URI.create(String str)` method documentation explicitely states that this convenience method should only be used when you are sure `str` is a well formed URI.
The issue I try to solve is that `HttpClient.newRequest(String str)` from Jetty expects a well formed URI as input, and it was called with a not checked URL. It is actually calling `URI.create(url)` internally. So I make sure it is a well formed URI before calling it. And indeed, that let to the `IllegalArgumentException`.
I didn't find any direct use of the `URL(String str)` constructor in this class.

> Can url be null?

As this is called from static public functions and never checked, I guess it could be. There are also no Null annotations. I will put in some checks for this.

> Then method `shouldUseProxy` could be problematic as well, as it may call `.matches` on a null string if exclusions are given.

This will also be solved by checking for a null url before calling the `executeUrlAndGetReponse` method.

mherwege:
> I didn't find any direct use of the `URL(String str)` constructor in this class.

I didn't look carefully. Replace now.

mherwege:
@holgerfriedrich I created a separate PR to remove all uses of the `URL` constructor from core.

holgerfriedrich (approved):
LGTM, thanks!

PR #4552, opened 2025-01-10
Ensure sitemap subscriptions are updated on Group Items

digitaldan:
Fixes an issue where sitemaps containing group items that change,  only update when a subscription times out (30 seconds)

I noticed while troubleshooting our IOS client that the sitemap i use for the apple watch would not update in realtime, only every 30 seconds.  Turns out this sitemap only has group items, and our SitemapResource was not subscribing to those events.  This is a pretty small change, i would recommend to back port as well.


wborn (approved):
Thanks!

PR #4555, opened 2025-01-11
Reduce SAT warnings

holgerfriedrich:
Address PMD warnings.
Cleanup PMD suppressions, add local `// NOPMD` tags where readability is preferred over PMD rule.

For the remaining SimplifyBooleanReturn warnings, no code change should be done. The code looks cleaner if we keep it as it is.
I did not try to silence this warnings yet. The `NOPMD` tag is not well suited, as it needs to be added to the line of the `if`. Looks ugly.

Before:
grafik

After:
grafik


bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/AbstractRemoteAddonService.java
```diff
@@ -298,11 +298,9 @@ protected boolean remoteEnabled() {
         try {
             Configuration configuration = configurationAdmin.getConfiguration("org.openhab.addons", null);
             Dictionary<String, Object> properties = configuration.getProperties();
-            if (properties == null) {
-                // if we can't determine a set property, we use true (default is remote enabled)
-                return true;
-            }
-            return ConfigParser.valueAsOrElse(properties.get(CONFIG_REMOTE_ENABLED), Boolean.class, true);
+            // if we can't determine a set property, we use true (default is remote enabled)
```
J-N-K, line 301: TBH, I don't think this is an improvement. Looking at the code before I can easily determine what it does, now I have to think about it.

PR #4563, opened 2025-01-18
Fix group calculations to use GroupItem's system unit

### LINKED ISSUE #4562: Improve consistency of calculation functions over sets of QuantityType values
> Indeed. And as mentioned above, it applies to ALL calculations based on sets of items with mixed units. So to be clear this is not really an issue with `toInvertibleUnit` .. it is rather an issue with the 'sum' algorithm when the set includes items having a non zero point offset.

That is not correct, as the calculation functions will be done on the absolute scale since https://github.com/openhab/openhab-core/pull/3792. So there is no unpredictability.

Looking at the code of the arithmetic group functions, it looks like items of another dimension than the first are excluded from the sum. So depending on the unit of the first item, you will have a different set of items used in the function if there are invertible units in it. It would probably be better to use the unit of the group base item, instead of the unit of the first item, and then allow the reciprocal unit in the calculation after conversion to the base unit. I notice in the history the use of toInvertibleUnit in that method has explicitly been added to cope with color temperature, but in my view it does not work at all at this time (and that is from reading code, not testing). Also it is not used in all methods, only a few.

_Originally posted by @mherwege in https://github.com/openhab/openhab-core/issues/4561#issuecomment-2599732127_

andrewfg:
Fixes #4562 

There was a bug whereby the calculations over a GroupItem's set of member Items are based on the Unit of the first member in the set. There are two problems: a) the order of the Items in the set is not fixed, and b) the Unit of the first member Item is also not definitive to the Unit of the expected result. Therefore these calculations do produce variable and unexpected results.

This PR fixes the bug by normalising the calculations over the group's set of member Items so that all calculations are based on the System Unit of the GroupItem's base Item.

Furthermore this PR allows conversion between non- inverted and inverted Units so invertible conversions (e.g. Kelvin <=> Mirek) are supported.


andrewfg:
@mherwege for info..

andrewfg:
TODO I need to add to the JUnit tests to confirm the functionality when the unit is taken from the group main item rather than from the group member items.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunction.java
```diff
@@ -86,23 +97,27 @@ public State calculate(@Nullable Set<Item> items) {
...
-                            sum = itemState; // initialise the sum from the first item
-                            count++;
-                        } else {
-                            itemState = itemState.toInvertibleUnit(sum.getUnit());
-                            if (itemState != null) {
-                                sum = sum.add(itemState);
-                                count++;
-                            }
-                        }
-                    }
```
mherwege, line 105: As long as this check remains, none of the conversions to base unit are necessary. And you can use toUnit instead of toInvertibleUnit, which I would prefer.

If you want this to work for inverse units as well, you need to check for same dimension and inverse dimension. And that’s where it gets tricky. You may have to convert the item unit to its systemUnit (see discussion in other PR) and then the inverse to see if that is in the same dimension.

andrewfg: I think we can simply continue if toInvertibleUnit returns null. (It does its own checks and falls through to toUnit anyway..)

andrewfg:
@mherwege while testing this, I discovered that the JUnit assertion test values are physically WRONG. e.g. the sum of 23.54 °C plus 89 °C plus 122.41 °C is absolutely NOT 234.95 °C .. rather it is 781.24 °C .. this assumes that the three absolute Kelvin values must be summed. => So this really provokes the question about what we are really trying to achieve here? Do we want a) just the sum of three numbers, or b) the sum of three physical temperatures?

```
       items.add(createNumberItem("TestItem1", Temperature.class, new QuantityType<>("23.54 °C")));
        items.add(createNumberItem("TestItem2", Temperature.class, UnDefType.NULL));
        items.add(createNumberItem("TestItem3", Temperature.class, new QuantityType<>("89 °C")));
        items.add(createNumberItem("TestItem4", Temperature.class, UnDefType.UNDEF));
        items.add(createNumberItem("TestItem5", Temperature.class, new QuantityType<>("122.41 °C")));

        GroupFunction function = new QuantityTypeArithmeticGroupFunction.Sum(Temperature.class, null);
        State state = function.calculate(items);

        assertEquals(new QuantityType<>("234.95 °C"), state);
```

mherwege:
> @mherwege while testing this, I discovered that the JUnit assertion test values are physically WRONG. e.g. the sum of 23.54 °C plus 89 °C plus 122.41 °C is absolutely NOT 234.95 °C .. rather it is 781.24 °C .. this assumes that the three absolute Kelvin values must be summed. => So this really provokes the question about what we are really trying to achieve here? Do we want a) just the sum of three numbers, or b) the sum of three physical temperatures?

Look at the PR I already linked several times. This is on purpose. The second and third arguments are interpreted ad differential values, not absolute values. And that’s what makes sense, as most would expect that adding 20 °C to 20 °C gives 40°C. And with the way it is done, it does. It also works for Fahrenheit, which is not the same scale as Kelvin or Celsius. This problem is only there if the 0 point of the respective units is not the same.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunction.java
mherwege: Yes, we can, but I don’t like it. If it calls this it kind of indicates this will also work with inverted units. And it does not because of the earlier check. Or we fix the checks and also allow inverse units, or we just call toUnit so it doesn’t create wrong expectations.

andrewfg:
> The .. arguments are interpreted as differential values

Ok. Got it.

I just committed support for the following..
- use the units of the group base item when available (in particular applies to `sum` and probably `median` functions
- added support for inverse units (in particular applies to color temperatures)
- added unit tests for the above.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunction.java
andrewfg: Done

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunction.java
```diff
@@ -62,11 +65,26 @@ public State[] getParameters() {
...
-                return isSameDimension(groupItem.getBaseItem());
+        protected boolean isCompatible(@Nullable Item member) {
+            if (member instanceof GroupItem groupItem) {
+                return isCompatible(groupItem.getBaseItem());
+            }
+            if ((baseItem instanceof NumberItem baseNI) && (member instanceof NumberItem memberNI)) {
+                Unit<? extends Quantity<?>> baseUnit = baseNI.getUnit();
+                Unit<? extends Quantity<?>> memberUnit = memberNI.getUnit();
+                if (baseUnit != null && memberUnit != null) {
+                    return baseUnit.isCompatible(memberUnit) || baseUnit.isCompatible(memberUnit.inverse());
```
mherwege, line 76:
```suggestion
                    return baseUnit.isCompatible(memberUnit) || baseUnit.isCompatible(memberUnit.getBaseUnit().inverse());
```

We discussed this previously. I think we should always use the base unit for inversion, which works in all cases we have identified so far.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunction.java
```diff
@@ -42,9 +43,11 @@ public interface QuantityTypeArithmeticGroupFunction extends GroupFunction {
     abstract class DimensionalGroupFunction implements GroupFunction {
 
         protected final Class<? extends Quantity<?>> dimension;
+        protected final @Nullable Item baseItem;
 
-        public DimensionalGroupFunction(Class<? extends Quantity<?>> dimension) {
+        public DimensionalGroupFunction(Class<? extends Quantity<?>> dimension, @Nullable Item baseItem) {
```
mherwege, line 48: I think `baseItem` shouldn't be `Nullable`. That further simplies the logic as well.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunction.java
andrewfg: I guess you meant `getSystemUnit()` here?

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunction.java
andrewfg: I kept it Nullable because the JUnit tests work on that that case (which causes it to default to the first member Item). However as far as I can tell, there is no instance in actual OH run-time where a null `GroupItem::baseItem` would ever occur. So I am happy to refactor this, but note that it would require modifying existing JUnit tests, which I was reluctant to do in order to avoid self delusion. => @mherwege are you Ok with such change?

PS if we would do this, then possibly the 'dimension' argument can also be removed from the constructor. ??

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunction.java
andrewfg: PS Insofar as all the JUnit tests succeeded I am wondering if the change is necessary? Or should we add another test for that case?

bundles/org.openhab.core/src/test/java/org/openhab/core/internal/i18n/TestKelvinProvider.java
```diff
@@ -0,0 +1,50 @@
...
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.internal.i18n;
```
mherwege, line 13: In my view, this has nothing to do with `i8n`. I understand you put the class there because there already is `TestUnitProvider`.  But I would keep it with the test you are doing.

bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunctionTest.java
```diff
@@ -380,4 +385,160 @@ private GroupItem createGroupItem(String name, Class<? extends Quantity<?>> dime
...
+        items.add(createNumberItem("TestItem2", Temperature.class, UnDefType.NULL));
+        items.add(createNumberItem("TestItem3", Temperature.class, new QuantityType<>("113 °F")));
+        items.add(createNumberItem("TestItem4", Temperature.class, UnDefType.UNDEF));
+        items.add(createNumberItem("TestItem5", Temperature.class, new QuantityType<>("294.15 K")));
+
+        Item baseItem = new NumberItem("Number:Temperature", "BaseItem", kelvinProvider);
+        GroupFunction function = new QuantityTypeArithmeticGroupFunction.Avg(Temperature.class, baseItem);
+        State state = function.calculate(items);
+
+        assertEquals(new QuantityType<>("55.33333333333333333333333333333334 °C"), state);
```
mherwege, line 424: This is prone to machine rounding differences when running the test. Try converting the state to °C, take the double value and use the assertEquals method with an extra argument (the tolerance).

bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunctionTest.java
```diff
@@ -380,4 +385,160 @@ private GroupItem createGroupItem(String name, Class<? extends Quantity<?>> dime
...
+
+    @ParameterizedTest
+    @MethodSource("locales")
+    public void testSumFunctionColorTemperatureDifferentUnitsWithBaseItem(Locale locale) {
+        Locale.setDefault(locale);
+
+        Set<Item> items = new LinkedHashSet<>();
+        items.add(createNumberItem("TestItem1", Temperature.class, QuantityType.valueOf(2000, Units.KELVIN)));
+        items.add(createNumberItem("TestItem2", Temperature.class, UnDefType.NULL));
+        items.add(createNumberItem("TestItem3", Temperature.class, QuantityType.valueOf(1726.85, SIUnits.CELSIUS)));
```
mherwege, line 473: Is color temperature really ever expressed in °C or °F? I know it now works, but I would just limit it to Kelvin and Mired.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunction.java
mherwege: Yes, I would refactor this. And yes, you could remove the dimension argument as well.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunction.java
mherwege: > I guess you meant `getSystemUnit()` here?

Yes, that's what I meant.


> PS Insofar as all the JUnit tests succeeded I am wondering if the change is necessary? Or should we add another test for that case?

Maybe try checking if `mK` (milliKelvin) `toInvertibleUnit` and then `toUnit(Mired)` works. I suspect that may give strange results as the inverse of milliKelvin is not in the same scale as Mired.

andrewfg: milliKelvin is wrong for two reasons .. a) the factor would be micro rather than milli, and b) the conversions are `mirek = 1e6 / kelvin` resp. `kelvin = 1e6 / mired` -- i.e. not mirek = `microKelvin` ..

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunction.java
andrewfg: Done

bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunctionTest.java
andrewfg: Done

bundles/org.openhab.core/src/test/java/org/openhab/core/internal/i18n/TestKelvinProvider.java
andrewfg: Done

bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunctionTest.java
andrewfg: I would rather keep the test. I know it should not be commonly used. But it addresses the common case where a user creates a plain vanilla `Number:Temperature` Item without specifying a unit. By default Number:Temperature items have unit=Fahrenheit in USA and unit=Celsius everywhere else. So we need to test that this won't fall over..

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunction.java
andrewfg: My prior JUnit tests were testing `toInvertibleUnit` with target Kelvin and they all pass. However I added a test that tests `toInvertibleUnit` with target Mirek `toInvertibleUnit` with the inverse target Mirek. This test currently fails. However it should work fine once #4561 is merged.

andrewfg: Maybe also keep in mind that `isCompatible()` works both ways even without falling back to `systemUnit`. The issue was that 'toInvertibleUnit` itself had a bug that requires an intermediate conversion via `systemUnit`.

So there is nothing to fix at this point in the code.

And #4561 provides the intermediate conversion via `systemUnit`.

bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunctionTest.java
mherwege: Yes, I understand. I am still worried that fixing that issue actually makes things worse. Right now the conversion from °C to mirek fails in the conversion as it correctly says it is not the same scale. With the changes, it will succeed. But looking at all of the discussion in the community post, that same user also assumes the state description min and max are in Kelvin when he doesn’t set the unit. Without the error, he would not have known.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunction.java
mherwege: My intention was to not start from the system unit (K), but from a derived unit (like mK or whatever multiplier symbol). But you are right with the fix in toInvertibleUnit, it should be fine.

mherwege: > My prior JUnit tests were testing `toInvertibleUnit` with target Kelvin and they all pass. However I added a test that tests `toInvertibleUnit` with the inverse target Mirek. This test currently fails. However it should work fine once #4561 is merged.

I don’t get why that fix would make a difference though. Did you try running the test with the patch applied as well?

bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunctionTest.java
andrewfg: I think you are referring to #4432

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunction.java
andrewfg: Ok I will patch and test.

andrewfg: I just tried the tests with the #4561 code patch applied, and I can confirm that it works fine.

mherwege:
What about the following approach:

Create a new method in `QuantityType`, `addAbsolute`, similar to `add` (https://github.com/openhab/openhab-core/blob/ce374252fa2c821103da888302f930c1c127f73c/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityType.java#L585)

It should be something like:
```
    public QuantityType<T> addAbsolute(QuantityType<T> state) {
        Quantity<T> quantity = Quantities.getQuantity(this.quantity.getValue(), this.quantity.getUnit(),
                Scale.ABSOLUTE);
        Quantity<T> quantitytoAdd = Quantities.getQuantity(state.quantity.getValue(), state.quantity.getUnit(),
                Scale.ABSOLUTE);
        return new QuantityType<>(quantity.add(quantityToAdd));
    }
```

You then call this method instead of `add` in the group arithmetic functions.
This way, adding in the group will always use the absolute 0 reference and the result becomes independent of the units and sequence. This will also yield the right average.

This will then be different from when you simply do the calculation in a rule, because there only the first argument is treated as ABSOLUTE. But I don't think this kind of logic applies in groups anyhow.

andrewfg:
> You then call this method instead of add in the group arithmetic functions.

@mherwege please calm down. As you say add is a relative function. It is based on the zero point offset of the initial unit. So in the case of Celsius it is a relative add. And in the case of Kelvin (and most other zero based units) the add function is at the same time both relative and absolute.

mherwege:
> please calm down

Honestly, I am pretty calm on this. I am just trying to achieve the best possible result. That's why I keep on questioning things. If we don't do that now, I know we hit issues at some point in time and we will have to find other workarounds again.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunction.java
```diff
@@ -218,12 +214,11 @@ public State calculate(@Nullable Set<Item> items) {
...
                     continue;
                 }
-                QuantityType itemState = item.getStateAs(QuantityType.class);
+                QuantityType itemState = normalizedQuantityType(item.getStateAs(QuantityType.class));
                 if (itemState == null) {
                     continue;
                 }
-                itemState = itemState.toInvertibleUnit(targetUnit);
-                if (itemState != null && (min == null || min.compareTo(itemState) > 0)) {
+                if (min == null || min.compareTo(itemState) > 0) {
```
mherwege, line 221: These comparison operators are the wrong way around. But that is because the compareTo method is wrong in the first place. See https://github.com/openhab/openhab-core/pull/4571

andrewfg:
> my worry is the different results depending on the unit used

We just need to add the following text to the readme and JavaDoc; then everything clear :)

_The `sum` resp. `add` function summates the series of values relative to the zero point offset of the target unit. Most units have an offset of zero in which case the result is an absolute sum of the values. However some units (e.g. Fahrenheit, Celsius) have a non zero offset; in these cases the result is a sum relative to the zero point e.g. 20 C + 30 C => 50 C_

mherwege:
> `add` function

So are you proposing to change the current logic of the add function? Because that is not what the add method in QuantityType currently does. Remember this method is used in other places than the group arithmetic functions.

mherwege:
> The `sum` resp. `add` function summates the series of values relative to the zero point offset of the target unit. Most units have an offset of zero in which case the result is an absolute sum of the values. However some units (e.g. Fahrenheit, Celsius) have a non zero offset; in these cases the result is a sum relative to the zero point e.g. 20 C + 30 C => 50 C

This is not what the current add method in QuantityType does.
Here is an example:
20 K add 20 °C will result in 40 K
Equally 20 °C add 20 K will result in 40 K

Within the group arithmetic functions, because you first convert all arguments to the base unit, you get (assuming base unit is K):
20 K + 20 °C = 20 K + 293.15 K = 313.15 K = 40 °C
Or if base unit is °C: -253.15 °C + 20 °C = -233.15 °C = 40 K
So the result becomes dependent on the base unit. I don’t think it should. So the only way around is to always use absolute values.

andrewfg:
> Equally 20 °C add 20 K will result in 40 K

Are you sure? I would expect a result of 40 °C ???

andrewfg:
@mherwege I have a doubt..

This PR has a JUnit test suite that runs 75 tests. It was about 57 tests in the prior version but I added 18 more. The initial 57 tests all pass with the assertion results as the prior version did. (So this PR did not break anything). And the 18 new tests are to check the new functionality. I checked the 18 new assertion values by hand calculator.

Therefore my doubt is whether your fears are justified.

If you have concerns about this PR can you therefore please submit your own proposed JUnit test cases so that I can write code that would pass your tests. Otherwise we are talking around in theoretical circles.

andrewfg:
See the following unit tests. They all produce IMHO the correct results. Or do you expect them to produce others??.
- [Sum Celsius Base](https://github.com/openhab/openhab-core/blob/203c4e2f56da4ba5e69a1d2e523bdb4d1bed2e96/bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunctionTest.java#L99)
- [Sum Kelvin Base](https://github.com/openhab/openhab-core/blob/203c4e2f56da4ba5e69a1d2e523bdb4d1bed2e96/bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunctionTest.java#L391)
- [Sum Kelvin Base with Mirek Input](https://github.com/openhab/openhab-core/blob/203c4e2f56da4ba5e69a1d2e523bdb4d1bed2e96/bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunctionTest.java#L483)
- [Sum Mirek Base with Mirek Input](https://github.com/openhab/openhab-core/blob/203c4e2f56da4ba5e69a1d2e523bdb4d1bed2e96/bundles/org.openhab.core/src/test/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunctionTest.java#L502)

mherwege:
@andrewfg So, I ran the simulations and here is a little table. The original unit of the inputs are on top, but the values in the cells are already recalculated to the target unit. These correspond to the tests with some extensions.

  |   |   |   | °C | °F | K | MK-1
-- | -- | -- | -- | -- | -- | -- | --
1 | testSumFunctionQuantityTypeDifferentUnitsBaseKelvin | 1054.4 | K | 296.69 | 362.15 | 395.56 |  
2 | testAvgFunctionQuantityTypeDifferentUnitsBaseKelvin | 328.4833 | K | 373.15 | 318.15 | 294.15 |  
3a | testSumFunctionColorTemperatureDifferentUnitsBaseKelvin | 8000 | K | 2000 | 2000 | 2000 | 2000
3b |  | 6907.4 | °C | 1726.85 | 1726.85 | 1726.85 | 1726.85
4 | testAvgFunctionQuantityTypeColorTempDifferentUnitsBaseKelvin | 2000 | K | 2000 | 2000 | 2000 | 2000
5 | testAvgFunctionQuantityTypeColorTempDifferentUnitsBaseMirek | 500 | MK-1 | 500 | 500 | 500 | 500
  |   |   |   |   |   |   |  
6a | testSumFunctionQuantityTypeDifferentUnits | 234.95 | °C | 23.54 | 89 | 122.41 |  
6b |   | 1054.4 | K | 296.69 | 362.15 | 395.56 |  
7a | testAvgFunctionQuantityTypeDifferentUnits | 55.33333 | °C | 100 | 45 | 21 |  
7b |   | 328.4833 | K | 373.15 | 318.15 | 294.15 |  

My first question is the difference in outcome between 6a and 6b (is the same as 1). In the current propsal, the base unit determines the way the sum is calculated. Before, the first element in the group did determine it. I would argue that even made the whole summing quantity types in groups unpredictable, but that is another story. At least now it is predictable.
My view was to always use absolute values, so always generate result (1 or 6b).

To that end, I also created a scenario 3b where the base unit is °C and there is a mirek input. So the sum would be off when you try to visualize this with e.g. a Kelvin state description pattern. This is another case you would have when the °C default is not changed.

I am just not so sure anymore it is the right thing to do to use group base item unit as the reference unit. It does make the result dependent on that unit setting. And so far the unit never really had an impact on any calculation we did. If someone changes the unit of an item, it has a broader impact than just this calculation. It will also impact persistence. That's why it got introduced in the first place. So if you change this to make the group sum calculation work, it impacts that as well. That's why I was somewhat advocating to always use the value against an absolute reference (hence the addAbsolute proposal). In that case base unit will not have an impact anymore.

The solution here is also loosly related to how to treat it for the state filter profiles. There is no natural base unit there. Summing in absolute units would avoid that problem there as well.

I don't want to be the final decision maker on this. I am not a maintainer, so won't be the final judge. I can live with your proposed solution as long as it is properly documented.

andrewfg:
Thanks for the analysis. AFAICT this issue only affects the SUM calculation. The MIN MAX AVG and MEDIAN functions anyway produce the same results. @mherwege can you please confirm re MEDIAN? For SUM I think we have three possible ways forward:

1. SUM = relative sum based on GroupItem unit
2. SUM = absolute sum
3. Add two new GroupItem calculation functions SUMABS and SUMREL(*). Plus assign the existing SUM as alias for SUMREL to achieve largely the same results as before.

*) on the case of SUMREL the relative base would be the zero point of the GroupItem's Unit, rather than the Unit of the "first" item in the Group members set.

mherwege:
> The MIN MAX AVG and MEDIAN functions anyway produce the same results. @mherwege can you please confirm re MEDIAN?

Indeed, the challenge is only with SUM.

Following wiki is an interesting read: https://github.com/unitsofmeasurement/unit-api/wiki/Arithmetic-rules-for-Difference-versus-Absolute-quantities

This is actually the reference for all Quantity calculations and how they should be treated. I quote:

> The golden rule is:
> 
> The numerical values of all arithmetic operations must be calculated as is all values were converted to system unit before calculation. The conversion (not the arithmetic) depends on the MeasurementType. For example the conversion of 0°C can be 273.15 K or 0 K depending if the measurement type is ABSOLUTE or INCREMENTAL respectively.

The QuantityType.add method treats the second argument in the add as a INCREMENTAL, and the first argument as ABSOLUTE. So it results in:
5 °C + delta 5 °C = 10 °C = 283.15 K, but also
5° C + delta 5 K = 278.15 K + delta 5 K = 283.15 K = 10 °C
So wether you convert before or after the addition does not matter.
This add method makes perfect sense in rules, where the second argument would typically be a differential. It makes no sense in group sums. All arguments need to be treated the same way.
With the current logic for group sums, we are getting:
target unit °C: 5° C + 5 °C = 10 °C = 283.15 K
target unit K: 278.15 K + 278.15 K = 556.30 K
The second result complies with the golden rule where all values are ABSOLUTE. The first one does not. And looking at it as if all are INCREMENTAL from a different reference depending on the unit, to me is a stretch where I fail to see the practical value.

If we step back for a moment, what would summing temperatures in a group really be used for? Honestly, I can't think of cases where it makes sense. Average or median, min or max I believe make a lot of sense. I am not sure summing temperatures in a group does to be honest. And unless someone finds an example where it does make sense, I would stick to the golden rule, which is convert everything to system unit before calculating. This avoids the extra complexity of adding (documenting, explaining, maintaining) another version of SUM, just for the temperature case, where I see little relevance.

So my vote goes for doing the calculations in system unit and not creating an extra flavor of SUM.

andrewfg:
@mherwege I am good with your proposal above. But I thought it was you who was insisting that the calculations must be relative (20 C + 30 C = 50 C) so I am a little surprised that you seem to have changed your mind. ??

mherwege:
> @mherwege I am good with your proposal above. But I thought it was you who was insisting that the calculations must be relative (20 C + 30 C = 50 C) so I am a little surprised that you seem to have changed your mind. ??

Yes, and I was turning in circles on this. I have advocated the opposite in some of my messages as well. My appologies for that. My itch came from the fact that I went the extra mile to make rules work with first argument absolute and second argument differential (relative). And that's reflected in the add method. The moment the conversion is done beforehand, you actually make them relative to a new base, not the first argument anymore, as the conversion is always converting absolule values, not differential values. And then you treat the first argument as absolute and the second (converted) argument as relative again in the add method.
Anyway, after turning this around in my head so many times, my conclusion was that relative does not make sense in group calculations. So we should not attempt it, and use absolute in all cases by first converting to system unit.

I actually think a extra javadoc on the add method explaining this, stating that the first argument is treated as absolute and the second argument as differential would be appropriate. Do you mind putting that in? You are touching the same classes.

Also, this would be a breaking change. The sum may be different for temperatures than it was before. E.g. if all your group items where in °C, your sum would be one value absolute and all the rest added as delta. With my proposed change all will be summed in K. While I would advocate the previous result does not make sense, it is a breaking change.

andrewfg:
> an extra javadoc on the add method explaining this, stating that the first argument is treated as absolute and the second argument as differential would be appropriate

I will do that.

andrewfg:
@mherwege I made a commit just now that should hopefully reflect our final agreement.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunction.java
```diff
@@ -62,208 +61,148 @@ public State[] getParameters() {
...
+         * current {@link State} is not a {@link QuantityType}. Convert any remaining {@link QuantityType} to the
+         * 'referenceUnit' and exclude any values that did not convert.
+         *
+         * @param items a list of {@link Item}
+         * @return a list of {@link QuantityType} converted to the 'referenceUnit'
+         */
+        @SuppressWarnings({ "rawtypes" })
+        protected List<QuantityType> referenceUnitQuantityTypes(Set<Item> items) {
+            return items.stream().map(i -> i.getStateAs(QuantityType.class)).map(s -> referenceUnitQuantityType(s))
+                    .filter(Objects::nonNull).map(s -> (QuantityType) s).toList();
```
mherwege, line 90:
```suggestion
    protected List<QuantityType<?>> referenceUnitQuantityTypes(Set<Item> items) {
        return items.stream().map(i -> i.getState()).map(s -> referenceUnitQuantityType(s))
                .filter(Objects::nonNull).collect(Collectors.toList());
    }
```
I think this solves the rawtypes problem. I still had it when using a straight `toList()`.

andrewfg: > I think this solves the rawtypes problem 

Umm. Yes it solves the problem on `toList()`. But the method still needs the annotation for another reason. Aargh!

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunction.java
```diff
@@ -62,208 +61,148 @@ public State[] getParameters() {
...
-                                sum = sum.add(itemState);
-                                count++;
-                            }
-                        }
-                    }
+            if (items != null) {
+                List<QuantityType> referenceUnitQuantities = referenceUnitQuantityTypes(items);
+                if (!referenceUnitQuantities.isEmpty()) {
+                    return referenceUnitQuantities.stream()
+                            .reduce(new QuantityType(0, referenceUnit), QuantityType::add)
```
mherwege, line 111:
```suggestion
                            .reduce(new QuantityType<>(0, referenceUnit), QuantityType::add)
```

andrewfg: Ok.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunction.java
```diff
@@ -62,208 +61,148 @@ public State[] getParameters() {
...
-                            itemState = itemState.toUnit(sum.getUnit());
-                            if (itemState != null) {
-                                sum = sum.add(itemState);
-                            }
-                        }
-                    }
+            if (items != null) {
+                List<QuantityType> referenceUnitQuantities = referenceUnitQuantityTypes(items);
+                if (!referenceUnitQuantities.isEmpty()) {
+                    return referenceUnitQuantities.stream().reduce(new QuantityType(0, referenceUnit),
```
mherwege, line 162:
```suggestion
                    return referenceUnitQuantities.stream().reduce(new QuantityType<>(0, referenceUnit),
```

andrewfg: Yup.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityTypeArithmeticGroupFunction.java
mherwege: Yes, figured it out as well. That's where I find streams combined with generics are a huge challenge. See my updated comment.

andrewfg: > You can just have i.getState()

Good catch. Done.

I think the remaining issues boil down to the different generics signatures between `QuantityTypeArithmeticGroupFunction.java` (i.e. `QuantityType<T>`) and `QuantityType.java` (i.e. `QuantityType<?>` .. the former is what we get from the stream functions, and latter is what we need when we call `QuantityType.add()`..

I did not fully follow you PR discussion about `compareTo` yet .. but I saw that `QuantityType.compareTo(QuantityType<T> o)` has a different signature than `QuantityType.internalCompareTo(QuantityType<?> o)` .. so I wonder if we could do the same (actually the opposite) here? And if it is actually necessary?

mherwege (approved):
LGTM

andrewfg:
@mherwege thanks for approving the PR. Unfortunately I had to revert your proposal from `.collect(Collectors.toList())` back to `.map(s -> (QuantityType) s).toList()` .. the reason is that whilst the former would compile locally in Eclipse IDE it failed the Maven build. The reverted version passes both Eclipse and Maven.

kaikreuzer (approved):
Many thanks, @andrewfg and @mherwege!

PR #4564, opened 2025-01-19
Set addons.xml as master addon info

jimtng:
This would prioritize the name and descriptions from addons.xml
which seem to be more complete.

Without this, many add-on data returned by the REST service /addons are missing their description.

The description will be used in add-on search and to show a short description for the binding in the UI

See
https://github.com/openhab/openhab-webui/issues/2870
https://github.com/openhab/openhab-webui/pull/3028


jimtng:
It seems that some addons (e.g. mielecloud) has `<name>@text/addon.ADDONID.name</name>` and the same with their description tag inside their addon.xml.

PR #4567, opened 2025-01-19
Reduce SAT warnings

holgerfriedrich:
Replaces #4555.
Address PMD warnings.
Cleanup PMD suppressions, add local // NOPMD tags where readability is preferred over PMD rule.

After @J-N-K s [comment](https://github.com/openhab/openhab-core/pull/4555#discussion_r1921530556) about code restructuring to avoid [SimplifyBooleanReturn](https://docs.pmd-code.org/pmd-doc-7.8.0/pmd_rules_java_design.html#simplifybooleanreturns), the related changes proposed in #4555 are no longer intended.
I agree that complying to this warning leads to code that is harder to read and understand. I already had ignored some of those warnings to stick with better readability when I created #4555. But maybe it is better not to go for those changes at all.

I could bring the rule set to core and do this change already now, but I think it does not make sense to duplicate the ruleset.
So I will propose to remove [SimplifyBooleanReturn](https://docs.pmd-code.org/pmd-doc-7.8.0/pmd_rules_java_design.html#simplifybooleanreturns) warning from the standard rule set in next SAT release.


kaikreuzer (approved):
Thanks!

PR #4568, opened 2025-01-19
Remove RegistryHook

holgerfriedrich:
Remove `RegistryHook` which is not longer used. Remove related methods from `ItemRegistry` and `ItemUiRegistry`.

See https://github.com/openhab/openhab-core/pull/4486#discussion_r1917043441 and https://github.com/openhab/openhab-core/pull/4486#discussion_r1921521767.

This is a public API, so we might go for deprecation first. With the target OH5, I have the feeling, we could skip this and remove it without deprecating it one release before.


wborn (approved):
Thanks!

PR #4570, opened 2025-01-20
Update logger to provide history and improve performance

cdjackson:
# Improve performance of logging and support reading history

- Reads the history so that the UI can populate logs on startup
- Manages high rate logging by sending logs in an array

# Description

This supports sending of history, and provides a method of managing message cadence.

When a client connects, it must send a filter request before the server will send any logs. This triggers the sending of history.

Live logs are sent as individual messages if they are received with sufficient spacing. When logs come in very quickly, they are clustered together and sent as an array after (up to) 100mS. With these changes I can view full zigbee debug logging which results in over 100 logs per second. The UI retains full responsiveness at this logging rate.

In principle the history size is configured in the `org.apache.felix.log.maxSize` property, but so far I've been unable to change this from the default of 100 entries.

See also https://github.com/openhab/openhab-webui/pull/3030 which provides updates to the UI

### DIFF
--- a/bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/log/LogDTO.java
+++ b/bundles/org.openhab.core.io.websocket/src/main/java/org/openhab/core/io/websocket/log/LogDTO.java
@@ -15,20 +15,34 @@ package org.openhab.core.io.websocket.log;
 import java.util.Date;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
-import org.eclipse.jdt.annotation.Nullable;
-import org.openhab.core.io.websocket.event.EventDTO;
 import org.osgi.service.log.LogLevel;
 
 /**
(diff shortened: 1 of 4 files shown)


cdjackson:
looks like this cherry pick from the 4.3 branch (which I'm developing against so I can test it in my production system) didn't come across cleanly. I'll take a look at this tomorrow.

kaikreuzer (approved):
lgtm, thanks!

Nadahar:
> In principle the history size is configured in the `org.apache.felix.log.maxSize` property, but so far I've been unable to change this from the default of 100 entries.

Does this mean that the maximum number of "historic entries" is 100? Because, I can't get "history" to show, but 100 entries is "nothing" with the logging I have enabled, which might explain why I don't see it.

edit: I did some additional testing, and this looks like the explanation. Whenever I enter the log viewer, it starts with 100 entries, which goes back less than 1 minute in time.

PR #4573, opened 2025-01-21
Revert "Fix addon name localisation before installation (#3908)"

jimtng:
This reverts commit 708a954081499754e3d36adc28cf16b7ffec7ad7
but with some modifications

This PR will use the name and description fields from addons.xml and return that to MainUI via REST. This is a change from before #3908 where the name and description fields from addons.xml were overwritten with "n/a". 

This data will be used for:

- This PR extends the search for addons within the description field too: https://github.com/openhab/openhab-webui/pull/3028
- Future improvements to addons store listing that can include addon description along with the title.

The addons.xml will be cleaned up so that all name/description fields of all addons contain the actual readable text, not i18n reference tags in this PR: https://github.com/openhab/openhab-addons/pull/18139


jimtng:
@mherwege

mherwege (approved):
LGTM

jimtng:
It seems that this "NA" thing now is never going to be used, so I've just removed it completely.
Before, NA was used to override the name and description from the addons.xml (which used to contain undesired i18n keys), but now since we don't need to do that, NA has no other purpose.

kaikreuzer (approved):
Thanks @jimtng!

PR #4574, opened 2025-01-22
Fix managed transformations retrieval

Nadahar:
As discussed [here](https://community.openhab.org/t/marketplace-versioning-with-embedded-resource/161421/97), after installing transformations from the marketplace, retrieval of managed transformations from storage during OH startup fails with:
```
Couldn't deserialize value 'org.openhab.core.storage.json.internal.StorageEntry@168c2c9b'. Root cause is: org.openhab.core.transform.ManagedTransformationProvider$PersistedTransformation cannot be found by org.openhab.core.storage.json_5.0.0.202501210532
```

This PR fixes it according to my testing, the problem is that the correct classloader isn't available to `JsonStorage`.

I can't comply with the DCO as I don't submit real name and e-mail in public, but this should be small enough for the exception to apply. It was a one-liner until Spotless interfered.


J-N-K (approved):

J-N-K:
Small Patch exception

Nadahar:
@J-N-K If there is another `4.3.x` release, I would consider backporting this one, since as far as I know, this is the only thing preventing marketplace transformations from working with `4.3.x`.

PR #4575, opened 2025-01-24
Persistence extensions, calculate QuantityType sums in absolute values

mherwege:
Sum calculations in persistence extensions for QuantityType items where done, summing the values in the configured unit.

This creates a challenge for temperatures, where the result would be different depending on the configured unit. To keep the results consistent, the sums should be done in the system unit for the dimension, wich has an absolute zero.

This PR changes that.

The only impact would be for summing temperatures, whereby it is highly unlikely someone would be summing temperatures over a time series.

This relates to https://github.com/openhab/openhab-core/pull/4563 and would make the logic consistent between summing for groups and summing over persisted values.

This PR would also change the returned units for QuantityType variance and deviation calculations to the  logically meaningful unit.

### DIFF
--- a/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
+++ b/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
@@ -1799,7 +1799,8 @@ public class PersistenceExtensions {
             Iterator<HistoricItem> it = result.iterator();
 
             Item baseItem = item instanceof GroupItem groupItem ? groupItem.getBaseItem() : item;
-            Unit<?> unit = baseItem instanceof NumberItem numberItem ? numberItem.getUnit() : null;
+            Unit<?> baseUnit = baseItem instanceof NumberItem numberItem ? numberItem.getUnit() : null;
+            Unit<?> unit = baseUnit != null ? baseUnit.getSystemUnit() : null;
 
             BigDecimal sum = BigDecimal.ZERO;
             while (it.hasNext()) {
--- a/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
+++ b/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
@@ -1242,12 +1242,19 @@ public class PersistenceExtensions {
         State averageState = internalAverageBetween(item, begin, end, effectiveServiceId);
 
         if (averageState != null) {
-            DecimalType dt = averageState.as(DecimalType.class);
+            Item baseItem = item instanceof GroupItem groupItem ? groupItem.getBaseItem() : item;
+            Unit<?> baseUnit = baseItem instanceof NumberItem numberItem ? numberItem.getUnit() : null;
+            Unit<?> unit = baseUnit != null ? baseUnit.getSystemUnit() : null;
+
+            DecimalType dt;
+            if (unit != null && averageState instanceof QuantityType<?> averageQuantity) {
+                averageQuantity = averageQuantity.toUnit(unit);
+                dt = averageQuantity != null ? averageQuantity.as(DecimalType.class) : null;
+            } else {
+                dt = averageState.as(DecimalType.class);
+            }
             BigDecimal average = dt != null ? dt.toBigDecimal() : BigDecimal.ZERO, sum = BigDecimal.ZERO;
             int count = 0;
-
-            Item baseItem = item instanceof GroupItem groupItem ? groupItem.getBaseItem() : item;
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
```diff
@@ -1799,7 +1807,8 @@ private static void internalPersist(Item item, TimeSeries timeSeries, @Nullable
             Iterator<HistoricItem> it = result.iterator();
 
             Item baseItem = item instanceof GroupItem groupItem ? groupItem.getBaseItem() : item;
-            Unit<?> unit = baseItem instanceof NumberItem numberItem ? numberItem.getUnit() : null;
+            Unit<?> baseUnit = baseItem instanceof NumberItem numberItem ? numberItem.getUnit() : null;
+            Unit<?> unit = baseUnit != null ? baseUnit.getSystemUnit() : null;
 
             BigDecimal sum = BigDecimal.ZERO;
```
andrewfg, line 1813: You could do this in one line :)
```
Unit<?> unit = (baseItem instanceof NumberItem numberItem) && (numberItem.getUnit() instanceof Unit<?> numberItemUnit) ?  numberItemUnit.getSystemUnit() : null;
```

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
```diff
@@ -1402,7 +1409,8 @@ private static void internalPersist(Item item, TimeSeries timeSeries, @Nullable
                 BigDecimal deviation = dt.toBigDecimal().sqrt(MathContext.DECIMAL64);
                 Item baseItem = item instanceof GroupItem groupItem ? groupItem.getBaseItem() : item;
                 if (baseItem instanceof NumberItem numberItem) {
-                    Unit<?> unit = numberItem.getUnit();
+                    Unit<?> baseUnit = numberItem.getUnit();
+                    Unit<?> unit = baseUnit != null ? baseUnit.getSystemUnit() : null;
```
andrewfg, line 1413: See below..

bundles/org.openhab.core.persistence/src/test/java/org/openhab/core/persistence/extensions/PersistenceExtensionsTest.java
```diff
@@ -75,6 +75,8 @@ public class PersistenceExtensionsTest {
     public static final String TEST_GROUP_QUANTITY_NUMBER = "testGroupQuantityNumber";
     public static final String TEST_SWITCH = "testSwitch";
 
+    public static final double KELVIN_OFFSET = 273.15;
```
andrewfg, line 78: Minor point, but I would initialise this constant from `QuantityType.valueOf(0, SIUnits.CELSIUS).toUnit(Units.KELVIN).doubleValue();`

andrewfg (changes requested):
Some minor suggestions

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java
```diff
@@ -1242,13 +1242,20 @@ private static void internalPersist(Item item, TimeSeries timeSeries, @Nullable
         State averageState = internalAverageBetween(item, begin, end, effectiveServiceId);
 
         if (averageState != null) {
-            DecimalType dt = averageState.as(DecimalType.class);
+            Item baseItem = item instanceof GroupItem groupItem ? groupItem.getBaseItem() : item;
+            Unit<?> baseUnit = baseItem instanceof NumberItem numberItem ? numberItem.getUnit() : null;
+            Unit<?> unit = baseUnit != null ? baseUnit.getSystemUnit() : null;
```
andrewfg, line 1247: Ditto

bundles/org.openhab.core.persistence/src/test/java/org/openhab/core/persistence/extensions/PersistenceExtensionsTest.java
mherwege: Yes, correct. But that gives a series of potential null pointer warnings for `toUnit`. It can't be null, it polutes the warnings, and it is more code to avoid it. As this offset will never change, and this is slightly more performant anyway, I will keep it.

andrewfg: Not going to fight it .. but just to be pedantic (it is a quiet day) .. a static final constant hardly has a performance hit .. and Objects.requireNonNull resolves sich wrong null warnings .. but as I said not going to fight it. :)

andrewfg (approved):
LGTM

kaikreuzer (approved):
Thanks!

PR #4577, opened 2025-01-26
Add logging to the StartLevelService

joerg1985:
This PR will add a debug log line in case a start level is not reached, there are small changes to update the code to use the `SequencedMap` interface added in JDK21.

Signed-off-by: Jörg Sautter joerg.sautter@gmx.net

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/service/StartLevelService.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/service/StartLevelService.java
@@ -13,12 +13,14 @@
 package org.openhab.core.service;
 
 import java.util.AbstractMap;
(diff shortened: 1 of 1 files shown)


J-N-K:
Why do we need the `SequenceMap` here? We could omit the sorting in `parseConfig` and use a `TreeMap` there, then we could just iterate over the `entrySet` and still get a sorted collection when checking the startlevel.

joerg1985:
@J-N-K The implementation does rely on the order of the map, so is used the interface reflecting this.
You are right everything works without the `SequencedMap` interface, so this is just syntactic sugar.

kaikreuzer (approved):
lgtm, thanks!

PR #4406, opened 2024-10-05
Upgrade Karaf to 4.4.7, Xtext/Xtend 2.37.0

holgerfriedrich:
This PR upgrades to the recently released Karaf version 4.4.7 🎉 and lifts Xtext/Xtend to 2.37.0.
Additionally, it removes workaround #4526 which should no longer be necessary.

This consists of 3 commits, PLEASE DO NOT SQUASH during merge.

Fixes #4283
Fixes various CVEs in BouncyCastle  1.77, Jetty, etc.

Upgrade Karaf from 4.4.6 to 4.4.7
* Sync runtime dependencies with Karaf 4.4.7, most notably:
   * PaxWeb 8.0.30
   * Jetty 9.4.56.v20240826
   * BouncyCastle 1.78.1
   * CXF 3.6.4
   * Jackson 2.18.2
   * JNA 5.16.0
   * JAXB 2.3.9
   * commons-io 2.17.0
   * commons-lang3 3.17.0
   * XBean 4.26
   * ASM 9.7.1
   * PaxLogging 2.2.8
 * Resolve itest runbundles

Upgrade Xtext/Xtend to 2.37.0
* Upgrade Xtext/Xtend from 2.36.0 to 2.37.0, see release notes:
  https://eclipse.dev/Xtext/releasenotes.html#/releasenotes/2024/11/19/version-2-37-0
  https://eclipse.dev/Xtext/xtend/releasenotes.html#/releasenotes/2024/11/19/version-2-37-0
* Upgrade dependencies
  * ecj from 3.36.0 to 3.39.0
  * gson from 2.10.1 to 2.11.0
  * classgraph to 4.8.176
  * guava from 3.33.0 to 3.33.1

PR #4581, opened 2025-01-28
Fix version of fasterxml.woodstox

holgerfriedrich:
* Add wstx version matching cxf 3.x

This seems to be required for the demo app to resolve properly.
Transitive dependencies from karaf libs brings fasterxml.woodstox 7.x, whereas cxf 3.6.4 still requires fasterxml.woodstox 6.x.
Adding this resolves the issue with the demo app.

Relates to latest karaf upgrade, #4406.
Relates to openhab/openhab-distro#1721.


wborn (approved):
Thanks!

Issue #4589, opened 2025-02-09
[Rule Templates] Support loading rule templates from file in conf


rkoshak:
Writing and publishing rule templates is tedious, error-prone and overly complicated. The current process is:

1. write a rule that you want to become a template
2. click the code tab and copy the YAML
3. *manually* modify the YAML to add the rule template fields, add properties, and modify the code with the {{propname}} replacements
4. create a post on the marketplace forum category
5. try to install your rule template and fail (I'm probably the most experienced rule template writer and I've never gotten it right the first time)
6. edit the marketplace posting
7. repeat 5-6 as many times as necessary to get just the format of the posting and the YAML correct
8. create a rule from the template
9. test the rule
10. repeat 6-9 as many times as required to make sure the rule itself works after being created and the properties applied.

This is excessively tedious and steps 4-10 is all done in public view where someone who has enabled "show unpublished" could install it before it's ready.

If we could write and test the rule templates locally instead of requiring us to publish it to the marketplace the edit/test loop can be significantly reduced. It will also make it much easier to maintain the rule template in the long run. It also opens up the opportunity for users to create rule templates of their own to improve code reuse.

This should support both JSON and YAML with YAML having priority. Bonus points if it supports more than one template per file.

## Your Environment

* Version used: (e.g., openHAB and add-on versions)
* Environment name and version (e.g. Chrome 111, Java 17, Node.js 18.15, ...):
* Operating System and version (desktop or mobile, Windows 11, Raspbian Bullseye, ...):

rkoshak:
Solved with #4591.

PR #4591, opened 2025-02-10
Rule Template installation fixes

seime:
Rule Template installation fixes

### DIFF
--- a/bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/automation/MarketplaceRuleTemplateProvider.java
+++ b/bundles/org.openhab.core.addon.marketplace/src/main/java/org/openhab/core/addon/marketplace/internal/automation/MarketplaceRuleTemplateProvider.java
@@ -19,14 +19,20 @@ import java.nio.charset.StandardCharsets;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.Locale;
+import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
+import org.openhab.core.automation.Module;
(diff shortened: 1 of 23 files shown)


Nadahar:
Ready for review/discussion.

This is a collaboration between @seime and me, so I'll add a bit of information.

There has been support for file-based rule templates (`conf/automation/templates`) in OH for many years, but it's been broken (not sure for how long), so it's probably unknown to most that it even exists. This is particularly handy for those making the rule templates. Today they have to upload their "work in progress" to the marketplace, install it and then test it. Rinse and repeat.

The reason it was broken is that the path used in the "file watcher" was wrong, so it responded by looking for the changed file in a different location and thus always failed to pick up anything. While we fixed this, we also found that there was another problem with the "file watcher" handling: The use of calls that potentially cause file system access as a way to filter out files, which fails for `DELETE` operations, because the file in question no longer exists.

To explain, it was for example checked if the file has hidden, and hidden files were filtered out. On Linux file systems, that work just fine, because you can deduct whether it's hidden from the file name itself (does it start with a `.`?). But, on other file systems, like NTFS, "hiddenness" is instead decided by file attributes stored in the file system. When trying to check this for a deleted file, a "file not found" `IOException` will be thrown and further processing halted.

The last commit in this PR is therefore strictly unrelated to rule templates, it just fixes the same issue in other places in core where the same problem was discovered.

Once the file-based rule templates were up and working, it turned out that only JSON was supported. Rule templates are often scripts like JavaScript. It's very inconvenient to write a whole script on one line separated by `\n`, so YAML support was desired in addition (as was already supported in the marketplace). Finally, the "parsers" for file and marketplace rule templates were "unified", giving the marketplace the additional functionality of supporting multiple rule templates in one file/add-on by putting them inside a JSON or YAML array.

rkoshak:
Solves #4589 .

seime:
Closes #4589

kaikreuzer (approved):
Many thanks @seime and @Nadahar for fixing this issue, fixing other bugs along the way and adding support for YAML templates - that's awesome!

PR #4593, opened 2025-02-10
Some other non null annotations

clinique:
and small code enhancements in org.op.enhab.core


kaikreuzer:
@clinique Could you please check the failing test?
```
TEST org.openhab.core.internal.events.OSGiEventManagerOSGiTest#testValidationTopic() <<< ERROR: Cannot invoke "String.isEmpty()" because the return value of "org.openhab.core.events.Event.getTopic()" is null
java.lang.NullPointerException: Cannot invoke "String.isEmpty()" because the return value of "org.openhab.core.events.Event.getTopic()" is null
	at org.openhab.core.internal.events.OSGiEventPublisher.assertValidArgument(OSGiEventPublisher.java:84)
```

kaikreuzer (approved):

PR #4596, opened 2025-02-14
Call unregisterCommand instead of registerCommand

jimtng:
I just came across this whilst reading the code. I'm not sure why it was calling registerCommand here, but it seems wrong.


kaikreuzer (approved):
I agree 😄

PR #4598, opened 2025-02-15
Fix JavaDoc, SAT and depreacation warnings

holgerfriedrich:
* Fix JavaDoc build
* Address SAT warnings introduced recently
* Address some of the deprecation warnings seen during compilation


kaikreuzer (approved):
lgtm, thanks!

PR #4602, opened 2025-02-16
JsonAddonService: Fix warning on startup if no URL configured

florian-h05:
When starting openHAB, I always get the following warning:

```
14:44:47.082 [WARN ] [tplace.internal.json.JsonAddonService] - JSON Addon Service invalid URL:
```

We should avoid this warning, I don't see a problem there, and we should not confuse users with unnecessary warnings.


kaikreuzer (approved):
Thanks!

holgerfriedrich:
@florian-h05 if you find some time, could you pls check if this is still necessary with the latest fixes? 

(See reference above)

florian-h05:
Yes, this is still neccassary with the current main branch (commit 343470d98a4f9ff0f594350c05984f2267b1a1fd),

PR #4605, opened 2025-02-17
Don't log a warning when JsonAddonService URL is blank

jimtng:


mherwege:
Isn’t this the same: https://github.com/openhab/openhab-core/pull/4602?

jimtng:
> Isn’t this the same: #4602?

Yup, identical, lol. I didn't see that one.

PR #4608, opened 2025-02-17
Revert ConfigUtil::normalizeType and fix null annotations

holgerfriedrich:
Partly reverts 8e597bde76ef17a7a4d47b1a39b8e44060c43e36.

This restores the semantics of `ConfigUtil::normalizeType` that was in place before #4505 was merged.

This should fix the NPE which crashes the add-on build.

```
java.lang.NullPointerException: Cannot invoke "Object.getClass()" because "value" is null
	at org.openhab.core.config.core.ConfigUtil.normalizeType(ConfigUtil.java:197)
	at org.openhab.core.config.core.ConfigUtil.normalizeTypes(ConfigUtil.java:234)
```


jimtng:
I wonder if this was also the cause of #4602

holgerfriedrich:
@openhab/core-maintainers could you please check this one? It should fix the broken add-ons build.

And could you please trigger a core build after merge.

Thanks!

kaikreuzer (approved):
Many thanks!

PR #4609, opened 2025-02-18
Add a quicker build instruction in README

jimtng:
Credit for the build command to @holgerfriedrich 
https://community.openhab.org/t/potential-problem-with-eclipse-and-latest-core/162024/12


kaikreuzer (approved):
Makes sense, thanks.

Issue #3662, opened 2023-06-18
Use Horizontal Connecting Lines in Charts for Persistence Strategy `everyChange`


dilyanpalauzov:
An item maps to the power consumption of a pump.  I persist changes on the power with `strategy=everyChange`.  In a sitemap I insert

> Chart item=e3_power period=4h refresh=60000

Then this URL is called http://openhab/chart?items=e3_power&period=4h&theme=bright&t=1687084184925 which produces

![chart4h](https://github.com/openhab/openhab-core/assets/4992947/259ef004-c369-49ef-8112-ffc61d0c3c1c)

The problem is that from the persisted data dots are created and these dots are then connected using tilted lines.  E.g. if from 8:00 to 8:59:20 the consumption was zero, and from 8:59:20 to 9:00:00 the consumption was 1000, there is no stored information that at e.g. 8:50:00 the consumption was zero, because the persistence strategy is `everyChange`.  The graph shall nevertheless make evident, that at 8:50:00 the consumption was zero.

Please make it possible to create charts, where the lines connecting the individual dots are only horizontal.  I understand that tilted lines have their merits to visualize interpolated data, but in the concrete use case with everyChange persistence strategy the diagram is inaccurate.  When the tilted lines are very close to each other (the pump is turned on and off in short periods of time) it is hardly possible to read from the graph, whether the device was all the time on, or it was also off in the meantime.  This gets even more complicated if I take a longer period, like `12h`.

My suggestion is: when the persistence strategy for an item is everyChange use in the charts only horizontal lines (no tilted lines). With other persistence strategies keep the current logic.

The above is written having sitemaps in mind.  I have not tried yet all possible options in “Main UI”.  In any case, in “Main UI” shall also exist an option to use horizontal lines to connect the dots, if there is no such option yet.

I use OpenHAB 3.4.2

dilyanpalauzov:
Just showing dots for the present data, avoiding any connecting lines would also work, as I can see what is stored.

rkoshak:
First of all, charting has not knowledge about the persistence strategy. All it sees are points. 

The charting servlet that generates the charts for the sitemap Chart element is very basic and simple. I don't know how feasible it is to actually provide any sort of configuration like that. 

The standard approach to solve this issue is to add a periodic strategy in addition to the everyChange strategy. That will add points to the graph in between the changes and flatten out the chart.

For even more configurability there is Grafana (there are lots of tutorials on the forum).

[...]

PR #4610, opened 2025-02-18
Sitemap interpolation parameter for charts

mherwege:
Closes https://github.com/openhab/openhab-core/issues/3662

This introduces an extra parameter for sitemap charts, `interpolation`, that can take two possible values: `linear` or `step`. If not set, the current behaviour will be used (`step` for OnOff or OpenClosed and `linear` for all other values).

Also fixes an issue where the `yAxisDecimalPattern` was not considered for managed sitemaps.

### DIFF
--- a/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/chart/ChartProvider.java
+++ b/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/chart/ChartProvider.java
@@ -55,7 +55,8 @@ public interface ChartProvider {
      * @param groups The groups to display on the chart
      * @param dpi The DPI (dots per inch) value, can be <code>null</code>
      * @param interpolation The interpolation between two datapoint (<code>linear</code> or <code>step</code>),
-     *            <code>null</code> defaults to <code>linear</code>)
+     *            <code>null</code> defaults to <code>linear</code> for numeric values and to <code>step</code> for
+     *            binary values
      * @param legend Show the legend? If <code>null</code>, the ChartProvider should make his own decision.
      * @return BufferedImage object if the chart is rendered correctly,
      *         otherwise null.
@@ -86,7 +87,8 @@ public interface ChartProvider {
      * @param dpi The DPI (dots per inch) value, can be <code>null</code>
      * @param yAxisDecimalPattern The format pattern for the y-axis labels
      * @param interpolation The interpolation between two datapoint (<code>linear</code> or <code>step</code>),
-     *            <code>null</code> defaults to <code>linear</code>)
+     *            <code>null</code> defaults to <code>linear</code> for numeric values and to <code>step</code> for
+     *            binary values
      * @param legend Show the legend? If <code>null</code>, the ChartProvider should make his own decision.
      * @return BufferedImage object if the chart is rendered correctly,
      *         otherwise null.
--- a/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/chart/ChartProvider.java
+++ b/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/chart/ChartProvider.java
@@ -54,6 +54,8 @@ public interface ChartProvider {
      * @param items The items to display on the chart
      * @param groups The groups to display on the chart
      * @param dpi The DPI (dots per inch) value, can be <code>null</code>
+     * @param interpolation The interpolation between two datapoint (<code>linear</code> or <code>step</code>),
+     *            <code>null</code> defaults to <code>linear</code>)
      * @param legend Show the legend? If <code>null</code>, the ChartProvider should make his own decision.
      * @return BufferedImage object if the chart is rendered correctly,
      *         otherwise null.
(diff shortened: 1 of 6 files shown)


lolodomo (approved):
LGTM

mherwege:
> @mherwege : did you see that issue? [openhab/openhab-webui#3025 (comment)](https://github.com/openhab/openhab-webui/issues/3025#issuecomment-2600768846)

I saw it a while ago, but didn't see the issue in core code.

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/chart/ChartProvider.java
```diff
@@ -54,6 +54,8 @@ public interface ChartProvider {
      * @param items The items to display on the chart
      * @param groups The groups to display on the chart
      * @param dpi The DPI (dots per inch) value, can be <code>null</code>
+     * @param interpolation The interpolation between two datapoint (<code>linear</code> or <code>step</code>),
+     *            <code>null</code> defaults to <code>linear</code>)
```
holgerfriedrich, line 58: Does this contradict what you wrote in the issue description?

>  If not set, the current behaviour will be used (step for OnOff or OpenClosed and linear for all other values).

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/chart/ChartProvider.java
```diff
@@ -82,6 +85,8 @@ BufferedImage createChart(@Nullable String service, @Nullable String theme, Zone
      * @param groups The groups to display on the chart
      * @param dpi The DPI (dots per inch) value, can be <code>null</code>
      * @param yAxisDecimalPattern The format pattern for the y-axis labels
+     * @param interpolation The interpolation between two datapoint (<code>linear</code> or <code>step</code>),
+     *            <code>null</code> defaults to <code>linear</code>)
```
holgerfriedrich, line 89: same question as above

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
```diff
@@ -220,6 +220,10 @@ protected Sitemap buildSitemap(RootUIComponent rootComponent) {
                 setWidgetPropertyFromComponentConfig(widget, component, "legend", SitemapPackage.CHART__LEGEND);
                 setWidgetPropertyFromComponentConfig(widget, component, "forceAsItem",
                         SitemapPackage.CHART__FORCE_AS_ITEM);
+                setWidgetPropertyFromComponentConfig(widget, component, "yAxisDecimalPattern",
```
holgerfriedrich, line 223: Can you comment on this? It seems missing before, why do we need it now?
(I am not familiar with the xchart properties)

holgerfriedrich (commented):
Overall: LGTM.
Just a few minor questions.

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
mherwege: This solves a bug I discovered when working on this, see my PR description:

> Also fixes an issue where the `yAxisDecimalPattern` was not considered for managed sitemaps.

Apparently, this parameter was missed when creating the managed sitemap provider. The UI has support for it, but it never worked. I could remove this and put it in a separate PR, but it is only a one line fix.

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/chart/ChartProvider.java
mherwege: Yes, indeed. I clarified it.

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/chart/ChartProvider.java
mherwege: Indeed.

mherwege:
Sigh, Eclipse insists on removing an explicit typecast. Without it, the build fails.

bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java
holgerfriedrich: fine for me, thanks for the clarification

holgerfriedrich (approved):
LGTM

schup011:
@mherwege 

I have now installed 5.0.0.0 M4 and this was a feature I had been waiting for.

Sorry to say that it does not seem to work in my case - or I did not really understand the concept. I thought

`interpolation="step"`

would really create steps until the next data point. But this does not look like it:

chart

I am ready to test or debug - could you give me some instructions?

In the log no complains about the new option. There was one when I put interpolation=step without quotes. Corresponding sitemap entry is:

`Chart item=Multisplit4_Power period=h   service="influxdb" legend=true refresh=60000 interpolation="step" visibility=[Periode==0]`

PR #4615, opened 2025-02-22
Semantic Tags: validate POINT/PROPERTY tags on Channel Types

andrewfg:
Addon developers may (or may not) set semantic tags in the `channel-type.xml` of their addons. The purpose of such tags is to provide information to Main UI concerning the POINT/PROPERTY semantic tags of the respective Channels. Currently the `channel-type.xml` allows developers to enter any free form string values. There is no schema check to confirm that the tags are actually real allowed POINT/PROPERTY tags instead of just random strings.

This PR adds XSD schema checks for correct allowed values of semantic tags on channel type definitions. The schema checks inputs against the current list of allowed POINT/PROPERTY tags. There is a separate PR #4626 that uses a script to update the actual list of allowed POINT/PROPERTY tags in the schema XSD whenever tags are added to (or eventually removed from) the set of OH Core standard tags.


jimtng:
If we want to be even more precise, only allow one Point, and one Property. This would avoid having two Points / two properties.

furthermore, perhaps Property can only be added with a Point, not alone.

jimtng:
The list of tags here can be updated by [generateTagClasses.groovy](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.semantics/model/generateTagClasses.groovy)

andrewfg:
> If we want to be even more precise, only allow one Point, and one Property. This would avoid having two Points / two properties.

Fully agreed. I just can't figure our the XSD syntax to specify that. Problem is that one could split the allowed values into two separate enums, but it is not allowed to have two xml elements of the same name 'tag' each being differently linked to those different enums. (If there are any XSD gurus please speak up).

andrewfg:
Please see #4617 also.

jimtng:
Maybe this can help? https://stackoverflow.com/questions/22354787/xsd-two-elements-with-the-same-name-but-different-attribute-value

jimtng:
Also another idea... probably crazy, we can probably write a validation script before build time.

andrewfg:
> Maybe this can help?

Uh. No. The accepted answer is as follows; and in our case we can neither give the elements different names, nor put them in different contexts. Unless we want to completely re-engineer the schema (which I do not not..)

> You can't declare two elements with the same name with different types in the same context, but I think I understand what you want to do.

> If you really had elements with very different contents, it would make sense to create two types (and it would also make sense for them to have different names or to at least occur in another context).

andrewfg:
> Also another idea... probably crazy, we can probably write a validation script before build time.

Not so crazy. I remember we did something similar with the addon suggestion finder code. AFAIR someone wrote a script that pools all the XML from all of the individual `addon.xml` files (all addons in the OH main distro) into a single `addons.xml` file, and I wrote a component that parses that file and logs anything bad. I think it would probably be trivial to extend that mechanism.

andrewfg:
Note: this PR causes a circular self referential CI build error. It is a change to thing-type XSD schema. The base file for the schema is located in `org.openhab.core.thing` and that will eventually be uploaded to `https://www.openhab.org/schemas/thing-description-1.0.0.xsd` .. however the CI build references the prior version from that url and gets confused.

andrewfg:
**Note**: The actual tag values are still under discussion in https://github.com/openhab/openhab-core/pull/4616, https://github.com/openhab/openhab-core/pull/4617, https://github.com/openhab/openhab-core/issues/4619, https://github.com/openhab/openhab-core/pull/4622 .. so the XSD enumeration restriction in the `thing-types.xml` schema will need to be finally updated to match those PRs.

andrewfg:
> The generator script could also update this xsd file

Wow that's cool. But I may need your help to write that script, as ot is a language I have not yet studied.

jimtng:
> > The generator script could also update this xsd file
> 
> Wow that's cool. But I may need your help to write that script, as ot is a language I have not yet studied.

Sure, I'll be happy to.

jimtng:
Does the xsd file not supposed to have one blank line at the end?

andrewfg:
I guess it should have a return at the end. Albeit that the current version does not have one.

jimtng:
Weird, my build is complaining about it and running spotless:apply added the missing blank line

kaikreuzer (approved):
Thanks.
Did you already check, if the addons build will still work once we actually publish this updated xsd at https://openhab.org/schemas/thing-description-1.0.0.xsd?

andrewfg:
@kaikreuzer thanks for merging.

> Did you already check, if the addons build will still work

Honestly not yet. But I expect there may be some failures that require fixes in some addons. I am not sure how to test this in a sandbox to prepare the necessary fixes. I think that maven caches the xsd somewhere so maybe I can figure it out..

andrewfg:
@kaikreuzer two things..

1. Eclipse: the eclipse compiler keeps a local cache version of the xsd file in the .m2 directory, and its IDE uses that to validate xml files. So you can see the validation against the new XSD in action in the screenshot below.

image

2. Maven: I have not yet figured out how to force Maven to validate against a local cached version of the XSD so I have not yet checked it for all 502 addons. => Any thoughts?

andrewfg:
@kaikreuzer OK. I did it manually using Eclipse to validate. There are 7 addons that fail. I will make a PR to cover them.

andrewfg:
@kaikreuzer / @jimtng we will need to merge the following before the whole automation is working, and before we can upload the new XSD: 

- #4695 to add some missing tags
- #4626 to inject the tags into the XSD file

kaikreuzer:
> OK. I did it manually using Eclipse to validate. There are 7 addons that fail.

Awesome, thanks!

The other two mentioned PRs are merged now as well.

PR #4616, opened 2025-02-23
Add `Airflow`, `Application`, `Channel`, and `Mode` Semantic Properties

jimtng:
Add new properties to support TV and Airconditioner points


jimtng:
@andrewfg fyi

lsiepel:
Please separate version upgrade from semantic tags.

jimtng:
I've split them up into separate commits

lsiepel:
> I've split them up into separate commits

All commits are pushed? As I still see the version change and the tag together

jlaur:
@jimtng - can you add `Airflow` to the PR title as well (for the release notes)?

jimtng:
> > I've split them up into separate commits
> 
> All commits are pushed? As I still see the version change and the tag together

Yes they are split up properly. There are 4 commits with airflow in a separate commit to the other 3 tags. Should I split each tag into a separate commit?

jimtng:
> @jimtng - can you add `Airflow` to the PR title as well (for the release notes)?

Title updated

rkoshak:
> Also wondering if adding these tags would need additional work in mainUI, @ghys ?

I can't definitively answer but given that custom tags are supported everywhere and I believe MainUI uses the REST API endpoint to get the lags there shouldn't be any wider changes required.

But I wonder if it makes sense to have a more deliberate discussion on what tags belong or not at a higher level and come up with a PR to address them all instead adding two here and four there in separate PRs without really designing the tags. 

For example, Battery, online status, pressure, VoC, Radon, air quality, are missing properties. There are a bunch of car related points and properties missing and car is missing from Equipment. 

On-the-other-hand, do we really need seven different types of Doors at the Equipment level and does it make sense to have smokealarm at all given the alarm/smoke point/property combo?

I worry that if we just add stuff here and there the lists will always remain inconsistent and  eventually grow beyond usability.

Note, I'm all for adding the four properties listed here. Even mode makkes sense: HVAC Equipment has a Control/Mode Point to set the heating/cooling/auto status.

andrewfg:
> I worry that if we just add stuff here and there the lists will always remain inconsistent and eventually grow beyond usability.

See #4615 in which we are seeking to apply an XSD schema restriction so that addon developers will be able to offer one “point” tag and one “property” tag; this offer would be a hint that users could choose to accept or override.

> Battery, online status, pressure, VoC, Radon, air quality, are missing properties.

@rkoshak may I therefore make a call to action on you to open a PR for those; or at least open an Issue..

rkoshak:
> @rkoshak may I therefore make a call to action on you to open a PR for those; or at least open an Issue..

That misses my whole point. 

But I was already creating a PR. https://github.com/openhab/openhab-core/issues/4619

lolodomo:
> Also upgrade groovy version to work with Java21
> Adjust copyright header

Please create a separate PR for that to have it merged quickly even if your proposal of new properties is finally not validated.

jimtng:
> Please create a separate PR for that to have it merged quickly even if your proposal of new properties is finally not validated.

#4621

lolodomo:
Should be rebased now that #4621 is merged.

jimtng:
rebased

lsiepel:
> I've split them up into separate commits

Bit late, but I meant PR’s not commits.

lsiepel (commented):
Changes look good. If core maintainers agree these tags should be added it is ready to be merged.

Only question i have (not a native english speaker) is the `Channel` tag specific enough? Do we agree it is fine to represent a TV channel, Radio channel or I/O Channel (and maybe there are more)

jimtng:
Channel should be applicable to TV Channel, and Radio Channel, or any other Channel. If they require a more specific tag, maybe that can be delegated to custom tags, or up for discussion in #4619 for future additions, and they could become a sub-tag of the Channel tag.

lsiepel:
Ping @openhab/core-maintainers can you look at this?

bundles/org.openhab.core.semantics/model/SemanticTags.csv
```diff
@@ -54,6 +55,9 @@ Property,Voltage,,Voltage,,
 Property,Current,,Current,,
 Property,Frequency,,Frequency,,
 Property,Gas,,Gas,,
+Property,Application,,Application,,
```
andrewfg, line 58: > Can you please confirm that it satisfies the new rules for semantic tags?

@jimtng I decided to review this myself against my current draft of the rules. The one where I have the biggest difficulty is the proposed `Application` property. Because it sounds to me more like a POINT than a PROPERTY. However when I look at your actual PR I see that CONTROL + CHANNEL is obvious for commanding the TV channel. And the analog for commanding the App should be CONTROL + APP .. rather than CONTROL + APPLICATION. In other words I think APP is obvious to everybody. Whereas APPLICATION is not so obvious. => Therefore I suggest changing APPLICATION to APP.

andrewfg (changes requested):
Suggested change.

bundles/org.openhab.core.semantics/model/SemanticTags.csv
jimtng: Isn't `App` merely an abbreviation of `Application`?
To me, `Application` sounds nicer and more obvious than `App`, although when you're talking about "Phone Apps" the abbreviated form is more common.

It used to be that "Desktop Applications" are also common, but lately "Desktop Apps" are also acceptable although more informal.

I don't quite understand your point (no pun intended!) about it sounding more like a POINT than a PROPERTY.

I _think_ it's a property, as you said Control (Point) + Application (Property) is what I think it should be. You are controlling the "Application" property of the device, similar to controlling the Volume property or Power property.

For example, on a physical remote control of my TV, I can press +/- to control the volume, press the red button to control its power, and there's a Netflix / Disney / Prime to control which application the TV should start.

It's all too abstract for me, that's why I haven't been getting involved too much in that documentation discussion.

andrewfg: For me the word “application” means an area of technical activity like lighting, heating, security, watering, etc. (i.e. POINT=APPLICATION + PROPERTY=LIGHTING). Therefore there is big scope for confusion over that term.

By contrast the word APP refers solely and explicitly to a piece of software running on a smart device. (i.e. POINT=CONTROL + PROPERTY=APP).

lsiepel: I think it is because `application` has multiple meanings:
image

While i personally think `Application` suits very good, i understand the case @jimtng points out. Maybe `App` is a better alternative as it does not seem to be multi interpretable :
image

andrewfg: Yes. I would like to avoid that developers may (wrongly) choose application as the POINT tag; and if it is named APP then there is no doubt that it should be the PROPERTY tag.

jimtng: Thanks for the clarifications. I see that `Application` has different meanings and therefore can be ambiguous / misinterpreted.

In regards to the change, should I add `Application` as a synonym, or should we completely drop it?

- Tag name: `App`
- Label: `App`
- Synonym: `Application` (?)
- Description: `Software program` (sounds weird?)

So, which of these should we use?
1. `Property,App,,App,,`
2. `Property,App,,App,Application,`
3. `Property,App,,App,Application,"Software program"`

andrewfg: IMHO probably 3. above.

andrewfg (approved):
LGTM

holgerfriedrich (approved):
LGTM, thanks!

PR #4617, opened 2025-02-23
Semantic Tags: set EQUIPMENT tags on Things

### LINKED ISSUE #4614: Add semantic equipment tag on things
Add support for `tag` element to thing type description xml.
In addition to, and analogous to, to the already existing tags for channel type descriptions.

The purpose is that thing definitions could define default 'Equipment' tags, in the same way as channel can define default 'Point'.

See https://github.com/openhab/openhab-docs/pull/2466/
See https://community.openhab.org/t/openhab-5-semantic-model-proposal/162526/83

#### Thing Type Definition (New)
```
<thing-type id="device">
    <supported-bridge-type-refs>
        <bridge-type-ref id="bridge-api2"/>
    </supported-bridge-type-refs>
    <label>Hue Device</label>
    <description>A Hue API v2 device with channels depending on its actual capabilities.</description>
    <semantic-equipment-tag>equipment.tag.name</semantic-equipment-tag>
    <channels>
        <channel id="illuminance" typeId="type-illuminance"/>
    </channels>
</thing-type>
```
#### Channel Type Definition (Already Existing)
```
<channel-type id="type-illuminance">
    <item-type>Number:Illuminance</item-type>
    <label>Illuminance</label>
    <description>Current illuminance.</description>
    <tags>
        <tag>point.tag.name</tag>
        <tag>property.tag.name</tag>
    </tags>
</channel-type>
```

andrewfg:
Addon developers may (or may not) set POINT/PROPERTY semantic tags in their addons. These tags are set either a) in the `channel-type.xml` of the addon, or b) dynamically at run-time via the addon Java code. The purpose of POINT/PROPERTY tags is for Main UI to offer appropriate tags on Items so that they will properly fit in the the user's hierarchical semantic model. Thus eliminating complexity for users.

Main UI also allows the user to apply EQUIPMENT tags when creating the user's hierarchical semantic model. However currently the user has to decide on, and apply, the EQUIPMENT tags manually themselves when setting up their system.

The purpose of this PR is to add a semantic EQUIPMENT tag to Things. This allows developers to set the EQUIPMENT tag on a Thing either a) in the `thing-type.xml` of the addon, or b) dynamically at run-time via the addon Java code. So that addons can offer the most appropriate EQUIPMENT tag via the Main UI. Thus further eliminating complexity for users.

This PR does the following:
1. It extends the XSD schema validation scheme to allow a semantic EQUIPMENT tag on Thing types; analogous to #4615
2. Adds code to permeate EQUIPMENT tags on Things throughout OH core wherever it may be needed. Specifically in the classes, DTOs and (de)serializers for thing types and things.

See also #4615
Resolves #4614
Used by https://github.com/openhab/openhab-webui/pull/3124


andrewfg:
@mherwege can I kindly ask you to be a sparring partner for this PR? My goal is to allow developers to set a (proposed) semantic (equipment) tag in their thing type xml, and have this permeated through the thing type instances to the actual thing instances so that the UI can use that semantic (equipment) tag (as a hint) in creating a semantic model.

mherwege:
> @mherwege can I kindly ask you to be a sparring partner for this PR? My goal is to allow developers to set a (proposed) semantic (equipment) tag in their thing type xml, and have this permeated through the thing type instances to the actual thing instances so that the UI can use that semantic (equipment) tag (as a hint) in creating a semantic model.

@andrewfg Sure. I like the idea as long as it does not then enforce anything on the item semantics. So I would think that it should be carried to the point of a having a tag on the thing, but then it should be up to the UI to pick this up as a suggested equipment semantic when creating the group item with its points (create model from thing).
As it should not be used on the thing directly, except when creating items, I am wondering what it should be called. Some day someone may think it carries meaning beyond a suggestion for an equipment group item, and may want to also show that in the UI...

bundles/org.openhab.core.config.discovery/src/main/java/org/openhab/core/config/discovery/DiscoveryResultBuilder.java
```diff
@@ -142,6 +143,17 @@ public DiscoveryResultBuilder withTTL(long ttl) {
         return this;
     }
 
+    /**
+     * Sets the semantic (equipment) tag.
+     *
+     * @param semanticEquipmentTag semantic (equipment) tag
+     * @return the updated builder
+     */
+    public DiscoveryResultBuilder withSemanticEquipmentTag(@Nullable String semanticEquipmentTag) {
```
jlaur, line 152: I'm wondering if it would be better to stay consistent and generic like e.g. here:
https://github.com/openhab/openhab-core/blob/6bf59605bdc25f57ce3792467075a09c567d349c/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/type/ChannelTypeBuilder.java#L54-L68

so simply `withTag`?

andrewfg: See https://github.com/openhab/openhab-core/pull/4617#issuecomment-2678795291 above.

mherwege: I am fine either way.

andrewfg: @jlaur some background..

In ChannelType we have a `tags` element that currently allows an unlimited number of `tag` sub-elements, which have no XSD schema validation. So this allows developers to have an open ended number of strings having open ended meanings.

I think it is the wish of OH maintainers to restrict the number and the meaning of these tags. So there should be a maximum of two `tag` sub-elements. And strictly speaking one of those `tag` elements is a `semantic point tag` and the other is a `semantic property tag`. And the developer choices when entering such values shall be subject to stricter XSD schema validation in order to restrict the choice to only valid tags.

```
// as it is today (open ended number and meaning)
<tags>
    <tag>blah</tag>
    ..
    <tag>blah</tag>
</tags>
```

In #4615 we are trying to do this constraint and validation. However since the `tag` elements have the same name we cannot do it fully via the xml. We would need to change the schema to do that (see below) which would be a breaking change. So I am loth to go all the way in doing that. But nevertheless we are moving towards more strictness rather than less.

```
// as it should be (constrained number and meaning)
<tags>
    <semanticPropertyTag>XSD validated Property</semanticPropertyTag>
    <semanticPointTag>XSD validated Point</semanticPointTag>
</tags>
```

Now -- finally to get to the point -- as we are moving towards more strictness the `ChannelType` `tag`s (`semantic property tag` resp. `semantic point tag`, we should not be moving to introduce a less strict `tag` in `ThingType`.

andrewfg:
Ok. I have now tested this, and I think it is ready for review.

image

Obviously it needs work now on the OH WebUI side in order to utilise this `semanticEquipmentTag` when creating the semantic model. => @mherwege :) 

Notes:
- The actual tag values are still under discussion in #4616, #4617, #4619, #4622 .. but this PR is primarily about implementation in things and DTOs.
- **Exception**: the XSD enumeration restriction in the `thing-types.xml` schema will need to be finally updated to match those PRs.

bundles/org.openhab.core.thing/schema/thing/thing-description-1.0.0.xsd
```diff
@@ -269,4 +270,65 @@
 			</xs:extension>
 		</xs:simpleContent>
 	</xs:complexType>
+
+	<xs:simpleType name="semanticEquipmentTag">
+		<xs:restriction base="xs:string">
+			<!-- Begin Allowed Semantic Equipment Tag Values -->
+			<!-- DO NOT EDIT THIS LIST - Generated by generateTagClasses.groovy -->
```
jimtng, line 277: This line shouldn't be in this PR. It will be added in #4626

mherwege:
> Obviously it needs work now on the OH WebUI side in order to utilise this `semanticEquipmentTag` when creating the semantic model. => @mherwege :)

I don't think this is a very complex change in the UI. I can take care of it, but would prefer to finish the core code first. There are a lot of moving parts and dependencies here, and creating a core that has it all for testing a UI change against is a bit of a challenge. It will be easier to do when we get the core changes over the line first.

andrewfg:
> It will be easier to do when we get the core changes over the line first.

@mherwege understood. IMHO all the changes in OH Core are ready to merge (via this PR). So it would be helpful if you can review this and give LGTM so that we can indeed proceed to merge it.

mherwege (approved):
LGTM

andrewfg:
Ping @ccutrer apropos our discussion on Equipment tags

kaikreuzer:
@andrewfg May I ask you to rebase this PR?

ccutrer (approved):

andrewfg:
> May I ask you to rebase this PR?

@kaikreuzer done! Apparently the CI / JavaDoc test failed but I don't know why.

holgerfriedrich:
@andrewfg the log of the javadoc job needs improvement. Running it locally, I see the message causing it to fail:

```
[WARNING] Javadoc Warnings
[WARNING] openhab-core\bundles\org.openhab.core.thing\src\main\java\org\openhab\core\thing\Thing.java:246: warning: @param argument "tag" is not a parameter name.
[WARNING] * @param tag the semantic (equipment) tag or {@code null} if no tag has been configured.
```

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/Thing.java
```diff
@@ -230,4 +231,19 @@ public interface Thing extends Identifiable<ThingUID> {
...
+     *
+     * @return the semantic (equipment) tag or {@code null} if no tag has been configured.
+     */
+    @Nullable
+    String getSemanticEquipmentTag();
+
+    /**
+     * Set the semantic (equipment) tag of the {@link Thing}.
+     *
+     * @param tag the semantic (equipment) tag or {@code null} if no tag has been configured.
```
holgerfriedrich, line 246:
```suggestion
     * @param semanticEquipmentTag the semantic (equipment) tag or {@code null} if no tag has been configured.
```

kaikreuzer (approved):
Thanks!

PR #4621, opened 2025-02-26
Upgrade Groovy to 3.0.23

jimtng:
This upgrade is needed so it would run on Java21.
Also change the generated header to `/*` instead of `/**`


wborn (approved):
Thanks!

Issue #4619, opened 2025-02-24
Scrub the semantic tags and add missing tags


rkoshak:
The default list of semantic tags are incomplete and inconsistent. Having a way to add custom tags helps some but the default list is the end user's first exposure to the model and it should help them get up to speed on how it works and support the most common use cases with the default list.

At a minimum we should have many more properties, a few more points, and more equipment. 

But there are also a bunch of tags that don't make much sense. For example, we probably only need two or three Door tags instead of seven. 

I think it makes a lot of sense to come up with some sort of rules for what makes a good tag or not, remove those that don't meet those rules and add any missing ones all in one go rather than fielding a bunch of separate issues to add two tags here and three tags there without any guiding principles. 

To deal with removed tags and backwards compatibility, the upgradeTool can search the tags and add any removed ones a user happens to have used as a custom tag. 

## Your Environment

* Version used: (e.g., openHAB and add-on versions)
* Environment name and version (e.g. Chrome 111, Java 17, Node.js 18.15, ...):
* Operating System and version (desktop or mobile, Windows 11, Raspbian Bullseye, ...):

jimtng:
My thinking was that we go through the _current_ needs from all the existing bindings, and see what tags are missing/needed, then perhaps we'll have a list of tags to consider. It would then be much easier to see / organise them. I've started this from the linked PR, and that's just from two bindings so far.

Then additionally we get input from users about what other tags they need / would suggest. Add them all to the pile and work from there. Posting this request on the forum would probably get us the most feedback.

rkoshak:
Looking at the bindings is a good approach. 

I can open a forum thread, but I'm not super optimistic how useful that will be in the long run. 

I think what I am hoping for is some sort of rules that we can establish and document which describes what makes a good tag compared to a bad tag. For example, is it meaningful to have "Patio", "Porch" and "Terrace" or does it make more sense to have just one of those and use synonyms and the Item label to distinguish between these outdoor areas. Or does it make sense to add "Lenai"?

Do we need "Front Door", "Back Door" and "Side Door" or does it make more sense to have "Exterior Door" and use synonyms and Item label to distinguish between the three? 

That's the sort of thing I'm hoping will come out of this issue. Of course, I have my opinions but even if my wishes are not met, I'll be overjoyed if we can have some rules in place to help maintainers decide what makes a good tag and what does not.

I'm worried that if we don't come up with those rules first, a post to the forum will just gather a ton of tags with lots of overlap.

[...]

PR #4622, opened 2025-02-26
Create static constants for the default semantic tags

jimtng:
This enables the default semantic tags to be referred to as constants by bindings to enable compile time check.


jimtng:
@andrewfg you might be able to use this

jimtng:
If we can be sure that the "simple" tag names must be unique, we can simplify the constants to just the last part of it, i.e.

`CAMERA` instead of `EQUIPMENT_CAMERA`.

andrewfg:
@jimtng here is an example piece of code from the Hue thing handler whereby the thing's semantic equipment tag gets set dynamically. Ideally I would pull in your EQUIPMENT_xx tags from this PR rather than using hand coded strings.

```
    /**
     * Update the thing's semantic equipment tag. The main determinant for the equipment type is the supported channels.
     * For lights use the product archetype to split between light bulbs and strip lights. Rooms and Zones are
     * considered to be (groups of) light bulbs.
     */
    // TODO use the semantic tag constants from OH core
    private void updateSemanticEquipmentTag() {
        String semanticEquipmentTag = null;

        // sensor equipment
        if (supportedChannelIdSet.contains(CHANNEL_2_LIGHT_LEVEL) //
                || supportedChannelIdSet.contains(CHANNEL_2_MOTION)
                || supportedChannelIdSet.contains(CHANNEL_2_TEMPERATURE)) {
            semanticEquipmentTag = "Sensor";
        } else

        // security equipment
        if (supportedChannelIdSet.contains(CHANNEL_2_SECURITY_CONTACT)) {
            semanticEquipmentTag = "Security";
        } else

        // button equipment
        if (supportedChannelIdSet.contains(CHANNEL_2_BUTTON_LAST_EVENT)) {
            semanticEquipmentTag = "Button";
        } else

        // rotary dial equipment
        if (supportedChannelIdSet.contains(CHANNEL_2_ROTARY_STEPS)) {
            semanticEquipmentTag = "RotaryDial";
        } else

        // rooms and zones are a super-set of light bulb equipment
        if (thisResource.getType() != ResourceType.DEVICE) {
            semanticEquipmentTag = "LightBulb";
        } else

        // everything else is individual light equipment
        if (thisResource.getProductData() instanceof ProductData productData) {
            if (STRIPLIGHT_ARCHETYPES.contains(productData.getProductArchetype())) {
                semanticEquipmentTag = "LightStrip";
            } else {
                semanticEquipmentTag = "LightBulb";
            }
        }

        logger.debug("{} -> updateSemanticEquipmentTag({})", resourceId, semanticEquipmentTag);
        updateThing(editThing().withSemanticEquipmentTag(semanticEquipmentTag).build());
    }
```

jimtng:
> Actually I strongly prefer the constants WITH prefixes. When developing a binding, if you want to apply a tag on a thing then the tag must be an EQUIPMENT_xx tag. Or if you want to apply tags to a channel they must be POINT_xx and/or PROPERTY_xx tags.

I believe that I understand your point.

My thinking is that when you're setting a tag here, you are selecting the "end point" of the tag without dictating the exact hierarchy of the tag. Yes, it must be a certain type (Equipment vs Point / Property) but in _here_ you are not dictating that Equipment Z _must_ be a subcategory of Y which is the subcategory of X. 

So instead of stating here `DefaultSemanticTags.EQUIPMENT_X_Y_Z`, you just say `DefaultSemanticTags.Z`. 

This frees up the ability later on to restructure Z within core to be either promoted or demoted e.g. to become EQUIPMENT_Z, EQUIPMENT_X_Z, or even EQUIPMENT_A_B_Z without having to make changes in all the consumer of the constants.

It is still `Z` to you here, due to the virtue of `Z` having to be unique. There can never be EQUIPMENT_A_Z and EQUIPMENT_B_Z to require specifying the full path.

The advantage is simpler looking code. You know you want a `FRIDGE`. You shouldn't have to care if it's EQUIPMENT_FREEZER or EQUIPMENT_WHITEGOODS_FREEZER, or anything else.

It has a 1 to 1 equivalent of using a plain string vs constant.

jimtng:
> A possible compromise could be to create the constants as TYPE_ENDNAME without having the hierarchy. So we'll have
> 
> `EQUIPMENT` (the root tag) `EQUIPMENT_FREEZER` (not `EQUIPMENT_WHITEGOODS_FREEZER`) `LOCATION_LAUNDRYROOM` (not `LOCATION_INDOOR_ROOM_LAUNDRYROOM`)

I've changed it to this scheme now. I recognise that this makes it very clear the type of the tag without getting into the hierarchy details.

jimtng:
@andrewfg and @jlaur - updated to
`DefaultSemanticTags.Location.SUMMER_HOUSE`

bundles/org.openhab.core.semantics/model/generateTagClasses.groovy
```diff
@@ -60,6 +64,62 @@ def appendLabelsFile(FileWriter file, def line, String tagSet) {
     file.write("\n")
 }
 
+def camelToUpperCasedSnake(def tag) {
```
andrewfg, line 67: Brilliant! One of my wishes (when I grow up) is to be able to do regexes like that..

bundles/org.openhab.core.semantics/src/main/java/org/openhab/core/semantics/model/DefaultSemanticTags.java
```diff
@@ -0,0 +1,678 @@
...
+import org.openhab.core.semantics.SemanticTag;
+import org.openhab.core.semantics.SemanticTagImpl;
+
+/**
+ * This class defines all the default semantic tags.
+ *
+ * @author Generated from generateTagClasses.groovy - Initial contribution
+ */
+@NonNullByDefault
+public class DefaultSemanticTags {
```
andrewfg, line 25: Yes, I like this structure. Thankyou.

andrewfg (approved):
LGTM subject to #4616 and #4619 being merged.

jimtng:
rebased, add new tags

andrewfg (approved):
LGTM

kaikreuzer (approved):

PR #4624, opened 2025-02-27
[WIP] Add Java24 early access build

holgerfriedrich:


holgerfriedrich:
Spotbugs >4.8 needed, as spotbugs is not using the maven.compiler setting an thus generates java24 class files.
Currently issue in 4.9.1 leasing to OOM in voice.googlestt addon. -> postponing until this is resolved

PR #4628, opened 2025-03-03
Upgrade XChart to 3.8.8

holgerfriedrich:
Upgrade XChart from 3.1.0 to 3.8.8
See changelog:
https://knowm.org/open-source/xchart/xchart-change-log/

### DIFF
--- a/bundles/org.openhab.core.ui/pom.xml
+++ b/bundles/org.openhab.core.ui/pom.xml
@@ -45,12 +45,6 @@
       <artifactId>org.openhab.core.bom.compile-model</artifactId>
       <type>pom</type>
     </dependency>
-    <dependency>
-      <groupId>org.junit.jupiter</groupId>
-      <artifactId>junit-jupiter-engine</artifactId>
-      <version>5.11.4</version>
-      <scope>test</scope>
-    </dependency>
   </dependencies>
 
 </project>
--- a/bundles/org.openhab.core.ui/pom.xml
+++ b/bundles/org.openhab.core.ui/pom.xml
@@ -19,7 +19,7 @@
       <groupId>org.knowm.xchart</groupId>
       <artifactId>xchart</artifactId>
       <version>3.8.8</version>
-      <!-- for for pdfbox issue https://issues.apache.org/jira/browse/PDFBOX-5722 merged but not included in 3.8.8 -->
+      <!-- for pdfbox issue https://issues.apache.org/jira/browse/PDFBOX-5722 merged but not included in 3.8.8 -->
       <exclusions>
         <exclusion>
           <groupId>org.junit.jupiter</groupId>
--- a/bundles/org.openhab.core.ui/pom.xml
+++ b/bundles/org.openhab.core.ui/pom.xml
@@ -19,6 +19,13 @@
       <groupId>org.knowm.xchart</groupId>
       <artifactId>xchart</artifactId>
       <version>3.8.8</version>
+      <!-- for for pdfbox issue https://issues.apache.org/jira/browse/PDFBOX-5722 merged but not included in 3.8.8 -->
+      <exclusions>
+        <exclusion>
+          <groupId>org.junit.jupiter</groupId>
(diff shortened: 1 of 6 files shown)


wborn:
Does the legend work correctly now?

See the previous attempt to upgrade it: https://github.com/openhab/openhab-core/pull/3291#issuecomment-1622207448

holgerfriedrich:
Thanks, @wborn , for the reference. I was not aware of this!

I cannot reproduce this on 3.8.8, the axis labels are shown. Though, I cannot tell if I have replaced all relevant jars in my dev system. There is at least a risk that I did not properly test this.

FYI:
Starting with version 3.8.7, the builds were failing in the surefire step. Log was not helpful. The surefire log from target folder was also not really helpful. Apparently, adding a new dependency for test to the bom file one folder above resolved this issue. Without, as at least two other bundles were failing in the surefire step.

wborn:
I see it got fixed with https://github.com/knowm/XChart/pull/774 but XChart 3.8.5 was not released in time for OH 4.0 so that's why it didn't get upgraded but was reverted instead.

bundles/pom.xml
```diff
@@ -131,6 +131,12 @@
       <type>pom</type>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-engine</artifactId>
+      <version>5.11.4</version>
+      <scope>test</scope>
+    </dependency>
```
wborn, line 139: Looks like the pdfbox artifact has junit artifacts as a transitive dependency.
Perhaps you can exclude them from the xchart dependency instead?

```
[INFO] +- org.knowm.xchart:xchart:jar:3.8.8:compile
[INFO] |  +- de.erichseifert.vectorgraphics2d:VectorGraphics2D:jar:0.13:compile
[INFO] |  +- de.rototor.pdfbox:graphics2d:jar:3.0.1:compile
[INFO] |  |  \- org.apache.pdfbox:pdfbox:jar:3.0.1:compile
[INFO] |  |     +- org.apache.pdfbox:pdfbox-io:jar:3.0.1:compile
[INFO] |  |     +- org.apache.pdfbox:fontbox:jar:3.0.1:compile
[INFO] |  |     +- commons-logging:commons-logging:jar:1.2:compile
[INFO] |  |     \- org.junit.jupiter:junit-jupiter:jar:5.10.1:compile
[INFO] |  |        \- org.junit.jupiter:junit-jupiter-engine:jar:5.10.1:runtime
[INFO] |  |           \- org.junit.platform:junit-platform-engine:jar:1.10.1:runtime
```

wborn: It's fixed in pdfbox 3.0.2 so you could also manage the dependency to that version, see:

https://issues.apache.org/jira/browse/PDFBOX-5722

xchart should upgrade to that if they haven't already.

bundles/org.openhab.core.ui/pom.xml
```diff
@@ -18,8 +18,14 @@
     <dependency>
       <groupId>org.knowm.xchart</groupId>
       <artifactId>xchart</artifactId>
-      <!-- Newer versions have issues with customized grid lines, see: https://github.com/knowm/XChart/issues/628 -->
-      <version>3.1.0</version>
+      <version>3.8.8</version>
+      <!-- for for pdfbox issue https://issues.apache.org/jira/browse/PDFBOX-5722 merged but not included in 3.8.8 -->
```
wborn, line 22:
```suggestion
      <!-- for pdfbox issue https://issues.apache.org/jira/browse/PDFBOX-5722 merged but not included in 3.8.8 -->
```

holgerfriedrich:
@wborn Thanks, excluding the transitive dependency is much cleaner.

Actually, XChart already did the dependency update to pdfbox 3.0.2 with knowm/XChart#851, but this is not yet part of a release.

wborn (approved):
Thanks for the upgrade!

PR #4632, opened 2025-03-05
OAuth RFC-8628 Device Code Grant Flow

andrewfg:
This PR adds [RFC-8628 Device Code Grant Flow](https://datatracker.ietf.org/doc/html/rfc8628) support to the OH Core OAuth service.

Specifically it adds new RFC-8628 specific classes and extends the functionality of already existing OAuth services to add support for the new RFC-8628 Device Code Grant Flow functionality.

This PR is required by https://github.com/openhab/openhab-addons/pull/18354 -- which has an implementation deadline of March 15th 2025.

Depends on https://github.com/openhab/openhab-core/pull/4647

image


andrewfg:
@mherwege @jlaur .. apropos https://github.com/openhab/openhab-addons/pull/18354

andrewfg:
Just to confirm that this is ready for review.

andrewfg:
@kritsel many thanks for pointing out the need for this.

andrewfg:
PS I am working on some tweaks to respond to server error responses..

andrewfg:
@openhab/add-ons-maintainers in case you need evidence of successful testing .. please see work flow below..

### 1. Thing offline: click link to servlet



### 2. Servlet: click link to authentication URL



### 3. Enter User Id



### 3. Enter credentials



### 4. Credentials confirmed



### 5. Thing online

lsiepel (changes requested):
Just one comment.

bundles/org.openhab.core.auth.oauth2client/src/main/java/org/openhab/core/auth/oauth2client/internal/OAuthConnectorRFC8628.java
```diff
@@ -0,0 +1,471 @@
...
+             * If we got here without exception then the (local or remote) DeviceCodeResponse is all good:
+             * => try to fetch an AccessTokenResponse for it from the remote server
+             * => if the fetch fails-soft (authentication not yet done) it returns a null AccessTokenResponse
+             * => if the fetch fails-hard it throws an exception
+             */
+            try {
+                atr = fetchAccessTokenResponse(dcr);
+                logger.trace("getDeviceCodeResponse() fetched from remote: {}", atr);
+            } catch (OAuthResponseException e) {
+            }
```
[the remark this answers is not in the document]
andrewfg: Umm. If `fetchAccessTokenResponse()` does not return a one-off result it simply means that the value of the `atr` variable remains null as it was before, and that we shall proceed further and start the polling cycle to try again repeatedly after 5 seconds. The OAuthResponseException here It should definitely NOT cause the method to "return null". However I will add the logging and explicitly set `atr = null`..

lsiepel (approved):

bundles/org.openhab.core.auth.oauth2client/src/main/java/org/openhab/core/auth/oauth2client/internal/OAuthStoreHandlerImpl.java
```diff
@@ -308,6 +367,16 @@ public Set<String> getAllHandlesFromIndex() {
                         logger.info("Unable to deserialize json, reset LAST_USED to now.  json:\n{}", value);
                         return Instant.now();
                     }
+                } else if (DEVICE_CODE_RESPONSE.equals(recordType)) {
+                    try {
+                        return gson.fromJson(value, DeviceCodeResponse.class);
+                    } catch (Exception e) {
+                        logger.error(
+                                "Unable to deserialize json, discarding DeviceCodeResponse.  "
```
holgerfriedrich, line 375:
```suggestion
                                "Unable to deserialize json, discarding DeviceCodeResponse. "
```

bundles/org.openhab.core.auth.oauth2client/src/main/java/org/openhab/core/auth/oauth2client/internal/OAuthStoreHandlerImpl.java
```diff
@@ -308,6 +367,16 @@ public Set<String> getAllHandlesFromIndex() {
                         logger.info("Unable to deserialize json, reset LAST_USED to now.  json:\n{}", value);
                         return Instant.now();
                     }
+                } else if (DEVICE_CODE_RESPONSE.equals(recordType)) {
+                    try {
+                        return gson.fromJson(value, DeviceCodeResponse.class);
+                    } catch (Exception e) {
```
holgerfriedrich, line 373: Is it necessary to catch this very generic `Exception` - according to the docs, [fromJson](https://www.javadoc.io/static/com.google.code.gson/gson/2.12.1/com.google.gson/com/google/gson/Gson.html#fromJson(com.google.gson.JsonElement,java.lang.Class)) should throw [JsonSyntaxException](https://www.javadoc.io/static/com.google.code.gson/gson/2.12.1/com.google.gson/com/google/gson/JsonSyntaxException.html) (at least this is what they declared, not sure what else could happen)

bundles/org.openhab.core.auth.oauth2client/src/main/java/org/openhab/core/auth/oauth2client/internal/OAuthConnectorRFC8628.java
```diff
@@ -0,0 +1,475 @@
...
+    }
+
+    /**
+     * Begins the RFC-8628 Device Code Grant Flow authentication process.
+     * Specifically it executes the following steps as described in the article in the link below:
+     *
+     * <ul>
+     * <li>Step 1: create a request and POST it to the 'device authorize url'</li>
+     * <li>Step 2: process the response and create a {@link DeviceCodeResponse}</li>
+     * <li>Step 3: the user goes off to authenticate themselves at the 'user authentication url</li>
```
holgerfriedrich, line 132: unmatched '

bundles/org.openhab.core.auth.oauth2client/src/main/java/org/openhab/core/auth/oauth2client/internal/OAuthConnectorRFC8628.java
```diff
@@ -0,0 +1,475 @@
...
+     * <li>Step 4: repeatedly create a request and POST it to the 'token url'</li>
+     * <li>Step 5: repeatedly read the response and eventually create a {@link AccessTokenResponse}</li>
+     * </ul>
+     *
+     * @see <a href="https://datatracker.ietf.org/doc/html/rfc8628">RFC-8628</a>
+     *
+     * @return either null or a {@link DeviceCodeResponse} containing the verification uri's where the
+     *         user is expected authenticate themselves
```
holgerfriedrich, line 140:
```suggestion
     *         users are expected to authenticate themselves
```

holgerfriedrich (changes requested):
Thanks, @andrewfg, for bringing this into core.

I left a few comments, mainly related to typos / spelling. Please consider those comments as a recommendation, not a mandatory change.

One comment about code (exception handling).

SAT is good (cannot avoid the warnings around the `pre` tag, spotless wants it this way). Not many compiler warnings left.

Nullable could be an improvement, but not necessary for the first merge.

Test is missing. I would appreciate to have at least one test of the overall functionality.
As there seems to be some time pressure, I would merge without tests.

andrewfg:
@holgerfriedrich many thanks for the feedback. I fixed all your issues except (so far) the following:

> Not many compiler warnings left.

I think the 'remaining' warnings were actually there prior to this PR. i.e. not my code. I could certainly fix them but I worry that the warnings may have been there for a good reason and that I might break something arcane. => WDYT?

> Nullable could be an improvement, but not necessary for the first merge.

Could you be more precise about where you suggest this? Is it the missing warnings above? (PS the new DeviceCodeResponse is a DTO so I did not annotate that as NonNullByDefault..)

> Test is missing. I would appreciate to have at least one test of the overall functionality.

Yes. I did do some local testing of e.g. serialization / deserialization of DTOs etc. but those were so obvious trivial that I did bother to put them in the test suite. As you say, the real test would be on the end to end functionality. This would involve mocking the Http transactions with potentially asynchronous timed responses, so to be honest I ducked the issue. But let me see if I can come up with something later today.

holgerfriedrich:
> @holgerfriedrich many thanks for the feedback. I fixed all your issues except (so far) the following:
> 
> > Not many compiler warnings left.
> 
> I think the 'remaining' warnings were actually there prior to this PR. i.e. not my code. I could certainly fix them but I worry that the warnings may have been there for a good reason and that I might break something arcane. => WDYT?

Out of scope of this PR.
We already had a few separate PRs in 5.0 to reduce warnings.
That should be handled separately. As you said, it might break something.

> 
> > Nullable could be an improvement, but not necessary for the first merge.
> 
> Could you be more precise about where you suggest this? Is it the missing warnings above? (PS the new DeviceCodeResponse is a DTO so I did not annotate that as NonNullByDefault..)

We have some suppressions for DTOs. If the file name of the DTO ends with `DTO.java`, the following checkstyle suppressions are in place:
https://github.com/openhab/openhab-core/blob/41720b4f0c025f8884064be0b18d48ab656ead3a/tools/static-code-analysis/checkstyle/suppressions.xml#L7

holgerfriedrich (approved):
LGTM

andrewfg:
@holgerfriedrich two things 

1. I renamed the DTO to make spotless more happy but the CI Build is now stuck on _"Build (Java 21, ubuntu-24.04) Expected — Waiting for status to be reported"_ => is there some way to un-stick it?
2. I am trying to write a Junit test but I am getting stuck trying to mock the Jetty HttpClient => do you have any tips? Or is there an example elsewhere in OH Core?

holgerfriedrich:
@andrewfg Build fail is not related to this PR, apparently there is an incident with one of the repos we use in our GHA build toolchain....
You should rebase after #4647 is merged.

holgerfriedrich:
GH has shut down one of the repos we use in our build action, thus the action is crashing in the preparation phase. This should be resolved when the PR above is merged and you can rebase. For now, you could cherry-pick it into your PR.

Regarding the test: I have no experience with any tests going beyond the standard unit tests...

andrewfg:
@holgerfriedrich I back ported your PR and at least the build starts OK; but it seems to have frozen on `142/142` integration tests this time.

holgerfriedrich:
Let's wait until #4647 is merged and then rebase.

Code looks good, thanks for adding the tests.

Not sure why the build failed, rebuild went fine. I will add a more realistic timeout to the workflow (IIRC default timeout is 6h, much too long for this job).

holgerfriedrich:
@andrewfg can you please remove the cherry-picked PR and rebase? Thanks!

holgerfriedrich:
Thanks, @andrewfg!

andrewfg:
@holgerfriedrich I wonder if it is possible resp. advisable to Cherry Pick this one back into v4.x? => WDYT?

https://github.com/openhab/openhab-addons/issues/18425#issuecomment-2746122779

lsiepel:
For the addon i don't see a problem. For core i also don't expect that it will introduce much risks, but it is up to a core maintainer to decide. Once this is backported in core, i can do the same for addon.

holgerfriedrich:
I backported the change, as the alternative is having a now broken plug-in in 4.3.x.
With this change, you should be able to restore existing functionality.

PR #4635, opened 2025-03-09
Provide Karaf Session object inside OSGiConsole

jimtng:
This PR provides Karaf's Session object through OSGiConsole object that's provided to `ConsoleCommandExtension`s.

The Karaf `Session` enables ConsoleCommandExtensions to prompt for user input. The simplest usage is provided through a convenience method `OSGiConsole.readLine()`.

A more advanced usage is also possible by constructing a custom org.jline.reader.LineReader object which requires the Terminal from console.getSession().getTerminal(). This is useful for more complex input scenarios, such as when you need to have sub-shell history, custom completion, etc.

This PR will enable the creation of a JRuby REPL inside the Karaf console. This REPL will run inside the same engine / environment as a normal JRuby rule, so it will have access to runtime Items, Things, events, etc, just like a normal openHAB script.


holgerfriedrich (approved):
LGTM, thanks!

PR #4640, opened 2025-03-13
Allow startlevel to increase past 80 when a bridge is disabled

jimtng:
Issue reported here: https://github.com/openhab/openhab-addons/issues/18311#issuecomment-2709903476

A Thing that's not ready because its bridge is disabled should not hold up the startlevel.

### DIFF
--- a/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
+++ b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
@@ -1173,40 +1173,45 @@ public class ThingManagerImpl implements ReadyTracker, ThingManager, ThingTracke
         updateInstructions.keySet().removeIf(key -> thingHandlerFactory.equals(key.factory()));
     }
 
+    private boolean allEnabledThingsAreInitialized() {
+        for (Thing thing : things.values()) {
+            if (!thing.isEnabled() || ThingHandlerHelper.isHandlerInitialized(thing)) {
+                continue;
+            }
+
+            int bridgeNestingLevel = 0;
+            boolean bridgeDisabled = false;
+            Bridge bridge = getBridge(thing.getBridgeUID());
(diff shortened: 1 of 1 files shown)


jimtng:
@J-N-K would you mind taking a look at this please?

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1175,7 +1176,30 @@ protected synchronized void removeThingHandlerFactory(ThingHandlerFactory thingH
...
+                int bridgeNestingLevel = 0;
+                boolean bridgeDisabled = false;
+                Bridge bridge = getBridge(thing.getBridgeUID());
+                while (bridge != null) {
+                    if (!bridge.isEnabled()) {
+                        bridgeDisabled = true;
+                        break;
+                    }
+                    bridge = getBridge(bridge.getBridgeUID());
+                    if (bridgeNestingLevel++ > MAX_BRIDGE_NESTING) {
```
J-N-K, line 1194: Why do you limit the nesting here? I agree that there is probably no use-case for 100 levels of bridge-nesting, but what would be the problem here? Potential stack overflow if nesting is too deep?

J-N-K (changes requested):
Otherwise looks good. A potential improvement would be to move the check to a separate method. Inlining complex is usually hard to read.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java
jimtng: avoid an endless loop due to circular reference

holgerfriedrich:
@jimtng can you please rebase? We can't get a CI build for the last commit. It still uses tj-actions, which have been forbidden on our repo when due to the hack a 3 weeks ago. Thanks!

PR #4641, opened 2025-03-13
Upgrade plugins

holgerfriedrich:
* maven-clean-plugin to 3.4.1 https://github.com/apache/maven-clean-plugin/releases/tag/maven-clean-plugin-3.4.1
* maven-compiler-plugin to 3.14.0 https://github.com/apache/maven-compiler-plugin/releases/tag/maven-compiler-plugin-3.14.0
* maven-deploy-plugin to 3.1.4 https://github.com/apache/maven-deploy-plugin/releases/tag/maven-deploy-plugin-3.1.4
* maven-intstall-plugin to 3.1.4 https://github.com/apache/maven-install-plugin/releases/tag/maven-install-plugin-3.1.4
* i18n-plugin:
  * upgrade maven-plugin-api to 3.6.3
  * upgrade maven-plugin-annotations to 3.15.1

### DIFF
--- a/tools/i18n-plugin/pom.xml
+++ b/tools/i18n-plugin/pom.xml
@@ -18,7 +18,6 @@
   <description>Generates translations files</description>
 
   <properties>
-    <maven.core.version>3.6.0</maven.core.version>
     <maven.plugin.api.version>3.9.6</maven.plugin.api.version>
     <maven.plugin.annotations.version>3.15.1</maven.plugin.annotations.version>
     <maven.plugin.plugin.version>3.15.1</maven.plugin.plugin.version>
--- a/tools/i18n-plugin/pom.xml
+++ b/tools/i18n-plugin/pom.xml
@@ -19,8 +19,8 @@
 
   <properties>
     <maven.core.version>3.6.0</maven.core.version>
-    <maven.plugin.api.version>3.6.0</maven.plugin.api.version>
-    <maven.plugin.annotations.version>3.6.0</maven.plugin.annotations.version>
+    <maven.plugin.api.version>3.9.6</maven.plugin.api.version>
+    <maven.plugin.annotations.version>3.15.1</maven.plugin.annotations.version>
     <maven.plugin.plugin.version>3.15.1</maven.plugin.plugin.version>
     <maven.plugin.compiler.version>3.8.1</maven.plugin.compiler.version>
   </properties>
--- a/pom.xml
+++ b/pom.xml
@@ -268,13 +268,13 @@ Import-Package: \\
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-clean-plugin</artifactId>
-          <version>3.4.0</version>
+          <version>3.4.1</version>
         </plugin>
 
         <plugin>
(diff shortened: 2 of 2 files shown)


tools/i18n-plugin/pom.xml
```diff
@@ -19,8 +19,8 @@
 
   <properties>
     <maven.core.version>3.6.0</maven.core.version>
-    <maven.plugin.api.version>3.6.0</maven.plugin.api.version>
-    <maven.plugin.annotations.version>3.6.0</maven.plugin.annotations.version>
+    <maven.plugin.api.version>3.9.9</maven.plugin.api.version>
+    <maven.plugin.annotations.version>3.15.1</maven.plugin.annotations.version>
```
wborn, line 23: AFAIK these API versions are usually not set to the latest version as it may also result into devs having to upgrade their Maven version. By not requiring the latest APIs you allow devs more freedom with the Maven version they use. If you really want to enforce everyone (reason?) to always use this Maven version the [enforcer config](https://github.com/openhab/openhab-core/blob/41720b4f0c025f8884064be0b18d48ab656ead3a/pom.xml#L634-L636) should also be updated.

holgerfriedrich: Thanks for having a look.

Ok, I have set `plugin.api.version` to 3.6.3 (all before are EOL).
Not sure about `plugin.annotations.version`, I thought it should match version of `plugin-plugin` (which is 3.15.1).

holgerfriedrich:
Just for the background:
I am trying to push plugins and properties to a recent level manually, based on the following scripts:

```
mvn versions:display-plugin-updates -nsu -Dmaven.version.ignore=".*-beta-.*,.*\.BETA[0-9]+,.*-M[0-9]+" -B -Dversions.outputLineWidth=120 |tee upd
cat upd|grep -v Downl|grep "\.\.\."|grep -v "openHAB Core"|grep -- "->" |sort -u


mvn versions:display-property-updates -nsu -Dmaven.version.ignore=".*-beta-.*,.*-alpha-.*,.*\.BETA[0-9]+,.*-M[0-9]+,.*-rc[-]?[0-9]" -B -Dversions.outputLineWidth=120 |tee props
cat props|grep -v Downl|grep "\.\.\."|grep -v "openHAB Core"|grep -- "->" |sort -u
```

This is likely a stupid and manual version of renovate and dependabot ;-)

tools/i18n-plugin/pom.xml
wborn: As https://github.com/openhab/openhab-core/pull/4664 got merged it now does make sense to update these versions.

tools/i18n-plugin/pom.xml
```diff
@@ -19,8 +19,8 @@
 
   <properties>
     <maven.core.version>3.6.0</maven.core.version>
```
holgerfriedrich, line 21: @wborn do you if this line has an effect at all? I could not find the define used in our code, nor as a documented parameter for a plugin....

wborn: Yes it seems to be unused so you can probably safely remove it.

holgerfriedrich: done

Issue #4649, opened 2025-03-15
Persistence Configuration with nested Groups: restore on startup broken


dandjo:
## Configuration

I have a new problem since OpenHAB 5.0.0-M1 (no issue with OpenHAB 4.3). Following persistence configuration:

```
configurations:
  - items:
      - mapdb_change_restore*
    strategies:
      - restoreOnStartup
      - everyChange
    filters: []
cronStrategies: []
defaultStrategies:
  - restoreOnStartup
  - everyChange
thresholdFilters: []
timeFilters: []
equalsFilters: []
includeFilters: []
```

I have an Item with the following configuration (a group switch for other switches) with ID `ventilation_management`:

```
label: Ventilation Management
type: Group
category: switch
groupNames:
  - mapdb_change_restore
  - influxdb_change
  - ventilation
tags:
  - Equipment
groupType: Switch
function:
  name: OR
  params:
    - ON
    - OFF
```

The group `ventilation` is another group (just for model grouping):

```
label: Ventilation
type: Group
category: fan
groupNames: []
tags:
  - Equipment
groupType: None
function: None
```

The group `mapdb_change_restore` looks as follows:

```
label: 000 MapDB Change Restore
type: Group
category: ""
groupNames: []
tags: []
groupType: None
function: None
```

The `ventilation_management` has a few children like this:

```
label: Ventilation CO2 Management
type: Switch
category: switch
groupNames:
  - ventilation
  - ventilation_management
tags:
  - Switch
```

## Expected Behavior

The Item `ventilation_management` gets persisted and restored on startup, because it is a member of the group `mapdb_change_restore` which is configured to persist.

## Current Behavior

The Item `ventilation_management` gets not persisted and not restored on startup. But its children do, even if they do not have a direct group membership of `mapdb_change_restore`. Normally, an active child triggers the state ON for the parent `ventilation_management`, but this is not the case when restoring from persistence service.

## Your Environment
* Version used: openHAB 5.0.0-M1
* Operating System and version: Ubuntu 24.04 on Raspberry Pi 4

## Discussion

https://community.openhab.org/t/mapdb-persistence-configuration-with-nested-groups/163129

mherwege:
Could you provide some logging showing the ventilation-management effectively changes? With the OR function and an everyChange strategy, itwill only turn OFF if all items in the group turn OFF.

Some debug and event logging from the changing group, persisting and restoring would be helpful. I don’t immediately see the cause of the issue.

[...]

PR #4651, opened 2025-03-17
Fix restoring nested group items from persistence

mherwege:
Closes https://github.com/openhab/openhab-core/issues/4649

https://github.com/openhab/openhab-core/pull/4463 introduced restoring `lastState`, `lastStateUpdate` and `lastStateChange` from persistence.
This broke restoring nested groups, as this relied on update events from the items to the groups. Update events would not allows group state updates to be restored with correct dates, or the `lastState` of the group to be restored.

This PR also saves the nested group itself when it has a state and uses this in restore.


kaikreuzer (approved):
Thanks!

PR #4653, opened 2025-03-18
Add support for concurrent multisink audio

kgoderis:
# Add support for concurrent multisink audio

Allow playing audio concurrently on multiple sinks. In the context of the Sonos binding, this eliminates the need to group or ungroup Sonos players in order to play audio synchronously on all players (e.g. in Rules). Audio is also played concurrently on all sinks using a ThreadPool that is configurable in services.cfg ("threadpool:audio=10"), thereby eliminating the "sequential" output when playing audio on a large group of sinks (e.g. doorbell on Sonos)

Furthermore, it add support for comma-separated lists of sinks or patterns to the various commands, so that arbitrary groupings can be made by the end-user, e.g. "Sonos:CONNECT:*,enhancedjavasound" for usage with playSound() in Rules or the "openhab:audio play" command

### DIFF
--- a/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioManager.java
+++ b/bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioManager.java
@@ -19,6 +19,8 @@ import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.library.types.PercentType;
 
+import io.reactivex.annotations.NonNull;
+
 /**
  * This service provides functionality around audio services and is the central service to be used directly by others.
  *
@@ -27,6 +29,7 @@ import org.openhab.core.library.types.PercentType;
  * @author Christoph Weitkamp - Added parameter to adjust the volume
  * @author Wouter Born - Added methods for getting all sinks and sources
  * @author Miguel Álvarez - Add record method
+ * @author Karel Goderis - Add multisink support
  */
 @NonNullByDefault
 public interface AudioManager {
@@ -52,6 +55,14 @@ public interface AudioManager {
     void play(@Nullable AudioStream audioStream, @Nullable String sinkId);
 
     /**
+     * Plays the passed audio stream on the given set of sinks.
+     *
+     * @param audioStream The audio stream to play or null if streaming should be stopped
+     * @param sinkIds The set of the audio sink ids to use
+     */
+    void play(@Nullable AudioStream audioStream, @NonNull Set<String> sinkIds);
+
+    /**
      * Plays the passed audio stream on the given sink.
      *
      * @param audioStream The audio stream to play or null if streaming should be stopped
@@ -59,6 +70,15 @@ public interface AudioManager {
      * @param volume The volume to be used or null if the default notification volume should be used
      */
     void play(@Nullable AudioStream audioStream, @Nullable String sinkId, @Nullable PercentType volume);
+
+    /**
+     * Plays the passed audio stream on the given set of sinks.
+     *
+     * @param audioStream The audio stream to play or null if streaming should be stopped
+     * @param sinkIds The set of the audio sink ids to use
+     * @param volume The volume to be used or null if the default notification volume should be used
+     */
+    void play(@Nullable AudioStream audioStream, Set<String> sinkIds, @Nullable PercentType volume);
 
     /**
      * Plays an audio file from the "sounds" folder using the default audio sink.
@@ -87,6 +107,15 @@ public interface AudioManager {
     void playFile(String fileName, @Nullable String sinkId) throws AudioException;
 
     /**
+     * Plays an audio file from the "sounds" folder using the given audio sink.
+     *
+     * @param fileName The file from the "sounds" folder
+     * @param sinkIds The set of the audio sink ids to use
+     * @throws AudioException in case the file does not exist or cannot be opened
+     */
+    void playFile(String fileName, @NonNull Set<String> sinkIds) throws AudioException;
+
+    /**
      * Plays an audio file with the given volume from the "sounds" folder using the given audio sink.
      *
      * @param fileName The file from the "sounds" folder
@@ -95,6 +124,16 @@ public interface AudioManager {
      * @throws AudioException in case the file does not exist or cannot be opened
      */
     void playFile(String fileName, @Nullable String sinkId, @Nullable PercentType volume) throws AudioException;
+
+    /**
+     * Plays an audio file with the given volume from the "sounds" folder using the given audio sink.
+     *
+     * @param fileName The file from the "sounds" folder
+     * @param sinkIds The set of the audio sink ids to use
+     * @param volume The volume to be used or null if the default notification volume should be used
+     * @throws AudioException in case the file does not exist or cannot be opened
+     */
+    void playFile(String fileName, @NonNull Set<String> sinkIds, @Nullable PercentType volume) throws AudioException;
 
     /**
      * Stream audio from the passed url using the default audio sink.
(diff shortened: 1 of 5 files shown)


kgoderis:
@wborn Can you help me explain why I get a test error whereas locally all runs fine:

```
[INFO] 
[INFO] --- surefire:3.5.2:test (default-test) @ org.openhab.core.audio ---
[INFO] Using auto detected provider org.apache.maven.surefire.junitplatform.JUnitPlatformProvider
[INFO] 
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running org.openhab.core.audio.internal.AudioServletTest
[INFO] Tests run: 11, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 6.450 s -- in org.openhab.core.audio.internal.AudioServletTest
[INFO] Running org.openhab.core.audio.internal.AudioFormatTest
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.003 s -- in org.openhab.core.audio.internal.AudioFormatTest
[INFO] Running org.openhab.core.audio.internal.AudioConsoleTest
[INFO] Tests run: 9, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.044 s -- in org.openhab.core.audio.internal.AudioConsoleTest
[INFO] Running org.openhab.core.audio.internal.AudioManagerServletTest
[main] WARN  o.o.c.a.internal.AudioManagerImpl - No audio sink provided for playback.
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.008 s -- in org.openhab.core.audio.internal.AudioManagerServletTest
[INFO] Running org.openhab.core.audio.internal.AudioManagerTest
[main] WARN  o.o.c.a.internal.AudioManagerImpl - No audio sink provided for playback.
[INFO] Tests run: 19, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.022 s -- in org.openhab.core.audio.internal.AudioManagerTest
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 46, Failures: 0, Errors: 0, Skipped: 0
```

kgoderis:
@wborn @kaikreuzer Something fishy is going on with the build process on Github. As you can see, the build here above fails on a NPE that does not occur when running tests locally with mvn or in the IDE

Copying over one of the /src/test/java files into a new .java file (without change, and keeping the original "faulty" .java as well) suddenly resolves the error, and the build completes. Then removing that copy again, and the build fails.

kgoderis:
@wborn Clear. so how to proceed with this PR ?

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioManager.java
```diff
@@ -19,6 +19,8 @@
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.library.types.PercentType;
 
+import io.reactivex.annotations.NonNull;
```
holgerfriedrich, line 22: Hi, I just briefly looked into this and I am wondering why you import an additional annotation framework.
As the main class is annotated with @ NonNullByDefault, you may not need this anyway.
Did I overlook something?

kgoderis: @holgerfriedrich You are right, but it is a bad habit of making things explicit in method signatures that comes into play.

holgerfriedrich: @kgoderis The import uses a different framework, I would expect `org.eclipse.jdt.annotation.NonNull` instead of `io.reactivex.annotations.NonNull`....

holgerfriedrich:
@kgoderis could you please rebase?

kgoderis:
> @kgoderis could you please rebase?

Yes, will do that, and fix also the NonNull import

dilyanpalauzov:
I suggest using `Stream.toList()` instead of `Stream.collect(Collectors.toList())`

wborn:
>  so how to proceed with this PR ?

There's https://github.com/openhab/openhab-core/pull/5781 now which may help.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/AudioManagerImpl.java
```diff
@@ -114,74 +125,192 @@ void modified(@Nullable Map<String, Object> config) {
...
+                logger.warn("Sink '{}' not found. Skipping.", sinkId);
+                return null; // Return null for missing sinks
+            }
+
+            // Handle volume adjustment for the current sink
+            Runnable restoreVolume = handleVolumeCommand(volume, sink);
+
+            try {
+                // Play the audio stream synchronously on this sink
+                sink.processAndComplete(audioStream);
```
wborn, line 197: The same `AudioStream` instance is passed to every sink here. `AudioStream` extends `InputStream` and is documented as often being a one-time-use instance, while existing sinks such as `enhancedjavasound` read directly from the supplied stream. With multiple consuming sinks, concurrent reads and ownership of the same object can split or truncate playback, and one sink can finish with the stream while another still needs it. Each sink needs an independent/replayable stream, or the multisink API needs a stream supplier/cache abstraction, before dispatching concurrently.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/AudioManager.java
```diff
@@ -51,6 +54,14 @@ public interface AudioManager {
      */
     void play(@Nullable AudioStream audioStream, @Nullable String sinkId);
 
+    /**
+     * Plays the passed audio stream on the given set of sinks.
+     *
+     * @param audioStream The audio stream to play or null if streaming should be stopped
+     * @param sinkIds The set of the audio sink ids to use
+     */
+    void play(@Nullable AudioStream audioStream, @NonNull Set<String> sinkIds);
```
wborn, line 63: Adding this `Set<String>` overload next to `play(AudioStream, String)` makes an existing call such as `audioManager.play(stream, null)` ambiguous at compile time. `null` is explicitly part of the existing API contract for selecting the default sink, and the same ambiguity is introduced for `playFile`, `stream`, and `playMelody`. Since the existing `String` argument can already represent a sink ID or pattern, consider extending that syntax to accept the comma-separated selector list directly and keep the `Set<String>` handling internal. A single sink then becomes the special case of a selector resolving to one sink, while the existing API remains source-compatible.

bundles/org.openhab.core.audio/src/main/java/org/openhab/core/audio/internal/AudioManagerImpl.java
```diff
@@ -114,74 +125,192 @@ void modified(@Nullable Map<String, Object> config) {
...
-                return null;
-            }).thenRun(restoreVolume);
-        } else {
-            logger.warn("Failed playing audio stream '{}' as no audio sink was found.", audioStream);
+                logger.error("Error playing audio stream '{}' on sink '{}': {}", audioStream, sinkId,
+                        exception.getMessage(), exception);
+                return null; // Handle the exception gracefully
+            }).thenRun(() -> {
+                restoreVolume.run(); // Ensure volume is restored after playback completes
+                logger.info("Audio stream '{}' has been successfully played on sink '{}'.", audioStream, sinkId);
```
wborn, line 170: `exceptionally` recovers a failed playback by returning `null`, so the following `thenRun` executes for both success and failure. A failed playback therefore logs the error and then also logs that it was successfully played. Keep the failure exceptional until after success-only logging, or use completion handling that distinguishes normal and exceptional completion.

wborn (changes requested):
**Additional AI review following the existing maintainer review:**

The multisink functionality is useful, but the current implementation still has correctness and source-compatibility issues that should be addressed before merge. In addition to the already-open `processAndComplete()` future issue, concurrent playback currently shares a single `AudioStream` instance across sinks, and the new `Set<String>` overloads make existing `null`-sink calls source-ambiguous.

The existing string sink selector can already naturally represent single- and multi-sink selection, so keeping collection handling internal would avoid expanding the public API and preserve compatibility.

Targeted multisink tests should cover independent stream consumption and delayed `processAndComplete()` completion.

No current CI/check run or static-analysis report is available for this HEAD, so current build/static-analysis validation could not be included in this review.

Issue #4645, opened 2025-03-14
[persistence] Upgrade to 5.0.0.M1 and later SNAPSHOT cause "Querying persistence service 'influxdb' to restore takes more than 5000ms"


Mark-VG:
Since updating to OH 5.0.0.M1 (and subsequent SNAPSHOT) I am seeing errors as follows: 

```
2025-03-14 09:13:49.863 [WARN ] [ence.internal.PersistenceManagerImpl] - Querying persistence service 'influxdb' to restore 'TestHueLight' takes more than 5000ms.
2025-03-14 09:13:54.870 [WARN ] [ence.internal.PersistenceManagerImpl] - Querying persistence service 'influxdb' to restore 'Partition4_PartitionState' takes more than 5000ms.
```
I reported the issue on the forums at https://community.openhab.org/t/openhab-5-0-milestone-discussion/162686/26

It was suggested to log here as it could be related to https://github.com/openhab/openhab-core/pull/4463

## Context
This occurs on starting/restarting OH since the update. I have updated to the latest influxdb as well as changed to Eclipse Adoptium

## Your Environment
* Version used: openHAB 5.0.0 Build #4542
* Environment name and version (e.g. Chrome 111, Java 17, Node.js 18.15, ...): Java Runtime Eclipse Adoptium 21.0.6 (Temurin-21.0.6+7)
* Operating System and version: Windows 11/10.0 (amd64)

Also seeing the following error in wrapper.log (not sure if this is related?)

```
STATUS | wrapper  | 2025/03/14 13:18:11 | --> Wrapper Started as Service
STATUS | wrapper  | 2025/03/14 13:18:11 | Launching a JVM...
INFO   | jvm 1    | 2025/03/14 13:18:12 | Wrapper (Version 3.2.3) http://wrapper.tanukisoftware.org
INFO   | jvm 1    | 2025/03/14 13:18:12 |   Copyright 1999-2006 Tanuki Software, Inc.  All Rights Reserved.
INFO   | jvm 1    | 2025/03/14 13:18:12 | 
INFO   | jvm 1    | 2025/03/14 13:18:13 | 13:18:13.711 [Framework Event Dispatcher: Equinox Container: b0d69dcc-7f74-4ac0-bc41-35d4ad630127] ERROR Events.Framework - FrameworkEvent ERROR
[...]

PR #4654, opened 2025-03-18
Revert iterating over history for restore on startup

mherwege:
Closes https://github.com/openhab/openhab-core/issues/4645

https://github.com/openhab/openhab-core/pull/4463 introduced restoring `lastState` and `lastStateUpdate` from persistence, introducing a new method `persistedItem` in the `QueryablePersistenceService`.

To achieve this, the default implementation would iterate backwards over the full persisted history. Specific services could override this and implement a better algorithm. This has been done for `mapdb` and `rrd4j`.

It turns out this unbounded iteration over the history can lead to performance problems, see issue. Specifically for `influxdb` there are optimizations when only retrieving the last value.

To avoid problems, I have partially reverted the logic of https://github.com/openhab/openhab-core/pull/4463. The default method will no longer restore `lastState` and `lastStateUpdate`, just `state` and the timestamp. Individual persistence services can implement a specific algorithm to retrieve all if it makes sense. With support for `mapdb`, this may not make much sense if `restoreOnStartup` is configured for `mapdb` anyway.


kaikreuzer (approved):
Many thanks!

PR #4657, opened 2025-03-19
Allow multiple default values to contain an escaped comma

jimtng:
This PR is needed in order to introduce a parameter with multiple="true" whose default values contain a comma character.

This PR allows a comma character to be a part of the value when escaped with a backslash. The backslash will be removed, i.e. `\,` -> `,`

To have an actual backslash and comma, add an extra backslash, i.e. `\\,` -> `\,`

A quick search in the openhab-addons repository didn't yield a current use of comma-separated default values for a `multiple="true"` parameter.

This fix will work in conjunction with a fix in webui to be submitted.

### DIFF
--- a/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/ConfigUtil.java
+++ b/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/ConfigUtil.java
@@ -23,6 +23,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
+import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
 import org.eclipse.jdt.annotation.NonNull;
@@ -51,7 +52,7 @@ import org.slf4j.LoggerFactory;
 @NonNullByDefault
 public class ConfigUtil {
 
-    private static final String DEFAULT_LIST_DELIMITER = ",";
+    private static final Pattern DEFAULT_LIST_SPLITTER = Pattern.compile("(?<!\\\\),");
 
     /**
      * Maps the provided (default) value of the given {@link ConfigDescriptionParameter} to the corresponding Java type.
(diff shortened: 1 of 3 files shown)


lolodomo:
If I correctly understand what you want to achieve, it will certainly require a change at this place:
https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/fileconverter/AbstractThingFileGenerator.java#L124

lolodomo:
And maybe also this method?
https://github.com/openhab/openhab-core/blob/baf79c5f27a7f69b70c21241c8aae68c2e8ca05a/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/ConfigUtil.java#L70

jimtng:
Yes it appears that more work is needed.

jimtng:
> If I correctly understand what you want to achieve, it will certainly require a change in that method?
> https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/fileconverter/AbstractThingFileGenerator.java#L112

I've changed getDefaultValueAsCorrectType to return a List when the param is a `multiple`. I think that this actually fixed the comparisons in https://github.com/openhab/openhab-core/blob/0f354e63f3d196f916a54f535fa764b6cd311829/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/fileconverter/AbstractThingFileGenerator.java#L124 - and here: https://github.com/openhab/openhab-core/blob/0f354e63f3d196f916a54f535fa764b6cd311829/bundles/org.openhab.core.model.item/src/org/openhab/core/model/item/internal/fileconverter/DslItemFileConverter.java#L239

Previously, for a `multiple` configparam, `value` would've been a `List` and `ConfigUtil.getDefaultValueAsCorrectType` would've always returned a converted TEXT / DECIMAL etc and the `equals` comparison param would've always been false because it would be comparing List vs String / BigDecimal etc.

kaikreuzer (approved):
Thanks!

PR #4658, opened 2025-03-20
Upgrade glassfish hk2 to 3.1.1, upgrade units

holgerfriedrich:
* Upgrade glassfish hk2 from 2.4.0-b34 to 3.1.1
* Upgrade jakarta.inject-api from 2.0.0 to 2.0.1 https://github.com/jakartaee/inject/releases/tag/2.0.1
* Upgrade jakarta.annotation-api from 2.0.0 to 2.1.1
* Upgrade si.uom:si-units and si-quantity from 2.1 to 2.2, including transition from jakarta.annotation-api 1.3.4 to 2.0.0
* Upgrade tech.units.indriya from 2.2 to 2.2.2

This is a big jump ahead for glassfish hk2.
I had to upgrade si-units to get in jakarta.annotation-api 2.0 (implicity required jakarta instead of javax namespace).
Depends on openhab/openhab-osgiify#60
Indriya package is upgraded as well to match si-units.

Intriya was not passing the tests, as it returns "yr" instead of "years" in our tests.
Not sure if it is the correct way to adapt the test..... So please comment on this.


wborn (approved):
LGTM! 👍 
Do you know if the Demo App in the distro still works with these upgrades?

holgerfriedrich:
> LGTM! 👍 Do you know if the Demo App in the distro still works with these upgrades?

In principle, yes. This is how I tested.
You will need a `mvn install` in the app folder to adapt to the current state of core.

Though, see openhab/openhab-distro#1734. This seems unrelated, I see it in much older snapshots as well.
Don't know if related to my setup.
One odd thing I have seen as well is that I could see add-ons from the marketplace but none of the official ones. Is this normal if I don't add them explicitly in thee config?

wborn:
> Though, see openhab/openhab-distro#1734. This seems unrelated, I see it in much older snapshots as well.

I'll have a look when I have some more time.

> One odd thing I have seen as well is that I could see add-ons from the marketplace but none of the official ones. Is this normal if I don't add them explicitly in thee config?

Yes it has always been like that because official add-ons are populated using Karaf features which are not available in the Demo App.

holgerfriedrich:
Thanks again, @wborn!

PR #4660, opened 2025-03-20
Improve GitHub Action

holgerfriedrich:
* pin external action stCarolas/setup-maven to specific SHA
* set global timeout
* comments
* Restore broken registration of problem matchers (this was forgotten when we removed Java17)


wborn (approved):
Thanks! Maybe you can also apply this in the other repos?

PR #4665, opened 2025-03-21
 Document use of older nrjavaserial in compile bom

holgerfriedrich:
Remove inconsistency by using latest version also in the compile bom.

### DIFF
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -290,6 +290,9 @@
     <dependency>
       <groupId>com.neuronrobotics</groupId>
       <artifactId>nrjavaserial</artifactId>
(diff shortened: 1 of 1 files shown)


wborn:
The inconsistency between compile/runtime BOMs is for a reason. When APIs do not change or are not used lower version allow for a greater range of OSGi bundle versions to be used.

See: https://github.com/openhab/openhab-core/pull/1464

holgerfriedrich:
Thanks, @wborn.
Sometime it feels like I missed 10 years of OH core development 🙈

The PR now changed completely - I'd like to add a comment, explaining why this should stay as it is.

wborn (approved):
Thanks for the clarification!

Issue #4662, opened 2025-03-21
Possible inconsistency on persistence Layer / possible Item State inconcistency


lo92fr:
Hello guys,

@Jlaur, @florian-h05 , @joerg1985, @mherwege, @J-N-K , @holgerfriedrich : I add you to this issue as I see you have worked recently on persistence layer, and on new timeseries functionality. 

@lolodomo, @clinique, @lsiepel : I add you as review of linky binding.

@weymann : also add you as you was first to respond my question on this issue !

I’m currently working on bindings (Linky) that use timeseries, and have notice some strange behavior. 

I’m currently not sure if I’ve done something wrong, or if there is really some inconsistency in timeseries layer, so I would appreciate your expertise on the good way to have thing done, and how everything should work.

To limit the scope, I would only expose a subset of the binding.

Let say I’ve got a month consumption collection, giving the consumption for (M-x, … M-2, M-1, MCurrent). 
This collection come from my service provider. 
Each value is timestamped, currently at the beginning of month, 00:00:00 time. 
The value is a Quantitype(Double, Units.KILOWATT_HOUR).

What I would like of course is to plot the series using graph, and expose the current item State value for MCurrent. 

Data is refreshing each night, so MCurrent will be update with a different value each morning.

Also, persistent layer is InfluxDB, but I don’t thing it’s really relevant, I observer similar behavior with InMemory persistence.

That is the context.

I’ve got a updateTimeSeries function that is quite simple:

Variant 1:  Just send the time series using SendTimeSeries.

```
TimeSeries timeSeries = new TimeSeries(Policy.REPLACE);
Foreach (value, valueTimestamp : values) {
timeseries.add(timestamp, new QuantityType(value, Units.K)
}
sendTimeSeries(channelId, timeseries);
```

Variant 2: add an updateState call after the sendTimeSeries.
```
TimeSeries timeSeries = new TimeSeries(Policy.REPLACE);
Foreach (value, valueTimestamp : values) {
timeseries.add(timestamp, new QuantityType(value, Units.K)
}
sendTimeSeries(channelId, timeseries);
updateState(channelId, new QuantityType(valueCollection[values[size-1].value, Units.K)
```

Variant 3: same as variant 2, but don’t put value for current month on sendTimeSeries.

Let’s go to test case!

1/ Scenerio 1

In first scenario, I use the variant 1 of updateTimeSeries.
What I observe is that timeseries is correctly initialized.
In influxDB, I see following values:

```
[...]

PR #4666, opened 2025-03-22
fixes proposal for #4662, part1, createDTO changes:

lo92fr:
# fix proposal for #4662, part1, createDTO changes:

# Description

The pull request is intend to fix the handling of itemState=true/false parameters in "rest/persistence/items/xxxx" service calls.
The idea is to verify if current item state is already in the DTO before adding it if itemState = true.
By extension, we also want to remove it from DTO if itemState=false to be fully consistent.

The current / awaited behavior is describe in issue #4662.

- If currentState is already in Dto, and itemState=true, do nothing.
- If currentState is already in Dto, and itemState=false, remove it.
- If currentState is not in Dto, and itemState=false, do nothing.
- If currentState is not in Dto, and itemState=true, add it.

Search for currentState in dto by looping the dto data for value before the current timestamp. Ignore all other data that are future state in Dtos.


mherwege:
If I am not mistaken, the itemState=true parameter was added to make sure a graph from the persisted values would show something at the end and not drop to zero (null value). Otherwise, itemState is false should always be used. So it is not so much about the state, but about the timestamp of the state for itemState=true.
So removing is definitely a no go in my opinion. Also not adding it if the last persisted state = item state is not right, because the intention of the parameter is to get the timestamp right.

Issue #4656, opened 2025-03-19
Possible inconsistency on persistence layer initialization,


lo92fr:
Hello All,

I'm currently working on the Linky binding, try to fix a reliability issue I've got with timeseries handling on the binding.

The binding exposes 4 different time series 👍

- Daily Consumptions
- Weekly Consumptions
- Monthly Consumptions
- Yearly Consumptions.

What I observe is that when I start the binding 👍

- Sometimes All 4 timeseries are correctly initialized.
- Sometimes only 1 or 2 timeseries are correctly initialized, over are empty.
- Sometimes All 4 timeseries are empty.

I first was thinking of course it was an error in my binding.
But after trace the issue, I've found the following issues:

I put a trace on GenericItem::applyTimeSeries(), and one another trace in 
PersistenceManageImpl:timeSeriesUpdated().

Normally, appyTimeSeries should called PersistentManagerImple.timeSeriesUpdated().

But what I find is that sometimes, the timeSeriesListeners in applyTimeSeries is not initialize when applyTimeSeries is called. It seems to come mainly of the fact that there is a lot of item in the registry, and the loop in PeristenceManagerImpl:allItemsChaned is not fully end when the timeseries update event occurs.

```
@OverRide
public void allItemsChanged(Collection oldItemNames) {
itemRegistry.getItems().forEach(this::added);
}
```

Of course, I can delay a little bit my addon initialization to prevent this condition.
But it seems strange that there is not some sort of locking / protection in the PersistenceManagerImpl to prevent this.

I take a full trace to explain more in details what happened

What we can see is that:

- at 10:48:41.531 : PersistenceManagerImpl.onReadyMarkerAdded() : the PersistenceManagerImpl is start a little after my binding initialization.
This function launch the loop to register the persistence listenner on all items.

- at 10:49:05.305, the loop only end at this point, near the end of the trace. The item registry contains around ~500 entries.

- at 10:48:44.314 : the binding start to send timeseries update.

- at 10:48:44.958 : we receive event in the core for timeseries update, but at this point, the timeseries listener is not yet initialized.

- at 10:48:45.010 : for another timeseries, this time listener is correctly initialized !

- at 10:48:45.319 : we see that only Month timeseries are correctly handle.

```
____________________________
Welcome to Apache Felix Gogo

g! 

10:48:27.499 [INFO ] [org.openhab.core.Activator          ] - Starting openHAB 5.0.0 (Unknown Build No.)

...

[...]

PR #4678, opened 2025-03-25
Fix timeseries restore init, #4656,

lo92fr:
# Fix timeseries restore init late initialization of listenners on GenericItem, #4656,

# Description

This PR is to fix lost value in timeseries update observe on issue #4656 on openhab startup.

The GenericItem implements a timeSeriesListener that is initialized from PersistentManagerImpl.allItemsChanged().

The initialization of this timeSeriesListerner can occurs lately after initialization of openhab because the init is in the same loop as the perstistence value restore from persistence layer. 

This persistence restore take long time because of multiple request to the persistence store and db, and so delayed the initialization.

The idea there is to break the loop into 2 parts:
One for the timeSeriesListener initialization which is very fast, and have to be done immedialty.
One for the value restore that can take more time.


lo92fr:
I closed this one, take wrong modification in the PR.

PR #4680, opened 2025-03-25
Adjust dependencies to fix JavaDoc build

holgerfriedrich:
* remove duplicate import
* add apiguardian-api, a transitive dependeny of junit, to compile bom to avoid JavaDoc warning


wborn (approved):
Thanks!

Issue #4625, opened 2025-02-28
[persistence] Persistence filter not accepting units of measure


mkroch:
In OH 4.3.3 in jdbc.persistance file, I have pasted following filter (based on documentation exapmle @ https://www.openhab.org/docs/configuration/persistence.html):

`Filters {
    temp_sanity : [] -100 200 C
}`

After save, log shows:

`Configuration model ‘jdbc.persist’ has errors, therefore ignoring it: [11,1]: mismatched input ‘}’ expecting ‘:’`

The same problem is when I change “C” to "°C" or “K” and so on. When not using units of measure I get a following error in log:

`2025-02-28 09:06:48.916 [WARN ] [ence.filter.PersistenceIncludeFilter] - Cannot compare 5.03125 °C to range -100 - 200`

At first I thought that, as mentioned above, the documentation might be wrong because it not only says that UoM in presistance filter are accepted (and uses them in examples) but stresses that they should be used:

`In case of dimensional items (e.g Number:Temperature) a unit MUST be provided.` 

But then no fillter is possible since comparison cannot be made.

I did contact the community (https://community.openhab.org/t/persistance-filter-not-accepting-uom/162764/6) and was directed here. 

## Expected Behavior
My intention was to include in database recording only readings of Number:Temperature item that are in the specified range. This way I could exclude readings that are abnormal like -1000°C which I occasionally receive due to sensor hardware connectivity issues.

## Current Behavior
Everything apart of the above works fine.

## Possible Solution
Is it a bug?

## Steps to Reproduce (for Bugs)
I am using oh on Ubuntu. Presistance is mariadb via jdbc. Checked it with rrd4j and it looked the same

## Your Environment
I am using 4.3.3 oh on Ubuntu. Presistance is mariadb via jdbc. Checked it with rrd4j and it looked the same. The issue arouse while using oh 4.3.2 update to 4.3.3. did not help apart of the fact that persistance filter started accepting negative values (as mentioned in https://github.com/openhab/openhab-core/pull/4400 ).

wborn:
It's transferred now.

mherwege:
@mkroch Can you try:

>  `Filters { temp_sanity : [] -100 200 “°C” }`

Looking at the code, it expects a String for the unit, and a String needs to be in quotes. If that solves the issue, a documentation change may be needed.

mkroch:
[...]

PR #4681, opened 2025-03-26
Persistence filters allow unquoted units

mherwege:
Closes https://github.com/openhab/openhab-core/issues/4625#issuecomment-2692693848

Persistence definition filters with units required the unit to be a valid string. The parser would not recognize units containing special characters as Strings (e.g. °C, kΩ) unless the unit string would be surrounded by quotes. The documentation makes no mention of this constraint and only provided examples with no special characters and without quotes (https://www.openhab.org/docs/configuration/persistence.html#threshold-filter).

With this PR, the parser will also recognize special characters that can be part of a unit definition when parsing the unit, removing the need to have units with special characters to be enclosed in quotes.

This has been tested with the following filter definitions:

```
Filters {
    test0 : > 5 A
    test1 : [] -100 200 °C
    test2 : [] -100 200 /°C
    test3 : ][ 100 200 kΩ
    test4a : > 5
    test4b : > 5.5 Å
    test5 : [] -100 200 '°C'
    test6 : [] -100 200 "°C"
    test7 : > 10 %
    test8 : > % 10 %
}
```

### DIFF
--- a/bundles/org.openhab.core.model.persistence/src/org/openhab/core/model/persistence/Persistence.xtext
+++ b/bundles/org.openhab.core.model.persistence/src/org/openhab/core/model/persistence/Persistence.xtext
@@ -40,27 +40,27 @@ TimeFilter:
 ;
 
 EqualsFilter:
-   '=' values+=STRING (',' values+=STRING)*
+	'=' values+=STRING (',' values+=STRING)*
 ;
 
 NotEqualsFilter:
-   '!' values+=STRING (',' values+=STRING)*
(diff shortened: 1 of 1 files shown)


kaikreuzer (approved):
Thanks!
Shall we backport this to 4.3.x?

mherwege:
> Shall we backport this to 4.3.x?

That would be nice, but there is always the work around using quotes. So I don’t think it is critical.

kaikreuzer:
It's not possible to cherry-pick it as is as it would result in merge conflicts. I'll therefore refrain from backporting it.

Issue #4470, opened 2024-12-03
Remove default persistence strategy and replace by suggested strategy


mherwege:
See discussion here: https://github.com/openhab/openhab-addons/pull/16496#issuecomment-2031662080

It would make sense not to have a default persistence strategy for a persistence service anymore, which automatically kicks in even if no items are defined. It could be replaced by a suggested strategy that is proposed when installing the service, but does not kick in automatically without confirming or changing the configuration.

As this would have been a major change, the linked PR was accepted. But this could be something to revisit for OH 5.0.

mherwege:
Partly. It would avoid this kind of behaviour by always requiring a strategy to explicitly be defined. This is a major breaking change, as the behaviour with default strategies has always been like that.

PR #4682, opened 2025-03-26
Persistence no default strategies

mherwege:
Closes https://github.com/openhab/openhab-core/issues/4470

This is a breaking change.

With this PR, the default persistence strategy is not applied automatically anymore when no other strategy is defined.

The `PersistenceService` interface still keeps the `getDefaultStrategies` method. Therefore no change is required in the persistence services implementations. But rather than applying this automatically, these  default strategies are now considered configuration suggestions.

An extra sub-endpoint to the `persistence` REST API endpoint is defined to retrieve these strategy suggestions (`GET persistence/strategysuggestions/{serviceId}`). This endpoint can be used by the UI to suggest strategies when configuring the service. Applying these suggestions will then become a clear user decision.

File based persistence configurations will have to be changed by the user if they use the default strategies.

To complete this functionality, the following is also needed:

- [ ] Breaking change notice (openhab-distro)
- [ ] Adapt the UI (openhab-web): remove default strategies and suggest when configuring, make setting a strategy required for the persistence configuration
- [ ] Upgrade logic for managed persistence configurations: if no strategies are defined, add the suggested ones to the configuration
- [ ] Change documentation

This PR should not be merged without the above also being available.

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/persistence/PersistenceResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/persistence/PersistenceResource.java
@@ -59,21 +59,19 @@ import org.openhab.core.persistence.FilterCriteria;
 import org.openhab.core.persistence.FilterCriteria.Ordering;
 import org.openhab.core.persistence.HistoricItem;
 import org.openhab.core.persistence.ModifiablePersistenceService;
-import org.openhab.core.persistence.PersistenceItemConfiguration;
 import org.openhab.core.persistence.PersistenceItemInfo;
 import org.openhab.core.persistence.PersistenceManager;
 import org.openhab.core.persistence.PersistenceService;
 import org.openhab.core.persistence.PersistenceServiceRegistry;
 import org.openhab.core.persistence.QueryablePersistenceService;
-import org.openhab.core.persistence.config.PersistenceAllConfig;
 import org.openhab.core.persistence.dto.ItemHistoryDTO;
 import org.openhab.core.persistence.dto.PersistenceServiceConfigurationDTO;
 import org.openhab.core.persistence.dto.PersistenceServiceDTO;
+import org.openhab.core.persistence.dto.PersistenceStrategyDTO;
 import org.openhab.core.persistence.registry.ManagedPersistenceServiceConfigurationProvider;
 import org.openhab.core.persistence.registry.PersistenceServiceConfiguration;
 import org.openhab.core.persistence.registry.PersistenceServiceConfigurationDTOMapper;
 import org.openhab.core.persistence.registry.PersistenceServiceConfigurationRegistry;
-import org.openhab.core.persistence.strategy.PersistenceStrategy;
 import org.openhab.core.types.State;
 import org.openhab.core.types.TypeParser;
 import org.openhab.core.types.UnDefType;
@@ -109,6 +107,7 @@ import io.swagger.v3.oas.annotations.tags.Tag;
  * @author Markus Rathgeb - Migrated to JAX-RS Whiteboard Specification
  * @author Wouter Born - Migrated to OpenAPI annotations
  * @author Mark Herwege - Implement aliases
+ * @author Mark Herwege - Make default strategy to be only a configuration suggestion
  */
 @Component
 @JaxrsResource
@@ -182,23 +181,10 @@ public class PersistenceResource implements RESTResource {
         PersistenceServiceConfiguration configuration = persistenceServiceConfigurationRegistry.get(serviceId);
         boolean editable = managedPersistenceServiceConfigurationProvider.get(serviceId) != null;
 
-        if (configuration == null) {
-            PersistenceService service = persistenceServiceRegistry.get(serviceId);
-            if (service != null) {
-                List<PersistenceStrategy> strategies = service.getDefaultStrategies();
-                List<PersistenceItemConfiguration> configs = List
-                        .of(new PersistenceItemConfiguration(List.of(new PersistenceAllConfig()), strategies, null));
-                Map<String, String> aliases = Map.of();
-                configuration = new PersistenceServiceConfiguration(serviceId, configs, aliases, strategies, strategies,
-                        List.of());
-                editable = true;
-            }
-        }
-
         if (configuration != null) {
             PersistenceServiceConfigurationDTO configurationDTO = PersistenceServiceConfigurationDTOMapper
                     .map(configuration);
-            configurationDTO.editable = editable;
+            configurationDTO.editable = managedPersistenceServiceConfigurationProvider.get(serviceId) != null;
             return JSONResponse.createResponse(Status.OK, configurationDTO, null);
         } else {
             return Response.status(Status.NOT_FOUND).build();
@@ -342,6 +328,23 @@ public class PersistenceResource implements RESTResource {
                     + DateTimeType.DATE_PATTERN_WITH_TZ_AND_MS + "]", required = true) @QueryParam("time") String time,
             @Parameter(description = "The state to store.", required = true) @QueryParam("state") String value) {
         return putItemState(serviceId, itemName, value, time);
+    }
+
+    @GET
+    @RolesAllowed({ Role.ADMIN })
+    @Produces({ MediaType.APPLICATION_JSON })
+    @Path("strategysuggestions/{serviceId: [a-zA-Z0-9]+}")
+    @Operation(operationId = "getPersistenceServiceStrategySuggestions", summary = "Gets a persistence service suggested strategies.", security = {
+            @SecurityRequirement(name = "oauth2", scopes = { "admin" }) }, responses = {
+                    @ApiResponse(responseCode = "200", description = "OK", content = @Content(array = @ArraySchema(schema = @Schema(implementation = PersistenceStrategyDTO.class), uniqueItems = true))),
+                    @ApiResponse(responseCode = "404", description = "Suggested strategies not found.") })
+    public Response httpGetPersistenceServiceStrategySuggestions(@Context HttpHeaders headers,
+            @Parameter(description = "Id of the persistence service.") @PathParam("serviceId") String serviceId) {
+        PersistenceService service = persistenceServiceRegistry.get(serviceId);
+        if (service != null) {
+            return JSONResponse.createResponse(Status.OK, service.getDefaultStrategies(), null);
+        }
+        return Response.status(Status.NOT_FOUND).build();
     }
 
     private ZonedDateTime convertTime(String sTime) {
--- a/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/dto/PersistenceServiceConfigurationDTO.java
+++ b/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/dto/PersistenceServiceConfigurationDTO.java
@@ -22,13 +22,13 @@ import org.eclipse.jdt.annotation.NonNullByDefault;
  * The {@link PersistenceServiceConfigurationDTO} is used for transferring persistence service configurations
  *
  * @author Jan N. Klug - Initial contribution
+ * @author Mark Herwege - Make default strategy to be only a configuration suggestion
  */
 @NonNullByDefault
 public class PersistenceServiceConfigurationDTO {
     public String serviceId = "";
     public Collection<PersistenceItemConfigurationDTO> configs = List.of();
     public Map<String, String> aliases = Map.of();
-    public Collection<String> defaults = List.of();
     public Collection<PersistenceCronStrategyDTO> cronStrategies = List.of();
     public Collection<PersistenceFilterDTO> thresholdFilters = List.of();
     public Collection<PersistenceFilterDTO> timeFilters = List.of();
--- a/bundles/org.openhab.core.model.persistence/src/org/openhab/core/model/persistence/Persistence.xtext
+++ b/bundles/org.openhab.core.model.persistence/src/org/openhab/core/model/persistence/Persistence.xtext
@@ -6,9 +6,7 @@ generate persistence "https://openhab.org/model/Persistence"
 
 PersistenceModel:
 	{PersistenceModel}
-	'Strategies' '{' strategies+=Strategy* 
-		('default' '=' defaults+=[Strategy|ID] (',' defaults+=[Strategy|ID])*)?
-	'}'
+	('Strategies' '{' strategies+=Strategy* '}')?
 	('Filters' '{' filters+=Filter* '}')?
 	('Items' '{' configs+=PersistenceConfiguration* '}')?
 	('Aliases' '{' aliases+=AliasConfiguration* '}')?
@@ -59,9 +57,8 @@ NotIncludeFilter:
 PersistenceConfiguration:
 	items+=(AllConfig | ItemConfig | GroupConfig | ItemExcludeConfig | GroupExcludeConfig)
 	   (',' items+=(AllConfig | ItemConfig | GroupConfig | ItemExcludeConfig | GroupExcludeConfig))* 
-	((':' ('strategy' '=' strategies+=[Strategy|ID] (',' strategies+=[Strategy|ID])*)? 
+	(':' ('strategy' '=' strategies+=[Strategy|ID] (',' strategies+=[Strategy|ID])*)
 		 ('filter' '=' filters+=[Filter|ID] (',' filters+=[Filter|ID])*)?) 
-		| ';')
 ;
 
 AllConfig:
--- a/bundles/org.openhab.core.model.persistence/src/org/openhab/core/model/persistence/internal/PersistenceModelManager.java
+++ b/bundles/org.openhab.core.model.persistence/src/org/openhab/core/model/persistence/internal/PersistenceModelManager.java
@@ -116,8 +116,7 @@ public class PersistenceModelManager extends AbstractProvider<PersistenceService
                 if (model != null) {
                     PersistenceServiceConfiguration newConfiguration = new PersistenceServiceConfiguration(serviceName,
                             mapConfigs(model.getConfigs()), mapAliases(model.getAliases()),
-                            mapStrategies(model.getDefaults()), mapStrategies(model.getStrategies()),
-                            mapFilters(model.getFilters()));
+                            mapStrategies(model.getStrategies()), mapFilters(model.getFilters()));
                     PersistenceServiceConfiguration oldConfiguration = configurations.put(serviceName,
                             newConfiguration);
                     if (oldConfiguration == null) {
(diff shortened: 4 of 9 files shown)


mherwege:
@J-N-K Can you let me know what you think? If this is accepted as the direction, I could start working on the UI side of this. But I don't want to spend time if the direction is not agreed upon.

J-N-K:
This is what I suggested about two years ago (IIRC). IMO a "proposal" is much better than some magic that is applied. However, there might be an issue with the "empty strategy". It would be better (but not so easy to implement) when the persistence service is not considered "ready" up without a valid configuration. It is necessary to allow the service to start, otherwise it can't be found and as a result the suggested strategy is not present (and the service can't be configured), but it should not be considered as "available".

mherwege:
> It would be better (but not so easy to implement) when the persistence service is not considered "ready" up without a valid configuration.

Yes, I agree that would be better. But it would require implementing a life cycle for persistence services (like for things) which is beyond what I feel comfortable doing. A persistence service persisting anything is not just dependent on a strategy being available, but also items to be persisted defined. What if you delete the all item definitions from an existing configuration? Should it go back to not ready? It should. I think it makes it more complicated then needed, and just not doing anything with an empty definition, or an item definition with no strategy is acceptable I believe.

mherwege:
@holgerfriedrich I saw https://github.com/openhab/openhab-core/pull/4693. This PR is already rebased on that PR, but I still see a javadoc failure. Should this be resolved?

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/PersistenceServiceProblem.java
```diff
@@ -0,0 +1,41 @@
...
+import org.eclipse.jdt.annotation.Nullable;
+
+/**
+ * This is a representation of a persistence service configuration problem.
+ *
+ * @param reason description for the persistence configuration problem, free text, but one of the public constants can
+ *            be used to allow the consumer (e.g. UI) to tailor the message.
+ * @param serviceId persistence service
+ * @param items list of persistence item definitions
+ * @param strategy persistence strategy selected for the list of items
```
holgerfriedrich, line 26:
```suggestion
```

holgerfriedrich:
@mherwege the log of the javadoc job needs improvement. Running it locally, I see the message causing it to fail:

```
[INFO] The goal 'org.apache.maven.plugins:maven-javadoc-plugin:3.11.2:javadoc' has not been previously called for the module: 'org.openhab.core.bom:org.openhab.core.bom.compile-model:pom:5.0.0-SNAPSHOT'. Trying to invoke it...
[WARNING] Javadoc Warnings
[WARNING] openhab-core\bundles\org.openhab.core.persistence\src\main\java\org\openhab\core\persistence\PersistenceServiceProblem.java:26: warning: @param argument "strategy" is not the name of a record component.
[WARNING] * @param strategy persistence strategy selected for the list of items

```

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/PersistenceServiceProblem.java
mherwege: This is good now I believe.

J-N-K:
What is the current state here? I see 4/4 tasks completed, but has everything been done and this can be considered as "ready for review"?

mherwege:
> What is the current state here? I see 4/4 tasks completed, but has everything been done and this can be considered as "ready for review"?

Yes, I consider this ready for review, in combination with the related PR's that are also ready.

mherwege:
@J-N-K Any chance to get this across the line?

mherwege:
Rebased to account for changes in UpgradeTool.

tools/upgradetool/src/main/java/org/openhab/core/tools/internal/PersistenceUpgrader.java
```diff
@@ -0,0 +1,96 @@
...
+
+        List.copyOf(persistenceStorage.getKeys()).forEach(serviceId -> {
+            PersistenceServiceConfigurationDTO serviceConfigDTO = Objects
+                    .requireNonNull(persistenceStorage.get(serviceId));
+            Collection<String> defaults = serviceConfigDTO.defaults;
+            if (defaults != null) {
+                Collection<PersistenceItemConfigurationDTO> configs = serviceConfigDTO.configs;
+                configs.forEach(config -> {
+                    Collection<String> strategies = config.strategies;
+                    if (strategies.isEmpty()) {
```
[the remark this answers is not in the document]
mherwege: `strategies` is defined as `@NonNullByDefault`, so can never be null.

bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/dto/PersistenceServiceConfigurationDTO.java
```diff
@@ -17,18 +17,20 @@
...
  * @author Jan N. Klug - Initial contribution
+ * @author Mark Herwege - Make default strategy to be only a configuration suggestion
  */
 @NonNullByDefault
 public class PersistenceServiceConfigurationDTO {
     public String serviceId = "";
     public Collection<PersistenceItemConfigurationDTO> configs = List.of();
     public Map<String, String> aliases = Map.of();
-    public Collection<String> defaults = List.of();
+    public @Nullable Collection<String> defaults; // Kept to be able to migrate previous version storage
```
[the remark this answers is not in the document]
mherwege: Adjusted the comment, but should not be removed in a future version.

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/persistence/PersistenceResource.java
```diff
@@ -344,6 +338,80 @@ public Response httpPutPersistenceItemData(@Context HttpHeaders headers,
...
+        return Response.status(Status.NOT_FOUND).build();
+    }
+
+    @GET
+    @RolesAllowed({ Role.ADMIN })
+    @Produces({ MediaType.APPLICATION_JSON })
+    @Path("persistencehealth")
+    @Operation(operationId = "getPersistenceHealth", summary = "Gets configuration problems with persistence services.", security = {
+            @SecurityRequirement(name = "oauth2", scopes = { "admin" }) }, responses = {
+                    @ApiResponse(responseCode = "200", description = "OK", content = @Content(array = @ArraySchema(schema = @Schema(implementation = PersistenceStrategyDTO.class), uniqueItems = true))),
```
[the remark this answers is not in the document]
mherwege: Adjusted.

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/persistence/PersistenceResource.java
```diff
@@ -344,6 +338,80 @@ public Response httpPutPersistenceItemData(@Context HttpHeaders headers,
...
+                    @ApiResponse(responseCode = "404", description = "Suggested strategies not found.") })
```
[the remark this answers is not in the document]
mherwege: Adjusted.

mherwege:
@wborn Rebased and copilot suggestions considered.

kaikreuzer:
@holgerfriedrich & @wborn As you both were active here already: Did any one of you do a (partial) review and is familiar with the code changes? I think we should get this merged, @mherwege is waiting since a long long while already.
If none of you feels confident, I can try to have a look as well.
@mherwege In the hope to do it a final time: May I ask you to rebase the PR to resolve the conflicts?

mherwege:
The javadoc failure looks like related to these warnings (that's what I get when running the build locally just for `upgradetool`):

```
Warning:  Javadoc Warnings
Warning: [WARNING] warning: unknown enum constant ReferenceCardinality.MULTIPLE
Warning:  reason: class file for org.osgi.service.component.annotations.ReferenceCardinality not found
Warning: [WARNING] warning: unknown enum constant ReferencePolicy.DYNAMIC
Warning:  reason: class file for org.osgi.service.component.annotations.ReferencePolicy not found
Warning:  Note: Custom tags that could override future standard tags:  @apiNote, @implNote. To avoid potential overrides, use at least one period character (.) in custom tag names.
Warning:  Note: Custom tags that were not seen:  @apiNote, @implNote
Warning: [WARNING] 2 warnings
```

I can't figure this out. I don't see how this is related to the javadoc in the `upgradetool` classes. Help is very much appreciated.

@kaikreuzer Apart from this, rebase is done.

kaikreuzer:
> I can't figure this out. I don't see how this is related to the javadoc in the upgradetool classes. Help is very much appreciated.

I had a look at it, but also have no clue why this might be failing, pretty weird...

> @kaikreuzer Apart from this, rebase is done.

Heck, and again I was too slow and I see three files in conflict state...? 😮

mherwege:
@kaikreuzer I rebased again, but the javadoc build issue still remains.

FWIW, while the javadoc build fails also locally with the error above I don't understand, the javadoc does get created and I don't see anything wrong with it.

holgerfriedrich:
this javadoc issue is strange. If I run `mvn javadoc:javadoc` in `tools/upgradetool`, it complains about missing `osgi.service.component.annotations.ReferenceCardinality` and `org.osgi.service.component.annotations.ReferencePolicy`.
If I import it and add the dependency to pom, it succeeds (if I disable spotless). Though, spotless wants to remove the imports, even when I reference it in javadoc text.

Maybe something is missing in the imported files Javadoc seems to process. I don't have time to check this out before the milestone build.

kaikreuzer:
This did the trick: https://github.com/openhab/openhab-core/pull/4682/commits/cea82896242c290941a62c519c618d4b702b260f

lolodomo:
What will be the impact of this change ? Is a manual action required for any existing OH installation? What must be change/added to restore the persistence?

kaikreuzer:
See description above:

> Upgrade logic for managed persistence configurations: if no strategies are defined, add the suggested ones to the configuration (included in this PR)

lolodomo:
And for users relying on config files, what is required to do ? A clear guide is necessary. To be included in release notes in the breaking changes section.

Ps: I will not install a milestone if I know that it will break my persistence and I have no idea how to fix that

kaikreuzer:
@mherwege also prepared https://github.com/openhab/openhab-distro/pull/1737/changes and https://github.com/openhab/openhab-docs/pull/2478/changes. Imho that is sufficient.

lolodomo:
The pointed links do not explain what has to be changed.
If it is already not clear for me what to change, imagine a little for a lambda user!
Is it the file rrd4j.persist that must be changed? What must be added/changed in that file?

I predict a disaster if we don't communicate clearly to the users and help them to resolve the breaking change.

lolodomo:
@mherwege : please provide clear instructions for me and every user having full setup in config files.

Should be included in Kai's message announcing the new milestone.

holgerfriedrich:
@kaikreuzer your last commit removed the persistence upgrader from the tool? Was that intended?

mherwege:
> This did the trick: [cea8289](https://github.com/openhab/openhab-core/commit/cea82896242c290941a62c519c618d4b702b260f)

@kaikreuzer It looks like you lost the PersistenceUpgrader in this commit. That’s probably why it worked this time. It is not included anymore. I think it is important to have that in.

mherwege:
> @mherwege : please provide clear instructions for me and every user having full setup in config files.
> 
> Should be included in Kai's message announcing the new milestone.

Actually, I don’t think much is needed. The default persistence is only applied if no specific persistence definition exists, either in a persist file, or managed (defined through the UI). If none exists, the upgrader will create a new managed persistence definition that does the same thing as the default persistence before.
If something was already defined in a persist file, nothing will change.

kaikreuzer:
> @kaikreuzer It looks like you lost the PersistenceUpgrader in this commit. That’s probably why it worked this time. It is not included anymore. I think it is important to have that in.

Errr, shit, that definitely was not intended... Let me check.

kaikreuzer:
https://github.com/openhab/openhab-core/pull/5190

JavaDoc build still succeeds locally for me - so let's see!

kaikreuzer:
> If it is already not clear for me what to change, imagine a little for a lambda user!

I understand your concerns - but I think they are mainly triggered by the first line in the description of this PR: "This is a breaking change.". Absolutely fair that this rings alarm bells for you.

But in this case, it is actually a change where the users do not have to become active themselves. They only need to behave differently in future (i.e. do not rely on a default strategy, which might anyhow caught many people by surprise, which is now not the case anymore).

I hope you are feeling better about it with @mherwege's explanation above - I didn't want to ignore your concerns, it was just the time pressure that had let me do the merge before properly responding to you; sorry for that.

lolodomo:
> If something was already defined in a persist file, nothing will change.

So we have nothing to change and persistence will continue to work as before with any required action ?
In that case, that is perfect but why do we consider that PR as a breaking change then ?

mherwege:
> So we have nothing to change and persistence will continue to work as before with any required action ?
> In that case, that is perfect but why do we consider that PR as a breaking change then ?

If the upgrade tool does not run when upgrading, you do lose the default persistence. Some people just copy file definitions into a new version, assuming running the upgrade tool is not required for full file based configurations. In that case, the new managed configuration replacing the default configuration will not be created.

And the main reason to call this a breaking change is that you now alway need to define items and strategies to persist anything when installing a persistence service. Before, for many persistence services, just installing it was sufficient.

mherwege:
@lolodomo @kaikreuzer It is too long ago I worked on this.

There is a case where changes in a .persist file are required:
If you only defined a item selection in the persist file, and no strategy (relying on the default strategy), you should now add a strategy, or nothing will be persisted anymore.
This possibility was already removed from the documentation with https://github.com/openhab/openhab-docs/pull/2478.
Loading he persist file will now fail without strategy defined.

Note that exactly this type of configuration lead to problems where the default kicked in before the persist file was loaded and all items got persisted for a short while.

The persistence heath check should also catch this missing strategy.

lolodomo:
Here is my .persist file. "everyMinute" is defined as strategy but not everyChange nor restoreOnStartup. Do I need to change something ?
```
// persistence strategies have a name and a definition and are referred to in the "Items" section
Strategies {
        // for rrd charts, we need a cron strategy
        everyMinute : "0 * * * * ?"
}

Items {
        GTemperature*, GSuiviConsoElec* : strategy = everyChange, everyMinute, restoreOnStartup
        GSupervision* : strategy = everyChange, everyMinute
        GVirtualData*, GParametres* : strategy = everyChange, restoreOnStartup
}
```

mherwege:
> Here is my .persist file. "everyMinute" is defined as strategy but not everyChange nor restoreOnStartup. Do I need to change something ?
> 
> ```
> // persistence strategies have a name and a definition and are referred to in the "Items" section
> Strategies {
>         // for rrd charts, we need a cron strategy
>         everyMinute : "0 * * * * ?"
> }
> 
> Items {
>         GTemperature*, GSuiviConsoElec* : strategy = everyChange, everyMinute, restoreOnStartup
>         GSupervision* : strategy = everyChange, everyMinute
>         GVirtualData*, GParametres* : strategy = everyChange, restoreOnStartup
> }
> ```

No, because you have strategy = … for each Items definition. The problem is when there is no strategy = for the Items definitions.

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/persistence/PersistenceResource.java
```diff
@@ -345,6 +339,79 @@ public Response httpPutPersistenceItemData(@Context HttpHeaders headers,
...
+        if (service != null) {
+            return JSONResponse.createResponse(Status.OK, service.getSuggestedStrategies(), null);
+        }
+        return Response.status(Status.NOT_FOUND).build();
+    }
+
+    @GET
+    @RolesAllowed({ Role.ADMIN })
+    @Produces({ MediaType.APPLICATION_JSON })
+    @Path("persistencehealth")
```
florian-h05, line 362: The semantics health endpoint is `/rest/semantics/health`.
`/rest/persistence/persistencehealth` also "sounds" a bit awkward to me.

Do we want to change to `/rest/persistence/health` for consistency?

mherwege: Yes, I would be fine with that.

florian-h05: @kaikreuzer Can we change that now in the feature freeze? It would be better to have it changed before the new REST endpoints are contained in a release.

kaikreuzer: @florian-h05 Yes, I agree that we should change it before the release.

helmar74:
It would be intressting what to do to get mapd persistence fixed in a filebased configuration…
Thanks ahead. Currently running OH 5.1.0 M4 with changed mapdb configuration.

For the moment, I commented out “default” in Strategies section:
```
Strategies{
 jedeMinute : "0 * * * * ?"
 jedeStunde : "20 0 * * * ?"
 // Geändert (auskommentiert) 15.12.2025 wegen fehlendem Restore beim startup
 //default = everyChange, restoreOnStartup
}
Items{
 * : strategy = everyChange, restoreOnStartup
.
.
.
}
```

Is this correct and everything I have to do?

lolodomo:
@mherwege and @kaikreuzer : maybe examples of what must be changed should be included in the breaking changes of our coming 5.1 release notes ?
As I imagined, many users will be lost and wil ask what to do. I was myself one of them last Sunday.

helmar74:
> @mherwege and @kaikreuzer : maybe examples of what must be changed should be included in the breaking changes of our coming 5.1 release notes ? As I imagined, many users will be lost and wil ask what to do. I was myself one of them last Sunday.

would be good to know for RC1 as well, for testing reasons

lolodomo:
@helmar74 : the alert message was already enhanced, look at this:
https://github.com/openhab/openhab-distro/pull/1841

lolodomo:
@mherwege : would it be possible for you to update this chapter in our documentation to explain that "default" strategy is now forbidden?
https://www.openhab.org/docs/configuration/persistence.html#strategies

mherwege:
> @mherwege : would it be possible for you to update this chapter in our documentation to explain that "default" strategy is now forbidden? https://www.openhab.org/docs/configuration/persistence.html#strategies

@lolodomo See https://github.com/openhab/openhab-docs/pull/2613

helmar74:
Could someone hav alook in my comments here:
https://community.openhab.org/t/openhab-5-1-milestone-discussion/166385/217

I think there'S a critical issue with persistence settings in RC1
Looks like filebased configuration is not used anymore

PR #4686, opened 2025-03-28
OAuth prevent concurrent token read / refresh

andrewfg:

### DIFF
--- a/bundles/org.openhab.core.auth.oauth2client/src/main/java/org/openhab/core/auth/oauth2client/internal/OAuthClientServiceImpl.java
+++ b/bundles/org.openhab.core.auth.oauth2client/src/main/java/org/openhab/core/auth/oauth2client/internal/OAuthClientServiceImpl.java
@@ -291,7 +291,7 @@ public class OAuthClientServiceImpl implements OAuthClientService {
     }
 
     @Override
-    public AccessTokenResponse refreshToken() throws OAuthException, IOException, OAuthResponseException {
+    public synchronized AccessTokenResponse refreshToken() throws OAuthException, IOException, OAuthResponseException {
         if (isClosed()) {
             throw new OAuthException(EXCEPTION_MESSAGE_CLOSED);
         }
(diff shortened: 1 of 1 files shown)


andrewfg:
@lsiepel / @jlaur re https://github.com/openhab/openhab-addons/pull/18450

lsiepel (approved):
Thanks, LGTM.

Should be backported to 4.3.x

bundles/org.openhab.core.auth.oauth2client/src/main/java/org/openhab/core/auth/oauth2client/internal/OAuthClientServiceImpl.java
```diff
@@ -333,7 +333,7 @@ public AccessTokenResponse refreshToken() throws OAuthException, IOException, OA
     }
 
     @Override
-    public @Nullable AccessTokenResponse getAccessTokenResponse()
+    public synchronized @Nullable AccessTokenResponse getAccessTokenResponse()
```
jlaur, line 336: Is this really needed when having `refreshToken()` synchronized?

andrewfg: > Is this really needed when having refreshToken() synchronized?

Yes I think do. If we just rely on the refreshToken sync, I can imagine that a second call to this method might still retrieve the old token before refreshToken has completed and stored the new one .. and then it would have to call refreshToken again..

lsiepel: Agree with @andrewfg

andrewfg:
@kaikreuzer thank you for this. Do you intend to back port into 4.3.4? If not then I need to doing something else at binding level for 4.3.x users..

lsiepel:
Please backport. That would be best.

kaikreuzer:
Backported!

andrewfg:
@kaikreuzer / @lsiepel I fear that I may have introduced a regression with this one. It seems that some users are experiencing thread locks (see forum thread below). I am not sure if the problem is due to the Tado auth server blocking on the refresh token call, or if there is a cross locking issue due to the added `synchronize` keywords. So I wonder if you could kindly look at my code and give me your thoughts. If the issue is indeed due to the Tado auth server blocking, then I guess we at least need to add a timeout on the Jetty client within OH Core OH.

https://community.openhab.org/t/many-items-stop-responding-after-a-while-since-upgrade-to-oh-4-3-4/163498/1

andrewfg:
Forked to new issue #4698

PR #4687, opened 2025-03-28
Speedup Build

holgerfriedrich:
* GitHub Action for build and test is split up:
  * build: spotless, build, test, itest
  * sat: build, sat (checkstyle, pmd, spotbugs)
  * javadoc: build javadoc to make sure it does not break

This reduces the build time from ~18min to ~13min, using 3 runners in parallel.
The initial idea was to split SAT from compilation. This works for checkstyle and PMD, but not for spotbugs (which requires compilation). So, compilation is done on both runners, but with different focus on checks and tests.

JavaDoc is new, and may prevent that we break core javadoc build from time to time.

Remark: Right now, SAT is the slowest part. Any idea what to save on this one, would speed up the build.
Skipping build completely with maven.main.skip and disabling the compilation of the generated code via profile did not work due to spotbugs.

I have played around with the number of cores and VM size for SAT, saving another 30 secs.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -113,7 +113,7 @@ jobs:
     timeout-minutes: 30
     env:
       MAVEN_OPTS: >-
-        -Xmx2g
+        -Xmx4g
         -Dmaven.wagon.http.retryHandler.count=5
         -Dmaven.wagon.httpconnectionManager.ttlSeconds=25
         -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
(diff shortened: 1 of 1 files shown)


holgerfriedrich:
@lsiepel this does not work well for addons repo, as spotbugs is the limiting factor. This takes 54min (compared to 1h5 we have now). Another solution would be needed there, eg. splitting the build in two parts which are done an different runners.

lsiepel:
> @lsiepel this does not work well for addons repo, as spotbugs is the limiting factor. Spotbugs needs class files, as it works on the byte code. This takes 54min (compared to 1h5 we have now). Another solution would be needed there, eg. splitting the build in two parts which are done an different runners.

Too bad, was hoping this would work for adding too.

kaikreuzer (approved):
Cool, thanks!

PR #4693, opened 2025-04-02
Fix JavaDoc build

holgerfriedrich:
JavaDoc build recently introduced recently (#4687) seems to be broken in certain cases.
xtend code generation and *compilation* is done- even though javadoc operates on java files.

The model seems to change, and `mvn javadoc:javadoc javadoc:aggregate` fails on CI.

Putting in `install` before javadoc solves the issue.
Skipping unnecessary steps keeps it still faster than the SAT build.


wborn (approved):
Thanks for the fix!

PR #4695, opened 2025-04-03
Semantic Tags: Property tags added

andrewfg:
Added Property tags as discussed in https://github.com/openhab/openhab-core/issues/4669

Makes #4688 and #4689 obsolete

Required by https://github.com/openhab/openhab-addons/pull/18483

Signed-off-by: Andrew Fiddian-Green <software@whitebear.ch>

PR #4694, opened 2025-04-03
Semantic Tags: Added EQUIPMENT tags

andrewfg:
Resolves #4617

Signed-off-by: Andrew Fiddian-Green <software@whitebear.ch>

Issue #4672, opened 2025-03-22
Semantic Tags: Delete duplicate/wrong POINT tags


andrewfg:
This thread lists POINT tags that should potentially be deleted; either because they are duplicates, or because they don't comply with the OH tagging rules.

andrewfg:
The following are POINT tags that are proposed to be deleted

- `LowBattery` - this should be relocated as `Alarm` with a PROPERTY context of `Voltage` or `Power`
- `OpenLevel` - this should be relocated as `Control` / `Setpoint` / `Measurement` with a PROPERTY context of `Position`
- `OpenState` - ditto as above
- `Tilt` - this should be relocated as `Control` / `Setpoint` / `Measurement` with a PROPERTY context of `Tilt`

andrewfg:
Note apropos `Tilt` : this has already been moved from Point => Property

andrewfg:
I think (hopefully) we have reached consensus on Point tags. So shall we close this?

PR #4700, opened 2025-04-05
Sort SemanticTags.csv and group children under their parent tag

jimtng:
Follow up to #4694 and #4695

- Sort SemanticTags.csv
- Sort the tags in the xsd alphabetically
- Remove Point_Sensor_Tilt to avoid duplicate tags, because a Property_Tilt was added in #4695


jimtng:
@andrewfg

andrewfg:
@jimtng Wow. That looks good. But I have a question/clarification: Obviously in this PR you ran the groovy script offline to produce the generated source files. So I am wondering if we can add an execution goal to the OH Maven build instructions so that this is done automatically during the `generate-sources` phase of the Maven [lifecycle](https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html)?

jimtng:
> So I am wondering if we can add an execution goal to the OH Maven build instructions so that this is done automatically during the `generate-sources` phase of the Maven [lifecycle](https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html)?

That is an idea. Perhaps post it as an issue, and someone who's familiar with the build system can comment?

andrewfg:
Umm. Actually it looks as though the auto-generate is already part of the Maven build configuration:

https://github.com/openhab/openhab-core/blob/693943e7c1f7d96ed882ddfaa08083f331452561/bundles/org.openhab.core.semantics/pom.xml#L64

jimtng:
I added 2 extra commits that:
- sorts the tags in the xsd alphabetically
- sorts the tags within each type alphabetically in DefaultSemanticTagProvider.java

I think the DefaultSemanticTags.java should remain sorted hierarchically because the hierarchy is visible in there.

jimtng:
> Actually it looks as though the auto-generate is already part of the Maven build configuration.
> 
> And if that is indeed the case, then you can remove all but the modified groovy script file from this PR.

If it does get generated automatically, then it doesn't matter whether they are included in this PR. However it shows the reviewers the result of the script, and that the generated files will still build (although if it generated during build, then this is irrelevant).

andrewfg:
Indeed.

bundles/org.openhab.core.semantics/src/main/java/org/openhab/core/semantics/model/DefaultSemanticTagProvider.java
```diff
@@ -39,59 +39,56 @@ public DefaultSemanticTagProvider() {
         defaultTags.add(DefaultSemanticTags.LOCATION);
```
andrewfg, line 39: The topmost tags are sorted Equipment => Location => Point =>, Property.

But by contrast your sort algorithm seems to be Location => Property => Point => Equipment.

Personally I suggest that both should be sorted Location => Point => Property => Equipment.

andrewfg (changes requested):

bundles/org.openhab.core.semantics/src/main/java/org/openhab/core/semantics/model/DefaultSemanticTagProvider.java
jimtng: Yeah this was the original ordering. I'll adjust

jimtng:
@andrewfg unrelated to this PR, but it seems that there are duplicate tags:

`Property_Tilt`
`Point_Status_Tilt`

Seems to have been added here: https://github.com/openhab/openhab-core/pull/4695

Should we remove the new Property, or the original Point_Status_Tilt?

jimtng:
I found out that DefaultSemanticTagProvider.java must be sorted hierarchically. At runtime the parent tags must have already been added before the children.

andrewfg:
> Property_Tilt
> Point_Status_Tilt

@jimtng you just removed the wrong one. The correct application is

- Point=Control + Property=Tilt, or
- Point=Status + Property=Tilt

A tag like Point_Status_Tilt is a total munge of Point versus Property. If anything that is the one that shall be removed. However that is under discussion in #4672. _Therefore I think you should not remove either of the tags here in this PR._

jimtng:
> Umm. Actually it looks as though the auto-generate is already part of the Maven build configuration.
> 
> And if that is indeed the case, then you can remove all but the modified groovy script file from this PR

Come to think about this further, the generated files still need to be included in the PR so that they get checked in to git, unless there is another process that actually commits them into the git repo. Otherwise, people would keep getting modified files when they check out and run a build process, because they aren't in .gitignore.

jimtng:
> However that is under discussion in #4672. _Therefore I think you should not remove either of the tags here in this PR._

I have removed the new Property Tilt. 

Had I not done this, build / tests will fail. So perhaps take note that this was removed again here, so the future PR that would change Tilt from Point to Property can do so by removing the Point at the same time.

I will add some code to check for duplicate tags so this situation can be avoided in the future, provided that changes to SemanticTags.csv will always need to be accompanied with running the tags generator (as done in this PR)

andrewfg:
> because they aren't in .gitignore.

Probably they should be. I think that is so for other cases where code-gen is used. I believe the convention is to save such files in the `target/generated-sources` directory rather than in `src/main`, the pom is adapted so it can pull in those file dependencies. So I think you would have the original SemanticTags.csv under `src/main` and the generated/sorted version of it under `target/generated-sources`. I am not an expert on this, so perhaps someone else can advise.

andrewfg:
> I have removed the new Property Tilt.
> Had I not done this, build / tests will fail. 
> So perhaps take note that this was removed again here, so the future PR that would change Tilt from Point to Property can do so by removing the Point at the same time.

I cannot accept that! I expressly asked to add the correct Tilt property as I have a backlog of binding pending tagging which require just that. If you remove it then I cannot approve this PR!

Just for info the boschhsc binding is the only one to use the tilt tag (see below) .. and as you can clearly see, it is used in the context of a Property rather than that of a Point. Therefore IMHO you should delete to Point Tilt and re-add the Property-Tilt.

```
		<tags>
			<tag>Alarm</tag>
			<tag>Tilt</tag>
		</tags>
```

jimtng:
> Probably they should be. I think that is so for other cases where code-gen is used. I believe the convention is to save such files in the `target/generated-sources` directory rather than in `src/main`, the pom is adapted so it can pull in those file dependencies. So I think you would have the original SemanticTags.csv under `src/main` and the generated/sorted version of it under `target/generated-sources`.

This is best kept as a separate issue.

andrewfg:
Apropos tilt, it is already listed as a candidate for removal, (see below), so I would remove it now.

https://github.com/openhab/openhab-core/issues/4672#issuecomment-2745342975

andrewfg:
Apropos code-gen see #4701

jimtng:
I have restored Tilt Property and removed Tilt Point

andrewfg (approved):
LGTM thank you.

kaikreuzer (approved):

jlaur:
Should the XSD also be uploaded to the website through openhab/website? See for example openhab/website#462.

andrewfg:
> Should the XSD also be uploaded to the website through openhab/website?

Yes absolutely. The file on the website is the one that is actually used by maven for xml xsd validation. We have been holding off updating the website version in order to avoid breaking CI builds. But I think everything is moderately stable now so it's probably ready now for the upload.

andrewfg:
@jlaur see https://github.com/openhab/website/issues/510

PR #4702, opened 2025-04-05
Check for duplicate tags in SemanticTags.csv

jimtng:
This will currently fail because there's a duplicate tag. The duplicate is removed in #4700, so once that's merged, the test in this PR should pass.

Currently duplicate tags would've caused some tests to fail, but the reason for the failure isn't clear without digging into surefire-reports.


jimtng:
FYI @andrewfg

andrewfg (approved):
LGTM

kaikreuzer (approved):

PR #4706, opened 2025-04-07
Semantic Tags: Update system channel tags to match latest schema

andrewfg:
Signed-off-by: Andrew Fiddian-Green <software@whitebear.ch>

Issue #4668, opened 2025-03-22
Semantic Tags: Add missing POINT tags


andrewfg:
This thread lists POINT tags that should potentially be added

ccutrer:
Honestly, `Control` (with `Switch` subclass), `SetPoint`, `Measurement`, and `Status` cover just about everything for Point for me. I view different Point types as a way to distinguish between multiple of the same Property on an Equipment, based on their intended usage. My only possible suggestion is a "Desired", "Target", or "Call" Point, as in the following example of a thermostat:
* 
* Thermostat (Thermostat)
  * Ambient temperature (Measurement, Temperature)
  * Set point (SetPoint, Temperature) - many thermostats have both a heating and a cooling set point for some sort of auto mode. Do we want to be able to distinguish between the two as a subclass of SetPoint?
  * Set Mode (Control, Mode) - i.e. is it supposed to be in heating, cooling, drying, auto, off, etc.
  * Current Mode (Status, Mode) - i.e. what it's currently doing - heating, cooling, drying, idle, off. In particular, this would never have "auto" mode, because auto isn't something it can actually do. And even when in heating set mode, it could be idle, or blower only, or something, because it's reached the set point. For multi-stage/multi-speed system, the speed of the compressor, or use of auxiliary heat, may or may not be exposed in the same point. I don't think it's particularly concerning which way it's modeled. For my main heat pump, it's a separate "aux heat" status. But interestingly, my primary zone (and _only_ my primary zone) has a set mode called "EHeat", which you can use to force aux heat on, even if the heat pump itself is not working for some reason.
  * Thermostat Call (???) - I have no idea how to classify this. It's basically what the thermostat _wants_ to do, given the set mode, but it may not be in that state yet. For example, a multi-zone unit might have one zone set to heat, and another set to cool. It cannot honor both at once. So one thermostat will be in heating set mode, heating current mode, and heating stage 1 call. The other thermostat would be in cooling set mode, idle current mode, and cooling stage 2 call. Another reason current and call can differ is if a compressor cannot cycle on for some period of time after it shut off for some reason. Another is if there's some sort of error lockout, preventing it from honoring the call. Right now I've modeled this as a new Point type called Desired, with a property of Mode. Not a lot of equipment exposes this level of detail to home automations systems, so I don't know how generally useful it is to even be able to define this. If others have suggestions of a better way to define this, or just think it should be labeled as a Status Point, or think it's too complicated, by all means, let's ignore this.

[...]

PR #4708, opened 2025-04-08
Semantic Tags: Tweaks to tags

andrewfg:
This PR contains some tweaks to Point and Property tags arising from #4706 and #4668


jimtng:
- No mediaplayer?
- I think you also need to run the generate-tag-classes plugin, whenever semantictags.csv is changed, and include the generated files in the same PR. That's how I usually do it.
`mvn clean install -Pgenerate-tag-classes -pl :org.openhab.core.semantics`

andrewfg:
@jimtng I have run the script

jimtng:
What would `OPENING` be for that isn't covered by OPEN_LEVEL and OPEN_STATE?

bundles/org.openhab.core.semantics/model/SemanticTags.csv
```diff
@@ -114,7 +114,9 @@ Equipment,Bed,,Bed,,
 Equipment,Blinds,,Blinds,"Rollershutter, Rollershutters, Roller shutter, Roller shutters, Shutter, Shutters",
 Equipment,Camera,,Camera,Cameras,
 Equipment,CleaningRobot,,Cleaning Robot,"Cleaning Robots, Vacuum robot, Vacuum robots",
+Equipment,CloudService,,Cloud Service,,
 Equipment,Computer,,Computer,,
+Equipment,Contact,,Contact,"Door Contact, Window Contact",
```
jimtng, line 119: I think that `Contact` should be a Point, not Equipment? The equipment would be `Door` and `Window`.

andrewfg: > The equipment would be Door and Window

It is a magnetic contact device that could be fitted to a door or a window

andrewfg: It is a separate magnet contact device that you can attach to a door or a window

andrewfg:
> What would OPENING be for that isn't covered by OPEN_LEVEL and OPEN_STATE?

Indeed. And we also have OpenClosed too. I think there should probably be just OpenState (binary) and OpenLevel (analog) that are children of Opening. => WDYT?

andrewfg:
> deleting it shouldn't be a problem?
Indeed.

bundles/org.openhab.core.semantics/model/SemanticTags.csv
lsiepel: The magnetic contact device can be applied to so many types of equipment, but is meaningless on its own. 
`Contact` is not a device / equipment. It is a state / property.

andrewfg: But if you are really so confused by the concept of `Equipment.CONTACT` as a piece of equipment we can call it  `Equipment.CONTACT_SENSOR` or `Equipment.CONTACT_DEVICE` instead, to make it clearer.

lsiepel: `CONTACT_SESNOR` would be a good start. Could you point me to the comment / PR part about this rule you refer to. I need to catch up and have trouble finding the right posts, thanks!

andrewfg: https://next.openhab.org/docs/developer/bindings/semantic-tags.html

bundles/org.openhab.core.semantics/model/SemanticTags.csv
```diff
@@ -59,25 +56,32 @@ Property,Airflow,,Airflow,,
...
 Property,Channel,,Channel,,
+Property,Color,,Color,,
 Property,ColorTemperature,,Color Temperature,,
 Property,Current,,Current,,
 Property,Duration,,Duration,,
 Property,Energy,,Energy,,
 Property,Frequency,,Frequency,,
 Property,Gas,,Gas,,
 Property,Heating,,Heating,,
+Property,HomeAway,,Home Away,,
```
jimtng, line 67:
```suggestion
Property,HomeOccupancy,,Home Occupancy,,
```

This can be a Switch (At home, vs not at home), or a Number Item (how many people are at home).
But then can't we just use the existing `Presence` ?

jimtng (commented):
I think `Motion` property is probably needed. We can apply Motion to PIR sensors, and Presence to mmwave sensors.

Another thing, I noticed `Equipment_TemperatureSensor`. Most of the time, such a device comes with Temperature, Humidity, and Barometric Pressure sensors, so naming it just TemperatureSensor seems like a misnomer. I haven't figured out a good name for it. Some candidates may be: EnvironmentalSensors, ClimateSensors, WeatherStation, etc.

andrewfg:
I think Equipment.MultiSensor is probably the best name. Or?

andrewfg:
> But then can't we just use the existing Presence ?

Umm. @lsiepel asked me not to use Presence .. since one is a measurement and the other a command.

andrewfg:
But in retrospect I think Presence is really the correct one. It can be either be switched by geosensing people's phones or by pir or other physical sensors. @lsiepel do you agree?

bundles/org.openhab.core.semantics/model/SemanticTags.csv
lsiepel: Thanks. `CONTACT_SENSOR` would be a good one to distinguish between the two.

bundles/org.openhab.core.semantics/model/SemanticTags.csv
andrewfg: https://github.com/openhab/openhab-addons/pull/18514#discussion_r2038079146

lsiepel:
> > But then can't we just use the existing Presence ?
> 
> Umm. @lsiepel asked me not to use Presence .. since one is a measurement and the other a command.

That was in a different context. Presence is perfectly fine as property. Like Measurement + Presence.

andrewfg:
> Presence is perfectly fine as property. Like Measurement + Presence.

We are talking about it as a property like Switch + Presence. (My Siemens thermostats can be switched to Presence mode either a) manually via the app, b) via geo sensing phone, c) through an external command from OH, or d) via a built in PIR. In all cases it is the same OH Channel involved.

andrewfg:
> Too generic

Ok please make a counter proposal. 4yi Hue sensors have motion, illuminance, temperature, / weather sensors have rain, sun, temperature, rh, zwave sensors have temperature and contact. So you need something .. well .. generic

lsiepel:
> We are talking about it as a property like Switch + Presence. (My Siemens thermostats can be switched to Presence mode either a) manually via the app, b) via geo sensing phone, c) through an external command from OH, or d) via a built in PIR. In all cases it is the same OH Channel involved.

Let's discuss this in the other PR.

bundles/org.openhab.core.semantics/model/SemanticTags.csv
```diff
@@ -97,6 +102,7 @@ Property,Vibration,,Vibration,,
 Property,Voltage,,Voltage,,
 Property,Water,,Water,,
 Property,Wind,,Wind,,
+Equipment,AlarmDevice,,Alarm Device,,
```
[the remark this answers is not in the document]
andrewfg: An alarm system has a panel with multiple alarm devices (inputs) and a siren (output) connected to it.

andrewfg:
> we can just use Sensor instead of adding another Equipment type, wdyt?

That is what I did already in the case of Hue. => See `sensorCount` in the code..

https://github.com/openhab/openhab-addons/blob/800f381ef9e47eee94061447cc32ff5b112cfc7b/bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/handler/Clip2ThingHandler.java#L1427

bundles/org.openhab.core.semantics/model/SemanticTags.csv
```diff
@@ -194,16 +201,18 @@ Equipment,AirQualitySensor,Sensor,Air Quality Sensor,,
 Equipment,CO2Sensor,Sensor,CO2 Sensor,,
 Equipment,COSensor,Sensor,CO Sensor,,
 Equipment,ElectricMeter,Sensor,Electric Meter,,
+Equipment,FireDetector,Sensor,Smoke Detector,Smoke Detectors,
```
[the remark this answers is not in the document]
andrewfg: The following are (some of) the possible fire detectors..

- Smoke detector (obscuration or ionisation)
- Heat detector (thermal)
- Flame detector (optical u/v)

I guess FireDetector should be the parent of them all.

Note there are also manual break glass switches.

bundles/org.openhab.core.semantics/model/SemanticTags.csv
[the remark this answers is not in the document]
jimtng: I see its use in https://github.com/openhab/openhab-addons/blob/800f381ef9e47eee94061447cc32ff5b112cfc7b/bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/handler/Clip2ThingHandler.java#L1445-L1446

Shouldn't that just be set to Equipment.ContactSensor?

bundles/org.openhab.core.semantics/model/SemanticTags.csv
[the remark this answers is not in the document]
jimtng: Perhaps remove all the sub-types and just have `FireDetector`?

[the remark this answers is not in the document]
andrewfg: > remove all the sub-types 

Umm that would be a breaking change. Better to add the other sub types I think.

jimtng:
@andrewfg it seems that in the last changes, Television and Projector lost its parent of `Screen`. Could we add it back please?

cc: @ccutrer

jimtng:
Also please set Blinds' parent to WindowCovering

andrewfg:
> Television and Projector lost its parent of Screen.

We added a level AudioVisual. Screen and Tv became children of that. But indeed projector should also be under AV. @jimtng do you consider Projector and TV as direct children of AV or sub children of Screen?

> Blinds

Yes. Weird but I will add it back.

jimtng:
```
Equipment_AudioVisual_Screen_Television
Equipment_AudioVisual_Screen_Projector
```

jimtng:
Put `Speaker` under AudioVisual

bundles/org.openhab.core.semantics/model/SemanticTags.csv
```diff
@@ -185,31 +193,33 @@ Equipment,UPS,PowerSupply,UPS,Uninterruptible Power Supply,
...
 Equipment,Printer3D,Printer,3D Printer,,
-Equipment,Projector,,Projector,"Projectors, Beamer, Beamers",
 Equipment,Pump,,Pump,Pumps,
 Equipment,WaterFeature,Pump,Water Feature,"Waterfall, Pond Pump",
 Equipment,RemoteControl,,Remote Control,Remote Controls,
 Equipment,Sensor,,Sensor,Sensors,
 Equipment,AirQualitySensor,Sensor,Air Quality Sensor,,
 Equipment,CO2Sensor,Sensor,CO2 Sensor,,
 Equipment,COSensor,Sensor,CO Sensor,,
+Equipment,ContactSensor,Sensor,Contact Sensor,"Door Contact, Window Contact",
```
jimtng, line 203: I think we should drop the synonyms, so that if someone wants to add custom Door Contact / Window Contact, they won't conflict with the synonyms here.
```suggestion
Equipment,ContactSensor,Sensor,Contact Sensor,,
```

jimtng (changes requested):
Unrelated to the changes here, I think having "AutoManual", "EnabledDisabled", "OnOff" is inconsistent with the rest, and I think they should be avoided. The equivalent would be OpenClosed (which we removed), LowHighBattery, EmptyFull, DarkBright, WetDry, UpDown, etc etc.

If we can't come up with a good name, perhaps we should just use `Mode` for them, and Status/`Power` for OnOff.

andrewfg:
> If we can't come up with a good name

The only name I can think of is `Binary` ..

bundles/org.openhab.core.semantics/model/SemanticTags.csv
andrewfg: Ok I will delete the synonyms. But lets wait for any final inputs apropos your other comment..

jimtng:
Binary doesn't tell you what _property_ it holds. Only the kind of possible values (binary vs analog/linear). This seems to violate the concept of "property".

andrewfg:
> Binary doesn't tell you what property it holds

Ok. Another suggestion would be `TwoState` .. but after that I am exhausted. Please also so the edits to my post above.

jimtng:
> We have `OpenState` => Should we delete it?

No, OpenState is good. OpenClosed isn't.



> We have `LowBattery` => We should not delete it.

This is unfortunately a legacy tag, although since we changed it to Property, it may cause some problems. In any case, I'm OK with LowBattery, but not "LowFullBattery". 

LowBattery can be assigned to a SwitchItem, when it's ON, it's low battery.

But if it was LowFullBattery, it's just awkward.

In general I'm trying to say that we should not have "OnestateOppositestate" format.

jimtng:
TwoState is no different to Binary / OpenClosed / HighLow / AutoManual / EnabledDisabled. They describe the possible values, but not what the values _mean_ (ie. what property they describe).

As I said, instead of AutoManual, the Item should be tagged with "Mode", and instead of EnabledDisabled, it should be tagged "State" (or Status), Instead of OnOff, it should be tagged Power, and so on.

andrewfg:
> we should not have "OnestateOppositestate" format

So lets change these three as follows..
- `OnOff` => `Power` (as in `Switch + Power` or `Status + Power`)
- `AutoManual` => `Mode`
- `EnabledDisabled` => `Enabled` (it's main use is for sensors which can be disabled)

Note: these three are new to this PR so changing or deleting them would only impact my own PRs. So let us make a quick decision now..

jimtng:
> So lets change these three as follows..
> 
> * `OnOff` => `Power` (as in `Switch + Power` or `Status + Power`)
> * `AutoManual` => `Mode`
> * `EnabledDisabled` => `Enabled` (it's main use is for sensors which can be disabled)

Yes, and Enabled shouldn't be a child of Mode. I think it should just stand alone.

jimtng (approved):
Thanks! LGTM

andrewfg:
I have uploaded the as per LGTM schema file to https://github.com/openhab/website/pull/511

ccutrer:
> `Television` and `Projector` lost its parent of `Screen`.

It depends on what me mean by screen. If it's "device that generates images", Projector and Television are both subclasses. If it's "the physical thing that images are displayed on", they are not. The difference is that a Television is both types of screens, but with a projector, you have the image generating device (the projector), and the screen the image is on. I don't think one interoperation is necessarily more correct than others (being able to automate the screen itself in a projector+screen setup is fairly common - it can roll out of the ceiling or other cavity, or have masks that are controllable). It was just previously Screen was the parent of Television and Projector, and I had rules that relied on that (when a receiver turns on, turn on the screen in that room, regardless of if it's a television or projector). I also have a motorized screen, which I just don't semantically tag (I have no reason to directly expose it in a UI, or need to generically find it and its properties in rules, because it's 100% controlled by dedicated automation rules).

> * `OnOff` => `Power` (as in `Switch + Power` or `Status + Power`)

I'm not sure "Power" is the right word here. The Power property refers to how much power (generally in watts) a device is using. Though maybe it's okay to overload it, since such a property would only ever be paired with the Measurement point, leaving Switch,Power to mean the point that can turn a device on and off, and Status,Power to mean a point you can't control, but indicates if the device is active at the moment based on some other means (such as a hot water heater that is always "on", but is now "heating" because it is below the setpoint temperature.

ccutrer:
Any objection to Display being the parent of Television and Projector, and Screen being separate (with a description of Projection Screen, so it's clear that's what is intended)?

andrewfg:
@ccutrer / @jimtng just to summarise: I need to have a unanimous proposal from you both on two topics..

1. the parent child hierarchy of tv, screen, projector, etc.
2. the property to use for on state

I have been dancing back and forwards on various conflicting proposals, it looks like I am wasting my time, so you two please make the decision. Fast.

ccutrer:
1. Just waiting on @jimtng to approve my latest proposal (Display -> (Television, Projector), Screen)
2. I've convinced myself I'm +1 on the current proposal of using Power for on state

jimtng:
> 1. Just waiting on @jimtng to approve my latest proposal (Display -> (Television, Projector), Screen)
> 2. I've convinced myself I'm +1 on the current proposal of using Power for on state

Yes I agree on Display for the parent of Television and Projector, and add Screen separately. This is good!
Should Screen also be a child of AudioVisual?

ccutrer:
> Should Screen also be a child of AudioVisual?

Yes

holgerfriedrich:
Sorry for joining the party late. I have not been involved in previous discussions about semantics, and would leave it to the other maintainers to review.

>> OnOff => Power (as in Switch + Power or Status + Power)
I'm not sure "Power" is the right word here. > The Power property refers to how much power (generally in watts) a device is using.

This confuses me.
I would interpret power in the same way as stated above, because this is a common term in physics.

In addition, there might be logical switches. I am not sure if power(ed) is the correct term for those cases. Or would you see this as a mode?

jimtng:
I understand that Power is used in physics and engineering with the unit of Watt. It is used to measure the rate of energy per unit time.

Power is also a very common terminology in consumer electronics to label a button which is used to turn something on or off.

Here is a photo of my power supply and a bench multimeter.

jimtng:
I understand that Power is used in physics and engineering with the unit of Watt. It is used to measure the rate of energy per unit time.

Power is also a very common terminology in consumer electronics to label a button which is used to turn something on or off.

Here is a photo of an appliance with a power button
image



Many devices / API uses `POWER` to turn it on / off, e.g. [Tasmota MQTT](https://tasmota.github.io/docs/Commands/), See how it's implemented in openhab's mqtt:
https://community.openhab.org/t/help-with-a-simple-mqtt-on-off-to-sonoff-tasmota-solved/98811/5

andrewfg:
Apropos Switch Power .. there are still channels where Power is not appropriate e.g. the Zwave binary sensor (where, as you see, I simply don’t declare a Property at all)..

https://github.com/andrewfg/org.openhab.binding.zwave/commit/174bb85f91f4ec694b95d6df3ddc08ebdab58f4b#diff-d8b76539195caa467bce060c1a64ebcad9d02640002ab809b761e11e102c52d0R1233

jimtng:
> Apropos Switch Power .. there are still channels where Power is not appropriate e.g. the Zwave binary sensor (where, as you see, I simply don’t declare a Property at all)..
> 
> [andrewfg/org.openhab.binding.zwave@174bb85#diff-d8b76539195caa467bce060c1a64ebcad9d02640002ab809b761e11e102c52d0R1233](https://github.com/andrewfg/org.openhab.binding.zwave/commit/174bb85f91f4ec694b95d6df3ddc08ebdab58f4b#diff-d8b76539195caa467bce060c1a64ebcad9d02640002ab809b761e11e102c52d0R1233)

For such a generic sensor, we don't need to assign any property at all, and leave it to the user to decide what that sensor _actually_ represents for them.

andrewfg:
> For such a generic sensor, we don't need to assign any property at all

Indeed.

And we will not delete the Power property from the set. It is a separate discussion in other PRs in addons about whether its use therein is appropriate. In some cases it is. And in some cases not.

andrewfg:
@ccutrer / @jimtng / @holgerfriedrich see latest commit. Are we done now?

jimtng (approved):

holgerfriedrich (approved):

holgerfriedrich:
Thanks for the discussion and the reviews!

andrewfg:
Phew that was a hard one. But many thanks to you all!

Issue #4711, opened 2025-04-09
UPNP Embedded device not properly handled in current implementations


lo92fr:
## Expected Behavior

Some devices like Denon / Marantz AV receiver expose their UPNP endpoints as embedded devices.
In the semple bellow, we have one denon AiosDevice that is compose of a Upnp Server, an Upnp Renderer, an ACT, and a AiosServices.
```
urn:schemas-denon-com:device:AiosDevice:1			uuid:bc3a33a3-38f6-1fcb-0080-0005cd7be0c6			
        urn:schemas-upnp-org:device:MediaServer:1		uuid:85f888a6-35ec-4bdf-86fe-c9ffa19a0d92
	urn:schemas-denon-com:device:ACT-Denon:1		uuid:662e4de5-90d5-6350-c29e-a7d70c404fd1				
	urn:schemas-denon-com:device:AiosServices:1		uuid:0bbd03ac-07b4-149a-0080-0005cd7be0c6
	urn:schemas-upnp-org:device:MediaRenderer:1		uuid:d807183e-9fd4-1457-0080-0005cd7be0c6
```	
This MediaRenderer and MediaServer should be shone in Inbox after an UPNP discovery.

## Current Behavior
Today the UPNP Server and renderer don't show in openhab Inbox after UPNP discovery because the embedded devices are not handled.

## Possible Solution
I push the following pull request to fix this:  #4712 

## Steps to Reproduce (for Bugs)
- Add the UPNPControl addon
- Trigger an UPNP Scan discovery
- Media Renderer and server does not show in Inbox.

## Your Environment
Windows 11 / Java 21 / Openhab 5.0-snapshot

andrewfg:
@lo92fr can you please mark this as closed via #4712 ?

lo92fr:
Fix by merge of #4712

lo92fr:
Hello @andrewfg,

Done.

Laurent.

PR #4712, opened 2025-04-09
Fix UPNP support to handle embededd devices

lo92fr:
# Fix UPNP support to handle embedded devices

# Description

As describe in issue #4711, the UPNP embedded device are today ignore in openhab code.
We need three modification to handle them:

- In discovery service, loop on the EmbeddedDevice collection to also handle there as discovery result.
- In UpnpIOServiceImpl, change call to getDevice, passing false instead of true to rootOnly parameter.
- In UpnpIOServiceImpl, change logic in remoteDeviceAdded and remoteDeviceRemoved to also handle embeddedDevice.

### DIFF
--- a/bundles/org.openhab.core.config.discovery.upnp/src/main/java/org/openhab/core/config/discovery/upnp/internal/UpnpDiscoveryService.java
+++ b/bundles/org.openhab.core.config.discovery.upnp/src/main/java/org/openhab/core/config/discovery/upnp/internal/UpnpDiscoveryService.java
@@ -149,6 +149,10 @@ public class UpnpDiscoveryService extends AbstractDiscoveryService
     protected void startScan() {
         for (RemoteDevice device : upnpService.getRegistry().getRemoteDevices()) {
             remoteDeviceAdded(upnpService.getRegistry(), device);
+
+            for (RemoteDevice childDevice : device.getEmbeddedDevices()) {
+                remoteDeviceAdded(upnpService.getRegistry(), childDevice);
+            }
         }
         upnpService.getRegistry().addListener(this);
         upnpService.getControlPoint().search();
--- a/bundles/org.openhab.core.io.transport.upnp/src/main/java/org/openhab/core/io/transport/upnp/internal/UpnpIOServiceImpl.java
+++ b/bundles/org.openhab.core.io.transport.upnp/src/main/java/org/openhab/core/io/transport/upnp/internal/UpnpIOServiceImpl.java
@@ -216,7 +216,7 @@ public class UpnpIOServiceImpl implements UpnpIOService, RegistryListener {
     }
 
     private Device getDevice(UpnpIOParticipant participant) {
-        return upnpService.getRegistry().getDevice(new UDN(participant.getUDN()), true);
+        return upnpService.getRegistry().getDevice(new UDN(participant.getUDN()), false);
     }
 
     @Override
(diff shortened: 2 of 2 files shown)


lo92fr:
Thanks andrew !

mherwege:
Thanks for this fix.

holgerfriedrich (approved):

lolodomo:
This change leads to new discovered things for Sonos devices that are not expected.
Does it mean all bindings based on UPnP discovery should now be updated ?
Unexpected things are _MR and  _MS.

andrewfg:
> Unexpected things are _MR and _MS.

MediaRenderer and MediaServer??

lolodomo:
> MediaRenderer and MediaServer??

Yes. They are embedded in the main device but they should no be added as things in the inbox.
@lo92fr : did you think about the case we don't want to consider the embedded devices ?

andrewfg:
I suppose most addon UPNP disco services already have some filters to only add wanted UPNP devices to the inbox?

lolodomo:
> I suppose most addon UPNP disco services already have some filters to only add wanted UPNP devices to the inbox?

I would be very surprised Sonos to be the only binding not doing it.

andrewfg:
There must be some filtering in Sonos disco otherwise it would anyway add all UPNP devices on the network.

lolodomo:
> There must be some filtering in Sonos disco otherwise it would anyway add all UPNP devices on the network.

What I try to say is that this PR impacted certain bindings, that is apparently the case of the Sonos binding but it probably impacted also other bindings.

andrewfg:
Do you want to revert this PR?

lolodomo:
> Do you want to revert this PR?

How could we evaluate its wrong impact ?
If only few bindings are impacted, the good option could be to rather fix them.
I have not yet checked but in the UPnP discovery class of a binding, I hope we have a way to filter on root device?

andrewfg:
I am wondering if we can add something so that binding disco services do not see embedded devices unless they expressly ask to see them. Let me look at it later today.

andrewfg:
@lolodomo see #4735

mherwege:
> This change leads to new discovered things for Sonos devices that are not expected. Does it mean all bindings based on UPnP discovery should now be updated ? Unexpected things are _MR and _MS.

Are these extra devices discovered by the sonos binding or the upnpcontrol binding? The whole point about this enhancement is to allow subdevices to be discovered by the upnpcontrol binding. I don’t see how you can disable it in the upnpcontrol binding only for Sonos. And it is a very valuable feature for upnpcontrol as I have had many requests in the past where subdevices where missed.

andrewfg:
> I don’t see how you can disable it in the upnpcontrol binding only for Sonos. And it is a very valuable feature for upnpcontrol as I have had many requests in the past where subdevices where missed.

@mherwege see #4735

mherwege:
> > I don’t see how you can disable it in the upnpcontrol binding only for Sonos. And it is a very valuable feature for upnpcontrol as I have had many requests in the past where subdevices where missed.
> 
> @mherwege see #4735

That would allow you to disable it for upnpcontrol, but it would be disabled for all embedded devices, not just the Sonos ones. If you enable it in upnpcontrol, the Sonos ones will again show up, hence my question above.

mherwege:
> > it would be disabled for all embedded devices, not just the Sonos ones
> 
> No. The **_discovery_** is always enabled. Only the **_notifications_** to the respective addons' discovery participants are disabled resp. optionally enabled.

Yes, you are right. But that effect is still what I wrote above.

andrewfg:
I don't understand your problem.

mherwege:
So @andrewfg PR will work to avoid this. The upnpcontrol binding will have to be adapted to get notifications about the embedded devices and will then discover the embedded Sonos media server and media renderer.

andrewfg:
@mherwege just to reiterate; the PR will in either case require changes in addons .. either a) for addons to opt out of receiving child device notifications, or b) with my other PR for addons to opt in for them. Personally I think it is more stable for knowledgeable developers to modify addons to explicitly opt in, rather than less knowledgeable ones maybe guessing about opting out..

PR #4718, opened 2025-04-11
Regenerate rules from templates

Nadahar:
# Regenerate rules from templates

Provide the ability to regenerate template based rules.

# Description

Currently, template based rules must be deleted and recreated from the same template if the template is updated or if there is a need to change some of the configuration parameters (like linking the rule to a different Item). This PR provides the base functionality to enable regeneration of rules without having to delete and recreate them.

This PR alone won't do much, it will only enable the required functionality through the REST API. A companion PR will be posted for MainUI that will make use of the provided functionality.

### DIFF
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/dto/RuleDTO.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/dto/RuleDTO.java
@@ -16,6 +16,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import org.eclipse.jdt.annotation.NonNull;
 import org.openhab.core.automation.Visibility;
 import org.openhab.core.config.core.dto.ConfigDescriptionParameterDTO;
 
@@ -26,15 +27,16 @@ import org.openhab.core.config.core.dto.ConfigDescriptionParameterDTO;
  */
 public class RuleDTO {
 
-    public List<TriggerDTO> triggers;
-    public List<ConditionDTO> conditions;
-    public List<ActionDTO> actions;
-    public Map<String, Object> configuration;
-    public List<ConfigDescriptionParameterDTO> configDescriptions;
+    public List<@NonNull TriggerDTO> triggers;
+    public List<@NonNull ConditionDTO> conditions;
+    public List<@NonNull ActionDTO> actions;
+    public Map<@NonNull String, @NonNull Object> configuration;
+    public List<@NonNull ConfigDescriptionParameterDTO> configDescriptions;
     public String templateUID;
+    public String templateState;
     public String uid;
     public String name;
-    public Set<String> tags;
+    public Set<@NonNull String> tags;
     public Visibility visibility;
     public String description;
 }
--- a/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleRegistryImpl.java
+++ b/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleRegistryImpl.java
@@ -24,6 +24,8 @@ import java.util.Set;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
+import org.openhab.core.automation.Action;
+import org.openhab.core.automation.Condition;
 import org.openhab.core.automation.ManagedRuleProvider;
 import org.openhab.core.automation.Module;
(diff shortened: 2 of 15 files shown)


Nadahar:
I've created https://github.com/openhab/openhab-addons/pull/18705 which should be merged if this is merged. It handles the small change to `RuleRegistry` this PR imposes. I've built the complete add-ons repo against this PR, and there seem to be no other issues.

Nadahar:
I have created two "companion PRs" that should go together with this:

https://github.com/openhab/openhab-addons/pull/18705

https://github.com/openhab/openhab-webui/pull/3197

The add-ons PR only modified one official add-on to accommodate the small change to `RuleRegistry` for this PR, it was the only problem I could find among all the official add-ons.

The UI PR makes the functionality for which this PR lays the foundation actually available to users.

Nadahar:
@kaikreuzer I rebased on latest `main` as well, so the comparison between before and after rebase is quite large. I only changed the two things found by Copilot though.

Nadahar:
I've been trying to find a way to run the integration test locally without building everything. Running `itest.bndrun` from Eclipse didn't work, I got a ton of errors. Running this from the command line (from the `itests` folder) succeeds, but I'm not confident that it has actually run the test:
```
mvn package -Dspotless.check.skip=true -pl :org.openhab.core.automation.integration.tests
```

Is there a way to run this test in isolation so that I can try to figure out what goes wrong?

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/dto/RuleDTO.java
```diff
@@ -16,6 +16,7 @@
 import java.util.Map;
 import java.util.Set;
 
+import org.eclipse.jdt.annotation.NonNull;
```
kaikreuzer, line 19: When adding null annotations, we typically add `@NonNullByDefault` to the class and then add `@Nullable` whereever appropriate.

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleRegistryImpl.java
```diff
@@ -336,6 +328,44 @@ public Collection<Rule> getByTags(String... tags) {
...
+            if (provider instanceof ManagedRuleProvider) {
+                update(resolvedRule);
+            } else {
+                updated(provider, rule, resolvedRule);
+            }
+            if (resolvedRule.getTemplateState() == TemplateState.TEMPLATE_MISSING) {
+                logger.warn("Failed to regenerate rule '{}' from template since the template is missing",
+                        rule.getUID());
+            } else {
+                logger.info("Rule '{}' was regenerated from template '{}'", rule.getUID(), rule.getTemplateUID());
```
kaikreuzer, line 361: This should better only be debug.

kaikreuzer (commented):
Many thanks @Nadahar, the code looks very clean.
I have only two very minor comment, then we should be good to merge!

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleRegistryImpl.java
Nadahar: This is the result of a user action, which is why I thought it would be nice to see a confirmation. But, I can change it if you think that's the best way.

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/dto/RuleDTO.java
kaikreuzer: Ok, if it is done similarly on other DTOs, let's keep it; I get your point.

bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/internal/RuleRegistryImpl.java
kaikreuzer: Ok, as long as there is no automatic or bulk regeneration possible, then it should be ok.

Nadahar:
@kaikreuzer I see that you squashed this when you merged. Is that the general practice, or is it because you considered the commits "too fine-grained"?

The reason I'm asking is that I spend quite a lot of time rebasing everything into what I consider "logical blocks" so that the commits represent "meaningful changes" if somebody is to attempt to understand how things come to be using Git history in the future. If commits are just squashed anyway, I can save myself this work.

PR #4719, opened 2025-04-11
Allow ThingHandlers to dynamically register services

ccutrer:
This is needed for dynamic things like Home Assistant devices that may or may not have particular services available, and it's not known until the thing has be initialized if that is the case. See also https://github.com/openhab/openhab-addons/issues/17992

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -496,30 +496,6 @@ Import-Package: \\
         </plugin>
 
         <plugin>
-          <groupId>org.openhab.tools.sat</groupId>
-          <artifactId>sat-plugin</artifactId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -495,30 +495,6 @@ Import-Package: \\
           <version>4.3.1</version>
         </plugin>
 
-        <plugin>
```
kaikreuzer, line 498: This change looks rather unrelated to the PR?

ccutrer: Yup. reverted.

holgerfriedrich (approved):
LGTM

PR #4724, opened 2025-04-17
Make `expire` support days, ISO8601 Duration and fully configurable through metadata configuration map

jimtng:
- Enrich duration format:

  - Duration format supports days e.g. `1d3h`
  - Supports ISO8601 Duration syntax e.g. `P1DT3H`

- Support configuring the expire manager through the metadata configuration map by adding:

  - `duration`: "1h 1m 1s"
  - `command`: ON
  - `state`: ON

- Fully backwards compatible. The old way of specifying expire value `1h5m,command=OFF` is still supported
- Specifying any setting (duration, command, state) in both value and metadata config will raise an error
- Specifying both command and state in the configuration will raise an error

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ExpireManager.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/ExpireManager.java
@@ -303,8 +303,8 @@ public class ExpireManager implements EventSubscriber, RegistryChangeListener<It
         protected static final String COMMAND_PREFIX = CONFIG_COMMAND + "=";
         protected static final String STATE_PREFIX = CONFIG_STATE + "=";
 
-        protected static final Pattern DURATION_PATTERN = Pattern
-                .compile("(?:([0-9]+)H)?\\s*(?:([0-9]+)M)?\\s*(?:([0-9]+)S)?", Pattern.CASE_INSENSITIVE);
+        protected static final Pattern DURATION_PATTERN = Pattern.compile(
+                "(?:([0-9]+)D)?\\s*(?:([0-9]+)H)?\\s*(?:([0-9]+)M)?\\s*(?:([0-9]+)S)?", Pattern.CASE_INSENSITIVE);
 
         final @Nullable Command expireCommand;
         final @Nullable State expireState;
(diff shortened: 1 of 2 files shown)


spacemanspiff2007:
Would it be possible to also add days to the duration string? 
Additionally if would be nice and I think [very easy to implement](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#parse-java.lang.CharSequence-) if parsing [ISO_8601 Durations](https://en.wikipedia.org/wiki/ISO_8601#Durations) would also be supported.


> Note: command takes precedence over state

Would it be possible to raise an error? 
This is very error prone and almost certainly a misconfiguration.

jimtng:
> Would it be possible to also add days to the duration string? 

Done

> Additionally if would be nice and I think [very easy to implement](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#parse-java.lang.CharSequence-) if parsing [ISO_8601 Durations](https://en.wikipedia.org/wiki/ISO_8601#Durations) would also be supported.

Done


> 
> > Note: command takes precedence over state
> 
> Would it be possible to raise an error? This is very error prone and almost certainly a misconfiguration.

Done. It is now no longer possible to have redundant config. It can only be specified once anywhere.

> 
> Do you think it makes sense to raise an error on additional unsupported keys in the map? E.g. I mistyped `statw` instead of `state`?

Done.

jimtng:
> Edit: Could you also add a test where the user specifies both `state` and `command` which is an invalid config, too?

https://github.com/jimtng/openhab-core/blob/befdb188a4f7f4e3112c1b2763b29d692c80c2c4/bundles/org.openhab.core/src/test/java/org/openhab/core/internal/items/ExpireManagerTest.java#L437-L438

spacemanspiff2007:
My bad - I somehow missed it. 

Perfect 👍

rkoshak:
Yes! I've wanted ISO8601 durations for a long time.

Got to be careful though as ISO8601 supports negative durations. We need to handle that case in a reasonable way. Ideas include treating a negative duration as 0 or throwing an error message. I think the latter is better but either would be reasonable I think.

There should probably be a test for that too.

(I briefly looked at the code and didn't see that case being handled, sorry if I missed it).

jimtng:
there's currently no check for negative duration. I'll add it

kaikreuzer (approved):
Thanks! Don't forget to do a PR for the docs as well now.

PR #4737, opened 2025-04-22
Semantic Tag: add `MuteState` property

jimtng:


jimtng:
FYI @andrewfg

andrewfg:
@jimtng I don't think you need this .. it is perfectly covered by `Switch:SoundVolume` ..

jimtng:
> @jimtng I don't think you need this .. it is perfectly covered by `Switch:SoundVolume` ..

You're right, Thanks!

Issue #4736, opened 2025-04-21
[Regression] States of nested group members are not restored at startup since 5.0.0.M2


LukasA83:
I am using MapDB to restore values at startup. Since 5 M2 the values of nested group members are not restored. 
Persistance config:

```
configurations:
  - items:
      - gRestoreOnStartup*
    strategies:
      - everyChange
      - restoreOnStartup
    filters: []
cronStrategies: []
defaultStrategies:
  - everyChange
thresholdFilters: []
timeFilters: []
equalsFilters: []
includeFilters: []
```

Steps to reproduce:
Group setup:
"gRestoreOnStartup" has a child group item called "gChildRestore". "gChildRestore" has a child called "xyz". State is getting restored in 5.0.0.M1 but not in M2.

PR #4743, opened 2025-04-23
Fix groupItem restore

mherwege:
Closes https://github.com/openhab/openhab-core/issues/4736
Closes https://github.com/openhab/openhab-addons/issues/18586

Regression of https://github.com/openhab/openhab-core/pull/4651

For group items with an aggregation function (state), nested items where not persisted and restored. These nested items where missed when building the list of items to persist/restore. It did work well when the group item did not have a state.


holgerfriedrich (approved):
Thanks!

PR #4745, opened 2025-04-24
Semantic tags: Just a few more

andrewfg:
This PR adds the following tags..

- `Point.Calculation`
- `Property.Info`
- `Property.Progress`
- `Equipment.Application`


andrewfg:
@mherwege / @jimtng .. writing the proper description texts will become a monster task :( .. as example see my first attempt below for `Point`..

```
Point,Alarm,,Alarm,Read only digital data point indicating an alarm condition,
Point,Calculation,,Read only analog calculated data point,,
Point,Control,,Control,Write-able analog data point,
Point,Switch,Control,Switch,Write-able digital data point,
Point,Forecast,,Read only analog predicted data point,,
Point,Measurement,,Read only analog data point from a sensor,,
Point,Setpoint,,Setpoint,Write-able analog data point for a controller,
Point,Status,,Status,Read only data point indicating a status condition,,
```

jimtng:
For this PR, just add those new tags.

Keep the description stuff in a separate PR so it doesn't hold up the new tag addition.
Also the description is the _last_ field in the CSV, not the label, nor synonyms.

andrewfg:
> Keep the description stuff in a separate PR so it doesn't hold up the new tag addition.

Ok. Then this PR is ready to go ;)

jimtng:
btw how did you mark a comment as outdated? :)

holgerfriedrich:
just asking, why do we have `Property_App` and `Equipment_Application`?

andrewfg:
The one describes the Thing (Equipment) and the other describes the Channel .. or specifically in this case the object (Property) which the Channel is changing.

holgerfriedrich:
Ok, but still the postfix is different: `_App`, `Application`. Is this intended?

bundles/org.openhab.core.semantics/model/SemanticTags.csv
```diff
@@ -103,6 +105,7 @@ Property,Water,,Water,,
 Property,Wind,,Wind,,
 Equipment,AlarmDevice,,Alarm Device,,
 Equipment,AlarmSystem,,Alarm System,Alarm Systems,
+Equipment,Application,,Application,"App, Software program",
```
jimtng, line 108:
```suggestion
Equipment,Application,,Application,Software program,
```
The synonym mustn't match another existing tag. All synonyms from all tags must be unique to avoid ambiguities.

andrewfg:
> Is this intended?

I believe so.

An Equipment.Application is something (a Thing) most often running on the same machine that OH is on. (Otherwise it would a NetworkAppliance or a Webservice or a SmartPhone). Which is the host for OH channels. (The use case is that a program running on the same Linux or Windows machine as OH, provides data to feed channels in OH).

A Property.App is the target of an OH channel command to some other device on the LAN. (Which is therefore probably a NetworkDevice).  The use case is that the OH property app channel sends a command to a firestick to start (say) the BBC iplayer app.

mherwege:
> Ok, but still the postfix is different: `_App`, `Application`. Is this intended?

In my understanding, the name (i.e. last part of the id) needs to be unique across the tags.

holgerfriedrich (approved):
Thanks for the feedback. LGTM.

andrewfg:
Thanks @holgerfriedrich

PR #4746, opened 2025-04-24
Enable semantic tag description translations

mherwege:
With all the work being done on adding better semantic tags, I was looking at making the semantic tag descriptions available in the UI, to help the user in deciding what tag to use.
It turned out very few had descriptions in the first place, and they cannot be localized.

This PR adds the ability to localize the semantic tag descriptions.

I would think an effort should be made to give the semantic tags relevant descriptions that help the user to understand what tag to use when. I think the best people to do that are also the ones that worked on changing and adding the semantic tags. @andrewfg @jimtng and others?

I intend to enhance the UI to make these descriptions visible when available. They are already in the the semantic tag maintenance UI proposed in https://github.com/openhab/openhab-webui/pull/3165.

### DIFF
--- a/bundles/org.openhab.core.semantics/model/generateTagClasses.groovy
+++ b/bundles/org.openhab.core.semantics/model/generateTagClasses.groovy
@@ -96,6 +96,10 @@ def appendLabelsFile(FileWriter file, def line, String tagSet) {
         file.write("," + line.Synonyms.replaceAll(", ", ","))
     }
     file.write("\n")
+    if (line.Description) {
+        file.write(tagSet + "__description=" + line.Description)
+        file.write("\n")
+    }
 }
 
 def camelToUpperCasedSnake(def tag) {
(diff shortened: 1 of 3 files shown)


andrewfg:
We do have a .properties file .. but it is auto generated on each build. So we need to remove it from the auto gen script profile. And start to maintain it manually. If @jimtng can do the former, then I will do the latter.

mherwege:
I don't think you need to change that auto generation. I enhanced the groovy script to generate keys for the descriptions in the properties file also.

andrewfg:
Don't we need to provide the English texts?

mherwege:
> Don't we need to provide the English texts?

They are in the csv, but only 2 are provided at the moment. So that’s the next step. Fill the csv with proper descriptions.

andrewfg:
That's what I meant. So you confirm the csv is (shall be) the origin of everything?

mherwege:
> That's what I meant. So you confirm the csv is (shall be) the origin of everything?

Yes, I do. I didn’t intend to, and didn’t, change that mechanism. I do feel that providing proper descriptions is best done by all involved adjusting/adding the previous semantic tags though. I know a lot of discussion went into what and why and that should somehow be reflected in the descriptions. I am not on top of all of that. While I can try, I would probably miss quite a few of the fine details. So I was looking at you?

andrewfg:
@mherwege as I have yet another PR currently open that touches the csv, I will take the opportunity to overload that PR with more extensive tag label texts => https://github.com/openhab/openhab-core/pull/4745

holgerfriedrich (approved):
LGTM, thanks!

PR #4748, opened 2025-04-25
YAMLThing: fix typo and minor refactor

jimtng:
The typo was `segement`.

@lolodomo

### DIFF
--- a/bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/internal/things/YamlThingDTO.java
+++ b/bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/internal/things/YamlThingDTO.java
@@ -33,9 +33,9 @@ import org.openhab.core.model.yaml.internal.util.YamlElementUtils;
 @YamlElementName("things")
 public class YamlThingDTO implements YamlElement, Cloneable {
 
-    private static final Pattern THING_UID_SEGMENT_PATTERN = Pattern.compile("^[a-zA-Z0-9_][a-zA-Z0-9_-]*$");
+    private static final Pattern THING_UID_SEGMENT_PATTERN = Pattern.compile("[a-zA-Z0-9_][a-zA-Z0-9_-]*");
     private static final Pattern CHANNEL_ID_PATTERN = Pattern
-            .compile("^[a-zA-Z0-9_][a-zA-Z0-9_-]*(#[a-zA-Z0-9_][a-zA-Z0-9_-]*)?$");
+            .compile("[a-zA-Z0-9_][a-zA-Z0-9_-]*(#[a-zA-Z0-9_][a-zA-Z0-9_-]*)?");
 
     public String uid;
     public Boolean isBridge;
@@ -90,11 +90,10 @@ public class YamlThingDTO implements YamlElement, Cloneable {
             ok = false;
         }
         for (String segment : segments) {
-            if (!THING_UID_SEGMENT_PATTERN.matcher(segment).find()) {
+            if (!THING_UID_SEGMENT_PATTERN.matcher(segment).matches()) {
                 if (errors != null) {
-                    errors.add(
-                            "thing %s: segement \"%s\" in uid is not matching the expected syntax [a-zA-Z0-9_][a-zA-Z0-9_-]*"
-                                    .formatted(uid, segment));
+                    errors.add("thing %s: segment \"%s\" in uid is not matching the expected syntax %s".formatted(uid,
+                            segment, THING_UID_SEGMENT_PATTERN.pattern()));
                 }
                 ok = false;
             }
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/internal/things/YamlThingDTO.java
```diff
@@ -33,9 +33,9 @@
 @YamlElementName("things")
 public class YamlThingDTO implements YamlElement, Cloneable {
 
-    private static final Pattern THING_UID_SEGMENT_PATTERN = Pattern.compile("^[a-zA-Z0-9_][a-zA-Z0-9_-]*$");
+    private static final Pattern THING_UID_SEGMENT_PATTERN = Pattern.compile("[a-zA-Z0-9_][a-zA-Z0-9_-]*");
     private static final Pattern CHANNEL_ID_PATTERN = Pattern
-            .compile("^[a-zA-Z0-9_][a-zA-Z0-9_-]*(#[a-zA-Z0-9_][a-zA-Z0-9_-]*)?$");
+            .compile("[a-zA-Z0-9_][a-zA-Z0-9_-]*(#[a-zA-Z0-9_][a-zA-Z0-9_-]*)?");
```
lolodomo, line 38: Did you test these changes ? I am sure it was not working as expected without the start and end character ?
Or is it working because you replaced find() by matches() ?

bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/internal/things/YamlThingProvider.java
```diff
@@ -176,7 +175,7 @@ public void deactivate() {
 
     @Override
     public Collection<Thing> getAll() {
-        return thingsMap.values().stream().flatMap(list -> list.stream()).collect(Collectors.toList());
+        return thingsMap.values().stream().flatMap(list -> list.stream()).toList();
```
lolodomo, line 178: I know both are not equivalent but I am never comfortable with what option should be used.
Are you sure of this change is correct in our context and if it is can you explain why please ?

jimtng: toList returns an unmodifiable list, whereas collect(Collectors.toList()) returns a modifiable one, although afaik, not guaranteed.

I am just following this: https://github.com/openhab/openhab-core/pull/3831

Many existing implementations of getAll() in core would return a Stream.toList().

In short, use Stream.toList() whenever possible. For me it's just shorter to type and read.

bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/internal/things/YamlThingDTO.java
jimtng: Yes, matches() will check the pattern against the entire string. See https://docs.oracle.com/javase/8/docs/api/java/util/regex/Matcher.html

> A matcher is created from a pattern by invoking the pattern's [matcher](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#matcher-java.lang.CharSequence-) method. Once created, a matcher can be used to perform three different kinds of match operations:

> The [matches](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Matcher.html#matches--) method attempts to match the entire input sequence against the pattern.

Tests:

```
jshell> Pattern.compile("[a-zA-Z0-9_][a-zA-Z0-9_-]*").matcher("ab12_332").matches()
$1 ==> true

jshell> Pattern.compile("[a-zA-Z0-9_][a-zA-Z0-9_-]*").matcher("1ab12_332").matches()
$2 ==> true

jshell> Pattern.compile("[a-zA-Z0-9_][a-zA-Z0-9_-]*").matcher(" ab12_332").matches()
$3 ==> false

jshell> Pattern.compile("[a-zA-Z0-9_][a-zA-Z0-9_-]*").matcher("-ab12_332").matches()
$4 ==> false

jshell> Pattern.compile("[a-zA-Z0-9_][a-zA-Z0-9_-]*").matcher("a-b12_332").matches()
$5 ==> true

jshell> Pattern.compile("[a-zA-Z0-9_][a-zA-Z0-9_-]*").matcher("a-b1#2_332").matches()
$6 ==> false

jshell> Pattern.compile("[a-zA-Z0-9_][a-zA-Z0-9_-]*").matcher("#a-b12_332").matches()
$7 ==> false

jshell> Pattern.compile("[a-zA-Z0-9_][a-zA-Z0-9_-]*").matcher("a-b12_332#").matches()
$8 ==> false
```

holgerfriedrich (approved):
LGTM, thanks!

bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/internal/things/YamlThingDTO.java
lolodomo: Ok, thank you, I will take care of that for the items ;)

PR #4749, opened 2025-04-25
Yaml Model Repository: Remove elements no longer in model

jimtng:
This fix two scenarios:
- When the new model is missing `version:` key but the old model had one, the file is considered invalid, but the previously loaded elements didn't get removed
- When the new model has a valid `version:` key, but the entire element type, e.g. `thing:` is removed, the previously loaded elements (e.g. things) didn't get removed.

Additionally, log a warning when an unknown element type is encountered, log a warning, e.g.
```
version: 2
thangs: # misspelled?
  x:y:z:d: xx
```

See https://github.com/openhab/openhab-core/issues/3666#issuecomment-2829347883

### DIFF
--- a/bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/internal/YamlModelRepositoryImpl.java
+++ b/bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/internal/YamlModelRepositoryImpl.java
@@ -81,6 +81,8 @@ import com.fasterxml.jackson.dataformat.yaml.YAMLParser;
 @Component(immediate = true)
 public class YamlModelRepositoryImpl implements WatchService.WatchEventListener, YamlModelRepository {
     private static final int DEFAULT_MODEL_VERSION = 2;
+    private static final String VERSION = "version";
+    private static final String READ_ONLY = "readOnly";
 
     private final Logger logger = LoggerFactory.getLogger(YamlModelRepositoryImpl.class);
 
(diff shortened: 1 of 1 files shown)


lolodomo (approved):
Good improvements, thank you

holgerfriedrich (approved):

lolodomo:
It would be appreciated if you could add a new unit test for the case when an element is removed to check that that listener are properly called.

lolodomo:
Your solution is not ideal at all because we got now unexpected warnings due to the fact that the models can be first loaded before the providers (listeners) are started. It leads to that:
```
09:10:46.850 [INFO ] [aml.internal.YamlModelRepositoryImpl] - Adding YAML model things\astro.yaml
09:10:46.851 [WARN ] [aml.internal.YamlModelRepositoryImpl] - Element 'things' in model things\astro.yaml is unknown.
09:10:46.852 [INFO ] [aml.internal.YamlModelRepositoryImpl] - Adding YAML model things\freebox.yaml
09:10:46.852 [WARN ] [aml.internal.YamlModelRepositoryImpl] - Element 'things' in model things\freebox.yaml is unknown.
09:10:46.855 [INFO ] [aml.internal.YamlModelRepositoryImpl] - Adding YAML model things\netatmo.yaml
09:10:46.855 [WARN ] [aml.internal.YamlModelRepositoryImpl] - Element 'things' in model things\netatmo.yaml is unknown.
09:10:46.857 [INFO ] [aml.internal.YamlModelRepositoryImpl] - Adding YAML model things\ntp.yaml
09:10:46.857 [WARN ] [aml.internal.YamlModelRepositoryImpl] - Element 'things' in model things\ntp.yaml is unknown.
09:10:47.261 [INFO ] [aml.internal.YamlModelRepositoryImpl] - Refreshing things from YAML model things\ntp.yaml
09:10:47.272 [INFO ] [aml.internal.YamlModelRepositoryImpl] - Refreshing things from YAML model things\freebox.yaml
09:10:47.289 [INFO ] [aml.internal.YamlModelRepositoryImpl] - Refreshing things from YAML model things\netatmo.yaml
09:10:47.366 [INFO ] [aml.internal.YamlModelRepositoryImpl] - Refreshing things from YAML model things\astro.yaml
...

jimtng:
> your solution is not ideal at all because we got now unexpected warnings due to the fact that the models can be first loaded before the providers (listeners) are started. It leads to that:

I'll see if we can change the load order.

jimtng:
It currently will still load them once the listener is loaded. It's just ideally we don't see those warnings.

lolodomo:
> It currently will still load them once the listener is loaded. It's just ideally we don't see those warnings.

Yes

jimtng:
Understood. I'm still thinking of how to best solve this.

PR #4751, opened 2025-04-25
Enable Dependabot PRs

holgerfriedrich:
This is a first step towards automating plugin and dependency updates. Dependabot will create PRs (which will still need review, no use of auto-merge intended!).

For GHA this will likely not require more configuration.

For Maven, I expect a bit more tuning, as we typically upgrade a lot of dependencies together/in sync with Karaf.
Not sure if I should add ignore statements here or answer the PRs with an ignore package comment. Let's try how this works out.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,16 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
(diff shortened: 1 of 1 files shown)


wborn (approved):
Let's test what happens. 😄

holgerfriedrich:
Great, no PRs ;-), no dependency upgrades pending....
Hm, there must be a syntax error in the config....
#4764

PR #4755, opened 2025-04-26
YAML Model Repository: Fix premature unknown element warning

jimtng:
Fix a problem introduced in #4749

Wait until all known elements are loaded before issuing warnings about unknown elements.

### DIFF
--- a/bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/internal/YamlModelRepositoryImpl.java
+++ b/bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/internal/YamlModelRepositoryImpl.java
@@ -30,6 +30,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


jimtng:
@lolodomo this isn't a perfect solution but it works for now. You'd need to add "items" to the list.

If we are anticipating that in the future, add-ons can add custom element types, we can defer the check using a scheduler for a pre-determined delay.

bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/internal/YamlModelRepositoryImpl.java
```diff
@@ -83,6 +84,7 @@ public class YamlModelRepositoryImpl implements WatchService.WatchEventListener,
     private static final int DEFAULT_MODEL_VERSION = 2;
     private static final String VERSION = "version";
     private static final String READ_ONLY = "readOnly";
+    private static final Set<String> KNOWN_ELEMENTS = Set.of("tags", "things");
```
lolodomo, line 87: Instead of hardcoded values, maybe you could get the annotation value from the DTO class ?

lolodomo (approved):
I have just a comment but except that looks good to me.

bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/internal/YamlModelRepositoryImpl.java
jimtng: I've made the change.

lolodomo (approved):
Looks good to me

jimtng:
Sorry I added two more changes:
- change the exception classname -> exception message like in the other pr
- Remove the unneeded temporary variable elementListeners

lolodomo (approved):

holgerfriedrich (approved):
Thanks!

PR #4757, opened 2025-04-27
Cleanup: Java21, SAT

holgerfriedrich:
fixing a few warnings added in the recent commits, should not include any functional change


wborn (approved):
Thanks!

PR #4760, opened 2025-04-27
[rest] Support sending Item command/state as JSON

florian-h05:
Adds support for sending command/state through media type `application/json` to the existing plain text endpoints.

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
@@ -104,6 +104,7 @@ import org.osgi.service.jaxrs.whiteboard.propertytypes.JaxrsResource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.gson.Gson;
 import com.google.gson.JsonObject;
 
 import io.swagger.v3.oas.annotations.Operation;
@@ -111,6 +112,7 @@ import io.swagger.v3.oas.annotations.Parameter;
 import io.swagger.v3.oas.annotations.media.ArraySchema;
 import io.swagger.v3.oas.annotations.media.Content;
 import io.swagger.v3.oas.annotations.media.Schema;
+import io.swagger.v3.oas.annotations.parameters.RequestBody;
 import io.swagger.v3.oas.annotations.responses.ApiResponse;
 import io.swagger.v3.oas.annotations.security.SecurityRequirement;
 import io.swagger.v3.oas.annotations.tags.Tag;
@@ -172,6 +174,7 @@ public class ItemResource implements RESTResource {
     }
 
     private final Logger logger = LoggerFactory.getLogger(ItemResource.class);
+    private final Gson gson = new Gson();
 
     private final DTOMapper dtoMapper;
     private final EventPublisher eventPublisher;
@@ -360,10 +363,19 @@ public class ItemResource implements RESTResource {
     @RolesAllowed({ Role.USER, Role.ADMIN })
     @Path("/{itemname: [a-zA-Z_0-9]+}/state")
     @Produces(MediaType.TEXT_PLAIN)
-    @Operation(operationId = "getItemState", summary = "Gets the state of an item.", responses = {
+    @Operation(operationId = "getOrSetItemState", summary = "Gets or sets the state of an item.", responses = {
             @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = String.class))),
+            @ApiResponse(responseCode = "202", description = "State accepted"),
+            @ApiResponse(responseCode = "400", description = "State cannot be parsed"),
             @ApiResponse(responseCode = "404", description = "Item not found") })
-    public Response getPlainItemState(@PathParam("itemname") @Parameter(description = "item name") String itemname) {
+    public Response getPlainItemState(
+            @HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) @Parameter(description = "language") @Nullable String language,
+            @PathParam("itemname") @Parameter(description = "item name") String itemname,
+            @QueryParam("state") @Parameter(description = "valid item state (e.g. ON, OFF) to set", required = false) @Nullable String state) {
+        if (state != null) {
+            return sendItemStateInternal(language, itemname, state);
+        }
+
         // get item
         Item item = getItem(itemname);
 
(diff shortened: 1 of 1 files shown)


spacemanspiff2007:
What about state updates? If commands are available these should work, too.

Are you aware of the [webhook binding](https://community.openhab.org/t/webhook-new-very-simple-binding-for-listening-incomming-http-requests/123597)?
Imho it tries to solve the same issue. Maybe it makes sense to consolidate into an official binding/endpoint?

TheNinth7:
> Are you aware of the [webhook binding](https://community.openhab.org/t/webhook-new-very-simple-binding-for-listening-incomming-http-requests/123597)?
> Imho it tries to solve the same issue. Maybe it makes sense to consolidate into an official binding/endpoint?

@spacemanspiff2007,

I’m the original requester for this change.
I’m developing an openHAB app for Garmin wearables. The Garmin SDK doesn’t support sending web requests with raw content, as required by the current API for sending a command.

The Webhook binding also doesn’t natively support sending commands to items. You need to create a custom Webhook setup, which involves a relatively complex configuration using a JEXL script in the Thing and additional JavaScript processing.

To make the Garmin app accessible to a wider audience, it would be much better if the needed API support were provided out of the box.

spacemanspiff2007:
@TheNinth7 
I'm not against using get requests to push data into openHAB, I explicitly support it.
This worked from OH1 - OH3 through Basic UI (unsupported) but then was unfortunately removed. 

All I'm saying is that
- state updates should be supported, too
- maybe it makes sense to implement a subset of the webhook binding
- maybe besides raw content json should also be supported for interacting with items, too

florian-h05:
> What about state updates? If commands are available these should work, too.

I didn’t notice the method for this, I will add that.

> maybe it makes sense to implement a subset of the webhook binding

I think there is some architectural difference here. The webhook binding won’t be allowed to directly control items as it is a binding, it provides a way to specify custom webhooks. I think its functionality being a binding only is fine, though it would be nice to be part of the distribution. 
The changes I do are to allow usage of the REST API like webhooks, so you only need HTTP GET for accessing some REST API functionality that currently requires the use of different HTTP methods.

> maybe besides raw content json should also be supported for interacting with items, too

That’s a good point, some clients only support posting JSON. It shouldn’t be a problem supporting multiple media types for the POST/PUT endpoints.

florian-h05:
So i am finished with adding the same functionality for states and JSON support for the existing send command/state endpoints.

splatch:
Just a generic comment around the merit - for rest interfaces GET requests should issue read requests (as verb suggests), rather than write.
It is not a bug issue for me, but it makes OH interface more of RPC.

kaikreuzer:
> Just a generic comment around the merit - for rest interfaces GET requests should issue read requests 

This is the reason why I always argued against such a "feature" - it completely goes against the [core principles of a REST API and GET requests being safe methods](https://restfulapi.net/http-methods/#get). I'm therefore still not a fan of adding that to our REST API. As @spacemanspiff2007 there was a "hidden" workaround in the past by having the Classic UI offering such an endpoint, which was not part of the REST API.

If the main use case is the support of webhooks, the approach with the webhook bindings looks much cleaner to me.

Just my two cents on the topic.

florian-h05:
What about splitting out the GET webhook functionality into a new core webhook API?
IMO it would be much more convenient to have support for sending commands and states directly without the need for a third party binding and linking channels.

TheNinth7:
From the perspective of the openHAB app I’m developing for Garmin, I can support both `PUT` and `POST` requests, with content types of either `application/x-www-form-urlencoded` or `application/json`. However, the Garmin SDK doesn’t always behave exactly as documented, so I’m happy to test in advance whichever approach you think best fits the openHAB REST API model before committing to an implementation.

spacemanspiff2007:
@kaikreuzer 
I fully agree with you that it's completely against the core principles of REST. 
But I'm also torn since it should be trivial to get data into openHAB.

Back in the days I had an old webcam on the only way to get notifications from it into openHAB was through a GET request because that was all the webcam offered. So it was either making that GET request work or nothing.
As you can guess I was not hesitant to use the workaround even though it was a dirty hack.

So I suggest that there is at least something or some way to provide more flexibility to ingest data, even though it's not the "proper" API.

Accepting json for commands / updates is imho a good idea in any case.
Maybe it makes sense to split this into two parts?

kaikreuzer:
> So I suggest that there is at least something or some way to provide more flexibility to ingest data, even though it's not the "proper" API.

Yes, I see the recurring request and see that there are many external services that don't follow REST principles, so I won't block any solution for this in core.

But I wouldn't advice to dilute the existing REST API and its endpoints by simply adding non-REST behavior to it [like here](https://github.com/openhab/openhab-core/pull/4760/files#diff-40fcf72cf6fd773547d30ba5271ba285feff86c9a761368c6e123065af8bce29L359-L364).
It might be better to maybe introduce a new dedicated endpoint like `/json-rpc/...` next to the `/rest/...` ones to keep these things apart.

digitaldan:
I also like the idea of a [JSON-RPC ](https://www.jsonrpc.org/) endpoint for this, its simple, follows a specification, and keeps the REST api clean.   I also would also definitely use this, i use  Caddy to proxy and rewrite requests from a few devices who only support get  requests for events.  Would be nice to remove those rules.

kaikreuzer (approved):
Thanks!

bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
```diff
@@ -459,12 +477,26 @@ public Response putItemState(
...
+        return sendItemCommandInternal(itemname, value);
+    }
+
+    @POST
+    @RolesAllowed({ Role.USER, Role.ADMIN })
+    @Path("/{itemname: [a-zA-Z_0-9]+}")
+    @Consumes(MediaType.APPLICATION_JSON)
+    public Response postItemCommandJson(@PathParam("itemname") String itemname, ValueContainer valueContainer) {
+        return sendItemCommandInternal(itemname, valueContainer.value());
+    }
```
jimtng, line 497: @lolodomo this made me think of you. I know we decided not to do it, but isn't this what you were trying to do but having trouble to do it? i.e. trying to accept either plain text and json.

florian-h05: Professing both MIME types is relatively simple, the most difficult part is getting Swagger work.

TheNinth7:
**If suppressing empty response errors is enabled:**

If that option is enabled in the [Settings](#-configuration), then the toast notification for the following errors will be suppressed.

* `INVRES` – The response was invalid (error code `-400`).
* `EMRES` – The response was empty.

If those errors persists for more than 10 seconds, they will be treated as fatal error and the full-screen error view will be displayed.

PR #4775, opened 2025-04-28
Tune Dependabot

holgerfriedrich:
Switch mvn to development deps only. This should cover mvn plugins.

Production deps are for now ignored, as Dependabot cannot update itests (thus PR builds would anyway fail), many deps are upgraded in sync with Karaf, etc.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,3 +19,8 @@ updates:
     open-pull-requests-limit: 10
     labels:
       - dependencies
+    allow:
+      - dependency-type: "development"
+    ignore:
+      - dependency-name: "org.openhab.core.*"
+      - dependency-name: "org.eclipse.smarthome.*"


holgerfriedrich:
@wborn We managed to summon the Bot 🧞, but as expected too many PRs.... 
Lets focus on the mvn plugins first - should not be too many.
GHA upgrades are currently hidden by the tons of mvn upgrades.

One open point is to get rid of the patch label, the bot adds if a dep has a patch level upgrade. Very misleading as we use this label differently.

.github/dependabot.yml
```diff
@@ -19,3 +19,8 @@ updates:
     open-pull-requests-limit: 10
     labels:
       - dependencies
```
holgerfriedrich, line 21:
```suggestion
    labels: [ ]
```

holgerfriedrich:
Labels hopefully gone with the latest change.... Let's see....

.github/dependabot.yml
```diff
@@ -17,5 +17,9 @@ updates:
...
       interval: "weekly"
     open-pull-requests-limit: 10
-    labels:
-      - dependencies
+    labels: [ ]
+    allow:
+      - dependency-type: "development"
+    ignore:
+      - dependency-name: "org.openhab.core.*"
+      - dependency-name: "org.eclipse.smarthome.*"
```
wborn, line 25: Why are these ignores added?
Also I can't find anything matching org.eclipse.smarthome?

holgerfriedrich:
```suggestion
    ignore:
      - dependency-name: "org.openhab.core.*"
```

holgerfriedrich:
Old deps were listed here:
https://github.com/openhab/openhab-core/network/dependencies
(which obviously includes all deps we have used before, there is not automatic removal). Thus I had excluded eclipse.smarthome.

Thanks for the info about resolving stuff automatically - I will try this on my repo, as this will need some iterations.....

My idea here was to go for the development stuff only - hopefully this targets the plugins. And care for the dependencies later.

holgerfriedrich:
@wborn can we try this to check if we get rid of the package upgrades?

wborn:
> Old deps were listed here:
https://github.com/openhab/openhab-core/network/dependencies
(which obviously includes all deps we have used before, there is not automatic removal). Thus I had excluded eclipse.smarthome.

I don't think this network is up to date? I can't find any ESH references as they have all been updated as that project is archived.

holgerfriedrich:
I did not realize that the network always contains all deps ever found in the project. There seems to be no way to clean this up.
That's why I added eclipse smart home to the ignore list in the beginning.

wborn:
The openHAB core ignore is probably also redundant. I don't think dependabot uses private repos unless specified.

.github/dependabot.yml
```diff
@@ -17,5 +17,8 @@ updates:
     schedule:
       interval: "weekly"
     open-pull-requests-limit: 10
-    labels:
-      - dependencies
+    labels: [ ]
+    allow:
+      - dependency-type: "development"
+    ignore:
+      - dependency-name: "org.openhab.core.*"
```
holgerfriedrich, line 24:
```suggestion
    ignore:
      - dependency-name: "org.openhab.core.*"
```

holgerfriedrich:
> The openHAB core ignore is probably also redundant. I don't think dependabot uses private repos unless specified.

Likely. I have removed it for now.

wborn (approved):
Thanks!

PR #4796, opened 2025-05-09
WatchService: use absolute path in processWatchEvent

jimtng:
# Problem:

Currently `processWatchEvent` is called with a relative path. This path is relative to the path given to `WatchService.registerListener`. For example:

```java
configWatchService.registerListener(this, Path.of("rules"));
```

Now when a file is created: `CONF/rules/test1.rules`

Will result in processWatchEvent being called with the `path` argument equals to `test1.rules`.

So far, all current implementations of `WatchService.WatchEventListener` have dealt with this by creating an absolute path that is relative to a pre-resolved absolute path that it passed to `registerListener`, for example: https://github.com/openhab/openhab-core/blob/8eb6eeae8473f8ba8ab5f74e53436dbedb53acc9/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptFileWatcher.java#L106

Without this pre-resolved absolute path, implementations of `processWatchEvent` cannot resolve the path properly given just the `watchService.getWatchPath()` when `registerListener` is given a non-blank path to watch.

This has never been an issue because nobody ever tried to call registerListener with multiple paths, which is supposed to be supported.

If you were to attempt this:

```java
configWatchService.registerListener(this, List.of(Path.of("rules"), Path.of("items")));
```

When two different files were created:
- `CONF/rules/foo`
- `CONF/items/foo`

`processWatchEvent` will be called twice, both with `path` argument set to `foo`. There  is no way for it to tell whether it's the rules/foo or items/foo.

I considered passing a relative path based on the given sub-path, e.g. `rules/foo` instead of an absolute path, but it will break unsuspecting implementations, whereas giving an absolute path will result in the following code to continue to work:

```java
basePathofSomeSort.resolve(path)
```

This is because `Path.resolve(q)` will simply return `q` when `q` is an absolute path.

# Change 

This PR fixes this problem by converting the path argument for processWatchEvent to an absolute path.

Prior to #3004, the old watch service also provided an absolute path.

I've looked at all the current implementations of the WatchService in addons, and I believe they will be unaffected by this change, because they're all using a fixed absolute path and actually ignoring the `path` argument.

- https://github.com/openhab/openhab-addons/blob/6ff50c5a32a41b16226b7dcad0c23cfc7a01f043/bundles/org.openhab.binding.exec/src/main/java/org/openhab/binding/exec/internal/ExecWhitelistWatchService.java#L65-L75
- https://github.com/openhab/openhab-addons/blob/6ff50c5a32a41b16226b7dcad0c23cfc7a01f043/bundles/org.openhab.binding.miio/src/main/java/org/openhab/binding/miio/internal/basic/MiIoDatabaseWatchService.java#L89-L95 - this uses the given path only to check the extension, so whether it's relative or absolute makes no difference. Its actual process method doesn't use the path argument, and instead uses its own internal path.
- https://github.com/openhab/openhab-addons/pull/11895 - the implementation isn't working because it doesn't call registerListener. It seems to be a bug in the addon.
- https://github.com/openhab/openhab-addons/blob/6ff50c5a32a41b16226b7dcad0c23cfc7a01f043/bundles/org.openhab.transform.exec/src/main/java/org/openhab/transform/exec/internal/ExecTransformationWhitelistWatchService.java#L63-L73 - this uses a pre-resolved absolute path, ignoring the given path argument
- There are various usage in JRubyScripting addons which I believe isn't working properly, and I suspect that it's caused partially by the problem that's fixed in this PR. I will investigate it further after this PR is merged. I had an open issue from a while back which might be related https://github.com/openhab/openhab-addons/issues/13922
- The JRuby helper library uses WatchService but it specifically converted the path to absolute so it won't be affected by this change. https://github.com/openhab/openhab-jruby/blob/3fc6e9a683e6a54a5b7d5da2022a5fd116ab5134/lib/openhab/dsl/rules/triggers/watch/watch_handler.rb#L96-L99

### DIFF
--- a/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptDependencyTracker.java
+++ b/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptDependencyTracker.java
@@ -79,7 +79,7 @@ public abstract class AbstractScriptDependencyTracker
 
     @Override
     public void processWatchEvent(WatchService.Kind kind, Path path) {
-        File file = libraryPath.resolve(path).toFile();
+        File file = path.toFile();
         if (kind == DELETE || (!file.isHidden() && file.canRead() && (kind == CREATE || kind == MODIFY))) {
             dependencyChanged(file.toString());
         }
--- a/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptDependencyTracker.java
+++ b/bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptDependencyTracker.java
@@ -78,8 +78,8 @@ public abstract class AbstractScriptDependencyTracker
     }
 
     @Override
-    public void processWatchEvent(WatchService.Kind kind, Path path) {
-        File file = path.toFile();
+    public void processWatchEvent(WatchService.Kind kind, Path fullPath) {
+        File file = fullPath.toFile();
         if (kind == DELETE || (!file.isHidden() && file.canRead() && (kind == CREATE || kind == MODIFY))) {
             dependencyChanged(file.toString());
         }
--- a/bundles/org.openhab.core.model.core/src/main/java/org/openhab/core/model/core/internal/folder/FolderObserver.java
+++ b/bundles/org.openhab.core.model.core/src/main/java/org/openhab/core/model/core/internal/folder/FolderObserver.java
@@ -59,6 +59,7 @@ import org.slf4j.LoggerFactory;
 @Component(name = "org.openhab.core.folder", immediate = true, configurationPid = "org.openhab.folder", configurationPolicy = ConfigurationPolicy.REQUIRE)
 public class FolderObserver implements WatchService.WatchEventListener {
     private final WatchService watchService;
+    private final Path watchPath;
     private final Logger logger = LoggerFactory.getLogger(FolderObserver.class);
 
     /* the model repository is provided as a service */
@@ -88,6 +89,7 @@ public class FolderObserver implements WatchService.WatchEventListener {
         this.modelRepository = modelRepo;
         this.readyService = readyService;
         this.watchService = watchService;
+        this.watchPath = watchService.getWatchPath();
     }
 
     @Reference(cardinality = ReferenceCardinality.AT_LEAST_ONE, policy = ReferencePolicy.DYNAMIC)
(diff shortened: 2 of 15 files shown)


bundles/org.openhab.core.automation.module.script.rulesupport/src/main/java/org/openhab/core/automation/module/script/rulesupport/loader/AbstractScriptDependencyTracker.java
```diff
@@ -79,7 +79,7 @@ public Path getLibraryPath() {
 
     @Override
     public void processWatchEvent(WatchService.Kind kind, Path path) {
```
J-N-K, line 81: Please also use `fullPath` here.

bundles/org.openhab.core.model.core/src/main/java/org/openhab/core/model/core/internal/folder/FolderObserver.java
```diff
@@ -287,7 +289,8 @@ private void checkPath(final Path path, final WatchService.Kind kind) {
     }
 
     @Override
-    public void processWatchEvent(WatchService.Kind kind, Path path) {
+    public void processWatchEvent(WatchService.Kind kind, Path absolutePath) {
```
J-N-K, line 292: Please use `fullPath` here (or change all to `absolutePath`, but it should be consistent with the naming in the interface).

bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/internal/YamlModelRepositoryImpl.java
```diff
@@ -157,18 +156,18 @@ public void deactivate() {
     @Override
     @SuppressWarnings({ "rawtypes", "unchecked" })
     public synchronized void processWatchEvent(Kind kind, Path path) {
```
J-N-K, line 158: see above

bundles/org.openhab.core.transform/src/test/java/org/openhab/core/transform/FileTransformationProviderTest.java
```diff
@@ -137,8 +139,8 @@ public void testMissingExtensionIsIgnored() throws IOException {
         Path extensionMissing = Path.of("extensionMissing");
```
J-N-K, line 139: I think you can use `Path extensionMissing = transformationPath.resolve("extensionMissing")` and remove `path` completely. (also in other places)

J-N-K (changes requested):
Thanks! Im general I agree. I have left some comments.

bundles/org.openhab.core.transform/src/test/java/org/openhab/core/transform/FileTransformationProviderTest.java
jimtng: good catch! I've adjusted it. There was only one other instance like this right below I think.

bundles/org.openhab.core.model.core/src/main/java/org/openhab/core/model/core/internal/folder/FolderObserver.java
jimtng: good suggestion! Whilst my gut feeling says it should be `absolutePath`, but it seems that `fullPath` is shorter and nicer, hence "easier" to type. I have changed them all to fullPath.

J-N-K (approved):
LGTM.

PR #4739, opened 2025-04-22
[GHA] Allow rebuild using label rebuild

holgerfriedrich:
* This restores functionality we had before we migrated the PR builds to GitHub Actions.
* Implement rebuild in ci_build workflow to make PR validation pass
* Remove label rebuild at the end of the workflow

The PR includes the latest fix from add-ons repo to make it work for the status check, see openhab/openhab-addons#18629.

PR #4808, opened 2025-05-12
[GHA] Permit rebuild from forks

holgerfriedrich:
Rebuild still did not work on PRs from forked repost.
This is to protect the repo from malicious PRs.

The way to do it is basically, to use the `pull_request_target` instead of `pull_request` (which always runs on read-only permissions).

This needs to be done carefully, see
https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/

Permissions have been reduced to the necessary ones. Persistence of checkout credentials has been disabled.

Regression of 
#4739 
#4785

### DIFF
--- a/.github/workflows/rebuild.yml
+++ b/.github/workflows/rebuild.yml
@@ -12,7 +12,7 @@
 name: Rebuild PR
 
 on:
-  pull_request:
+  pull_request_target:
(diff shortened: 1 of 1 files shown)


kaikreuzer (approved):
Thanks!

PR #4809, opened 2025-05-12
Fix CVE-2025-24813

th555555:
The vulnerability in the CorsFilter allows any arbitrary origin to access REST API resources by blindly reflecting the Origin header without validation. The fix adds proper origin validation against a configurable whitelist and implements security best practices following the CORS specification

The current implementation has a Cross-Origin Resource Sharing (CORS) misconfiguration vulnerability that allows any arbitrary origin to access the resources by reflecting the origin header without validation:

No Origin Validation: In the processPreflight and processRequest methods, the code takes the origin directly from the request headers and reflects it back in the response without checking if it's allowed.

Missing Origin Whitelist: There is no whitelist of allowed origins or any validation mechanism.

Lack of Security Controls: The implementation doesn't follow standard CORS security practices, which recommend validating origins against a whitelist.

Impact of This Vulnerability
This vulnerability could allow attackers to:

Execute cross-site request forgery (CSRF) attacks
Access sensitive data through cross-origin requests Interact with the API from unauthorized websites
The Fix I've Implemented
I've made the following security improvements:

Added Origin Validation:

Added an isOriginAllowed() method to validate origins against a whitelist Only set CORS headers if the origin is allowed
Added Origin Configuration:

Added configuration properties for allowed origins Added processing of a comma-separated list of allowed origins Added support for a wildcard "*" to allow any origin (with warnings) Default to localhost if not configured
Improved Request Method Validation:

Added validation of the requested methods against allowed methods Only set CORS headers if both origin and method are valid Added Better Logging:

Added informative logs about the CORS configuration Added security warnings for permissive configurations Added Proper Vary Header:

Consistently added the Vary header to both preflight and regular requests This helps browsers properly cache the responses
These changes bring the implementation in line with CORS security best practices while maintaining backward compatibility with existing functionality.

References:
https://github.com/apache/tomcat/commit/52382ebfbce20a98b01cd9d37184a12703987a5a
https://nvd.nist.gov/vuln/detail/CVE-2025-24813

### DIFF
--- a/bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/internal/filter/CorsFilter.java
+++ b/bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/internal/filter/CorsFilter.java
@@ -100,6 +100,11 @@ public class CorsFilter implements ContainerResponseFilter {
      *
      * @param requestContext
      * @param responseContext
+     */    /**
+     * Process the CORS request and response.
+     *
+     * @param requestContext The request context
+     * @param responseContext The response context
      */
     private void processRequest(ContainerRequestContext requestContext, ContainerResponseContext responseContext) {
         // Process the request only if if is an acceptable request method and if it is different from an OPTIONS request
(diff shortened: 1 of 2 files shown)


kaikreuzer:
The code looks generated by a very bad coding assistant, since it breaks a lot of things.
If there is indeed a critical vulnerability that needs to be fixed, I'd wish for a PR that looks trustworthy and which shows that the author knows what he is doing.

rkoshak:
CVE-2025-24813 is against Tomcat, not Jetty or Karaf, as I understand it. That doesn't necessarily mean it doesn't also exist in OH, but that should be verified if this is picked up again. It's a 9.8 vulnerability so if it does exist in OH it should be addressed.

PR #4733, opened 2025-04-20
Tweak x-axis labels of charts

holgerfriedrich:
* Avoid overlapping of labels

Fixes #4685, which is caused by the recent upgrade of XChart library.

This might need some more iterations for special cases.
In my tests is worked fine on different resolutions and DPI settings.

Issue #4685, opened 2025-03-27
Overlapping axis labels in Sitemap graphs in openHAB App


dandjo:
## Expected Behavior

Readable aligned horizontal labeling in graphs.

## Current Behavior

Overlapping text in labels in graphs on the x-axis using openHAB App (latest from Play-Store on Android).

![Image](https://github.com/user-attachments/assets/07c4989f-e863-4759-bf69-6f59ac9b2805)

## Your Environment
* Version used: openHAB 5.0.0 Build 4558
```
runtimeInfo:
  version: 5.0.0
  buildString: "Build #4558"
locale: en-AT
systemInfo:
  configFolder: /etc/openhab
  userdataFolder: /var/lib/openhab
  logFolder: /var/log/openhab
  javaVersion: 21.0.6
  javaVendor: Ubuntu
  osName: Linux
  osVersion: 6.8.0-1020-raspi
  osArchitecture: aarch64
  availableProcessors: 4
  freeMemory: 311492568
  totalMemory: 471859200
  uptime: 1310
  startLevel: 100
addons:
  - automation-jsscripting
  - binding-enigma2
  - binding-http
  - binding-mielecloud
  - binding-modbus
  - binding-mqtt
  - binding-netatmo
  - binding-tado
  - misc-openhabcloud
  - persistence-influxdb
  - persistence-mapdb
  - persistence-rrd4j
  - transformation-jsonpath
  - transformation-regex
  - ui-basic
clientInfo:
  device:
    ios: false
    android: false
    androidChrome: false
    desktop: true
    iphone: false
    ipod: false
    ipad: false
    edge: false
    ie: false
    firefox: false
    macos: false
    windows: false
    cordova: false
    phonegap: false
    electron: false
    nwjs: false
    webView: false
    webview: false
    standalone: false
    pixelRatio: 1
    prefersColorScheme: dark
  isSecureContext: false
  locationbarVisible: true
  menubarVisible: true
  navigator:
    cookieEnabled: true
    deviceMemory: N/A
    hardwareConcurrency: 12
    language: en-GB
    languages:
      - en-GB
      - en-US
      - en
      - de
    onLine: true
    platform: Linux x86_64
  screen:
    width: 3440
    height: 1440
    colorDepth: 24
  support:
    touch: false
    pointerEvents: true
    observer: true
    passiveListener: true
    gestures: false
    intersectionObserver: true
  themeOptions:
    dark: dark
    filled: true
    pageTransitionAnimation: default
[...]

PR #4830, opened 2025-05-28
Tweak x-axis labels of charts

holgerfriedrich:
More tuning to avoid overlapping of labels, follow-up to #4733

This is increases the spacing a bit to handle the issue shown in
https://github.com/openhab/openhab-core/issues/4685#issuecomment-2893957503

I don't want to increase it too much to avoid getting only very few labels for other DPI settings.


holgerfriedrich:
@kaikreuzer could we please include this before the milestone?

PR #4839, opened 2025-06-04
Resolve runbundles of linuxsysfs itest

holgerfriedrich:
Itests are still failing silently.
This seems to happen if the feature verification passes, but the itest has issues resolving bundles.
Github hides this output and marks the test as passed.

You can find the output in Jenkins logs, just look for `Unresolved requirement` in 
https://ci.openhab.org/job/openHAB-Core/154/consoleText


wborn (approved):
LGTM! 👍

PR #4843, opened 2025-06-08
Use gmavenplus-plugin

holgerfriedrich:
* Replace groovy-maven-plugin by gmavenplus-plugin
* Upgrade groovy-all from 3.0.25 to 4.0.27

groovy-maven-plugin has not been updated for over 5 years, see https://mvnrepository.com/artifact/org.codehaus.gmaven/groovy-maven-plugin/2.1.1.

This change is only visible if profile `generate-tag-classes` is set and will not be validated by CI.

Refs: openhab/openhab-distro#1756

### DIFF
--- a/bundles/org.openhab.core.semantics/pom.xml
+++ b/bundles/org.openhab.core.semantics/pom.xml
@@ -45,19 +45,14 @@
       <build>
         <plugins>
           <plugin>
-            <groupId>org.codehaus.gmaven</groupId>
-            <artifactId>groovy-maven-plugin</artifactId>
-            <version>2.1.1</version>
+            <groupId>org.codehaus.gmavenplus</groupId>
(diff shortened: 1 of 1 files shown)


andrewfg:
> This change is only visible if profile `generate-tag-classes` is set and will not be validated by CI.

@jimtng as the expert on that profile script, could you please check this?

jimtng:
I tried it and it seems to work fine, as I'm sure @holgerfriedrich had tried it also.

Excerpt of log output:

```
[INFO] Using Groovy 4.0.27 to perform execute.
[INFO] Running Groovy script from /home/openhab/projects/openhab-core/bundles/org.openhab.core.semantics/model/generateTagClasses.groovy.
Processing Tag Indoor
Processing Tag Apartment
...
```

kaikreuzer (approved):

PR #4847, opened 2025-06-15
Set property oh.java.version

holgerfriedrich:
Property `oh.java.version` was removed in #4161, but is still used in `features/karaf/pom.xml`.
With some configurations of the dev machine, this leads to wrong javase being pulled in, causing the verification to fail.

This PR reverts this part of #4161.


kaikreuzer (approved):
Thanks!

PR #4848, opened 2025-06-15
[addonservice] Use valid component name

holgerfriedrich:
This fixes the following warning during build:

```
[INFO] --- bnd:7.1.0:bnd-process (default) @ org.openhab.core.addon ---
[WARNING] openhab-core\bundles\org.openhab.core.addon\pom.xml [0:0]: Invalid component name: JAR-File add-on service
```
Other component names do not contain spaces, I changed that to get rid of the warning.

Not sure about possible side effects.


kaikreuzer (approved):
Thanks!

PR #4849, opened 2025-06-15
Skip antlr-generator-3.2.0-patch.jar download

wborn:
Uses antlr-generator-3.2.0-patch.jar in the repository as it currently can no longer be downloaded causing builds to fail.

See also: https://github.com/eclipse-xtext/xtext/issues/3443


holgerfriedrich (approved):

wborn:
If I revert f695acfc4c45f99508133a12086cb90a66897ac4 the test works. So I think the failing itests are caused by https://github.com/openhab/openhab-core/pull/4513.

holgerfriedrich:
Agreed, lets merge this and deal with the itest separately.

PR #4856, opened 2025-06-17
Fix itests after #4513

florian-h05:
Fixes itest after #4513.

I do not understand why the new ScriptedItemProvider is returned when retrieving the ItemProvider service, as this did not happen with the ManagedItemProvider and the YamlItemProvider. Anyway, this change fixes the problem by explicitly retrieving the GenericItemProvider.


holgerfriedrich (approved):
Thanks for fixing this test - hopefully we can get a green nightly again.

Do you expect side effects, do we need to investigate further what is going on?

florian-h05:
> Do you expect side effects, do we need to investigate further what is going on?

No, since I cannot find any places where `@Reference ItemProvider` or `getService(ItemProvider.class)` is used to gain access to an ItemProvider implementation (I also can't imagine a use case for that) I don't expect side effects.

PR #4859, opened 2025-06-19
NonNullByDefault for all classes

holgerfriedrich:
This basically removes all CheckStyle warnings "NullAnnotationsCheck: Classes/Interfaces/Enums should be annotated with `@NonNullByDefault`.
Annotations are now there for all "non-DTO" classes.

It was not an easy task to get everything sound and accepted by the compiler, without requiring bigger changes in the add-ons or webui repo (of course cannot tell much about other bindings, not part of the repo, but breaking things should be avoided if possible.)

The biggest decision to be made in this PR is then handling of properties / `Configuration` data. At some places, this seems to be written in a way values could be `@Nullable`, but most of the time is seems used only with non-null values. I played both to the end, leading to much more efforts getting the non-null for values in place (not to speak about the add-ons repo, here many bindings just did not compile anymore).

This is not somehow only half part of the story, leaving a lot of warnings about null mismatches or dead code during compilation. But for now, I tried the keep the impact to active code as minimal as possible (I have a feeling this should be handled in a separate PR).

Just to give you an impression how the SAT report now looks like (including #4863 and using the unreleased snapshot of SAT which does no longer warn about SimplifyBooleanReturns:
grafik

### DIFF
--- a/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/TypeIntrospections.java
+++ b/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/TypeIntrospections.java
@@ -15,6 +15,8 @@ package org.openhab.core.config.core.internal.validation;
 import java.math.BigDecimal;
 import java.util.Map;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.config.core.ConfigDescriptionParameter.Type;
 
 /**
@@ -23,6 +25,7 @@ import org.openhab.core.config.core.ConfigDescriptionParameter.Type;
  *
  * @author Thomas Höfer - Initial contribution
  */
+@NonNullByDefault
 final class TypeIntrospections {
 
     private static final Map<Type, TypeIntrospection> INTROSPECTIONS = Map.of( //
@@ -59,15 +62,15 @@ final class TypeIntrospections {
     abstract static class TypeIntrospection {
 
         private final Class<?> clazz;
-        private final MessageKey minViolationMessageKey;
-        private final MessageKey maxViolationMessageKey;
+        private final @Nullable MessageKey minViolationMessageKey;
+        private final @Nullable MessageKey maxViolationMessageKey;
 
         private TypeIntrospection(Class<?> clazz) {
             this(clazz, null, null);
         }
 
-        private TypeIntrospection(Class<?> clazz, MessageKey minViolationMessageKey,
-                MessageKey maxViolationMessageKey) {
+        private TypeIntrospection(Class<?> clazz, @Nullable MessageKey minViolationMessageKey,
+                @Nullable MessageKey maxViolationMessageKey) {
             this.clazz = clazz;
             this.minViolationMessageKey = minViolationMessageKey;
             this.maxViolationMessageKey = maxViolationMessageKey;
@@ -81,7 +84,7 @@ final class TypeIntrospections {
          *
          * @return true, if the given value is less than the given min attribute, otherwise false
          */
-        boolean isMinViolated(Object value, BigDecimal min) {
+        boolean isMinViolated(Object value, @Nullable BigDecimal min) {
             if (min == null) {
                 return false;
             }
@@ -133,7 +136,7 @@ final class TypeIntrospections {
          *
          * @return true, if the given value is a string, otherwise false
          */
-        final boolean isStringInstance(Object value) {
+        final boolean isStringInstance(@Nullable Object value) {
             return value instanceof String;
         }
 
@@ -144,7 +147,7 @@ final class TypeIntrospections {
          *
          * @return true, if the given value is a big decimal, otherwise false
          */
-        final boolean isBigDecimalInstance(Object value) {
+        final boolean isBigDecimalInstance(@Nullable Object value) {
             return value instanceof BigDecimal;
         }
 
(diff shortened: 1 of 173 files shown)


bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/internal/validation/TypeIntrospections.java
```diff
@@ -60,15 +63,17 @@ static TypeIntrospection get(Type type) {
     abstract static class TypeIntrospection {
 
         private final Class<?> clazz;
+        @Nullable
         private final MessageKey minViolationMessageKey;
+        @Nullable
         private final MessageKey maxViolationMessageKey;
```
wborn:
```suggestion
        private final @Nullable MessageKey minViolationMessageKey;
        private final @Nullable MessageKey maxViolationMessageKey;
```

bundles/org.openhab.core.io.transport.modbus/src/main/java/org/openhab/core/io/transport/modbus/internal/ModbusManagerImpl.java
```diff
@@ -308,7 +307,8 @@ public void accept(AggregateStopWatch timer, WriteTask task, ModbusSlaveConnecti
             .accept(new ModbusSlaveEndpointVisitor<EndpointPoolConfiguration>() {
 
                 @Override
-                public @NonNull EndpointPoolConfiguration visit(ModbusTCPSlaveEndpoint modbusIPSlavePoolingKey) {
+                @Nullable
+                public EndpointPoolConfiguration visit(ModbusTCPSlaveEndpoint modbusIPSlavePoolingKey) {
```
wborn:
```suggestion
                public @Nullable EndpointPoolConfiguration visit(ModbusTCPSlaveEndpoint modbusIPSlavePoolingKey) {
```

holgerfriedrich:
TODO:
- [X] check if webui repo still validates
- [x] check if add-ons repo still validates (PR openhab/openhab-addons#18915)
- [x] review changes again
- [x] TODOs, i.e. 2 disabled test cases
- [ ] review build warnings
- [x] test OH distro built with that

lsiepel:
> This is not somehow only half part of the story, leaving a lot of warnings about null mismatches or dead code during compilation. But for now, I tried the keep the impact to active code as minimal as possible (I have a feeling this should be handled in a separate PR).

Totally agree. Otherwise the change would be way too big and will never finish. 

Can't fully asses the impact on addons, but i guess it will be big. I hope we can merge this quickly, together with the absolute minimum to addons. As there will be a conflict nightmare if we keep it open for a while. I was also about the start a series of PR's to reduce the annotation warnings in addons, but wil wait now to see how this PR evolves.

We should also communicatie clearly to marketplace contributors. While the API won;t change, they will need to take account of the added constrains.

Another option - i just want to mention - is that we first add annotations to the addon repo, and when finished, we check the (limited) impact of this PR on that repo? 

Regarding the decision about handling of properties, i don't have much to add. Unless someone has a brilliant idea, it will be a bit painfull to add annotations and/or Objects.requireNonNull etc.

holgerfriedrich:
Just a quick reply on this:
 
> Can't fully asses the impact on addons, but i guess it will be big. I hope we can merge this quickly, together with the absolute minimum to addons. As there will be a conflict nightmare if we keep it open for a while. I was also about the start a series of PR's to reduce the annotation warnings in addons, but wil wait now to see how this PR evolves.

The PR to get add-ons to compile after this affects only 70 files, openhab/openhab-addons#18915. So more than 85% of the add-ons could stay unchanged. There might be more compiler warnings all over the repo, I must admit.


> 
> We should also communicatie clearly to marketplace contributors. While the API won;t change, they will need to take account of the added constrains.

Good point.

holgerfriedrich:
Finally, this one is ready for review.
@lsiepel FYI @jlaur 

I am not sure if I took the correct decisions in all places, but as discussed above, this is the direction with a viable number of changes to add-ons as well.

I did not follow-up the warnings nor did I remove the now unneeded if null checks.
As said before, I would introduce that in a second step.

Any feedback is welcome!

lsiepel:
@holgerfriedrich i like to look at this, but it is also massive. Can you rebase? I'll have some time to have a closer look at this the next few weeks.

holgerfriedrich:
@lsiepel I fear this has to be delayed for the next few weeks. I have a few other things in the queue, touching a lot of files. I am trying to do a Karaf 4.5 port including the switch to Jakarta and Jetty 12 - this keeps me busy and drives me crazy ;-)
On top, I have an incomplete port to the new MapDB (upgrader still not ready), and a change of the Modbus library to something modern...

After that, we could maybe revive this PR - but likely not in one piece but split to smaller chunks.

lsiepel:
Ok, good to know. I was just gathering some courage to start reviewing this. The other things you mention, especially karaf 4.5 would be awesome. Maybe not the right place to discuss this. But as 4.5 is such a big change, i think it will be smart to move that to openHAB 5.3 (and maybe jump version to 5.5.x to highlight the significant change)

jlaur:
> Maybe not the right place to discuss this. But as 4.5 is such a big change, i think it will be smart to move that to openHAB 5.3 (and maybe jump version to 5.5.x to highlight the significant change)

Or 6.0 with Java 25. 😄

holgerfriedrich:
@jlaur Java 25 is not yet possible with Karaf snapshot, but still planned for 4.5. SecurityManager still needs to be removed in snapshot. But let's say 2/3 big changes are done (Jetty and Javax to Jakarta).

PR #4862, opened 2025-06-21
Upgrade Jollyday to 1.5.4

wborn:
Upgrades Jollyday from 1.5.0 to 1.5.4.

For release notes, see:

https://github.com/focus-shift/jollyday/releases

### DIFF
--- a/bom/compile/pom.xml
+++ b/bom/compile/pom.xml
@@ -360,7 +360,7 @@
     <dependency>
       <groupId>de.focus-shift</groupId>
       <artifactId>jollyday-jackson</artifactId>
-      <version>1.5.0</version>
+      <version>1.5.4</version>
       <scope>compile</scope>
     </dependency>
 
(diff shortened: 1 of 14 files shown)


holgerfriedrich (approved):
LGTM

holgerfriedrich:
@wborn Do you think we should backport this to 4.3.x? I assume, 4.3.x will be longer in use than other previous releases due to 32/64 bit transition. WDYT?

wborn:
It may help everyone a little bit and probably won't hurt.

holgerfriedrich:
This would mean changes to all repos (feature, and bndrun if we want the I tests to work, demo app); and we need to upload the new lib to jfrog.

Do you expect problems as we may break marketplace add-ons for 4.3.x?

wborn:
It will probably be fine. But I haven't seen anyone commenting about any issues with Jollyday in 4.3.x and if so there are ways to override the default holidays. So in the end it may not be worth the hassle for a patch release.

holgerfriedrich:
@wborn Thanks for the assesment.

I think I will not continue backporting this for now.

PR #4866, opened 2025-06-22
Refactor timestamps to `Instant` in discovery methods

jlaur:
Continuation of #4026, this time finishing the job by changing the API.

Methods related to discovery now use `Instant` rather than `long` for timestamps. This provides type safety, clarity and integration with the modern `java.time` API.

I have kept all existing methods, but immediately deprecated them. The `Instant` overload for `removeOlderResults` in the `DiscoveryListener` interface has a default implementation delegating to the `long` overload, so it is non-breaking now since an implementation is not required.

The proposed next phases:
1. Refactor add-ons to use the new methods. Prepared here: openhab/openhab-addons#18838.
2. Minor release: Invert the delegation: Make the `long` method a default method delegating to `Instant`.
   Make the `Instant` method abstract — implementors must now provide it.
3. Major release: Remove the deprecated `long` method entirely from the interface.
   Implementors who didn’t remove `@Override` on the legacy method in previous phase will get a compile error.

### DIFF
--- a/bundles/org.openhab.core.config.discovery/src/main/java/org/openhab/core/config/discovery/AbstractDiscoveryService.java
+++ b/bundles/org.openhab.core.config.discovery/src/main/java/org/openhab/core/config/discovery/AbstractDiscoveryService.java
@@ -352,8 +352,22 @@ public abstract class AbstractDiscoveryService implements DiscoveryService {
      * as timestamp.
      *
      * @param timestamp timestamp, older results will be removed
+     * @deprecated Use {@link #removeOlderResults(Instant)} instead.
      */
+    @Deprecated(since = "5.0", forRemoval = true)
     protected void removeOlderResults(long timestamp) {
+        removeOlderResults(Instant.ofEpochMilli(timestamp));
+    }
+
+    /**
+     * Call to remove all results of all {@link #supportedThingTypes} that are
+     * older than the given timestamp. To remove all left over results after a
+     * full scan, this method could be called {@link #getTimeOfLastScan()}
+     * as timestamp.
+     *
+     * @param timestamp timestamp, older results will be removed
+     */
+    protected void removeOlderResults(Instant timestamp) {
         removeOlderResults(timestamp, null, null);
     }
 
@@ -365,8 +379,23 @@ public abstract class AbstractDiscoveryService implements DiscoveryService {
      *
      * @param timestamp timestamp, older results will be removed
      * @param bridgeUID if not {@code null} only results of that bridge are being removed
+     * @deprecated Use {@link #removeOlderResults(Instant, ThingUID)} instead.
      */
+    @Deprecated(since = "5.0", forRemoval = true)
     protected void removeOlderResults(long timestamp, @Nullable ThingUID bridgeUID) {
+        removeOlderResults(Instant.ofEpochMilli(timestamp), null);
+    }
+
+    /**
+     * Call to remove all results of all {@link #supportedThingTypes} that are
+     * older than the given timestamp. To remove all left over results after a
+     * full scan, this method could be called {@link #getTimeOfLastScan()}
+     * as timestamp.
+     *
+     * @param timestamp timestamp, older results will be removed
+     * @param bridgeUID if not {@code null} only results of that bridge are being removed
+     */
+    protected void removeOlderResults(Instant timestamp, @Nullable ThingUID bridgeUID) {
         removeOlderResults(timestamp, null, bridgeUID);
     }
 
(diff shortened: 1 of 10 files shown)


jlaur:
In the addons repo there are many usages like this:
```java
removeOlderResults(getTimestampOfLastScan());
```

I'm tempted to suggest skipping all these proposed compatibility phases and instead prepare full addons PR. Usages like this one wouldn't need to be changed at all, whereas with the multi-phased approach I need to use a new name for `getTimestampOfLastScan`:
```java
removeOlderResults(getTimeOfLastScan());
```

The downside is that it also affects bindings in other repositories. I guess they still need to be adapted to 5.0 anyway, and the required changes would be simple and few.

WDYT?

**EDIT**: The only breaking change I would actually propose to do initially is changing the return type of `getTimeOfLastScan()` to `Instant` to avoid introducing another method with a different name, leading to many changes as a consequence. I believe this can be done in a safe way - see openhab/openhab-addons#18845. After merging that PR, nothing should break. Let me know if you agree, and I can adapt this PR to remove `getTimeOfLastScan()` again and change `getTimestampOfLastScan()` to return `Instant`.

jlaur:
openhab/openhab-addons#18845 has been merged, so I could go ahead and adapt this PR to remove `getTimeOfLastScan()` again and change `getTimestampOfLastScan()` to return `Instant`. Would that be acceptable?

wborn:
> So I could go ahead and adapt this PR to remove getTimeOfLastScan() again and change getTimestampOfLastScan() to return Instant. Would that be acceptable?

If you make breaking changes you might as well not keep the deprecated interfaces at all. I'm fine with that as long as the code in all OH repos keeps compiling. :slightly_smiling_face:

jlaur:
> > So I could go ahead and adapt this PR to remove getTimeOfLastScan() again and change getTimestampOfLastScan() to return Instant. Would that be acceptable?

> If you make breaking changes you might as well not keep the deprecated interfaces at all. I'm fine with that as long as the code in all OH repos keeps compiling. 🙂

I have now made the breaking change for `getTimestampOfLastScan()` so that it now returns `Instant`. With openhab/openhab-addons#18845 already merged, I believe the PR in its current form will not break openhab-addons or the zigbee/zwave repos.

I have verified by building a few bindings, both some of those prepared by openhab/openhab-addons#18845 and those using `removeOlderResults(getTimeOfLastScan());`. However, I'm currently facing some issues with a full addons build:
```
[INFO] --- karaf:4.4.7:verify (karaf-feature-verification) @ org.openhab.automation.groovyscripting ---
[INFO] Using repositories: https://openhab.jfrog.io/openhab/libs-release@id=openhab-release,https://openhab.jfrog.io/openhab/libs-snapshot@id=openhab-snapshot@noreleases@snapshots,https://repo1.maven.org/maven2@id=central
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  EXCEPTION_ACCESS_VIOLATION (0xc0000005) at pc=0x00007ffd6da8d536, pid=9168, tid=16980
#
# JRE version: Java(TM) SE Runtime Environment (21.0.5+9) (build 21.0.5+9-LTS-239)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (21.0.5+9-LTS-239, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, windows-amd64)
# Problematic frame:
# V  [jvm.dll+0x13d536]
```

so I have been so far unable to fully verify.

I agree with your suggestion to not keep the deprecated interfaces at all, but with this approach there will be a strict dependency between the repos, so this PR would break addons until openhab/openhab-addons#18838 is adapted and merged as well. I can do that, but would it be acceptable as well to go through the iterations in an accelerated way, so that this PR is merged first, and I will then immediately prepare the next steps so we'll reach the same result within the next day or two? I think this will reduce the risk a bit and put less stress on me. 😄 Both approaches are fine me with as well though.

jlaur:
> so I have been so far unable to fully verify.

I'm running a full build now after upgrading to JDK 21.0.7, which seems to work better.

jlaur:
> > so I have been so far unable to fully verify.
> 
> I'm running a full build now after upgrading to JDK 21.0.7, which seems to work better.

Everything seems fine so far, although I had to create openhab/openhab-addons#18890 because a new binding got merged in the meantime. In any case, that PR must be merged before this one.

wborn:
> this PR is merged first, and I will then immediately prepare the next steps so we'll reach the same result within the next day or two? 

Sound good to me!

wborn (approved):
Thanks for the refactoring!

jlaur:
> > this PR is merged first, and I will then immediately prepare the next steps so we'll reach the same result within the next day or two?
> 
> Sound good to me!

Excellent, thanks! Thanks for the review and merge - perhaps you can trigger a new core build, so that I can proceed with the addons changes and make the build pass?

wborn:
Yes I already triggered it. 😉

lolodomo:
@jlaur : looks like one of your very recent changes in core framework broke the remote openHAB binding:
https://community.openhab.org/t/openhab-5-0-snapshot-discussion/161404/26
It could be this PR or one of few linked PRs.

This PR should be marked as breaking change, no ?

Edit: I have now a doubt, it could have no link with your recent changes, it is just the REST API to get thing that is failing for the user, maybe because it did not set properly the required token.

Nadahar:
There is a regression here - I just noticed that the discovery timestamps in the JSONDB and in the log are in the wrong timezone. While `Date` "cheats" and uses the JVM default timezone in `toString()` (because internally it too is in UTC), `Instant` doesn't - and the result is that the stored and logged information doesn't make sense to those that don't happen to use UTC.

PR #4867, opened 2025-06-23
[semantic tags] Add undefined point, equipment, location tags

andrewfg:
This PR adds generic undefined Point, Equipment, and Location tags. (For backwards compatibility).


andrewfg:
Ping @jimtng / @mherwege

mherwege:
These tags were never there, so I don’t think you should introduce them. I believe past behavior is still preserved, but do need some time to check.

jimtng:
This seems like the wrong solution, although I am still catching up and trying to understand what the problem is.

mherwege:
This is a solution to a non-existing problem. I called out `Location`, `Equipment`, `Point` and `Property` are valid semantic tags. Nothing has changed in the code accepting these as valid tags (`SemanticsMetadataProvider.java`), so this should keep working as before without any required change. In my view, this should be closed.

andrewfg:
> This is a solution to a non-existing problem

PR #4868, opened 2025-06-24
Move expire's Duration parser to a separate utility class and support `ms` for milliseconds

jimtng:
so it can be used elsewhere

This was prompted by the need to standardize Duration parsing across openHAB, including addons, such as in https://github.com/openhab/openhab-addons/pull/18301

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/util/DurationUtils.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/util/DurationUtils.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright (c) 2010-2025 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.core.util;
+
+import java.time.Duration;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
+/**
+ * Static utility methods that are helpful when dealing with {@link Duration}.
+ *
+ * @author Jimmy Tanagra - Initial contribution
+ */
+@NonNullByDefault
+public class DurationUtils {
+
+    private static final Pattern DURATION_PATTERN = Pattern
+            .compile("(?:([0-9]+)D)?\\s*(?:([0-9]+)H)?\\s*(?:([0-9]+)M)?\\s*(?:([0-9]+)S)?", Pattern.CASE_INSENSITIVE);
+
+    /**
+     * Parses a duration string in ISO-8601 format or a custom format like
+     * "1d 1h 15m 30s" where 'd' stands for days, 'h' for hours,
+     * 'm' for minutes, and 's' for seconds.
+     *
+     * @param durationString the string representation of the duration
+     * @return a Duration object representing the parsed duration
+     * @throws IllegalArgumentException if the string cannot be parsed into a valid duration
+     */
+    public static Duration parse(String durationString) throws IllegalArgumentException {
+        try {
+            return Duration.parse(durationString);
+        } catch (Exception e) {
+            // ignore
+        }
+
+        Matcher m = DURATION_PATTERN.matcher(durationString);
+        if (!m.matches() || (m.group(1) == null && m.group(2) == null && m.group(3) == null && m.group(4) == null)) {
+            throw new IllegalArgumentException(
+                    "Invalid duration: " + durationString + ". Expected something like: '1d 1h 15m 30s'");
+        }
+
+        Duration duration = Duration.ZERO;
+        if (m.group(1) != null) {
+            duration = duration.plus(Duration.ofDays(Long.parseLong(m.group(1))));
+        }
+        if (m.group(2) != null) {
+            duration = duration.plus(Duration.ofHours(Long.parseLong(m.group(2))));
+        }
+        if (m.group(3) != null) {
+            duration = duration.plus(Duration.ofMinutes(Long.parseLong(m.group(3))));
+        }
+        if (m.group(4) != null) {
+            duration = duration.plus(Duration.ofSeconds(Long.parseLong(m.group(4))));
+        }
+        return duration;
+    }
+}
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/util/DurationUtils.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/util/DurationUtils.java
@@ -13,6 +13,7 @@
 package org.openhab.core.util;
 
 import java.time.Duration;
+import java.time.temporal.ChronoUnit;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -26,13 +27,20 @@ import org.eclipse.jdt.annotation.NonNullByDefault;
 @NonNullByDefault
 public class DurationUtils {
 
-    private static final Pattern DURATION_PATTERN = Pattern
-            .compile("(?:([0-9]+)D)?\\s*(?:([0-9]+)H)?\\s*(?:([0-9]+)M)?\\s*(?:([0-9]+)S)?", Pattern.CASE_INSENSITIVE);
+    private static final Pattern DURATION_PATTERN = Pattern.compile(
+            "(?:([0-9]+)D)?\\s*(?:([0-9]+)H)?\\s*(?:([0-9]+)M)?\\s*(?:([0-9]+)S)?\\s*(?:([0-9]+)MS)?",
+            Pattern.CASE_INSENSITIVE);
+    private static final ChronoUnit[] DURATION_UNITS = { ChronoUnit.DAYS, ChronoUnit.HOURS, ChronoUnit.MINUTES,
+            ChronoUnit.SECONDS, ChronoUnit.MILLIS };
 
     /**
      * Parses a duration string in ISO-8601 format or a custom format like
-     * "1d 1h 15m 30s" where 'd' stands for days, 'h' for hours,
-     * 'm' for minutes, and 's' for seconds.
+     * "1d 1h 15m 30s 500ms" where
+     * 'd' stands for days,
+     * 'h' for hours,
+     * 'm' for minutes,
+     * 's' for seconds, and
+     * 'ms' for milliseconds.
      *
      * @param durationString the string representation of the duration
      * @return a Duration object representing the parsed duration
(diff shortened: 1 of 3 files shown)


andrewfg (approved):
LGTM

mherwege:
At one point, I introduced formatting for durations in QuantityType (https://github.com/openhab/openhab-core/pull/4169). Would it make sense to also move this to this utility class? That way we may be able to keep it all together and use it everywhere in the same way.

rkoshak:
I'd really like to see support for ISO8601 in a utility like this. It's slightly more flexible compared to Expire's format and it's an official standard. Though it is slightly less clear because of the whole PT stuff.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/DurationUtils.java
```diff
@@ -0,0 +1,73 @@
...
+     * 'm' for minutes,
+     * 's' for seconds, and
+     * 'ms' for milliseconds.
+     *
+     * @param durationString the string representation of the duration
+     * @return a Duration object representing the parsed duration
+     * @throws IllegalArgumentException if the string cannot be parsed into a valid duration
+     */
+    public static Duration parse(String durationString) throws IllegalArgumentException {
+        try {
```
andrewfg, line 50: Suggested addition:

```
        if (QuantityType.valueOf(durationString) instanceof QuantityType<?> time
                && time.toUnit(MetricPrefix.MILLI(Units.SECOND)) instanceof QuantityType<?> msec) {
            return Duration.ofMillis(msec.longValue());
        }
```

andrewfg: See https://github.com/openhab/openhab-addons/pull/18301#discussion_r2164592436

jimtng: From what I can tell, the difference between QuantityType parsing of duration vs expire's current format:
- `1 h` vs `1h` - QuantityType supports spaces between the numbers and the unit
- `1 min` vs `1m` - QuantityType uses `min` instead of `m` (which would be meter, unit of length in QuantityType world)

So to "support" QuantityType formatting, we can just support spaces between number and unit and `min` in addition to `m` - or actually get QuantityType to try to parse it. But is this going to be a gain for specifying duration strings? I think it's a loss of clarity and added confusion and inconsistencies:

- You can use `1 min` but not `1 hr` nor `1 hour` nor `1 sec`

@mherwege @rkoshak wdyt?

By supporting QuantityType's syntax, you're declaring that "we support QuantityType syntax, and also the old expire syntax for specifying durations", but then "We don't support the old expire syntax when parsing it with QuantityType.valueOf". That seems inconsistent.

I'd rather separate QuantityType and Duration parsing completely, and use Duration when you need a Duration.

jimtng:
> At one point, I introduced formatting for durations in QuantityType (#4169). Would it make sense to also move this to this utility class? That way we may be able to keep it all together and use it everywhere in the same way.

Are you thinking of `public String DurationUtils.format(Duration value, String format)`? In any case it should be a separate follow up PR.



> I'd really like to see support for ISO8601 in a utility like this. It's slightly more flexible compared to Expire's format and it's an official standard. Though it is slightly less clear because of the whole PT stuff.

ISO8601 is already implemented, and I think in one of your other replies, you've also realised this, but I'd reply here to clarify for other readers.

bundles/org.openhab.core/src/main/java/org/openhab/core/util/DurationUtils.java
jimtng: Furthermore, QuantityType is not a suitable substitute for specifying "durations" that we often expect (machine-based duration, not calendar based durations). For example, QuantityType.valueOf("1 year") converts to 31536000 seconds, but this is not always correct depending on whether the current year has leap year etc.

This is the reason why Java splits its Temporal classes into Period and Duration, with the prior supporting years, months, days, and Duration supporting days, hours, minutes, seconds, etc.

See: https://docs.oracle.com/javase/tutorial/datetime/iso/period.html

mherwege:
> Are you thinking of `public String DurationUtils.format(Duration value, String format)`? In any case it should be a separate follow up PR.

Yes, indeed. And a follow-up PR would be fine.

> By supporting QuantityType's syntax, you're declaring that "we support QuantityType syntax, and also the old expire syntax for specifying durations", but then "We don't support the old expire syntax when parsing it with QuantityType.valueOf". That seems inconsistent.
> 
> I'd rather separate QuantityType and Duration parsing completely, and use Duration when you need a Duration.

I don't think it harms to allow QuantityType syntax for parsing here. Anything allowing a user to use the same syntax in different places is a plus to me and avoids another question on the forum.
I don't think we should try doing it the other way around (QuantityType to support expire syntax), as the syntax in QuantityType is not under our direct control and it would lead to dimension confusion quickly (minute vs metre). I don't see how we can solve that in QuantityType parsing pre-processing without knowing the dimension in advance.
Even if we allow it in the syntax creating a Duration, I wouldn't put it upfront in the documentation (if we document it at all).
I have no big issue ignoring the `1 year` problem with QuantityType and just accepting the QuantityType conversion. This is an unlikely case to be relevant.

jimtng:
QuantityType also supports `1 month` which doesn't translate to 30 * 24 * 60 * 60 

```
irb(main):033> QuantityType.new("1 month").to_unit(Units::SECOND)
=> 2629746 s
irb(main):037> 2629746.0 / 60 / 60 / 24
=> 30.436875
```

So using QuantityType may lead to issues.

If the goal is to make Duration parsing as "user friendly" as possible, I can just add support for 
- Allow spaces, e.g. `1 h`
- d|day|days
- h|hr|hour|hours
- m|min|minute|minutes
- s|sec|second|seconds
- ms|millisecond|milliseconds

Without asking QuantityType to perform the parsing for us.

What other aspects of QuantityType are desired?

andrewfg:
It appears that 1 month is defined as 1/12 of 1 year. (And probably 1 year is defined as 365.25 days). In any case these must be handled in the same way by both the duration parser and the quantity parser. If that is not the case then we need to fix it so that they do both produce identical results.

---

EDIT: and therefore in any case you need to add junit tests (if not already present) to confirm that the two parsers do indeed produce identical results in all cases.

andrewfg:
> It appears that 1 month is defined as 1/12 of 1 year. (And probably 1 year is defined as 365.25 days).

Yup.

```
 package tech.units.indriya.unit;

	/**
	 * A time unit accepted for use with SI units (standard name <code>yr</code> ).
	 */
	public static final Unit<Time> YEAR = AbstractSystemOfUnits.Helper.addUnit(INSTANCE.units,
			Units.DAY.multiply(365.2425), "Year", "yr");

	/**
	 * A unit of duration equal to 1/12 {@link #YEAR} (common name <code>mo</code>).
	 * @since 2.3
	 */
	public static final Unit<Time> MONTH = AbstractSystemOfUnits.Helper.addUnit(INSTANCE.units, 
			YEAR.divide(12), "Month", "mo");
```

---

> 1 month doesn't translate to 30 * 24 * 60 * 60

Indeed. Nor should it ever.

> => 30.436875

Correct:  365.2425 / 12 = 30.436875 days; = 2629746 seconds

jimtng:
> I don't think it harms to allow QuantityType syntax for parsing here. Anything allowing a user to use the same syntax in different places is a plus to me and avoids another question on the forum.

They will ask why 
- `1 min` works but not `1 hr 1 min` (not currently supported by the expire parser)
- `1000000 ns` works, but not `1 s 5000000 ns` doesn't work (expire parser doesn't support ns)
- `200 ns` is accepted, but it doesn't work

It just opens up too many corner cases

mherwege:
@jimtng Fair enough. Keep it separate.

I do like your suggestion of allowing the alternatives as outlined here:

> * Allow spaces, e.g. `1 h`
> * d|day|days
> * h|hr|hour|hours
> * m|min|minute|minutes
> * s|sec|second|seconds
> * ms|millisecond|milliseconds

jimtng:
I intended this PR to be a straight forward refactoring with minimal changes, so that it could get merged quickly.

Further enhancements, such as adding QuantityType support can be added later in a separate PR if you believe that's a good idea.

As for https://github.com/openhab/openhab-addons/pull/18301 I don't see why it can't just support the expiry syntax for now, and then later if DurationUtils.parse supports QuantityType format it can automatically have that too. But that's a discussion that belongs in the addons PR, not here.

jimtng:
> I do like your suggestion of allowing the alternatives as outlined here:
> 
> > * Allow spaces, e.g. `1 h`
> > * d|day|days
> > * h|hr|hour|hours
> > * m|min|minute|minutes
> > * s|sec|second|seconds
> > * ms|millisecond|milliseconds

I can add that in a separate PR later.

andrewfg:
Personally I don't care if you use QuantityType as a sub helper to do the parsing, or whether you write your own code to do it. However I do insist that if you add support for weeks, months, years, etc. then it MUST produce the same results as the indriya (QuantityType) code does. e.g. one month is NOT treated as 30 days, and one year is NOT treated as 365 days. However as far as I can tell you are (currently) stopping at days, so the current code is probably OK.

andrewfg:
PS apropos the expire parameter, the duration should permit being with or without spaces in order to comply with the documentation https://www.openhab.org/docs/configuration/items.html#parameter-expire

rkoshak:
Assuming java's Duration implementation of ISO8601 matches the spec, the value of "P1M" will depend on current context, meaning the amount of time in the current month (i.e. 30 days for June). The value of  P2M" will be the current month plus the next month (i.e. 61 days). The same goes for "P1Y" which takes into account leap days. 

Unless QuantityType does something similar, it is rare that the value from an ISO8601 Duration is going to match QuantityType's 1 month.

So are you basically saying ISO8601 cannot be used at all and QuantityType must become the OH wide standard for the length of 1 month? Beacause we are definitely not going to change the implementation of java.time.Duration to violate the ISO8601 spec.

mherwege:
> Assuming java's Duration implementation of ISO8601 matches the spec, the value of "P1M" will depend on current context, meaning the amount of time in the current month (i.e. 30 days for June). The value of P2M" will be the current month plus the next month (i.e. 61 days). The same goes for "P1Y" which takes into account leap days.
> 
> Unless QuantityType does something similar, it is rare that the value from an ISO8601 Duration is going to match QuantityType's 1 month.
> 
> So are you basically saying ISO8601 cannot be used at all and QuantityType must become the OH wide standard for the length of 1 month? Beacause we are definitely not going to change the implementation of java.time.Duration to violate the ISO8601 spec.

The Java `Duration` class only parses part of the ISO8601 spec (days and time: `PnDTnHnMn.nS`), whereby days are treated as exactly 24h. It avoids the month and year problem. That's where the `Period` class comes into play. You cannot know what the exact duration is of a period if you do not use it relative to a base date (adding/substracting to/from a date). I don't think an ISO  string with years or months is accepted for expiry configuration at this point in time.

andrewfg:
.. whereas indriya defines one month as exactly 365.2425 days divided by 12 -- which is therefore invariant.

rkoshak:
Interesting. JS uses jods-js whose Duration implementation does provide full ISO8601 support. I can't imagine anyone would use weeks or months for a timeout in OH so the whole discussion is probably moot. Maybe just don't support anything more than days and call it quits. 

> .. whereas indriya defines one month as exactly 365.2425 days divided by 12 -- which is therefore invariant.

And it's also wrong for every possible way that humans actually use month. It doesn't match the calendar nor does it match the moon. A month isn't an invariant measurement and forcing it to be such in a user facing interface is not a good idea. 

I'm thinking anything more than weeks should simply not be supported. Since java.time.Duration doesn't support anything more than days that should probably be the max unit for both Expire and the ISO8601. QuantityType can be converted to any unit so it probably doesn't really matter what the max unit it supports is, if QuantityType can parse it, it should be OK. 

But making the Expire String for 1 month be 365.2425 / 12 is wrong. That number doesn't correspond to a month in any of the ways that humans refer to it and this is a human facing interface. One month from now is July 25 at 14:00 MDT, not July 26 at 00:29 MDT.

So I vote, for simplicity and consistency, let's follow the lead of java.time.Duration and simply not support anything more than days, at least for Expire (it's already not supported for Duration apparently). Anything more than a few days is going to be impractical from an OH perspective anyway.

andrewfg:
> it's also wrong for every possible way that humans actually use month

Please take that up with indriya. Which is BTW (one of) the standard(s) that OH adopted.

andrewfg:
FWIW ISO8601 defines Instants. Not Durations. Durations are defined as the delta between two Instants. As the expire property does not define the reference Instant for the Duration calculation, this whole discussion about enforcing ISO8601 is out of context, and irrelevant to this use case, and thus a waste of time. I suggest terminating the discussion. I will anyway unsubsribe.

jimtng:
We'll stop at days, so won't add weeks / months / years.

jimtng:
@mherwege: #4873

mherwege:
> Interesting. JS uses jods-js whose Duration implementation does provide full ISO8601 support.

Looking at the documentation, it provides exactly the same support as the Java Duration class: https://js-joda.github.io/js-joda/class/packages/core/src/Duration.js~Duration.html#static-method-parse

mherwege (approved):
LGTM

rkoshak:
> Looking at the documentation, it provides exactly the same support as the Java Duration class: https://js-joda.github.io/js-joda/class/packages/core/src/Duration.js~Duration.html#static-method-parse

Interesting, I've used `P1W` with it without error before. :man_shrugging: 

Anyway, the only reason I came here was I was asked. I'll be sure to avoid any issue or PR where @andrewfg is participating in the future. It's clear I'm not wanted.

mherwege:
> Interesting, I've used `P1W` with it without error before.

That would probably work because this would return a Period, not a Duration: https://js-joda.github.io/js-joda/class/packages/core/src/Period.js~Period.html#static-method-parse
Both Period and Duration extend TemporalAmount. But there is not parse for TemporalAmount. In calculations with dates, you would use TemporalAmount, which would be a Period or a Duration.

andrewfg:
> I'll be sure to avoid any issue or PR where @andrewfg is participating in the future.

I was still lurking. And @rkoshak I therefore extend my humble apologies for being rude to you.

> I've used P1W with it without error before 

Sure, but you would still need to use invariants if you would use P2Y or P3M

kaikreuzer (approved):
Thanks!

PR #4873, opened 2025-06-26
Allow extended unit names for days, hours, minutes, seconds, milliseconds and allow spaces before the unit in expire duration parser

jimtng:
Extend the expire/duration parser syntax to allow optional spaces between the number and its unit and accept alternative units:
- d|day|days
- h|hr|hrs|hour|hours
- m|min|mins|minute|minutes
- s|sec|secs|second|seconds
- ms|millisecond|milliseconds

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/util/DurationUtils.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/util/DurationUtils.java
@@ -43,7 +43,7 @@ public class DurationUtils {
             ChronoUnit.SECONDS, ChronoUnit.MILLIS };
 
     /**
-     * Parses a duration string in ISO-8601 format or a custom format like
+     * Parses a duration string in ISO-8601 duration format or a custom format like
      * "1d 1h 15m 30s 500ms".
      *
      * When specifying a duration, the units must be specified in the order of
@@ -62,22 +62,29 @@ public class DurationUtils {
      * <li>'ms|millisecond|milliseconds' for milliseconds.
      * </ul>
      *
-     * Examples:
+     * Examples of valid duration strings:
      * <ul>
      * <li>"1h" represents 1 hour
      * <li>"15m" represents 15 minutes
      * <li>"1h15m" represents 1 day and 15 minutes. It can also be written as "1h 15m", "1 h 15 m", "1 hr 15 mins",
      * "1hour 15 minutes", etc.
      * <li>"1d 1h 30s" represents 1 day, 1 hour, and 30 seconds
-     * <li>"PT1H30M" represents 1 hour and 30 minutes
-     * <li>"PT1D" represents 1 day
      * </ul>
      *
-     * The ISO-8601 format is also supported, but only days, hours, minutes, seconds, and milliseconds.
-     * Specifically, years, months, and weeks are not supported.
-     * For example, "PT1H30M" is valid and represents 1 hour and 30 minutes,
-     * however, "P1M" (1 month) is not supported.
-     * The numbers in the ISO-8601 format must be integers.
+     * The ISO-8601 duration format is supported, but only the following units are recognized:
+     * days, hours, minutes, and seconds.
+     * Units such as years, months, and weeks are not supported.
+     * The number of days, hours and minutes must parse to a long.
+     * The number of seconds must parse to a long with optional fraction.
+     * The decimal point may be either a dot or a comma.
+     * The fractional part may have from zero to 9 digits.
+     *
+     * Examples of ISO-8601 durations:
+     * <ul>
+     * <li>"PT1H30M" represents 1 hour and 30 minutes
+     * <li>"PT1D" represents 1 day
+     * <li>"PT0.5S" represents 0.5 seconds (500 milliseconds)
+     * </ul>
      *
      * @param durationString the string representation of the duration
      * @return a Duration object representing the parsed duration
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/util/DurationUtils.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/util/DurationUtils.java
@@ -27,20 +27,29 @@ import org.eclipse.jdt.annotation.NonNullByDefault;
 @NonNullByDefault
 public class DurationUtils {
 
-    private static final Pattern DURATION_PATTERN = Pattern.compile(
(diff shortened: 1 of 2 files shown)


mherwege (approved):
LGTM

bundles/org.openhab.core/src/main/java/org/openhab/core/util/DurationUtils.java
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Static utility methods that are helpful when dealing with {@link Duration}.
+ *
+ * @author Jimmy Tanagra - Initial contribution
+ */
+@NonNullByDefault
+public class DurationUtils {
+
+    private static final Pattern DURATION_PATTERN = Pattern.compile("""
+            (?:([0-9]+)\\s*(?:d|days?))?
```
andrewfg, line 31: If you google "alternate unit symbol for day" (resp. second, hour, minute, etc.) it suggests "day", "sec", "hr", "min" -- without terminating "s"..

andrewfg: .. perhaps the `s?` in the regex does allow them without the terminating s ??

andrewfg: .. to answer my own question .. yes it does => so LGTM

andrewfg (approved):
LGTM

andrewfg:
@jimtng I did not check it but does this also allow spaces? Presumably meaning spaces between the integer and the unit string. https://github.com/openhab/openhab-core/pull/4868#issuecomment-3003710052 .. as `3 days 2 hrs 1 minute` .. and maybe also `3 days, 2 hrs, 1 minute` with commas would make sense here as well..

jimtng:
> @jimtng I did not check it but does this also allow spaces? Presumably meaning spaces between the integer and the unit string. [#4868 (comment)](https://github.com/openhab/openhab-core/pull/4868#issuecomment-3003710052) .. as `3 days 2 hrs 1 minute` .. and maybe also `3 days, 2 hrs, 1 minute` with commas would make sense here as well..

Spaces are allowed and noted in the initial post: 
> Extend the expire/duration parser syntax to allow optional spaces between the number and its unit and

I will add it to PR title too.

Commas might be problematic, because it's used in expire to separate the duration and the options, e.g. `1h,state=OFF`.
So `1h, 1m, state=OFF` might look a bit strange, assuming we can make the parser correctly handle something like `1h,1m,state=abc,def`

andrewfg:
Perfect. LGTM (again).

jimtng:
rebased to main

andrewfg:
I am not sure if this is a big deal but QuantityType can parse `0.5 h` whereas this code falls over..

jimtng:
The doc stated that it supports long, not float.

Do we want to support float? Maybe? But that would be a new thing.

jimtng:
~~`PT0.5H` would work here btw~~ EDIT: No it doesn't support floating point

mherwege:
> `PT0.5H` would work here btw

Are you sure about this? The Javadoc says these parts of the ISO string should parse to a long (except for the seconds part): https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/time/Duration.html#parse(java.lang.CharSequence)

jimtng:
> > `PT0.5H` would work here btw
> 
> Are you sure about this? The Javadoc says these parts of the ISO string should parse to a long (except for the seconds part): https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/time/Duration.html#parse(java.lang.CharSequence)

I stand corrected. it only supports long.

```
jshell> Duration.parse("PT0.5H")
|  Exception java.time.format.DateTimeParseException: Text cannot be parsed to a Duration
|        at Duration.parse (Duration.java:419)
|        at (#2:1)

jshell> Duration.parse("PT5H")
$3 ==> PT5H
```

andrewfg:
Whereas one of the examples on the Oracle doc site shows a case with a float. ???

jimtng:
I have no problem supporting float in the Xh Xm Xs syntax, but perhaps it should go in a separate PR again for granularity. The PR title here is already long as it is.

andrewfg:
I guess either a) allow floats or b) make it clear in the documentation and javadoc that floats are not supported.

mherwege:
> Whereas one of the examples on the Oracle doc site shows a case with a float. ???

Link?

jimtng:
I've added more details in javadoc

andrewfg:
> Link

https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/time/Duration.html#parse(java.lang.CharSequence)

jimtng:
> > Link
> 
> https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/time/Duration.html#parse(java.lang.CharSequence)

Heh, I knew I've seen fractions somewhere before. I didn't pay attention to _which_ part can be fractions:

> The number of days, hours and minutes must parse to a long. The number of seconds must parse to a long with optional fraction. The decimal point may be either a dot or a comma. The fractional part may have from zero to 9 digits.

andrewfg:
But as already as ascertained commas are difficult for other reasons.

mherwege:
> Are you sure about this? The Javadoc says these parts of the ISO string should parse to a long **_(except for the seconds part)_**:

jimtng:
Allowing fractions for the duration format might enable complications / confusions. For example, `1.5d 3h`. I guess in that case it's `(1.5 * 24 + 3) hours` but it makes the actual duration not immediately apparent. So I think it isn't the best idea.

With the introduction of `ms` unit, it is not necessary to allow fractions for seconds, and we don't really need higher precision than milliseconds to allow fractional ms, or ns.

mherwege:
> With the introduction of `ms` unit, it is not necessary to allow fractions for seconds, and we don't really need higher precision than milliseconds to allow fractional ms, or ns.

I totally agree. I don't see a reason to complicate this further. If you use an ISO String, you do need the fractional seconds because that's the way the millis are represented in the ISO String. But that will be handled properly. In all other cases, work with longs.

andrewfg:
Ok. LGTM. Again. Again. :)

kaikreuzer:
Sorry, this has slipped through the code freeze. It looks like a very low risk PR, though and I would partially even consider it as a bug fix. Would you agree that we should merge it?

jimtng:
If you're asking me, I'd agree :)

kaikreuzer (approved):

PR #4874, opened 2025-06-27
[rest] Add editable field to metadata

florian-h05:
This is required for the UI to properly handle metadata that is not editable in the UI.

### DIFF
--- a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
+++ b/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java
@@ -75,6 +75,7 @@ import org.openhab.core.items.ItemBuilderFactory;
 import org.openhab.core.items.ItemNotFoundException;
 import org.openhab.core.items.ItemRegistry;
 import org.openhab.core.items.ManagedItemProvider;
+import org.openhab.core.items.ManagedMetadataProvider;
 import org.openhab.core.items.Metadata;
 import org.openhab.core.items.MetadataKey;
 import org.openhab.core.items.MetadataRegistry;
@@ -183,6 +184,7 @@ public class ItemResource implements RESTResource {
     private final LocaleService localeService;
     private final ManagedItemProvider managedItemProvider;
     private final MetadataRegistry metadataRegistry;
+    private final ManagedMetadataProvider managedMetadataProvider;
     private final MetadataSelectorMatcher metadataSelectorMatcher;
     private final SemanticTagRegistry semanticTagRegistry;
     private final TimeZoneProvider timeZoneProvider;
@@ -203,6 +205,7 @@ public class ItemResource implements RESTResource {
             final @Reference LocaleService localeService, //
             final @Reference ManagedItemProvider managedItemProvider,
             final @Reference MetadataRegistry metadataRegistry,
+            final @Reference ManagedMetadataProvider managedMetadataProvider,
             final @Reference MetadataSelectorMatcher metadataSelectorMatcher,
             final @Reference SemanticTagRegistry semanticTagRegistry,
             final @Reference TimeZoneProvider timeZoneProvider) {
@@ -213,6 +216,7 @@ public class ItemResource implements RESTResource {
         this.localeService = localeService;
         this.managedItemProvider = managedItemProvider;
         this.metadataRegistry = metadataRegistry;
+        this.managedMetadataProvider = managedMetadataProvider;
         this.metadataSelectorMatcher = metadataSelectorMatcher;
         this.semanticTagRegistry = semanticTagRegistry;
         this.timeZoneProvider = timeZoneProvider;
@@ -267,7 +271,7 @@ public class ItemResource implements RESTResource {
             Stream<EnrichedItemDTO> itemStream = getItems(type, tags).stream() //
                     .map(item -> EnrichedItemDTOMapper.map(item, false, null, uriBuilder, locale, zoneId)) //
                     .peek(dto -> addMetadata(dto, namespaces, null)) //
-                    .peek(dto -> dto.editable = isEditable(dto.name));
+                    .peek(dto -> dto.editable = isEditable(dto));
             itemStream = dtoMapper.limitToFields(itemStream,
                     "name,label,type,groupType,function,category,editable,groupNames,link,tags,metadata,commandDescription,stateDescription");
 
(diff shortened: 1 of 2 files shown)


spacemanspiff2007:
I've always thought that metadata is editable even when defined in a file.
Is this not the case any more?

florian-h05:
Metadata defined from files is provided by the `GenericMetadataProvider`. This provider does not support modifying its metadata during runtime.
If you "edit" such metadata in the UI, it is added to the `ManagedMetadataProvider`.
Now you have to providers trying to provide the same thing, what you receive depends on who comes first, which very likely is not deterministic due to the partly undeterministic startup.

I have just tried to edit metadata provided through .items file from the UI, my changes are not visible/effective though I have saved them several times.

rkoshak:
It may be the case this has been broken for some time and I've not noticed it, but it used to be possible to add/edit metadata defined in files. I took advantage of this fact in several different ways back when OHRT was Python instead of JavaScript. 

It may have been a trick of timing and I lucked out, but it worked for me.

I personally don't do this anymore, so there's no impact on me personally, but I just helped someone on the forum the other day who wanted to edit metadata from a rule and was unable to because the metadata is first defined in .items files. 

I'm not sure what the correct answer is, but this approaches a breaking change, though maybe it's already changed since that users on the forum already can't do it? that doesn't mean it shouldn't be fixed, just care needs to be taken.

florian-h05:
> I'm not sure what the correct answer is, but this approaches a breaking change, though maybe it's already changed since that users on the forum already can't do it? that doesn't mean it shouldn't be fixed, just care needs to be taken.

This PR doesn't change the core behaviour, it only allows the UI to properly handle it and thus avoid confusing users.
I for my part never overwrote metadata from the UI since I never expected this to work, but I have to admit I am aware of the implementation, which is not the case for most users.
Anyway, even if the change in UI behaviour (not allowing to edit file-provided metadata anymore) is considered breaking, 5.0 is the time to break things ;-)

rkoshak:
But users don't like it when you take away capabilities, even if they never really worked reliably in the first place.

florian-h05:
That's a good point, but having users confused or unsatisfied if it doesn't work reliably in the first place also is not great.

Keep in mind users can still add metadata through the UI to file-based Items, just don't overwrite what's defined in the file.
I will also do it in a way users can still try out changes to the widget metadata, just not allowing them to click save in the UI.

rkoshak:
If possible, some looking into the error thrown when one tries to write to a read only metadata from a rule could be more clear. 

The user I helped on the forum got a "does not exist" error when trying to use `replaceMetadata()` from a rule, which really made no sense at the time. The metadata existed, it just wasn't writeable. I didn't look all the way down in the code, but I think the error was generated in Java, not OH, so here is where it should be addressed. But I could be long, I only briefly looked.

florian-h05:
> The user I helped on the forum got a "does not exist" error when trying to use replaceMetadata() from a rule, which really made no sense at the time. The metadata existed, it just wasn't writeable.

Did the user get an error or a warning?

In case of the warning, I can explain what happened. replaceMetadata() in openhab-js queries the MetadataRegistry for the metadata, and depending on whether null or Metadata is returned, metadataRegistry.update() or metadataRegistry.add(). If you call metadataRegistry.update(), managedMetadataProvider.update() is called under the hood, and this one logs a warning if it cannot find the updated metadata in its storage.

kaikreuzer:
This change is not breaking anything - so if anybody thinks the UI should allow editing file-defined metadata, the discussion should be continued on https://github.com/openhab/openhab-webui/pull/3250.

ghys:
I just wanted to add my experience when dealing with this issue, when I was building the Main UI.

Adding JSONDB (ie. managed) metadata to .items-files defined items was always technically a possibility indeed, and even a use case (for instance, you might prefer to define your item in an .items file, but add Alexa or Siri or Google Assistant metadata to it with the UI because it's more convenient - you get lists of options to choose from etc.)

And external tools could even add metadata in some namespaces to your items through the API, no matter how they were defined.

There was some confusion when you had both unmanaged and managed metadata added to an item for a particular namespace - shouldn't happen, but which would take precedence if it were the case? It was never clear, you'd have to look into the code for that.

Also since the items and metadata are managed pretty much separately, what happens when you change the .items files (rename, delete items, shuffle them around) but the managed data remains the same? You delete ItemB and rename ItemA to ItemB, and suddenly the managed metadata attached to the former ItemB gets attached to the newly renamed ItemA. You might think it's okay, when you're aware of this, but will everyone get this?

rkoshak:
> https://github.com/openhab/openhab-webui/pull/3250

What about editing it from rules? Personally that's my main concern. If we can no longer modify metadata defined in .items files from rules, that *is* a breaking change. But it's maybe a breaking change that has already been made? If not, it's definitely made now this has been merged, and it needs to be included in the list of breaking changes in the release notes. 

> Did the user get an error or a warning?

It was reported as an error, but they never posted the full log statement so maybe it was a warning. In the JS Scripting call to `replaceMetadata` it reported a "does not exist" log message.

https://community.openhab.org/t/js-replacemetadata/164646

> Could not update element with key expectedState:EG_Kueche_Beschattung_West in ManagedMetadataProviderImpl, because it does not exist.

florian-h05:
> Adding JSONDB (ie. managed) metadata to .items-files defined items was always technically a possibility indeed, and even a use case (for instance, you might prefer to define your item in an .items file, but add Alexa or Siri or Google Assistant metadata to it with the UI because it's more convenient - you get lists of options to choose from etc.)

This is still possible.

> There was some confusion when you had both unmanaged and managed metadata added to an item for a particular namespace - shouldn't happen, but which would take precedence if it were the case? It was never clear, you'd have to look into the code for that.

Main UI doesn't allow saving metadata through the UI anymore if that metadata has been created from files, so this shouldn't happen anymore.

> What about editing it from rules? Personally that's my main concern. If we can no longer modify metadata defined in .items files from rules, that is a breaking change. But it's maybe a breaking change that has already been made? 

I checked the code of both `AbstractManagedProvider::update` (where the "does not exist" log message comes from) and `AbstractRegistry::update` (which is called by openhab-js `replaceMetadata` if the metadata already exists), in both cases the code hasn't been touched since a37cceab6767313e1b99643739bdc450428dcbde in Jan 2019.
I hence can't imagine how it should have been possible to edit metadata defined in .items files from rules.

> If not, it's definitely made now this has been merged, and it needs to be included in the list of breaking changes in the release notes.

Again: This PR does change nothing wrt to rules, all it does is to add an additional field to the REST response so the UI knows wheter metadata is editable or not.

PR #4877, opened 2025-06-29
AbstractRegistry: hide exception stacktrace on IAE when adding an element.

jimtng:
Hide the scary stacktrace from user. The message is sufficient to describe the problem, which is usually a syntax error, e.g. not satisfying the regex check.

### DIFF
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/common/registry/AbstractRegistry.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/common/registry/AbstractRegistry.java
@@ -201,6 +201,12 @@ public abstract class AbstractRegistry<@NonNull E extends Identifiable<K>, @NonN
         }
         try {
             onAddElement(element);
+        } catch (final IllegalArgumentException ex) {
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.core/src/main/java/org/openhab/core/common/registry/AbstractRegistry.java
```diff
@@ -201,6 +201,12 @@ private boolean added(Provider<E> provider, E element, Collection<E> providerEle
         }
         try {
             onAddElement(element);
+        } catch (final IllegalArgumentException ex) {
+            logger.warn("Cannot add \"{}\" with key \"{}\": {}", element.getClass().getSimpleName(), uid,
```
wborn: It can be a one liner if you provide null or the exception based on the log level like:

https://github.com/openhab/openhab-core/blob/5c29932f1f30ab753e29cc45011543a5aeb199d6/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/chart/ChartServlet.java#L327

jimtng: Brilliant, thanks!

wborn (approved):
Thanks!

PR #4881, opened 2025-07-03
Fix persistence threshold filter with UoM

ea21fd6:
The [`PersistenceThresholdFilter`](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/filter/PersistenceThresholdFilter.java) currently yields unexpected results when converting temperatures. In particular the current (buggy) behavior depends on the unit of the threshold (e.g. one would expect "K" and "°C" to be interchangeable).

The first commit adds test cases, which include a currently [failing](https://github.com/ea21fd6/openhab-core/actions/runs/16045241135/job/45274898945#step:8:1639) one that wants the filter to exclude a "change" from "0&nbsp;°C" to "0&nbsp;°C" for a threshold of "10&nbsp;K".

The proposed fix uses `toUnitRelative()` to convert the difference between old and current state to the threshold unit. This method ignores different offsets of involved units, which fixes the above test case.


J-N-K (approved):
LGTM

PR #4899, opened 2025-07-14
Yaml Model Repository: skip non-existent paths during the initial scan instead of logging a warning

jimtng:
Resolve an issue reported in https://community.openhab.org/t/openhab-5-0-milestone-discussion/162686/277

### DIFF
--- a/bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/internal/YamlModelRepositoryImpl.java
+++ b/bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/internal/YamlModelRepositoryImpl.java
@@ -130,7 +130,10 @@ public class YamlModelRepositoryImpl implements WatchService.WatchEventListener,
         // read initial contents
         WATCHED_PATHS.forEach(watchPath -> {
             Path fullPath = mainWatchPath.resolve(watchPath);
-            if (!Files.exists(fullPath) || !Files.isDirectory(fullPath)) {
+            if (!Files.exists(fullPath)) {
+                return;
+            } else if (!Files.isDirectory(fullPath)) {
+                logger.warn("Expecting '{}' to be a directory, but it's a file. Ignoring it.", fullPath);
                 return;
             }
             try {
--- a/bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/internal/YamlModelRepositoryImpl.java
+++ b/bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/internal/YamlModelRepositoryImpl.java
@@ -130,6 +130,9 @@ public class YamlModelRepositoryImpl implements WatchService.WatchEventListener,
         // read initial contents
         WATCHED_PATHS.forEach(watchPath -> {
             Path fullPath = mainWatchPath.resolve(watchPath);
+            if (!Files.exists(fullPath) || !Files.isDirectory(fullPath)) {
+                return;
+            }
             try {
                 Files.walkFileTree(fullPath, new SimpleFileVisitor<>() {
                     @Override
--- a/bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/internal/YamlModelRepositoryImpl.java
+++ b/bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/internal/YamlModelRepositoryImpl.java
@@ -150,7 +150,14 @@ public class YamlModelRepositoryImpl implements WatchService.WatchEventListener,
                     @Override
                     public FileVisitResult visitFileFailed(@NonNullByDefault({}) Path file,
                             @NonNullByDefault({}) IOException exc) throws IOException {
-                        logger.warn("Failed to process {}: {}", file.toAbsolutePath(), exc.getMessage());
+                        String message = exc.getMessage();
+                        if (file.toString().equals(message)) {
+                            // If the message is just the path, we do not need to log it again.
+                            // This is the case for FileNotFoundException, AccessDeniedException, etc.
+                            // Instead of the path, we log the exception class to provide additional details.
+                            message = exc.getClass().getSimpleName();
+                        }
+                        logger.warn("Failed to process {}: {}", file.toAbsolutePath(), message);
                         return FileVisitResult.CONTINUE;
                     }
                 });


jimtng:
Ping @lolodomo and @Nadahar

bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/internal/YamlModelRepositoryImpl.java
```diff
@@ -130,6 +130,9 @@ public YamlModelRepositoryImpl(@Reference(target = WatchService.CONFIG_WATCHER_F
         // read initial contents
         WATCHED_PATHS.forEach(watchPath -> {
             Path fullPath = mainWatchPath.resolve(watchPath);
+            if (!Files.exists(fullPath) || !Files.isDirectory(fullPath)) {
```
Nadahar: I *think* `Files.exists()` covers folders as well as files.

Nadahar:
This should work, but there is a principle that some people swear by that says that instead of trying to anticipate the future, you should handle the exception when something unexpected happens. What this means is that even if you check that the file or folder exists before you start iterating, this can change during iteration. The information is essentially volatile.

This is from `Files.exists()` JavaDoc:

> Note that the result of this method is immediately outdated. If thismethod indicates the file exists then there is no guarantee that asubsequent access will succeed. Care should be taken when using thismethod in security sensitive applications.

It is thus still theoretically possible to get a `FileNotFoundException` during iteration, so "the most correct" is probably to avoid logging the warning if a `FileNotFoundException` is the cause of the "failure" regardless of if a check has been done upfront or not.

jimtng:
> It is thus still theoretically possible to get a `FileNotFoundException` during iteration, so "the most correct" is probably to avoid logging the warning if a `FileNotFoundException` is the cause of the "failure" regardless of if a check has been done upfront or not.

My intention was to still log FileNotFoundException in all other cases, so I don't want to mute it in `visitFileFailed`

bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/internal/YamlModelRepositoryImpl.java
jimtng: I'd like to exclude it if there happens to be a _file_ called `yaml` instead of a directory. But perhaps we should log a warning in that case

jimtng:
> This should work, but there is a principle that some people swear by that says that instead of trying to anticipate the future, you should handle the exception when something unexpected happens. What this means is that even if you check that the file or folder exists before you start iterating, this can change during iteration. The information is essentially volatile.

That is correct, and in most situations that's perhaps very much the case. In this case for scanning for a predefined list of watched paths though, in what circumstances do you think it can happen? I don't like suppressing an entire type of error to prevent a specific issue from happening, namely that the user doesn't have the known directories. Doing so would cause other potential issues to go unnoticed.

In the case when the user _does_ have say `conf/yaml` directory at the start of the scan, but for some reason that directory disappeared, I would very much want them to see a warning being logged.

Nadahar:
> In this case for scanning for a predefined list of watched paths though, in what circumstances do you think it can happen?

It won't happen "under normal circumstances", the principle is rather about trying to handle the unexpected too. It is possible that a user deletes or moves one or more folders while others are being parsed. Would logging it in such a case be beneficial? I'd say that in such a situation, the content of the log might be the least of the user's problems. 

Some of the problem with `FileNotFoundException` is also how terrible the error message itself is - just the path. It doesn't really give any information that is useful as to "what went wrong".

> I don't like suppressing an entire type of error to prevent a specific issue from happening, namely that the user doesn't have the known directories. Doing so would cause other potential issues to go unnoticed.

I agree with that in general, but I'm not sure that it applies here. If a file (or folder) can't be opened or read, if there's not enough permissions etc., other exceptions will be thrown - not `FileNotFoundException`. So, these cases would still be handled. Also, `Files.walkFileTree()` recursively processes all files and folder under the "start path", which means that under normal circumstances, the only place you *can* get a `FileNotFoundException` is if "start path" doesn't exist. All the rest are iterated because they have been returned as children of that path, by the OS - so per definition they do exist, at least when the iteration starts. So, it seems that again, the only way to experience this is if the file system is being manipulated *while* the recursive iteration runs.

jimtng:
> Some of the problem with `FileNotFoundException` is also how terrible the error message itself is - just the path. It doesn't really give any information that is useful as to "what went wrong".

I think this is a useful pointer. Perhaps we should log the exception class name before the message, e.g. `FileNotFoundException: <path>`, `AccessControlException: <msg>` ?


> So, it seems that again, the only way to experience this is if the file system is being manipulated _while_ the recursive iteration runs.

Which I think we should log, should it occur.

Nadahar:
> I think this is a useful pointer. Perhaps we should log the exception class name before the message, e.g. `FileNotFoundException: <path>`, `AccessControlException: <msg>` ?

That would absolutely be helpful for `FileNotFoundException`, but I don't think this is the situation for most other `IOExceptions`. Usually they have a somewhat helpful/descriptive message. That said, I don't know them all, there are no absolutes 😉 

> Which I think we should log, should it occur.

For what purpose? A file, or folder, that was there at the start of the iteration, but is no more, wasn't parsed. That doesn't warrant a warning to me: obviously the file or folder won't be parsed if it doesn't exist. You delete/move it if you don't want it to be parsed, so it not being parsed is what the user would expect - thus, this isn't something we need to "warn" the user about.

jimtng:
This is splitting hair about a situation that would probably never happen, but here is my thinking:

- Scenario 1: User didn't have the directory, or had deleted it before the scan -> no warning
- Scenario 2: User had the directory, then deleted it right at the exact moment after the initial check, but before being visited -> log a warning. Yes the user should know about it so a warning seems redundant, but not totally unexpected.
- Scenario 3: same as scenario 2, but something else deleted it at that exact moment, without user knowledge -> I think the user would like to know this.

So the choice is whether we want to silence scenario2 or scenario3. I prefer not to silence scenario3 at the expense of scenario2 being logged. But once again they're both a rare situation.

The benefit of your proposal is that we don't have to perform checks in each iteration, and only when handling an error, and that's a good thing.

The benefit of my current implementation is that it warns user if the known directory is a file, which I think is also a good thing.

The bottom line is, other than the scenario where the known list of directories not existing, I do not want to specifically silence FileNotFoundExceptions.

Nadahar:
> * Scenario 3: same as scenario 2, but something else deleted it at that exact moment, without user knowledge -> I think the user would like to know this.

This absolutely is splitting hairs, but I'd say that in the above situation, I don't think logging a warning that the file couldn't be processes is what would help the user figure out what (rogue) process is deleting the OH configuration files.

When it comes to warning if it's not a directory, that could be done independently of exception handling/logging during iteration. I don't know how relevant that situation is either, these paths aren't user specified, are they? I thought they were hardcoded, e.g `conf/yaml`, `conf/things` etc., so chances that a user would create a file with one of these paths should be pretty slim.

The check could be performed "in iteration" while only silencing the warning on the "root level", you can simply compare the provided `file` with `fullPath` and only silence the warning when they are equal. That way you could "get the logging as you want it" while avoiding relying on "outdated information". The reason I didn't suggest that is that I couldn't imagine a situation where logging a warning for a `FileNotFoundException` would actually be useful, and I still can't.

But, as I said initially, it will work regardless of which approach is used. I just wanted to mention why some people want to avoid making assumptions about file systems retaining their current state over time.

lolodomo:
I did not ready the full discussion in details but the change looks good to me.

rkoshak:
It looks like the changes apply the same for all folders in $OH_CONF, correct? For example, if I decided to only use `yaml` and no longer needed the `tags`, `items`, and `things` folders, I wouldn't get the warning logged, right? I'm pretty sure the answer is "yes" but want to confirm. "Yes" is the answer I'd want. If the folder is not used by the end user, it shouldn't be a warning if they remove it.

Nadahar:
> It looks like the changes apply the same for all folders in $OH_CONF, correct? For example, if I decided to only use `yaml` and no longer needed the `tags`, `items`, and `things` folders, I wouldn't get the warning logged, right? I'm pretty sure the answer is "yes" but want to confirm. "Yes" is the answer I'd want. If the folder is not used by the end user, it shouldn't be a warning if they remove it.

Yes, but this only applies to the YAML parser. Other parsers might complain.

rkoshak:
> Yes, but this only applies to the YAML parser. Other parsers might complain.

That's what I wasn't sure about. In that case the actual answer to my question is probably "no". I don't like that inconsistency but fixing the other parsers is probably too big for this. 

Ideally all the folders should be treated the same. The end users don't know anything about different parsers and stuff like that.

Nadahar:
I don't *know* if other parsers would complain. It should be pretty easy to test. I'm just saying that this particular PR will only deal with the YAML parser.

rkoshak:
I tested it before I replied. It does generate a warning.

```
2025-07-15 10:10:01.297 [WARN ] [.core.internal.folder.FolderObserver] - Directory 'items' does not exist in '/openhab/conf'. Please check your configuration settings!
```

lolodomo:
I fixed model parsers last weekend just before releasing milestone 4 to allow starting OH properly even when a folder like conf/items is missing. And yes, there is a warning log. This warning could be removed later, after OH 5 is released. Have in mind we are in feature freeze now.

Nadahar:
> Have in mind we are in feature freeze now.

Is log tweaking considered "a feature"?

lolodomo:
No but we have to be certain to break nothing when merging a PR after feature freeze, especially when it is to change something minor.

rkoshak:
> No but we have to be certain to break nothing when merging a PR after feature freeze, especially when it is to change something minor.

Then that should apply to this PR too then, right?

lolodomo:
> Then that should apply to this PR too then, right?

Yes.

@jimtng : is the YAML parser fully broken when cont/yaml is missing ? In that case, that is a severe bug and we have to merge ASAP that PR. If not, we can live with the warning and merge the improvement after the release.
That is only my suggestion, I am not a maintainer here, so any core maintainer could have a different analysis and could decide to merge.

jimtng:
> is the YAML parser fully broken when cont/yaml is missing ?

No, yaml files in non missing directories are still loaded / processed.

lolodomo:
I checked again the code change and I believe the risk is ultra low to merge it (I also assume @jimtng tested the change).
@openhab/core-maintainers : we let you decide but I think that will be better to merge this PR before OH 5.0 release, it will avoid the warning log for all users having openHAB already installed.

jimtng:
> I checked again the code change and I believe the risk is ultra low to merge it

I agree :)

lolodomo:
@jimtng : if the folder conf/yaml is created after openHAB is started, is the folder then automatically considered or is a OH server restart necessary ? For other (DSL) model parsers, a restart could be required (to be verified).

jimtng:
> if the folder conf/yaml is created after openHAB is started, is the folder then automatically considered

For the yaml parser:

Yes. You can have a missing conf/yaml (or any of the other folders) at start up, and once OH is running, create it, and start adding files into it. The new files will be loaded/processed.

You can then delete / rename the directory, which triggers the related objects to be removed from OH. Then rename it back, and the files will be processed again.

An OH restart is not required.

Nadahar:
> I tested it before I replied. It does generate a warning.
> 
> ```
> 2025-07-15 10:10:01.297 [WARN ] [.core.internal.folder.FolderObserver] - Directory 'items' does not exist in '/openhab/conf'. Please check your configuration settings!
> ```

I took a quick look at `FolderObserver` - removing the warning is very easy, but it looks to me like the logic doesn't really handle if the folder is added later (you'd have to restart OH). That should also be changed IMO, there's no reason why the parser can't "be ready" to watch a folder if it is created later. I'm not sure what the rationale behind this logic is, and if changing it has some other consequences.

lolodomo:
> but it looks to me like the logic doesn't really handle if the folder is added later (you'd have to restart OH).

@jimtng said the opposite just before you.
Who really tested that ?

**I hope this discussion is not preventing the merge of the PR.**

Nadahar:
> @jimtng said the opposite just before you.
> Who really tested that ?

I thought @jimtng was referring to the YAML parser. I was referring to `FolderObserver` that takes care of parsing the DSL formats.

This discussion shouldn't prevent merging, it just came up because @rkoshak mentioned that a warning is still logged if one of the "DSL folders" are missing.

I'm hesitant to remove the warning if it's true that the parser won't start processing the folder if it is created, which was all I tried to say above. I think it would be better to modify `FolderObserver` so that it *does* handle if the folder is created later, and *then* remove the warning.

lolodomo:
> I'm hesitant to remove the warning if it's true that the parser won't start processing the folder if it is created, which was all I tried to say above. I think it would be better to modify `FolderObserver` so that it _does_ handle if the folder is created later, and _then_ remove the warning.

It was just an hypothesis I did, I am not sure, I did not yet check that. We lack time to change that properly before 5.0.

But once again, we are off-topic, this PR is not about DSL but about YAML, so let's go for a merge.

Nadahar:
> It was just an hypothesis I did, I am not sure, I did not yet check that.

I didn't base what I said on your hypothesis. I studied the code in `FolderObserver`, and it looks to me like it won't be picked up without a restart. The code is a bit convoluted, so it's not immediately apparent how it will behave, but I'm relatively confident that it *won't* process a folder that wasn't present at startup.

I hope reviewers aren't prevented from merging this PR because of this "discussion". It doesn't impact whether this PR should be merged, but it is a closely related topic. Maybe someone should create an issue for `FolderObserver`s behavior instead.

lolodomo:
>  It doesn't impact whether this PR should be merged, but it is a closely related topic. 

Yes, I agree.

> Maybe someone should create an issue for FolderObservers behavior instead.

Yes please.
I suggest to stop that discussion in that PR.

PR #4902, opened 2025-07-15
DSL Item: fix item name validation pattern

jimtng:
Item names must not start with a digit, and must not contain a dash.

The looser check done here in the past didn't cause a problem because the item name is also checked by the AbstractRegistry and will be rejected from being added to the registry if it doesn't conform to the required syntax.

Since it has been like this for a long time, perhaps wait until after 5.0 before merging this in case it causes some unexpected issues.


jimtng:
@lolodomo

lolodomo:
Take care, "ID" is used not only for item name but also for other tokens.

jimtng:
I'm not sure if this is a gain. Fixing the pattern in items.xtext causes a more obscure error message from Xtext compared to the warning message issued by AbstractRegistry. Based on this, I'm closing this.

PR #4633, opened 2025-03-05
Rule file provider

Nadahar:
# Rule file provider

- Create rule file provider for `JSON` and `YAML` files
- Fix `DELETE` event bug in `FolderObserver`

# Description

The primary task of this PR is to provide file-based `YAML` and `JSON` rules. It supports multiple rules in one file if they are supplied within an array, which means enclosed in `[` and `]` and separated with comma. 

This PR is very loosely coupled to #4591 in that they are both split out as PRs from the same "work branch", and this PR use the `AbstractJacksonYAMLParser` introduced in #4591. Therefore, the `FolderObserver` file watcher `DELETE` fix has been included here, since it was discovered after #4591 was submitted.

There is one issue that might need to be discussed, and that is the location to look for the `YAML` and `JSON` files. There are two "obvious candidates", `conf/automation/rules` and `conf/rules`. The latter is the folder already in use for Rules DSL files, while the former isn't currently used (or automatically created when OH is installed). In either case, some minor changes to the documentation should be made as well, but it's hard to write that until the final location has been decided.

Currently, this PR uses `conf/rules`, but that's very easy to change. From a user's perspective, this might be the most logical, as they are still rules, although the format is different. It doesn't seem like this causes any conflicts, but a debug log message is created both from the Rules DSL parser (that it ignores the `.yaml` or `.json` file) and from the `AbstractFileProvider` that it doesn't have a parser for `.rules` files. None of these are shown by default. The involved classes could be modified to not log these cases, as they are expected, but both are "generic in nature" and introducing exceptions to concrete file extensions doesn't really "fit".

If `conf/automation/rules` are used instead, the "conflict" with two parsers monitoring the same folder would be avoided, but it might be more confusing for the end user to have to place the rule files in two different locations. It should also be considered if this folder should be created when the installation is set up, like with many of the other folders.

It should also be mentioned that [the documentation currently claims support for `JSON` based rules](https://www.openhab.org/docs/developer/module-types/#via-json-files), but no evidence that this exists, or has existed in previous versions, has been found in the code:

>The automation engine reads rule json files from the `{openhab-dir}/automation/*.json` directory.

If this actually is wrong, the documentation should be corrected.

PR #4910, opened 2025-07-17
Add tests for rules and rule templates JSON parsers

Nadahar:
This is another split out from #4633. It should be safe to merge, it only provides tests for the existing classes `RuleGSONParser` and `TemplateGSONParser`, a small log tweak and changes `HashSet` to `LinkedHashSet` in the two tested classes to make iteration order predictable, which makes testing easier.


Nadahar:
@holgerfriedrich This has been dormant for a while, while I expected it to be very uncontroversial. All it really does is add tests for some code that I have provided earlier, and some code that has been here for years.

I can rebase it on latest `main` if that is desirable/helpful, just let me know.

PR #4913, opened 2025-07-18
Apply default semantic tags from linked channels to items

### LINKED ISSUE #4898: Semantic Tags for item not inherited from linked channel (OH 5)
@andrewfg : I just made a quick test, removing some semantic tags from few items in my .items file.
I was expecting semantic tags being automatically inherited from the linked channel.
I just discover that it does not work, I mean my items are not present in Properties view in Main UI.
Wasn't it the goal with all the changes done with semantic tags in OH 5.0 ?

andrewfg:
Resolves #4898 

When an Item is linked to a Channel, and the Item does not have its own Point and/or Property tags, and the Item Channel Link is configured with `useTags=true`, then the default Point and/or Property tags of the Channel are applied to the Item.

This enhancement simplifies configuration for users by reducing the need for manual tag assignment.

The `useTags=true` setting can be applied individually for each channel link of a given Item. Or alternatively `useTags=true` can be applied system-wide globally for all Items via a setting in `runtime.cfg`


andrewfg:
@lolodomo / @jimtng this relates to #4898 and is currently work in progress, but perhaps you could have a quick look at it to see if I am on the right track?

andrewfg:
@lolodomo / @jimtng

1. I added tests for this new feature to `org.openhab.core.model.item.tests` and the code now passes those tests.
2. However unfortunately it causes `org.openhab.core.model.thing.tests` to fail; the stack trace is humungous and hard to understand (so any advice would be welcome).

jimtng:
I wonder if this should be done in the itemregistry instead. That way it's consistent across all Item providers, so we don't have to duplicate the logic for yaml and others (ruby, js).

andrewfg:
> wonder if this should be done in the itemregistry

I did wonder that too. My concern is that if it is done in the item registry it might (as you suggest) apply the same logic to items from all sources, and particularly if the user has manually applied tags to managed items, those tags would continually be overwritten by the default tags again.

EDIT to be more specific if the user had applied DIFFERENT tags, the current logic would not override them. But if the user had DELETED the tags, the current logic would always restore the default tags. .. But perhaps that is actually the correct approach?

jimtng:
> those tags would continually be overwritten by the default tags again.

Then don't set `useTags=true`. The same rule applies.

andrewfg:
@jimtng my edit to my prior comment crossed over with your subsequent comment, so please look up.

jimtng:
I would expect that the tags are combined with the custom tags that have been specified. But if they deleted the tags, especially in managed items (UI), then I think it should keep coming back on again until they remove the `useTags`.

In any case, I don't think it's currently possible to add custom config in the UI, so unless this is added, this is a moot issue?

jimtng:
To clarify, 

```
String MyItem [MyCustomTag] { channel="x:y:z" [ useTags=true ] }
```

Would result in MyItem having MyCustomTag + whatever the tags provided by the channel.

And in the UI, all the tags would show up.

andrewfg:
> Would result in MyItem having MyCustomTag + whatever the tags provided by the channel.

That is not how I coded it currently; it is currently either/or. However I guess that the union of the two tag sets is as you say really the proper solution. (However we need to think of the case where there are two channels links both with useTags..)

jimtng:
> However we need to think of the case where there are two channels links both with useTags

As we discussed, this case will result in a warning and perhaps the subsequent channel tags being ignored.

andrewfg:
@jimtng I moved the code to the item registry; however -- maybe I am missing something obvious -- but I can't get it to compile due to circular references between the item registry, the item channel link registry and the thing registry..

lolodomo:
You introduced a circular reference by referencing the channel link registry in the item registry.
Playing with channel links can be a kind of nightmare because this concept being in core.thing bundle is at the frontier of the thing (channel) being in core.thing bundle and the item being in core bundle.

jimtng:
How about instead of doing this in ItemRegistry, you could do it in ItemChannelLinkRegistry and hook into the `added` and `updated` methods. Maybe that will also resolve the circular reference issue.

Although it might be tricky to figure out whether the item has already 'used' tags from another link.

bundles/org.openhab.core.model.item/src/org/openhab/core/model/item/internal/GenericItemProvider.java
```diff
@@ -403,7 +403,6 @@ public void modelChanged(String modelName, EventType type) {
                 case MODIFIED:
                     Map<String, Item> oldItems = toItemMap(itemsMap.get(modelName));
                     Map<String, Item> newItems = toItemMap(getItemsFromModel(modelName));
-                    itemsMap.put(modelName, newItems.values());
```
lolodomo, line 422: Such change will have to be seriously commented.
Is it related to your new feature ?
Or do you use this PR to fix a bug you discovered in the DSL item provider?

lolodomo:
In my opinion, the item registry should not be updated, it contains the tags set by users.
What should be updated are the REST API returning information about items. These APIs should combine data from item registry, data from channel link registry and data from thing registry. The bundle containing these APIs has an access to all these registries.

So the item registry contains tags set by users. No need to change it.
The channel link registry will contain a configuration parameter to know if tags from channel links have to be considered or not. No need to change it.
The thing registry contains the tags from the channel types. No need to change it.
And the code behind the REST APIs will collect that data and retrieve if necessary tags from the corresponding channel and fill the result. It is what needs to be changed.

andrewfg:
> the item registry should not be updated, it contains the tags set by users.

The tags can only be updated in the item registry if the user has set 'useTags' on the item channel link configuration, **and** if there is not already a Point or Property tag on it.

jimtng:
> So the item registry contains tags set by users. No need to change it.
> The channel link registry will contain a configuration parameter to know if tags from channel links have to be considered or not. No need to change it.
> The thing registry contains the tags from the channel types. No need to change it.
> And the code behind the REST APIs will collect that data and retrieve if necessary tags from the corresponding channel and fill the result. It is what needs to be changed.

That would create an inconsistency/confusion. It would "appear" in the UI that the item has the tags, but in scripts, they don't have the tags.

Either add them to the item as planned (when useTags is true), or if that's not desirable, don't.

Personally, I would rather manually and explicitly define all the tags myself and not rely on useTags, because it wouldn't be clear from reading the textual definition, what the added tags are going to be.

andrewfg:
Aargh! I ran spotless and it chas changed 142 pom.xml files !!

andrewfg:
@jimtng @lolodomo 

Current status update
1. I moved the tagging operation to the ItemChannelLinkRegistry
2. There are now only two files in the PR
3. As you can see I added `ItemChannelLinkAssignDefaultTagsOSGiTest.java` .. HOWEVER during a local Maven build there is no evidence that this test suite ever gets called. **_Indeed the is no evidence of any of the OSGi tests in `org.openhab.core.thing.tests` being called._** -- Is this a bug? Or is there some switch necessary to run the tests?

andrewfg:
I am not sure why my code removed the casts 'Stream<xyz>' .. perhaps spotless did it .. but anyway it seems correct.

lolodomo:
There is certainly something to do also in notifyListenersAboutRemovedElement.

andrewfg:
> something to do also in notifyListenersAboutRemovedElement

I guess that you mean that if an item channel link would be removed then you want to revert to whatever were the prior tags? Or?

That could be quite difficult, since we don't directly know what the prior tags were. I guess we could implement some kind of cache in the item channel link registry, but it would be quite complicated. Alternatively I suppose we can assume that (in the case of .items files) if a link has been removed, it explicitly means that the .items file entry has been entirely reloaded, so there might be no need for local caching in the link registry..

EDIT: (I had a sleepless night) insofar as the channel link tags would only be applied if the item has no native tags, we can probably just remove the channel’s point and property from the item’s native tag set if a link would be removed. And if there might be a second channel link we could check that and eventually apply that channels tags instead.

itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/link/ItemChannelLinkAssignDefaultTagsOSGiTest.java
```diff
@@ -0,0 +1,201 @@
...
+                .formatted(TEST_ITEM_NAME, ItemChannelLinkRegistry.USE_TAGS);
+
+        modelRepository.addOrRefreshModel(TEST_MODEL_NAME, new ByteArrayInputStream(input.getBytes()));
+
+        Item item = itemRegistry.get(TEST_ITEM_NAME);
+        assertNotNull(item);
+        assertEquals(2, item.getTags().size());
+        assertTrue(item.getTags().contains("tag1"));
+        assertTrue(item.getTags().contains("tag2"));
+        assertTrue(item.getTags().contains("this should fail"));
```
andrewfg: @lolodomo / @jimtng as you can see this test does NOT fail .. which is a proof that the build system is not actually running this test suite. Indeed by placing a similar failing test in other OSGi test classes in `org.openhab.core.thing.tests` we can easily see that **_none of the OSGI tests in the module are actually being run._**

andrewfg:
We need to test the timing too. Specifically to confirm that the item is in the registry before the link is created..

itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/link/ItemChannelLinkAssignDefaultTagsOSGiTest.java
lolodomo: Open a dedicated issue for that.

lolodomo:
> I guess that you mean that if an item channel link would be removed then you want to revert to whatever were the prior tags?

Yes

> We need to test the timing too. Specifically to confirm that the item is in the registry before the link is created..

I have the same thought yesterday.
I hope it is the case.
To be checked for all item providers.

lolodomo:
> Alternatively I suppose we can assume that (in the case of .items files) if a link has been removed, it explicitly means that the .items file entry has been entirely reloaded, so there might be no need for local caching in the link registry..

Yes, makes sense for DSL and YAML.

lolodomo:
And please rename the title of your PR as your change is not specific to .items file but to all config files defining items.

jimtng:
> > Alternatively I suppose we can assume that (in the case of .items files) if a link has been removed, it explicitly means that the .items file entry has been entirely reloaded, so there might be no need for local caching in the link registry..
> 
> Yes, makes sense for DSL and YAML.

We can't rely on this. Links can be added and removed on the fly, onto existing items, not just when the items were first created.



> Specifically to confirm that the item is in the registry before the link is created..

I'm not sure I understand what you're saying or what you intended to do, but links can exist even though items don't exist. We can have orphaned links. One example would be the links loaded from jsondb. This is why there's "Health" reporting in mainui.

andrewfg:
> Links can be added and removed on the fly, onto existing items

@lolodomo / @jimtng I have added support for dynamic link removal.
And I think one can't apply a `useTags=true` config to on the fly created links, so @jimtng your point is perhaps also moot??

> not sure I understand what you're saying or what you intended to do, but links can exist even though items don't exist.

@jimtng when a link is created (or updated) we scan the linked channel for possible tags, and then we update the item in the item registry with those tags. So this tagging can only work if there is an actual item in the item registry. Therefore when a new item is created from (say) a .items file the sequence of object creation becomes important:

1. Item object created and added to item registry, then link object(s) created and added to link registry. => Works!
2. Link object(s) created and added to link registry, then item object created and added to item registry. => Fails!

>  please rename the title of your PR 

Done. And I added a description to the first post (as the linked issue describes only the .items sub case).

itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/link/ItemChannelLinkAssignDefaultTagsOSGiTest.java
andrewfg: https://github.com/openhab/openhab-core/issues/4921

andrewfg:
> when a new item is created from (say) a .items file the sequence of object creation becomes important

In the meantime I checked GenericItemProvider and YamlItemProvider .. and at least those two are first adding the item to the item registry and then adding the links to the link registry. So at least in those two cases the timing should be Ok.

Unfortunately OH core is too dense for me to figure out the timing for managed items. I horribly suspect that the items and the links are loaded separately and asynchronously (perhaps even in multiple threads). So I could imagine that we might have the cases of links being created before items here. I will think about possible methods to handle timing crossovers tomorrow. But in the meantime the good news is that managed items are probably not going to use such tag importation anyway, since the tags come via the UI.

EDIT: I think if we add a listener to the item registry added event we could detect cases where we had already got a link but not yet an item..

... however it would be nice if either 1) we could build OH Core to test on a live system, and/or 2) we could run the OSGi tests .. hopefully tomorrow ??

jimtng:
> 2\. Link object(s) created and added to link registry, then item object created and added to item registry. => Fails!

You may also need to apply the tags when an item is created. That would cover both possibilities.



> EDIT: I think if we add a listener to the item registry added event we could detect cases where we had already got a link but not yet an item..

Yes, that's what I meant :)

But this is getting more and more complex, is it worth it for the feature we're trying to add? e.g. is this something you or @lolodomo will use?


> 1) we could build OH Core to test on a live system

This might be a silly way to do it, and there may be a better way, but I set up a "dev" docker container, and then mount my local .m2 repository onto it. 

```yaml
    volumes:
      - /home/openhab/.m2/repository/org/openhab/core/bundles:/openhab/runtime/system/org/openhab/core/bundles
```

This way I just need to build/install core once, then build/install the bundle I'm working on. Then stop the docker container, delete cache, then start it again. 

> 2) we could run the OSGi tests

This should work, and will be the easier way perhaps.

andrewfg:
> this is getting more and more complex, is it worth it for the feature we're trying to add?

It was complex. But the code is written now. => See my last commit.
 
> Is this something you or @lolodomo will use?

Sure.

andrewfg:
@lolodomo / @jimtng status update..

- code complete
- CI build running again
- still no OSGi tests running => hopefully someone will resolve #4921 as it is rather important
- tomorrow I will set up a new OH 5.1 test system and try to do the tests the hard way

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/link/ItemChannelLinkRegistry.java
```diff
@@ -229,4 +252,139 @@ public enum ItemChannelLinkProblem {
...
+        try {
+            if (itemRegistry.getItem(link.getItemName()) instanceof ActiveItem activeItem) {
+                removeChannelDefaultTags(link, activeItem);
+            }
+        } catch (ItemNotFoundException e) {
+        }
+    }
+
+    /**
+     * If the linked channel has 'useTags=true' in its configuration return it's default tags.
```
jimtng:
```suggestion
     * If the linked channel has 'useTags=true' in its configuration return its default tags.
```

jimtng (commented):
I haven't looked into this closely, just a quick question:

This is a hypothetical scenario (haven't thought of exactly how to do this)
an Item is manually tagged with "CustomTag, Status, Power", then linked to a channel with useTags=true, and the channel provides "Status, Power" (matching what's set manually in the item).

Then what would happen when the link is removed. I would expect the item continues to have all 3 tags as manually set?

andrewfg:
> This is a hypothetical scenario

Your hypothesis is absolutely correct. This is indeed the one edge case where the deletion of a useTags channel link would indeed wrongly eliminate the native tags. (The custom tags would be preserved, but the Point and Property would not). This edge case could eventually be eliminated by adding a field to the item (say) ‘nativeTags’ but it would have to propagate through all aspects of items .. and this is probably indeed the limit point beyond which the extra development effort would exceed the extra benefit.

jimtng:
> his edge case could eventually be eliminated by adding a field to the item (say) ‘nativeTags’

How about in the itemchannellink itself, save the "original tags" before it added the channel's tags, and restore this on removal? This way you are keeping this data/logic contained within the same object. 

Or even better: save the list of _actually_ added tags (i.e. don't save it if the tag already exists prior to adding). Then only remove these.

- There will still be another edge case: after all this, then that tag got manually added to the item... thus invalidating your saved tags list...

andrewfg:
Yeah. I thought of that just after making my prior post. We could store it in the Configuration of the item channel link

> another edge case: after all this

The configuration would have ‘useTags’ and (say) ‘tagsWereIndeedUsed’ .. (I will think of a better name later..)

jimtng:
It could just be in a private variable invisible to the user

andrewfg:
> private variable invisible to the user

Yup. See my latest commit.

lolodomo:
I will check the last version of code later.
Is the inheritance of tags from channels the default ? It was my expectation to avoid adding the new parameter to all my channels and rather set it only when I don't like developer proposal.

andrewfg:
> Is the inheritance of tags from channels the default ?

Currently not. It requires useTags to be explicitly set. I did it that way to avoid any breaking changes. However the logic only applies channel tags if the user had not already defined point or property tags on the item. So the risk of breaking changes may be quite low. Therefore I will defer to you concerning whether we make it the default setting or not.

jimtng:
Making it the default (with the caveat that it won't overwrite existing point/property) is convenient. It can also be a "surprise" although I think it should be a good kind of surprise. But I'm not really sure of this. It would certainly cause an unexpected behaviour. So the prudent way is to make it opt-in.



> Yup. See my latest commit.

How does it handle the following scenarios:

Scenario A: Item defined with "Point", and "Property", linked to a channel with "Point" and "Property", then channel is unlinked -> Item should retain its "Point" and "Property" tags.

Scenario B: Item defined with "Point" only, linked to a channel with "Point" and "Property". When channel is unlinked -> Item should retain its "Point", whereas "Property" is removed.

andrewfg:
@jimtng I don't agree with either of your scenarios. If the item has native tags then they are never changed. But if the item has no native tags, then the channel tags are applied/removed all or nothing as the channel link is applied/removed.

jimtng:
> @jimtng I don't agree with either of your scenarios. If the item has native tags then they are never changed. But if the item has no native tags, then the channel tags are applied/removed all or nothing as the channel link is applied/removed.

Can you elaborate / give an example please?

andrewfg:
> Can you elaborate / give an example please?

#### Scenario 1

- Item: [] => Add Channel: [Switch, Power] => Result: [Switch, Power] @tagsLinked == true
- Item: [Switch, Power] => Remove Channel: [Switch, Power] @tagsLinked == true =>Result: [] @tagsLinked == false

#### Scenario 2

- Item: [CustomTag] => Add Channel: [Switch, Power] => Result: [CustomTag, Switch, Power] @tagsLinked == true
- Item: [CustomTag, Switch, Power] => Remove Channel: [Switch, Power] @tagsLinked == true => Result: [CustomTag] @tagsLinked == false

#### Scenario 3

- Item: [CustomTag, Switch, Power] => Add Channel: [Switch, Power] => Result: [CustomTag, Switch, Power] @tagsLinked == false
- Item: [CustomTag, Switch, Power] => Remove Channel: [Switch, Power] @tagsLinked == false => Result: [CustomTag, Switch, Power] @tagsLinked == false

#### Scenario 4

- Item: [CustomTag, Switch, Power] => Add Channel: [Measurement, Temperature] => Result: [CustomTag, Switch, Power] => @tagsLinked == false
- Item: [CustomTag, Switch, Power] => Remove Channel: [Measurement, Temperature] @tagsLinked == false => [Result: CustomTag, Switch, Power]

#### Scenario 5

- Item: [CustomTag, Switch] => Add Channel: [Switch, Power] => Result: [CustomTag, Switch] @tagsLinked == false
- Item: [CustomTag, Switch] => Remove Channel: [Switch, Power] @tagsLinked == false => Result: [CustomTag, Switch] @tagsLinked == false

#### Scenario 6

- Item: [CustomTag, Power] => Add Channel: [Switch, Power] => Result: [CustomTag, Power] @tagsLinked == false
- Item: [CustomTag, Power] => Remove Channel: [Switch, Power] @tagsLinked == false => Result: [CustomTag, Power] @tagsLinked == false

#### Scenario 7

- Channel: [Switch, Power] => Add Item: [CustomTag] => Result: [CustomTag, Switch, Power] @tagsLinked == true
- Item: [CustomTag, Switch, Power] => Remove Item => Result: N/A @tagsLinked == false

#### Scenario 8

- Item: [] => Add Channel1: [Switch, Power] => Result: [Switch, Power] @tagsLinked1 == true
- Item: [Switch, Power] => Add Channel2: [Measurement, Temperature] => Result: [Switch, Power] @tagsLinked2 == false
- Item: [Switch, Power] => Remove Channel1: =>Result: [Measurement, Temperature] @tagsLinked1 == n/a / @tagsLinked2 == true

jimtng:
Thanks for the comprehensive examples. It looks perfect!

jimtng:
> Making it the default (with the caveat that it won't overwrite existing point/property) is convenient. It can also be a "surprise" although I think it should be a good kind of surprise. But I'm not really sure of this. It would certainly cause an unexpected behaviour. So the prudent way is to make it opt-in.

@rkoshak, wdyt regarding whether we make this automatic tagging from the channel's default tags be the default (opt-out), or opt-in?

andrewfg:
Apropos opt-in vs opt-out: We could consider having a system wide setting which if absent causes opt-out to be assumed and if present forces opt-in to be applied. But perhaps this would be too meta-physical (a configuration that changes the behaviour of other configurations) for the average user to understand. ??

jimtng:
> Apropos opt-in vs opt-out: We could consider having a system wide setting which if absent causes opt-out to be assumed and if present forces opt-in to be applied. But perhaps this would be too meta-physical (a configuration that changes the behaviour of other configurations) for the average user to understand. ??

That would make people from both camps to be happy!
It could go in the runtime.cfg - since we're dealing with text-file people, it shouldn't be a problem. No need to expose it through the UI settings.
Maybe: `org.openhab.core.things.ItemChannelLinkRegistry:addTags` defines the default which can be overridden in the individual link config, and the default can be true.

andrewfg:
> make people from both camps to be happy!

Are you sure that it is not too metaphysical? It would be simpler to have a fixed behavior..

jimtng:
> > make people from both camps to be happy!
> 
> Are you sure that it is not too metaphysical? It would be simpler to have a fixed behavior..

Well... it can be ignored and it would make no difference, but if you do want to change it, it's possible. Best of both worlds.

andrewfg:
Ok. I did my first test on a live system. And I can confirm that the Item does take over both the native tags of the item (green color) and the tags of the linked channel (red color) as shown below. Note: I am not sure why both sets of tags are listed in the UI under the title "Non-Semantic Tags" (blue color) .. perhaps that is another issue to be opened?

#### Items file
```
Color Office_Lamp_Colour "Office Lamp Colour" [foo, bar] {channel="hue:device:001788fffe2157c7:24e7a04e-ad36-4192-961e-ca7f22103cbc:color" [ useTags=true ] }
```

#### Log
```
2025-07-24 12:14:56.538 [DEBUG] [e.thing.link.ItemChannelLinkRegistry] - Item 'Office_Lamp_Colour' assigned tags '[Control, Color]' from channel 'hue:device:001788fffe2157c7:24e7a04e-ad36-4192-961e-ca7f22103cbc:color'.
```

#### User Interface

image

lolodomo:
Maybe you added the tags and not replaced them?

lolodomo:
Why tags from channels are considered in that case?
If item has already tags set, we have to ignore tags from channel.

jimtng:
When adding tags, did you tell the item to notify listeners that it had changed?

lolodomo:
foo+bar should not be changed if the user assigned them to the item.

andrewfg:
> Why tags from channels are considered in that case?
> If item has already tags set, we have to ignore tags from channel.

No. If the item has **Point** or **Property** tags then indeed we ignore tags from channel. But if the item has only Custom tags then the result is the **union** of the Custom tag from the item plus the Point / Property tags from the channel. => _see Scenario 2 above._

> foo+bar should not be changed if the user assigned them to the item.

Correct => _see Scenarios above_

> When adding tags, did you tell the item to notify listeners that it had changed?

The code does indeed call `itemRegistry.update()` .. but the OH code assumes that elements from .items files are read only so unfortunately it forces to call to `managedProvider.update()` which logs an error instead of actually notifying the listeners.

rkoshak:
> @rkoshak, wdyt regarding whether we make this automatic tagging from the channel's default tags be the default (opt-out), or opt-in?

Any way to make it opt-in for existing users and opt-out for new users? That way on an upgrade OH will continue working as expected without surprises for those who have already configured their Items, but new users will get the tags by default. I don't know if that's feasible or not, but I think that would be ideal. apt/yum/et al don't usually mess around with runtime.cfg I think, so I'm not sure that's feasible. 

Short of that, I'm not a files users so am probably not the best person to answer. But personally I'd prefer an opt-in over an opt-out for something like this. If I use files it's because I don't want spooky magic mucking with my configuration. I want to know where everything comes from, and how it's applied. 

As an aside, to reduce surprises, is there any way to do an info level log when the suggested semantic tags are applied? That can help with surprises, unless of course it clutters the log too much because it happens too often or something like that. 

My biggest objection with how stateDescription is handled and what I would want to avoid with this is the hiddenness of it. It just happens and there is no indication anywhere about when it happens and how to control or override it until after it happens that first time and one needs to research or ask on the forum "where did these tags come from?". Making it opt in goes a long way towards addressing that since the user needs to make a positive decision to use the features. Creating some sort of log, even if it's just one time "applying binding semantic tags to Items" could also help address that even if it's opt out instead.

This makes me wonder though, even some UI users will want to opt out of this feature. Right now to opt out they need to remove the suggested tags for each Item. It would be a better experience for those users if this opt-out flag also let those users disable the suggested tags over all. But that's for a different issue in another repo, but it does go to your comment about not needing to expose the config in the UI. I think eventually it will need to be.

jimtng:
> This makes me wonder though, even some UI users will want to opt out of this feature. Right now to opt out they need to remove the suggested tags for each Item.

Ahhh yes, making it opt-out (applied by default) will affect UI users badly. They won't even be able to opt out.

So I think it's much safer to make this opt-in, @andrewfg, for many reasons.

andrewfg:
> So I think it's much safer to make this opt-in, @andrewfg, for many reasons.

I reverted to opt in. But we need confirmation from @lolodomo. However I will also try to address @lolodomo 's concerns about not having to type in useTags=true for all items by using a system wide override configuration.

andrewfg:
> is there any way to do an info level log when the suggested semantic tags are applied

@rkoshak sure. It is currently logged at debug level but I have just changed it.

andrewfg:
> OH code assumes that elements from .items files are read only so unfortunately it forces to call to managedProvider.update() which logs an error instead of actually notifying the listeners

Ok. I fixed that too.

#### Item definition 
```
Color Office_Lamp_Colour "Office Lamp Colour" [foo, bar] {channel="hue:device:001788fffe2157c7:24e7a04e-ad36-4192-961e-ca7f22103cbc:color" [ useTags=true ] }
```
#### Log

```
2025-07-24 17:27:31.207 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model '24g.items'
2025-07-24 17:27:31.320 [INFO ] [e.thing.link.ItemChannelLinkRegistry] - Item 'Office_Lamp_Colour' assigned tags '[Control, Color]' from channel 'hue:device:001788fffe2157c7:24e7a04e-ad36-4192-961e-ca7f22103cbc:color'.
```

#### UI

image

mherwege:
What happens if you have the same item linked to 2 channels with conflicting default tags?

andrewfg:
> What happens if you have the same item linked to 2 channels with conflicting default tags?

It iterates over all possible sources of Point/Property tags (starting with the Item native tags, and followed by each channel link in order) and whichever is the first to contain a Point or Property tag (with useTags) will be taken (and logged info), and any subsequent ones are logged warn.

andrewfg:
PS while testing this, I discovered that there is an already existing bug in `ModelRepositoryImpl` where it tries to parse native item tags. Specifically the tags [Status, foo, bar] load OK but the tags [Switch, foo, bar] cause an error..

#### Test case 1 (loads OK)

```
// an item with native tags [Status, foo, bar] loads OK
Color Office_Lamp_Colour "Office Lamp Colour" [Status, foo, bar] {channel="hue:device:001788fffe2157c7:24e7a04e-ad36-4192-961e-ca7f22103cbc:color" [ useTags=true ] }
```
#### Test case 2 (does not load)

```
// an item with native tags [Switch, foo, bar] causes an error
Color Office_Lamp_Colour "Office Lamp Colour" [Switch, foo, bar] {channel="hue:device:001788fffe2157c7:24e7a04e-ad36-4192-961e-ca7f22103cbc:color" [ useTags=true ] }

// log entry below
2025-07-24 18:40:36.360 [WARN ] [el.core.internal.ModelRepositoryImpl] - Configuration model '24g.items' has errors, therefore ignoring it: [1,48]: no viable alternative at input 'Switch'
[1,54]: extraneous input ',' expecting RULE_ID
```

Reported in https://github.com/openhab/openhab-core/issues/4927

andrewfg:
> That would make people from both camps to be happy!
> It could go in the runtime.cfg - since we're dealing with text-file people, it shouldn't be a problem.

@jimtng I don't know which OH core module is responsible for parsing such configs and making the available to other modules. => Can you give me a clue please?

andrewfg:
@lolodomo I added support for a system wide configuration entry `alwaysUseTags` in `runtime.cfg` which should avoid you having to set `useTags=true` individually on every item channel link; but note that its default is false, so you would still need to modify one file..

```
org.openhab.core.things.ItemChannelLinkRegistry:alwaysUseTags=true
```

jimtng:
> alwaysUseTags

I'm just thinking about making this as self-explanatory as possible. Alternatives are: `applyChannelTags`, `applyChannelTagsToItem`, `addChannelTags`, and`addChannelTagsToItem`

andrewfg:
My logic is that useTags applies with a certain syntax for individual links and that alwaysUseTags applies _with the exact same syntax_ for all links in the registry. If we rename the one, we should rename them both.

EDIT: we could call them BOTH useTags indeed; where their scope is evident from the location where they are declared.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/link/ItemChannelLinkRegistry.java
```diff
@@ -49,18 +62,44 @@
  * @author Markus Rathgeb - Rewrite collection handling to improve performance
  */
 @NonNullByDefault
-@Component(immediate = true, service = ItemChannelLinkRegistry.class)
-public class ItemChannelLinkRegistry extends AbstractLinkRegistry<ItemChannelLink, ItemChannelLinkProvider> {
+@Component(immediate = true, service = ItemChannelLinkRegistry.class, configurationPid = "org.openhab.core.things.ItemChannelLinkRegistry")
```
lolodomo: If you search "configurationPid", you will see that org.openhab.xxx is generally used instead of org.openhab.core.xxx.
Maybe "org.openhab.ItemChannelLinks" could be fine ?

andrewfg: Ok

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/link/ItemChannelLinkRegistry.java
```diff
@@ -229,4 +270,171 @@ public enum ItemChannelLinkProblem {
...
+                alreadyHasPointOrPropertyTag = true;
+                break;
+            }
+        }
+
+        Set<String> channelDefaultTags = getChannelDefaultTags(link);
+        if (!channelDefaultTags.isEmpty()) {
+            if (alreadyHasPointOrPropertyTag) {
+                logger.warn("Item '{}' already tagged; forbidden to assign tags from channel '{}'.",
+                        activeItem.getName(), link.getLinkedUID());
```
lolodomo: Why a warning?!
This is just a normal case, certainly even the standard case until this new feature is added.
This would generate massive logs for no reason as there is nothing to warn.
In my opinion, there is nothing to log, or just a DEBUG log.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/link/ItemChannelLinkRegistry.java
```diff
@@ -229,4 +270,171 @@ public enum ItemChannelLinkProblem {
...
+     * if the linked channel is the actual source of the item's tags then remove those tags from
+     * the item. If the item had any native custom tags they shall NOT be removed. Finally iterate
+     * over any other linked channels so they may eventually provide new tags.
+     */
+    private void removeChannelDefaultTags(ItemChannelLink oldLink, ActiveItem activeItem) {
+        if (oldLink.tagsLinked()) {
+            Set<String> newTags = new HashSet<>(activeItem.getTags());
+
+            // remove old link's tags
+            Set<String> oldLinkTags = getChannelDefaultTags(oldLink);
```
lolodomo, line 343: Check that `oldLinkTags` is not empty (like you did in `assignChannelDefaultTags`.
If empty, nothing to do, this will also avoid an unexpected log message.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/link/ItemChannelLinkRegistry.java
andrewfg: It means that in the .items file the user self intentionally applied **both** Point/Property at the native item level and also declared useTags at the Channel level. So we have to warn the user that they cannot have both.

HOWEVER having said that, if we go the way of applying alwaysUseTags at global level then indeed we should probably not log it.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/link/ItemChannelLinkRegistry.java
```diff
@@ -229,4 +270,171 @@ public enum ItemChannelLinkProblem {
...
+
+            // iterate over other links in case one may assign new tags
+            boolean alreadyHasPointOrPropertyTag = false;
+            for (ItemChannelLink otherLink : getLinks(activeItem.getName())) {
+                if (!otherLink.getUID().equals(oldLink.getUID())) {
+                    Set<String> otherLinkTags = getChannelDefaultTags(otherLink);
+                    if (!otherLinkTags.isEmpty()) {
+                        if (alreadyHasPointOrPropertyTag) {
+                            logger.warn("Item '{}' already tagged; forbidden to assign tags from channel '{}'.",
+                                    activeItem.getName(), otherLink.getLinkedUID());
```
lolodomo: Same here, no need for a warning

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/link/ItemChannelLinkRegistry.java
```diff
@@ -229,4 +270,171 @@ public enum ItemChannelLinkProblem {
...
+            Set<String> newTags = new HashSet<>(activeItem.getTags());
+
+            // remove old link's tags
+            Set<String> oldLinkTags = getChannelDefaultTags(oldLink);
+            newTags.removeAll(oldLinkTags);
+            logger.info("Item '{}' removed tags '{}' supplied from channel '{}'.", activeItem.getName(), oldLinkTags,
+                    oldLink.getLinkedUID());
+
+            // iterate over other links in case one may assign new tags
+            boolean alreadyHasPointOrPropertyTag = false;
```
lolodomo, line 352: Here, you are not checking if a Point/Property tag is defined on the item itself like you did in `assignChannelDefaultTags`. There is something missing I believe.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/link/ItemChannelLinkRegistry.java
andrewfg: No need. The test `if (oldLink.tagsLinked()) {` above explicitly eliminates such a case.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/link/ItemChannelLinkRegistry.java
andrewfg: It is not necessary to do that, since (as mentioned above) if the given channel has NOT provided tags, we would never get to this point in the code.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/link/ItemChannelLinkRegistry.java
andrewfg: Same comment as above..

lolodomo:
> > alwaysUseTags
> 
> I'm just thinking about making this as self-explanatory as possible. Alternatives are: `applyChannelTags`, `applyChannelTagsToItem`, `addChannelTags`, and`addChannelTagsToItem`

My pref: applyChannelTagsToItem

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/link/ItemChannelLinkRegistry.java
lolodomo: > HOWEVER having said that, if we go the way of applying alwaysUseTags at global level then indeed we should probably not log it.

Yes
I assume that most of users will enable this setting if they are informed of its existence, at least of course I will myself.

andrewfg:
> My pref: applyChannelTagsToItem

Ok. But then we should use the same `applyChannelTagsToItem` for both the config at individual link level and also the config at global registry level. (My reason for having a shorter key name was to avoid too much typing for the user..)

lolodomo:
Yes, applyChannelTagsToItem is certainly too long to write on each channel link

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/link/ItemChannelLinkRegistry.java
lolodomo: Yes, OK

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/link/ItemChannelLinkRegistry.java
lolodomo: OK, understood.

andrewfg:
So again I suggest to use useTags for both the individual links scope and the global all links scope.

lolodomo:
Yes, `useTags` is not so bad.

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/link/ItemChannelLinkRegistry.java
andrewfg: Moi aussi

andrewfg: See https://github.com/openhab/openhab-distro/pull/1774 -- so all you need to do is remove one # character from the .cfg file..

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/link/ItemChannelLinkRegistry.java
andrewfg: Is it now OK? With the conditional check?

lolodomo:
Code looks good to me.

andrewfg:
Many thanks @lolodomo 

Now all I need to do is fix the iTests not running problem in #4921 (and of course pass them) .. and in light of the knowledge gained from the discussion in #4928 I will try to fix the test class to use the proposed hack.

I will let you know when it is ready..

andrewfg:
> Now all I need to do is fix the iTests

I have spent a whole day on this, and made no progress. I think I simply do not have the skill set to write a working iTest suite for this. The biggest difficulty is that creation of an `ItemChannelLinkRegistry` component in the test environment depends on creation of a `ModelRepository`, a `ThingRegistry`, an `ItemRegistry`, and an `ItemBuilderFactory` component (at least), and these components in turn require the creation of other dependent sub-components themselves. And with all these components being instantiated in internal parts of other modules, it becomes a complete rats nest of circular dependencies etc. So I think it will eventually be quicker and easier to use an actual running instance of OH to do the tests; where all the proper components are magically instantiated by Karaf.

To this end I shall do the following..

1. The test suite is based on a common well known `binding:thing-type:channel` -- namely a color Channel on a Philips hue light Thing; where we know precisely that such channel has specifically [Control, Color] tags.
2. Provide a .things file containing the above
3. Provide a .items file containing several Items each with different combinations of native tags (point, property, custom) and different combinations of useTags settings (true, false, absent)
4. Provide comments in the .items file for each Item what you expect to see in the Item UI
5. Do a visual check to confirm

It is not a great way forward, but I cannot figure out a better way..

andrewfg:
Status update: I am currently live testing it. It works perfect when the .items file is edited or dropped from scratch. But there is an issue on system restart where some tags are missed. I am guessing it is a timing issue that the channels are not yet fully loaded when the items are loaded. So I probably need to register to receive notifications from the thing registry..

andrewfg:
Ok. We are nearly there. Below are the test cases that I am using. I had to make a few more changes, but it does now pass all the tests.

<details><summary>Test case .things file</summary>

```
// in the definitions below you need to set the applicationKey and resourceIds to match your actual hardware
Bridge hue:bridge-api2:bridge "Bridge" [ipAddress="123.123.123.123", applicationKey="REDACTED"] {
    Thing device light "Light" [resourceId="REDACTED"]
    Thing device dimmer "Dimmer" [resourceId="REDACTED"]
}
```

</details>

<details><summary>Test case .items file</summary>

```
String Item_01 "Expected: String" {channel="hue:device:bridge:light:color" }

String Item_02 "Expected: String" {channel="hue:device:bridge:light:color" [useTags=false] }
String Item_03 "Expected: String" {channel="hue:device:bridge:light:color" [useTags="false"] }

String Item_04 "Expected: String · Point > Control > Color" {channel="hue:device:bridge:light:color" [useTags=true] }
String Item_05 "Expected: String · Point > Control > Color" {channel="hue:device:bridge:light:color" [useTags="true"] }

String Item_06 "Expected: String · Point > Control > Color; Custom" ["Custom"] {channel="hue:device:bridge:light:color" [useTags=true] }

// property without a point: OH magically provides 'Control' property
String Item_07 "Expected: String · Point > 'Control' > Power; Custom" ["Power", "Custom"] {channel="hue:device:bridge:light:color" }
String Item_08 "Expected: String · Point > 'Control' > Power; Custom" ["Power", "Custom"] {channel="hue:device:bridge:light:color" [useTags=true] }

String Item_09 "Expected: String · Point > Switch; Custom" ["Switch", "Custom"] {channel="hue:device:bridge:light:color" }
String Item_10 "Expected: String · Point > Switch; Custom" ["Switch", "Custom"] {channel="hue:device:bridge:light:color" [useTags=true] }

String Item_11 "Expected: String · Point > Switch > Power; Custom" ["Switch", "Power", "Custom"] {channel="hue:device:bridge:light:color" }
String Item_12 "Expected: String · Point > Switch > Power; Custom" ["Switch", "Power", "Custom"] {channel="hue:device:bridge:light:color" [useTags=true] }

String Item_13 "Expected: String · Point > Alarm > LowBattery" {channel="hue:device:bridge:dimmer:battery-low" [useTags=true] }

String Item_14 "Expected: String · Point > Control > Color" {channel="hue:device:bridge:light:color" [useTags=true], channel="hue:device:bridge:dimmer:battery-low" [useTags=true] }
String Item_15 "Expected: String · Point > Alarm > LowBattery" {channel="hue:device:bridge:light:color" [useTags=false], channel="hue:device:bridge:dimmer:battery-low" [useTags=true] }
```

</details>

<details><summary>Logger output</summary>

```
2025-07-26 16:26:51.460 [DEBUG] [e.thing.link.ItemChannelLinkRegistry] - Item 'Item_12' already tagged; ignoring tags supplied by channel 'hue:device:bridge:light:color'.
2025-07-26 16:26:51.462 [DEBUG] [e.thing.link.ItemChannelLinkRegistry] - Item 'Item_08' already tagged; ignoring tags supplied by channel 'hue:device:bridge:light:color'.
2025-07-26 16:26:51.463 [INFO ] [e.thing.link.ItemChannelLinkRegistry] - Item 'Item_05' adding tags '[Control, Color]' supplied by channel 'hue:device:bridge:light:color'.
2025-07-26 16:26:51.469 [DEBUG] [e.thing.link.ItemChannelLinkRegistry] - Item 'Item_10' already tagged; ignoring tags supplied by channel 'hue:device:bridge:light:color'.
2025-07-26 16:26:51.469 [INFO ] [e.thing.link.ItemChannelLinkRegistry] - Item 'Item_04' adding tags '[Control, Color]' supplied by channel 'hue:device:bridge:light:color'.
2025-07-26 16:26:51.472 [INFO ] [e.thing.link.ItemChannelLinkRegistry] - Item 'Item_14' adding tags '[LowBattery, Alarm]' supplied by channel 'hue:device:bridge:dimmer:battery-low'.
2025-07-26 16:26:51.475 [DEBUG] [e.thing.link.ItemChannelLinkRegistry] - Item 'Item_14' already tagged; ignoring tags supplied by channel 'hue:device:bridge:light:color'.
2025-07-26 16:26:51.475 [INFO ] [e.thing.link.ItemChannelLinkRegistry] - Item 'Item_06' adding tags '[Control, Color]' supplied by channel 'hue:device:bridge:light:color'.
2025-07-26 16:26:51.477 [INFO ] [e.thing.link.ItemChannelLinkRegistry] - Item 'Item_15' adding tags '[LowBattery, Alarm]' supplied by channel 'hue:device:bridge:dimmer:battery-low'.
2025-07-26 16:26:51.480 [INFO ] [e.thing.link.ItemChannelLinkRegistry] - Item 'Item_13' adding tags '[LowBattery, Alarm]' supplied by channel 'hue:device:bridge:dimmer:battery-low'.
```

</details>

Note: as you can see in the log for for `Item_14` when the item is bound to two channels and both of them have `useTags=true` then it takes the tags from whichever one of them came first in the set. So there is a 50:50 chance of it being the 'right' one. And in that case the user can see it in the log and manually suppress the 'wrong' one by applying `useTags=false` on it.

andrewfg:
@lolodomo / @jimtng in the meantime this is now ready to go; but perhaps you want to do a final test?

andrewfg:
See https://github.com/openhab/openhab-docs/pull/2529

andrewfg:
I am not sure why the CI build fails with an error on the voice integration tests. I did nothing that should impact that part. ??

andrewfg:
I also finally figured out how to add the OSGi integration test (and it passes).

andrewfg:
This pull request introduces significant enhancements to the `ItemChannelLinkRegistry` and related classes to support automatic assignment and management of default semantic tags for items based on their linked channels. Additionally, it includes minor dependency updates in test configurations. Below is a summary of the most important changes:

### Enhancements to `ItemChannelLinkRegistry` and Tag Management:

1. **Automatic Tag Assignment and Management**:
   * Introduced methods to assign default semantic tags from linked channels to items, provided the channel has `useTags=true` and the item does not already have conflicting tags. Warnings are logged for conflicting tags or multiple linked channels with `useTags=true`. (`[bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/link/ItemChannelLinkRegistry.javaR285-R487](diffhunk://#diff-3c6274c126dc7186b400a8a006e8f8849b086463034900a11730973419ec003fR285-R487)`)
   * Added logic to remove tags supplied by a channel when the link is removed or updated, ensuring native custom tags remain intact. (`[bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/link/ItemChannelLinkRegistry.javaR285-R487](diffhunk://#diff-3c6274c126dc7186b400a8a006e8f8849b086463034900a11730973419ec003fR285-R487)`)
   * Implemented global and per-channel configuration for tag usage via the `USE_TAGS` property. (`[[1]](diffhunk://#diff-3c6274c126dc7186b400a8a006e8f8849b086463034900a11730973419ec003fR70-R114)`, `[[2]](diffhunk://#diff-3c6274c126dc7186b400a8a006e8f8849b086463034900a11730973419ec003fR285-R487)`)

2. **Event Handling and Registry Integration**:
   * Registered the `ItemChannelLinkRegistry` as a listener for item registry changes and system start level events to manage tags dynamically during item lifecycle events (addition, removal, updates). (`[[1]](diffhunk://#diff-3c6274c126dc7186b400a8a006e8f8849b086463034900a11730973419ec003fR70-R114)`, `[[2]](diffhunk://#diff-3c6274c126dc7186b400a8a006e8f8849b086463034900a11730973419ec003fR285-R487)`)
   * Added `notifyListenersAboutItemExternalUpdate` to `ItemRegistry` for notifying listeners of external item updates, ensuring proper synchronization. (`[[1]](diffhunk://#diff-6b9dd17d5481dad9b22edc83b356d73477460b15c8b4f4bb30ba2fdd64983158R467-R471)`, `[[2]](diffhunk://#diff-856755163fef4f78f17c08b696bc172894935c6761a66c9a1205454dee2d043cR108-R114)`)

3. **Refactoring and Code Simplification**:
   * Simplified stream operations in `getLinkedItems` and `getBoundThings` methods for better readability. (`[[1]](diffhunk://#diff-3c6274c126dc7186b400a8a006e8f8849b086463034900a11730973419ec003fL89-R141)`, `[[2]](diffhunk://#diff-3c6274c126dc7186b400a8a006e8f8849b086463034900a11730973419ec003fL100-R152)`)
   * Added a transient `tagsLinked` flag to `ItemChannelLink` to track whether tags were assigned by a specific channel. (`[[1]](diffhunk://#diff-a985152020c0e792d8079ee081e17eae000b8d2e0a8bfbb1b927fe9252eecc2fR34-R35)`, `[[2]](diffhunk://#diff-a985152020c0e792d8079ee081e17eae000b8d2e0a8bfbb1b927fe9252eecc2fR77-R87)`)

### Dependency Updates in Test Configurations:

4. **Test Configuration Updates**:
   * Updated dependencies in `itest.bndrun` files to include `org.openhab.core.model.persistence.runtime` and remove unused `org.openhab.core.model.rule.runtime`. (`[[1]](diffhunk://#diff-0d15e70053365c51dfcffdb077af6e8c4c79c365cb4ddb2fe3e085033806f079L127-R132)`, `[[2]](diffhunk://#diff-e4b49351566f44aa46c31af7a2cee504af1b7441368afc2a4296b6049c1cb3bfL134-R135)`, `[[3]](diffhunk://#diff-9eaac1e2084455180924aa70459c1bf05c71197553c619d9bb75229564507186L133-R138)`, `[[4]](diffhunk://#diff-41e516a93b6d481f8b4babc9d49c8ebd022942000d046d65f73f8d769d672f34L136-R141)`)

andrewfg:
I am guessing the OH core maintainers are looking for signs of qualified support from other OH experts. So @lolodomo it might help if you want this if you could give a formal LGTM review..

lolodomo:
> I am guessing the OH core maintainers are looking for signs of qualified support from other OH experts. So @lolodomo it might help if you want this if you could give a formal LGTM review..

How can I identify all your commits after my previous "Code looks good to me" ?

andrewfg:
> How can I identify all your commits after my previous "Code looks good to me" ?

Umm. Probably with great difficulty. So I would suggest to look at the last final state of the 12 modified files. (Of which 2 are unit tests, and 6 are auto generated bndrun files .. so really you only need to check 4 files)

lolodomo:
Why are you not pushing commits as we all do which means the timing is kept and we as reviewers immediately see what commits are not yet read ?

andrewfg:
> Why are you not pushing commits as we all do

Sorry but I don't understand the question. Can you please explain?

EDIT: perhaps you are referring to the date tags on my commits? And indeed they all show the same (today's) date. The reason for that is that I had to do a rebase in order to resolve a merge conflict on the bndrun files. And git destroys the original commit date on such commits. If you know some way to prevent git from doing that (or providing the original date as well) then please let me know.

andrewfg:
@lolodomo I made this PR at your wish; so if you don’t support it anymore then I should probably close it; or??

lolodomo:
Of course I am supporting your PR and of course I am still interested.
Sorry if you consider I am not enough involved in openHAB and not spending enough of my free time as contributor and reviewer. I consider myself this spent time as already a lot of my free time but I am not complaining at all as I do it for fun meaning without any constraint, when I can and when I want.

andrewfg:
I just want to move this forward. And the OH core maintainers won't do that unless they see some agreement that it is necessary.

itests/org.openhab.core.model.item.tests/itest.bndrun
```diff
@@ -128,4 +127,13 @@ Fragment-Host: org.openhab.core.model.item
...
+	org.yaml.snakeyaml;version='[2.4.0,2.4.1)',\
+	org.openhab.core.model.persistence.runtime;version='[5.1.0,5.1.1)',\
+	org.osgi.service.cm;version='[1.6.0,1.6.1)'
+	org.osgi.service.cm;version='[1.6.0,1.6.1)',\
+	org.ops4j.pax.logging.pax-logging-api;version='[2.2.8,2.2.9)',\
+	com.fasterxml.jackson.dataformat.jackson-dataformat-yaml;version='[2.18.2,2.18.3)',\
+	org.yaml.snakeyaml;version='[2.3.0,2.3.1)',\
+	de.focus_shift.jollyday-core;version='[1.5.4,1.5.5)',\
+	de.focus_shift.jollyday-jackson;version='[1.5.4,1.5.5)',\
+	org.openhab.core.model.persistence.runtime;version='[5.1.0,5.1.1)'
```
lolodomo: I see in your changed itest.bndrun that you added duplicated lines, that looks strange.

It is the case in this file but also in few others.

andrewfg: I see it too. The files are auto generated via `mvn -DwithResolver` and not manually created. So maybe there is an issue with the mvn profile?

itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/link/ItemChannelLinkTaggingOSGiTest.java
```diff
@@ -0,0 +1,176 @@
...
+        assertEquals(Set.of("Control", "Color"), item.getTags());
+
+        item = itemRegistry.get("Item_06");
+        assertNotNull(item);
+        assertEquals(Set.of("Control", "Color", "Custom"), item.getTags());
+
+        item = itemRegistry.get("Item_07");
+        assertNotNull(item);
+        assertTrue(item.getTags().contains("Power"));
+        assertTrue(item.getTags().contains("Custom"));
```
lolodomo: Why don't you use
`assertEquals(Set.of("Power", "Custom"), item.getTags());`
in that case and also for Item_08 ?

lolodomo:
Except changes in itests.bndrun for which I have not the skills to comment (even if I see strange duplicated lines introduced) and a minor comment in a test, code looks good to me.

andrewfg:
@holgerfriedrich / @wborn could you please have a look at the bndrun files in this PR? The changes were created via `mvn -DwithResolver` but that seems to have created duplicate entries for the same dependency with different versions. And added more dependencies than I would have expected. => Perhaps I should not have used -DwithResolver? What is the correct approach?

holgerfriedrich:
That looks strange.I messed up the hivemq upgrade, so netty deps are broken  in core. I will work on it, but not before later in the evening.

Using the resolver is fine. Your PR seems to inclue many changes. Resolver acts differently depending on OS, so deps like osgi.cm or rs-api may show up or vanish. I sometimes change it back manually.

andrewfg:
@holgerfriedrich I think the issue was due to missing trailing backslashes `\` whereby `-DwithResolver` considers the following two lines to be different, so it keeps the former rather than overwriting it with the latter. Now fixed.

```
org.some.thing;version='[1.0.0,1.0.1)'
org.some.thing;version='[1.0.0,1.0.1)'\
```

itests/org.openhab.core.thing.tests/src/main/java/org/openhab/core/thing/link/ItemChannelLinkTaggingOSGiTest.java
andrewfg: Well spotted. I am happy to change to your version if you wish, but I thought I should explain my logic and let you make the final decision..
 
A Property tag without a Point tag is generally not allowed, so `[Power, Custom]` is invalid. When I wrote the test I was imagining that (perhaps at some time in the future) the item parser might "auto-magically" insert an appropriate (guessed) Point. However such Point would be of type unknown to the test suite. So I specifically wrote them to check only the Property tag and the Custom tag. With the current design of the item parser, both your and my version of the tests do produce the same result, so we can easily adopt yours instead of mine.

=> @lolodomo please let me know if you want me to adopt your version?

andrewfg: Indeed test 08 could be an example of the "not" auto-magic use case .. the Item has ["Power", "Custom"] and the Channel has ["Control", "Color"] .. in some universes you could imagine useTags would (wrongly) merge the Point from the Channel with the Property from the Item to produce ["Control", "Power"] -- but the good news is that the current code does NOT do that!

```
String Item_08 "'Control', Power, Custom" ["Power", "Custom"] {channel="hue:device:dummy:color" [useTags=true] }
```

lolodomo: Ok
I was just surprised of the different approach compared to other items.

jimtng: There was a piece of code in core that automatically adds a Point tag (Measurement, or Control) if an item only has Property tag.

andrewfg: @jimtng you say "was" rather than "is". Is that code likely to return? In which case I should probably revert the tests. Or??

jimtng: I just checked, I think it's still there

lolodomo: Where please ?

andrewfg: It is obviously in the codepath after the code used by the itest. (But I suppose that could change; so I think it is better to revert these two tests just in case)..

lolodomo: I don't understand. This is not code in core framework?

andrewfg: The itest does indeed use (part of) the core model parser to create its test Items. However it apparently takes the Items from quite deep in the core code, presumably before any auto magic corrections are applied, and before any validation checks and failures would be applied.

The purpose of these tests is NOT to test the model parser but to test the useTags functionality. So we should structure the tests to ONLY check stuff that may or may not have been contributed via useTags.

jimtng: I finally found it. It doesn't actually add the tag, but returning an item as a point even when a Point tag is missing, with only a property tag

https://github.com/openhab/openhab-core/blob/8537a1b82278159d7ec69c1bb45508ca2c9537a1/bundles/org.openhab.core.semantics/src/main/java/org/openhab/core/semantics/SemanticTags.java#L64-L82

This has the effect that the MainUI will show the item with Point and Property, but it doesn't affect what item.getTags() would return, and you could do `assertEquals(Set.of("Power", "Custom"), item.getTags())` like @lolodomo mentioned if you prefer. But I'm not making a suggestion one way or another.

andrewfg: Ok. I will leave the test as is, in @lolodomo's form, and mark this again as resolved.

lolodomo:
@openhab/core-maintainers : this PR looks globally OK to me and seems to be really ready for a final review by you with two particular points I would suggest to take attention because I am not an expert:

1. the changes in itests.bndrun 
2. the usage of the start level

andrewfg:
> in cases where the user has configured useTags=true at a global level, then it shall be possible to set useTags=false at item level in order to configure an item specific exception to the global rule.

EDIT: Done!

lolodomo:
> TODO in cases where the user has configured useTags=true at a global level, then it shall be possible to set useTags=false at item level in order to configure an item specific exception to the global rule.

Ah yes, if not yet possible, I missed something very important..

itests/org.openhab.core.model.script.tests/itest.bndrun
```diff
@@ -135,4 +134,11 @@ Fragment-Host: org.openhab.core.model.script
 	com.fasterxml.jackson.core.jackson-databind;version='[2.19.2,2.19.3)',\
 	com.fasterxml.jackson.dataformat.jackson-dataformat-xml;version='[2.19.2,2.19.3)',\
 	com.fasterxml.jackson.dataformat.jackson-dataformat-yaml;version='[2.19.2,2.19.3)',\
-	org.yaml.snakeyaml;version='[2.4.0,2.4.1)'
+	org.yaml.snakeyaml;version='[2.4.0,2.4.1)',\
+	org.openhab.core.model.persistence.runtime;version='[5.1.0,5.1.1)'
```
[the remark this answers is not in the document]
andrewfg: Done

itests/org.openhab.core.model.script.tests/itest.bndrun
```diff
@@ -135,4 +134,11 @@ Fragment-Host: org.openhab.core.model.script
...
+	org.openhab.core.model.rule.runtime;version='[5.1.0,5.1.1)',\
+	org.ops4j.pax.logging.pax-logging-api;version='[2.2.8,2.2.9)',\
+	com.fasterxml.jackson.dataformat.jackson-dataformat-yaml;version='[2.18.2,2.18.3)',\
+	de.focus_shift.jollyday-core;version='[1.5.4,1.5.5)',\
+	de.focus_shift.jollyday-jackson;version='[1.5.4,1.5.5)',\
+	org.openhab.core.model.persistence.runtime;version='[5.1.0,5.1.1)'
```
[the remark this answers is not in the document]
andrewfg: Done

bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/link/ItemChannelLinkRegistry.java
```diff
@@ -229,4 +282,208 @@ public enum ItemChannelLinkProblem {
...
+     * if the item does already have a Point and/or a Property tag then we write a warning message in
+     * the log. The warning is also logged if the item has more than one linked channel with 'useTags=true'.
+     * And in any case if the item has native custom tags then those tags remain.
+     */
+    private void assignChannelDefaultTags(ItemChannelLink link, ActiveItem activeItem) {
+        boolean alreadyHasPointOrPropertyTag = false;
+
+        for (String tag : activeItem.getTags()) {
+            Class<? extends Tag> type = SemanticTags.getById(tag);
+            if ((type != null) && (Point.class.isAssignableFrom(type) || Property.class.isAssignableFrom(type))) {
```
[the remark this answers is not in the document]
andrewfg: Caching is not a good idea. Suggestion rejected.

andrewfg:
@wborn it seems that @holgerfriedrich is pushing dependency updates faster than I can follow with fixing the bndrun files ;) so I can't get it to run the iTests => let us wait until tomorrow..

holgerfriedrich:
> @wborn it seems that @holgerfriedrich is pushing dependency updates faster than I can follow with fixing the bndrun files ;) so I can't get it to run the iTests => let us wait until tomorrow..

@andrewfg sorry for that! It seems today is patchday, bringing you all the latest features (and bugs) of the week ;-) ;-)
We are almost finished with the planned upgrades for 5.1, so it should be more relaxed in a few days....

andrewfg:
@holgerfriedrich / @wborn many thanks for the support; it is now all fixed and builds without error..

holgerfriedrich:
@andrewfg There are again some conflicts in the itests, caused by me in #5005.... Sorry for that.

lolodomo:
@andrewfg : not available to fix the conflicts soon ? The milestone build will start very soon

andrewfg:
> not available to fix the conflicts soon ? The milestone build will start very soon

@lolodomo I finally just did it

andrewfg:
@holgerfriedrich can you please do a final check that my fix to your break of my pr did not break yours..

andrewfg:
@lolodomo / @holgerfriedrich  / @kaikreuzer => Can you please not leave it until the last minute before a milestone before changing the itests thus breaking the build and preventing this PR from ever being merged? Can you please either merge it now into a snapshot, or at least give me 24 hours to sort out last minute merge conflicts?

lolodomo:
@openhab/core-maintainers : can you merge it please? I would like to test it in a snapshot before the next milestone.

PR #4922, opened 2025-07-22
AbstractScriptModuleHandler: Recompile scripts on dependency change

florian-h05:
Fixes https://github.com/openhab/openhab-addons/issues/19091.

When a script's dependency changes, it should recompile the compiled script similarly to resetting the engine for uncompiled scripts. Fixing this behaviour fixes an issue where compiled scripts stopped working after a dependency changed.

This also simplifies the code a bit.

### DIFF
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/AbstractScriptModuleHandler.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/AbstractScriptModuleHandler.java
@@ -93,15 +93,10 @@ public abstract class AbstractScriptModuleHandler<T extends Module> extends Base
      * {@link Compilable}.
      */
     protected void compileScript() throws ScriptException {
-        if (compiledScript.isPresent()) {
+        if (compiledScript.isPresent() || script.isEmpty()) {
             return;
         }
-        if (!scriptEngineManager.isSupported(this.type)) {
-            logger.debug(
-                    "ScriptEngine for language '{}' could not be found, skipping compilation of script for identifier: {}",
-                    type, engineIdentifier);
-            return;
-        }
+
         Optional<ScriptEngine> engine = getScriptEngine();
         if (engine.isPresent()) {
             ScriptEngine scriptEngine = engine.get();
@@ -119,11 +114,20 @@ public abstract class AbstractScriptModuleHandler<T extends Module> extends Base
 
     /**
      * Reset the script engine to force a script reload
-     *
      */
     public synchronized void resetScriptEngine() {
         scriptEngineManager.removeEngine(engineIdentifier);
         scriptEngine = Optional.empty();
+        compiledScript = Optional.empty();
+    }
+
+    /**
+     * Gets the unique identifier of the rule this module handler is used for.
+     * 
+     * @return the UID of the rule
+     */
+    public String getRuleUID() {
+        return ruleUID;
     }
 
     /**
@@ -135,10 +139,20 @@ public abstract class AbstractScriptModuleHandler<T extends Module> extends Base
         return engineIdentifier;
     }
 
+    /**
+     * Get the script engine instance used by this module handler.
+     * 
+     * @return the script engine instance if available, otherwise Optional.empty()
+     */
     protected Optional<ScriptEngine> getScriptEngine() {
         return scriptEngine.isPresent() ? scriptEngine : createScriptEngine();
     }
 
+    /**
+     * Creates a new script engine for the type defined in the module configuration.
+     * 
+     * @return the script engine if available, otherwise Optional.empty()
+     */
     private Optional<ScriptEngine> createScriptEngine() {
         ScriptEngineContainer container = scriptEngineManager.createScriptEngine(type, engineIdentifier);
 
@@ -203,13 +217,15 @@ public abstract class AbstractScriptModuleHandler<T extends Module> extends Base
     }
 
     /**
-     * Evaluates the passed script with the ScriptEngine.
+     * Evaluates the script with the given script engine.
      *
      * @param engine the script engine that is used
-     * @param script the script to evaluate
      * @return the value returned from the execution of the script
      */
-    protected @Nullable Object eval(ScriptEngine engine, String script) {
+    protected @Nullable Object eval(ScriptEngine engine) {
+        if (script.isEmpty()) {
+            return null;
+        }
         try {
             if (compiledScript.isPresent()) {
                 logger.debug("Executing pre-compiled script of rule with UID '{}'", ruleUID);
--- a/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/ScriptActionHandler.java
+++ b/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/ScriptActionHandler.java
@@ -91,7 +91,7 @@ public class ScriptActionHandler extends AbstractScriptModuleHandler<Action> imp
             }
             try {
                 setExecutionContext(scriptEngine, context);
-                Object result = eval(scriptEngine, script);
+                Object result = eval(scriptEngine);
                 resultMap.put("result", result);
                 resetExecutionContext(scriptEngine, context);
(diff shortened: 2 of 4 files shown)


bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/AbstractScriptModuleHandler.java
```diff
@@ -119,11 +114,20 @@ public void dispose() {
...
         scriptEngineManager.removeEngine(engineIdentifier);
         scriptEngine = Optional.empty();
+        compiledScript = Optional.empty();
+    }
+
+    /**
+     * Gets the unique identifier of the rule this module handler is used for.
+     * 
+     * @return the UID of the rule
+     */
```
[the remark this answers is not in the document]
florian-h05: The non null by default annotation requires ruleUID to be not null, Copilot seems to not understand that annotation.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/AbstractScriptModuleHandler.java
```diff
@@ -93,15 +93,10 @@ private static String getValidConfigParameter(String parameter, Configuration co
...
-        if (compiledScript.isPresent()) {
-            return;
-        }
-        if (!scriptEngineManager.isSupported(this.type)) {
-            logger.debug(
-                    "ScriptEngine for language '{}' could not be found, skipping compilation of script for identifier: {}",
-                    type, engineIdentifier);
+        if (compiledScript.isPresent() || script.isEmpty()) {
             return;
         }
```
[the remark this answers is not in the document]
florian-h05: The child classes return if script is empty before calling the eval method of the AbstractScriptModuleHandler, and even if not, there is no harm in not having the compiled script.

florian-h05:
Can you please backport?

florian-h05:
@openhab/core-maintainers Can we also backport this to 4.3.x?

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/AbstractScriptModuleHandler.java
```diff
@@ -93,15 +93,10 @@ private static String getValidConfigParameter(String parameter, Configuration co
...
      */
     protected void compileScript() throws ScriptException {
-        if (compiledScript.isPresent()) {
-            return;
-        }
-        if (!scriptEngineManager.isSupported(this.type)) {
-            logger.debug(
-                    "ScriptEngine for language '{}' could not be found, skipping compilation of script for identifier: {}",
-                    type, engineIdentifier);
+        if (compiledScript.isPresent() || script.isEmpty()) {
```
Nadahar, line 96: @florian-h05  Could this change be related to the errors that are now logged and are discussed in #4965? I don't know exactly what's happening here, but it seems that there used to be a check if the script engine was available before attempting compilation that disappeared with this change, which could theoretically lead to what we're seeing.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/ScriptActionHandler.java
```diff
@@ -91,7 +91,7 @@ public void compile() throws ScriptException {
             }
             try {
                 setExecutionContext(scriptEngine, context);
-                Object result = eval(scriptEngine, script);
+                Object result = eval(scriptEngine);
                 resultMap.put("result", result);
```
Nadahar, line 95: There is a bug here - `eval()` might return `null`, in which case the result shouldn't be put in the map, since the map is annotated as not containing `null` entries. The "protection" of checking if `script` is "empty" above isn't sufficient, since `eval()` might return `null` under other circumstances as well.

florian-h05: That behaviour isn’t modified here though, the result entry of the map did contain null in many cases since I had a first look.
I would rather say the annotation is wrong then.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/AbstractScriptModuleHandler.java
florian-h05: Yeah, that very likely is the cause if I think about this a bit further.
compileScript might get called during startup by the RuleEngineImpl before the engine is ready. When doing this change I removed that check because the changed code does the same job, but the check was indeed required to avoid the log.

Can you create a PR adding back that check together with a comment why it’s needed?

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/ScriptActionHandler.java
Nadahar: > I would rather say the annotation is wrong then.

Yeah, if `null` results are "normal", the annotation is definitely wrong. But then it's wrong in the `ActionHandler` interface as well. It should have been:
```java
    @Nullable Map<String, @Nullable Object> execute(Map<String, Object> context);
```

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/AbstractScriptModuleHandler.java
Nadahar: > Can you create a PR adding back that check together with a comment why it’s needed?

I can, but I'm not sure that I have quite enough overview to do this and be 100% confident that I made the correct change. But, I'll study it and see what I can come up with.

Nadahar: > Can you create a PR adding back that check together with a comment why it’s needed?

#4971 created. Please note that most of my core PRs are untouched, so you probably need to pull some strings to get it to move.

bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/handler/ScriptActionHandler.java
Nadahar: I included the annotation change in the PR, but now I'm starting to wonder: Is it actually useful to put null entries in the map? It has quite limited use, you have to use `map.contains(key)` to even know that the entry exists, if you try to retrieve it using `map.get(key)`, you will get `null` - the exact same as if the entry doesn't exist, so most code will then conclude that there's no entry for said key. To add to that, although `HashMap` supports `null` values, "most" `Map`s don't. So, anybody handing this `Map` must make sure not to copy it into some other type that don't support `null` values.

It's very easy to instead modify the code to not `put` they key/value pair if the value is `null`, to the question is: what is the correct action? What other code inserts `null` values in there, and what are they used for?

I'm asking just to make sure that changing the annotation really is the right thing to do.

Nadahar: I get that there might be a point in returning the fact that an output got a `null` value, I can only really find two places this `Map` is used, `ThingActionsResource` (REST) and `RuleEngineImpl`. While `ThingActionsResource` seems to be able to handle `null` values (assuming that the JSON serializer doesn't choke on them), `RuleEngineImpl` doesn't seem to expect `null` values. It goes on to put them in `context` with different keys but the same values. The `context` map is also annotated to not allow `null` values, but the static analyzer doesn't seem to be able to pick this up, so it doesn't complain.

When such annotations are used, it's somewhat important that they are accurate IMO, especially since `null` values in `Map`s are so rare in Java, because it's natural for a developer dealing with this to assume no to expect `null` values when the annotation says so.

Issue #4903, opened 2025-07-15
Check of item name is wrong in DSL/YAML providers


lolodomo:
Check should be aligned to what is finally accepted by the items registry.

PR #4928, opened 2025-07-25
DSL Items Parser: Fix incorrect parsing of keywords encountered in the wrong context

### LINKED ISSUE #4927: ModelRepositoryImpl error parsing (some) native item tags
There is a bug in `ModelRepositoryImpl` where it tries to parse native item tags. Specifically the tag set [Status, foo, bar] loads OK but the tag set [Switch, foo, bar] causes an error..

#### Test case 1 (loads OK)

```
// an item with native tags [Status, foo, bar] loads OK
Color Office_Lamp_Colour "Office Lamp Colour" [Status, foo, bar] {channel="hue:device:001788fffe2157c7:24e7a04e-ad36-4192-961e-ca7f22103cbc:color" [ useTags=true ] }
```
#### Test case 2 (does not load)

```
// an item with native tags [Switch, foo, bar] causes an error
Color Office_Lamp_Colour "Office Lamp Colour" [Switch, foo, bar] {channel="hue:device:001788fffe2157c7:24e7a04e-ad36-4192-961e-ca7f22103cbc:color" [ useTags=true ] }

// log entry below
2025-07-24 18:40:36.360 [WARN ] [el.core.internal.ModelRepositoryImpl] - Configuration model '24g.items' has errors, therefore ignoring it: [1,48]: no viable alternative at input 'Switch'
[1,54]: extraneous input ',' expecting RULE_ID
```

_Originally posted by @andrewfg in https://github.com/openhab/openhab-core/issues/4913#issuecomment-3114335143_

jimtng:
When specifying tag names that match one of the valid item types, e.g. `Switch`, the parser incorrectly treated it as a new start of an item definition.

Fix #4927

~~Consider backporting to 4.3.x~~ This is a bigger change than I originally thought, it shouldn't be backported

Longer explanation:

The standard way xtext parser / grammar deals with this is to prefix conflicting literals such as `Switch`, `AND`, `Group`, etc with a caret `^`. So even without this PR, this would work:

```
Switch Test1 [^Switch]
```

This PR makes this possible:
```
Switch Test1 [Switch]
```

See Copilot review for more details.

Also "fixed": removed space in the generated DSL item format, from:

```
Group:Switch:OR (ON,OFF) Name
```

to

```
Group:Switch:OR(ON,OFF) Name // remove space before (
```

Copilot Summary

## Pull Request Overview
This PR fixes incorrect parsing of DSL items when keywords matching valid item types (like "Switch", "Group", etc.) are used in other contexts such as tag names. Previously, the parser would incorrectly treat these as new item definitions.

* Simplifies the grammar by removing separate `ModelGroupItem` and `ModelNormalItem` types in favor of a unified `ModelItem` approach
* Updates validation logic to properly validate item types, group functions, and dimensions with improved error messages
* Refactors code generation to remove unnecessary space in group function syntax (e.g., `Group:Switch:OR(ON,OFF)` instead of `Group:Switch:OR (ON,OFF)`)

### Reviewed Changes
Copilot reviewed 16 out of 16 changed files in this pull request and generated 2 comments.

Show a summary per file
| File                                                                                                                     | Description                                                                                 |
|--------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------|
| bundles/org.openhab.core.model.item/src/org/openhab/core/model/Items.xtext                                               | Unified grammar to support flexible item type parsing with up to 4 colon-separated segments |
| bundles/org.openhab.core.model.item/src/org/openhab/core/model/item/internal/GenericItemProvider.java                    | Major refactor to handle unified item creation logic with improved group item parsing       |
| bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend                           | Enhanced validation with comprehensive item type, group function, and dimension checking    |
| bundles/org.openhab.core/src/main/java/org/openhab/core/items/GroupFunction.java                                         | Added constants for valid group functions to centralize function validation                 |
| bundles/org.openhab.core/src/main/java/org/openhab/core/library/CoreItemFactory.java                                     | Exposed valid item types as a public constant for validation use                            |
| itests/org.openhab.core.model.item.tests/src/main/java/org/openhab/core/model/item/internal/GenericItemProviderTest.java | Added comprehensive tests for keyword parsing edge cases and group type syntax              |

### DIFF
--- a/bundles/org.openhab.core.model.item/src/org/openhab/core/model/Items.xtext
+++ b/bundles/org.openhab.core.model.item/src/org/openhab/core/model/Items.xtext
@@ -14,31 +14,18 @@ ItemModel:
 ;
 
 ModelItem:
-	(ModelNormalItem | ModelGroupItem) name=ID
+	type=ModelItemType ('(' args+=(ID|STRING) (',' args+=(ID|STRING))* ')')?
+	name=ID
 	(label=STRING)?
 	('<' icon=Icon '>')?
-	('(' groups+=ID (',' groups+=ID)* ')')? 
+	('(' groups+=ID (',' groups+=ID)* ')')?
 	('[' tags+=(ID|STRING) (',' tags+=(ID|STRING))* ']')?
-	('{' bindings+=ModelBinding (',' bindings+=ModelBinding)* '}')? 
+	('{' bindings+=ModelBinding (',' bindings+=ModelBinding)* '}')?
 ;
 
-ModelGroupItem:
-	{ModelGroupItem} 'Group' (':' type=ModelItemType ( ':' function=ModelGroupFunction ('(' args+=(ID|STRING) (',' args+=(ID|STRING))* ')')?)?)?
-;
-
-enum ModelGroupFunction:
-	EQUALITY='EQUALITY' | AND='AND' | OR='OR' | NAND='NAND' | NOR='NOR' | XOR='XOR' | AVG='AVG' | MEDIAN='MEDIAN' | SUM='SUM' | MAX='MAX' | MIN='MIN' | COUNT='COUNT' | LATEST='LATEST' | EARLIEST='EARLIEST'
-;
-
-ModelNormalItem:
-	type=ModelItemType
-;
-
+// Supports item types with up to 4 colon-separated segments, e.g., Group:Number:Dimension:MAX
 ModelItemType:
-	BaseModelItemType | ('Number' (':' ID)?)
-;
-BaseModelItemType:
-	'Switch' | 'Rollershutter' | 'String' | 'Dimmer' | 'Contact' | 'DateTime' | 'Color' | 'Player' | 'Location' | 'Call' | 'Image'
+   ID (':' ID (':' ID (':' ID )?)?)?
 ;
 
 ModelBinding:
@@ -56,7 +43,7 @@ ValueType returns ecore::EJavaObject:
     STRING | NUMBER | BOOLEAN
 ;
 
-BOOLEAN returns ecore::EBoolean: 
+BOOLEAN returns ecore::EBoolean:
     'true' | 'false'
 ;
 
--- a/bundles/org.openhab.core.model.item/src/org/openhab/core/model/formatting/ItemsFormatter.xtend
+++ b/bundles/org.openhab.core.model.item/src/org/openhab/core/model/formatting/ItemsFormatter.xtend
@@ -25,8 +25,10 @@ class ItemsFormatter extends AbstractDeclarativeFormatter {
 	@Inject extension ItemsGrammarAccess
 
 	override protected void configureFormatting(FormattingConfig c) {
-		c.setLinewrap(1, 1, 2).before(modelGroupItemRule)
-		c.setLinewrap(1, 1, 2).before(modelNormalItemRule)
+		c.setLinewrap(1, 1, 2).before(modelItemRule)
+
+		// No space between the item type and the opening parenthesis for the group function arguments
+		c.setNoSpace().between(modelItemTypeRule, modelItemAccess.leftParenthesisKeyword_1_0)
 
 		c.setNoSpace().withinKeywordPairs("<", ">")
 		c.setNoSpace().withinKeywordPairs("(", ")")
--- a/bundles/org.openhab.core.model.item/src/org/openhab/core/model/item/internal/fileconverter/DslItemFileConverter.java
+++ b/bundles/org.openhab.core.model.item/src/org/openhab/core/model/item/internal/fileconverter/DslItemFileConverter.java
@@ -35,6 +35,7 @@ import org.openhab.core.config.core.ConfigUtil;
 import org.openhab.core.items.GroupFunction;
 import org.openhab.core.items.GroupItem;
 import org.openhab.core.items.Item;
+import org.openhab.core.items.ItemUtil;
 import org.openhab.core.items.Metadata;
 import org.openhab.core.items.fileconverter.AbstractItemFileGenerator;
 import org.openhab.core.items.fileconverter.ItemFileGenerator;
@@ -45,8 +46,6 @@ import org.openhab.core.model.item.internal.GenericMetadataProvider;
 import org.openhab.core.model.items.ItemModel;
 import org.openhab.core.model.items.ItemsFactory;
 import org.openhab.core.model.items.ModelBinding;
-import org.openhab.core.model.items.ModelGroupFunction;
-import org.openhab.core.model.items.ModelGroupItem;
 import org.openhab.core.model.items.ModelItem;
 import org.openhab.core.model.items.ModelProperty;
 import org.openhab.core.types.State;
@@ -115,26 +114,25 @@ public class DslItemFileConverter extends AbstractItemFileGenerator implements I
 
     private ModelItem buildModelItem(Item item, List<Metadata> channelLinks, List<Metadata> metadata,
             @Nullable String stateFormatter, boolean hideDefaultParameters) {
-        ModelItem model;
+        ModelItem model = ItemsFactory.eINSTANCE.createModelItem();
         if (item instanceof GroupItem groupItem) {
-            ModelGroupItem modelGroup = ItemsFactory.eINSTANCE.createModelGroupItem();
-            model = modelGroup;
             Item baseItem = groupItem.getBaseItem();
+            List<String> groupType = new ArrayList<>();
+            groupType.add(groupItem.getType());
             if (baseItem != null) {
-                modelGroup.setType(baseItem.getType());
+                groupType.add(baseItem.getType());
                 GroupFunction function = groupItem.getFunction();
                 if (function != null) {
-                    ModelGroupFunction modelFunction = ModelGroupFunction
-                            .getByName(function.getClass().getSimpleName().toUpperCase());
-                    modelGroup.setFunction(modelFunction);
+                    groupType.add(function.getClass().getSimpleName().toUpperCase());
                     State[] parameters = function.getParameters();
                     for (int i = 0; i < parameters.length; i++) {
-                        modelGroup.getArgs().add(parameters[i].toString());
+                        model.getArgs().add(parameters[i].toString());
                     }
                 }
+
             }
+            model.setType(groupType.stream().collect(Collectors.joining(ItemUtil.EXTENSION_SEPARATOR)));
         } else {
-            model = ItemsFactory.eINSTANCE.createModelNormalItem();
             model.setType(item.getType());
         }
 
--- a/bundles/org.openhab.core.model.item/src/org/openhab/core/model/item/internal/fileconverter/DslItemFileConverter.java
+++ b/bundles/org.openhab.core.model.item/src/org/openhab/core/model/item/internal/fileconverter/DslItemFileConverter.java
@@ -129,7 +129,6 @@ public class DslItemFileConverter extends AbstractItemFileGenerator implements I
                         model.getArgs().add(parameters[i].toString());
                     }
                 }
-
             }
             model.setType(groupType.stream().collect(Collectors.joining(ItemUtil.EXTENSION_SEPARATOR)));
         } else {
--- a/bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
+++ b/bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
@@ -72,8 +72,7 @@ class ItemsValidator extends AbstractItemsValidator {
             case 2: checkGroupBaseType(item, segments.get(1))
             case 3: checkGroupWithOneParam(item, segments.get(1), segments.get(2))
             case 4: checkGroupWithTwoParams(item, segments.get(1), segments.get(2), segments.get(3))
-            // The xtext grammar will not allow more than 4 segments for Group types, so the default case will never be reached
-            default: error("Item '" + item.name + "' has an invalid Group type: too many segments: " + item.type, ItemsPackage.Literals.MODEL_ITEM__TYPE)
+            // The xtext grammar will not allow more than 4 segments for Group types, so no default case is needed
         }
     }
 
--- a/bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
+++ b/bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
@@ -61,7 +61,7 @@ class ItemsValidator extends AbstractItemsValidator {
     }
 
     def checkBasicItemType(ModelItem item) {
-        if (!CoreItemFactory.SUPPORTED_ITEM_TYPES.contains(item.type)) {
+        if (!CoreItemFactory.VALID_ITEM_TYPES.contains(item.type)) {
             error("Item '" + item.name + "' has an invalid type: '" + item.type + "'", ItemsPackage.Literals.MODEL_ITEM__TYPE)
         }
     }
@@ -78,7 +78,7 @@ class ItemsValidator extends AbstractItemsValidator {
     }
 
     def checkGroupBaseType(ModelItem item, String baseType) {
-        if (!CoreItemFactory.SUPPORTED_ITEM_TYPES.contains(baseType)) {
+        if (!CoreItemFactory.VALID_ITEM_TYPES.contains(baseType)) {
             error("Item '" + item.name + "' has an invalid base item type: '" + baseType + "'", ItemsPackage.Literals.MODEL_ITEM__TYPE)
         }
     }
--- a/bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
+++ b/bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
@@ -12,15 +12,18 @@
  */
 package org.openhab.core.model.validation
 
-import org.openhab.core.model.items.ModelItem
 import org.eclipse.xtext.validation.Check
+import org.openhab.core.items.GroupFunction
+import org.openhab.core.items.ItemUtil
+import org.openhab.core.library.CoreItemFactory
 import org.openhab.core.model.items.ItemsPackage
+import org.openhab.core.model.items.ModelItem
 import org.openhab.core.types.util.UnitUtils
 
 /**
- * Custom validation rules. 
+ * Custom validation rules.
  *
- * see http://www.eclipse.org/Xtext/documentation.html#validation
+ * see https://eclipse.dev/Xtext/documentation/303_runtime_concepts.html#validation
  */
 class ItemsValidator extends AbstractItemsValidator {
 
@@ -29,23 +32,96 @@ class ItemsValidator extends AbstractItemsValidator {
 		if (item === null || item.name === null) {
 			return
 		}
-		if (item.name.contains("-")) {
-			error('Item name must not contain dashes.', ItemsPackage.Literals.MODEL_ITEM__NAME)
-		}
+        if (!ItemUtil.isValidItemName(item.name)) {
+			error('Item name "' + item.name + '" is invalid. It must begin with a letter or underscore followed by alphanumeric characters and underscores, and must not contain any other symbols.', ItemsPackage.Literals.MODEL_ITEM__NAME)
+        }
 	}
-	
-	@Check
-	def checkDimension(ModelItem item) {
-	    if (item === null || item.type === null) {
+
+    @Check
+    def checkValidItemType(ModelItem item) {
+        if (item === null || item.type === null) {
             return
         }
-        if (item.type.startsWith("Number:")) {
-            var dimension = item.type.substring(item.type.indexOf(":") + 1)
-            try {
-                UnitUtils.parseDimension(dimension)                
-            } catch (IllegalArgumentException e) {
-                warning("'" + dimension + "' is not a valid dimension.", ItemsPackage.Literals.MODEL_ITEM__TYPE)
-            }
+
+        val segments = item.type.split(":")
+        val mainType = segments.get(0)
+        switch mainType {
+            case "Group": checkGroupType(item, segments)
+            case "Number": checkNumberType(item, segments)
+            default: checkBasicItemType(item)
         }
-	}
+    }
+
+    def checkNumberType(ModelItem item, String[] segments) {
+        switch (segments.size) {
+            case 1: return // Just "Number", valid
+            case 2: checkDimension(item, segments.get(1)) // "Number:<dimension>"
+            default: error("Item: '" + item.name + "' has an invalid Number type, too many segments: " + item.type, ItemsPackage.Literals.MODEL_ITEM__TYPE)
+        }
+    }
+
+    def checkBasicItemType(ModelItem item) {
+        if (!CoreItemFactory.SUPPORTED_ITEM_TYPES.contains(item.type)) {
+            error("Item '" + item.name + "' has an invalid type: '" + item.type + "'", ItemsPackage.Literals.MODEL_ITEM__TYPE)
+        }
+    }
+
+    def checkGroupType(ModelItem item, String[] segments) {
+        switch (segments.size) {
+            case 1: return // Just "Group", valid
+            case 2: checkGroupBaseType(item, segments.get(1))
+            case 3: checkGroupWithOneParam(item, segments.get(1), segments.get(2))
+            case 4: checkGroupWithTwoParams(item, segments.get(1), segments.get(2), segments.get(3))
+            default: error("Item '" + item.name + "' has an invalid Group type: too many segments", ItemsPackage.Literals.MODEL_ITEM__TYPE)
+        }
+    }
+
+    def checkGroupBaseType(ModelItem item, String baseType) {
+        if (!CoreItemFactory.SUPPORTED_ITEM_TYPES.contains(baseType)) {
+            error("Item '" + item.name + "' has an invalid base item type: '" + baseType + "'", ItemsPackage.Literals.MODEL_ITEM__TYPE)
+        }
+    }
+
+    def checkGroupWithOneParam(ModelItem item, String baseType, String param) {
+        checkGroupBaseType(item, baseType)
+
+        if (item.args.size == 0 && baseType == "Number") {
+            if (!isValidGroupFunction(param)) {
+                // Group:Number:Dimension
+                checkDimension(item, param)
+            }
+        } else {
+            checkGroupFunction(item, param)
+        }
+    }
+
+    def checkGroupWithTwoParams(ModelItem item, String baseType, String dimension, String function) {
+        checkGroupBaseType(item, baseType)
+
+        if (baseType == "Number") {
+            checkDimension(item, dimension)
+        } else {
+            error("Item '" + item.name + "' with type '" + item.type + "' cannot have a dimension. Dimensions are only valid for Number base type. The dimension '" + dimension + "' is ignored.", ItemsPackage.Literals.MODEL_ITEM__TYPE)
+        }
+
+        checkGroupFunction(item, function)
+    }
+
+    def checkDimension(ModelItem item, String dimension) {
+        try {
+            UnitUtils.parseDimension(dimension)
+        } catch (IllegalArgumentException e) {
+            error("Item '" + item.name + "' has an unknown dimension: '" + dimension + "'. The Item will not be created.", ItemsPackage.Literals.MODEL_ITEM__TYPE)
+        }
+    }
+
+    def checkGroupFunction(ModelItem item, String function) {
+        if (!isValidGroupFunction(function)) {
+            warning("Item '" + item.name + "' has an unknown group function: '" + function + "'. Using EQUALITY instead.", ItemsPackage.Literals.MODEL_ITEM__TYPE)
+        }
+    }
+
+    def isValidGroupFunction(String value) {
+        return GroupFunction.VALID_FUNCTIONS.contains(value)
+    }
 }
--- a/bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
+++ b/bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
@@ -72,7 +72,8 @@ class ItemsValidator extends AbstractItemsValidator {
             case 2: checkGroupBaseType(item, segments.get(1))
             case 3: checkGroupWithOneParam(item, segments.get(1), segments.get(2))
             case 4: checkGroupWithTwoParams(item, segments.get(1), segments.get(2), segments.get(3))
-            default: error("Item '" + item.name + "' has an invalid Group type: too many segments", ItemsPackage.Literals.MODEL_ITEM__TYPE)
+            // The xtext grammar will not allow more than 4 segments for Group types, so the default case will never be reached
+            default: error("Item '" + item.name + "' has an invalid Group type: too many segments: " + item.type, ItemsPackage.Literals.MODEL_ITEM__TYPE)
         }
     }
 
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/CoreItemFactory.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/CoreItemFactory.java
@@ -12,6 +12,8 @@
  */
 package org.openhab.core.library;
 
+import java.util.Set;
+
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.i18n.UnitProvider;
@@ -57,6 +59,11 @@ public class CoreItemFactory implements ItemFactory {
     public static final String ROLLERSHUTTER = "Rollershutter";
     public static final String STRING = "String";
     public static final String SWITCH = "Switch";
+
+    public static final Set<String> SUPPORTED_ITEM_TYPES = Set.of( //
+            CALL, COLOR, CONTACT, DATETIME, DIMMER, IMAGE, LOCATION, NUMBER, PLAYER, ROLLERSHUTTER, STRING, SWITCH //
+    );
+
     private final UnitProvider unitProvider;
 
     @Activate
@@ -90,7 +97,6 @@ public class CoreItemFactory implements ItemFactory {
 
     @Override
     public String[] getSupportedItemTypes() {
-        return new String[] { SWITCH, ROLLERSHUTTER, CONTACT, STRING, NUMBER, DIMMER, DATETIME, COLOR, IMAGE, PLAYER,
-                LOCATION, CALL };
+        return SUPPORTED_ITEM_TYPES.toArray(new String[0]);
     }
 }
--- a/bundles/org.openhab.core/src/main/java/org/openhab/core/library/CoreItemFactory.java
+++ b/bundles/org.openhab.core/src/main/java/org/openhab/core/library/CoreItemFactory.java
@@ -60,7 +60,7 @@ public class CoreItemFactory implements ItemFactory {
     public static final String STRING = "String";
     public static final String SWITCH = "Switch";
 
-    public static final Set<String> SUPPORTED_ITEM_TYPES = Set.of( //
+    public static final Set<String> VALID_ITEM_TYPES = Set.of( //
             CALL, COLOR, CONTACT, DATETIME, DIMMER, IMAGE, LOCATION, NUMBER, PLAYER, ROLLERSHUTTER, STRING, SWITCH //
     );
 
@@ -97,6 +97,6 @@ public class CoreItemFactory implements ItemFactory {
 
     @Override
     public String[] getSupportedItemTypes() {
-        return SUPPORTED_ITEM_TYPES.toArray(new String[0]);
+        return VALID_ITEM_TYPES.toArray(new String[0]);
     }
 }
--- a/bundles/org.openhab.core.model.core/src/main/java/org/openhab/core/model/core/internal/ModelRepositoryImpl.java
+++ b/bundles/org.openhab.core.model.core/src/main/java/org/openhab/core/model/core/internal/ModelRepositoryImpl.java
@@ -330,7 +330,17 @@ public class ModelRepositoryImpl implements ModelRepository {
                     final org.eclipse.emf.common.util.Diagnostic diagnostic = safeEmf
                             .call(() -> Diagnostician.INSTANCE.validate(resource.getContents().getFirst()));
                     for (org.eclipse.emf.common.util.Diagnostic d : diagnostic.getChildren()) {
-                        warnings.add(d.getMessage());
+                        if (d.getSeverity() == org.eclipse.emf.common.util.Diagnostic.ERROR) {
+                            criticalErrors.append(d.getMessage() + "\n");
+                        } else {
+                            warnings.add(d.getMessage());
+                        }
+                    }
+                    if (!criticalErrors.isEmpty()) {
+                        return criticalErrors.toString();
+                    } else if (!warnings.isEmpty()) {
+                        logger.info("Validation issues found in configuration model '{}', using it anyway:\n{}", name,
+                                String.join("\n", warnings));
                     }
                 } catch (NullPointerException e) {
                     // see https://github.com/eclipse/smarthome/issues/3335
--- a/bundles/org.openhab.core.model.core/src/main/java/org/openhab/core/model/core/internal/ModelRepositoryImpl.java
+++ b/bundles/org.openhab.core.model.core/src/main/java/org/openhab/core/model/core/internal/ModelRepositoryImpl.java
@@ -331,16 +331,13 @@ public class ModelRepositoryImpl implements ModelRepository {
                             .call(() -> Diagnostician.INSTANCE.validate(resource.getContents().getFirst()));
                     for (org.eclipse.emf.common.util.Diagnostic d : diagnostic.getChildren()) {
                         if (d.getSeverity() == org.eclipse.emf.common.util.Diagnostic.ERROR) {
-                            criticalErrors.append(d.getMessage() + "\n");
+                            errors.add(d.getMessage());
                         } else {
                             warnings.add(d.getMessage());
                         }
                     }
-                    if (!criticalErrors.isEmpty()) {
-                        return criticalErrors.toString();
-                    } else if (!warnings.isEmpty()) {
-                        logger.info("Validation issues found in configuration model '{}', using it anyway:\n{}", name,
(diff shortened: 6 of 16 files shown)


andrewfg:
@jimtng many thanks for picking this up. It seems that the tests, although running, are now failing, but you are obviously on the right track. I have no knowledge of xtext so can’t really help any more with the details..

jimtng:
This passed now.  However I've updated the original post with some further explanation

andrewfg:
@jimtng I think the ^ prefix is a horrible hack, and IMHO not a solution. The parser should break on new lines, and if the word Switch is at the start of the line it should be interpreted as an item type; and if anywhere else it should be interpreted as a string (label or tag name). I think a simple regex could easily identify the two cases.

bundles/org.openhab.core.model.item/src/org/openhab/core/model/Items.xtext
```diff
@@ -17,9 +17,10 @@ ModelItem:
 	(ModelNormalItem | ModelGroupItem) name=ID
 	(label=STRING)?
 	('<' icon=Icon '>')?
-	('(' groups+=ID (',' groups+=ID)* ')')? 
-	('[' tags+=(ID|STRING) (',' tags+=(ID|STRING))* ']')?
-	('{' bindings+=ModelBinding (',' bindings+=ModelBinding)* '}')? 
+	('(' groups+=ID (',' groups+=ID)* ')')?
+	// Add BaseModelItemType as valid tags otherwise the parser treats [Switch] as start of item definition
+	('[' tags+=(ID|STRING|BaseModelItemType) (',' tags+=(ID|STRING|BaseModelItemType))* ']')?
+	('{' bindings+=ModelBinding (',' bindings+=ModelBinding)* '}')?
```
lolodomo, line 23: Can you check what was changed at lines 20 and 22 ?
Did you replace spaces by tab or vice-versa ?

lolodomo:
Note that if this PR is not merged, the code generating DSL format for item has to be fixed to add the character `^ when expected when generating tags.
https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.model.item/src/org/openhab/core/model/item/internal/fileconverter/DslItemFileConverter.java#L146

andrewfg:
^ and we would probably need for safety sake to fix all item getTag(s) and setTag(s) methods to eliminate any stray carets. As I say, I think it is a horrible hack, and I strongly argue to solve this properly. We all know the proper syntax since years, and we should not allow a compromise in order to avoid a fix.

bundles/org.openhab.core.model.item/src/org/openhab/core/model/Items.xtext
andrewfg: Maybe a CRLF issue?

andrewfg:
If the current parser really cannot distinguish between Switch at the start of a line and Switch anywhere else, the we should implement a regex pre-parser to do that in an inter- phase between loading the file and passing the stream to the existing parser.

lolodomo:
But when the user writes ^Switch, I guess that the XText parser automatically removes the ^. So the resulting tag is Switch as we expect.

andrewfg:
Maybe. Who knows. My argument is to make the code accept the syntax that every user knows since OH v1.0 and not introduce some queer kludge to avoid fixing a bug in the code that will confuse everybody.

lolodomo:
Apparently noone was annoyed by this "bug" since several years ;)

I just checked how I myself solved this problem in my own file. I did not use ^ character at all but all my tags are between double quotes like for example: `[ "Switch" ]`

IMHO, as string are accepted in addition to ID by the parser, it is fine and there is nothing to fix. If the user wants to use a tag that is a special keyword, he just has to surround it with double-quotes.

lolodomo:
But @jimtng proposal is also an improvement as it allows a syntax that was not possible before, so why not.
I see that more as an enhancement than as a bug fix.

andrewfg:
Can either of you explain why the parser is not capable to detect the difference between `Switch` at the beginning of a line and `Switch` elsewhere? It cannot be so difficult.

lolodomo:
I also checked for an item what is the generated code from Main UI and everything is OK, this is not a string that is used but the escape character:
```
Switch bypassed "ignorée" <switch> [^Switch] { channel="xxxxxx" }
```
So nothing to change/fix in the DSL generator.

andrewfg:
So Copilot suggests to create and use a custom token source that injects a synthetic token like `terminal LINE_START: 'synthetic:LINE_START'` at the start of each line. Then the parser can differentiate between those tokens that start with that synthetic token and  those that do not.

```
import org.antlr.runtime.Token;
import org.antlr.runtime.TokenSource;
import org.antlr.runtime.CommonToken;
import org.eclipse.xtext.parser.antlr.XtextTokenStream;
import org.eclipse.xtext.parser.antlr.ITokenDefProvider;

public class LineAwareTokenSource extends XtextTokenStream {

    private boolean atLineStart = true;

    public LineAwareTokenSource(TokenSource tokenSource, ITokenDefProvider tokenDefProvider) {
        super(tokenSource, tokenDefProvider);
    }

    @Override
    public Token nextToken() {
        Token next = super.nextToken();

        // Check for newline to reset line start flag
        if (next.getType() == getTokenType("WS") && next.getText().contains("\n")) {
            atLineStart = true;
        }

        // Inject LINE_START before first visible token on a new line
        if (atLineStart && isVisibleToken(next)) {
            atLineStart = false;
            CommonToken synthetic = new CommonToken(getTokenType("LINE_START"), "synthetic:LINE_START");
            synthetic.setLine(next.getLine());
            synthetic.setCharPositionInLine(0);
            return synthetic;
        }

        return next;
    }

    private boolean isVisibleToken(Token token) {
        int type = token.getType();
        return type != getTokenType("WS") &&
               type != getTokenType("SL_COMMENT") &&
               type != getTokenType("ML_COMMENT");
    }

    private int getTokenType(String name) {
        return getTokenDefProvider().getTokenDefMap().entrySet().stream()
            .filter(e -> name.equals(e.getValue()))
            .map(e -> e.getKey())
            .findFirst()
            .orElse(-1);
    }
}
```

lolodomo:
I would suggest to not start a big change with risk while this is something that everyone using .items already knows how to handle. Maybe something could be added in our documentation to help someone who could encounter that error for the first time.

andrewfg:
> this is something that everyone using .items already knows how to handle

I don't agree.

jimtng:
For tags, including generating DSL syntax, I'd suggest just using quoted strings instead of caret or bare.

I tried finding a syntax or solution but haven't found one that's simple/straight forward.

jimtng:
> I also checked for an item what is the generated code from Main UI and everything is OK, this is not a string that is used but the escape character:
> 
> ```
> Switch bypassed "ignorée" <switch> [^Switch] { channel="xxxxxx" }
> ```
> 
> So nothing to change/fix in the DSL generator.

indeed, that's good to know!

jimtng:
I tried creating an item in the UI:
- Name: Switch
- Tag: Switch

That worked and generated the correct DSL: `String ^Switch "Switch" [^Switch]` -> DSL Parser is happy to load it

Then I set the icon to `Switch`, DSL generated this: `String ^Switch "Switch" <Switch> [^Switch]` -> DSL Parser is not happy. 
It needs to be manually adjusted to `String ^Switch1 "Switch" <^Switch> [^Switch]` to make DSL parser happy

jimtng:
Marking this PR as Draft. I'm not happy with the current solution.

lolodomo:
The icon is generally in lowercase. Does it work with switch as icon ?

`String ^Switch "Switch" <switch> [^Switch]`

bundles/org.openhab.core.model.item/src/org/openhab/core/model/Items.xtext
jimtng: They had a trailing space, my editor is set to automatically remove them on save.

jimtng:
> The icon is generally in lowercase. Does it work with switch as icon ?
> 
> `String ^Switch "Switch" <switch> [^Switch]`

Yes it works with lowercase <switch>. However, the UI accepts `Switch` and shows a switch icon -> there must be some lowercasing happening somewhere.

Even if we ignore this, basically this problem occurs everywhere, and not just Switch, e.g.

```
String ^Switch "Switch" <switch> [^Switch] { Switch=test }
```

We need to make the literal parsing context sensitive, ie only beginning of line, and in the case of the other literals, only where they're needed. I haven't figured this out yet. I don't want to have to add custom java pre-parser which still won't fully solve the issue anyway.

rkoshak:
> We need to make the literal parsing context sensitive, ie only beginning of line
There is no requirement that an Item is all on one line. The following is a legal Item definition:

```
Switch MySwitch
      [ Tag1,
       Tag2]
```

If you replace "Tag2" with "Switch", it will be the first word on the line. A newline is not sufficient to tell when Switch is being used to start the definition of an Item or being used as a tag. Changing this would be a breaking change as many users define their Items across multiple lines. 

I would have expected the parser to treat everything between [ and ] as tags though. The parser doesn't keep state and know where it is in the Item definition?

jimtng:
> ```
> Switch MySwitch
>       [ Tag1,
>        Tag2]
> ```
> 
> 
>     
>   
> 
> If you replace "Tag2" with "Switch", it will be the first word on the line. A newline is not sufficient to tell when Switch is being used to start the definition of an Item or being used as a tag

Ahh yes thanks! That saved me from going down the wrong rabbit hole.

jimtng:
I've made some progress, although haven't ironed out all the issues yet. This is now working as expected:

```
Switch Switch "Switch" <Switch> [Switch] { Switch="Switch" }
```

This also works (exactly the same but broken into multiple lines):
```
Switch
Switch
"Switch"
<Switch>
[
Switch
] {
    Switch="Switch" 
}
```

jimtng:
Note, this PR may also address #4903 for DSL. Will fix YAML in a separate pr

jimtng:
I've made progress and ironed out most of the issues. The only remaining issue is with the `Group` literal. 

```
Number Number // this is fine
Switch Switch [Switch] // also fine
Group MyGroup // this is fine as before
Group Group // ERROR! the literal 'Group' is still being parsed incorrectly here
Switch Item1 [Group] // ERROR! This is also suffering from the same issue
```

This would do for now, I think.

bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
```diff
@@ -29,23 +36,42 @@ class ItemsValidator extends AbstractItemsValidator {
...
+    @Check
+    def checkValidItemType(ModelNormalItem item) {
+        if (!CoreItemFactory.SUPPORTED_ITEM_TYPES.contains(item.type) && !item.type.startsWith("Number:")) {
+            error("Invalid item type: " + item.type, ItemsPackage.Literals.MODEL_ITEM__TYPE)
+        }
+    }
+
+    @Check
+    def checkValidGroupFunction(ModelGroupItem item) {
+        if (item.function !==  null && !GroupFunction.VALID_FUNCTIONS.contains(item.function)) {
```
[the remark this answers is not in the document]
jimtng: fixed

bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
```diff
@@ -29,23 +36,42 @@ class ItemsValidator extends AbstractItemsValidator {
...
-	
+
 	@Check
 	def checkDimension(ModelItem item) {
 	    if (item === null || item.type === null) {
             return
         }
         if (item.type.startsWith("Number:")) {
             var dimension = item.type.substring(item.type.indexOf(":") + 1)
+            if (item instanceof ModelGroupItem && GroupFunction.VALID_FUNCTIONS.contains(dimension)) {
```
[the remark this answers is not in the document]
jimtng: not quite a duplicate, and `GroupFunction.VALID_FUNCTIONS` _is_ the "shared utility method" that avoids code duplication already.

andrewfg:
@jimtng this looks mega! If I understand correctly you shifted the item type validation from the front-end xtext input stream parser to the java back-end. This makes huge sense. And .. dare I say it .. with the benefit of hindsight .. so obvious that I am surprised nobody thought of it before. It takes a great mind.

jimtng:
> I am surprised nobody thought of it before

it came out of necessity!

To resolve the `Group` issue, I'd have to shift more of the syntax parsing out of xtext. Not sure if it's worth it, but it wouldn't be too difficult either.

The good thing is that now there's only one list of item types and function names, not duplicated everywhere.

jimtng:
After over 2 days trying to figure out how to solve the Group parsing properly, I've decided to take the `Group` literal out of xtext grammar too. It simplifies a lot of things.

The bonus is that errors in item type definition will no longer cause the entire .items file to be discarded. The file will still be processed and other valid item definitions will load.

This is a good thing, I think. However, it may cause a surprise, i.e. items previously not loaded will now be loaded.

Another case:

```
Group:Switch:INVALIDFUNCTION TestItem
```

Previously this will cause the entire .items file to be discarded. Now, the file will be processed, and a group TestItem will be created with EQUALITY function - this is the [default behaviour](https://github.com/openhab/openhab-core/blob/6f7d46caab56b88ce14e9036e47f868fa10dd6c1/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/GroupFunctionHelper.java#L170) which had already been implemented prior to this PR.

I've also improved on the error messages when parsing .items file. Instead of the cryptic:

```
Configuration model 'test.items' has errors, therefore ignoring it: [2,20]: no viable alternative at input 'sss'
[2,24]: missing EOF at 'G1'
```

```
Validation issues found in configuration model 'test.items', using it anyway:
Item 'G1' has an unknown group function: 'sss'. <-- previously, parsing stops at the first error
Item 'G3' has an invalid dimension: 'a' <-- and these two wouldn't even be parsed
Item 'G3' has an unknown group function: 'bOR'.

later on when the item is created...

11:38:23.133 [ERROR] [re.internal.items.GroupFunctionHelper] - Unknown group function 'sss'. Using Equality instead.
11:38:23.144 [ERROR] [re.internal.items.GroupFunctionHelper] - Unknown group function 'sss'. Using Equality instead.
```

bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
```diff
@@ -29,23 +32,96 @@ class ItemsValidator extends AbstractItemsValidator {
 		if (item === null || item.name === null) {
 			return
 		}
-		if (item.name.contains("-")) {
-			error('Item name must not contain dashes.', ItemsPackage.Literals.MODEL_ITEM__NAME)
-		}
+        if (!ItemUtil.isValidItemName(item.name)) {
+			error('Item name "' + item.name + '" is invalid. It must begin with a letter or underscore followed by alphanumeric characters and underscores, and must not contain any other symbols.', ItemsPackage.Literals.MODEL_ITEM__NAME)
```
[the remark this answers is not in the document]
jimtng: the rule hasn't changed

jimtng:
@lolodomo would you mind taking a look at this please?

lolodomo:
I will during the coming weekend.

bundles/org.openhab.core.model.item/src/org/openhab/core/model/formatting/ItemsFormatter.xtend
```diff
@@ -25,8 +25,10 @@ class ItemsFormatter extends AbstractDeclarativeFormatter {
 	@Inject extension ItemsGrammarAccess
 
 	override protected void configureFormatting(FormattingConfig c) {
-		c.setLinewrap(1, 1, 2).before(modelGroupItemRule)
-		c.setLinewrap(1, 1, 2).before(modelNormalItemRule)
+		c.setLinewrap(1, 1, 2).before(modelItemRule)
+
+		// No space between the item type and the opening parenthesis for the group function arguments
+		c.setNoSpace().between(modelItemTypeRule, modelItemAccess.leftParenthesisKeyword_1_0)
```
lolodomo, line 31: No risk of confusion with the left parenthesis for groups ?

bundles/org.openhab.core.model.item/src/org/openhab/core/model/item/internal/fileconverter/DslItemFileConverter.java
```diff
@@ -95,26 +94,25 @@ public synchronized void generateFileFormat(OutputStream out, List<Item> items,
...
-                            .getByName(function.getClass().getSimpleName().toUpperCase());
-                    modelGroup.setFunction(modelFunction);
+                    groupType.add(function.getClass().getSimpleName().toUpperCase());
                     State[] parameters = function.getParameters();
                     for (int i = 0; i < parameters.length; i++) {
-                        modelGroup.getArgs().add(parameters[i].toString());
+                        model.getArgs().add(parameters[i].toString());
                     }
                 }
+
```
lolodomo: Minor: added blank line to be removed

bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
```diff
@@ -29,23 +32,96 @@ class ItemsValidator extends AbstractItemsValidator {
...
+        }
+    }
+
+    def checkGroupType(ModelItem item, String[] segments) {
+        switch (segments.size) {
+            case 1: return // Just "Group", valid
+            case 2: checkGroupBaseType(item, segments.get(1))
+            case 3: checkGroupWithOneParam(item, segments.get(1), segments.get(2))
+            case 4: checkGroupWithTwoParams(item, segments.get(1), segments.get(2), segments.get(3))
+            default: error("Item '" + item.name + "' has an invalid Group type: too many segments", ItemsPackage.Literals.MODEL_ITEM__TYPE)
```
lolodomo: You could log item.type like you do in `checkNumberType` (for consistency)

bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
```diff
@@ -29,23 +32,96 @@ class ItemsValidator extends AbstractItemsValidator {
...
+    def checkGroupBaseType(ModelItem item, String baseType) {
+        if (!CoreItemFactory.SUPPORTED_ITEM_TYPES.contains(baseType)) {
+            error("Item '" + item.name + "' has an invalid base item type: '" + baseType + "'", ItemsPackage.Literals.MODEL_ITEM__TYPE)
+        }
+    }
+
+    def checkGroupWithOneParam(ModelItem item, String baseType, String param) {
+        checkGroupBaseType(item, baseType)
+
+        if (item.args.size == 0 && baseType == "Number") {
```
lolodomo, line 88: Just for checking: is it == or === ?

bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
```diff
@@ -29,23 +32,96 @@ class ItemsValidator extends AbstractItemsValidator {
...
+        } else {
+            checkGroupFunction(item, param)
         }
-	}
+    }
+
+    def checkGroupWithTwoParams(ModelItem item, String baseType, String dimension, String function) {
+        checkGroupBaseType(item, baseType)
+
+        if (baseType == "Number") {
```
lolodomo, line 101: Just for checking: is it == or === ?

bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
```diff
@@ -29,23 +32,96 @@ class ItemsValidator extends AbstractItemsValidator {
...
+            checkDimension(item, dimension)
+        } else {
+            error("Item '" + item.name + "' with type '" + item.type + "' cannot have a dimension. Dimensions are only valid for Number base type. The dimension '" + dimension + "' is ignored.", ItemsPackage.Literals.MODEL_ITEM__TYPE)
```
lolodomo, line 104: If it is ignored, is it an error or a warning ?

bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
jimtng:
```
Equality Operators
In Xtend the equals operators (==,!=) are bound to Object.equals. So you can write:

if (name == 'Homer')
	println('Hi Homer')
Java’s identity equals semantic is mapped to the tripple-equals operators === and !== in Xtend.

if (someObject === anotherObject)
	println('same objects')
```
https://eclipse.dev/Xtext/xtend/documentation/203_xtend_expressions.html

bundles/org.openhab.core/src/main/java/org/openhab/core/library/CoreItemFactory.java
```diff
@@ -57,6 +59,11 @@ public class CoreItemFactory implements ItemFactory {
     public static final String ROLLERSHUTTER = "Rollershutter";
     public static final String STRING = "String";
     public static final String SWITCH = "Switch";
+
+    public static final Set<String> SUPPORTED_ITEM_TYPES = Set.of( //
```
lolodomo: Now that you introduce this set, it could certainly also use in the YAML parser when I check the item type is valid in item and channel.

bundles/org.openhab.core.model.item/src/org/openhab/core/model/formatting/ItemsFormatter.xtend
jimtng: AFAIK, no. This is between the item type and the parenthesis, e.g.
```
Group:Switch:OR ________ (ON,OFF)     ItemName     (GroupThatItBelongsTo)
                 ^^^^
```

So unless I'm mistaken, it shouldn't be confused with the open parenthesis for the group, although I don't remember whether I tested with the group.

lolodomo:
My review is finished.

bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
jimtng: IMO, it should be an error that needs to be corrected.

bundles/org.openhab.core.model.item/src/org/openhab/core/model/formatting/ItemsFormatter.xtend
lolodomo: What I mean is that if no parameters are set for the function, wouldn't the parenthesis for groups be considered by your rule ? Leading to some spaces being removed when not expected.

bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
lolodomo: So it would be ok, comment to be ignored

bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
lolodomo: So it would be ok, comment to be ignored

bundles/org.openhab.core/src/main/java/org/openhab/core/library/CoreItemFactory.java
lolodomo: I mean this constant

](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/internal/util/YamlElementUtils.java#L34)

and as a consequence this method

https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.model.yaml/src/main/java/org/openhab/core/model/yaml/internal/util/YamlElementUtils.java#L74

bundles/org.openhab.core.model.item/src/org/openhab/core/model/formatting/ItemsFormatter.xtend
lolodomo: Like for example
`Group:Switch:EQUALITY     ItemName     (GroupThatItBelongsTo)`

lolodomo: Maybe it is fine, I don't know exactly what is `modelItemAccess.leftParenthesisKeyword_1_0`

jimtng: I will test this case and report back

bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
lolodomo: So you should not mention that the dimension will be ignored as in fact this item will be fully rejected.
It is just the message to adjust.

lolodomo:
Regarding #4903 , this PR fixed the item name checking for DSL but not for YAML, correct ?

lolodomo:
I imagine you have done all tests for your added checks in class `ItemsValidator` ?

bundles/org.openhab.core.model.item/src/org/openhab/core/model/formatting/ItemsFormatter.xtend
jimtng: I can confirm that this works fine with the Group membership parentheses.

```
Group:Switch:EQUALITY ItemName (G1)
```


I noticed that there's a space between the two options in the comma. I didn't notice this before but it does exist in the current `main` branch, e.g.

```
Group:Switch:OR(ON, OFF) G2 (G1)
                   ^
```

It does get loaded properly, so I'll just leave that alone for now and it's not the topic of this PR.

jimtng: To clarify, in the main branch the output is actually with space before the left parenthesis (because that's fixed in this PR)

```
Group:Switch:OR (ON, OFF) G2 (G1)
               ^    ^ The main branch has both these spaces
```

lolodomo: Fine for your removal of the space before the left parenthesis, it is something I previously saw but did not take time to change (or maybe I did not know how to change it simply).

bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
jimtng: The item is not fully rejected. The dimension is indeed ignored.

Given this definition in an .items file:
```
Group:Switch:Power:OR(ON,OFF) G1 "Group"
```

The user will see this in the log:
```
20:46:42.357 [INFO ] [del.core.internal.ModelRepositoryImpl] - Loading model 'test.items'
20:46:42.367 [INFO ] [del.core.internal.ModelRepositoryImpl] - Validation issues found in configuration model 'test.items', using it anyway:
Item 'G1' with type 'Group:Switch:Power:OR' cannot have a dimension. Dimensions are only valid for Number base type. The dimension 'Power' is ignored.
```

And the Item G1 is still registered but with the dimensions ignored as logged.

AFAIK, the items validator has no ability to reject an item. All it can do is log messages.

Even with the `error` function call, it still ends up as an `INFO` log.

Whereas currently, this will cause a syntax error because it violates the xtend grammar, so we get:
```
20:50:27.487 [WARN ] [del.core.internal.ModelRepositoryImpl] - Configuration model 'test.items' has errors, therefore ignoring it: [50,18]: missing RULE_ID at ')'
[52,14]: no viable alternative at input 'Power'
[52,19]: missing EOF at ':'
```

and when this happens, the _entire_ file is rejected, and none of the other items will be loaded at all.

bundles/org.openhab.core/src/main/java/org/openhab/core/library/CoreItemFactory.java
jimtng: I think it's a good idea to make the change in this PR so that it is all contained in one commit. I'll make the changes shortly.

bundles/org.openhab.core.model.item/src/org/openhab/core/model/item/internal/fileconverter/DslItemFileConverter.java
jimtng: blank line removed

bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
jimtng: added item.type for consistency, however I discovered that I can never trigger the default case. The xtext grammar prevents it from having more than 4 segments. I added a comment - should I just remove the default case since it's currently redundant?

bundles/org.openhab.core/src/main/java/org/openhab/core/library/CoreItemFactory.java
jimtng: changes to Yaml made.

jimtng:
> Regarding #4903 , this PR fixed the item name checking for DSL but not for YAML, correct ?

Correct. This is what happens when the name pattern is invalid:
- It gets noted by the ItemsValidator
- AbstractRegistry will issue a warning and reject it.
```
21:16:25.902 [INFO ] [del.core.internal.ModelRepositoryImpl] - Validation issues found in configuration model 'test.items', using it anyway:
Item name "1item" is invalid. It must begin with a letter or underscore followed by alphanumeric characters and underscores, and must not contain any other symbols.
21:16:25.903 [WARN ] [core.common.registry.AbstractRegistry] - Cannot add "SwitchItem" with key "1item": The specified name of the item '1item' is not valid!
```

jimtng:
> I imagine you have done all tests for your added checks in class `ItemsValidator` ?

I added some extra tests in GenericItemProviderTest.java - included in this PR

bundles/org.openhab.core/src/main/java/org/openhab/core/library/CoreItemFactory.java
lolodomo: OK for me

bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
lolodomo: Yes, I think you can remove it.

jimtng: I've removed it. There's also a similar case in GenericItemProvider.java, but I think that one should remain just in case, wdyt?

bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
lolodomo: In validateModel, we log all validation messages. But that is not yet clear for me, what is the difference if you call error method or warn method in a check ?

https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.model.core/src/main/java/org/openhab/core/model/core/internal/ModelRepositoryImpl.java#L299

lolodomo:
> Correct. This is what happens when the name pattern is invalid:
> 
>     * It gets noted by the ItemsValidator
> 
>     * AbstractRegistry will issue a warning and reject it.

Ok, this is because all these checks are finally considered as warnings and not errors, so the loading of model is not stopped.
Fine.

The only thing that remains fully obscure to me is what is the difference if you call warn vs error in a check in ItemsValidator .

bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
jimtng: I guess it just grabs the message, and ignoring the severity.

https://github.com/eclipse-emf/org.eclipse.emf/blob/master/plugins/org.eclipse.emf.common/src/org/eclipse/emf/common/util/Diagnostic.java

jimtng:
> The only thing that remains fully obscure to me is what is the difference if you call warn vs error in a check in ItemsValidator .

See my comment above https://github.com/openhab/openhab-core/pull/4928#discussion_r2295880670

bundles/org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
lolodomo: So using error or warn makes no difference with our current code, correct ?

jimtng: Yes, that's my brief understanding.

lolodomo:
So, I understand now that you changed the item provider to accept invalid item names and invalid item types (in particular), something is logged to announce the problem but the invalid item is finally provided to the registry. I don't think it is a good approach as it is of the responsibility of the provider to provide something correct.

Maybe we have to consider the severity of the validations and reject if severity is error.

jimtng:
> Maybe we have to consider the severity of the validations and reject if severity is error.

Good idea. Let me try that out.

jimtng:
I've made some changes. Here's what will happen now:

```
String Item1
String 1Item // invalid item name
```

will generate this warning, and the entire file isn't getting loaded (the same behaviour as `main` when there's an xtend syntax error)

```
22:55:39.591 [WARN ] [del.core.internal.ModelRepositoryImpl] - Configuration model 'test.items' has errors, therefore ignoring it: Item name "1item" is invalid. It must begin with a letter or underscore followed by alphanumeric characters and underscores, and must not contain any other symbols.
```

So Item1 isn't loaded even though it has the correct syntax.

Are you happy with this behavior?

jimtng:
Since this change is made in ModelRepositoryImpl, I think it would also affect other Xtext stuff like sitemap, Things.

lolodomo:
> will generate this warning, and the entire file isn't getting loaded (the same behaviour as `main` when there's an xtend syntax error)
> 
> ```
> 22:55:39.591 [WARN ] [del.core.internal.ModelRepositoryImpl] - Configuration model 'test.items' has errors, therefore ignoring it: Item name "1item" is invalid. It must begin with a letter or underscore followed by alphanumeric characters and underscores, and must not contain any other symbols.
> ```
> 
> So Item1 isn't loaded even though it has the correct syntax.
> 
> Are you happy with this behavior?

It is not ideal but it is the behaviour we have since the first day so at least we are consistent with the past.

> Since this change is made in ModelRepositoryImpl, I think it would also affect other Xtext stuff like sitemap, Things.

Yes, we have to check if error or warning was used for other models. I believe we don't have too many checks of this kind.

lolodomo:
If I search for `@Check`, I can find commented checks for rules, scripts and persistence using warning().
You have error() being used for item, thing and sitemap.
For thing, it is only one check that checks the id so maybe it is fine, to be checked.
For sitemap, I imagine most of checks should be change to warning() but to be checked.
For item, that is the file you changed. To be checked when error() and when warning() should be used.

So really not a lot of changes to do.

jimtng:
Thing seems to have this weird allowance for syntax like this:

```
Bridge hue:bridge:myBridge @ "basement" [ ip = "1.2.3.4", username = "123" ]
LCT001 bulb1 (hue:bridge:myBridge) [ lightId = "1" ] { Switch : notification }
```

This is allowed, and there is a test that specifically tests it. However the validator currently flags it as an error, hence the test failure.

I'll change it to warn so it will keep its old behaviour, but if this syntax is officially supported, why issue any warnings / logs at all?

jimtng:
The change in ThingValidator from error to warning worked and satisfied the existing Thing test.

jimtng:
> For sitemap, I imagine most of checks should be change to warning() but to be checked.

I had a look through the sitemap's validation. IMO they should all be errors instead of warnings, because unlike in Things - the validation issues will cause the relevant widgets non-functional.

It would give those who have been ignoring the errors a bit of a surprise because now their sitemap won't load at all. I do think the user should fix them, but where do we want to stand on this: keep it lax and change to warning, despite the problems? My vote is to keep them as errors.

andrewfg:
In the past you got a log message that something had failed "somewhere" in a 1000 line file. Which was very hard to fix. So not failing on errors was a good "solution". However if the log message is now clear about the exact nature of the error and its line number, then I think it is better to fail so the user can fix it.

lolodomo:
I will have a look to the checks done for thing and sitemap.
For item, it looks ok to use error() as these cases previously failed in current code and it is logic to fail in case of invalid item name or invalid item type or invalid group function. I just have to check again the new check where you used warning(), I already do not remember what it is.

lolodomo:
If the wrong value will not lead to problem later but will just be ignored or replace by a default value, we can have a warning and continue.
If not, we should generate an error and fail the loading of the file.

The granularity for DSL is the file and not the entity like in YAML, I doubt we will change that soon.

lolodomo:
> For item, it looks ok to use error() as these cases previously failed in current code and it is logic to fail in case of invalid item name or invalid item type or invalid group function. I just have to check again the new check where you used warning(), I already do not remember what it is.

You used warning() for an invalid group function. Did you check that our provider code is then just considering EQUALITY ?

jimtng:
> You used warning() for an invalid group function. Did you check that our provider code is then just considering EQUALITY ?

Yes

https://github.com/openhab/openhab-core/blob/8537a1b82278159d7ec69c1bb45508ca2c9537a1/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/items/GroupFunctionHelper.java#L169-L172

In fact, this gets logged _twice_, and I haven't tried to dig deeper to figure out why or how to make it only log once.

jimtng:
> If the wrong value will not lead to problem later but will just be ignored or replace by a default value, we can have a warning and continue. If not, we should generate an error and fail the loading of the file.
> 
> The granularity for DSL is the file and not the entity like in YAML, I doubt we will change that soon.

We could just change all error() to warning() for Thing and Sitemap. That will keep the same behaviour as before.

lolodomo:
I had a look to sitemap checks. We have to take care in particular with the check of frame in frame, I believe it is supported by some of the sitemap UIs. A warning() could then be more appropriate for this one. To be investigated. I believe the other are really errors.

I believe I discovered a little bug in `checkInterpolationParameter`:  `SitemapPackage.Literals.INPUT` is used while this check is for chart and not input widget. To be fixed in a separate PR.
What is interesting in that check is the way to retrieve the line number. We should probably apply it to our other checks starting with the new checks you added for item. Can you please have a look if you can use `NodeModelUtils.getNode` and `node.getStartLine()`:
```
            val node = NodeModelUtils.getNode(i)
            val line = node.getStartLine()
```

https://github.com/openhab/openhab-core/blob/8537a1b82278159d7ec69c1bb45508ca2c9537a1/bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/validation/SitemapValidator.xtend#L168

lolodomo:
> We could just change all error() to warning() for Thing and Sitemap. That will keep the same behaviour as before.

Yes of course that would be the easier solution that change nothing compared to today.
But at least for sitemaps I would like to really do the right choice and my current feeling is that( errors would be more appropriate than warnings; except maybe for one case. But more investigation is probably necessary.

jimtng:
> We should probably apply it to our other checks starting with the new checks you added for item. Can you please have a look if you can use `NodeModelUtils.getNode` and `node.getStartLine()`:


This is a great find. 

If I were to add the line number to the error messages, I would make it formatted so they are clickable in VS Code that would take me straight to the file / line, so that means revamping even the existing line numbers in sitemap validator too.

However, I think this should be added in a separate PR that goes through all the validators. Adding them here would extend the scope of this PR too wide, I think.



> I had a look to sitemap checks. We have to take care in particular with the check of frame in frame, I believe it is supported by some of the sitemap UIs. A warning() could then be more appropriate for this one. To be investigated. I believe the other are really errors.

I will change Frame in Frame to `warning` here.

lolodomo:
> We could just change all error() to warning() for Thing and Sitemap. That will keep the same behaviour as before.

As your PR is not about thing an sitemap, I believe it is the proper way in your PR to replace error() by warning(), just to not change the current behaviour.
If I want to change sitemap behaviour, I will create a next PR that restore error() where I think it is necessary.
So I am in favour you replace everywhere in thing (already done) and in sitemap.

jimtng:
> As your PR is not about thing an sitemap, I believe it is the proper way in your PR to replace error() by warning(), just to not change the current behaviour.
> If I want to change sitemap behaviour, I will create a next PR that restore error() where I think it is necessary.
> So I am in favour you replace everywhere in thing (already done) and in sitemap.

I agree. Will push another commit to this effect.

lolodomo:
> This is a great find.
> 
> If I were to add the line number to the error messages, I would make it formatted so they are clickable in VS Code that would take me straight to the file / line, so that means revamping even the existing line numbers in sitemap validator too.
> 
> However, I think this should be added in a separate PR that goes through all the validators. Adding them here would extend the scope of this PR too wide, I think.

Except if you have no time or you think it could be complicated, it would be great if you do it for item in that PR as you revisited all validation checks for item.
For thing and sitemap, ok to plan a separate PR.
WDYT?

lolodomo:
If not done, please add an explanation in commit and PR initial message to explain the change done with DSL model validation checks .

jimtng:
I've included Copilot's summary in the original PR message. I think it does a good job at summarizing it.



> Except if you have no time or you think it could be complicated, it would be great if you do it for item in that PR as you revisited all validation checks for item.
> For thing and sitemap, ok to plan a separate PR.
> WDYT?

I'm not sure I understand what you meant here. Did you want me to add the line numbers to the ItemsValidator in _this_ PR #4928?

lolodomo:
> I'm not sure I understand what you meant here. Did you want me to add the line numbers to the ItemsValidator in _this_ PR #4928?

Yes, that would be great

jimtng:
I think it would be better to do it all in one PR, and not in this one.

lolodomo:
@openhab/core-maintainers : for your information, this PR looks very good to me and ready for a merge.

lolodomo:
@jimtng : can you please resolve conflicts?

lolodomo:
@jimtng : Kai just merged PR #4793 (and that is a very good news for our pending stuff in UI)  and that means you will have to adjust your PR carefully.
I will check it again after you solve the conflicts.

lolodomo:
Build failed, you need to apply spotless in /org/openhab/core/model/item

kaikreuzer (approved):
Thanks!

